PR #1, opened 2019-02-05
add instructions for running against minishift

dhellmann:
Update the README with instructions for setting up the controller in a
dev environment using minishift.

Update the role binding YAML file to allow a developer to run the
controller from outside the cluster. We will likely want to do
something different when we get to really packaging this for
deployment.

Signed-off-by: Doug Hellmann <doug@doughellmann.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,2 +1,46 @@
 Bare Metal Operator
 ===================
+
+## Setup with minishift
+
+1. Install and launch minishift
+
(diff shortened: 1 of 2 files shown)


hardys:
Not tested but lgtm :)

russellb:
Similar instructions could be provided for how to try this with vanilla Kube using minikube.

If OpenShift is the dev env, I’d rather point to OpenShift 4, using the libvirt backend.

PR #2, opened 2019-02-05
Start operator

dhellmann:
This series of commits starts sketching in parts of the BareMetalHost operator like the spec and status fields as well as some minimal logic for handling deletes and demonstrating basic error reporting.


deploy/crds/metalkube_v1alpha1_baremetalhost_cr.yaml
```diff
@@ -4,4 +4,7 @@ metadata:
   name: example-baremetalhost
 spec:
   # Add fields here
-  size: 3
+  bmc:
```
russellb: probably need a "type" field somewhere in the future.  "ipmi" vs "idrac" etc

russellb: Btw, as this evolves, can you update the docs repo too?  There's a section at the bottom about the API.

https://github.com/metalkube/metalkube-docs

PR #3, opened 2019-02-06
Add test framework

dhellmann:
This pulls in the code from #2 as well so we should probably not merge this as-is.


dhellmann:
Rebased after #2 merged so it should be clean now

PR #6, opened 2019-02-08
Rework reconcile loop and add hardware-profile label

dhellmann:


pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -141,54 +138,91 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (reconcile
...
 			}
+			return reconcile.Result{Requeue: true}, nil
+		}
+	}
+
+	// Set the hardware profile name.
+	//
+	// FIXME(dhellmann): This should pull data from Ironic and compare
+	// it against known profiles.
+	if instance.SetLabel(metalkubev1alpha1.HardwareProfileLabel, "unknown") {
```
russellb: maybe we just let this get set manually in the meantime?

dhellmann: Yeah, I thought about that. We're going to have at least 2 profiles to start (libvirt and hardware). Before I can do much with profile matching, I want to see what data Ironic returns from a scan, so I was just putting it off for now.

PR #7, opened 2019-02-08
Manage BMC credentials as secrets

dhellmann:

PR #9, opened 2019-02-11
Watch secrets

dhellmann:


dhellmann:
This addresses a comment left on #7

mhrivnak:
Just to check in on the goal, why is it that we want to trigger a reconcile for a BareMetalHost when the secret changes? What action do we want to take in response to that secret changing?

Would it be sufficient to instead just retrieve the current value of the secret at whatever time the credentials are needed?

dhellmann:
@mhrivnak if the secret was previously invalid (missing a value or incorrect) we need to try again to reconcile the state of the host. Similarly, if it was valid and is now invalid, we would want to reflect that error status.

mhrivnak:
Makes sense. The reconcile function will need to know when the secret has changed so it can try the credentials, and not try them on every reconcile. Perhaps storing the version of the secret in the status of the CR, when an auth succeeds, would solve that. If that `CredsSuccessVersion` field (or whatever you want to name it) is empty or has a different value than the current secret's actual version, the Reconcile will know to re-test.

dhellmann:
Good point. I'll look at adding that field to the status block.

mhrivnak:
Another question: Just so I understand, what was the reasoning to go with a one-to-many relationship of secret to host? Do we expect that to be the most common case where many hosts use the same credentials? Is that something the UX could potentially handle for us instead of modeling it this way? How and when will the secret get created and deleted? When a new host gets discovered (either automatically or through a manual process), how will it get associated with a credential secret?

I only ask because our lives get easier (within the controller at least) if we model it as one Secret per Host. The Secret would have an owner reference to the Host, and then its lifecycle is driven 100% on the Host CR's lifecycle. No need to worry about identifying and cleaning up an orphaned Secret for example. This is a very common use case, where a little bit of data in a CR is sensitive and needs to be stored in a subordinate Secret. A change to the Secret can then directly queue an event for the owning CR without a need for the fan-out logic.

There's of course nothing wrong with one-to-many if that's the best fit for the data and how it will be used. But we'll need to manage the lifecycle of the Secrets separately, identifying orphans, doing the event fan-out you're implementing here, and make sure the UX tooling understands how to associate and disassociate a credentials secret from hosts.

dhellmann:
Good questions about ownership and 1:1 vs. 1:n. I don't know. I expect the installer to create 1 secret per host, since that's going to be easier than figuring out if the user has entered the same values. The in-cluster UI could also do that. If the user is using the API directly, though, how would we guarantee that? Would we check to see if the secret they give us is already being used, and refuse to reuse it?

I see 2 version-related fields in the ObjectMeta struct. Should I be using the ResourceVersion or the Generation value? Or maybe there's a method I should call instead?

mhrivnak:
If we expect one secret per host, that can be a lot easier then. I suggest we go with a 1:1 modeling.

The Reconcile function should ensue that the Secret that corresponds to a Host has an owner reference to the Host. If it finds an owner reference to some other Host, that would be an error that it should both log and reflect in the Status of the Host. If it finds no owner reference, it should add one.

Your Watch can just use the EnqueueRequestForOwner handler and not need any of the fan-out logic.

It's common to establish a naming convention in this sort of case where the name of the secret is derived from the name of the host. For example, you've probably noticed that Pod names are derived from the name of the Deployment and ReplicaSet that created them. (a Deployment with name "wordpress" ends up with Pods named things like "wordpress-abcdef") We could reasonably establish a hard rule here that the Secret should have the same name as the Host, or something like that, which will help ensure correctness.

To further guarantee correctness, we could create a validating web hook to actually prevent an API user from establishing an invalid Host->Secret relationship. That's the standard way to do validation in k8s.

dhellmann:
So API users would need to create the secret with a specific name, and then refer to that secret within the BMC block of the BareMetalHost as well? Or we would use the naming convention to establish the relationship?

mhrivnak:
> I see 2 version-related fields in the ObjectMeta struct. Should I be using the ResourceVersion or the Generation value? Or maybe there's a method I should call instead?

ResourceVersion is what we want to use.

mhrivnak:
> So API users would need to create the secret with a specific name, and then refer to that secret within the BMC block of the BareMetalHost as well? Or we would use the naming convention to establish the relationship?

Looking at a few examples, it seems that the convention is to store the name of the secret on the owning object and not enforce a particular name. See the uses here for example:

https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#secretreference-v1-core

dhellmann:
Is this the webhook you mean? https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#admission-webhooks

mhrivnak:
> Is this the webhook you mean? https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#admission-webhooks

That's it.

dhellmann:
Do you have any examples of other controllers implementing web hooks like that? Maybe it's less work than the tutorial makes it seem.

mhrivnak:
> Do you have any examples of other controllers implementing web hooks like that? Maybe it's less work than the tutorial makes it seem.

I don't know of a good example, but we might find one by asking in slack. Here are some docs on controller-runtime's support for webhooks: https://book.kubebuilder.io/beyond_basics/what_is_a_webhook.html

Ultimately the API is very simple, even though the docs drag on about details that make it seem more complicated than it really is. After the boilerplate is taken care of, you'll just receive an [AdmissionRequest](https://godoc.org/k8s.io/api/admission/v1beta1#AdmissionRequest) and return an [AdmissionResponse](https://godoc.org/k8s.io/api/admission/v1beta1#AdmissionResponse).

I wouldn't prioritize making one for now, as it would be an easy task to add one later. But I can appreciate wanting to understand them now and how they'll fit in. Let me know what other questions you have.

dhellmann:
I've added another change to the end of this PR to make the hosts own their credentials and assume a 1:1 relationship.

dhellmann:
I guess the "complexity" I was seeing with the web hook had to do with setting up the SSL endpoint  giving it credentials, and deploying it in a separate pod. Maybe that's all part of the boilerplate you were talking about, though. I wonder if it's possible to just put that hook inside the operator, for example? If that's the preferred way of doing API validation, it would be cool if the operator-sdk had a mode to generate the stub for the hook.

dhellmann:
I think I'm done tweaking this and adding tests, and it's ready for another review.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -53,6 +55,16 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
+	// Watch for changes to secrets being used by hosts
+	err = c.Watch(&source.Kind{Type: &v1.Secret{}},
+		&handler.EnqueueRequestForOwner{
+			IsController: true,
+			OwnerType:    &metalkubev1alpha1.BareMetalHost{},
+		})
+	if err != nil {
```
mhrivnak: As a nit, I would make this line `return err` and remove the remaining lines in the function.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -288,7 +324,6 @@ func (r *ReconcileBareMetalHost) getBMCCredentials(request reconcile.Request, in
 	if secretKey.Namespace == "" {
```
mhrivnak: I think we should require that the secret be in the same namespace as the host object. Aside from being the likely case and a sensible way to keep related data together, OwnerRefs cannot traverse namespaces.

dhellmann: Should I not use a reference then, and just ask for the secret name in the spec?

pkg/apis/metalkube/v1alpha1/baremetalhost_types.go
```diff
@@ -124,6 +134,28 @@ func (host *BareMetalHost) SetOperationalStatus(status string) bool {
...
 }
 
+// compare the secret with the last one known to work and report if
+// they are the same
+func (host *BareMetalHost) CredentialsHaveChanged(currentSecret corev1.Secret) bool {
+	currentRef := host.Status.GoodCredentials.Reference
+	currentVersion := host.Status.GoodCredentials.Version
+	newRef := host.Spec.BMC.Credentials
+
+	if currentRef == nil {
```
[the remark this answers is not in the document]
dhellmann: Done

pkg/apis/metalkube/v1alpha1/baremetalhost_types.go
```diff
@@ -124,6 +134,28 @@ func (host *BareMetalHost) SetOperationalStatus(status string) bool {
 	return host.SetLabel(OperationalStatusLabel, status)
 }
 
+// compare the secret with the last one known to work and report if
+// they are the same
+func (host *BareMetalHost) CredentialsHaveChanged(currentSecret corev1.Secret) bool {
```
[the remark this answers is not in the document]
dhellmann: Done

pkg/apis/metalkube/v1alpha1/baremetalhost_types.go
```diff
@@ -124,6 +134,28 @@ func (host *BareMetalHost) SetOperationalStatus(status string) bool {
 	return host.SetLabel(OperationalStatusLabel, status)
 }
 
+// compare the secret with the last one known to work and report if
```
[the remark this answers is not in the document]
dhellmann: Done

pkg/controller/baremetalhost/baremetalhost_controller.go
dhellmann: Done

pkg/controller/baremetalhost/baremetalhost_controller.go
mhrivnak: I think just the name would be fine, and it would match what I've seen other operators do.

dhellmann: Done.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -53,7 +53,13 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
 	}
 
-	return nil
+	// Watch for changes to secrets being used by hosts
+	err = c.Watch(&source.Kind{Type: &v1.Secret{}},
+		&handler.EnqueueRequestForOwner{
+			IsController: true,
+			OwnerType:    &metalkubev1alpha1.BareMetalHost{},
+		})
+	return err
```
russellb: If I'm reading this right, it looks like we watch for updates to Secrets owned by a BareMetalHost.  That ownership is set by the BareMetalHost reconciler when it matches them together.

I'm imagining the case where a BareMetalHost is created and refers to a Secret that does not yet exist.  Later, that Secret gets created.  What will ensure that the BMH reconciler runs again to pick up this secret?

or am I misinterpreting the second argument to Watch() ?

russellb:
I think we should merge fast and do follow-ups at this stage, so I'm going to just merge.

pkg/controller/baremetalhost/baremetalhost_controller.go
dhellmann: If the host has no secret or the secret can't be loaded Reconcile() exits with an error and that triggers reconciliation again. That will just keep happening until the secret is there. If we have a secret, and its contents are invalid, then the host is marked with an error and the reconciliation loop stops until the secret reference is changed or the contents of the secret are changed.

PR #10, opened 2019-02-11
API docs

dhellmann:


dhellmann:
I realized I left out the label information, so I've added that.

PR #15, opened 2019-02-19
add a method to host to retrieve the operational status

dhellmann:
Instead of expecting users of the BareMetalHost to understand how to
fetch the status from the label, add a method.

### DIFF
--- a/pkg/apis/metalkube/v1alpha1/baremetalhost_types.go
+++ b/pkg/apis/metalkube/v1alpha1/baremetalhost_types.go
@@ -167,15 +167,34 @@ func (host *BareMetalHost) SetLabel(name, value string) bool {
 	return false
 }
 
+// getLabel returns the value associated with the given label. If
+// there is no value, an empty string is returned.
+func (host *BareMetalHost) getLabel(name string) string {
+	if host.Labels == nil {
+		return ""
+	}
+	return host.Labels[name]
+}
+
 // SetOperationalStatus updates the OperationalStatusLabel and returns
 // true when a change is made or false when no change is made.
 func (host *BareMetalHost) SetOperationalStatus(status string) bool {
 	return host.SetLabel(OperationalStatusLabel, status)
 }
 
-// GetCredentialsKey returns a NamespacedName suitable for loading the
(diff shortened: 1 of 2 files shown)


pkg/apis/metalkube/v1alpha1/baremetalhost_types.go
```diff
@@ -167,12 +167,31 @@ func (host *BareMetalHost) SetLabel(name, value string) bool {
 	return false
 }
 
+// GetLabel returns the value associated with the given label. If
+// there is no value, an empty string is returned.
+func (host *BareMetalHost) GetLabel(name string) string {
```
russellb: I was just reading the Methods chapter in the Go programming language book.  The last page of that chapter (chapter 6) talks about method naming conventions.  It implied that the typical naming convention is to omit the "Get" prefix.

"However, when naming a getter method, we usually omit the Get prefix.  This preference for brevity extends to all methods, not just field accessors, and to other redundant prefixes as well, such as Fetch, Find, and Lookup."

I don't have a strong personal preference but based on the above, I'd remove "Get" from these two new methods.

dhellmann: Ugh, I keep forgetting that. Methods are the verbs. They should *have* verbs in the names. I'll change it, though.

dhellmann: I fixed these and updated some other docstrings and methods to be consistent.

russellb: One of the new methods still has "get" in the name.  Intentional?

dhellmann: I'm concerned that "label" can be seen as both a verb and a noun, and since the method fetches a label and does not set one, I thought being explict about using "get" would be good. I made it a private method, because it really should exist on the ObjectMeta structure rather than the host object. I should probably make a similar change to SetLabel().

russellb: OK, thanks

PR #22, opened 2019-02-22
add step to deploy the new image

dhellmann:

### DIFF
--- a/docs/publishing-images.md
+++ b/docs/publishing-images.md
@@ -59,3 +59,15 @@ your development fork.
       branch at the end, like this:
 
           quay.io/dhellmann/baremetal-operator:update-operator-deployment
+
+6. Launch the deployment by applying the new file.
+
+   1. Make sure you have run the [setup steps](dev-setup.md) to set up
+      the service account, role, and mapping.
+   2. Apply the new deployment:
+
+       oc apply -f deploy/dev-operator.yaml
+
+To monitor the operator, use `oc get pods` to find the pod name for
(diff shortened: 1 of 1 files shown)


docs/publishing-images.md
```diff
@@ -59,3 +59,15 @@ your development fork.
...
 
           quay.io/dhellmann/baremetal-operator:update-operator-deployment
+
+6. Launch the deployment by applying the new file.
+
+   1. Make sure you have run the [setup steps](dev-setup.md) to set up
+      the service account, role, and mapping.
+   2. Apply the new deployment:
+
+       oc apply -f deploy/dev-operator.yaml
```
russellb: I don't see dev-operator.yaml in the tree?  Should this instruct how to create it?  I imagine it's a copy of deploy/operator.yaml, but pointing to the custom image?

Also, should this mention dealing with a workflow where you've already got the operator running and now want to run a new version?

dhellmann: The file is created in an earlier step. I didn't want to check mine in, because it refers to my image, and I would expect each individual developer to have their own. I could set up an ignore pattern, I guess.

Yes, this should talk about how to deal with that "switch over" situation. I haven't encountered that, so I'm not sure what I would do. Delete the existing deployment, I guess?

russellb: Oh, so I should read the step immediately before this one?  :-p

Yeah, delete the deployment, or maybe scale it down and back up?  I'm not sure.  :-)

We can just merge this for now.

PR #29, opened 2019-02-25
WIP: start sketching in a provisioner api

dhellmann:
Start defining an API for managing the more complex provisioning
steps.

### DIFF
--- a/pkg/provisioning/provisioning.go
+++ b/pkg/provisioning/provisioning.go
@@ -1,0 +1,70 @@
+package provisioning
+
+import (
+	"github.com/metalkube/baremetal-operator/pkg/bmc"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+)
+
+var log = logf.Log.WithName("provisioner")
+
+/*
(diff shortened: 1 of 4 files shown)


dhellmann:
I'm not sure I like the direction this is going. Having yet another representation of the host (and having to come up with a name for it) feels awkward, but I was trying to avoid tying the library to the BareMetalHost class. Maybe it's cleaner to just use that?

pkg/provisioning/provisioning.go
```diff
@@ -0,0 +1,86 @@
...
+var log = logf.Log.WithName("provisioner")
+
+/*
+Package provisioning defines the API for talking to the provisioning backend.
+*/
+
+// Provisioner holds the state information for talking to the
+// provisioning backend.
+//
+// NOTE(dhellmann): Provisioner will eventually become an interface,
```
mhrivnak: Using an interface makes testing a lot easier, so I would suggest doing that now. It's a common approach for enabling unit tests of hard API boundaries like this.

dhellmann:
Let's go with the approach in #30 for now.

PR #30, opened 2019-02-25
Provisioning api with hosts

dhellmann:
This is an alternative to #29 that uses the existing BareMetalHost type instead of defining a new Target type.


dhellmann:
This needs tests, but I think it demonstrates the direction we want to go.

dhellmann:
This needs tests, but is ready for a preliminary review.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -214,6 +241,40 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (reconcile
...
 	}
 
+	// Set the hardware profile name.
+	if host.Status.HardwareDetails == nil {
+		if dirty, err = r.provisioner.InspectHardware(host); err != nil {
+			reqLogger.Error(err, "failed to inspect hardware")
+			return reconcile.Result{}, err
+		}
+		if dirty {
+			reqLogger.Info("saving status after inspecting hardware")
```
mhrivnak: It looks like this saves the host, not the status.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -214,6 +241,40 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (reconcile
...
+			if err := r.client.Update(context.TODO(), host); err != nil {
+				reqLogger.Error(err, "failed to save host during inspection")
+				return reconcile.Result{}, err
+			}
+			if host.Status.HardwareDetails != nil {
+				reqLogger.Info("saving hardware details")
```
mhrivnak: I suggest using similar language to the error that comes two lines below, so it would be clear to someone looking in a log that this is the operation that failed. "Doing X" ... "Failed to do X"

pkg/provisioning/provisioning.go
```diff
@@ -0,0 +1,98 @@
...
+var log = logf.Log.WithName("provisioner")
+
+/*
+Package provisioning defines the API for talking to the provisioning backend.
+*/
+
+// Provisioner holds the state information for talking to the
+// provisioning backend.
+//
+// NOTE(dhellmann): Provisioner will eventually become an interface,
```
mhrivnak: As noted on the other PR, I think we'll want the interface as soon as we write any tests.

pkg/provisioning/provisioning.go
```diff
@@ -0,0 +1,98 @@
...
+	if host.Status.HardwareDetails == nil {
+		reqLogger.Info("continuing inspection by setting details")
+		host.Status.HardwareDetails = &metalkubev1alpha1.HardwareDetails{}
+		return true, nil
+	}
+
+	return false, nil
+}
+
+// Deprovision cleans the host and removes it from the external
```
mhrivnak: I noticed earlier that you named the function `Remove`, and I see here from the comment that I assume you also considered `Deprovision`. I sort of like `Deprovision` when applying the verb to a host, but could go either way. I guess `provisioner.Remove()` implies that the provisioner is removing whatever it had previously added? 

In any case, either name is fine, but this comment should match.

mhrivnak: IMO stating that this function "removes it from the external provisioning tool" reveals too much about the implementation details of a particular provisioner. I suggest we stick to describing what changes will be made to the host itself when calling each function.

pkg/apis/metalkube/v1alpha1/baremetalhost_types.go
```diff
@@ -134,6 +141,14 @@ type BareMetalHostStatus struct {
 	ErrorMessage string `json:"errorMessage"`
 }
 
+// ProvisionStatus holds the state information for a single target.
+type ProvisionStatus struct {
+	// FIXME(dhellmann): This should be an enum of some sort.
+	State string `json:"state"`
+	// UUID in ironic
+	ID string `json:"ID"`
```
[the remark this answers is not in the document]
dhellmann: It's a unique identifier that can be used to ensure we're manipulating the right object when we talk to Ironic later.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -130,10 +142,17 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (reconcile
...
+		if dirty, err = r.provisioner.Remove(host); err != nil {
+			reqLogger.Error(err, "failed to deprovision")
+			return reconcile.Result{}, err
+		}
+		if dirty {
+			if err := r.saveStatus(host); err != nil {
+				reqLogger.Error(err, "failed to clear host status on deprovision")
+				return reconcile.Result{}, err
+			}
+			return reconcile.Result{Requeue: true}, nil
```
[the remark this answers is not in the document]
dhellmann: I'm anticipating Remove() taking several steps, and us wanting to wait for it if we are wiping the server.

pkg/controller/baremetalhost/baremetalhost_controller.go
dhellmann: ah, yeah, some of these messages got moved around. I'll clean all of that up.

pkg/provisioning/provisioning.go
dhellmann: I waffled on the name. I'll switch it back to Deprovision().

pkg/apis/metalkube/v1alpha1/baremetalhost_types.go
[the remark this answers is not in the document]
dhellmann: I expanded the description of the field a bit.

pkg/controller/baremetalhost/baremetalhost_controller.go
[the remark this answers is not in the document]
dhellmann: I added a comment to explain this expectation.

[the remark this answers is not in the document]
mhrivnak: Gotcha. Probably then we'll want to use the `RequeueAfter` option so that there is some delay before reconcile happens again. Is there any way for the operator to estimate how long it will need to wait? If not, we can just select a sensible polling interval.

dhellmann:
I think this is ready to go. I want to defer the work to set up the interface to a separate PR, where I will also work on making unit tests that run faster than the current tests do.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -25,6 +27,15 @@ import (
...
 var log = logf.Log.WithName("controller_baremetalhost")
 
+// DeprovisionRequeueDelay controls the amount of time the controller
+// waits between attempts to determine if the deprovisioning operation
+// has been completed.
+//
+// FIXME(dhellmann): This is public for now so we can change it in the
+// test suite. When we figure out a better unit test setup we can make
+// it private.
+var DeprovisionRequeueDelay = time.Second * 10
```
mhrivnak: I think this would be a reasonable attribute of `ReconcileBareMetalHost`. Feel free to do that now or later as you prefer.

dhellmann: good call

dhellmann: I'm not sure how I'd be able to get to the reconciler being used by the current tests until after the conversion to unit tests, so I'll hold off on that

PR #46, opened 2019-03-05
Add test provisioner

dhellmann:
This split is a prerequisite to making the ironic provisioner really depend on having ironic present.

### DIFF
--- a/pkg/provisioner/fixture/fixture.go
+++ b/pkg/provisioner/fixture/fixture.go
@@ -19,7 +19,7 @@ var deprovisionRequeueDelay = time.Second * 10
 type provisionerFactory struct{}
 
 // New returns a new test ProvisionerFactory
-func New() provisioner.ProvisionerFactory {
+func NewFactory() provisioner.ProvisionerFactory {
 	return &provisionerFactory{}
 }
 
--- a/pkg/provisioner/fixture/fixture.go
+++ b/pkg/provisioner/fixture/fixture.go
@@ -1,0 +1,181 @@
+package fixture
+
+import (
+	"time"
+
+	"github.com/pkg/errors"
+
+	"github.com/go-logr/logr"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+
+	metalkubev1alpha1 "github.com/metalkube/baremetal-operator/pkg/apis/metalkube/v1alpha1"
+	"github.com/metalkube/baremetal-operator/pkg/provisioner"
+)
+
+var log = logf.Log.WithName("ironic")
+var deprovisionRequeueDelay = time.Second * 10
+
+// Test fixture provisioner factory
+type provisionerFactory struct{}
+
+// New returns a new test ProvisionerFactory
+func New() provisioner.ProvisionerFactory {
+	return &provisionerFactory{}
+}
+
+// Provisioner implements the provisioning.Provisioner interface
+// and uses Ironic to manage the host.
+type fixtureProvisioner struct {
+	// the host to be managed by this provisioner
+	host *metalkubev1alpha1.BareMetalHost
+	// a logger configured for this host
+	log logr.Logger
+}
+
+// New returns a new Ironic Provisioner
+func (f *provisionerFactory) New(host *metalkubev1alpha1.BareMetalHost) (provisioner.Provisioner, error) {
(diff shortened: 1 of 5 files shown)


pkg/provisioner/provisioner.go
```diff
@@ -10,29 +10,34 @@ import (
 Package provisioning defines the API for talking to the provisioning backend.
 */
 
+type ProvisionerFactory interface {
```
mhrivnak: Consider making this a function-based type rather than an interface, since I think that's all you really need.

```
type ProvisionerFactory func(host *metalkubev1alpha1.BareMetalHost) (Provisioner, error)
```

That'll make usage a little simpler too I think. For example, `ironic.New` can be the factory function, returning an ironic provisioner. It eliminates the need to have a constructor just for your factory.

You can then do things like this:

```
	return &ReconcileBareMetalHost{
		client:             mgr.GetClient(),
		scheme:             mgr.GetScheme(),
		provisionerFactory: ironic.New,
        }
```

Let me know what you think of that approach.

dhellmann: I'm working on the ironic version of the factory, and I'm anticipating wanting to have the controller pass in configuration values like the IPA image names, ironic IP address, etc. We could do that with a function-based interface, but it just moves all of those settings to the factory's New() method and means the reconciler has to keep up with them, instead of just keeping up with the factory itself.

dhellmann: See https://github.com/dhellmann/baremetal-operator/commit/5f067534702aa3815abeb09250914084c00666f8 for an example of what I mean.

dhellmann:
I rebased this on top of Honza's changes in #47

pkg/provisioner/provisioner.go
mhrivnak: If the provisioner factory will ever need multiple methods, then an interface is the way to go. But if it will only ever have the New method, which seems likely, this function-based approach may still pay off. The above example would become:

```
return &ReconcileBareMetalHost{
		client:             mgr.GetClient(),
		scheme:             mgr.GetScheme(),
		provisionerFactory: ironic.NewFactory(x, y, z).New, // renamed to NewFactory just for clarity of the example
        }
```

or you can move the factory into a closure and do this, which is another common pattern in go code:

```
return &ReconcileBareMetalHost{
		client:             mgr.GetClient(),
		scheme:             mgr.GetScheme(),
		provisionerFactory: func(host *metalkubev1alpha1.BareMetalHost) (Provisioner, error) {
				return ironic.New(x, y, z, host)
		},
        }
```

It ends up being quite flexible, and I like that it reduces the "New depth" from `somepackage.New(config).New(host)`

Here is an example of the closure-factory pattern: https://github.com/operator-framework/operator-sdk/blob/v0.5.0/pkg/ansible/runner/runner.go#L193-L196

If you prefer to stick with the interface, as a nit-pick I suggest considering a name like `NewFactory`, since the common assumption is that if I call `thing.New()`, I'll get a new "thing", not a "thing factory".

pkg/provisioner/fixture/fixture.go
```diff
@@ -0,0 +1,181 @@
...
+// PowerOn ensures the server is powered on independently of any image
+// provisioning operation.
+func (p *fixtureProvisioner) PowerOn() (dirty bool, err error) {
+	p.log.Info("ensuring host is powered on")
+
+	if dirty, err = p.ensureExists(); err != nil {
+		return dirty, errors.Wrap(err, "could not power on host")
+	}
+
+	if p.host.Status.Provisioning.State != "powered on" {
```
mhrivnak: "powered on" and "powered off" are good candidates for a constant, presumably defined where the `State` struct is defined.

pkg/provisioner/provisioner.go
dhellmann: I've renamed the factory creation functions for now. If we want to rewrite this later to be function based, that's fine, but I don't want to spin on that this sprint.

pkg/provisioner/fixture/fixture.go
dhellmann: Yeah, that'll happen when those methods are doing something useful.

imain:
I talked to Doug in person.  This looks good to me.

PR #48, opened 2019-03-05
Create ironic nodes

dhellmann:
This depends on #46


pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -30,6 +30,15 @@ import (
 
 var log = logf.Log.WithName("controller_baremetalhost")
 
+// FIXME(dhellmann): These values should probably come from
+// configuration settings and something that can tell the IP address
+// of the ironic server.
+const (
+	instanceImageSource   = "http://172.22.0.1/images/redhat-coreos-maipo-latest.qcow2"
+	instanceImageChecksum = "97830b21ed272a3d854615beb54cf004"
```
dtantsur: Ironic has just added support for loading checksums from a URL. You may want to use it.

deploy/crds/worker-0.yaml
```diff
@@ -0,0 +1,20 @@
...
+
+---
+apiVersion: metalkube.org/v1alpha1
+kind: BareMetalHost
+metadata:
+  name: worker-0
+spec:
+  online: true
+  bmc:
+    address: ipmi://192.168.122.1:6233/?mac=00:73:49:3a:76:8e;net=provisioning
```
dtantsur: What are `mac=..` and `net=..` parts doing here? If it's for creating ports, why not have a separate field?

dhellmann: It is indeed for the port. See https://github.com/metalkube/metalkube-docs/blob/master/design/bmc-address.md for details about using the URL. So far we only need MAC and net values for libvirt, so I will eventually create a libvirt "type" but the type parsing isn't there, yet.

pkg/controller/baremetalhost/baremetalhost_controller.go
dhellmann: Is that available in the version of ironic we'll be packaging for this?

imain: Ultimately these should come from a config somewhere.. I think the values are probably ok for now, except it might randomly break if the latest image gets updated

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -132,6 +145,51 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (reconcile
...
+		Username: string(bmcCredsSecret.Data["username"]),
+		Password: string(bmcCredsSecret.Data["password"]),
+	}
+
+	// Verify that the secret contains the expected info.
+	validCreds, reason := bmcCreds.AreValid()
+	if !validCreds {
+		reqLogger.Info("invalid BMC Credentials", "reason", reason)
+		err := r.setErrorCondition(request, host, reason)
+		return reconcile.Result{}, errors.Wrap(err, "failed to set error condition")
```
imain: I don't understand this part.. is the error message correct?  Do we not want to return an error message about invalid creds?

pkg/provisioner/ironic/ironic.go
```diff
@@ -1,46 +1,318 @@
...
+					"ipmi_port":     p.bmcAccess.Port(),
+					"ipmi_username": p.bmcCreds.Username,
+					"ipmi_password": p.bmcCreds.Password,
+					"ipmi_address":  p.bmcAccess.Hostname(),
+					// FIXME(dhellmann): The names of the images are tied
+					// to the version of ironic we are using and are
+					// likely to change.
+					//
+					// FIXME(dhellmann): We need to get our IP on the
+					// provisioning network from somewhere.
```
imain: Yeah my original thought is we could just put the full address in a config file and have that be generated, but I guess we don't have a way to template things ATM?.. maybe ignition? (I don't like ignition myself tho :))

pkg/provisioner/ironic/ironic.go
```diff
@@ -1,46 +1,318 @@
...
+			ironicNode.UUID,
+			nodes.ProvisionStateOpts{
+				Target: nodes.TargetManage,
+			})
+		if changeResult.Err != nil {
+			return false, errors.Wrap(changeResult.Err,
+				"failed to change provisioning state to manage")
+		}
+	}
+
```
imain: We actually need to hit 'available' state.  Manageable just means there is enough information to manage them.  They then need to be introspected to become 'available' and then we can actually use them.  I see some 'availab'e stuff below but it doesn't look like we are verifying that it's in that state?  Maybe it's just the way the diff is layed out..

pkg/controller/baremetalhost/baremetalhost_controller.go
dhellmann: We talked about this via chat, but for other reviewers: This section stores the error state on the host object and then only throws an error if it cannot complete that save operation. The reason for that pattern is that a thrown error causes the reconciler to try again, and "bad credentials" isn't the type of error where trying again is going to do any good.

pkg/provisioner/ironic/ironic.go
dhellmann: A config file might be fine, or we could change the startup script for the container to figure it out (or even the main function of the operator process).

pkg/provisioner/ironic/ironic.go
dhellmann: Good point. For now, I'm just putting the register and delete code up for review. The changes were starting to get big, and I wanted some early validation before going any further.

deploy/crds/worker-0.yaml
dhellmann: @dtantsur, the latest patch in the pull request shows what I intend for those extra values.

pkg/bmc/access.go
```diff
@@ -0,0 +1,88 @@
...
+	} else {
+		addr.bmcType = "ipmi"
+	}
+
+	// NOTE(dhellmann): libvirt-based hosts used for dev and testing
+	// require a MAC address and network name, specified as query
+	// parameters in the URL.
+	params := parsedURL.Query()
+	if len(params["mac"]) >= 1 {
+		// FIXME(dhellmann): There is probably a safer way to
```
mhrivnak: Here it is: https://golang.org/pkg/net/url/#Values.Get

dhellmann: Done

imain:
This looks good to me.

pkg/controller/baremetalhost/baremetalhost_controller.go
hardys: I think we should be using a stein (or later) Ironic for this, but either way we shouldn't hard-code this checksum, because we exposed a file alongside the qcow2 that contains it:

https://github.com/metalkube/dev-scripts/blob/master/04_setup_ironic.sh#L76

So we should either read http://172.22.0.1/images/redhat-coreos-maipo-latest.qcow2.md5sum in the operator, or pass that URL to ironic so it can read it?

deploy/crds/worker-0.yaml
hardys: So just to clarify, the argument is that the UI will be simpler if we encode the port details in the BMC URL string?

I'm really not clear how that can be the case, since it means we'll either need some special validation, or the UI can't validate the fields at all?

The BMC connection is logically separate from the PXE provisioning mac/network so combining them seems weird to me, but perhaps I'm missing some context from previous discussions.

dhellmann:
I have updated this series of changes to move the MAC information to its own field and to deal with error handling a little more nicely.

dhellmann:
And now I have rebased it on top of Michael's recent change to add the Available() method to BMH.

pkg/controller/baremetalhost/baremetalhost_controller.go
dtantsur: We'll need IPA images from Stein (this feature is in IPA, not in Ironic itself). It's a bit tricky to figure out whether the RDO images have been updated already, but they certainly will have by Stein final.

docs/dev-setup.md
```diff
@@ -50,3 +50,24 @@ install the operator-sdk tools.
     ```
     oc apply -f deploy/crds/metalkube_v1alpha1_baremetalhost_cr.yaml
     ```
+
+## Using libvirt VMs with Ironic
+
+In order to use VMs as hosts, they need to be connected to vmbc and
```
dtantsur: nit: s/vmbc/vbmc/ (also maybe link to virtualbmc docs?)

deploy/crds/worker-0.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+---
+apiVersion: metalkube.org/v1alpha1
+kind: BareMetalHost
+metadata:
+  name: worker-0
+spec:
+  online: true
+  bmc:
+    address: libvirt://192.168.122.1:6233/
```
dtantsur: I wonder if using `libvirt` schema will make people assume we actually use `libvirt` in ironic (as opposed to using IPMI-to-libvirt).

pkg/provisioner/ironic/ironic.go
```diff
@@ -1,62 +1,313 @@
...
+	if err != nil {
+		return nil, err
+	}
+	bmcAccess, err := bmc.NewAccessDetails(host.Spec.BMC.Address)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to parse BMC address information")
+	}
+	// Ensure we have a microversion high enough to get the features
+	// we need.
+	client.Microversion = "1.50"
```
dtantsur: 1.50 is pretty random. Use 1.56 which is very likely to be Stein final? We will need it for configdrive building through Ironic (if we plan to use it).

pkg/provisioner/ironic/ironic.go
```diff
@@ -1,62 +1,313 @@
...
+
+		driverInfo := p.bmcAccess.DriverInfo(p.bmcCreds)
+		// FIXME(dhellmann): The names of the images are tied
+		// to the version of ironic we are using and are
+		// likely to change.
+		//
+		// FIXME(dhellmann): We need to get our IP on the
+		// provisioning network from somewhere.
+		driverInfo["deploy_kernel"] = "http://172.22.0.1/images/tinyipa-stable-rocky.vmlinuz"
+		driverInfo["deploy_ramdisk"] = "http://172.22.0.1/images/tinyipa-stable-rocky.gz"
```
dtantsur: We should not use tinyipa in the final version (nor rocky).

pkg/provisioner/ironic/ironic.go
```diff
@@ -1,62 +1,313 @@
...
+
+		ironicNode, err = nodes.Create(
+			p.client,
+			nodes.CreateOpts{
+				Driver:        "ipmi",
+				BootInterface: "pxe",
```
dtantsur: "ipxe" is actually a much better default. it will be calculated automatically by ironic though, so let's remove this.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1,62 +1,313 @@
...
+		if p.host.Spec.BootMACAddress != "" {
+			enable := true
+			p.log.Info("creating port for node in ironic", "MAC",
+				p.host.Spec.BootMACAddress)
+			_, err := ports.Create(
+				p.client,
+				ports.CreateOpts{
+					NodeUUID:   ironicNode.UUID,
+					Address:    p.host.Spec.BootMACAddress,
+					PXEEnabled: &enable,
```
dtantsur: If one day we want to support spine/leaf, we'll need the PhysicalNetwork field as well.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1,62 +1,313 @@
...
+				nodes.UpdateOperation{
+					Op:    nodes.AddOp,
+					Path:  "/instance_info/image_source",
+					Value: p.instanceImageSource,
+				},
+				nodes.UpdateOperation{
+					Op:    nodes.AddOp,
+					Path:  "/instance_info/image_checksum",
+					Value: p.instanceImageChecksum,
+				},
```
dtantsur: You need to support `/instance_info/root_gb` (I don't remember why we need it for whole disk images, but can investigate) and `properties/root_device` (optional, but much recommended for multi-disk machines).

pkg/provisioner/ironic/ironic.go
```diff
@@ -1,62 +1,313 @@
...
 	}
+
+	// Ensure the node is marked manageable.
+	//
+	// FIXME(dhellmann): Do we need to check other states here?
+	ironicNode, err = nodes.Get(p.client, p.status.ID).Extract()
+	if err != nil {
+		return false, errors.Wrap(err, "failed to get provisioning state in ironic")
+	}
+	if ironicNode.ProvisionState == string(nodes.Enroll) && ironicNode.TargetProvisionState != nodes.Manageable {
```
dtantsur: I don't think `provision_state == enroll && target_provision_state == manageable` can ever happen, so the second part of the check is redundant. The transient state after `enroll` is `verifying`.

pkg/controller/baremetalhost/baremetalhost_controller.go
dhellmann: I agree we should update the images. I will open a separate ticket to track that.

docs/dev-setup.md
dhellmann: Done

deploy/crds/worker-0.yaml
dhellmann: Only developers will ever see it. I think we're OK, at least for now.

pkg/provisioner/ironic/ironic.go
dhellmann: I made a note in the ticket about updating to stein to include updating the microversion.

pkg/provisioner/ironic/ironic.go
dhellmann: See above

pkg/provisioner/ironic/ironic.go
dhellmann: I changed the default to ipxe. I think we want to keep explicit values for as much as we can to avoid unexpected behaviors.

pkg/provisioner/ironic/ironic.go
dhellmann: Let's address that when we get there.

pkg/provisioner/ironic/ironic.go
dhellmann: What values should be used for those?

pkg/provisioner/ironic/ironic.go

pkg/provisioner/ironic/ironic.go
dtantsur: Good question. I went through the sources yesteday, and I'm pretty sure this value should NOT be required for whole-disk images. I'm going to report and fix it today, but meanwhile you can put any positive integer. Say, 10.

dtantsur: Ehhhm, I was talking about `root_gb`. The `root_device` should either come from a user or be detected using introspection data. We don't have a good default.

pkg/provisioner/ironic/ironic.go
dtantsur: ++

pkg/provisioner/ironic/ironic.go
dtantsur: `root_gb`: https://storyboard.openstack.org/#!/story/2005165

dhellmann:
I'm going to go ahead and merge this. I'm hoping to get this repo into a state where it's possible to iterate with small changes, so I will address any other comments with other patches.

Issue #31, opened 2019-02-26
Determine primary source of readiness for actuator to consume


russellb:
The baremetal-operator should provide a simple way for the Machine actuator to determine whether a given BareMetalHost is in a state ready to be provisioned.  This will be used in selection criteria in the Machine actuator.

russellb:
This can be a Ready() method on the BareMetalHost, for example

russellb:
The interface was implemented.  It will be tweaked over time but we can call this done.

PR #49, opened 2019-03-05
adds Available() method stub to BareMetalHost

mhrivnak:
re #31

### DIFF
--- a/pkg/apis/metalkube/v1alpha1/baremetalhost_types.go
+++ b/pkg/apis/metalkube/v1alpha1/baremetalhost_types.go
@@ -165,6 +165,16 @@ type BareMetalHost struct {
 	Status BareMetalHostStatus `json:"status,omitempty"`
 }
 
+// Available returns true if the host is available to be provisioned.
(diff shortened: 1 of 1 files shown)


pkg/apis/metalkube/v1alpha1/baremetalhost_types.go
```diff
@@ -165,6 +165,16 @@ type BareMetalHost struct {
 	Status BareMetalHostStatus `json:"status,omitempty"`
 }
 
+// Available returns true if the host is available to be provisioned.
+func (host *BareMetalHost) Available() (bool, error) {
```
dhellmann: What kinds of cases are you thinking might return an error? The response will be based on data in the status of the object, right?

mhrivnak: Yep, good point. We don't want outside code initiating any kind of active inspection, so I'll get rid of the error.

dhellmann:
This could use a unit test, but we can add that in another PR.

PR #50, opened 2019-03-07
Add fields to NIC and Storage types

honza:
Opening this to start a discussion: are these the right fields?

I was going by the wireframes which are in turn based on the kubevirt CRD.

Once we are happy with the fields, I will also update the defaults.

### DIFF
--- a/pkg/apis/metalkube/v1alpha1/baremetalhost_types.go
+++ b/pkg/apis/metalkube/v1alpha1/baremetalhost_types.go
@@ -99,8 +99,8 @@ type Storage struct {
 	// The size of the disk in gigabyte
 	SizeGiB int `json:"sizeGiB"`
 
-	// Storage class, e.g. "Storage Class 1"
-	Class string `json:"class"`
+	// Hardware model
+	Model string `json:"model"`
 }
 
 // NIC describes one network interface on the host.
--- a/pkg/apis/metalkube/v1alpha1/baremetalhost_types.go
+++ b/pkg/apis/metalkube/v1alpha1/baremetalhost_types.go
@@ -90,20 +90,38 @@ type CPU struct {
 
 // Storage describes one storage device (disk, SSD, etc.) on the host.
 type Storage struct {
-	Name      string `json:"name"`
+	// A name for the disk, e.g. "disk 1 (boot)"
+	Name string `json:"name"`
+
+	// Interface
 	Interface string `json:"interface"`
-	SizeGiB   int    `json:"sizeGiB"`
-	Class     string `json:"class"`
+
+	// The size of the disk in gigabyte
+	SizeGiB int `json:"sizeGiB"`
+
+	// Storage class, e.g. "Storage Class 1"
+	Class string `json:"class"`
 }
 
 // NIC describes one network interface on the host.
 type NIC struct {
-	Name      string `json:"name"`
-	Model     string `json:"model"`
-	Network   string `json:"network"`
-	MAC       string `json:"mac"`
-	IP        string `json:"ip"`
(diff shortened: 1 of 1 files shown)


pkg/apis/metalkube/v1alpha1/baremetalhost_types.go
```diff
@@ -90,12 +90,17 @@ type CPU struct {
 
 // Storage describes one storage device (disk, SSD, etc.) on the host.
 type Storage struct {
-	SizeGiB int    `json:"sizeGiB"`
-	Info    string `json:"info"`
+	Name      string `json:"name"`
+	Interface string `json:"interface"`
+	SizeGiB   int    `json:"sizeGiB"`
+	Class     string `json:"class"`
```
dhellmann: What is "class"? Where will that come from? Is it derived from information we can see when we inspect the hardware on the host?

pkg/apis/metalkube/v1alpha1/baremetalhost_types.go
```diff
@@ -90,12 +90,17 @@ type CPU struct {
...
 }
 
 // NIC describes one network interface on the host.
 type NIC struct {
+	Name      string `json:"name"`
+	Model     string `json:"model"`
+	Network   string `json:"network"`
```
dhellmann: Is network the CIDR?

pkg/apis/metalkube/v1alpha1/baremetalhost_types.go
honza: Storage class; it seemed redundant to call it `Storage.StorageClass`.

pkg/apis/metalkube/v1alpha1/baremetalhost_types.go
honza: The wireframes have a value of "Pod Networking".

pkg/apis/metalkube/v1alpha1/baremetalhost_types.go
dhellmann: Sure, the name is fine. I just don't know where the value comes from. :-)

dhellmann:
I think it would help if each of the fields included a comment about what type of data we expect it to hold and where we expect to find that data.

julienlim:
@honza 
For the disk, do we know what type of disk it is?  I see that we know if it's HDD by the "rotational" (true).  Can we tell if it's SSD or NVMe, or do we figure this out by parsing "name"?

Here's what @dhellmann shared as an example of the discovery data we get from ironic about a disk:
{
       "size": 599550590976,
       "serial": "6b083fe0d2d0eb001bd2cc7e0fc68486",
       "wwn": "0x6b083fe0d2d0eb00",
       "rotational": true,
       "vendor": "DELL",
       "name": "/dev/sda",
       "wwn_vendor_extension": "0x1bd2cc7e0fc68486",
       "hctl": "0:2:0:0",
       "wwn_with_extension": "0x6b083fe0d2d0eb001bd2cc7e0fc68486",
       "by_path": "/dev/disk/by-path/pci-0000:01:00.0-scsi-0:2:0:0",
       "model": "PERC H710"
     }

For the details (when we show serial number, it'd be nice to also things like the following for SSDs if possible):
- remaining rated write endurance
- power on hours

andybraren:
Here's the very latest Disks table screen based on our discussion.

3-3-host-disks-errors

dhellmann:
The introspection tool doesn't gather those stats. I don't know if Ceph could or not.

julienlim:
@dhellmann ceph-volume doesn't provide them today, and also not sure if Ceph can get those stats either.  

@jolmomar@redhat.com @pcuzner@redhat.com

PR #55, opened 2019-03-09
add result type for provisioner API responses

dhellmann:
Update the Provisioner API so that all of the potentially multi-step
operations return the same Result type with instructions for the
caller about whether to store the host and how long to wait before
retrying the call.


dhellmann:
rebased

PR #56, opened 2019-03-09
unit tests for bmc package

dhellmann:


pkg/bmc/access_test.go
```diff
@@ -0,0 +1,127 @@
...
+	if acc.hostname != "192.168.122.1" {
+		t.Fatalf("unexpected hostname: %q", acc.hostname)
+	}
+	if acc.portNum != "6233" {
+		t.Fatalf("unexpected port: %q", acc.portNum)
+	}
+}
+
+func TestParseHostPortIPv6(t *testing.T) {
+	acc, err := NewAccessDetails("[fe80::fc33:62ff:fe83:8a76]:6233")
```
mhrivnak: :100:

pkg/bmc/access_test.go
```diff
@@ -0,0 +1,127 @@
...
+func TestParseHost(t *testing.T) {
+	acc, err := NewAccessDetails("192.168.122.1")
+	if err != nil {
+		t.Fatalf("unexpected parse error: %v", err)
+	}
+	if acc.Type != "ipmi" {
+		t.Fatalf("unexpected type: %q", acc.Type)
+	}
+	if acc.hostname != "192.168.122.1" {
+		t.Fatalf("unexpected hostname: %q", acc.hostname)
```
mhrivnak: Is it worth having a test or two that use an actual hostname instead of using only IP addresses? Might not be our use case, but was just a thought given the field name.

pkg/bmc/access.go
```diff
@@ -19,26 +21,57 @@ type AccessDetails struct {
...
-		hostname: parsedURL.Hostname(),
+		// We failed to parse the URL, but it may just be a host or
+		// host:port string (which the URL parser rejects because ":"
+		// is not allowed in the first segment of a
+		// path. Unfortunately there is no error class to represent
+		// that specific error, so we have to guess.
+		if strings.Contains(address, ":") {
+			// If we can parse host:port, carry on with those
+			// values. Otherwise, report the original parser error.
+			var err2 error
```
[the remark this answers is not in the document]
dhellmann: Done

pkg/bmc/access_test.go
dhellmann: I added one for the default case where that's all they enter. I assume the golang URL and host:port parsing routines can cope with hostnames.

PR #57, opened 2019-03-10
create AccessDetails interface

dhellmann:
Rather than having methods with lots of if-then statements, define an
interface so we can have multiple implementations and just select the
right one in NewAccessDetails().

### DIFF
--- a/pkg/bmc/ipmi.go
+++ b/pkg/bmc/ipmi.go
@@ -1,0 +1,43 @@
+package bmc
+
+type ipmiAccessDetails struct {
+	bmcType  string
+	portNum  string
+	hostname string
+}
+
+const ipmiDefaultPort = "623"
+
+func (a *ipmiAccessDetails) Type() string {
+	return a.bmcType
+}
+
+// NeedsMAC returns true when the host is going to need a separate
+// port created rather than having it discovered.
+func (a *ipmiAccessDetails) NeedsMAC() bool {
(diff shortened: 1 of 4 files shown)


pkg/bmc/ipmi.go
```diff
@@ -0,0 +1,37 @@
...
+	return a.bmcType == "libvirt"
+}
+
+// DriverInfo returns a data structure to pass as the DriverInfo
+// parameter when creating a node in Ironic. The structure is
+// pre-populated with the access information, and the caller is
+// expected to add any other information that might be needed (such as
+// the kernel and ramdisk locations).
+func (a *ipmiAccessDetails) DriverInfo(bmcCreds Credentials) map[string]interface{} {
+	return map[string]interface{}{
```
mhrivnak: Do the contents change sufficiently among implementations that we cannot make a struct for this return value, and likewise for the interface?

dhellmann: Not really. These APIs are very "anything goes." The Ironic REST API expects a mapping as argument, where the values may be strings or other nested objects and the keys depend on the driver. To cope with the Ironic REST API, the gophercloud API takes a map[string]interface as argument, so I figured returning that here would make it easier for the caller to use it.

mhrivnak: :+1:

dhellmann:
Rebased and tweaked a bit

PR #61, opened 2019-03-12
turn off power management sections of reconcile

dhellmann:
We may not need it at all and for now it interferes with testing
provisioning.


russellb:
How about just drop the code instead of leaving it commented out?

dhellmann:
> How about just drop the code instead of leaving it commented out?

Done

dhellmann:
> > How about just drop the code instead of leaving it commented out?
> 
> Done

Oops, I missed the tests the first time.

PR #62, opened 2019-03-12
update test wait loop to look for delayed requeue

dhellmann:
Sometimes the result from a step is a request to requeue after a
delay. In those cases, the Requeue flag is not set but the
RequeueAfter value is non-zero.

### DIFF
--- a/pkg/controller/baremetalhost/baremetalhost_controller_test.go
+++ b/pkg/controller/baremetalhost/baremetalhost_controller_test.go
@@ -124,7 +124,8 @@ func tryReconcile(t *testing.T, r *ReconcileBareMetalHost, host *metalkubev1alph
 			break
 		}
 
-		if !result.Requeue {
+		t.Logf("result: %v", result)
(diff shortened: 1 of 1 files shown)


pkg/controller/baremetalhost/baremetalhost_controller_test.go
```diff
@@ -124,7 +124,8 @@ func tryReconcile(t *testing.T, r *ReconcileBareMetalHost, host *metalkubev1alph
 			break
 		}
 
-		if !result.Requeue {
+		t.Logf("result: %v", result)
```
russellb: It's not clear to me if this is a leftover line from debugging, or if you meant to leave it

dhellmann: I did mean to leave that, to show the result in the loop. It's only printed to the console if the test fails.

russellb: OK, thanks

PR #65, opened 2019-03-13
add -test-mode command line flag

dhellmann:
The UX team needs a way to run the operator without the full
provisioning environment. This switch lets them have a provisioner
that does properly set up the host objects, and does not require
ironic.


honza:
I just took it for a spin, and it works great!

PR #72, opened 2019-03-15
Add ironic to baremetal operator deployment.

imain:
Still a bit of a work in progress but it does launch the containers
and it should be close to working.

### DIFF
--- a/deploy/operator_ironic.yaml
+++ b/deploy/operator_ironic.yaml
@@ -1,0 +1,169 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: metal3-baremetal-operator
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      name: metal3-baremetal-operator
+  template:
+    metadata:
+      labels:
+        name: metal3-baremetal-operator
+    spec:
+      serviceAccountName: metal3-baremetal-operator
+      hostNetwork: true
+      containers:
+        - name: baremetal-operator
+          image: quay.io/metal3-io/baremetal-operator:master
+          ports:
+            - containerPort: 60000
+              name: metrics
+          command:
+            - /baremetal-operator
+          imagePullPolicy: Always
+          env:
+            - name: WATCH_NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
+            - name: POD_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.name
+            - name: OPERATOR_NAME
+              value: "baremetal-operator"
+            - name: DEPLOY_KERNEL_URL
+              valueFrom:
+                configMapKeyRef:
+                  name: ironic-bmo-configmap
+                  key: deploy_kernel_url
+            - name: DEPLOY_RAMDISK_URL
+              valueFrom:
+                configMapKeyRef:
+                  name: ironic-bmo-configmap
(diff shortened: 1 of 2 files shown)


docs/dev-setup.md
```diff
@@ -51,6 +51,20 @@ install the operator-sdk tools.
     oc apply -f deploy/crds/metalkube_v1alpha1_baremetalhost_cr.yaml
     ```
 
+7. Launch with Ironic in the Deployment
```
dhellmann: Maybe we should document these steps separately as "production" use?

docs/dev-setup.md
```diff
@@ -51,6 +51,20 @@ install the operator-sdk tools.
...
+
+Alternatively to steps 5 and 6, add the baremetal-operator service account to the 'privileged' security context:
+
+    ```
```
rustyrobot: nit: this won't be rendered correctly, see https://github.com/metalkube/baremetal-operator/pull/82/files

dhellmann:
Note to self: We need https://github.com/metalkube/metalkube-ironic/pull/16 and https://github.com/metalkube/metalkube-ironic/pull/18 before approving this.

imain:
I think this can be merged any time now.  The pod is working, we just need to make sure the BMO can be wired into it all.

To test see: https://github.com/imain/dev-scripts/tree/ironic_imain

zaneb:
Should we be removing the proxy in https://github.com/metal3-io/baremetal-operator/blob/master/deploy/operator.yaml#L45-L48 at the same time?

Is there a plan for doing the same for ironic-inspector? I assume if we don't do them at the same time then inspection will break (since it'll be talking to a different Ironic.)

imain:
Well, my thinking is that we leave operator.yaml as is so it can continue to work in dev-scripts.  Once we verify this can work, I'll post the patch to update dev-scripts and we'll transition to this deployment.  I'm not even sure if operator.yaml will be needed afterwards.

I should look into ironic-inspector more.. didn't think it would be an issue?  I thought whoever controlled dhcp + pxe would control the boot sequence?  I'll check it out.  Thanks Zane!

hardys:
Hey @imain can you clarify how we manage the kernel/ramdisk image and the qcow for deployment in this case?

I'm wondering the best way to handle that generally, e.g if we launch ironic on the bootstrap host or VM to deploy the masters, we'll need to download the image (or build it into a huge container image/layer), then download it again when launching Ironic to deploy the workers?

I'm also thinking about how we'll do destroy ref https://github.com/openshift-metal3/kni-installer/issues/74 - in that case I guess we'll do the deprovision steps the same as when scaling down, but we'll probably need to start the container outside the cluster to enable deprovisioning the masters.  In that case we'll want the IPA kernel/ramdisk to enable cleaning, but not the overhead of downloading a giant container image with a qcow2 embedded.

deploy/operator_ironic.yaml
```diff
@@ -0,0 +1,169 @@
...
+              valueFrom:
+                secretKeyRef:
+                  name: mariadb-password
+                  key: password
+            - name: HTTP_PORT
+              valueFrom:
+                configMapKeyRef:
+                  name: ironic-bmo-configmap
+                  key: http_port
+            - name: IP
```
dhellmann: We talked today about the need to auto-detect this IP based on a CIDR value, since the pod may move around within the cluster. But let's merge this and iterate on that step, since it will require changes to the image definitions.

PR #77, opened 2019-03-19
add go fmt job to travis

dhellmann:


dhellmann:
I copied this from the cluster-api-provider-baremetal Makefile. I thought it exited non-0 if not then there's no real reason to run it.

PR #78, opened 2019-03-19
stop using labels for status values

dhellmann:
This PR includes a couple of related changes to clean up the way we handle status values that we were storing in labels but that we want to store in the status fields instead.


dhellmann:
This PR drops some of the unused status values. I intend to have a separate PR to restore those after talking with @jtomasek

PR #81, opened 2019-03-21
Don't hardcode instanceImageSource into templateBody

yprokule:
instanceImageSource is constant and can be used within
 templates' declaration.


dhellmann:
Thanks!

PR #84, opened 2019-03-21
Add iDRAC support

zaneb:
When the BMC address is of the form idrac://... the iDRAC driver in
Ironic will be used instead of the default IPMI driver.

### DIFF
--- a/pkg/bmc/access.go
+++ b/pkg/bmc/access.go
@@ -1,6 +1,7 @@
 package bmc
 
 import (
+	"fmt"
 	"net"
 	"net/url"
 	"strings"
@@ -21,6 +22,10 @@ type AccessDetails interface {
 	// port created rather than having it discovered.
 	NeedsMAC() bool
 
+	// The name of the driver to instantiate the BMC with. This may differ
+	// from the Type - both the ipmi and libvirt types use the ipmi driver.
+	Driver() string
+
 	// DriverInfo returns a data structure to pass as the DriverInfo
 	// parameter when creating a node in Ironic. The structure is
 	// pre-populated with the access information, and the caller is
@@ -29,7 +34,19 @@ type AccessDetails interface {
 	DriverInfo(bmcCreds Credentials) map[string]interface{}
 }
 
-func getTypeHostPort(address string) (bmcType, host, port string, err error) {
+// UnknownBMCTypeError is returned when the provided BMC address cannot be
+// mapped to a driver.
+type UnknownBMCTypeError struct {
+	address string
+	bmcType string
+}
+
+func (e UnknownBMCTypeError) Error() string {
+	return fmt.Sprintf("Unknown BMC type '%s' for address %s",
+		e.bmcType, e.address)
+}
+
+func getTypeHostPort(address string) (bmcType, host, port, path string, err error) {
 	// Start by assuming "type://host:port"
 	parsedURL, err := url.Parse(address)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/bmc/access.go
```diff
@@ -29,7 +34,19 @@ type AccessDetails interface {
...
+	address string
+	bmcType string
+}
+
+func (e UnknownBMCTypeError) Error() string {
+	return fmt.Sprintf("Unknown BMC type '%s' for address %s",
+		e.bmcType, e.address)
+}
+
+func getTypeHostPort(address string) (bmcType, host, port, path string, err error) {
```
dhellmann: Should we change this function name to include that it returns a path? Or should we define some type that it can return so we can keep adding fields?

pkg/bmc/idrac.go
```diff
@@ -0,0 +1,56 @@
...
+// DriverInfo returns a data structure to pass as the DriverInfo
+// parameter when creating a node in Ironic. The structure is
+// pre-populated with the access information, and the caller is
+// expected to add any other information that might be needed (such as
+// the kernel and ramdisk locations).
+func (a *iDracAccessDetails) DriverInfo(bmcCreds Credentials) map[string]interface{} {
+	host := a.hostname
+	if strings.ContainsRune(host, ':') {
+		// Hostname is an IPv6 address
+		host = fmt.Sprintf("[%s]", host)
```
dhellmann: I'm surprised the URL class doesn't do this wrapping for us.

pkg/bmc/idrac.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+	if a.portNum != "" {
+		host = fmt.Sprintf("%s:%s", host, a.portNum)
+	}
+
+	address := url.URL{
+		Scheme: "http",
```
dhellmann: Not https?

pkg/bmc/access.go
zaneb: Honestly it should probably return a URL object (or even be moved to the ipmi/idrac files as separate, driver-specific New() functions) but it didn't seem worth rewriting everything that was there and works adequately.

dhellmann: Yeah, I just made one function because the type logic would be the same. Using a URL object might make sense.

pkg/bmc/idrac.go
zaneb: Yeah, it's annoying. It seems to be more designed for going in the other direction (splitting a URL). There is the [Hostname()](https://golang.org/pkg/net/url/#URL.Hostname) function that extracts the IPv6 address, but no corresponding method for constructing the Host from a DNS/IPv4/IPv6 addresss + port :(

pkg/bmc/idrac.go
zaneb: Good question. Does DRAC support https? Will we have the right certificates? One disadvantage of using the scheme to indicate DRAC vs. IPMI is that we lose the ability to let the user tell us.

dhellmann: Yeah, someone (@dtantsur ?) pointed out that we could do idrac+https://... if we have to.

dhellmann:
This looks good, and we can keep iterating as we need to.

pkg/bmc/access.go
zaneb: Also I convinced myself that there's something to be said for constructing it from parts like this, because there's no way to sneak through weird stuff (query strings, fragment ids, username + password) that we aren't expecting.

pkg/bmc/idrac.go
zaneb: That's a good idea, I can implement that.

russellb:
Where should this be documented, so people know what's currently supported through the API?

dhellmann:
> Where should this be documented, so people know what's currently supported through the API?

Good question. We should probably start a user guide of some sort. See #87

PR #89, opened 2019-03-22
let the baremetal operator download the checksum from a URL

dhellmann:
The unit tests for the actuator are failing because we require it to
provide a full checksum for the image being provisioned. This will
allow it to just pass us the URL, and have the operator download it to
learn the checksum from there. Eventually we can just pass the URL to
ironic and let it do the download, but we have to upgrade to a newer
ironic for that.

### DIFF
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -438,7 +438,13 @@ func (p *ironicProvisioner) Provision(userData string) (result provisioner.Resul
 		if err != nil {
 			return result, errors.Wrap(err, "Could not fetch image checksum")
 		}
-		defer resp.Body.Close()
+		if resp.Close {
+			defer resp.Body.Close()
+		}
+		if resp.StatusCode != 200 {
+			return result, fmt.Errorf("Failed to fetch image checksum from %s: [%d] %s",
+				checksum, resp.StatusCode, resp.Status)
+		}
 		checksumBody, err := ioutil.ReadAll(resp.Body)
 		if err != nil {
 			return result, errors.Wrap(err, "Could not read image checksum")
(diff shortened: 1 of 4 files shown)


pkg/provisioner/ironic/ironic.go
```diff
@@ -414,6 +425,27 @@ func (p *ironicProvisioner) Provision(userData string) (result provisioner.Resul
...
+		return result, errors.Wrap(err, "Could not understand image checksum")
+	}
+	if isURL {
+		p.log.Info("looking for checksum for image", "URL", checksum)
+		resp, err := http.Get(checksum)
+		if err != nil {
+			return result, errors.Wrap(err, "Could not fetch image checksum")
+		}
+		defer resp.Body.Close()
+		checksumBody, err := ioutil.ReadAll(resp.Body)
```
mhrivnak: I know this code only was relocated and isn't new, but I just noticed that we need to check the response code. Non-2xx codes won't result in an error.

dhellmann: Good catch. I pushed a fixup patch to address that.

pkg/provisioner/ironic/ironic.go
```diff
@@ -414,6 +425,33 @@ func (p *ironicProvisioner) Provision(userData string) (result provisioner.Resul
...
+	if err != nil {
+		return result, errors.Wrap(err, "Could not understand image checksum")
+	}
+	if isURL {
+		p.log.Info("looking for checksum for image", "URL", checksum)
+		resp, err := http.Get(checksum)
+		if err != nil {
+			return result, errors.Wrap(err, "Could not fetch image checksum")
+		}
+		if resp.Close {
```
mhrivnak: I don't think this is related to closing the body of the message. It's related to closing the connection itself, which is different.

The docs for `Get` say "When err is nil, resp always contains a non-nil resp.Body. Caller should close resp.Body when done reading from it." So I think if the error is nil, you can just always close the body when you're done. No need to check this.

PR #93, opened 2019-03-26
add active power state monitoring

dhellmann:
Regularly check the power status of the host to ensure it matches the
desired state.

### DIFF
--- a/pkg/controller/baremetalhost/baremetalhost_controller.go
+++ b/pkg/controller/baremetalhost/baremetalhost_controller.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"time"
 
 	"github.com/pkg/errors"
 
@@ -386,6 +387,39 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (reconcile
 		}
 	}
 
+	// Check the current power status against the desired power
+	// status.
+	if provResult, err = prov.UpdateHardwareState(); err != nil {
+		return reconcile.Result{}, errors.Wrap(err, "failed to update the hardware status")
+	}
+	if provResult.Dirty {
+		if err := r.saveStatus(host); err != nil {
+			return reconcile.Result{}, errors.Wrap(err,
+				"failed to update host hardware state")
+		}
+		return reconcile.Result{RequeueAfter: provResult.RequeueAfter}, nil
+	}
+	if host.Status.PoweredOn != host.Spec.Online {
+		reqLogger.Info("power state change needed",
+			"expected", host.Spec.Online, "actual", host.Status.PoweredOn)
+		if host.Spec.Online {
+			provResult, err = prov.PowerOn()
+		} else {
+			provResult, err = prov.PowerOff()
+		}
+		if err != nil {
+			return reconcile.Result{}, errors.Wrap(err, "failed to manage power state of host")
+		}
+		if provResult.Dirty {
+			if err := r.saveStatus(host); err != nil {
+				return reconcile.Result{}, errors.Wrap(err,
+					"failed to save host status after power change")
+			}
+			return reconcile.Result{Requeue: true, RequeueAfter: provResult.RequeueAfter}, nil
+		}
+		return reconcile.Result{RequeueAfter: provResult.RequeueAfter}, nil
+	}
+
 	// If we reach this point we haven't encountered any issues
 	// communicating with the host, so ensure the error message field
 	// is cleared.
@@ -413,9 +447,9 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (reconcile
 		}
 	}
 
-	// Pod already exists - don't requeue
+	// Come back in 60 seconds to keep an eye on the power state
 	reqLogger.Info("Done with reconcile")
-	return reconcile.Result{}, nil
+	return reconcile.Result{RequeueAfter: time.Second * 60}, nil
 }
 
 func (r *ReconcileBareMetalHost) saveStatus(host *metalkubev1alpha1.BareMetalHost) error {
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -329,13 +329,12 @@ func (p *ironicProvisioner) ValidateManagementAccess() (result provisioner.Resul
 func (p *ironicProvisioner) InspectHardware() (result provisioner.Result, err error) {
 	p.log.Info("inspecting hardware", "status", p.host.OperationalStatus())
 
-	_, err = p.findExistingHost()
+	ironicNode, err := p.findExistingHost()
 	if err != nil {
 		return result, errors.Wrap(err, "failed to find existing host")
 	}
-	if p.host.HasError() {
-		p.log.Info("host has error, skipping hardware inspection")
-		return result, nil
+	if ironicNode == nil {
+		return result, fmt.Errorf("no ironic node for host")
 	}
 
 	if p.host.OperationalStatus() != metalkubev1alpha1.OperationalStatusInspecting {
@@ -415,9 +414,8 @@ func (p *ironicProvisioner) UpdateHardwareState() (result provisioner.Result, er
 	if err != nil {
 		return result, errors.Wrap(err, "failed to find existing host")
 	}
-	if p.host.HasError() {
-		p.log.Info("host has error, skipping hardware state update")
-		return result, nil
+	if ironicNode == nil {
+		return result, fmt.Errorf("no ironic node for host")
 	}
 
 	var discoveredVal bool
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -42,8 +42,10 @@ const (
 	stateProvisioning         = "provisioning"
 	stateProvisioned          = "provisioned"
 	stateDeprovisioning       = "deprovisioning"
-	powerOn                   = "power on"
-	powerOff                  = "power off"
+	// See nodes.Node.PowerState for details
+	powerOn   = "power on"
+	powerOff  = "power off"
+	powerNone = "None"
 )
 
 // Provisioner implements the provisioning.Provisioner interface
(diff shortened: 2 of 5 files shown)


dhellmann:
I wonder if we only want to power the server on if there is an image associated with it.

juliakreger:
> I wonder if we only want to power the server on if there is an image associated with it.

That might not mesh well with  our path to uploading information into inspection with discovery is through power operations and if the node is already on with ironic 12.1.0, we can just jump through some of the power cycles.  Just thinking in the grand scheme.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -385,6 +386,39 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (reconcile
...
+		return reconcile.Result{}, errors.Wrap(err, "failed to update the hardware status")
+	}
+	if provResult.Dirty {
+		if err := r.saveStatus(host); err != nil {
+			return reconcile.Result{}, errors.Wrap(err,
+				"failed to update host hardware state")
+		}
+		return reconcile.Result{RequeueAfter: provResult.RequeueAfter}, nil
+	}
+	if host.Status.PoweredOn != host.Spec.Online {
```
juliakreger: If I grok this correctly. If it is not powered on, and we expect it ?deployed? then proceed?

pkg/provisioner/ironic/ironic.go
```diff
@@ -728,19 +765,57 @@ func (p *ironicProvisioner) Deprovision(deleteIt bool) (result provisioner.Resul
...
+		ironicNode.UUID,
+		nodes.PowerStateOpts{
+			Target: target,
+		})
+
+	switch changeResult.Err.(type) {
+	case nil:
+		result.Dirty = true
+		p.log.Info("power change OK")
+	case gophercloud.ErrDefault409:
```
juliakreger: I like!

dhellmann:
> Looks pretty good. Makes sense to me, but another set of eyes would naturally be really good since I'm not super familiar with the operator code. Slightly concerned about not being able to actuate power unless the node is online, but I might not correctly grok the spec data model.

"Online" is a poorly named field that means "bring the host online" and is in fact just the software-based power switch. I'm definitely open to suggestions for a new name for that field.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -385,6 +386,39 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (reconcile
 		}
 	}
 
+	// Check the current power status against the desired power
+	// status.
+	if provResult, err = prov.UpdateHardwareState(); err != nil {
```
mhrivnak: Is this something that accesses the BMC, and thus needs to be throttled to ~once a minute? Is something below this call doing throttling? I didn't see it, but maybe it's happening in ironic or somewhere else I didn't notice.

pkg/provisioner/ironic/ironic.go
```diff
@@ -728,19 +765,57 @@ func (p *ironicProvisioner) Deprovision(deleteIt bool) (result provisioner.Resul
...
+		"current", p.host.Status.PoweredOn,
+		"target", ironicNode.TargetPowerState)
+
+	if ironicNode.PowerState != "power on" {
+		if ironicNode.TargetPowerState == "power on" {
+			p.log.Info("waiting for power status to change")
+			result.RequeueAfter = powerRequeueDelay
+			return result, nil
+		}
+		result, err = p.changePower(ironicNode, nodes.PowerOn)
```
mhrivnak: need to check this `err` ?

pkg/provisioner/ironic/ironic.go
```diff
@@ -751,14 +826,19 @@ func (p *ironicProvisioner) PowerOn() (result provisioner.Result, err error) {
...
-	if p.host.Status.PoweredOn {
-		p.host.Status.PoweredOn = false
-		result.Dirty = true
+	if ironicNode.PowerState != "power off" {
+		if ironicNode.TargetPowerState == "power off" {
+			p.log.Info("waiting for power status to change")
+			result.RequeueAfter = powerRequeueDelay
+			return result, nil
+		}
+		result, err = p.changePower(ironicNode, nodes.PowerOff)
```
mhrivnak: need to check this `err` ?

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -385,6 +386,39 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (reconcile
...
+			return reconcile.Result{}, errors.Wrap(err, "failed to manage power state of host")
+		}
+		if provResult.Dirty {
+			if err := r.saveStatus(host); err != nil {
+				return reconcile.Result{}, errors.Wrap(err,
+					"failed to save host status after power change")
+			}
+			return reconcile.Result{Requeue: true, RequeueAfter: provResult.RequeueAfter}, nil
+		}
+		return reconcile.Result{Requeue: false, RequeueAfter: provResult.RequeueAfter}, nil
```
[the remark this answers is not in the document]
dhellmann: Oh, I think I was just being explicit while I worked out the logic and didn't clean up after myself.

pkg/controller/baremetalhost/baremetalhost_controller.go
dhellmann: The Provisioner is responsible for returning a result with a delay value (see line 399 for where it is used). In the current implementation, it's just a database read because Ironic is polling the BMC on its own schedule.

pkg/provisioner/ironic/ironic.go
dhellmann: d'oh

pkg/provisioner/ironic/ironic.go
dhellmann: d'oh

pkg/controller/baremetalhost/baremetalhost_controller.go
mhrivnak: Perfect. I did see the delay from line 399, but wanted to make sure we weren't depending on the controller's queue to guarantee that it won't reconcile sooner.

pkg/provisioner/ironic/ironic.go
```diff
@@ -395,6 +396,42 @@ func (p *ironicProvisioner) InspectHardware() (result provisioner.Result, err er
...
+
+	var discoveredVal bool
+	switch ironicNode.PowerState {
+	case "power on":
+		discoveredVal = true
+	case "power off":
+		discoveredVal = false
+	default:
+		p.log.Info("unknown power state", "value", ironicNode.PowerState)
+		return result, nil
```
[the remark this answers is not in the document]
dhellmann: I think not. I think the only reason we would hit that is if I guessed wrong at the spelling of "power on" and "power off".

pkg/provisioner/ironic/ironic.go
```diff
@@ -728,19 +765,57 @@ func (p *ironicProvisioner) Deprovision(deleteIt bool) (result provisioner.Resul
...
 	}
 
-	if !p.host.Status.PoweredOn {
-		p.host.Status.PoweredOn = true
-		result.Dirty = true
+	p.log.Info("checking current state",
+		"current", p.host.Status.PoweredOn,
+		"target", ironicNode.TargetPowerState)
+
+	if ironicNode.PowerState != "power on" {
```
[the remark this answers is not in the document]

dhellmann:
Addressed all comments and rebased.

pkg/provisioner/ironic/ironic.go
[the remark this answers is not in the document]
zaneb: If hitting this indicates a programming error then I'd tend to think it should be logged at a higher level than Info.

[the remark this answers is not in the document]
dhellmann: The logger doesn't support Warning and logging an error requires having an error object. Should I create an error here?

pkg/controller/baremetalhost/baremetalhost_controller.go
dhellmann: If the desired power state (in the spec) is not the same as the actual power state (in the status) then we want to change the state.

pkg/provisioner/ironic/ironic.go
```diff
@@ -395,6 +396,42 @@ func (p *ironicProvisioner) InspectHardware() (result provisioner.Result, err er
...
+func (p *ironicProvisioner) UpdateHardwareState() (result provisioner.Result, err error) {
+	p.log.Info("updating hardware state")
+
+	ironicNode, err := p.findExistingHost()
+	if err != nil {
+		return result, errors.Wrap(err, "failed to find existing host")
+	}
+	if p.host.HasError() {
+		p.log.Info("host has error, skipping hardware state update")
+		return result, nil
```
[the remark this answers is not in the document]
dhellmann: I reviewed the reason for this, and I think it's another leftover from earlier error handling in the middle of a development phase. I've removed the check in the most recent commit.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -386,6 +387,39 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (reconcile
...
+		}
+		if err != nil {
+			return reconcile.Result{}, errors.Wrap(err, "failed to manage power state of host")
+		}
+		if provResult.Dirty {
+			if err := r.saveStatus(host); err != nil {
+				return reconcile.Result{}, errors.Wrap(err,
+					"failed to save host status after power change")
+			}
+			return reconcile.Result{Requeue: true, RequeueAfter: provResult.RequeueAfter}, nil
```
mhrivnak: It looks like the `Requeue` field is [ignored](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.1.10/pkg/internal/controller/controller.go#L221-L229) if `RequeueAfter` is set.

dhellmann: Depending on the state of things, the delay won't always be set by the provisioner, so explicitly setting Requeue to true here ensures that we're requeued.

PR #97, opened 2019-03-27
Provision only online

dhellmann:


dhellmann:
This should not merge until https://github.com/metalkube/cluster-api-provider-baremetal/pull/46 does

pkg/apis/metalkube/v1alpha1/baremetalhost_types.go
```diff
@@ -342,6 +342,10 @@ func (host *BareMetalHost) CredentialsNeedValidation(currentSecret corev1.Secret
 // status and returns true when more work is needed or false
 // otherwise.
 func (host *BareMetalHost) NeedsProvisioning() bool {
+	if !host.Spec.Online {
```
mhrivnak: It would be helpful to describe this in the docs for the host spec. Mentioning this on the `Online` field would be a good idea, but it feels like we're at a point where it would be good to summarize what things a user needs to do to their Spec in order to initiate provisioning.

It also might help to express in the host status that it's not provisioning because the host is powered off.

dhellmann: I've added the docs.
I'm not sure about updating the host status to say why it isn't doing something that isn't being requested, but we can talk more about that.

docs/api.md
```diff
@@ -149,3 +149,18 @@ hardware does not match a known profile, the value "unknown" is used.
...
   the error condition.
+
+## Triggering Provisioning
+
+Several conditions must be met in order to initiate provisioning.
+
+1. The host spec must have an image with a URL pointing to a valid
+   image file visible from within the cluster and from the host
+   receiving the image.
+2. The host provisioning image URL field must be empty, indicating
```
mhrivnak: Nit: I suggest rewording this to focus on the state of the world primarily, and the contents of the status as a reflection. Something like "The host must not have an image currently provisioned, as reflected by `status.provisioning.image`."

docs/api.md
```diff
@@ -149,3 +149,18 @@ hardware does not match a known profile, the value "unknown" is used.
...
+   that no other image has been provisioned to the host. To reuse an
+   existing host with a different image, deprovision the host first.
+3. The host must be marked online so that the operator will keep the
```
mhrivnak: nit: I suggest being more specific about the API: "The host spec must have `Online` set to `true` so that the controller knows to keep it powered on."

pkg/apis/metalkube/v1alpha1/baremetalhost_types.go
mhrivnak: Sounds good. We can work on that status later. Perhaps we should even do validation and prevent someone from setting the image while power is off. Generally, I'm just looking at it as an invalid state; the user has asked us to ensure that the host is provisioned while the power is off. We can't do both of those.

docs/api.md
dhellmann: Done

docs/api.md
dhellmann: Done

PR #100, opened 2019-03-28
Demo provisioner and controller logic cleanup

dhellmann:
In the course of building this demo provisioner, I found some additional ways to clean up the controller logic for error handling. I have prepared separate patches in a series so they are easier to review.

### DIFF
--- a/pkg/controller/baremetalhost/baremetalhost_controller.go
+++ b/pkg/controller/baremetalhost/baremetalhost_controller.go
@@ -11,6 +11,7 @@ import (
 	metalkubev1alpha1 "github.com/metalkube/baremetal-operator/pkg/apis/metalkube/v1alpha1"
 	"github.com/metalkube/baremetal-operator/pkg/bmc"
 	"github.com/metalkube/baremetal-operator/pkg/provisioner"
+	"github.com/metalkube/baremetal-operator/pkg/provisioner/demo"
 	"github.com/metalkube/baremetal-operator/pkg/provisioner/fixture"
 	"github.com/metalkube/baremetal-operator/pkg/provisioner/ironic"
 	"github.com/metalkube/baremetal-operator/pkg/utils"
@@ -37,9 +38,12 @@ const (
 )
 
 var runInTestMode bool
+var runInDemoMode bool
 
 func init() {
 	flag.BoolVar(&runInTestMode, "test-mode", false, "disable ironic communication")
+	flag.BoolVar(&runInDemoMode, "demo-mode", false,
+		"use the demo provisioner to set host states")
 }
 
 var log = logf.Log.WithName("controller_baremetalhost")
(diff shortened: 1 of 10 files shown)


pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -167,6 +171,9 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (reconcile
 		return reconcile.Result{}, nil
 	}
 
+	// Clear any error so we can recompute it
+	host.ClearError()
```
zaneb: I'm trying to work through the implications of ignoring the return code here.

Shouldn't we save the result of this and then check that on line 455 instead of checking the result of clearing it a second time?

dhellmann: Hrm, yes, you're right. Updating...

dhellmann: Ah, no, I remember why that is now. If we had an error and clear it at the top of the loop, then do not reset the error, we're going to end up saving the status somewhere else through the process because whatever failed before will have worked this time. So by the time we get to the bottom of the loop there won't be anything to clear and that entire block needs to come out.

dhellmann:
After this merges I plan to rework Reconcile() into a series of explicit phases that call other methods so that it's clear what error handling is common and what is unique to each phase. I hope that will make the overall logic easier to follow.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -279,6 +286,14 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (reconcile
 			return result, nil
 		}
 
+		if host.HasError() {
+			// We have tried to register and validate the host and
+			// that failed in a way we assume is not retryable, so do
+			// not proceed to any other steps.
+			reqLogger.Info("registration error")
```
mhrivnak: Do we assume that the error information in the CR status gives the user some direction about what they should do to fix the situation?

And is it safe to assume that fixing it will always involve some API action that causes reconciliation to run again? I just want to make sure there is some path to a user taking action to fix the problem and then re-reconciling.

tools/clean_demo_hosts.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash -x
+
+for host in $(oc get baremetalhosts --no-headers | grep '^demo-' | awk '{print $1}')
+do
+    oc delete baremetalhost $host
```
mhrivnak: Just as an idea, you could put a label on all of them, and just `oc delete baremetalhost -l metalkubedemo=true` or something like that. This certainly gets the job done though.

dhellmann: Good idea. See #108

PR #101, opened 2019-03-29
look for busy status from ironic

dhellmann:
Ironic may return a 409 status indicating that the host is busy any
time a change is requested. Add explicit handling for those errors so
they are treated as expected and so we wait before trying again.

### DIFF
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -100,7 +100,7 @@ func (p *ironicProvisioner) validateNode(ironicNode *nodes.Node) (ok bool, err e
 	p.log.Info("validating node settings in ironic")
 	validateResult, err := nodes.Validate(p.client, ironicNode.UUID).Extract()
 	if err != nil {
-		return false, errors.Wrap(err, "failed to validate node settings in ironic")
+		return false, err // do not wrap error so we can check type in caller
 	}
 	if !validateResult.Boot.Result {
 		validationErrors = append(validationErrors, validateResult.Boot.Reason)
@@ -280,7 +280,14 @@ func (p *ironicProvisioner) ValidateManagementAccess() (result provisioner.Resul
 			nodes.ProvisionStateOpts{
 				Target: nodes.TargetManage,
 			})
-		if changeResult.Err != nil {
+		switch changeResult.Err.(type) {
+		case nil:
+		case gophercloud.ErrDefault409:
+			p.log.Info("could not change state of host, busy")
+			result.Dirty = true
+			result.RequeueAfter = provisionRequeueDelay
+			return result, nil
+		default:
 			return result, errors.Wrap(changeResult.Err,
 				"failed to change provisioning state to manage")
 		}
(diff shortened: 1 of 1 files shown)


dhellmann:
I've personally already seen 409 errors just through normal use, so I want to handle those. If 503 errors are more rare, I'm comfortable leaving those to the default error handling which also triggers a retry after logging the error. If we start seeing those more often, we can handle them in a similar way to the 409 errors.

juliakreger:
> Would it help to have some additional details about the host in the log msg (input to p.log.Info)? Something akin to a hostname ? Might help us diagnose more involved issues later?

We would really need the conductor and api logs to pair up together and just compare the locking/request activity. It may also just be that we're trying to do things while Ironic is in the middle of intermediate actions like checking if a node can be talked to.

sadasu:
> > Would it help to have some additional details about the host in the log msg (input to p.log.Info)? Something akin to a hostname ? Might help us diagnose more involved issues later?
> 
> We would really need the conductor and api logs to pair up together and just compare the locking/request activity. It may also just be that we're trying to do things while Ironic is in the middle of intermediate actions like checking if a node can be talked to.

Agreed.

dhellmann:
> Would it help to have some additional details about the host in the log msg (input to p.log.Info)? Something akin to a hostname ? Might help us diagnose more involved issues later?

The logger is constructed so that all messages include the name and namespace of the CRD being reconciled. Those can be mapped to the UUID of the node that Ironic knows about. Here's some example log output:

```
2019-03-29T13:35:17.508-0400	INFO	controller_baremetalhost	Reconciling BareMetalHost	{"Request.Namespace": "openshift-machine-api", "Request.Name": "openshift-worker-1"}
2019-03-29T13:35:17.508-0400	INFO	ironic	updating hardware state	{"host": "openshift-worker-1"}
2019-03-29T13:35:17.508-0400	INFO	ironic	looking for existing node by ID	{"host": "openshift-worker-1", "ID": "a89e4594-1e2d-45fa-8c7a-fb9b0dc5129b"}
2019-03-29T13:35:17.527-0400	INFO	ironic	found existing node by ID	{"host": "openshift-worker-1"}
```

dhellmann:
> The prospect of having to handle more error types later makes me wonder if it's worth trying to abstract this out into a function, but it would probably result in some loss of fidelity in the error/log messages.

I thought about trying to do that, but since there were other errors to be handled at some times I decided it would be clearer to just throw switch statements everywhere. I'm open to ideas for a function to do that, though. It all makes me miss python's function decorators where we could trap that exception neatly. Maybe we could do something similar, if we look inside the wrapped error?

sadasu:
> > Would it help to have some additional details about the host in the log msg (input to p.log.Info)? Something akin to a hostname ? Might help us diagnose more involved issues later?
> 
> The logger is constructed so that all messages include the name and namespace of the CRD being reconciled. Those can be mapped to the UUID of the node that Ironic knows about. Here's some example log output:
> 
> ```
> 2019-03-29T13:35:17.508-0400	INFO	controller_baremetalhost	Reconciling BareMetalHost	{"Request.Namespace": "openshift-machine-api", "Request.Name": "openshift-worker-1"}
> 2019-03-29T13:35:17.508-0400	INFO	ironic	updating hardware state	{"host": "openshift-worker-1"}
> 2019-03-29T13:35:17.508-0400	INFO	ironic	looking for existing node by ID	{"host": "openshift-worker-1", "ID": "a89e4594-1e2d-45fa-8c7a-fb9b0dc5129b"}
> 2019-03-29T13:35:17.527-0400	INFO	ironic	found existing node by ID	{"host": "openshift-worker-1"}
> ```

Thanks @dhellmann ! This output really helps!

PR #102, opened 2019-03-29
Fix CR reference in the docs

bcrochet:
In the minikube section of the docs, the final step of creating a CR
references a file that no longer exists. Point it to example-host.yaml
instead.


dhellmann:
D'oh, thanks!

PR #104, opened 2019-03-29
Move inspecting status location

dhellmann:
This change should make tracking "activity" more consistent because it all happens in 1 field. The main status field is then left for tracking whether a host is OK or not. Note that only the most recent patch in the series is part of this PR. The others are from #100, on which this is based.

### DIFF
--- a/pkg/provisioner/provisioner.go
+++ b/pkg/provisioner/provisioner.go
@@ -66,3 +66,33 @@ type Result struct {
 	// Dirty is also true.
 	RequeueAfter time.Duration
 }
+
+const (
+	// StateNone means the state is unknown
+	StateNone = ""
+	// StateRegistrationError means there was an error registering the
+	// host with the backend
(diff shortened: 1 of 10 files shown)


dhellmann:
This depends on #100

pkg/provisioner/provisioner.go
```diff
@@ -95,4 +95,7 @@ const (
 	// StateDeprovisioning means we are removing an image from the
 	// host's disk(s)
 	StateDeprovisioning = "deprovisioning"
+	// StateInspecting means we are running the agent on the host to
+	// learn about the hardware components available there
+	StateInspecting = "inspecting"
```
zaneb: Not an issue with this patch per se, but shouldn't these have a type?

dhellmann: Yes, good point. I'll do that after this PR merges so I can get them all at once.

PR #106, opened 2019-04-01
tool for creating worker nodes

dhellmann:


russellb:
Hm, this seems to be OpenShift specific.  Should it go somewhere else, instead?

dhellmann:
I removed the bits that are openshift-specific. We want folks to use the actuator anyway.

docs/dev-setup.md
```diff
@@ -111,5 +111,34 @@ spec:
...
+
+---
+apiVersion: metalkube.org/v1alpha1
+kind: BareMetalHost
+metadata:
+  name: worker-99
+spec:
+  online: true
+  bmc:
+    address: 1.2.3.4
```
russellb: I guess it assumes IPMI without a prefix?

russellb:
lgtm

docs/dev-setup.md
dhellmann: Yes, see https://github.com/metalkube/baremetal-operator/blob/master/pkg/bmc/access.go#L49

PR #108, opened 2019-04-01
label all demo hosts for more robust cleanup

dhellmann:
Instead of relying on the names of the hosts, add a label that we can
use in the cleanup script to find them again.

### DIFF
--- a/tools/clean_demo_hosts.sh
+++ b/tools/clean_demo_hosts.sh
@@ -1,6 +1,3 @@
 #!/bin/bash -x
 
-for host in $(oc get baremetalhosts --no-headers | grep '^demo-' | awk '{print $1}')
-do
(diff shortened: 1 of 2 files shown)


tools/clean_demo_hosts.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash -x
 
-for host in $(oc get baremetalhosts --no-headers | grep '^demo-' | awk '{print $1}')
+for host in $(oc get baremetalhosts --no-headers -l metalkubedemo | awk '{print $1}')
```
mhrivnak: I think you can skip the `for` loop and just do `oc delete baremetalhosts -l metalkubedemo`.

dhellmann: Done

PR #109, opened 2019-04-01
Split Reconcile() into phases

dhellmann:
This is an attempt to restructure the Reconcile() method based on a series of phases so we can add consistent handling for the need to save data or report errors. I have it split up into separate commits to make reviews easier, and the tests all worked at each commit point.

### DIFF
--- a/pkg/controller/baremetalhost/baremetalhost_controller.go
+++ b/pkg/controller/baremetalhost/baremetalhost_controller.go
@@ -276,6 +276,7 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (reconcile
 		{name: "inspect hardware", action: r.phaseInspectHardware},
 		{name: "hardware profile", action: r.phaseSetHardwareProfile},
 		{name: "provisioning", action: r.phaseProvisioning},
+		{name: "deprovisioning", action: r.phaseDeprovisioning},
 	}
 	for _, phase := range phases {
 		ctx.log = reqLogger.WithValues("phase", phase.name)
@@ -308,23 +309,6 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (reconcile
 			"after", (*phaseResult).RequeueAfter,
 		)
 		return *phaseResult, nil
-	}
-
-	// Start/continue deprovisioning if we need to.
-	if host.NeedsDeprovisioning() {
-		if provResult, err = prov.Deprovision(false); err != nil {
-			return reconcile.Result{}, errors.Wrap(err, "failed to deprovision")
-		}
-		if provResult.Dirty {
-			if err := r.saveStatus(host); err != nil {
-				return reconcile.Result{}, errors.Wrap(err,
-					"failed to clear host status on deprovision")
-			}
-			// Go back into the queue and wait for the Deprovision() method
-			// to return false, indicating that it has no more work to
-			// do.
-			return reconcile.Result{RequeueAfter: provResult.RequeueAfter}, nil
-		}
 	}
 
 	// Check the current power status against the desired power
@@ -568,6 +552,23 @@ func (r *ReconcileBareMetalHost) phaseProvisioning(ctx reconcileContext) (result
 			RequeueAfter: provResult.RequeueAfter,
 		}
 		return result, nil
+	}
+
+	return nil, nil
+}
+
+func (r *ReconcileBareMetalHost) phaseDeprovisioning(ctx reconcileContext) (result *reconcile.Result, err error) {
+	var provResult provisioner.Result
+
+	if !ctx.host.NeedsDeprovisioning() {
+		return nil, nil
+	}
+
+	if provResult, err = ctx.provisioner.Deprovision(false); err != nil {
+		return nil, errors.Wrap(err, "failed to deprovision")
+	}
+	if provResult.Dirty {
+		return &reconcile.Result{RequeueAfter: provResult.RequeueAfter}, nil
 	}
 
 	return nil, nil
--- a/pkg/controller/baremetalhost/baremetalhost_controller.go
+++ b/pkg/controller/baremetalhost/baremetalhost_controller.go
@@ -115,7 +115,7 @@ type ReconcileBareMetalHost struct {
 
 // Instead of passing a zillion arguments to the action of a phase,
 // hold them in a context
-type reconcileContext struct {
+type reconcileInfo struct {
 	log            logr.Logger
 	host           *metalkubev1alpha1.BareMetalHost
 	provisioner    provisioner.Provisioner
@@ -129,7 +129,7 @@ type reconcileContext struct {
 // - Return a result if the host should be saved and requeued without error.
 // - Return error if there was an error.
 // - Return double nil if nothing was done and processing should continue.
-type reconcileAction func(context reconcileContext) (*reconcile.Result, error)
+type reconcileAction func(info reconcileInfo) (*reconcile.Result, error)
 
 // One step of reconciliation
 type reconcilePhase struct {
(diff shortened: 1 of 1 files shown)


pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -230,248 +259,327 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (reconcile
...
+			return reconcile.Result{}, errors.Wrap(err, "failed to initialize status block")
 		}
 	}
-	if host.Status.Provisioning.State == provisioner.StateRegistrationError {
-		// We have tried to register and validate the host and that
-		// failed, so do not proceed to any other steps.
-		reqLogger.Info("registration error, stopping")
-		return reconcile.Result{}, nil
+
+	// Come back in 60 seconds to keep an eye on the power state
```
mhrivnak: Just while it's on my mind: if it were ever desirable to do this 60-second check of power state without reconciling the whole object, we could make that happen by creating a custom event source and having a dedicated loop that just checks power state every 60s. We can talk through it if that ever becomes desirable.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -230,248 +259,327 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (reconcile
...
+	reqLogger.Info("Done with reconcile")
+	return reconcile.Result{RequeueAfter: time.Second * 60}, nil
+}
+
+// Handle delete operations.
+func (r *ReconcileBareMetalHost) phaseDelete(ctx reconcileContext) (result *reconcile.Result, err error) {
```
mhrivnak: Same context issue as above. It's very common for the first argument of a function to be `func foo(ctx context.Context)`. This could be confusing. It's not as bad as naming the first parameter of a non-method python function `self`, but maybe 30% of the way there.

zaneb:
Have you considered structuring this code in the form of a finite state machine? Is there a reason we are trying to handle multiple phases in a single call, rather than just re-queueing after each action?

Not having everything in one giant function is definitely an improvement, but I still find the code very hard to follow and I suspect I will continue to unless we adopt something where the state transitions are much more structured.

pkg/controller/baremetalhost/baremetalhost_controller.go
dhellmann: Most of the rest of the logic bypasses the checks very quickly, but if that changes I could definitely see the usefulness of having a second reconcile loop just for power.

dhellmann:
> Have you considered structuring this code in the form of a finite state machine? Is there a reason we are trying to handle multiple phases in a single call, rather than just re-queueing after each action?
> 
> Not having everything in one giant function is definitely an improvement, but I still find the code very hard to follow and I suspect I will continue to unless we adopt something where the state transitions are much more structured.

I did think about a state machine when I started all of this, especially with the provisioner API. Looking at the state diagram, though, it's almost entirely a series of nodes with 1 edge between them and self-referencing loops on most of them (keep trying to inspect hardware until that's done, then keep trying to provision until that's done, then keep trying to power on until that's done, etc.). There were also a few stop states for errors, and I guess a branch for provision vs. deprovision, but given the largely linear nature of it, dealing with explicit state machine code felt like overkill.

I'm happy to revisit all of that, though, because I agree this is still not as clear as I would like for it to be.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -230,248 +259,327 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (reconcile
...
-
-		// Remove finalizer to allow deletion
-		reqLogger.Info("cleanup is complete, removing finalizer")
-		host.ObjectMeta.Finalizers = utils.FilterStringFromList(
-			host.ObjectMeta.Finalizers, metalkubev1alpha1.BareMetalHostFinalizer)
-		if err := r.client.Update(context.TODO(), host); err != nil {
-			return reconcile.Result{}, errors.Wrap(err, "failed to remove finalizer")
-		}
-		return reconcile.Result{}, nil // done
+	ctx := reconcileContext{
```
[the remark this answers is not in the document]
dhellmann: I replaced "ctx" and "context" with "info". I'm open to more descriptive suggestions.

dhellmann:
> > Have you considered structuring this code in the form of a finite state machine? Is there a reason we are trying to handle multiple phases in a single call, rather than just re-queueing after each action?
> > Not having everything in one giant function is definitely an improvement, but I still find the code very hard to follow and I suspect I will continue to unless we adopt something where the state transitions are much more structured.
> 
> I did think about a state machine when I started all of this, especially with the provisioner API. Looking at the state diagram, though, it's almost entirely a series of nodes with 1 edge between them and self-referencing loops on most of them (keep trying to inspect hardware until that's done, then keep trying to provision until that's done, then keep trying to power on until that's done, etc.). There were also a few stop states for errors, and I guess a branch for provision vs. deprovision, but given the largely linear nature of it, dealing with explicit state machine code felt like overkill.
> 
> I'm happy to revisit all of that, though, because I agree this is still not as clear as I would like for it to be.

Would it make more sense if a reconcilePhase object had methods like isNeeded() and doWork() instead of combining those into the current action() function? We could define reconcilePhase as an interface, and implement each phase as its own class.

zaneb:
> I did think about a state machine when I started all of this, especially with the provisioner API. Looking at the state diagram, though, it's almost entirely a series of nodes with 1 edge between them and self-referencing loops on most of them (keep trying to inspect hardware until that's done, then keep trying to provision until that's done, then keep trying to power on until that's done, etc.). There were also a few stop states for errors, and I guess a branch for provision vs. deprovision, but given the largely linear nature of it, dealing with explicit state machine code felt like overkill.

That actually sounds ideal in terms of implementation simplicity to me :)
I might have a go at putting together a PoC after I've finished the introspection stuff.

mhrivnak:
FWIW I suspect that in most cases, by the time you figure out if a given phase is needed, you're most of the way done with actually doing its work. It'll likely be simpler to just let one function do whatever it thinks is necessary, however much or little that is, like the current code in the PR does. After all, that's the pattern of the `Reconcile` function itself.

I may be biased by all the time I've spent staring at various controller code, but to me the PR as-is looks quite readable and easy to understand.

pkg/controller/baremetalhost/baremetalhost_controller.go
mhrivnak: Rather than make a second reconcile loop, we'd just start a goroutine that periodically checks power, and if it observes a change, triggers the normal reconcile loop.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -110,6 +113,30 @@ type ReconcileBareMetalHost struct {
...
+//
+// - Return a result if the host should be saved and requeued without error.
+// - Return error if there was an error.
+// - Return double nil if nothing was done and processing should continue.
+type reconcileAction func(info reconcileInfo) (*reconcile.Result, error)
+
+// One step of reconciliation
+type reconcilePhase struct {
+	name   string
+	action reconcileAction
```
mhrivnak: FWIW I probably would have in-lined the type here, rather than define `reconcileAction`. It doesn't seem like you need it as a separate type, so you could condense it into what's below. But there's not any harm in breaking it out like you did, so feel free to go either way.

```
type reconcilePhase struct {
	name   string

        // - Return a result if the host should be saved and requeued without error.
        // - Return error if there was an error.
        // - Return double nil if nothing was done and processing should continue.
	action func(info reconcileInfo) (*reconcile.Result, error)
}
```

PR #110, opened 2019-04-02
clear the provisioning state when inspection is done

dhellmann:


dhellmann:
"Ready" means it is available to be used, so I used None as an intermediate step between this and inspection.

dhellmann:
I rebased this and updated the new state to ready instead of none.

dhellmann:
I think this is going to be obsolete, based on the discussion we had yesterday about moving state management out of the drivers and into Reconcile().

PR #111, opened 2019-04-02
make it safe to register masters

dhellmann:


dhellmann:
Only the last 2 patches in this PR are relevant for this. The rest are part of #109

dhellmann:
This will make it safe to do the 2-step registration of masters, first to set the spec and then to set the status.

PR #120, opened 2019-04-05
Document ProvisioningState transitions of BaremetalHost

zaneb:
Add a graphviz representation of the possible ProvisioningState
transitions of the BaremetalHost object.

I thought there might be a way of previewing the output on GitHub, but there is not :(


dhellmann:
Do you want to check in a PNG? Should we put this in the docs repo as a design doc? Or here as a design doc?

zaneb:
I think it's better here as a design doc, because as we update the code we can update the diagram as we go.
Before I saw your comment I updated the patch to add generation of the PNG to the makefile (with `make docs`). I normally try to avoid putting stuff that can be built in git, but I can check in the PNG as well to make it more useful to view in GitHub.
Also updated the graph to take into account the fact that #110 didn't merge.

PR #121, opened 2019-04-05
rely more on ironic state within the provisioner

dhellmann:


dhellmann:
@jtomasek This change is going to remove a couple of host states. I'm still working on error handling, but it looks like we may not update the provisioning state field on error, and just update the operational status and error message fields.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -251,80 +250,93 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (reconcile
 		return reconcile.Result{}, nil
 	}
 
+	// Pick the action to perform
```
dhellmann: We may need to add some provisioner APIs to probe some of these states. For example, if we provision a host and then the Pod is killed and restarted, we won't have it registered and will need to do that without reprovisioning it.

dhellmann:
I added another patch that makes the ironic driver deal better with error states.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -251,80 +250,93 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (reconcile
...
+		result, err = r.actionProvisioning(prov, info)
+	case metalkubev1alpha1.StateDeprovisioning:
+		result, err = r.actionDeprovisioning(prov, info)
+	case metalkubev1alpha1.StateProvisioned:
+		result, err = r.actionManageHostPower(prov, info)
+	case metalkubev1alpha1.StateReady:
+		result, err = r.actionManageHostPower(prov, info)
+	default:
+		// Probably a provisioning error state?
+		return reconcile.Result{}, fmt.Errorf("Unrecognized action %q", actionName)
```
zaneb: This can't actually happen AFAICT

pkg/apis/metalkube/v1alpha1/baremetalhost_types.go
```diff
@@ -51,17 +51,13 @@ const (
...
+	// StateMatchProfile means we are comparing the discovered details
+	// against known hardware profiles
+	StateMatchProfile ProvisioningState = "match profile"
+
 	// StateReady means the host can be consumed
 	StateReady ProvisioningState = "ready"
 
-	// StatePreparingToProvision means we are updating the host to
-	// receive its image
-	StatePreparingToProvision ProvisioningState = "preparing to provision"
```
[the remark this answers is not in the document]
dhellmann: I wonder if the fact that that validation is also a separate step is an implementation detail? I honestly would have expected a provisioning tool API to have at least one entry point where *all* of the info could be passed in to have it start doing the work. The fact that we're making a bunch of calls before Ironic does any work seems like something we don't need to show the user.

That said, if we want a PrepareForProvisioning() method in the provisioner API, I could see how to implement it.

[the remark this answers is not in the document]
zaneb: I don't think it really matters. It's fine to treat it as an implementation detail. Since we already had it there anyway, I thought there may be some value to the user in distinguishing between "there's no way we're doing anything with that nonsense you gave us" vs. "we tried but Ironic says no" - it could narrow down where to start debugging.
It would only really matter one way or the other if it significantly simplified the code. e.g. if we fail in validation we might be able to go straight back into attempting provisioning again once data changes, but if it fails during provisioning itself we might have deprovision first? If that were the case then having separate states would allow us to handle those transitions in the top-level Reconcile state machine, which would reduce complexity overall.

dhellmann:
I rebased this on master to get the image build fixes so quay would stop telling me it can't build the image from my branch when I update the PR.

pkg/apis/metalkube/v1alpha1/baremetalhost_types.go
```diff
@@ -323,7 +324,17 @@ func (host *BareMetalHost) getLabel(name string) string {
 	return host.Labels[name]
 }
 
-// SetHardwareProfile updates the HardwareProfileLabel and returns
+// NeedsHardwareProfile returns true if the profile is not set
+func (host *BareMetalHost) NeedsHardwareProfile() bool {
```
mhrivnak: Can you just elaborate on why these two methods are valuable?

dhellmann: I was trying not to encode the comparison to an empty string in the place where NeedsHardwareProfile() is called. It's trivial, but it's parallel to some of the other Needs*() methods.

mhrivnak: Gotcha. `""` is the zero value for a string in go, so while not explicitly an "unset" state, I think it's pretty close and regularly gets used as such. I think the code would be more idiomatic without these two methods, but that's subjective, and they're hardly showstoppers.

PR #122, opened 2019-04-06
Support changing the provisioned image

zaneb:
Previously to change the provisioned image, you would have to first set
the image URL to "" and wait for deprovisioning to complete before
reprovisioning with a new image URL. With this change, a change to the
image URL will cause the server to be deprovisioned and reprovisioned
with the new image.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -158,10 +158,7 @@ Several conditions must be met in order to initiate provisioning.
 1. The host `spec.image.url` field must contain a URL for a valid
    image file that is visible from within the cluster and from the
    host receiving the image.
-2. The host must not have an image provisioned, as reflected by the
-   `status.provisioning.image.URL` field being empty. To reuse an
-   existing host with a different image, deprovision the host first.
-3. The host must have `online` set to `true` so that the operator will
+2. The host must have `online` set to `true` so that the operator will
    keep the host powered on.
 
 To initiate deprovisioning, clear the image URL from the host spec.
--- a/pkg/apis/metalkube/v1alpha1/baremetalhost_types.go
+++ b/pkg/apis/metalkube/v1alpha1/baremetalhost_types.go
@@ -413,8 +413,6 @@ func (host *BareMetalHost) NeedsProvisioning() bool {
 		// We have an image set, but not provisioned.
 		return true
 	}
-	// FIXME(dhellmann): Compare the provisioned image against the one
-	// we are supposed to have to make sure they match.
 	return false
 }
 
@@ -427,7 +425,7 @@ func (host *BareMetalHost) NeedsDeprovisioning() bool {
 	if host.Spec.Image == nil {
 		return true
 	}
-	if host.Spec.Image.URL == "" {
+	if host.Spec.Image.URL != host.Status.Provisioning.Image.URL {
 		return true
 	}
 	return false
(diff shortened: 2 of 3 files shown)


hardys:
Interested to know how we can be very, very sure an image change won't rebuild all the nodes at once, do we get evacuation and rolling rebuild via the cluster/machine API layers?

zaneb:
I would have thought it was the responsibility of something at a higher level than the baremetal operator to ensure that we don't change all of the image URLs at once? The BaremetalMachine knows about just one machine, it's not its job to know what is going on in the cluster (or even that there is a cluster).

dhellmann:
Yeah, this fix is about recovering from an error if the initial settings are wrong. Orchestrating a cluster update is outside the scope of this operator. FWIW, I wouldn't expect the actuator to *ever* change the image setting on a host once it was successfully provisioned.

hardys:
Ack thanks - it'd be good to confirm exactly how that works in due course as I'm having flashbacks to an early TripleO bug where you could change a single Heat parameter and rebuild your entire cloud :)

dhellmann:
> Ack thanks - it'd be good to confirm exactly how that works in due course as I'm having flashbacks to an early TripleO bug where you could change a single Heat parameter and rebuild your entire cloud :)

The intent is to have something set the image as part of the provider spec for the MachineSet, so it is then copied to the Machine. The actuator would duplicate the settings from the Machine to the Host when the host is allocated. After that, the actuator would not modify the values on the host.

@mhrivnak , please correct me if I got any of that summary wrong.

dhellmann:
> Previously to change the provisioned image, you would have to first set
> the image URL to "" and wait for deprovisioning to complete before
> reprovisioning with a new image URL. With this change, a change to the
> image URL will cause the server to be deprovisioned and reprovisioned
> with the new image.

I mis-commented earlier, thinking this was a different change.

Does this actually trigger deprovisioning? Or does it simply write over the image that's there? Maybe Ironic does the deprovisioning automatically and that's why I'm not seeing it?

hardys:
> Does this actually trigger deprovisioning? Or does it simply write over the image that's there? Maybe Ironic does the deprovisioning automatically and that's why I'm not seeing it?

AFAIK Ironic nodes always have to be explicitly deprovisioned via an API call, just changing the properties or instance_info won't be destructive AFAIK.  Where we've had issues in the past is when layers above Ironic (e.g Nova and orchestration e.g Heat/Ansible) assume an image change means a rebuild.

hardys:
cc @stbenjam for info on the gophercloud workflow around this

zaneb:
> Does this actually trigger deprovisioning? Or does it simply write over the image that's there? Maybe Ironic does the deprovisioning automatically and that's why I'm not seeing it?

It triggers deprovisioning, by returning true from NeedsDeprovisioning() if the image is mismatched (including if the target image is "" but the machine is currently provisioned). Once deprovisioning is complete, NeedsProvisioning() will start returning true and the machine will get reprovisioned.

It may be that we don't actually want to allow the baremetal operator to do this. I'm happy to close this PR if that's the case. I just saw a TODO that I knew how to fix :)

dhellmann:
> > Does this actually trigger deprovisioning? Or does it simply write over the image that's there? Maybe Ironic does the deprovisioning automatically and that's why I'm not seeing it?
> 
> It triggers deprovisioning, by returning true from NeedsDeprovisioning() if the image is mismatched (including if the target image is "" but the machine is currently provisioned). Once deprovisioning is complete, NeedsProvisioning() will start returning true and the machine will get reprovisioned.
> 
> It may be that we don't actually want to allow the baremetal operator to do this. I'm happy to close this PR if that's the case. I just saw a TODO that I knew how to fix :)

Ah, of course, yeah.

I think it's fine to include this feature. I don't know if KNI will use it, but I would expect it to be needed as part of a more general provisioning tool.

dhellmann:
It would be useful to have a test for this case.

PR #123, opened 2019-04-08
State diagram updates

dhellmann:


docs/BaremetalHost_ProvisioningState.dot
```diff
@@ -1,5 +1,7 @@
 digraph BaremetalHost {
-    None -> Registering
+    None -> Discovered [label="(without BMC access details)"]
+    Discovered -> Registering [label="set BMC access details"]
```
zaneb: I think this could just be label=done for consistency.

docs/BaremetalHost_ProvisioningState.dot
```diff
@@ -13,6 +13,7 @@ digraph BaremetalHost {
     Inspecting -> Ready [label="done"]
 
     RegistrationError -> Deleting [label="!DeletionTimestamp.IsZero()"]
+    RegistrationError -> Registering [label="(user edits settings)"]
```
zaneb: Is there a succinct snippet of code to detect when the user has edited the settings that we could use as a label?

dhellmann:
> Is the intention here to document a spec for what we want to evolve the code towards? If so I think it looks great (although some more cleanup could also happen around the PreparingToProvision/MakingAvailable states, which your other PR already removes).
>
> However, my original intention with adding this diagram was to describe the code as it exists, and update it incrementally as the code changes (this would help in reviews, since in each case we could check 1. that the diagram change does accurately reflect the code change, and 2. that the new state diagram looks sane). That's also why I labelled the edges with actual code snippets. I wonder if we actually need two diagrams - perhaps a prescriptive one should live in the design docs repo and have the edges labelled with the transistions that are meaningful from the outside (e.g. "Delete requested" rather than `!DeletionTimestamp.IsZero()`).

This is a bit of both. Maybe we should focus on getting it to be something like what we think it ought to be, as a way to hash that out, and then we can make the code in my other PR match it? I'm happy to do that here in this repo.

zaneb:
That sounds good

docs/BaremetalHost_ProvisioningState.dot
dhellmann: Not really. When they edit the settings, we're just notified by having Reconcile() called again, and we figure out that there aren't any saved "good" credentials and try again.

docs/BaremetalHost_ProvisioningState.dot
dhellmann: I changed it to the same code snippet that I used earlier.

dhellmann:
The new version emphasizes the states other than delete a bit better. I think it's clearer, but if you all think we should keep the edges going to the single deleted node at the bottom I can remove that 1 commit.

dhellmann:
I added "discovered" as a steady state and rebased the series to avoid errors from quay.io builds.

PR #129, opened 2019-04-10
Demo unit tests

dhellmann:
Fix the demo provisioner based on the new logic for Reconcile() and add unit tests to ensure it stays up to date.

### DIFF
--- a/pkg/provisioner/demo/demo.go
+++ b/pkg/provisioner/demo/demo.go
@@ -1,7 +1,6 @@
 package demo
 
 import (
-	"fmt"
 	"time"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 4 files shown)


pkg/provisioner/demo/demo.go
```diff
@@ -59,15 +76,12 @@ func (p *demoProvisioner) ValidateManagementAccess() (result provisioner.Result,
...
-	case registrationErrorHost:
-		if p.host.SetErrorMessage("failed to register new host") {
-			p.log.Info("setting registration error")
-			p.publisher("RegistrationError", "Failed to register new host")
-			result.Dirty = true
-		}
+	case RegistrationErrorHost:
 		// We have set an error, so Reconcile() will stop
+		result.ErrorMessage = "failed to register new host"
+		p.log.Info("setting registration error")
```
zaneb: Was this change meant to be part of the next patch?
Does it even matter on GitHub?

dhellmann: Ugh, maybe, but probably not. I should stop trying to decompose these things and just go with the PR flow, I guess.

PR #133, opened 2019-04-11
DO NOT MERGE: Hard code root_device settings.

russellb:
This was a hard coded hack to test VM deployment of worker nodes.  This is posted for others to test if they'd like to, but this should be resolved with parameters from BareMetalHost.


hardys:
If folks want to test with this there's a build at https://quay.io/repository/shardy/baremetal-operator

PR #137, opened 2019-04-12
use the hardware profile to configure ironic to provision correctly

dhellmann:
Add some temporarily hacky logic to guess the hardware profile so we
can get a root device. The real profile will eventually come from
matching inspected inventory, when we implement that. In the mean
time, use the BMC type to detect virt vs. physical, and provide a
field in the host spec to override the guess of a profile if we get it
wrong.

Also rewrite the logic for setting individual host and image
properties so they are all checked and added or updated as required.


dhellmann:
This is not quite how I envisioned having hardware profiles (https://github.com/metalkube/metalkube-docs/pull/4) but I think it will do for now.

PR #141, opened 2019-04-15
Register master in an active state

dhellmann:
This PR adds a new active state, "externally provisioned" and places nodes in it when they have a machine reference but no image reference. This is how we expect master hosts to appear after they are registered with the cluster in which they are running.


pkg/apis/metalkube/v1alpha1/baremetalhost_types.go
```diff
@@ -439,6 +443,15 @@ func (host *BareMetalHost) WasProvisioned() bool {
 	return false
 }
 
+// WasExternallyProvisioned returns true when we think something else
+// is managing the image running o n the host.
```
rustyrobot: nit: s/o n/on/

dhellmann: fixed

PR #142, opened 2019-04-16
Align make-bm-worker with dev-script's version

yprokule:
Commit https://github.com/openshift-metalkube/dev-scripts/commit/470a7bcbdc76db9ac7e9cf4086db79b418215fcf#diff-6aaa4e12b6fddc1b93b9ee994b50a49b introduced
ability to register both master and worker nodes.
This change brings in custom changes to make-bm-worker utility
to the BMO's repo.

### DIFF
--- a/cmd/make-bm-worker/main.go
+++ b/cmd/make-bm-worker/main.go
@@ -29,14 +29,22 @@ spec:
   bmc:
     address: {{ .BMCAddress }}
     credentialsName: {{ .Name }}-bmc-secret
+{{- if .WithMachine }}
+  machineRef:
+    name: {{ .Machine }}
+    namespace: {{ .MachineNamespace }}
+{{- end }}
 `
 
 // TemplateArgs holds the arguments to pass to the template.
 type TemplateArgs struct {
-	Name            string
-	BMCAddress      string
-	EncodedUsername string
-	EncodedPassword string
+	Name             string
+	BMCAddress       string
+	EncodedUsername  string
+	EncodedPassword  string
+	WithMachine      bool
+	Machine          string
+	MachineNamespace string
 }
 
 func encodeToSecret(input string) string {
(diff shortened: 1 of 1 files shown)


cmd/make-bm-worker/main.go
```diff
@@ -9,6 +9,11 @@ import (
 	"text/template"
 )
 
+const (
+	instanceImageSource      = "http://172.22.0.1/images/rhcos-ootpa-latest.qcow2"
+	instanceImageChecksumURL = instanceImageSource + ".md5sum"
```
dhellmann: The image related information is specific to the dev-scripts setup and using this code to deploy RHCOS nodes. We don't want to put that in this upstream repo.

cmd/make-bm-worker/main.go
```diff
@@ -29,14 +34,35 @@ spec:
   bmc:
     address: {{ .BMCAddress }}
     credentialsName: {{ .Name }}-bmc-secret
+{{- if .MAC }}
+  bootMACAddress: {{ .MAC }}
+{{ end -}}{{- if .WithImage }}
+  userData:
+    namespace: openshift-machine-api
```
dhellmann: This setting is also specific to the environments we have downstream.

cmd/make-bm-worker/main.go
```diff
@@ -29,14 +34,35 @@ spec:
   bmc:
     address: {{ .BMCAddress }}
     credentialsName: {{ .Name }}-bmc-secret
+{{- if .MAC }}
+  bootMACAddress: {{ .MAC }}
```
dhellmann: We shouldn't need MAC addresses for bare metal servers.

cmd/make-bm-worker/main.go
```diff
@@ -29,14 +34,35 @@ spec:
...
+    namespace: openshift-machine-api
+    name: worker-user-data
+  image:
+    url: "{{ .ImageSourceURL }}"
+    checksum: "{{ .Checksum }}"
+{{ end -}}{{ if .WithMachine }}
+  machineRef:
+    name: {{ .Machine }}
+    namespace: {{ .MachineNamespace }}
+{{ end }}
```
dhellmann: The machine ref feature is OK.

cmd/make-bm-worker/main.go
yprokule: Updated

cmd/make-bm-worker/main.go
yprokule: Updated

Issue #139, opened 2019-04-14
Automatically determine IP address on the provisioning network 


russellb:
At the moment, `pkg/provisioner/ironic/ironic.go` includes a hard coded IP address for the baremetal-operator to use on its provisioning network:

```
		// FIXME(dhellmann): We need to get our IP on the
		// provisioning network from somewhere.
		driverInfo["deploy_kernel"] = "http://172.22.0.1/images/ironic-python-agent.kernel"
		driverInfo["deploy_ramdisk"] = "http://172.22.0.1/images/ironic-python-agent.initramfs"
```

As noted by the FIXME comment there, we need to add a way to get this IP that makes this easier to use different environments.

russellb:
I think the env vars PR (#143) solved this problem well enough

PR #143, opened 2019-04-16
add environment variables for configuring the operator

dhellmann:
We don't always want these values to change, but sometimes they must,
so provide a way to do it.

### DIFF
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"os"
 	"strings"
 	"time"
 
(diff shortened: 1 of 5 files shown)


dhellmann:
This is related to #139 but does not solve it in the way proposed there.

pkg/provisioner/ironic/ironic.go
```diff
@@ -29,15 +30,32 @@ var log = logf.Log.WithName("baremetalhost_ironic")
...
 	// See nodes.Node.PowerState for details
 	powerOn   = "power on"
 	powerOff  = "power off"
 	powerNone = "None"
 )
 
+func init() {
+	deployKernelURL = os.Getenv("DEPLOY_KERNEL_URL")
+	if deployKernelURL == "" {
+		panic("No DEPLOY_KERNEL_URL variable set")
```
mhrivnak: nit: I'm not a fan of `panic` unless stuff really hits the fan. It outputs a stacktrace along with the message, which isn't helpful to an end user. They just need to know that they should set the environment variable. I think it would be better to log a message and then `os.Exit(1)`.

dhellmann: Fixed

PR #145, opened 2019-04-17
improve support for externally provisioned hosts

dhellmann:
This PR includes 2 related changes to better support hosts that have been provisioned before the operator was told about them. The first gives the host information to ironic to ensure the hardware is monitored. The second makes it possible to delete externally provisioned hosts without deprovisioning them.


pkg/provisioner/ironic/ironic.go
```diff
@@ -794,6 +779,85 @@ func (p *ironicProvisioner) Deprovision(deleteIt bool) (result provisioner.Resul
...
+	p.log.Info("deleting host",
+		"ID", ironicNode.UUID,
+		"lastError", ironicNode.LastError,
+		"current", ironicNode.ProvisionState,
+		"target", ironicNode.TargetProvisionState,
+		"deploy step", ironicNode.DeployStep,
+	)
+
+	if ironicNode.ProvisionState == nodes.Available {
+		// Move back to manageable so we can delete it cleanly.
```
dtantsur: You can delete available nodes (it can be prevented, but it's allowed by default), there is no need for that.

pkg/provisioner/ironic/ironic.go
```diff
@@ -794,6 +779,85 @@ func (p *ironicProvisioner) Deprovision(deleteIt bool) (result provisioner.Resul
...
+		return p.changeNodeProvisionState(
+			ironicNode,
+			nodes.ProvisionStateOpts{Target: nodes.TargetManage},
+		)
+	} else if !ironicNode.Maintenance {
```
dtantsur: This will try to set maintenance for nodes in `manageable` state. I think you missed something like `ironicNode.ProvisionState != nodes.Manageable`. Well, you can also delete nodes in `enroll` state...

pkg/provisioner/ironic/ironic.go
dhellmann: I got an error when I tried it, and this transition is easy enough to do.

pkg/provisioner/ironic/ironic.go
dhellmann: Does setting the flag in any of those other states matter? It's going to delete the node, right?

pkg/provisioner/ironic/ironic.go
dtantsur: This is suspicious, something is going wrong. Available nodes can be deleted. Maybe you should make sure you don't keep `instance_uuid` on them when unprovisioning?

pkg/provisioner/ironic/ironic.go
dtantsur: It doesn't, mostly for consistency sake

dhellmann: OK. By the time execution gets to this point, there's really no turning back. The host object in k8s is going to be removed as soon as we release it. So I want to ensure that Ironic's database is cleaned up without a lot of fuss. There's a separate path for real deprovisioning still, so in the vast majority of cases we're going to follow the more complete and expected process.

pkg/provisioner/ironic/ironic.go
russellb: @dhellmann did you want to follow up on @dtantsur 's last suggestion here, to see if the instace_uuid prevented deleting the node?

dhellmann: I could do that. Will Ironic do something different during delete based on the state of the node?

dtantsur: Note deletion us literally just deleting a record from the database (not to be confused with `deleted` action, which is an unfortunate name for unprovisioning).

dhellmann: OK, that's pretty much what we want at this point. If the node is available, we haven't provisioned it so we don't need to clean it. And we want ironic not to argue with us when we say to delete it. I'll investigate whether removing the instance_uuid setting makes this possible without the state transition, but since this is working as it is I would like to proceed with this PR and do that other work separately.

dhellmann: Looking into ironic's code, I see that destroy_node() in the conductor [checks a configuration option](https://opendev.org/openstack/ironic/src/branch/master/ironic/conductor/manager.py#L2274) before allowing Available nodes to be deleted. I think it's safer to keep the explicit state transition here.

PR #146, opened 2019-04-17
make registration error a steady state

dhellmann:
If a host fails registration with the same error message a second
time, that means we have reconciled it twice with the same data and it
still does not work. Rather than keep trying, which toggles the host
between "registering" and "registration error" indefinitely, leave the
host in the "registration error" state.


dhellmann:
The test failure was a transient travis issue. This is working now.

PR #147, opened 2019-04-17
Add hardware introspection support using ironic-inspector

zaneb:
Note that https://github.com/metalkube/baremetal-operator/commit/fb54f2943004e7364cffd1f438768462adb1192e is a fix to the vendored gophercloud that is awaiting upstream acceptance in https://github.com/gophercloud/gophercloud/pull/1548

### DIFF
--- a/pkg/controller/baremetalhost/baremetalhost_controller.go
+++ b/pkg/controller/baremetalhost/baremetalhost_controller.go
@@ -471,10 +471,11 @@ func (r *ReconcileBareMetalHost) actionRegistering(prov provisioner.Provisioner,
 // Ensure we have the information about the hardware on the host.
 func (r *ReconcileBareMetalHost) actionInspecting(prov provisioner.Provisioner, info *reconcileInfo) (result reconcile.Result, err error) {
 	var provResult provisioner.Result
+	var details *metal3v1alpha1.HardwareDetails
 
 	info.log.Info("inspecting hardware")
 
-	provResult, err = prov.InspectHardware()
+	provResult, details, err = prov.InspectHardware()
 	if err != nil {
 		return result, errors.Wrap(err, "hardware inspection failed")
 	}
@@ -486,21 +487,17 @@ func (r *ReconcileBareMetalHost) actionInspecting(prov provisioner.Provisioner, 
 		return result, nil
 	}
 
+	if details != nil {
+		info.host.Status.HardwareDetails = details
+		result.Requeue = true
+		return result, nil
+	}
+
 	if provResult.Dirty {
 		info.host.ClearError()
 		result.Requeue = true
 		result.RequeueAfter = provResult.RequeueAfter
-		return result, nil
 	}
-
-	// FIXME(dhellmann): Since we test the HardwareDetails pointer
-	// before calling function, perhaps it makes sense to have
-	// InspectHardware() return a value and store it here in this
-	// function. That would eliminate duplication in the provisioners
-	// and make this phase consistent with the structure of others.
-
-	// Line up a requeue if we could set the hardware profile
-	result.Requeue = info.host.NeedsHardwareProfile()
 
 	return result, nil
 }
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -424,38 +424,41 @@ func getHardwareDetails(data *introspection.Data) *metal3v1alpha1.HardwareDetail
 // details of devices discovered on the hardware. It may be called
 // multiple times, and should return true for its dirty flag until the
 // inspection is completed.
-func (p *ironicProvisioner) InspectHardware() (result provisioner.Result, err error) {
+func (p *ironicProvisioner) InspectHardware() (result provisioner.Result, details *metal3v1alpha1.HardwareDetails, err error) {
 	p.log.Info("inspecting hardware", "status", p.host.OperationalStatus())
 
 	ironicNode, err := p.findExistingHost()
 	if err != nil {
-		return result, errors.Wrap(err, "failed to find existing host")
+		err = errors.Wrap(err, "failed to find existing host")
+		return
 	}
 	if ironicNode == nil {
-		return result, fmt.Errorf("no ironic node for host")
+		return result, nil, fmt.Errorf("no ironic node for host")
 	}
 
 	status, err := introspection.GetIntrospectionStatus(p.inspector, ironicNode.UUID).Extract()
 	if err != nil {
 		if _, isNotFound := err.(gophercloud.ErrDefault404); isNotFound {
 			p.log.Info("starting new hardware inspection")
-			return p.changeNodeProvisionState(
+			result, err = p.changeNodeProvisionState(
 				ironicNode,
 				nodes.ProvisionStateOpts{Target: nodes.TargetInspect},
 			)
+			return
 		}
-		return result, errors.Wrap(err, "failed to extract hardware inspection status")
+		err = errors.Wrap(err, "failed to extract hardware inspection status")
+		return
 	}
 	if !status.Finished {
 		p.log.Info("inspection in progress", "started_at", status.StartedAt)
 		result.Dirty = true // make sure we check back
 		result.RequeueAfter = introspectionRequeueDelay
-		return result, nil
+		return
 	}
 	if status.Error != "" {
 		p.log.Info("inspection failed", "error", status.Error)
 		result.ErrorMessage = status.Error
-		return result, nil
+		return
 	}
 
 	// Introspection is ongoing
@@ -463,15 +466,14 @@ func (p *ironicProvisioner) InspectHardware() (result provisioner.Result, err er
 	introData := introspection.GetIntrospectionData(p.inspector, ironicNode.UUID)
 	data, err := introData.Extract()
 	if err != nil {
-		return result, errors.Wrap(err, "failed to retrieve hardware introspection data")
+		err = errors.Wrap(err, "failed to retrieve hardware introspection data")
+		return
 	}
 	p.log.Info("received introspection data", "data", introData.Body)
 
-	p.host.Status.HardwareDetails = getHardwareDetails(data)
+	details = getHardwareDetails(data)
 	p.publisher("InspectionComplete", "Hardware inspection completed")
-	result.Dirty = true
-
-	return result, nil
+	return
 }
 
 // UpdateHardwareState fetches the latest hardware state of the server
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -14,6 +14,7 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/baremetal/v1/nodes"
 	"github.com/gophercloud/gophercloud/openstack/baremetal/v1/ports"
 	noauthintrospection "github.com/gophercloud/gophercloud/openstack/baremetalintrospection/noauth"
+	"github.com/gophercloud/gophercloud/openstack/baremetalintrospection/v1/introspection"
 
 	nodeutils "github.com/gophercloud/utils/openstack/baremetal/v1/nodes"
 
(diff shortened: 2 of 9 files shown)


pkg/provisioner/ironic/ironic.go
```diff
@@ -334,54 +416,51 @@ func (p *ironicProvisioner) InspectHardware() (result provisioner.Result, err er
...
+		p.log.Info("getting hardware details from inspection")
+		introData := introspection.GetIntrospectionData(p.inspector, ironicNode.UUID)
+		data, err := introData.Extract()
+		if err != nil {
+			return result, errors.Wrap(err, "failed to retrieve hardware introspection data")
+		}
+		if jsonData, err := json.Marshal(introData.Body); err == nil {
+			p.log.Info("Received introspection data", "data", jsonData)
+		} else {
+			p.log.Info("Error marshalling JSON introspection data", "error", err)
```
dhellmann: Should this error case return right away?

pkg/provisioner/ironic/ironic.go
```diff
@@ -319,13 +341,73 @@ func (p *ironicProvisioner) changeNodeProvisionState(ironicNode *nodes.Node, opt
...
+			SizeGiB: metalkubev1alpha1.GiB(disk.Size / (1024 * 1024 * 1024)),
+			Model:   fmt.Sprintf("%s %s", disk.Vendor, disk.Model),
+		}
+	}
+	return storage
+}
+
+func getCPUDetails(cpudata *introspection.CPUType) []metalkubev1alpha1.CPU {
+	var freq float64
+	fmt.Sscanf(cpudata.Frequency, "%fGHz", &freq)
```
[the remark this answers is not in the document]
zaneb: I was going to do that, but it turns out there's nothing to do - if it fails then freq will be 0, which seems like the only sane thing you could set it to anyway.

pkg/provisioner/ironic/ironic.go
```diff
@@ -319,13 +341,73 @@ func (p *ironicProvisioner) changeNodeProvisionState(ironicNode *nodes.Node, opt
...
+	return nics
+}
+
+func getStorageDetails(diskdata []introspection.RootDiskType) []metalkubev1alpha1.Storage {
+	storage := make([]metalkubev1alpha1.Storage, len(diskdata))
+	for i, disk := range diskdata {
+		storage[i] = metalkubev1alpha1.Storage{
+			Name:    disk.Name,
+			Type:    map[bool]string{true: "HDD", false: "SSD"}[disk.Rotational],
+			SizeGiB: metalkubev1alpha1.GiB(disk.Size / (1024 * 1024 * 1024)),
```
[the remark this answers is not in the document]
zaneb: Yes, we should.
Also we should report the RAM size in MB imho.

pkg/provisioner/ironic/ironic.go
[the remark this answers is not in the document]
dhellmann: That makes sense. I guess there's no point in logging a debug message.

pkg/provisioner/ironic/ironic.go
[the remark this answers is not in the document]
dhellmann: OK, let's go ahead and change the data type here (unless you'd rather do that as a separate PR). The structures I threw together were meant to be a starting point, and are likely wrong or incomplete in lots of ways.

pkg/provisioner/ironic/ironic.go
```diff
@@ -319,13 +341,73 @@ func (p *ironicProvisioner) changeNodeProvisionState(ironicNode *nodes.Node, opt
...
+func getCPUDetails(cpudata *introspection.CPUType) []metalkubev1alpha1.CPU {
+	var freq float64
+	fmt.Sscanf(cpudata.Frequency, "%fGHz", &freq)
+	cpu := metalkubev1alpha1.CPU{
+		Type:     cpudata.Architecture,
+		SpeedGHz: metalkubev1alpha1.GHz(freq),
+	}
+
+	cpus := make([]metalkubev1alpha1.CPU, cpudata.Count)
+	for i := range cpus {
```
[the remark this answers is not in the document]
zaneb: The data we get from ironic-inspector doesn't allow for the possibility that they could be different, so even if they could we wouldn't have a way to get that data. So yes, we should get rid of the array.

pkg/provisioner/ironic/ironic.go
```diff
@@ -319,13 +341,73 @@ func (p *ironicProvisioner) changeNodeProvisionState(ironicNode *nodes.Node, opt
...
+
 // InspectHardware updates the HardwareDetails field of the host with
 // details of devices discovered on the hardware. It may be called
 // multiple times, and should return true for its dirty flag until the
 // inspection is completed.
 func (p *ironicProvisioner) InspectHardware() (result provisioner.Result, err error) {
 	p.log.Info("inspecting hardware", "status", p.host.OperationalStatus())
 
+	// Introspection not started
+	if p.host.Status.Provisioning.State != metalkubev1alpha1.StateInspecting {
```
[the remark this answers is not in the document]
zaneb: I don't disagree that it should, but nothing in #121 attempts to do that AFAICT. That seems like something to do in a separate PR?

pkg/provisioner/ironic/ironic.go
zaneb: It's only an error in the logging, so I'd say no.

pkg/provisioner/ironic/ironic.go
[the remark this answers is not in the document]
dhellmann: Maybe I pointed to the wrong PR. I don't think ironic.go sets Provisioning.State anywhere any more, so I was suggesting we not add a new place where it does.

pkg/provisioner/ironic/ironic.go
dhellmann: It would be an error from json.Marshal(), right?

pkg/provisioner/ironic/ironic.go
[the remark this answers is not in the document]
zaneb: It was still being set on line 368, this just moves it to line 351.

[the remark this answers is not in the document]
dhellmann: OK, we can fix that in another PR then

pkg/provisioner/ironic/ironic.go
zaneb: Yes, but we're marshalling it purely so that we can log the data in JSON format. Which we need to do purely because gophercloud stores the body as deserialised data, even though it was received as serialised JSON. IOW it can't fail and even if it did, it only means that you wouldn't get to see the log message.

I think this whole JSON serialisation thing might actually be a mistake though, because the logging code seems to base64 encode the JSON. It is a structured logging thing, so maybe it would let us just pass the deserialised objects?

dhellmann: Ah,  I see, I was not reading the variable names closely and saw the call to getHardwareDetails() a few lines down and assumed it was using the results of the Marshall call.

Yes, I think if you pass the structured object to the log call it will figure out how to do the right thing with it. That may be quite a lot of data, though. Do we really want to log all of it in one go?

zaneb: It is a ton of data :/ I was tempted to take the log out before posting the PR, but this will likely be invaluable for debugging - every time somebody installs on some new weird hardware and it doesn't show up the way we want, we'll be able to just ask them for the log.

dhellmann: What if we log the sections as we try to interpret them? "Here's all the CPU info we found: blob", etc.?

pkg/provisioner/ironic/ironic.go
```diff
@@ -319,13 +341,73 @@ func (p *ironicProvisioner) changeNodeProvisionState(ironicNode *nodes.Node, opt
...
+	}
+	return nics
+}
+
+func getStorageDetails(diskdata []introspection.RootDiskType) []metalkubev1alpha1.Storage {
+	storage := make([]metalkubev1alpha1.Storage, len(diskdata))
+	for i, disk := range diskdata {
+		storage[i] = metalkubev1alpha1.Storage{
+			Name:    disk.Name,
+			Type:    map[bool]string{true: "HDD", false: "SSD"}[disk.Rotational],
```
alanmeadows: Is it safe to say the type is either "HDD" or "SSD"? I know -- we live in the modern world but I assume there is some thought behind ironic's `rotational` bool.  I'm just speculating, but would the inspector pick up a USB thumb drive someone left in? If so, ironic would correctly call that non-rotational.  The code here would call that an SSD?

zaneb: Are you suggesting that thumb drives are not SSDs? ;)
`rotational` comes straight from udev so it's probably reasonably universal.

pkg/provisioner/ironic/ironic.go
zaneb: I'll look into that. The data is quite spread around though, so not sure if it is feasible.

mhrivnak: Is there a way to have ironic log it in its own logs? That would at least keep the "noise" separate from this controller.

Or we could write it out to disk in the container and just ask someone to exec into the pod if they really need the data.

dhellmann: Let's leave it in the operator logs for now. I know we have more data to pull out of the structure, and it's going to make development easier. We can clean it up before we go GA.

pkg/provisioner/ironic/ironic.go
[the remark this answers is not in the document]
zaneb: Actually I'm wrong, it was added in the previous patch (at least after rebasing). I see now where it's getting set in #121 (it wasn't where I expected, and I think I thought it wasn't happening because it no longer records an event). I'll squash those two patches and make sure we're doing all state changes only in the controller.

pkg/provisioner/ironic/ironic.go
zaneb: +1 for clean it up later

zaneb:
This is working for me and addresses the functional comments. Let's leave updating the content of the BareMetalHost resource to future PRs.

pkg/provisioner/ironic/ironic.go
```diff
@@ -345,74 +365,118 @@ func (p *ironicProvisioner) changeNodeProvisionState(ironicNode *nodes.Node, opt
...
-						Type:     "x86",
-						SpeedGHz: 3,
-					},
-				},
+	status, err := introspection.GetIntrospectionStatus(p.inspector, ironicNode.UUID).Extract()
+	if err != nil {
+		if _, isNotFound := err.(gophercloud.ErrDefault404); isNotFound {
+			p.log.Info("starting new hardware inspection")
+			manageBoot := true
+			start := introspection.StartIntrospection(p.inspector,
```
dtantsur: You should use ironic's `inspect` transition rather than ironic-inspector API directly here. It has many nice benefits like actually updating the Node's `provision_state`.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -477,22 +478,18 @@ func (r *ReconcileBareMetalHost) actionInspecting(prov provisioner.Provisioner,
 		return result, nil
 	}
 
+	if details != nil {
+		info.host.Status.HardwareDetails = details
+		result.Requeue = true
```
dhellmann: Should this set RequeueAfter as well?

pkg/provisioner/ironic/ironic.go
```diff
@@ -58,6 +62,11 @@ func init() {
 		log.Error(fmt.Errorf("No IRONIC_ENDPOINT variable set"), "Cannot start")
 		os.Exit(1)
 	}
+	inspectorEndpoint = os.Getenv("IRONIC_INSPECTOR_ENDPOINT")
+	if inspectorEndpoint == "" {
+		log.Error(fmt.Errorf("No IRONIC_INSPECTOR_ENDPOINT variable set"), "Cannot start")
```
dhellmann: This needs to use fmt.Fprintf(). See #153 for details.

pkg/controller/baremetalhost/baremetalhost_controller.go
zaneb: I don't see any reason to delay - we're not waiting for anything external at this point. We just want to write the new data and then requeue immediately so that we can check if there are any other steps that need to be taken.

dhellmann:
I think this got caught up in the rename, so now it needs a rebase.

pkg/controller/baremetalhost/baremetalhost_controller.go
dhellmann: Yeah, that makes sense.

PR #153, opened 2019-04-23
switch startup error reporting to Fprintf()

dhellmann:
Logging is not configured when init() is running, so using the log API
to report errors means they are not output at all. Switch to simple
print statements.


russellb:
tested - works for me.  Thanks!

PR #156, opened 2019-04-24
Update dev documents by adding required env variables

longkb:
When we try to bring **baremetal-operator** up via **operator-sdk** CLI, it raises an error due to the missing of some environment variables to tell it where ironic is. This PR aims to update the doc by including these env variables.


russellb:
Thank you for the patch!

PR #157, opened 2019-04-24
Add a dell-raid profile.

russellb:
Someone reported trying to deploy a host where the root disk needed to
be 0:2:0:0.  Add another built in profile for this.

### DIFF
--- a/pkg/hardware/profile.go
+++ b/pkg/hardware/profile.go
@@ -76,6 +76,16 @@ func init() {
 		CPUArch: "x86_64",
 	}
 
+	profiles["dell-raid"] = Profile{
+		Name: "dell-raid",
+		RootDeviceHints: RootDeviceHints{
+			HCTL: "0:2:0:0",
+		},
+		RootGB:  10,
+		LocalGB: 50,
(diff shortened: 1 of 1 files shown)


pkg/hardware/profile.go
```diff
@@ -76,6 +76,16 @@ func init() {
 		CPUArch: "x86_64",
 	}
 
+	profiles["hctl-0-2-0-0"] = Profile{
```
dhellmann: The goal was to have the profile names be more meaningful (or generic?) than "use the second disk". Is there something about this hardware that we could use to describe it other than the HCTL entry?

sreichar: This is the first disk behind a per on teh del r720xd

russellb: "dell2" ?  :-)

I figured all of these were just hacks until we add defining these dynamically via CRs, right?

dhellmann: Maybe "dell-raid"?

russellb: done!

PR #159, opened 2019-04-25
Add hardwareProfile to the bmhost

### LINKED ISSUE #158: missing 'hardwareProfile'
https://github.com/metal3-io/baremetal-operator/blob/b72e991c3409e44ca471bb4d27a02ac092c83248/cmd/make-bm-worker/main.go#L27

For baremetal hosts, it is required to have a 'hardwareProfile' in the spec section.

e-minguez:
Fixes #158

Please bear with me I have 0 golang experience

### DIFF
--- a/cmd/make-bm-worker/main.go
+++ b/cmd/make-bm-worker/main.go
@@ -26,6 +26,9 @@ metadata:
   name: {{ .Name }}
 spec:
   online: true
+{{- if .WithHardwareProfile }}
+  hardwareProfile: {{ .HardwareProfile }}
+{{- end }}
   bmc:
     address: {{ .BMCAddress }}
     credentialsName: {{ .Name }}-bmc-secret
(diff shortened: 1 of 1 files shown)


cmd/make-bm-worker/main.go
```diff
@@ -26,6 +26,7 @@ metadata:
   name: {{ .Name }}
 spec:
   online: true
+	hardwareProfile: {{ .HardwareProfile }}
```
elfosardo: fix indentation here

e-minguez: Fixed (I believe)

cmd/make-bm-worker/main.go
```diff
@@ -81,11 +84,16 @@ func main() {
 		os.Exit(1)
 	}
 
+	if *hardwareProfile == "" {
+		*hardwareProfile = "libvirt"
```
russellb: Let’s leave it empty if not specified and use the default

e-minguez: Fixed!

cmd/make-bm-worker/main.go
```diff
@@ -26,6 +26,7 @@ metadata:
   name: {{ .Name }}
 spec:
   online: true
+  hardwareProfile: {{ .HardwareProfile }}
```
russellb: You can make this conditional where hardwareProfile doesn’t show up at all if it’s empty. See some examples in the dev-scripts version

PR #165, opened 2019-04-25
handle inspection states during delete

dhellmann:


pkg/provisioner/ironic/ironic.go
```diff
@@ -931,6 +931,12 @@ func (p *ironicProvisioner) Deprovision(deleteIt bool) (result provisioner.Resul
 			nodes.ProvisionStateOpts{Target: nodes.TargetManage},
 		)
 
+	case nodes.Inpsecting, nodes.InspectWait:
```
dtantsur: typo: `Inspecting`?

pkg/provisioner/ironic/ironic.go
```diff
@@ -931,6 +931,12 @@ func (p *ironicProvisioner) Deprovision(deleteIt bool) (result provisioner.Resul
...
+		p.log.Info("waiting for inspection to complete")
```
dtantsur: FYI inspection can be aborted

pkg/provisioner/ironic/ironic.go
dhellmann: The typo is in the gophercloud code. When we update to a new version, the const name will be fixed and we can update this spot, too.

pkg/provisioner/ironic/ironic.go
dhellmann: Based on https://docs.openstack.org/ironic/latest/_images/states.svg it looks like I have to tell it to "abort" and then the node will enter the "inspect failed" state and I can then "manage" it to move it back to manageable. Is that right?

dhellmann: The latest version does try to abort. It seems to be told the node is busy a lot, but it does eventually work.

pkg/provisioner/ironic/ironic.go
```diff
@@ -931,6 +931,26 @@ func (p *ironicProvisioner) Deprovision(deleteIt bool) (result provisioner.Resul
...
+
+	case nodes.InspectWait:
+		p.log.Info("cancelling inspection")
+		return p.changeNodeProvisionState(
+			ironicNode,
+			nodes.ProvisionStateOpts{Target: nodes.TargetAbort},
+		)
+
+	case nodes.InspectFail:
+		p.log.Info("inspection cancelled")
```
zaneb: It can fail for other reasons as well, so this log may not always reflect what actually happened.

dhellmann: I updated the message to say "failed or cancelled".

pkg/provisioner/ironic/ironic.go
```diff
@@ -931,6 +931,26 @@ func (p *ironicProvisioner) Deprovision(deleteIt bool) (result provisioner.Resul
 			nodes.ProvisionStateOpts{Target: nodes.TargetManage},
 		)
 
+	case nodes.Inpsecting:
```
elfosardo: s/Inpsecting/Inspecting

pkg/provisioner/ironic/ironic.go
elfosardo: @dhellmann that is correct

pkg/provisioner/ironic/ironic.go
```diff
@@ -931,6 +931,26 @@ func (p *ironicProvisioner) Deprovision(deleteIt bool) (result provisioner.Resul
...
+		p.log.Info("waiting for inspection to complete")
+		result.Dirty = true
+		result.RequeueAfter = introspectionRequeueDelay
+		return result, nil
+
+	case nodes.InspectWait:
```
elfosardo: FYI if the node is in InspectWait, it still can successfully finish inspection and go to manageable

pkg/provisioner/ironic/ironic.go
elfosardo: nvm, didn't see the comment above

elfosardo: btw I just checked and it seems fixed in gophercloud
https://github.com/gophercloud/gophercloud/blob/master/openstack/baremetal/v1/nodes/requests.go#L37

pkg/provisioner/ironic/ironic.go
dhellmann: When we're deleting the host in k8s, we want to tell Ironic to stop doing whatever it is doing, power the node off, and then delete the node record. We don't want it to finish, because nothing is going to use the data.

pkg/provisioner/ironic/ironic.go
dhellmann: OK, good. I will update that in a separate PR.

zaneb: Yes, I fixed it a couple of weeks ago.

PR #166, opened 2019-04-25
store the CPU model name

dhellmann:


dhellmann:
> Should also set in the fixture and demo provisioners?

Done

PR #168, opened 2019-04-25
add vendor and serial number for storage devices

dhellmann:


pkg/provisioner/ironic/ironic.go
```diff
@@ -387,10 +387,12 @@ func getStorageDetails(diskdata []introspection.RootDiskType) []metal3v1alpha1.S
...
-			Name:    disk.Name,
-			Type:    map[bool]string{true: "HDD", false: "SSD"}[disk.Rotational],
-			SizeGiB: metal3v1alpha1.GiB(disk.Size / (1024 * 1024 * 1024)),
-			Model:   fmt.Sprintf("%s %s", disk.Vendor, disk.Model),
+			Name:         disk.Name,
+			Type:         map[bool]string{true: "HDD", false: "SSD"}[disk.Rotational],
+			SizeGiB:      metal3v1alpha1.GiB(disk.Size / (1024 * 1024 * 1024)),
+			Vendor:       disk.Vendor,
+			Model:        disk.Model,
+			SerialNumber: disk.Serial,
```
zaneb: It would be interesting to know whether potential consumers of this are expecting the serial number or the WWN. (TBH I'm unclear on the relationship between them - possibly the WWN is just the vendor indentifier + the serial number?)

dhellmann: My understanding was that they wanted serial number visible to make replacing bad components in the hardware easier. I thought WWN was more of a fibre channel or iSCSI identifier similar to a MAC, rather than something that could be used for part replacement. That said, there's no reason we can't have both in the data model.

dhellmann: I went ahead and added all of the fields that are available through the inspector.

PR #169, opened 2019-04-29
Clear *provisionNet* to get MAC address

longkb:
Current, *provisionNet* is set to *provisioning* as default values that makes the our *main.go* cannot get the MAC address [1] of virt-host. This PR intends to clear this default to make our tool work smoothly.

[1] https://github.com/metal3-io/baremetal-operator/blob/master/cmd/make-virt-host/main.go#L161-L163

### DIFF
--- a/cmd/make-virt-host/main.go
+++ b/cmd/make-virt-host/main.go
@@ -118,7 +118,7 @@ type VBMC struct {
 
 func main() {
 	var provisionNet = flag.String(
-		"provision-net", "provisioning", "use the MAC on this network")
+		"provision-net", "", "use the MAC on this network")
 	var machine = flag.String(
 		"machine", "", "specify name of a related, existing, machine to link")
 	var machineNamespace = flag.String(


dhellmann:
The change looks OK, but I'm not sure how it fixes the problem you describe. Setting the default to "" would require all users to always pass the -provision option, right? Or is the network name in your case ""?

longkb:
> The change looks OK, but I'm not sure how it fixes the problem you describe. Setting the default to "" would require all users to always pass the -provision option, right? Or is the network name in your case ""?
No, it is Bridge's name that virtual host is attached [1]. If you set the default is "", the users is no need to always enter the --provision option as you said :) And if you keep this default value, the error will be **Could not find MAC for bm0 on network provisioning**

[1] https://github.com/metal3-io/baremetal-operator/blob/master/cmd/make-virt-host/main.go#L161

dhellmann:
> > The change looks OK, but I'm not sure how it fixes the problem you describe. Setting the default to "" would require all users to always pass the -provision option, right? Or is the network name in your case ""?
> 
> No, it is Bridge's name that virtual host is attached [1]. If you set the default is "", the users is no need to always enter the --provision option as you said :) And if you keep this default value, the error will be **Could not find MAC for bm0 on network provisioning**
> 
> [1] https://github.com/metal3-io/baremetal-operator/blob/master/cmd/make-virt-host/main.go#L161

If the default is set to "" how does the tool find the MAC address of the VM? I guess the bridge has a name other than "provisioning" and that is not "", but how does the tool know what that name is?

longkb:
> 
> 
> > > The change looks OK, but I'm not sure how it fixes the problem you describe. Setting the default to "" would require all users to always pass the -provision option, right? Or is the network name in your case ""?
> > 
> > 
> > No, it is Bridge's name that virtual host is attached [1]. If you set the default is "", the users is no need to always enter the --provision option as you said :) And if you keep this default value, the error will be **Could not find MAC for bm0 on network provisioning**
> > [1] https://github.com/metal3-io/baremetal-operator/blob/master/cmd/make-virt-host/main.go#L161
> 
> If the default is set to "" how does the tool find the MAC address of the VM? I guess the bridge has a name other than "provisioning" and that is not "", but how does the tool know what that name is?

+1 You was right. 
My VM does not attached to any bridge, so bridge's name is empty. 
I think it would be better if we have a script to config provisioning network in local env :)

dhellmann:
OK, good, I'm glad you figured out what was happening with the VM. 

The defaults here are set up to work with the environment created by https://github.com/metal3-io/metal3-dev-env or https://github.com/openshift-metalkube/dev-scripts/ to give us some consistency.

PR #174, opened 2019-05-01
update make-bm-worker tool to include bootMacAddress

rlopez133:
Currently the make-bm-worker tool does not include the bootMACAddress that is required in order to specify the provisioning NIC when adding a worker node to your KNI deployment. This code includes the bootMACAddress as a field that needs to be filled out when using the bm-worker tool.


cmd/make-bm-worker/main.go
```diff
@@ -39,13 +40,18 @@ func main() {
 		fmt.Fprintf(os.Stderr, "Missing -address argument\n")
 		os.Exit(1)
 	}
+	if *macAddress == "" {
```
dhellmann: The MAC address is not always required. In fact, when inspection is available it won't be required for any bare metal systems. So I think this option should stay optional.

russellb: It's discovery that makes this optional, not introspection, right?

rlopez133: @dhellmann - when troubleshooting adding a worker, this is one thing @russellb mentioned to me that I was missing within my yaml file. The MAC address of this host for its interface on the provisioning network. I guess I'm unsure then what I would or would not need to include it.

dhellmann: > It's discovery that makes this optional, not introspection, right?

Introspection happens on discovered hosts, but it now also happens on registered hosts. My understanding from @dtantsur and @juliakreger is that if the user enters the BMC settings and we power it on and do inspection, the user does not have to enter the MAC.

dhellmann: > @dhellmann - when troubleshooting adding a worker, this is one thing @russellb mentioned to me that I was missing within my yaml file. The MAC address of this host for its interface on the provisioning network. I guess I'm unsure then what I would or would not need to include it.

My understanding is that you only have to enter the value for VMs or in cases where inspection isn't running (so for older versions of the baremetal-operator).

cmd/make-bm-worker/main.go
```diff
@@ -14,6 +14,7 @@ func main() {
 	var password = flag.String("password", "", "password for BMC")
 	var bmcAddress = flag.String("address", "", "address URL for BMC")
 	var hardwareProfile = flag.String("hardwareprofile", "", "hardwareProfile to be used")
+	var macAddress = flag.String("bootmacaddress", "", "bootMACAddress to be used")
```
russellb: There's another version of this tool in dev-scripts, which includes "-boot-mac".  Can we use the same name to match?

rlopez133: Done.

cmd/make-bm-worker/main.go
russellb: seems like the MAC is still needed to be able to configure the PXE environment so it'll boot into the IPA image for introspection?

anyway, seems fine to make this optional in the script

rlopez133: @dhellmann - I went ahead and removed bootmac from being required to being optional.

russellb:
Feedback seems addressed, and works for me!

russellb:
One last thing ... can you squash these commits down to 1 commit?  You'd do something like `git rebase -i origin/master` and then "squash" all but one commit.

cmd/make-bm-worker/main.go
dhellmann: > seems like the MAC is still needed to be able to configure the PXE environment so it'll boot into the IPA image for introspection?
> 
> anyway, seems fine to make this optional in the script

The DHCP server is configured to boot the IPA image for any MAC it doesn't recognize. That allows discovery to work, and collect the MAC address(es) of the host. Those values can then be used to reconfigure the DHCP server for that one host to be given more specific instructions (it is possible to return a PXE response to boot directly from the local disk, for example), although I don't know if we're actually doing that.

rlopez133:
Feel

PR #175, opened 2019-05-02
Add short names for BareMetalHost.

russellb:
Add some short names for use at the command line that are a few less
keystrokes than "baremetalhost".


dhellmann:
Nice, I didn't realize short names were a feature!

PR #196, opened 2019-05-27
Minor API docs formatting fix

markmc:
As currently written, the markdown is formatted as if the operational-status values
are other labels.


markmc:
Ok, operational status isn't even a label any more - working on a separate PR to fix that

PR #201, opened 2019-05-28
WIP: fix the e2e tests

markmc:
It looks like these tests have gone stale - perhaps they have less importance now that many of the tests are moved to the unit tests, but I got sucked into trying to fix them ...

Notes:

* The single test still doesn't actually pass - it times out waiting for inspection to complete
* I never figured out why the secret data is getting double-encoded (i.e. why the `base64.StdEncoding.EncodeToString()` appears not to be needed)
* I added a way to use the virtual nodes created by metal3-dev-env - not terribly easy to use though

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -12,6 +12,12 @@ oc --as system:admin apply -f test/e2e/role_binding.yaml
 
 ### Run the e2e tests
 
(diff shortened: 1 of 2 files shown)


kanwar-saad:

nordixinfra:

PR #205, opened 2019-05-31
Update Ironic image tag and Ironic-inspector shared folder

longkb:
Podman is trying to pull image with tag latest but there is no this tag at this moment here: https://quay.io/repository/metal3-io/ironic?tag=latest&tab=tags. Only **master** and **no-cleaning** are available.
This PR also configure a shared folder for ironic-inspector container


russellb:
Thank you!

PR #213, opened 2019-06-11
Add hostname from introspection to hardware details

bcrochet:


bcrochet:
Depends on https://review.opendev.org/#/c/663991/

pkg/provisioner/ironic/ironic.go
```diff
@@ -429,6 +429,7 @@ func getHardwareDetails(data *introspection.Data) *metal3v1alpha1.HardwareDetail
 	details.NIC = getNICDetails(data.Inventory.Interfaces)
 	details.Storage = getStorageDetails(data.Inventory.Disks)
 	details.CPU = getCPUDetails(&data.Inventory.CPU)
+	details.Hostname = &data.Hostname
```
dhellmann: I think you don't want the & here, because both types are just strings and you can assign between them directly without worrying about pointers.

dhellmann: Oh, and to have access to the new field we will have to update the gophercloud dependency. Ping me on slack if you need help with that.

zaneb: Done in PR #218

PR #219, opened 2019-06-13
Improve HardwareDetails

zaneb:
Various tweaks to how we display the inspection data.


pkg/apis/metal3/v1alpha1/baremetalhost_types.go
```diff
@@ -204,6 +222,13 @@ type Storage struct {
 	HCTL string `json:"hctl,omitempty"`
 }
 
+type VLANID int16
+
+type VLAN struct {
+	ID   VLANID `json:"id"`
+	Name string `json:"name,omitempty"`
```
dhellmann: What is a VLAN name? Does the switch report that?

zaneb: Yes, the switch usually allows you to configure a name for the VLAN (by default it'll be something boring like `vlan4092`), which it then reports over LLDP.

PR #226, opened 2019-06-19
ServiceMonitor for ironic-prometheus-exporter

iurygregory:
This PR adds the ServiceMonitor for the ironic-promeheus-exporter

We need a `yaml` to NameSpace deploy so we can add the label `openshift.io/cluster-monitoring=true`
like in [1]

[1] https://github.com/openshift/cluster-monitoring-operator/blob/master/manifests/01-namespace.yaml#L8

### DIFF
--- a/deploy/baremetal-operator_servicemonitor.yaml
+++ b/deploy/baremetal-operator_servicemonitor.yaml
@@ -1,0 +1,15 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: baremetal-operator
+  namespace: openshift-machine-api
+spec:
+  endpoints:
+  - port: "5001-tcp"
+    scheme: http
+    path: /metrics
+    targetPort: 5001
+  namespaceSelector:
+    matchNames:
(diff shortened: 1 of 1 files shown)


russellb:
The ServiceMonitor is an OpenShift integration specific thing, so I don't think it should go in this repository.  Instead, I'd expect it as something configured by the machine-api-operator either as part of https://github.com/openshift/machine-api-operator/pull/302 or as a follow-up PR

deploy/baremetal-operator_servicemonitor.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: baremetal-operator
```
dhellmann: We should make sure this name is consistent with the name for the component we're using during installation. Talk to @sadasu about what name we settled on.

sadasu: Currently, we are using the name "metal3-baremetalhost-controller". This code is under development and this name could change during review.

Issue #230, opened 2019-06-27
Allow hardware interfaces to be set by vendor driver


tiendc:
Currently boot interface is hard-code of 'ipxe' when create an Ironic node. However vendor driver iRMC does not support ipxe. It's nice if baremetal-operator allows vendor driver to set this interface.

zaneb:
Closed by #231.

PR #231, opened 2019-06-27
Allows boot interface to be customizable

tiendc:
Currently boot interface is hard-code of 'ipxe'. This patch allows
boot interface to be set by vendor driver.

#230 


dhellmann:
Nice improvement, thanks!

Issue #234, opened 2019-07-02
Support Fujitsu driver iRMC


tiendc:

dhellmann:
I think this is done with #235 . If not, please reopen the issue.

PR #235, opened 2019-07-02
Add support for Fujitsu driver iRMC

tiendc:
#234


russellb:
Where should we be documenting that this is a valid option?  How about adding it here: https://github.com/metal3-io/baremetal-operator/blob/master/docs/api.md

tiendc:
> Where should we be documenting that this is a valid option? How about adding it here: https://github.com/metal3-io/baremetal-operator/blob/master/docs/api.md

Done with the last update. Thanks.

russellb:
Thanks!

PR #236, opened 2019-07-02
remove machineRef field from host crd

dhellmann:


pkg/apis/metal3/v1alpha1/baremetalhost_types_test.go
```diff
@@ -37,22 +37,6 @@ func TestHostAvailable(t *testing.T) {
 			Expected:    false,
 			FailMessage: "host with error returned as available",
 		},
-		{
```
mhrivnak: Why remove this test case?

pkg/apis/metal3/v1alpha1/baremetalhost_types_test.go
```diff
@@ -113,19 +97,6 @@ func TestHostNeedsHardwareInspection(t *testing.T) {
 		t.Error("expected to not need hardware inspection")
 	}
 
-	hostWithMachine := BareMetalHost{
```
mhrivnak: Why remove this test case?

pkg/apis/metal3/v1alpha1/baremetalhost_types_test.go
dhellmann: A previous PR that added the consumerRef field added a redundant test in parallel. This one references the field that was removed, so the test can be removed without losing coverage.

pkg/apis/metal3/v1alpha1/baremetalhost_types_test.go
dhellmann: Ditto.

PR #237, opened 2019-07-02
log component version on startup

### LINKED ISSUE #216: add version info to operator startup logs
It would be useful to have version information available in the logs from the operator. It is possible to tell the go compiler to replace a string in a module with a different value (see https://github.com/openshift-metal3/kni-installer/blob/master/hack/build.sh#L35 for an example). Let's see if that technique works for the operator, too.

dhellmann:
Fixes #216

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -12,7 +12,7 @@ var (
 	// String is the human-friendly representation of the version.
 	String = fmt.Sprintf("metal3-io/baremetal-operator %s", Raw)
 
-	// Commit is the commit hash from which the installer was built.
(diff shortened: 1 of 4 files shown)


pkg/version/version.go
```diff
@@ -0,0 +1,18 @@
+// Package version includes the version information.
+package version
+
+import "fmt"
+
+var (
+	// Raw is the string representation of the version. This will be replaced
+	// with the calculated version at build time.
+	// set in hack/build.sh
```
russellb: set in Makefile

dhellmann: oops, fixed!

PR #239, opened 2019-07-03
Idempotent reboot request

zaneb:
Allow consumers to request a reboot of the host independently of the power state.


dhellmann:
Part of the reason I think we need to deal with this outside of the operator is the "fencing" operation may need more than a simple reboot. We may have to leave the host powered off past some timeout value to ensure Pods are all restarted elsewhere, for example. And if a single reboot doesn't fix the problem, we may want to reprovision the host and bring it up again. I don't want to build that much logic into this operator, because it involves making decisions based on data the operator doesn't necessarily have.

zaneb:
Yeah, that would all have to be handled at a higher level. The idea here is to provide a low-level interface that the higher level stuff can use.
The only part that seems like it would be a problem is if there is indeed a minimum time that the host needs to be powered down. Do we know that is a requirement?

dhellmann:
I don't know for sure that the power of time is a requirement.

Why is the existing on/off switch API insufficient for some outside thing to manage the reboot?

zaneb:
The only way to completely prevent race conditions is to implement it at the lowest level.
This way is robust against multiple reboot requests coming in at different times, and simultaneous manipulation of the power state by the machine actuator. That means that any controller that needs to reboot the server can be completely independent of the Machine API.

pkg/apis/metal3/v1alpha1/baremetalhost_types.go
```diff
@@ -385,6 +397,16 @@ func (host *BareMetalHost) ClearError() (dirty bool) {
 	return dirty
 }
 
+// RecordBoot records the current time as the LastBooted time.
+func (host *BareMetalHost) RecordBoot() {
+	now := metav1.Now()
+	host.Status.LastBooted = &now
+	if !host.Status.PendingRebootSince.IsZero() &&
```
dhellmann: If PendingRebootSince is nil what will this do? Does IsZero() handle that?

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -363,6 +372,31 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (result re
...
 
+// Record reboot requests
+func checkUpdatedRebootRequest(host *metal3v1alpha1.BareMetalHost) (dirty bool) {
+	requestTime := host.Spec.RequestRebootAt
+	if requestTime.IsZero() ||
+		requestTime.Equal(host.Status.PendingRebootSince) {
+		return
+	}
+
+	// Bring forward future-dated requests to the present
```
dhellmann: If something schedules a reboot for the future, shouldn't we wait to do it?

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -363,6 +372,31 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (result re
 	return result, nil
 }
 
+// Record reboot requests
+func checkUpdatedRebootRequest(host *metal3v1alpha1.BareMetalHost) (dirty bool) {
```
dhellmann: Should we refuse to reboot a host that hasn't been provisioned? Or do we want to allow this as an option for resolving issues with inspection?

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -706,15 +741,20 @@ func (r *ReconcileBareMetalHost) actionManageHostPower(prov provisioner.Provisio
 	// a delay.
 	result.RequeueAfter = time.Second * 60
 
-	if info.host.Status.PoweredOn == info.host.Spec.Online {
+	needsReboot := (!info.host.Status.PendingRebootSince.IsZero() &&
```
dhellmann: I'm not sure how I feel about mixing the reboot logic with the standard power management logic. It limits us to rebooting after provisioning is done, for one thing, but it also complicates the logic here, which has to figure out if we're in the middle of a reboot operator, and if so where. I wonder if the logic of an actionReboot() function would be easier to follow because we would know the context is a reboot.

pkg/apis/metal3/v1alpha1/baremetalhost_types.go
zaneb: Yes, it's handled: https://github.com/metal3-io/baremetal-operator/blob/master/vendor/k8s.io/apimachinery/pkg/apis/meta/v1/time.go#L67

pkg/controller/baremetalhost/baremetalhost_controller.go
zaneb: Maybe. It's not really an intended use of the API, so it seemed safer not to allow it.
The problem is that there's no way to cancel a request (although you can always request it to happen sooner), so if you request a reboot next year then next year the server will reboot and if you change your mind at any point in that year there's nothing you can do about it.
I'm happy to discuss though, because I don't know what the Right Thing would be.

pkg/controller/baremetalhost/baremetalhost_controller.go
zaneb: Good question.
If the host has _never_ been provisioned then LastBooted won't be set, so we won't try to reboot.
However, if the host has been provisioned and then deprovisioned I think we will try to reboot it. If so that's an inconsistency that I need to fix.

pkg/controller/baremetalhost/baremetalhost_controller.go
dhellmann: To cancel a future request wouldn't I just need to clear that timestamp field by setting it back to nil?

pkg/controller/baremetalhost/baremetalhost_controller.go
dhellmann: It feels like asking for a power cycle can be independent of having an image provisioned to the host. At the very least we need to support externally provisioned hosts, but I don't see any particular reason to restrict it at all. If we prioritize processing a reboot request so it happens before any of the other operations like provisioning or profile matching, then it might work as remediation for failure of some other operation.

pkg/controller/baremetalhost/baremetalhost_controller.go
zaneb: The code that copies the `Spec.RequestRebootAt` field to `Status.PendingRebootSince` only does so if it's an earlier but non-zero time. Otherwise I think someone just updating the CRD from a yaml file could accidentally cancel any pending reboots.

dhellmann: Isn't being able to cancel a reboot that is scheduled for the future ("9 am tomorrow") a feature?

zaneb: Indeed, it's a feature you'd certainly want if you had the feature of scheduling a future reboot - along with several other features, like the ability schedule multiple reboots at different times, and to configure whether a reboot prior to the scheduled time counts or not. The only way to implement those features (without requiring client-side merging logic) would be to create a separate BareMetalHostScheduledRebootRequest CRD that waits until the scheduled time and then triggers this interface. The more I think about it, the more that seems like the right way to implement a scheduled reboot if somebody actually requests it (and until then YAGNI).

dhellmann: Sure, that's just like the remediation controller proposal.

If we're not actually going to manage the schedule, I still think it makes sense to just wait for the requested time instead of doing this bring-forward logic. That will make it behave the way the field names imply.

pkg/apis/metal3/v1alpha1/baremetalhost_types.go
```diff
@@ -147,6 +147,10 @@ type BareMetalHostSpec struct {
 
 	// Description is a human-entered text used to help identify the host
 	Description string `json:"description"`
+
+	// RequestRebootAt requests a reboot after the given timestamp
+	// +optional
+	RequestRebootAt *metav1.Time `json:"requestRebootAt,omitempty"`
```
mhrivnak: I assume the timestamp here is so you can determine if the reboot was achieved or not. If a reboot was requested yesterday, and another one today, the updated timestamp is used to identify that a new request was made?

It seems awkward, and has the ambiguity @dhellmann identified below around future-dated requests.

Did you consider making a separate CRD to represent a unique reboot request? A small Reconcile function in the BMO, a new "controller", would watch and reconcile the new resource. Each reboot would be initiated by creating one of those resources. Its status would reflect whether the power off and on had taken place or not. In a pinch, the creation timestamp could be compared to the host's `LastBooted` timestamp to determine if the reboot had happened.

Seems like a more explicit way to represent a series of discrete actions. It would also enable history to be preserved.

Example:

```
kind: BareMetalHostPowerCycle
metadata:
  name: example-pwmdhf
  namespace: default
spec:
  hostReference:
    name: example-host
    namespace: default
status:
  powerOff: <timestamp>
  powerOn: <timestamp>
```

WDYT?

zaneb: > I assume the timestamp here is so you can determine if the reboot was achieved or not. If a reboot was requested yesterday, and another one today, the updated timestamp is used to identify that a new request was made?

I don't really care about that. The timestamp is firstly to ensure that requests from different clients are unique in a way that just setting e.g. `rebootRequest: true` wouldn't be (in that sense it could just as easily be a random UUID or something), and secondly to reduce unnecessary reboots by not initiating any more once the last boot time is later than the last requested reboot.

It is a little awkward, but that's because it's not an interface designed for users, but rather a low-level interface designed to be an implementation detail for some higher-level external CRD like the one you proposed. (Can we use webhooks to limit access to this field to specific classes of CRDs? I'd be fine with that.)

The requirements for fencing are that at some point in time the fencing code decides that a host must be fenced, and it must be able to:

1. stop everything that is running on the host; and
2. determine when everything running on the host has been stopped;
3. regardless of what else might be happening in the system

For this use case the interface is actually quite elegant - the fencing code simply has to copy over the timestamp of the fencing decision and the operator will guarantee the rest.

I'd be interested in being proven wrong, but I don't currently believe that it's possible even in theory to implement those guarantees without having an interface in the BareMetalHost itself that looks pretty similar to this one. All of the other proposed solutions appear to sacrifice correctness in favour of a more general-purpose user-facing interface for sysadmins (something which AFAIK is not actually a requirement, and which in any event could be implemented on top of this as a CRD - and as you point out that is the correct way to implement it anyway). IMHO that isn't a good tradeoff.

pkg/controller/baremetalhost/baremetalhost_controller.go
zaneb: Maybe the name is the problem? We can't control what users put in there, but my intention was to telegraph the fact that they should be putting the current time (in the case of fencing it would be the time of the fencing decision, for example). Maybe `rebootRequestTimestamp` would be better? I picture it working similarly to the `deletedTimestamp`, though of course you generally don't interact with that one directly thanks to some syntactic sugar in the client.

pkg/apis/metal3/v1alpha1/baremetalhost_types.go
mhrivnak: I would not be too concerned about drawing a line between high-level vs. low-level reboot APIs, or anything "user-facing" vs not. We just need a declarative way to reboot a host through the k8s API. I'm not seeing value in making multiple levels of APIs to do that.

If we want to limit who or what can initiate reboot, having a dedicated CRD and using standard RBAC is the most straight-forward way to accomplish that. A webhook could potentially be used to restrict access to a specific field based on custom rules, but that's a much less obvious behavior. There's no way to programatically express to a client that its update will be rejected; it has to try the update and then deal with it being potentially rejected, and it won't know why. With RBAC, it's all clear up-front whether a client has the ability to update a resource, and it's very easy to give that permission.

Am I understanding correctly that whatever *implements* reboot will just do these things?
* set `Spec.Online = false`
* wait for `Status.PoweredOn == false`
* set `Spec.Online = true`

In the separate CRD case, the fencing code would just create an instance of that CRD instead of writing a timestamp to the BMH. Seems simple either way. Is there some aspect of correctness that would be sacrificed?

zaneb: > Am I understanding correctly that whatever implements reboot will just do these things?

Conceptually yes, but with the key constraint that `Spec.Online` can't actually change, because if it does then multiple reboot requests can interfere with each other and with unrelated power management updates from the user.

Say for example that the user decides they want to power down the host, and the request hits during a reboot. The reboot will turn the power back on at the end even though the user wanted the server to remain powered off, which is not correct. AFAIK this is not solvable by an external CRD without adding some other field in the BareMetalHost spec that is orthogonal to the `Online` field.

mhrivnak: Any time you have multiple actors that modify the spec of something, which is fairly common, those things can disagree about the desired state, and one of them will lose. It gets especially exciting if they're both controllers and they end up in a reconcile fight. We at least try to avoid that. :)

The best we can do is have a clear declarative API. If one actor requests a reboot at approximately the same time as another actor requests power off, we can just hope that whoever loses will keep watching and try again if appropriate.

zaneb: Right, but twiddling the Online field to effect a reboot is not a declarative API; it's the epitome of a workflow hacked in to a supposedly-declarative system (something I regrettably have a lot of experience with). Hence this patch, which adds a truly declarative API.

kanwar-saad:

beekhof:
Based on recent events and the direction metal3-io/metal3-docs#48 is taking, I think we should resurrect most of this patch.

Having a CRD introduces a number of problems, including lifetime, de-duplication, the absence of an operator that can install it, and the opportunity for race conditions when state is tracked separately from the BMH object.

There were some concerns around ``requestRebootAt`` values that represent a scheduled reboot, but fundamentally we cannot trust any timestamp from the user since even a well-meaning admin may have generated it on a machine that is not synchronised with the cluster or uses a different timezone.  So I think Zane's handling is correct and that scheduled reboots should not be a supported feature of the API.

PR #241, opened 2019-07-04
 Introduce gosec for security checks

JAORMX:
gosec does static code analysis and checks for common security issues in golang codebases.

This PR introduces the tool, and sets it up as a test target in the Makefile; it also ignores a couple of warnings that were found which are probably not very problematic.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,6 +54,11 @@ lint:
 	golint -set_exit_status pkg/... cmd/...
 	go vet ./pkg/... ./cmd/...
 
+.PHONY: test-sec
+test-sec:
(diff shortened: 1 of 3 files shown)


dhellmann:
> This is great, thanks Ozz! I'd personally be in favour of going further and making this a part of the 'lint' target so that it gets run by Travis on every PR.

Yes, that seems like a good idea. @JAORMX do you have time to do that?

JAORMX:
I'll have time... In two weeks. I'm on PTO right now and didn't bring my
laptop :D. I'll get back to this though.

PR #247, opened 2019-07-08
Configure metrics endpoint.

### LINKED ISSUE #245: add prometheus metrics for the operator
The operator should expose a /metrics endpoint to allow for monitoring via prometheus. It should include the standard status about a go process as well as any stats that may be unique to itself (such as the number of provisioning operations performed, duration, etc.).

russellb:
This patch configures a prometheus metrics endpoint for the
baremetal-operator.  The metrics endpoint only listens on localhost by
default, but this can be configured through a command line argument.

This can be tested with `make run` and then run the following command:
  $ curl http://localhost:8080/metrics

Closes #245

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -31,6 +31,7 @@ func printVersion() {
 
 func main() {
 	devLogging := flag.Bool("dev", false, "enable dev logging")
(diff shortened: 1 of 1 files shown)


russellb:
Thanks, Doug.  I'll open a follow up issue.

PR #249, opened 2019-07-08
use data table to drive tests

dhellmann:


kanwar-saad:

pkg/controller/baremetalhost/baremetalhost_controller_test.go
```diff
@@ -341,77 +341,105 @@ func TestDiscoveredHost(t *testing.T) {
...
-		&metal3v1alpha1.BareMetalHostSpec{
-			BMC: metal3v1alpha1.BMCDetails{
-				Address:         "ipmi://192.168.122.1:6233",
-				CredentialsName: "bmc-creds-no-user",
+	type HostFactory func() *metal3v1alpha1.BareMetalHost
+
+	testCases := []struct {
+		Scenario string
+		Secret   *corev1.Secret
+		Factory  HostFactory
```
mhrivnak: Most examples I've seen like this just put the definition inline instead of making a new type, but there's nothing at all wrong with making it an explicit type.

mhrivnak: That said, why does this need to be a function? Could you just make the host up-front and have a reference to it in this struct?

dhellmann: I've replaced the factory with a host object in this function.

Issue #252, opened 2019-07-09
metrics endpoint has port conflict


hardys:
Since https://github.com/metal3-io/baremetal-operator/commit/fa1d0e1d0118476038f6f3bbf1fe35b1b7e84ff2 landed I'm seeing https://github.com/metal3-io/baremetal-operator/pull/212 and https://github.com/openshift-metal3/dev-scripts/pull/635 fail starting the BMO like:

```
$ oc logs metal3-baremetal-operator-58b4c4d558-q28ll -c baremetal-operator --namespace openshift-machine-api
{"level":"error","ts":1562673586.4834816,"logger":"cmd","msg":"","error":"error listening on 127.0.0.1:8080: listen tcp 127.0.0.1:8080: bind: address already in use","stacktrace":"github.com/metal3-io/baremetal-operator/vendor/github.com/go-logr/zapr.(*zapLogger).Error\n\t/go/src/github.com/metal3-io/baremetal-operator/vendor/github.com/go-logr/zapr/zapr.go:128\nmain.main\n\t/go/src/github.com/metal3-io/baremetal-operator/cmd/manager/main.go:73\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:198"}
```

So we need to either configure a different port in that environment, or choose a new default here I think?

PR #253, opened 2019-07-09
Avoid port conflict for metrics endpoint

hardys:
In the OpenShift integrated case this conflicts with coredns which binds
to :::8080, switching to 8085 should avoid this conflict with the
default.

Fixes: #252

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -31,7 +31,7 @@ func printVersion() {
 
 func main() {
 	devLogging := flag.Bool("dev", false, "enable dev logging")
-	metricsAddr := flag.String("metrics-addr", "127.0.0.1:8080", "The address the metric endpoint binds to.")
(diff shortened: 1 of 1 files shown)


hardys:
Ah thanks @dhellmann I wasn't sure how the ports were assigned so I just chose one which was free in my test environment ;)

Seems we need to add the ironic exporter to https://github.com/prometheus/prometheus/wiki/Default-port-allocations then update this again

russellb:
> Ah thanks @dhellmann I wasn't sure how the ports were assigned so I just chose one which was free in my test environment ;)
> 
> Seems we need to add the ironic exporter to https://github.com/prometheus/prometheus/wiki/Default-port-allocations then update this again

Hmm, that page seems to be aimed at exporters.  I've added the Ironic Prometheus Exporter, though.

Issue #238, opened 2019-07-03
deprovisioning a host marks it as "externally provisioned"


dhellmann:
As reported in https://github.com/metal3-io/metal3-dev-env/issues/26 when a host is deprovisioned the operator treats it as externally provisioned and does not fully deprovision it. As described in https://github.com/metal3-io/cluster-api-provider-baremetal/issues/95 we need to detect the case of a host that has previously been provisioned and is being deprovisioned and not treat it as externally provisioned.

dhellmann:
This was resolved by #255

PR #255, opened 2019-07-09
add explicit flag for externally provisioned hosts

dhellmann:
We want to add an externallyProvisioned flag to the host specification to make it easier to keep track of hosts that have already had an image written to them, so we don't overwrite it.

Addresses #238


docs/api.md
```diff
@@ -30,6 +30,12 @@ power state on the physical host.
 host. For example, a Machine when the host is being used by the
 machine-api.
 
+*externallyProvisioned* -- A boolean indicating whether something else
+is managing the image running on the host. When set, if no image is
+provided, the host's power status and hardware inventory will be
+monitored. If the *image* contains image parameters, this field is
+ignored.
```
russellb: why not just always ignore the image field if externallyProvisioned is true?

dhellmann: The user can change the externallyProvisioned field at any time. So, they could start to provision a host in the normal way and then while it's in the middle of provisioning set externallyProvisioned=true. At that point, the logic managing the provisioning workflow would be circumvented and the host status would never update properly.

We _could_ allow that, but it's easier to reason about the host if we say that giving an image and saying a host is externally provisioned are mutually exclusive, but the image has precedence if both are provided. We end up with fewer overall transitions in the state diagram that way.

russellb: OK, that makes sense, thanks.

dhellmann: I added some comments in the code to explain the reasoning for this choice.

dhellmann:
I tested this locally with the current CAPB by scaling down the worker MachineSet to delete the machine and it worked. Then I scaled the MachineSet back up, provisioned a worker, and then deleted the Machine directly. The worker was deprovisioned and then reprovisioned by the new Machine created by the MachineSet.

Issue #220, opened 2019-06-14
add firmware details to host CRD


dhellmann:
The firmware details from inspection are [available in gophercloud](https://github.com/gophercloud/gophercloud/blob/master/openstack/baremetalintrospection/v1/introspection/results.go#L259), but not exposed in the host CRD.

On a libvirt/kvm VM I get:

```
    "firmware": {
      "bios": {
        "date": "04/01/2014",
        "version": "1.11.0-2.el7",
        "vendor": "SeaBIOS"
      }
    },
```

It would be useful to be able to display that information about a host.

PR #259, opened 2019-07-10
Persist firmware and bios data during introspection.

alanmeadows:
Introduce handling of the gophercloud's Firmware ExtraHardwareDataSection.  

Today, this only includes known BIOS fields.  If we are able to determine all likely data potentially within the Firmware section we can expand the data and fields handled.

Resolves: #220

### DIFF
--- a/pkg/apis/metal3/v1alpha1/baremetalhost_types.go
+++ b/pkg/apis/metal3/v1alpha1/baremetalhost_types.go
@@ -262,10 +262,29 @@ type NIC struct {
 	PXE bool `json:"pxe"`
 }
 
+// Firmware describes the firmware on the host.
+type Firmware struct {
+	// The BIOS for this firmware
+	BIOS BIOS `json:"bios"`
+}
+
+// BIOS describes the BIOS version on the host.
+type BIOS struct {
+	// The release/build date for this BIOS
+	Date string `json:"date"`
+
+	// The vendor name for this BIOS
+	Vendor string `json:"vendor"`
+
+	// The version of the BIOS
+	Version string `json:"version"`
+}
+
 // HardwareDetails collects all of the information about hardware
 // discovered on the host.
 type HardwareDetails struct {
 	SystemVendor HardwareSystemVendor `json:"systemVendor"`
+	Firmware     Firmware             `json:"firmware"`
 	RAMMebibytes int                  `json:"ramMebibytes"`
 	NIC          []NIC                `json:"nics"`
 	Storage      []Storage            `json:"storage"`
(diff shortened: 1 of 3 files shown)


pkg/apis/metal3/v1alpha1/baremetalhost_types.go
```diff
@@ -262,10 +262,29 @@ type NIC struct {
...
+// Firmware describes the firmware on the host.
+type Firmware struct {
+	// The BIOS for this firmware
+	BIOS BIOS `json:"bios"`
+}
+
+// BIOS describes the BIOS version on the host.
+type BIOS struct {
+	// The release/build date for this BIOS
+	Date string `json:"date"`
```
dhellmann: Should this field be a metav1.Time instead? I don't know if we have any idea if the input format is standard enough to be able to parse it, so maybe that's why you're using a string?

alanmeadows: I would love to use a proper `Time` field but I believe these versions, supplied directly from the vendor during introspection will be in a variety of formats.  If there is a belief we should come up with a good way to generically try and process them I wouldn't argue.  We would have to have some empty fallback for this process because I feel quite certain we will run into a date that we cannot parse at some point.

dhellmann: That's more or less what I thought. If we start seeing a few of them and it looks like maybe we could add a best-effort parser later, we can add a separate field to hold the parsed value.

dhellmann:
I'm running some tests on this now with the intent to merge it this afternoon.

dhellmann:
I'm fighting with my dev environment, but still intend to get to this PR this week.

dhellmann:
Thanks, Alan!

pkg/provisioner/ironic/ironic.go
```diff
@@ -465,8 +465,38 @@ func getCPUDetails(cpudata *introspection.CPUType) metal3v1alpha1.CPU {
...
+	// handle bios optionally
+	var bios metal3v1alpha1.BIOS
+
+	if _, ok := firmwaredata["bios"]; ok {
+
+		// we do not know if all fields will be supplied
+		// as this is not a structured response
+		// so we must handle each field conditionally
+		if _, ok := firmwaredata["bios"]["vendor"]; ok {
+			bios.Vendor = firmwaredata["bios"]["vendor"].(string)
```
zaneb: This will panic if the data contains something other than a string. I don't think that can happen in practice today, but it'd be safer to handle that case. I'll submit a follow-up PR.

PR #262, opened 2019-07-16
Minor fixups to firmware introspection data

zaneb:
Check in the updated deepcopy code, and avoid panicking if the data is not in string format.


pkg/provisioner/ironic/ironic.go
```diff
@@ -470,22 +470,13 @@ func getFirmwareDetails(firmwaredata introspection.ExtraHardwareDataSection) met
 	// handle bios optionally
 	var bios metal3v1alpha1.BIOS
 
-	if _, ok := firmwaredata["bios"]; ok {
-
+	if biosdata, ok := firmwaredata["bios"]; ok {
 		// we do not know if all fields will be supplied
 		// as this is not a structured response
 		// so we must handle each field conditionally
```
russellb: Is this comment still accurate?

pkg/provisioner/ironic/ironic.go
```diff
@@ -470,22 +470,13 @@ func getFirmwareDetails(firmwaredata introspection.ExtraHardwareDataSection) met
...
-		if _, ok := firmwaredata["bios"]["version"]; ok {
-			bios.Version = firmwaredata["bios"]["version"].(string)
-		}
-
-		if _, ok := firmwaredata["bios"]["date"]; ok {
-			bios.Date = firmwaredata["bios"]["date"].(string)
-		}
+		bios.Vendor, _ = biosdata["vendor"].(string)
+		bios.Version, _ = biosdata["version"].(string)
+		bios.Date, _ = biosdata["date"].(string)
```
russellb: what did you change here that makes this better handle unexpected types?

pkg/provisioner/ironic/ironic.go
zaneb: More or less, yes.

pkg/provisioner/ironic/ironic.go
zaneb: Doing:

    x, _ = y.(string)

will set `x` to the zero value (an empty string) if `y` is not a string, but:

    x = y.(string)

will panic if `y` is not a string.

There's also a simplification, because doing `biosdata["vendor"]` will return `nil` (which is not a string) if the `"vendor"` key does not exist. So one check suffices for testing both whether the key exists and whether the value is a string.

russellb: Thanks!

PR #263, opened 2019-07-18
Adopt externally-provisioned hosts in Ironic

### LINKED ISSUE #185: adopt externally provisioned hosts
Set the state of externally provisioned hosts (masters) so they appear as adopted systems.

zaneb:
Fixes #185

### DIFF
--- a/pkg/provisioner/demo/demo.go
+++ b/pkg/provisioner/demo/demo.go
@@ -182,6 +182,13 @@ func (p *demoProvisioner) UpdateHardwareState() (result provisioner.Result, err 
 	return result, nil
 }
 
+// Adopt allows an externally-provisioned server to be adopted.
+func (p *demoProvisioner) Adopt() (result provisioner.Result, err error) {
+	p.log.Info("adopting host")
+	result.Dirty = false
+	return
+}
+
 // Provision writes the image from the host spec to the host. It may
 // be called multiple times, and should return true for its dirty flag
 // until the deprovisioning operation is completed.
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -834,6 +834,43 @@ func (p *ironicProvisioner) startProvisioning(ironicNode *nodes.Node, checksum s
 	return p.changeNodeProvisionState(ironicNode, opts)
 }
 
+// Adopt allows an externally-provisioned server to be adopted by Ironic.
+func (p *ironicProvisioner) Adopt() (result provisioner.Result, err error) {
+	var ironicNode *nodes.Node
+
+	if ironicNode, err = p.findExistingHost(); err != nil {
+		err = errors.Wrap(err, "could not find host to adpot")
+		return
+	}
+	if ironicNode == nil {
+		err = fmt.Errorf("no ironic node for host")
+		return
+	}
+
+	switch nodes.ProvisionState(ironicNode.ProvisionState) {
+	case nodes.Enroll, nodes.Verifying:
+		err = fmt.Errorf("Invalid state for adopt: %s",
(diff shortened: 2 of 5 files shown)


pkg/provisioner/ironic/ironic.go
```diff
@@ -843,6 +843,43 @@ func (p *ironicProvisioner) startProvisioning(ironicNode *nodes.Node, checksum s
...
+			},
+		)
+	case nodes.Adopting:
+		result.RequeueAfter = provisionRequeueDelay
+		result.Dirty = true
+	case nodes.AdoptFail:
+		result.ErrorMessage = fmt.Sprintf("Host adoption failed: %s",
+			ironicNode.LastError)
+		result.Dirty = true
+	case nodes.Available:
```
dtantsur: This is also an invalid state? The target state for adoption is `active`.

kanwar-saad:

pkg/provisioner/demo/demo.go
```diff
@@ -182,6 +185,27 @@ func (p *demoProvisioner) UpdateHardwareState() (result provisioner.Result, err
 	return result, nil
 }
 
+// Adopt allows an externally-provisioned server to be adopted.
+func (p *demoProvisioner) Adopt() (result provisioner.Result, err error) {
+	hostName := p.host.ObjectMeta.Name
+
+	switch hostName {
+	default:
+		p.host.ObjectMeta.Name = AdoptingHost
```
dhellmann: Why change the object name here?

pkg/provisioner/demo/demo.go
```diff
@@ -182,6 +185,27 @@ func (p *demoProvisioner) UpdateHardwareState() (result provisioner.Result, err
...
+		result.Dirty = true
+		result.RequeueAfter = time.Second * 5
+	case AdoptingHost:
+		p.log.Info("adopting host")
+		p.host.ObjectMeta.Name = ReadyHost
```
dhellmann: Same question?

pkg/provisioner/demo/demo.go
zaneb: To demonstrate the multi-stage nature of the adoption (i.e. it's not complete after the first call to Adopt()).

dhellmann: The point of the demo driver is to ensure that we can create a bunch of example host CRs that will get "stuck" in certain states, so that someone (esp. the UI team) could use the API to see what they look like in those cases. So, rather than changing the host name, I think we want to just return the same stable state over and over for the hosts based on their name (even if that causes reconciliation again).

The fixture driver is the place to deal with the fact that we expect adoption to require multiple iterations, so that the unit tests will take the host through the expected flow.

dhellmann: It would also be good to update deploy/crds/demo-hosts.yaml with the new cases.

zaneb: Got rid of this in the new patches. Do we still need to do anything in demo-hosts.yaml?

dhellmann: Yeah, ideally we would add 1 new host names to the list of special names. "adopting-host" would always appear to be in the process of being adopted and would stay dirty (so the demo driver would always return dirty when the name matches). We can do that in a separate PR, I think.

zaneb: Oh, gotcha. I'm not sure that's interesting from a UI perspective though, because the visible state of the Host doesn't actually change - it just sits in ExternallyProvisioned either way and this is more of an implementation detail, sorting out the bookkeeping in Ironic.
Anyway, I agree we can deal with that in a separate PR if we need to.

Issue #265, opened 2019-07-19
Node properties are not supported


redixin:
It is not possible to provide `properties` when creating a node.

See https://developer.openstack.org/api-ref/baremetal/?expanded=create-node-detail#create-node

And https://github.com/gophercloud/gophercloud/blob/v0.2.0/openstack/baremetal/v1/nodes/requests.go#L220

russellb:
Can you expand on your use case that needs feeding data into this API?

redixin:
@russellb e.g. if uefi is the only option `--property capabilities='boot_mode:uefi'`

russellb:
OK - a design goal is to keep Ironic hidden as an implementation detail.  Is support for UEFI only hosts the issue?  Should we rename the issue something like: `Need support for UEFI only hosts` ?

dtantsur:
Aren't we going to switch to UEFI-only anyway?

alexz-kh:
>Aren't we going to switch to UEFI-only anyway?
isn't it will be bad idea? not all deployment require uefi and not even all hw support it properly(at least "old" hw). i guess, it might drop some compatibility layer..

dhellmann:
Will this be resolved if we finish the work described for #207?

zaneb:
I think this is covered by #207.

Issue #206, opened 2019-05-31
Support RAID setup for baremetal server


furukawa3:
We'd like to deploy baremetal server with RAID configuration.  In order to
setup/unset RAID, using vendor driver is necessary.  This issue proposes a new yaml
attribute to setup RAID with Fujitsu PRIMERGY server by using iRMC driver.

dhellmann:
I wonder if it makes sense to fold RAID setup into the hardware profile? Or would we potentially want hosts that match the same profile to have different RAID configurations?

furukawa3:
Hi.  I think we should have 1 RAID configuration with 1 bare metal server.  So, I agree with your 1st proposal.  Here is a my draft idea of hardware profile.  Just adding "targetRAID".   

```yaml
apiVersion: metal3.io/v1alpha1
kind: BareMetalHost
metadata:
  name: bm1
spec:
  online: true
  bmc:
    address: libvirt://192.168.122.1:6230/
    credentialsName: bm1-bmc-secret
  bootMACAddress: 52:54:00:c3:40:2e
  targetRAID: 5
```

dhellmann:
@dtantsur and @juliakreger do you have any thoughts on this?

dtantsur:
Should this tie into the deploy templates work done for Train? We could just use a deploy template name here, that would cover RAID, BIOS and any other customization from deploy templates.

```yaml
apiVersion: metal3.io/v1alpha1
kind: BareMetalHost
metadata:
  name: bm1
spec:
  online: true
  bmc:
    address: libvirt://192.168.122.1:6230/
    credentialsName: bm1-bmc-secret
  bootMACAddress: 52:54:00:c3:40:2e
  deployTemplates:
     - CUSTOM_RAID5
     - CUSTOM_BIOS_VMX
```

Of course, we'd need something to create deploy templates themselves from their JSON definition.

Reference: https://docs.openstack.org/ironic/latest/admin/node-deployment.html#deploy-templates

juliakreger:
@dhellmann @furukawa3 I guess a target + a hardware profile would work, the potential issue that comes to mind is that raid configuration traditionally takes some details to achieve. See https://docs.openstack.org/ironic/latest/admin/drivers/irmc.html#raid-configuration-support

furukawa3:
I'm so sorry I was late reply!!

@dtantsur @juliakreger Long time no see since Denver PTG :)

> Should this tie into the deploy templates work done for Train? 

[...]

PR #266, opened 2019-07-19
Make use of nodes.CreateOpts.Properties (#265)

redixin:
This may be useful when you need to pass `properties` when creating a node in ironic, e.g. `boot_mode:uefi`

### DIFF
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -2,6 +2,12 @@ package ironic
 
 import (
 	"fmt"
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/baremetal/noauth"
+	"github.com/gophercloud/gophercloud/openstack/baremetal/v1/nodes"
+	"github.com/gophercloud/gophercloud/openstack/baremetal/v1/ports"
+	noauthintrospection "github.com/gophercloud/gophercloud/openstack/baremetalintrospection/noauth"
+	"github.com/gophercloud/gophercloud/openstack/baremetalintrospection/v1/introspection"
 	"io/ioutil"
 	"net/http"
 	"net/url"
@@ -9,13 +15,6 @@ import (
 	"sort"
 	"strings"
 	"time"
-
-	"github.com/gophercloud/gophercloud"
-	"github.com/gophercloud/gophercloud/openstack/baremetal/noauth"
-	"github.com/gophercloud/gophercloud/openstack/baremetal/v1/nodes"
-	"github.com/gophercloud/gophercloud/openstack/baremetal/v1/ports"
-	noauthintrospection "github.com/gophercloud/gophercloud/openstack/baremetalintrospection/noauth"
-	"github.com/gophercloud/gophercloud/openstack/baremetalintrospection/v1/introspection"
 
 	nodeutils "github.com/gophercloud/utils/openstack/baremetal/v1/nodes"
 
@@ -214,6 +213,12 @@ func (p *ironicProvisioner) ValidateManagementAccess() (result provisioner.Resul
 		return result, errors.Wrap(err, "failed to find existing host")
 	}
 
+	properties := make(map[string]interface{})
+
+	if p.host.Spec.BootUEFI {
+		properties["capabilities"] = "boot_mode:uefi"
+	}
+
 	// If we have not found a node yet, we need to create one
 	if ironicNode == nil {
 		p.log.Info("registering host in ironic")
@@ -235,6 +240,7 @@ func (p *ironicProvisioner) ValidateManagementAccess() (result provisioner.Resul
 				BootInterface: p.bmcAccess.BootInterface(),
 				Name:          p.host.Name,
 				DriverInfo:    driverInfo,
+				Properties:    properties,
 			}).Extract()
 		// FIXME(dhellmann): Handle 409 and 503? errors here.
 		if err != nil {
--- a/pkg/apis/metal3/v1alpha1/baremetalhost_types.go
+++ b/pkg/apis/metal3/v1alpha1/baremetalhost_types.go
@@ -142,6 +142,9 @@ type BareMetalHostSpec struct {
 	// data to be passed to the host before it boots.
 	UserData *corev1.SecretReference `json:"userData,omitempty"`
 
+	// Whether or not use UEFI boot mode (efiboot ipmitool option)
+	BootUEFI bool `json:"bootUEFI,omitempty"`
+
 	// Description is a human-entered text used to help identify the host
 	Description string `json:"description"`
 


redixin:
This is not tested yet.

pkg/provisioner/ironic/ironic.go
```diff
@@ -2,6 +2,12 @@ package ironic
 
 import (
 	"fmt"
+	"github.com/gophercloud/gophercloud"
```
redixin: These lines has been moved by Goland automaticly :-\

russellb:
Thanks for the patch!

I mentioned this over in the issue you filed (#265), but a design goal is to hide Ironic as an implementation detail.  Do we need a way to note on a `BareMetalHost` object that the host can only support UEFI?  I imagine that's something we can't discover out of band before trying to boot an image?

russellb:
@juliakreger @dtantsur FYI - some Ironic advice here would be great

alexz-kh:
@russellb 
>Do we need a way to note on a BareMetalHost object that the host can only support UEFI?
why so ? that's actually just additional property for ironic-conductor, to make correct ipmi call for hardware.
I don't think we need to lock resource somehow - some supermicro hosts(as exampl), can provide DUAL uefi\bios processing in same time, and try to boot in series- so, user may chose one or another solution just changing boot opts via ipmtiool.

zaneb:
I agree with Russell and Dmitry - a BootUEFI option makes sense; exposing an unstable Ironic-specific API in the BaremetalHost interface does not.

alexz-kh:
The issue here, that those option applicable only for ipmi proto case only.
So, we need to add some binding for  ipmi driver, which extend obkect to even more complicated :\

dtantsur:
Which option do you mean? `boot_mode` exists for all drivers, although they may handle it differently

alexz-kh:
Hmm, just double-grep-ed - looks so. ok, so it make sense.

alexz-kh:
New solution also, at least with super-micro X10SRW-F in native UEFI  mode.

russellb:
I wonder if this should actually be part of the `hardwareProfile` interface?

redixin:
@dtantsur what do you think about `BaremetalHostSpec.bootUEFI`?

dhellmann:
Is there any way at all for us to detect this setting through the BMC so that the user does not have to give it to us?

Failing that, I would like to wait to add this field until we can resolve some of the questions about how we are going to define hardware profiles and make them user-configurable. As @russellb points out, this seems like an excellent example of a setting that should go on the profile, rather than on each individual host.

dtantsur:
> Is there any way at all for us to detect this setting through the BMC so that the user does not have to give it to us?

We do it though inspection. However, this setting may also provide a *desired* state for hardware that supports changing between UEFI and Legacy out-of-band (e.g. redfish).

alexz-kh:
>We do it though inspection

This stage, is to late. since in case clear UEFI boot, you need to perform _boot for inspection_ already via uefi-boot-fix. (before inspection will be finished, and ironic|metal3 can realize hardware type)

dhellmann:
> > Is there any way at all for us to detect this setting through the BMC so that the user does not have to give it to us?
> 
> We do it though inspection. However, this setting may also provide a _desired_ state for hardware that supports changing between UEFI and Legacy out-of-band (e.g. redfish).

Are there any other boot settings that the user may need to give us like this, that would make us want to create a BootOptions struct inside the Spec?

Are there modes other than BIOS and UEFI, which would indicate that this parameter should use a type other than a boolean?

zaneb:
> Are there any other boot settings that the user may need to give us like this, that would make us want to create a BootOptions struct inside the Spec?

Secure boot is one.

> Are there modes other than BIOS and UEFI, which would indicate that this parameter should use a type other than a boolean?

Not on x86-64. Other architectures have different bootloaders (e.g. Open Firmware). But AFAIK x86-64 is the only one where you might have to select between BIOS vs. UEFI.

juliakreger:
> > Are there modes other than BIOS and UEFI, which would indicate that this parameter should use a type other than a boolean?
> 
> Not on x86-64. Other architectures have different bootloaders (e.g. Open Firmware). But AFAIK x86-64 is the only one where you might have to select between BIOS vs. UEFI.

Power has no concept between the two.

Arm is purely UEFI at this point, so it is implied. We have no special code for such, but there is also no third party ARM CI at present for upstream Ironic. We only know it works from occasional contributor check-ins with specialized hardware.

dhellmann:
> > > Are there modes other than BIOS and UEFI, which would indicate that this parameter should use a type other than a boolean?
> > 
> > 
> > Not on x86-64. Other architectures have different bootloaders (e.g. Open Firmware). But AFAIK x86-64 is the only one where you might have to select between BIOS vs. UEFI.
> 
> Power has no concept between the two.

Does that mean it supports something else, that it only supports one of BIOS or UEFI, or that it somehow supports both transparently?

> Arm is purely UEFI at this point, so it is implied. We have no special code for such, but there is also no third party ARM CI at present for upstream Ironic. We only know it works from occasional contributor check-ins with specialized hardware.

OK, we know we're going to need to support both Power and ARM downstream for metal3 eventually, so I want to design its API with those architectures in mind from the start.

What "owns" requiring a specific boot mode? Is it the image? Is there any reason to think that someone would want to take 2 identical x86_64 servers and boot them using different protocols? It seems unlikely in a given cluster, but remember we're not limiting the operator to deploying images as part of a cluster.

redixin:
> Are there modes other than BIOS and UEFI, which would indicate that this parameter should use a type other than a boolean?

This particular option is to decide whether we add `efiboot` option to `ipmitool` or not.

juliakreger:
> > > > Are there modes other than BIOS and UEFI, which would indicate that this parameter should use a type other than a boolean?
> > > 
> > > 
> > > Not on x86-64. Other architectures have different bootloaders (e.g. Open Firmware). But AFAIK x86-64 is the only one where you might have to select between BIOS vs. UEFI.
> > 
> > 
> > Power has no concept between the two.
> 
> Does that mean it supports something else, that it only supports one of BIOS or UEFI, or that it somehow supports both transparently?

Their boot loader is intelligent and is flashed in. It loads kernel/ramdisk and looks for an explicit partition format layout that is performed with the architecture type. I would say it is something different, but so far outside our context that it would make our heads hurt to do anything special to account for it beyond setting the CPU architecture for PPC. If you want details on this, I think mjturek on freenode irc is a good person to ask.

> > Arm is purely UEFI at this point, so it is implied. We have no special code for such, but there is also no third party ARM CI at present for upstream Ironic. We only know it works from occasional contributor check-ins with specialized hardware.
> 
> OK, we know we're going to need to support both Power and ARM downstream for metal3 eventually, so I want to design its API with those architectures in mind from the start.
> 
> What "owns" requiring a specific boot mode? Is it the image? Is there any reason to think that someone would want to take 2 identical x86_64 servers and boot them using different protocols? It seems unlikely in a given cluster, but remember we're not limiting the operator to deploying images as part of a cluster.

I guess the owner would be a union of the admin setting the CPU architecture, user specifying what they desire (x86 only), and ultimately the image. In the metal3 case, I think boot order will be fairly consistent across the fleet of hardware.

juliakreger:
> > Are there modes other than BIOS and UEFI, which would indicate that this parameter should use a type other than a boolean?
> 
> This particular option is to decide whether we add `efiboot` option to `ipmitool` or not.

Slightly more complex than this and there are some caveats with IPMI. Anyway, I think we want to move mostly to redfish as time moves forward.

dhellmann:
> > Are there modes other than BIOS and UEFI, which would indicate that this parameter should use a type other than a boolean?
> 
> This particular option is to decide whether we add `efiboot` option to `ipmitool` or not.

I'm trying to get us to look beyond the immediate need to set 1 option so that we don't end up with an API that has a bunch of different flags like this, all of them mutually exclusive or where mixing them in certain ways doesn't make sense. In this case, if we want to set a "boot mode" I think perhaps we want to ask the user to provide a string that we validate using an enum that is defined as part of the host object API, instead of just using a boolean. But I'm trying to ask questions to get you to help me understand what those modes may be, what dictates which mode to use, etc. because my knowledge in this area is a little thin.

dhellmann:
> > > Power has no concept between the two.
> > 
> > Does that mean it supports something else, that it only supports one of BIOS or UEFI, or that it somehow supports both transparently?
> 
> Their boot loader is intelligent and is flashed in. It loads kernel/ramdisk and looks for an explicit partition format layout that is performed with the architecture type. I would say it is something different, but so far outside our context that it would make our heads hurt to do anything special to account for it beyond setting the CPU architecture for PPC. If you want details on this, I think mjturek on freenode irc is a good person to ask.

OK, it sounds like it is probably sufficient to treat Power as not needing to be told what boot mode to use and say whatever field we add to set it for x86 is ignored by Power systems.

> > What "owns" requiring a specific boot mode? Is it the image? Is there any reason to think that someone would want to take 2 identical x86_64 servers and boot them using different protocols? It seems unlikely in a given cluster, but remember we're not limiting the operator to deploying images as part of a cluster.
> 
> I guess the owner would be a union of the admin setting the CPU architecture, user specifying what they desire (x86 only), and ultimately the image. In the metal3 case, I think boot order will be fairly consistent across the fleet of hardware.

OK, so the set of valid modes depends on the architecture, but the actual mode necessary depends on the image.

I see 4 options:

1. Add a standalone field in the host's spec, as in this PR.
2. Add a field to the hardware profile.
3. Add a field to the the existing Image struct as part of the host's spec.
4. Create a new BootOptions struct and add a field there.

Option 1 seems like it is going to lead us to a somewhat disorganized API, as we add more and more fields over time.

Option 2 hides the field and doesn't give the user any control, but it sounds like that control is needed. I think we can rule this approach out.

Option 3 ties the new field to the image, which seems like the thing that dictates the required value.

Option 4 also gives us a place to put the secure boot option @zaneb mentioned, but we could alternatively put that in the "bios" struct described in #206.

I think I like option 3 at this point.

What do people think? Are there other approaches to consider? Do any of the options I've described make more sense than the others?

juliakreger:
> > > Are there modes other than BIOS and UEFI, which would indicate that this parameter should use a type other than a boolean?
> > 
> > 
> > This particular option is to decide whether we add `efiboot` option to `ipmitool` or not.
> 
>  In this case, if we want to set a "boot mode" I think perhaps we want to ask the user to provide a string that we validate using an enum that is defined as part of the host object API, instead of just using a boolean. 

I think this is actually the best option because the operator can be clueful enough to know "oh, PPC or ARM, the backend has that" and then the user could say "I want x86_64 bios mode" or "i want x86_64 uefi mode" using the same field.

dhellmann:
> > > > Are there modes other than BIOS and UEFI, which would indicate that this parameter should use a type other than a boolean?
> > > 
> > > 
> > > This particular option is to decide whether we add `efiboot` option to `ipmitool` or not.
> > 
> > 
> > In this case, if we want to set a "boot mode" I think perhaps we want to ask the user to provide a string that we validate using an enum that is defined as part of the host object API, instead of just using a boolean.
> 
> I think this is actually the best option because the operator can be clueful enough to know "oh, PPC or ARM, the backend has that" and then the user could say "I want x86_64 bios mode" or "i want x86_64 uefi mode" using the same field.

It sounds like you're describing a pass-through field, and I don't think we want that. It requires the user to understand implementation details of metal3 ("the value I place here is passed to Ironic as X"). We will provide a better experience if we explicitly list the supported modes in our documentation and in the OpenAPI validation for the host CRD.

juliakreger:
> I think I like option 3 at this point.
> 
> What do people think? Are there other approaches to consider? Do any of the options I've described make more sense than the others?

Option 3 as an enum that is a blend of options I think works best to fulfill the needs. It allows us to have each option as a focused area of that is supported and validated for use as opposed to creating a weird support matrix. This also makes it more pluggable as if someone didn't want to use ironic, they could still map those enums to settings that worked for their tooling.

dhellmann:
> > I think I like option 3 at this point.
> > What do people think? Are there other approaches to consider? Do any of the options I've described make more sense than the others?
> 
> Option 3 as an enum that is a blend of options I think works best to fulfill the needs. It allows us to have each option as a focused area of that is supported and validated for use as opposed to creating a weird support matrix. This also makes it more pluggable as if someone didn't want to use ironic, they could still map those enums to settings that worked for their tooling.

When you say "a blend of options", what do you mean? I was anticipating having values like "BIOS" and "UEFI", then for architectures where the mode isn't required allowing an empty string (or using "automatic" or making the field a pointer so it can be nil). Do you think we should add more information to the enum values?

juliakreger:
So in theory, there is "BIOS", "BIOS w/Trusted Boot", "UEFI", "UEFI with SecureBoot" <-- I think.

But that matrix only applies to x86_64 processors... I guess automatic could work for other things, but I'd kind of prefer to take a route of "we know this should be good" as opposed to "this is the passthrough, and good luck"

dhellmann:
> So in theory, there is "BIOS", "BIOS w/Trusted Boot", "UEFI", "UEFI with SecureBoot" <-- I think.
> 
> But that matrix only applies to x86_64 processors... I guess automatic could work for other things, but I'd kind of prefer to take a route of "we know this should be good" as opposed to "this is the passthrough, and good luck"

If a platform only supports 1 mode, or figures things out on its own, then the user shouldn't have to specify anything.

Does the image dictate whether trusted/secure boot should be used? Or can an image that expects UEFI be booted with or without SecureBoot?

juliakreger:
> > So in theory, there is "BIOS", "BIOS w/Trusted Boot", "UEFI", "UEFI with SecureBoot" <-- I think.
> > But that matrix only applies to x86_64 processors... I guess automatic could work for other things, but I'd kind of prefer to take a route of "we know this should be good" as opposed to "this is the passthrough, and good luck"
> 
> If a platform only supports 1 mode, or figures things out on its own, then the user shouldn't have to specify anything.

Only the architecture would need to be defined.

> 
> Does the image dictate whether trusted/secure boot should be used? Or can an image that expects UEFI be booted with or without SecureBoot?

Images can be dual or even triple purposed. SecureBoot is slightly different in that it is wired into the boot loader configuration and end effect depends on external configuration (if keys match, tpm set to halt the machine, or if just the bit tracking secure state flips, etc.), but generally the differences enforcement with some of those settings like secureboot/trustedboot need to be conveyed to the bmc if that is a supported feature. That being said, for IPMI nothing explicit is done beyond "boot from uefi" or "boot from bios".

iLOs are a good example where additional information allows for policy enforcement with secureboot to take place which needs that driver.

kanwar-saad:

nordixinfra:

PR #275, opened 2019-08-05
improve error handling in power state management code

dhellmann:
When the host is in certain provisioning states we cannot change its
power status. Rather than erroring and trying over and over, report
the issue and delay retrying to give the host a chance to move into a
state where power can be changed.

### DIFF
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -1176,9 +1176,15 @@ func (p *ironicProvisioner) Delete() (result provisioner.Result, err error) {
 func (p *ironicProvisioner) changePower(ironicNode *nodes.Node, target nodes.TargetPowerState) (result provisioner.Result, err error) {
 	p.log.Info("changing power state")
 
-	// If we're here, we're going to change the state and we should
-	// wait to try that again.
-	result.RequeueAfter = powerRequeueDelay
+	if ironicNode.TargetProvisionState != "" {
+		p.log.Info("host in state that does not allow power change, try again after delay",
+			"state", ironicNode.ProvisionState,
+			"target state", ironicNode.TargetProvisionState,
+		)
+		result.Dirty = true
+		result.RequeueAfter = powerRequeueDelay
+		return result, nil
+	}
 
 	changeResult := nodes.ChangePowerState(
 		p.client,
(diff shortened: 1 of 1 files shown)


pkg/provisioner/ironic/ironic.go
```diff
@@ -1176,9 +1176,14 @@ func (p *ironicProvisioner) Delete() (result provisioner.Result, err error) {
...
 
-	// If we're here, we're going to change the state and we should
-	// wait to try that again.
-	result.RequeueAfter = powerRequeueDelay
+	switch nodes.ProvisionState(ironicNode.ProvisionState) {
+	case nodes.Inspecting, nodes.InspectWait, nodes.InspectFail, nodes.Cleaning, nodes.CleanWait:
+		p.log.Info("host in state that does not allow power change, try again after delay",
+			"state", ironicNode.ProvisionState,
+		)
+		result.Dirty = true
```
mhrivnak: Should this also be setting `result.RequeueAfter`?

pkg/provisioner/ironic/ironic.go
```diff
@@ -1176,9 +1176,14 @@ func (p *ironicProvisioner) Delete() (result provisioner.Result, err error) {
 func (p *ironicProvisioner) changePower(ironicNode *nodes.Node, target nodes.TargetPowerState) (result provisioner.Result, err error) {
 	p.log.Info("changing power state")
 
-	// If we're here, we're going to change the state and we should
-	// wait to try that again.
-	result.RequeueAfter = powerRequeueDelay
+	switch nodes.ProvisionState(ironicNode.ProvisionState) {
+	case nodes.Inspecting, nodes.InspectWait, nodes.InspectFail, nodes.Cleaning, nodes.CleanWait:
```
dtantsur: There is an easier way to find it out: inspect the `node.target_provision_state` field. If it's not `null`, you probably cannot (and definitely should not) change its power state.

pkg/provisioner/ironic/ironic.go
dhellmann: Good eye, yes.

pkg/provisioner/ironic/ironic.go
dhellmann: That's much simpler, yes.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1176,9 +1176,15 @@ func (p *ironicProvisioner) Delete() (result provisioner.Result, err error) {
 func (p *ironicProvisioner) changePower(ironicNode *nodes.Node, target nodes.TargetPowerState) (result provisioner.Result, err error) {
 	p.log.Info("changing power state")
 
-	// If we're here, we're going to change the state and we should
-	// wait to try that again.
-	result.RequeueAfter = powerRequeueDelay
+	if ironicNode.TargetProvisionState != "" {
```
dtantsur: Are you sure it's an empty string? Ironic API returns JSON `null` in this case.

dhellmann: The go type defined by the client is a string, not a pointer to a string. It can't be nil.

PR #283, opened 2019-08-21
add coverage info to unit test output

Jaakko-Os:
-two Makefile targets added
   -coverage statistics written to file
   -code coverage results launched to browser

fix: #283

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,18 +18,20 @@ export IRONIC_INSPECTOR_ENDPOINT=http://localhost:5050/v1/
 .PHONY: help
 help:
 	@echo "Targets:"
-	@echo "  test         -- run unit tests and linter"
-	@echo "  unit         -- run the unit tests"
-	@echo "  unit-verbose -- run unit tests with verbose flag enabled"
-	@echo "  lint         -- run the linter"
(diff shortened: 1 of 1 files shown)


maelk:
Can you please link the issue ? Do we want the code coverage to be an independent target or should it be included in the existing targets ?

Makefile
```diff
@@ -46,8 +48,18 @@ travis: test-verbose lint
 unit:
 	go test $(GO_TEST_FLAGS) ./cmd/... ./pkg/...
 
+.PHONY: unit-cover
+unit-cover:
+	go test -coverprofile=cover.out $(GO_TEST_FLAGS) ./cmd/... ./pkg/...
+	go tool cover -func=cover.out -o coverage-stats.out
```
dhellmann: Ideally we would have a target that reports the stats to the console, so we can run it in CI and see the results in the logs. Is there a separate mode for doing that, or do we just need to cat the output file here?

Jaakko-Os: If there is no output file defined, the stats will be reported directly to he console. I'll do that.

dhellmann:
This looks good, thanks!

PR #284, opened 2019-08-23
Add redfish BMC support in baremetal operator

maelk:
accepts bmc address like
`redfish+https://my.host:8080/redfish/v1/Systems/MySystem`

### DIFF
--- a/pkg/bmc/redfish.go
+++ b/pkg/bmc/redfish.go
@@ -1,0 +1,75 @@
+package bmc
+
+import (
+	"net/url"
+	"strings"
+)
+
+func init() {
+	registerFactory("redfish", newRedfishAccessDetails)
+	registerFactory("redfish+http", newRedfishAccessDetails)
+	registerFactory("redfish+https", newRedfishAccessDetails)
+}
+
+func newRedfishAccessDetails(parsedURL *url.URL) (AccessDetails, error) {
+	return &redfishAccessDetails{
+		bmcType:  parsedURL.Scheme,
+		host:     parsedURL.Host,
+		path:     parsedURL.Path,
+	}, nil
+}
+
+type redfishAccessDetails struct {
+	bmcType  string
+	host     string
+	path     string
+}
+
+const redfishDefaultScheme = "https"
+
+func (a *redfishAccessDetails) Type() string {
+	return a.bmcType
+}
+
+// NeedsMAC returns true when the host is going to need a separate
+// port created rather than having it discovered.
+func (a *redfishAccessDetails) NeedsMAC() bool {
+	// For the inspection to work, we need a MAC address
+	// https://github.com/metal3-io/baremetal-operator/pull/284#discussion_r317579040
+	return true
+}
+
+func (a *redfishAccessDetails) Driver() string {
+	return "redfish"
+}
+
+// DriverInfo returns a data structure to pass as the DriverInfo
+// parameter when creating a node in Ironic. The structure is
+// pre-populated with the access information, and the caller is
+// expected to add any other information that might be needed (such as
+// the kernel and ramdisk locations).
+func (a *redfishAccessDetails) DriverInfo(bmcCreds Credentials) map[string]interface{} {
+	redfishAddress := []string{}
+	schemes := strings.Split(a.bmcType, "+")
+	if len(schemes) > 1 {
+		redfishAddress = append(redfishAddress, schemes[1])
+	} else {
+		redfishAddress = append(redfishAddress, redfishDefaultScheme)
+	}
+	redfishAddress = append(redfishAddress, "://")
+	redfishAddress = append(redfishAddress, a.host)
+
+	result := map[string]interface{}{
+		"redfish_system_id":     a.path,
+		"redfish_username": bmcCreds.Username,
+		"redfish_password": bmcCreds.Password,
+		"redfish_address": strings.Join(redfishAddress, ""),
+	}
+
+	return result
+}
+
+// That can be either pxe or redfish-virtual-media
+func (a *redfishAccessDetails) BootInterface() string {
+	return "pxe"
+}
--- a/docs/api.md
+++ b/docs/api.md
@@ -18,6 +18,10 @@ in place of `https://` in the iDRAC URL; only the hostname or IP address is
 required - `idrac://host.example` is equivalent to
 `idrac+https://host.example:443/wsman`. Fujitsu iRMC is also supported,
 by using the scheme `irmc://<host>:<port>` with `<port>` is optional.
+Redfish is also supported, by using the scheme `redfish://` (or
+`redfish+http://` to disable TLS) in place of `https://` in the Redfish URL;
+the hostname or IP address, and the path to the system ID are required -
+for example `redfish://myhost.example/redfish/v1/Systems/MySystemExample`
 
 *bmc.credentials* -- A reference to a Secret containing the connection
 data, at least username and password, for the BMC.
--- a/pkg/bmc/access_test.go
+++ b/pkg/bmc/access_test.go
@@ -180,6 +180,57 @@ func TestParse(t *testing.T) {
 			Hostname: "192.168.122.1",
 			Path:     "",
 		},
+
+		{
+			Scenario: "redfish url",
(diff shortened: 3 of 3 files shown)


pkg/bmc/redfish.go
```diff
@@ -0,0 +1,96 @@
...
+func (a *redfishAccessDetails) DriverInfo(bmcCreds Credentials) map[string]interface{} {
+	redfishAddress := []string{}
+	schemes := strings.Split(a.bmcType, "+")
+	if len(schemes) > 1 {
+		redfishAddress = append(redfishAddress, schemes[1])
+	} else {
+		redfishAddress = append(redfishAddress, redfishDefaultScheme)
+	}
+	redfishAddress = append(redfishAddress, "://")
+	redfishAddress = append(redfishAddress, a.hostname)
```
dtantsur: IIRC (@etingof could confirm) the `/redfish/v1` bit has to be present in both `redfishAddress` and `redfish_system_id`.

pkg/bmc/redfish.go
```diff
@@ -0,0 +1,96 @@
...
+// NeedsMAC returns true when the host is going to need a separate
+// port created rather than having it discovered.
+func (a *redfishAccessDetails) NeedsMAC() bool {
+	// libvirt-based hosts used for dev and testing require a MAC
+	// address, specified as part of the host, but we don't want the
+	// provisioner to have to know the rules about which drivers
+	// require what so we hide that detail inside this class and just
+	// let the provisioner know that "some" drivers require a MAC and
+	// it should ask.
+	return false
```
dtantsur: Must be `true`, inspection lookup won't work otherwise.

pkg/bmc/redfish.go
maelk: We used this as a base for implementation: https://docs.openstack.org/ironic/stein/admin/drivers/redfish.html There the redfish/v1 does not seem to be needed for redfish_address.

pkg/bmc/redfish.go
maelk: Could you please explain this ? From the ipmi.go, it returns `return a.bmcType == "libvirt"` meaning that when we use "ipmi://xxx" addresses such as dev env, it returns false. And what about libvirt/ non-libvirt based hosts ?

dhellmann: I think one of the goals is to replace the current IPMI-based dev environment with one that uses Redfish. That would mean using the Redfish driver when the type is "libvirt". And that would mean this function should have logic similar to the current IPMI implementation, at least.

Unless @dtantsur's comment means that all Redfish servers always need a MAC pre-defined in order to have their boot process managed?

dtantsur: To be clear, PXE booting requires MAC address to be present. We can use introspection to discover NICs (and thus MAC addresses), however, it has a chicked-and-egg problems: without both MAC addresses and IPMI address in our records, how do we know which introspection data corresponds to which node?

TL;DR Redfish requires a MAC address with and without inspection for the time being.

pkg/bmc/redfish.go
dtantsur: You're probably right, sorry for confusion.

pkg/bmc/redfish.go
dhellmann: @dtantsur why is that different for redfish from IPMI (which only requires a MAC for VMs)?

maelk: Aren't we using the IPMI driver for dev env (and not Libvirt) : https://github.com/metal3-io/metal3-dev-env/blob/master/vm-setup/roles/libvirt/templates/ironic_nodes.json.j2#L15 And in that case the needsMAC returns false. So how does the dev env works if the MAC should be required ?

maelk: Aren't we using the IPMI driver for dev env (and not Libvirt) : https://github.com/metal3-io/metal3-dev-env/blob/master/vm-setup/roles/libvirt/templates/ironic_nodes.json.j2#L15 then used here : https://github.com/metal3-io/metal3-dev-env/blob/master/lib/common.sh#L133 . And in that case the needsMAC returns false. So how does the dev env works if the MAC should be required ?

maelk: and what if we switched from PXE to redfish-virtual-media boot interface ?

dhellmann: The scripts for building the dev environment populate the MAC address, so even if the driver isn't requiring it, it's there.

dtantsur: > why is that different for redfish from IPMI (which only requires a MAC for VMs)?

Because on real hardware we can detect IPMI address from within the machine and use it for matching data to nodes. It's much more complex for Redfish.

> what if we switched from PXE to redfish-virtual-media boot interface ?

Might work, needs testing.

dtantsur: > What additional work would be needed in this repo to make it work with ironic configured that way?

To clarify on my comment: inspection is independent of the boot method in ironic, it always uses PXE and requires a port or an IPMI address for matching. Virtual media only helps if we don't use inspection.

maelk: ok thank you @dtantsur and @dhellmann . This is now fixed.

maelk:
@dtantsur @dhellmann Any further comments ?

pkg/bmc/redfish.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+func newRedfishAccessDetails(bmcType, portNum, hostname, path string) (AccessDetails, error) {
+	// If the hostname is an ipv6 address, it needs to be in square brackets
+	// as we are forming a URL out of it.
+	addresses, err := net.LookupIP(hostname)
+	if err == nil {
+		for _, address := range addresses {
+			if address.String() == hostname && address.To4() == nil {
+				hostname = strings.Join([]string{"[",hostname,"]"}, "")
```
dhellmann: What happens here if a host has multiple IPv6 addresses?

dtantsur: Why do we even resolve the address?

maelk: Resolving the hostname is to find out if the domain name given is an ipv6 (that will need brackets around) or not. If we are given a domain name and the host has multiple addresses, then none will match the domain name. If we are given an IPv6, the resolved IP will match the given IP and the brackets will be added.

dhellmann: OK, the purpose of that loop was not obvious to me at all, even with the comment.

I think if we take the other PR that changes the access API to use a parsed URL, we would not need to perform this check here, right?

pkg/bmc/redfish.go
```diff
@@ -0,0 +1,92 @@
...
+func init() {
+	registerFactory("redfish", newRedfishAccessDetails)
+	registerFactory("redfish+http", newRedfishAccessDetails)
+	registerFactory("redfish+https", newRedfishAccessDetails)
+}
+
+func newRedfishAccessDetails(bmcType, portNum, hostname, path string) (AccessDetails, error) {
+	// If the hostname is an ipv6 address, it needs to be in square brackets
+	// as we are forming a URL out of it.
+	addresses, err := net.LookupIP(hostname)
```
zaneb: It seems like you should be using `net.ParseIP()`, not `net.LookupIP()` here.

pkg/bmc/redfish.go
zaneb: Correct, with that other PR you could just use `.Host` instead of `.Hostname()` and you get the unmodified host+port.
This seems like a good reason to accept that PR.

maelk: Ok, so we'll first focus on URL PR and then modify this one

maelk:
now built on the URL PR, we got rid of the lookup part.

maelk:
running the integration CI loop would not test anything from this change as metal3-dev-env uses IPMI. A PR to use Redfish for one vm is under work

dhellmann:
> running the integration CI loop would not test anything from this change as metal3-dev-env uses IPMI. A PR to use Redfish for one vm is under work

If it's easier to just switch them all then I think a patch that added a variable to control that would be very welcome.

maelk:
Any further comments ? (@zaneb ? )

PR #285, opened 2019-08-23
Model Host with a finite state machine

zaneb:
Implement transitions of the Host state using a finite state machine. This means that the next action of the Host depends on the current state as well as any changed data.

This means that e.g. a deprovisioning operation will run to completion even if the user restores the image details to the spec. Currently the operation will complete in Ironic but the Host would return to the Provisioned state.

The allowable state transitions are based on those in the state diagram design doc, with some modifications to prevent substantive changes from existing behaviour. (For example, the Host can transition to the Registering state from any other state, and then transition back again once registration is complete.) We can iterate toward removing these differences over time.

As an added bonus, coverage analysis can now tell us what portion of the state transitions are covered by unit tests:

```
host_state_machine.go:128:		handleNone			100.0%
host_state_machine.go:136:		handleRegistering		66.7%
host_state_machine.go:160:		handleRegistrationError		0.0%
host_state_machine.go:169:		handleInspecting		100.0%
host_state_machine.go:177:		handleMatchProfile		100.0%
host_state_machine.go:185:		handleExternallyProvisioned	57.1%
host_state_machine.go:201:		handleReady			100.0%
host_state_machine.go:213:		handleProvisioning		77.8%
host_state_machine.go:228:		handleProvisioningError		0.0%
host_state_machine.go:239:		handleProvisioned		0.0%
host_state_machine.go:248:		handlePowerManagementError	0.0%
host_state_machine.go:261:		handleDeprovisioning		0.0%
host_state_machine.go:280:		handleDeleting			100.0%
```


zaneb:
@dhellmann @mhrivnak After playing around with this branch for a while this feels like the right direction to be moving in (though I don't know if this is the right time to be merging a change like this).

If y'all agree then the next step after this is to decide what to do about registration errors. Currently we bounce to the Registering state and back again when credentials change, meaning we're effectively still relying on the old non-stateful behaviour to figure out what state to go back to. Fixing that is relatively straightforward (just re-register outside of the state machine without changing the state), but if an error occurs we move to the RegistrationFailed state and it's less clear what to do about that. (Similar problems exist, but to a much lesser extent, with the PowerManagementError state.)

One option would be to get rid of the error provisioning states altogether and rely on the ErrorMessage string and the operational state to convey the information. This seems like it might be what we were hinting at when creating the current state diagram? (And it does seem wrong that after a failure of some sort you can no longer tell at a glance what state the host is in.) @jtomasek how would that potentially affect the UI?

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -364,67 +261,82 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (result re
...
+	// we requeue the host as we will not know if they create the secret
+	// at some point in the future.
+	case *ResolveBMCSecretRefError:
+		saveErr := r.setErrorCondition(request, host, err.Error())
+		if saveErr != nil {
+			return reconcile.Result{Requeue: true}, saveErr
+		}
+		// Only publish the event if we do not have an error
+		// after saving so that we only publish one time.
+		r.publishEvent(request, host.NewEvent("BMCCredentialError", err.Error()))
```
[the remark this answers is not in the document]
zaneb: I submitted #313 to fix this.

jtomasek:
Not necessarily related to this work, but would it be hard to change the `Ready` state name to `Available`? `Ready` is also used by Node resource status and there are potential cases in the UI where these status names collide. I am able to do this change quite simply on the UI side but it would be ideal if states which UI displays matched the BM resource.

zaneb:
Good thought. That seems easy enough. I opened #315 so we don't forget it.

pkg/controller/baremetalhost/host_state_machine.go
```diff
@@ -0,0 +1,285 @@
...
+	NextState           metal3v1alpha1.ProvisioningState
+	Reconciler          *ReconcileBareMetalHost
+	Provisioner         provisioner.Provisioner
+	RequeueDespiteError bool
+}
+
+func newHostStateMachine(host *metal3v1alpha1.BareMetalHost,
+	reconciler *ReconcileBareMetalHost,
+	provisioner provisioner.Provisioner) *hostStateMachine {
+	currentState := host.Status.Provisioning.State
```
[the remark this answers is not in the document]
dhellmann: OK, I get it. Adding a comment about that would help.

pkg/controller/baremetalhost/host_state_machine.go
```diff
@@ -0,0 +1,294 @@
...
+	} else {
+		info.log.Info("No handler found for state", "state", initialState)
+		err = fmt.Errorf("No handler found for state \"%s\"", initialState)
+		return
+	}
+
+	return
+}
+
+func (hsm *hostStateMachine) shouldInitiateDelete() bool {
```
dhellmann: This function is named in a way that makes me think it answers a question, but it has a side-effect of actually setting up the next state if delete should happen. I don't know if I want to ask to have the function renamed or have the 2 responsibilities divided into separate functions.

pkg/controller/baremetalhost/host_state_machine.go
```diff
@@ -0,0 +1,294 @@
...
+		// Allow state machine to run to continue deprovisioning.
+		return false
+	case metal3v1alpha1.StateDeleting:
+		// Already in deleting state. Allow state machine to run.
+		return false
+	}
+	return true
+}
+
+func (hsm *hostStateMachine) shouldInitiateRegister(info *reconcileInfo) bool {
```
dhellmann: The same comment about side-effects applies here, too.

zaneb: I have an upcoming [patch](https://github.com/zaneb/baremetal-operator/commit/a65c1d107dece244490157db31dcfc76882fd8a6) that replaces this function with one called `ensureRegistered`, so that should resolve this.

pkg/controller/baremetalhost/host_state_machine.go
zaneb: Will fix in a follow-up. `checkInitiateDelete()` maybe?

pkg/controller/baremetalhost/host_state_machine.go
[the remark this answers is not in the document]
zaneb: Ack, will add one in a follow-up.

zaneb:
I had to push some changes to make the bot run the tests, so I just added the trivial fixups on the end. Reapproving.

/lgtm

zaneb:
sigh.

PR #287, opened 2019-08-26
Parameterize the BMC address

maelk:
Retrieve the query elements if any from the BMC address and pass them to the BMC drivers.
For redfish, add the URL parameters to the ironic options.

This is built on and depends on https://github.com/metal3-io/baremetal-operator/pull/284

### DIFF
--- a/pkg/bmc/access.go
+++ b/pkg/bmc/access.go
@@ -10,7 +10,7 @@ import (
 
 // AccessDetailsFactory describes a callable that returns a new
 // AccessDetails based on the input parameters.
-type AccessDetailsFactory func(bmcType, portNum, hostname, path string) (AccessDetails, error)
+type AccessDetailsFactory func(parsedURL *url.URL) (AccessDetails, error)
 
 var factories = map[string]AccessDetailsFactory{}
 
@@ -48,9 +48,9 @@ type AccessDetails interface {
 	BootInterface() string
 }
 
-func getTypeHostPort(address string) (bmcType, host, port, path string, err error) {
+func getParsedURL(address string) (parsedURL *url.URL, err error) {
 	// Start by assuming "type://host:port"
-	parsedURL, err := url.Parse(address)
+	parsedURL, err = url.Parse(address)
 	if err != nil {
 		// We failed to parse the URL, but it may just be a host or
 		// host:port string (which the URL parser rejects because ":"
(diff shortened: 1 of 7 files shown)


pkg/bmc/access.go
```diff
@@ -93,14 +94,14 @@ func getTypeHostPort(address string) (bmcType, host, port, path string, err erro
...
-	return bmcType, host, port, path, nil
+	return bmcType, host, port, path, query, nil
 }
 
 // NewAccessDetails creates an AccessDetails structure from the URL
 // for a BMC.
 func NewAccessDetails(address string) (AccessDetails, error) {
 
-	bmcType, host, port, path, err := getTypeHostPort(address)
+	bmcType, host, port, path, query, err := getTypeHostPort(address)
```
dhellmann: What do you think about passing the parsed URL to the factory directly? We could replace getTypeHostPort() with a function that tries to parse a URL string, and if it can't, construct a parsed URL for itself based on the "host:port" syntax.

maelk: that would be much cleaner indeed. I'll make the change and check the CI

maelk:
I extracted the modification to user url.URL instead of the list of strings passed to the driver. This PR is mostly for discussion on https://github.com/metal3-io/baremetal-operator/issues/286 . If #284  and #288 are merged, then I'll rebase this PR that will have only minimum changes in redfish driver. But following @zaneb 's comments I am looking into adding a "insecure" flag. If we go with that, then I'll just close this pull request

hemanthnakkina:
@maelk Do you have plans to add InspectInterface to each of the BMC types? 
From ironic perspective, the inspect interface can be set to redfish and use Ironic API to get the hardware inspection details instead of Ironic Introspection API.

maelk:
@hemanthnakkina as discussed offline, we do not have plans related to this. Go ahead.

dhellmann:
What did we decide about the approach for this? Do we want URLs with parameters, or do we want explicit fields?

zaneb:
#288 was merged, so if this were rebased it would be a lot simpler now. However, I'm still not a big fan of the query parameters and stand by my comment on #286.

maelk:
I'll resume this work soon. I can implement the "insecure" flag as a comparison

maelk:
/close
An insecure flag will be added in #396

PR #288, opened 2019-08-27
Modify BMC drivers to handle url.URL instead of separate arguments and refactor BMC tests

maelk:
This enhancements changes the BMC drivers factory method to accept a url.URL object instead of separate string arguments, allowing more flexibility in passing information to the BMC drivers.

### DIFF
--- a/pkg/bmc/access_test.go
+++ b/pkg/bmc/access_test.go
@@ -16,17 +16,23 @@ func TestParse(t *testing.T) {
 		Address     string
 		Type        string
 		Host        string
+		Hostname    string
 		Port        string
 		Path        string
 		ExpectError bool
+		Query       map[string][]string
 	}{
 		{
 			Scenario: "libvirt url",
-			Address:  "libvirt://192.168.122.1:6233/",
+			Address:  "libvirt://192.168.122.1:6233/?abc=def",
 			Type:     "libvirt",
 			Port:     "6233",
 			Host:     "192.168.122.1",
+			Hostname: "192.168.122.1:6233",
 			Path:     "/",
+			Query: map[string][]string{
+				"abc": []string{"def"},
+			},
 		},
 
 		{
@@ -35,6 +41,7 @@ func TestParse(t *testing.T) {
 			Type:     "ipmi",
 			Port:     "",
 			Host:     "192.168.122.1",
+			Hostname: "192.168.122.1",
 			Path:     "",
 		},
 
@@ -44,6 +51,7 @@ func TestParse(t *testing.T) {
 			Type:     "ipmi",
 			Port:     "",
 			Host:     "my.favoritebmc.com",
+			Hostname: "my.favoritebmc.com",
 			Path:     "",
 		},
 
@@ -53,6 +61,7 @@ func TestParse(t *testing.T) {
 			Type:     "ipmi",
 			Port:     "6233",
 			Host:     "192.168.122.1",
+			Hostname: "192.168.122.1:6233",
 			Path:     "",
 		},
 
@@ -63,6 +72,7 @@ func TestParse(t *testing.T) {
 			Port:     "6233",
 			Host:     "fe80::fc33:62ff:fe83:8a76",
 			Path:     "",
+			Hostname: "[fe80::fc33:62ff:fe83:8a76]:6233",
 		},
 
 		{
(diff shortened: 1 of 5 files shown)


maelk:
@dhellmann I created this pull request to separate this from the on-going discussion on Redfish configuration. Built on https://github.com/metal3-io/baremetal-operator/pull/284.

maelk:
@zaneb I fixed the issues in #288 and rebased this on top. Please use #288 to discuss about using *url.URL as driver parameter.

maelk:
@zaneb any further comments ?

pkg/bmc/redfish.go
```diff
@@ -22,23 +26,30 @@ func newRedfishAccessDetails(bmcType, portNum, hostname, path string) (AccessDet
 			}
 		}
 	}
```
zaneb: Lines 18-28 should just be deleted.

maelk: Ok, I'll rebase on master

pkg/bmc/access_test.go
```diff
@@ -11,162 +11,192 @@ func init() {
...
-		t.Fatalf("unexpected path: %q", A)
+	if len(url.Query()) != 0 {
+		t.Fatalf("unexpected query length: %q", len(url.Query()))
 	}
 }
 
 func TestParseHostNoPortIPv6(t *testing.T) {
-	// They either have to give us a port or a URL scheme with IPv6.
-	_, _, _, _, err := getTypeHostPort("[fe80::fc33:62ff:fe83:8a76]")
+	// TBy default we fallback to IPMI, also with IPv6.
```
zaneb: typo, but also this new comment doesn't seem to match the code.

pkg/bmc/access_test.go
```diff
@@ -11,162 +11,192 @@ func init() {
 }
 
 func TestParseLibvirtURL(t *testing.T) {
-	T, H, P, A, err := getTypeHostPort("libvirt://192.168.122.1:6233/")
+	url, err := getParsedURL("libvirt://192.168.122.1:6233/?abc=def")
```
zaneb: Rather than change every one of these tests it might be easier to just define a local `getTypeHostPort()` function.

dhellmann: I want to convert all of these tests to be table-driven anyway, but I can do that in a separate PR. If I do that first, you'll have to rebase this, but you'll have fewer renames.

maelk: I can take care of modifying them to be table-driven at the same time if you wish.

pkg/bmc/access_test.go
maelk: thanks

pkg/bmc/access_test.go
dhellmann: See #300

maelk: I have now modified the tests to be table-driven @dhellmann.

maelk: Sorry, I missed this, I'll rebase on it and take it further

dhellmann: @maelk I went ahead and merged the other change. That should make reviewing your rebase a little easier. And thanks for the additional changes you made in that file. I like the way you cleaned up the other tests, too.

maelk: @dhellmann Thanks for the merge, I have now rebased this PR that should now be ready for review. Thank you

pkg/bmc/access_test.go
```diff
@@ -176,338 +196,258 @@ func TestParse(t *testing.T) {
...
 
-			if P != tc.Port {
-				t.Fatalf("expected port %q but got %q", tc.Port, P)
+			if url.Port() != tc.Port {
+				t.Fatalf("expected port %q but got %q", tc.Port, url.Port())
 			}
 
-			if H != tc.Host {
-				t.Fatalf("expected host %q but got %q", tc.Host, H)
+			if url.Hostname() != tc.Host {
```
dhellmann: It's unfortunate that the test data maps url.Hostname() to tc.Host and url.Host to tc.Hostname. We can live with it for now, and maybe fix it in another PR.

zaneb: LOL.
Personally I don't think it even necessary to check url.Host. If url.Hostname() and url.Port() are right then it can only be because we got the right data.

maelk: ok, I can create another PR later to fix this. If this is ok and gets merged then we can focus on #284 . I'll do the fix after. Thank you

Issue #273, opened 2019-08-05
update power management to default to soft shutdown


dhellmann:
The current power-off behavior when Online=false is to just turn off the power. We should default to a soft power off, which would allow the BMC to signal to the OS that the power is going to be pulled and let the OS shutdown cleanly. We will need to choose a reasonable timeout for that shutdown process.

https://docs.openstack.org/api-ref/baremetal/?expanded=change-node-power-state-detail

Xenwar:
I will look into this

PR #294, opened 2019-08-29
Default to use soft power off instead of hard power off

tiendc:
#273

### DIFF
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -32,6 +32,7 @@ var deprovisionRequeueDelay = time.Second * 10
 var provisionRequeueDelay = time.Second * 10
 var powerRequeueDelay = time.Second * 10
 var introspectionRequeueDelay = time.Second * 15
+var softPowerOffTimeout = time.Second * 180
 var deployKernelURL string
 var deployRamdiskURL string
 var ironicEndpoint string
@@ -39,9 +40,10 @@ var inspectorEndpoint string
 
 const (
 	// See nodes.Node.PowerState for details
-	powerOn   = "power on"
-	powerOff  = "power off"
-	powerNone = "None"
+	powerOn      = "power on"
+	powerOff     = "power off"
+	softPowerOff = "soft power off"
+	powerNone    = "None"
 )
 
 func init() {
@@ -1144,12 +1146,17 @@ func (p *ironicProvisioner) changePower(ironicNode *nodes.Node, target nodes.Tar
 		return result, nil
 	}
 
+	powerStateOpts := nodes.PowerStateOpts{
+		Target: target,
+	}
+	if target == softPowerOff {
+		powerStateOpts.Timeout = int(softPowerOffTimeout.Seconds())
+	}
+
 	changeResult := nodes.ChangePowerState(
 		p.client,
 		ironicNode.UUID,
-		nodes.PowerStateOpts{
-			Target: target,
-		})
+		powerStateOpts)
 
 	switch changeResult.Err.(type) {
 	case nil:
@@ -1159,7 +1166,11 @@ func (p *ironicProvisioner) changePower(ironicNode *nodes.Node, target nodes.Tar
 		p.log.Info("host is locked, trying again after delay", "delay", powerRequeueDelay)
 		result.Dirty = true
 		result.RequeueAfter = powerRequeueDelay
-		return result, nil
+		return result, HostLockedError{Address: p.host.Spec.BMC.Address}
+	case gophercloud.ErrDefault400:
+		// Error 400 Bad Request means target power state is not supported by vendor driver
+		p.log.Info("power change error", "message", changeResult.Err)
+		return result, SoftPowerOffUnsupportedError{Address: p.host.Spec.BMC.Address}
 	default:
 		p.log.Info("power change error", "message", changeResult.Err)
 		return result, errors.Wrap(changeResult.Err, "failed to change power state")
@@ -1205,6 +1216,29 @@ func (p *ironicProvisioner) PowerOn() (result provisioner.Result, err error) {
 func (p *ironicProvisioner) PowerOff() (result provisioner.Result, err error) {
 	p.log.Info("ensuring host is powered off")
 
+	result, err = p.softPowerOff()
+	if err != nil {
+		switch err.(type) {
+		// In case of soft power off is unsupported or has failed,
+		// we activate hard power off.
+		case SoftPowerOffUnsupportedError, SoftPowerOffFailed:
+			return p.hardPowerOff()
+		case HostLockedError:
+			result.RequeueAfter = powerRequeueDelay
+			result.Dirty = true
+			return result, nil
+		default:
+			result.RequeueAfter = powerRequeueDelay
+			return result, err
+		}
+	}
+	return result, nil
+}
+
+// hardPowerOff sends 'power off' request to BM node and waits for the result
+func (p *ironicProvisioner) hardPowerOff() (result provisioner.Result, err error) {
+	p.log.Info("ensuring host is powered off by \"hard power off\" command")
+
 	ironicNode, err := p.findExistingHost()
 	if err != nil {
 		return result, errors.Wrap(err, "failed to find existing host")
@@ -1223,6 +1257,44 @@ func (p *ironicProvisioner) PowerOff() (result provisioner.Result, err error) {
 			return result, errors.Wrap(err, "failed to power off host")
 		}
 		p.publisher("PowerOff", "Host powered off")
+	}
+
+	return result, nil
+}
+
+// softPowerOff sends 'soft power off' request to BM node.
+// If soft power off is not supported, the request ends with an error.
+// Otherwise the request ends with no error and the result should be
+// checked later via node fields "power_state", "target_power_state"
+// and "last_error".
+func (p *ironicProvisioner) softPowerOff() (result provisioner.Result, err error) {
+	p.log.Info("ensuring host is powered off by \"soft power off\" command")
+
+	ironicNode, err := p.findExistingHost()
+	if err != nil {
+		return result, errors.Wrap(err, "failed to find existing host")
+	}
+
+	if ironicNode.PowerState != powerOff {
+		targetState := ironicNode.TargetPowerState
+		// If the target state is either powerOff or softPowerOff, then we should wait
+		if targetState == powerOff || targetState == softPowerOff {
+			p.log.Info("waiting for power status to change")
+			result.RequeueAfter = powerRequeueDelay
+			result.Dirty = true
+			return result, nil
+		}
+		// If the target state is unset while the last error is set,
+		// then the last execution of soft power off has failed.
+		if targetState == "" && ironicNode.LastError != "" {
+			return result, SoftPowerOffFailed{Address: p.host.Spec.BMC.Address}
+		}
+		result, err = p.changePower(ironicNode, nodes.SoftPowerOff)
+		if err != nil {
+			result.RequeueAfter = powerRequeueDelay
+			return result, err
+		}
+		p.publisher("PowerOff", "Host soft powered off")
 	}
 
 	return result, nil
--- a/pkg/provisioner/ironic/errors.go
+++ b/pkg/provisioner/ironic/errors.go
@@ -1,0 +1,38 @@
+package ironic
+
+import (
+	"fmt"
+)
+
+// SoftPowerOffUnsupportedError is returned when the BMC does not
+// support soft power off.
+type SoftPowerOffUnsupportedError struct {
+	Address string
+}
+
+func (e SoftPowerOffUnsupportedError) Error() string {
+	return fmt.Sprintf("Soft power off is unsupported on BMC %s",
+		e.Address)
+}
+
+// SoftPowerOffFailed is returned when the soft power off command
+// finishes with failure.
+type SoftPowerOffFailed struct {
+	Address string
+}
+
+func (e SoftPowerOffFailed) Error() string {
+	return fmt.Sprintf("Soft power off has failed on BMC %s",
+		e.Address)
+}
+
+// HostLockedError is returned when the BMC host is
+// locked.
+type HostLockedError struct {
+	Address string
+}
+
+func (e HostLockedError) Error() string {
+	return fmt.Sprintf("BMC %s host is locked",
+		e.Address)
+}


nordixinfra:

derekhiggins:
Is this the call that is used for fencing? If so should it remain a hard power off?

pkg/provisioner/provisioner.go
```diff
@@ -72,6 +72,10 @@ type Provisioner interface {
 	// PowerOff ensures the server is powered off independently of any image
 	// provisioning operation.
 	PowerOff() (result Result, err error)
+
+	// SoftPowerOff tries to perform soft shutdown first
+	// If soft shutdown is unspported, it performs hard shutdown
+	SoftPowerOff() (result Result, err error)
```
dhellmann: Do we need to add a new method? Could we change the implementation of PowerOff() instead?

pkg/provisioner/ironic/ironic.go
```diff
@@ -1346,3 +1356,51 @@ func (p *ironicProvisioner) PowerOff() (result provisioner.Result, err error) {
...
+
+		if ironicNode.TargetPowerState == powerOff || ironicNode.TargetPowerState == softPowerOff {
+			p.log.Info("waiting for power status to change")
+			result.RequeueAfter = powerRequeueDelay
+			return result, nil
+		}
+		result, err = p.changePower(ironicNode, nodes.SoftPowerOff)
+		if err != nil {
+			// Soft power off is not supported by vendor driver, uses PowerOff()
+			if strings.HasPrefix(err.Error(), "driver does not support target power state") {
```
dhellmann: Is there some way for us to detect whether the driver supports soft power off before we get here and try to use it? Could we store a setting in the Status section of the host, so the user knows what to expect when the ask for the host to be powered off?

pkg/provisioner/ironic/ironic.go
```diff
@@ -1346,3 +1356,51 @@ func (p *ironicProvisioner) PowerOff() (result provisioner.Result, err error) {
...
+// If soft shutdown is unspported, it performs hard shutdown
+func (p *ironicProvisioner) SoftPowerOff() (result provisioner.Result, err error) {
+	p.log.Info("ensuring host is soft powered off")
+
+	ironicNode, err := p.findExistingHost()
+	if err != nil {
+		return result, errors.Wrap(err, "failed to find existing host")
+	}
+
+	if ironicNode.PowerState != powerOff {
```
dhellmann: See if you can reorganize the logic of this function to include more early exits and fewer levels of indentation. For example, here, if ironicNode.PowerState == powerOff { ... return } would mean we would not need an "else" block, and all of the code on lines 1370-1399 could be dedented one level.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1346,3 +1356,51 @@ func (p *ironicProvisioner) PowerOff() (result provisioner.Result, err error) {
...
+			return result, nil
+		}
+		result, err = p.changePower(ironicNode, nodes.SoftPowerOff)
+		if err != nil {
+			// Soft power off is not supported by vendor driver, uses PowerOff()
+			if strings.HasPrefix(err.Error(), "driver does not support target power state") {
+				return p.PowerOff()
+			} else {
+				result.RequeueAfter = powerRequeueDelay
+				return result, errors.Wrap(err, "failed to soft power off host")
```
dhellmann: Errors are retried immediately. If this case is something that might mean no retry is reasonable or that a retry should be delayed, the error message should be stored in the result so it can be attached to the host later.

dhellmann:
> Is this the call that is used for fencing? If so should it remain a hard power off?

A soft power off is the default, but if that fails we still yank the power.

dhellmann:
@tiendc thank you for working on this!

derekhiggins:
> > Is this the call that is used for fencing? If so should it remain a hard power off?
> 
> A soft power off is the default, but if that fails we still yank the power.

ack, thanks

pkg/provisioner/ironic/ironic.go
tiendc: I've checked the Ironic API, currently there is no API for retrieving supported power states.

docs/api.md
```diff
@@ -164,6 +164,15 @@ operation.  One of:
   * *power management error* -- The host could not be powered on or
     off, as requested through the *online* field setting.
 
+*softPowerOffStatus* -- The status of the soft power off command.
+Value is one of the following:
+  * *<empty string>* -- Soft power off is not used on the node.
+  * *unsupported* -- Soft power off is not supported on the node.
+  * *triggered* -- Soft power off is triggered on the node but
```
dhellmann: Do we really need to track the soft power off status separately?

If we do, I think instead of reflecting it in a new status field, we should see if we can combine it with the provisioning status and make that a top level field on the status structure.

We would never have something in "provisioning" with a soft power off status of "triggered", for example, right?

pkg/provisioner/ironic/ironic.go
```diff
@@ -1325,6 +1336,19 @@ func (p *ironicProvisioner) PowerOn() (result provisioner.Result, err error) {
 func (p *ironicProvisioner) PowerOff() (result provisioner.Result, err error) {
 	p.log.Info("ensuring host is powered off")
 
+	// Tries soft power off first, if it fails, performs hard power off
+	result, err = p.softPowerOff()
+	if err != nil {
```
dhellmann: We only want to switch to the hard power off mode if we get the very specific 400 error. If we get a 409 we want to pause and try the soft power off again, for example.

I think we want to define a new error type so that we can convert the 400 error from line 1291 to our custom type, and then check for that type here instead of just checking against nil.

tiendc: In my test environment, I have a Fujitsu BM server. It requires the OS must install a specific agent to allow soft power off. So if the agent is not install, any try to perform soft power off will fail regardless of support from Ironic and Fujitsu driver for Ironic. I think if we retry the action when failed, we should limit the number of it, say 3 times. Do you have any suggestion?

deploy/crds/metal3.io_baremetalhosts_crd.yaml
```diff
@@ -436,6 +436,20 @@ spec:
               - ID
               - state
               type: object
+            softPowerOff:
```
dhellmann: Instead of adding these new fields, could we use the existing power status field? It would only allow us to try using soft power off one time, but I think that's OK.

pkg/apis/metal3/v1alpha1/baremetalhost_types.go
```diff
@@ -89,6 +89,17 @@ const (
...
 	// ready for deletion
 	StateDeleting ProvisioningState = "deleting"
+
+	// StateSoftPowerOffTriggered means host if soft powered off
+	// but not yet finished
+	StateSoftPowerOffTriggered ProvisioningState = "soft power off triggered"
+
+	// StateSoftPowerOffUnsupported means soft powered off command
+	// is not supported by the driver
+	StateSoftPowerOffUnsupported ProvisioningState = "soft power off unsupported"
```
dhellmann: I don't think we need this state at all, do we? If we receive a power off request and the driver does not support soft power off we should just use the hard power off API.

pkg/apis/metal3/v1alpha1/baremetalhost_types.go
```diff
@@ -89,6 +89,17 @@ const (
...
+
+	// StateSoftPowerOffError means soft powered off fails
+	StateSoftPowerOffError ProvisioningState = "soft power off error"
```
dhellmann: We don't have a separate hard power off error. Do we need to track errors with soft power off?

tiendc:
It seems I did something wrong with git, so the pull request contains a commit that is not mine. I will try to fix it.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1295,3 +1338,53 @@ func (p *ironicProvisioner) PowerOff() (result provisioner.Result, err error) {
...
+		// Although request is "soft power off", target state is still "power off"
+		if ironicNode.TargetPowerState == powerOff {
+			p.log.Info("waiting for power status to change")
+			result.RequeueAfter = powerRequeueDelay
+			result.Dirty = true
+			return result, nil
+		}
+		// Checks last error of ironic node, if there is one happened
+		// then soft power off has failed
+		if p.host.Status.SoftPowerOffStatus == metal3v1alpha1.StateSoftPowerOffTriggered {
```
zaneb: If the Ironic API docs are correct, then we don't need to store this status in the CRD. Avoiding that would be preferable if it's possible.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1295,3 +1338,53 @@ func (p *ironicProvisioner) PowerOff() (result provisioner.Result, err error) {
...
+			if ironicNode.LastError != "" {
+				err = errors.New(ironicNode.LastError)
+				err = errors.Wrap(err, "soft power off failed")
```
zaneb: I'm not sure this is the sort of thing we should be returning an error on. I think this should just be an ordinary requeue.
This would be a good thing to count in Prometheus though.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1295,3 +1338,53 @@ func (p *ironicProvisioner) PowerOff() (result provisioner.Result, err error) {
...
+				p.host.Status.SoftPowerOffStatus = metal3v1alpha1.StatePowerManagementError
```
zaneb: For starters, this will never get written if we return an error.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1295,3 +1338,53 @@ func (p *ironicProvisioner) PowerOff() (result provisioner.Result, err error) {
...
+func (p *ironicProvisioner) softPowerOff() (result provisioner.Result, err error) {
+	p.log.Info("ensuring host is powered off by \"soft power off\" command")
+
+	ironicNode, err := p.findExistingHost()
+	if err != nil {
+		return result, errors.Wrap(err, "failed to find existing host")
+	}
+
+	if ironicNode.PowerState != powerOff {
+		// Although request is "soft power off", target state is still "power off"
```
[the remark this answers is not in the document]
tiendc: When I test in practical, I only see the target power state is "power off" for "soft power off" command. Please see the below links. I will re-check it and confirm to you later. 

https://github.com/openstack/ironic/blob/master/ironic/conductor/manager.py#L322
https://github.com/openstack/ironic/blob/master/ironic/conductor/utils.py#L267
https://github.com/openstack/ironic/blob/master/ironic/conductor/utils.py#L158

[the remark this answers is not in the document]
zaneb: Yeah, it sure looks like the API docs are wrong doesn't it.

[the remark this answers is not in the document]
zaneb: I opened an [issue](https://storyboard.openstack.org/#!/story/2007089) and proposed a [patch to fix the docs](https://review.opendev.org/702542). We'll see what the Ironic team says about it.

pkg/provisioner/ironic/ironic.go
zaneb: So the docs were incorrect.
Still, with no local state required, couldn't we do something like:
* get the status
* if the target power state is not `power off`:
  * start a soft power off with timeout
* else (target state is `power off`):
  * if there's an error:
    * start a hard power off
  * else (no error):
    * keep waiting

How could that fail? The only way I can think of is the periodic power state check in Ironic overwriting the error before we see it. But if our timeout << power state check period, then I think we should always end up doing the Right Thing eventually.

tiendc: I change to check the power_state and target_power_state instead of last_error. It seems if soft power off fails, target_power_state will be unset.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1295,3 +1327,46 @@ func (p *ironicProvisioner) PowerOff() (result provisioner.Result, err error) {
...
+		targetState := ironicNode.TargetPowerState
+		// If the target state is either powerOff or softPowerOff, then we should wait
+		if targetState == powerOff || targetState == softPowerOff {
+			p.log.Info("waiting for power status to change")
+			result.RequeueAfter = powerRequeueDelay
+			result.Dirty = true
+			return result, nil
+		}
+		// If the target state is unset, then soft power off has failed
+		if p.host.Status.SoftPowerOffTriggered && targetState == "" {
```
zaneb: Can we check the target_state + the last_error to distinguish between the cases where we haven't started a power off vs. a previous attempt has timed out, instead of having to set Status.SoftPowerOffTriggered?

tiendc: I've tried your suggestion. It seems fine for the common cases I test. However, I think using the LastError may not be reliable as using a "trigger" flag.

q384566678:
@zaneb @dhellmann 
Hi, I will take over from @tiendc  to continue this work, and I made some changes to the code based on your previous comments. PTAL.

q384566678:
@dhellmann PTAL

q384566678:
@zaneb @dhellmann  PTAL

q384566678:
@zaneb @dhellmann PTAL

pkg/provisioner/ironic/ironic.go
```diff
@@ -1205,6 +1216,28 @@ func (p *ironicProvisioner) PowerOn() (result provisioner.Result, err error) {
...
 
+	result, err = p.softPowerOff()
+	if err != nil {
+		switch err.(type) {
+		// In case of soft power off is unsupported or has failed,
+		// we activate hard power off.
+		case SoftPowerOffUnsupportedError, SoftPowerOffFailed:
+			return p.hardPowerOff()
+		case HostLockedError:
+			return p.softPowerOff()
```
zaneb, line 1227: I don't think we want to call again straight away in this case, but rather requeue with a delay so we can try again.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1205,6 +1216,28 @@ func (p *ironicProvisioner) PowerOn() (result provisioner.Result, err error) {
 func (p *ironicProvisioner) PowerOff() (result provisioner.Result, err error) {
 	p.log.Info("ensuring host is powered off")
 
+	result, err = p.softPowerOff()
```
zaneb, line 1219: The result values returned here are not actually used anywhere, so why bother returning them?

pkg/provisioner/ironic/ironic.go
```diff
@@ -1227,3 +1260,41 @@ func (p *ironicProvisioner) PowerOff() (result provisioner.Result, err error) {
...
+		if targetState == powerOff || targetState == softPowerOff {
+			p.log.Info("waiting for power status to change")
+			result.RequeueAfter = powerRequeueDelay
+			result.Dirty = true
+			return result, nil
+		}
+		// If the target state is unset while the last error is set,
+		// then the last execution of soft power off has failed.
+		if targetState == "" && ironicNode.LastError != "" {
+			return result, SoftPowerOffFailed{Address: p.host.Spec.BMC.Address}
```
zaneb, line 1289: This just logs an error and requeues. There's no reason to think we won't just end up back here in an infinite loop, is there?
This is where we want to do a hard power off I think.

pkg/provisioner/ironic/ironic.go
q384566678: These two values are returned as the result of the `PowerOff` function, I refer to the `PowerOn` and `PowerOff` functions, they will have these two return values.

pkg/provisioner/ironic/ironic.go
q384566678:
```
		result, err = p.changePower(ironicNode, nodes.SoftPowerOff)
		if err != nil {
			result.RequeueAfter = powerRequeueDelay
			return result, err
		}
```
In the `softPowerOff `function, when these errors occur, `powerRequeueDelay` will be added, is it necessary to add additional delay?

pkg/provisioner/ironic/ironic.go
q384566678: I do n’t know if you marked the wrong code. In the code you marked, `SoftPowerOffFailed` is returned, and then go back to execute `hardPowerOff`.
```
		// In case of soft power off is unsupported or has failed,
		// we activate hard power off.
		case SoftPowerOffUnsupportedError, SoftPowerOffFailed:
			return p.hardPowerOff()
```

pkg/provisioner/ironic/ironic.go
zaneb: That's a delay after calling it a second time. So you call it twice in succession, then wait to go around again. It's extremely unlikely that the second call will succeed if the first didn't, so it's a waste. Just requeue with a delay.

pkg/provisioner/ironic/ironic.go
zaneb: Sorry, I got confused about what gets called from where. You are correct.

pkg/provisioner/ironic/ironic.go
q384566678: Updated, before calling `p.softPowerOff ()`, I added a `sleep` function. PTAL

zaneb: Sleeping blocks the controller from reconciling any other resources. Just return, we'll get back here soon enough.

q384566678: thanks, updated, PTAL.

q384566678: @zaneb PTAL

q384566678:
@zaneb @dhellmann @maelk Can someone help review this patch? Thanks a lot!

q384566678:
@zaneb Why has this PR been stuck in this state? Is this PR currently waiting for the result of test-integration (almost a week has passed)? Is there anything I can do to promote this PR?

zaneb:
I'm not sure why the integration test didn't run. @maelk any idea?

dhellmann:
> I'm not sure why the integration test didn't run. @maelk any idea?

Perhaps only org members can trigger the test job?

pkg/provisioner/ironic/ironic.go
dhellmann: The 409 error just means that ironic itself is too busy to handle our request (or more likely that it is already doing something with the host and cannot send multiple instructions). But your point about only retrying a few times makes a lot of sense, for other types of errors.

zaneb:
> Perhaps only org members can trigger the test job?

I also tried unsuccessfully to trigger it last week, but on reflection that may have been before the regex was changed to allow it not to be the only line in the comment.

q384566678:
@zaneb @dhellmann test-integration has passed. Please continue to review, thanks.

q384566678:
@zaneb PTAL, thanks.

q384566678:
@zaneb @dhellmann @maelk At present, this PR has obtained an approve, and test-integration has passed. Currently requires lgtm label. Is there anything else I can do to advance this PR?

Issue #215, opened 2019-06-12
show which cleaning step is running during (de)provision operations


dhellmann:
Cleaning can take a little while, so we may want to expose which step is running. The [ironic docs](https://docs.openstack.org/ironic/latest/admin/cleaning.html#what-cleaning-step-is-running) explain which field has the information. We can probably just update the human-readable provisioning status message with that detail when it changes.

maelk:
I'll look into this

Jaakko-Os:
Hi. Please give me feedback in case I somehow missed the "beef" of this task. Testing was run in metal3-dev-env with the tests from upstream.

PR #299, opened 2019-09-09
show which cleaning step is running during (de)provision operations

Jaakko-Os:
The cleaning step is set in the ironic client, running in bare metal operator. BMO fetches the host data which .i.a contains the step info.
The cleaning step field can be an empty/non-empty string.

Closes: #215

### DIFF
--- a/deploy/crds/metal3_v1alpha1_baremetalhost_crd.yaml
+++ b/deploy/crds/metal3_v1alpha1_baremetalhost_crd.yaml
@@ -50,7 +50,6 @@ spec:
     status: {}
   validation:
     openAPIV3Schema:
-      description: BareMetalHost is the Schema for the baremetalhosts API
       properties:
         apiVersion:
           description: 'APIVersion defines the versioned schema of this representation
@@ -65,7 +64,6 @@ spec:
         metadata:
           type: object
         spec:
-          description: BareMetalHostSpec defines the desired state of BareMetalHost
           properties:
             bmc:
               description: How do we connect to the BMC?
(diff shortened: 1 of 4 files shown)


zaneb:
What sets the `host.Status.Provisioning.Step` field?

pkg/provisioner/ironic/ironic.go
```diff
@@ -1167,12 +1166,26 @@ func (p *ironicProvisioner) Deprovision() (result provisioner.Result, err error)
 		p.log.Info("cleaning")
 		result.Dirty = true
 		result.RequeueAfter = deprovisionRequeueDelay
+		stepClean, stepCleanOk := ironicNode.CleanStep["step"]
+
+		if stepCleanOk == true {
+			p.host.Status.Provisioning.Step = fmt.Sprintf("%v", stepClean)
```
dhellmann: According to the [API docs](https://docs.openstack.org/api-ref/baremetal/?expanded=show-node-details-detail) the clean step is a string. Why do we need the Sprintf() call? Is the go value not a string?

pkg/provisioner/ironic/ironic.go
```diff
@@ -1167,12 +1166,26 @@ func (p *ironicProvisioner) Deprovision() (result provisioner.Result, err error)
 		p.log.Info("cleaning")
 		result.Dirty = true
 		result.RequeueAfter = deprovisionRequeueDelay
+		stepClean, stepCleanOk := ironicNode.CleanStep["step"]
```
dhellmann: Ideally we would update the host before setting result.Dirty so that it's clear why we're setting that flag to true.

pkg/provisioner/ironic/ironic.go
Jaakko-Os: CleanStep is of type 'map[string]interface{}' and thus stepClean is of type 'interface{}'.
p.host.Status.Provisioning. Step is 'string' and I found Sprintf() to be a convenient way for type cast.

pkg/provisioner/ironic/ironic.go
Jaakko-Os: Done

pkg/provisioner/ironic/ironic.go
dhellmann: Are those the types gophercloud uses? I wonder if we should change the library upstream if we know the value will only ever be a string.

Jaakko-Os: Yes the types are from gophercloud.
I don't have an answer for that if 'step' will be a string in the future. My proposal is to merge this PR and possibly have a separate task for the typing.

Jaakko-Os:
Hi!
Any chance of getting lgtm and approved labels for this PR.
Thanks.

zaneb:
This PR needs to be rebased before it can merge.

Jaakko-Os:
rebase done, tested in metal3-dev-env

Jaakko-Os:
rebase done

Jaakko-Os:
Rebased. Please review

deploy/crds/metal3_v1alpha1_baremetalhost_crd.yaml
```diff
@@ -0,0 +1,387 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
dhellmann: How is this file different from deploy/crds/metal3.io_baremetalhosts_crd.yaml

Do we need both? Was this generated automatically?

zaneb: I think the file was renamed at some point. This looks like a rebase artifact.

Jaakko-Os: To me it seems like the this PR has turned out to be obsolete. After several rebases the affected code is quite different compared to where it started from and thus the whole task should be reconsidered.
I would propose to close this PR.
What do you think?

PR #300, opened 2019-09-09
simplify bmc access tests

dhellmann:
Convert the separate test functions for parsing BMC URLs to a table
test to cut down on the repetition.

### DIFF
--- a/pkg/bmc/access_test.go
+++ b/pkg/bmc/access_test.go
@@ -10,163 +10,188 @@ func init() {
 	logf.SetLogger(logf.ZapLogger(true))
 }
 
-func TestParseLibvirtURL(t *testing.T) {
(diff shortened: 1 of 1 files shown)


pkg/bmc/access_test.go
```diff
@@ -10,163 +10,188 @@ func init() {
...
-	}
-	if P != "6233" {
-		t.Fatalf("unexpected port: %q", P)
-	}
-	if A != "" {
-		t.Fatalf("unexpected path: %q", A)
+func TestParse(t *testing.T) {
+	for _, tc := range []struct {
+		Scenario    string
+		URL         string
```
zaneb: Maybe name this something like `bmcAddr`, since we're about to change the output to an actual URL object and this could get confusing?

dhellmann: Done

Issue #298, opened 2019-09-05
let ironic look up the image checksum value


dhellmann:
The ironic provisioner code looks at the checksum value we have and, if it is a URL, fetches the content to pass to ironic. The version of ironic we have in the images now can do that step for itself, so we should let it.

Jaakko-Os:
I'll look into this

hardys:
@derekhiggins did your test issues related to this get resolved yesterday?

We switched the metal3 ironic images to train yesterday, and downstream we also have train-based images available now, so I think this should be good to go, modulo feedback from Derek re his test issues.

PR #304, opened 2019-09-17
let ironic look up the image checksum value

Jaakko-Os:
The Stein version of Ironic supports passing a URL.
When we upgrade, we can stop doing this work ourself.

Closes: #298

### DIFF
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -2,9 +2,6 @@ package ironic
 
 import (
 	"fmt"
-	"io/ioutil"
-	"net/http"
-	"net/url"
 	"os"
 	"sort"
 	"strings"
@@ -276,12 +273,7 @@ func (p *ironicProvisioner) ValidateManagementAccess(credentialsChanged bool) (r
 		}
 
 		if p.host.Spec.Image != nil && p.host.Spec.Image.URL != "" {
-			// FIXME(dhellmann): The Stein version of Ironic supports passing
-			// a URL. When we upgrade, we can stop doing this work ourself.
-			checksum, err := p.getImageChecksum()
-			if err != nil {
-				return result, errors.Wrap(err, "failed to retrieve image checksum")
-			}
+			checksum := p.host.Spec.Image.Checksum
 
 			p.log.Info("setting instance info",
 				"image_source", p.host.Spec.Image.URL,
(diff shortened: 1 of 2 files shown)


elfosardo:
this should work with final builds, since we're installing Stein
it doesn't work on CI images though, as at the moment Rocky is what we're getting there

dhellmann:
> this should work with final builds, since we're installing Stein
> it doesn't work on CI images though, as at the moment Rocky is what we're getting there

OK, if this is going to break our internal CI I would like to wait to merge it until we can resolve that issue.

hardys:
@derekhiggins did your test issues related to this get resolved yesterday?

We switched the metal3 ironic images to train yesterday, and downstream we also have train-based images available now, so I think this should be good to go, modulo feedback from Derek re his test issues.

derekhiggins:
> @derekhiggins did your test issues related to this get resolved yesterday?
> 
> We switched the metal3 ironic images to train yesterday, and downstream we also have train-based images available now, so I think this should be good to go, modulo feedback from Derek re his test issues.

from what I'm told, 

Looks like we have to wait for the CI builds to move to OSP16, the patch to do this merged 
https://github.com/openshift/release/pull/5316

but it has to be deployed and we need other 2 PRs:
https://github.com/openshift/release/pull/5355
https://github.com/openshift/release/pull/5354

hardys:
@derekhiggins I think we should be good to merge this now, then test the backport to openshift/baremetal-operator which will consume the downstream images mentioned?

derekhiggins:
> @derekhiggins I think we should be good to merge this now, then test the backport to openshift/baremetal-operator which will consume the downstream images mentioned?

I havn't tried this patch but yes, I think the problems we had with checksum checking are now resolved.

hardys:
Ok I tested this locally and all works OK, we've got the images updated upstream and downstream now so lets go ahead and merge this.  Thanks for the patience @Jaakko-Os while we worked out the image updates :)

hardys:
Hmm for some reason the check jobs aren't running, @Jaakko-Os perhaps you can push a rebase to see if the tide job then works?

Jaakko-Os:
> Hmm for some reason the check jobs aren't running, @Jaakko-Os perhaps you can push a rebase to see if the tide job then works?

Done. Looks better.

PR #305, opened 2019-09-18
Fix doc of setting up development env with minikube

sshukun:
Example host should be created in namespace metal3 because it is the namespace the operator is watching for changes.


nordixinfra:

longkb:
LGTM :)

dhellmann:
Thanks!

PR #306, opened 2019-09-23
Set Soft Power Off to be the default shutdown mode

Xenwar:
Co-authored-by: Jaakko Kuuskoski <jaakko.kuuskoski@est.tech>
Co-authored-by: maelk <mael.kimmerlin@est.tech>

### DIFF
--- a/pkg/controller/baremetalhost/baremetalhost_controller.go
+++ b/pkg/controller/baremetalhost/baremetalhost_controller.go
@@ -726,8 +726,21 @@ func (r *ReconcileBareMetalHost) manageHostPower(prov provisioner.Provisioner, i
 
 	if info.host.Spec.Online {
 		provResult, err = prov.PowerOn()
+		if err = r.unMarkLastSoftPowerOff(info.host); err != nil {
+			return reconcile.Result{}, errors.Wrap(err, fmt.Sprintf("failed to mark last soft power off time %q", err.Error()))
+		}
+
 	} else {
-		provResult, err = prov.PowerOff()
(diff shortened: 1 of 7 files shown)


nordixinfra:

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -827,6 +840,41 @@ func (r *ReconcileBareMetalHost) saveStatus(host *metal3v1alpha1.BareMetalHost)
...
+}
+
+// Remove soft power off time stamp
+func (r *ReconcileBareMetalHost) unMarkLastSoftPowerOff(host *metal3v1alpha1.BareMetalHost) error {
+	host.Status.LastSoftPowerOff = nil
+	return r.client.Status().Update(context.TODO(), host)
+}
+
+// Check if Soft Power Off takes longer than timeout treshold, for any reason.
+func (r *ReconcileBareMetalHost) softPowerOffTimeoutNotReached(host *metal3v1alpha1.BareMetalHost) bool {
```
dhellmann: Won't ironic manage the timeout for us?

Xenwar: It does. But, we need to cover the case where ironic does not move to hard power off, for any reason. 
Basically, trying to enable the control to make the final say(regardless of ironic)

Xenwar:
> How does this work relate to #294?

Em, there seems to be a duplicate work. I think we can go with #294.

hardys:
> Em, there seems to be a duplicate work. I think we can go with #294.

Closing per this comment

PR #307, opened 2019-09-24
Set error message on inspection error 

zaneb:
When an error occurred during inspection, we weren't setting the Requeue
flag, which is required for the event to actually be published and the
CR to be updated (although it doesn't actually cause the reconcile to be
requeued, because of the error).

Had we published the event, it would have wrongly stated that it was a
registration error.

Finally, we weren't actually changing the Host's ProvisionState to
RegistrationError (since changes weren't written), so don't start now.

### DIFF
--- a/pkg/controller/baremetalhost/baremetalhost_controller.go
+++ b/pkg/controller/baremetalhost/baremetalhost_controller.go
@@ -516,9 +516,10 @@ func (r *ReconcileBareMetalHost) actionInspecting(prov provisioner.Provisioner, 
 	}
 
 	if provResult.ErrorMessage != "" {
-		info.host.Status.Provisioning.State = metal3v1alpha1.StateRegistrationError
-		info.host.SetErrorMessage(provResult.ErrorMessage)
-		info.publishEvent("RegistrationError", provResult.ErrorMessage)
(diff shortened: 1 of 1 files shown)


pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -516,9 +516,10 @@ func (r *ReconcileBareMetalHost) actionInspecting(prov provisioner.Provisioner,
 	}
 
 	if provResult.ErrorMessage != "" {
-		info.host.Status.Provisioning.State = metal3v1alpha1.StateRegistrationError
```
russellb: I know you mention it in the commit message, but I wonder why not keep this in place?

zaneb: It's the wrong state anyway (i.e. the error is not due to a registration problem), and our ultimate goal is to get rid of these error states altogether (in #285), so introducing a new place where we switch to this state (which this effectively would be, since the code did nothing before) makes things worse and our future plans more complicated.

zaneb:
Actually it turns out that events are always published regardless of whether we update the CR. I'll update the commit message to be less misleading.

PR #308, opened 2019-09-26
doc: add more literature to api doc

tripledes:
After our chat over email (@hardys) , I'm sending this PR that adds:

* diagram in SVG for future modifications
* diagram in PNG for linking
* Markdown document with the explanation and showing the diagram

I could also provide the original Google Drawing if that's preferred over the SVG.

Let me know if anything would need amending or changed completely :nerd_face: 

**UPDATE**:
* Tried to merge both api.md and my previous addition
* Add missing or erroneous fields
* Add a complete example of the BareMetalHost resource

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -2,202 +2,266 @@
 
 ## BareMetalHost
 
-The BareMetalHost resource defines the properties of a physical host
-necessary to manage and provision it.
+**Metal³** introduces the concept of **BareMetalHost** resource, which
+defines a physical host and its properties. The **BareMetalHost** embeds
+two well differentiated sections, the bare metal host specification
+and its current status.
 
-### Spec Fields
+### BareMetalHost spec
 
-*bmc* -- The connection information for the BMC controller on the host.
+The *BareMetalHost's* *spec* defines the desire state of the host. It contains
+mainly, but not only, provisioning details.
 
-*bmc.address* -- The URL for communicating with the BMC controller. When
-communicating over IPMI, this should be a URL of the form
-`ipmi://<host>:<port>` (an unadorned `<host>:<port>` is also accepted).
-Specifying the port is optional; the default port is 623. Dell iDRAC is also
-supported, by using the scheme `idrac://` (or `idrac+http://` to disable TLS)
-in place of `https://` in the iDRAC URL; only the hostname or IP address is
-required - `idrac://host.example` is equivalent to
-`idrac+https://host.example:443/wsman`. Fujitsu iRMC is also supported,
-by using the scheme `irmc://<host>:<port>` with `<port>` is optional.
-Redfish is also supported, by using the scheme `redfish://` (or
-`redfish+http://` to disable TLS) in place of `https://` in the Redfish URL;
-the hostname or IP address, and the path to the system ID are required -
-for example `redfish://myhost.example/redfish/v1/Systems/MySystemExample`
+#### Spec fields
 
-*bmc.credentials* -- A reference to a Secret containing the connection
-data, at least username and password, for the BMC.
+* *bmc* -- The connection information for the BMC (Baseboard Management
+  Controller) on the host.
+  * *address* -- The URL for communicating with the BMC controller, based
+    on the provider being used, the URL will look as follows:
+    * IPMI
+      * `ipmi://<host>:<port>`, an unadorned `<host>:<port>` is also accepted
+        and the port is optional, if using the default one (623).
+    * Dell iDRAC
+      * `idrac://` (or `idrac+http://` to disable TLS).
+    * Fujitsu iRMC
+      * `irmc://<host>:<port>`, where `<port>` is optional if using the default.
+    * Redfish
+      * `redfish://` (or `redfish+http://` to disable TLS), the hostname
+        or IP address, and the path to the system ID are required,
+        for example `redfish://myhost.example/redfish/v1/Systems/MySystemExample`
 
-*online* -- A boolean indicating whether the host should be powered on
-(true) or off (false). Changing this value will trigger a change in
-power state on the physical host.
+  * *credentialsName* -- A reference to a *secret* containing the
+    username and password for the BMC.
 
-*consumerRef* -- A reference to another object that is using the
-host. For example, a Machine when the host is being used by the
-machine-api.
+* *online* -- A boolean indicating whether the host should be powered on
+  (true) or off (false). Changing this value will trigger a change in
+  power state on the physical host.
 
-*externallyProvisioned* -- A boolean indicating whether something else
-is managing the image running on the host. When set, if no image is
-provided, the host's power status and hardware inventory will be
-monitored. If this flag is set, no provisioning or deprovisioning
-operations are performed for the host.
+* *consumerRef* -- A reference to another resource that is using the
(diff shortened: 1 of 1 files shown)


nordixinfra:

zaneb:
It would be good if you could add something to the Makefile for automatically converting the SVG to the PNG.

TBH I'm not sure I see the value of the diagram, since it is not generated from the code it will quickly become out of date. The docs look good, although a lot of the information seems duplicated by https://github.com/metal3-io/baremetal-operator/blob/master/docs/api.md

tripledes:
@zaneb my original intention making the diagram and the doc wasn't really submitting a PR but as mentioned, @hardys suggested to send it over. I completely agree this should be kept up to date as it will stale quickly and if it's already (almost) covered makes no sense to duplicate content. Please feel free to drop the PR :+1:

russellb:
The current api.md doc has a lot of field reference info, but not so much of the general explanation of the objects.  Can you look at how you might integrate your new content into that file?

As for the diagram, it's nice, but I'd agree with the concern about keeping it up to date.  Maybe split that out into a new PR and we can think about ways to auto generate something similar?  or just drop it for now ...

tripledes:
@russellb makes sense, I think I'll drop the diagram for now and give it a thought on how to auto-generate it. I'll see how to integrate the doc on `api.md` and update the PR, thanks!

tripledes:
@zaneb yeah, I'm still polishing it a bit, I'll remove the WIP (probably today by the EOD) and ping you for review. Thanks!

tripledes:
Hi! I tried to merge both but respecting `api.md` to the best of my knowledge on the language and the code :)

Let me know if you find anything missing (I'm aware I skipped few things, hopefully not that important) or anything wrong....I'll be more than happy to fix/add/remove anything.

zaneb:
One issue with this is that you've changed the names from the ones that appear in the API (which start with lower case) to field names or sometimes types (e.g. BareMetalHostStatus) that appear only in the Golang code. Since this is the API document I think we should continue to use the names as they appear in JSON.
The extra details and hierarchy are fantastic though, this is a lot more readable now that the API has grown quite a bit.

tripledes:
@zaneb makes sense, I understood the API as the package API, but I get your point, let me fix it and re-push it.

tripledes:
@zaneb I reverted all the names and changed the references to `BareMetalHostSpec` and `BareMetalHostStatus`, let me know if that'd be enough. I used the json annotations on the types file to make the modifications.

docs/api.md
```diff
@@ -2,202 +2,266 @@
...
+      * `idrac://` (or `idrac+http://` to disable TLS).
+    * Fujitsu iRMC
+      * `irmc://<host>:<port>`, where `<port>` is optional if using the default.
+    * Redfish
+      * `redfish://` (or `redfish+http://` to disable TLS), the hostname
+        or IP address, and the path to the system ID are required,
+        for example `redfish://myhost.example/redfish/v1/Systems/MySystemExample`
+
+  * *credentialsName* -- A reference to a *secret* containing the connection
+    details, at least username and password, for the BMC.
```
```suggestion
    username and password for the BMC.
```

We only want the username and password, so saying "at least" is confusing here.

docs/api.md
```diff
@@ -2,202 +2,266 @@
...
+
+* *online* -- A boolean indicating whether the host should be powered on
+  (true) or off (false). Changing this value will trigger a change in
+  power state on the physical host.
+
+* *consumerRef* -- A reference to another resource that is using the
+  host, it could be empty if the host is not being currently used.
+  For example, a *Machine* resource when the host is being used by the
+  [*machine-api*](https://github.com/openshift/machine-api-operator).
```
```suggestion
  [*machine-api*](https://github.com/kubernetes-sigs/cluster-api).
```

We should link to the upstream definition of the machine-api.

docs/api.md
```diff
@@ -2,202 +2,266 @@
...
+
+* *externallyProvisioned* -- A boolean indicating whether the host
+  provisioning and deprovisioning are managed externally. When set,
+  if no image is provided, the host's power status and hardware
```
```suggestion
  the host's power status and hardware
```

It looks like I missed a doc update when changing the way this flag works. If it is set, the image isn't checked at all.

docs/api.md
```diff
@@ -2,202 +2,266 @@
...
+  inventory will be monitored but no provisioning or deprovisioning
+  operations are performed on the host.
+
+* *image* -- Holds details for the image to be deployed on a given host.
+  * *url* -- The URL of an image to deploy to the host.
+  * *checksum* -- The actual checksum or a URL to a file containing
+    the checksum for the image at *image.url*.
+
+* *userData* -- A reference to the Secret containing the user provided data
```
```suggestion
* *userData* -- A reference to the Secret containing the cloudinit user data
```

It's not "user provided data" it's the user data value for cloudinit.

docs/api.md
```diff
@@ -2,202 +2,266 @@
...
+### BareMetalHost status
+
+Moving onto the next block, the *BareMetalHost's* *status* which represents
+the host's current state. Including tested credentials, current hardware
+details, etc.
+
+#### Status fields
+
+* *goodCredentials* -- A reference to the secret and its namespace
+  holding the last set of credentials the system was able to validate
```
```suggestion
  holding the last set of BMC credentials the system was able to validate
```

docs/api.md
```diff
@@ -2,202 +2,266 @@
...
+  as working.
+
+* *triedCredentials* -- A reference to the secret and its namespace
+  holding the last set of credentials that were sent to
```
```suggestion
  holding the last set of BMC credentials that were sent to
```

docs/api.md
```diff
@@ -2,202 +2,266 @@
...
+
+* *lastUpdated* -- The timestamp of the last time the status of the
+  host was updated.
+
+* *operationalStatus* -- The status of the server. Value is one of the
+  following:
+  * *OK* -- Indicates all the details for the host are known and working,
+    meaning the host is correctly configured and manageable.
+  * *discovered* -- Implies some of the host's details are either
+    not working correctly missing. For exmample, the BMC address is known
```
```suggestion
    not working correctly or missing. For example, the BMC address is known
```

docs/api.md
```diff
@@ -2,202 +2,266 @@
...
+  consumerRef:
+    apiVersion: machine.openshift.io/v1beta1
+    kind: Machine
+    name: bmo-master-0
+    namespace: bmo-project
+  externallyProvisioned: true
+  hardwareProfile: default
+  image:
+    checksum: http://172.16.1.100/images/myOS.qcow2.md5sum
+    url: http://172.16.1.100/images/myOSv1/myOS.qcow2
```
dhellmann: The paths in these values don't match. Should they?

docs/api.md
tripledes: @dhellmann I used a kinda generic term cause I assumed it could be other technologies, like for instance ignition, is not that the case?

docs/api.md
tripledes: I guess they don't have to, but it was my intention that they'd match, fixing it, thanks!

tripledes:
@dhellmann I made all the modifications. Let me know if there is anything else!

docs/api.md
dhellmann: For RHCOS the data is consumed by Ignition, but it's passed in cloudinit format.

PR #312, opened 2019-10-03
Check namespace when matching Secret references

zaneb:
Previously when checking if a Secret has changed we could have confused
two Secrets with the same name in different namespaces.


russellb:
Hopefully RBAC would have prevented looking at secrets in other namespaces, but this is a good improvement in any case!

/lgtm

PR #316, opened 2019-10-06
Get make generate to regenerate the yaml file.

jbrette:
- New version of controller-gen aborts instead of silently ignoring
  errors. This gets both controller-gen and operator-sdk to fail
  generation of yaml
- Name of the crd.yaml has changed. New version of operator-sdk does
  not detect it needs to regenerate the crd.yaml
- demo-host.yaml is conflicting with operator-sdk because was saved
  in the deploy/crds folder.
- float64 are not supported according to latest controller-gen code.

Changes have been tested with latest version of controller-gen and operator-sdk


jbrette:
Can't figure out how to get newer version of operator-sdk to build from travis. Closing the PR.

PR #317, opened 2019-10-07
Add disk/by_path

redixin:
Add disk/by_path (e.g. e.g. /dev/disk/by-path/pci-0000:00:07.0) to available introspection results.

### DIFF
--- a/pkg/apis/metal3/v1alpha1/baremetalhost_types.go
+++ b/pkg/apis/metal3/v1alpha1/baremetalhost_types.go
@@ -205,6 +205,9 @@ type Storage struct {
 	// A name for the disk, e.g. "disk 1 (boot)"
 	Name string `json:"name"`
 
+	// Disk path, e.g. /dev/disk/by-path/pci-0000:00:07.0
+	ByPath string `json:"by_path"`
+
 	// Whether this disk represents rotational storage
 	Rotational bool `json:"rotational"`
 
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -202,7 +202,7 @@
   version = "v0.2.0"
 
 [[projects]]
-  digest = "1:f60e7e9be98ff9d00371b24fcbe92e019b809b592d7d10b4cb61965ba59610f6"
+  digest = "1:f0b8f85d41e3633aab3a2b6082fb9152fe0bc87f63c4a96d8dbf1254cdd10744"
   name = "github.com/gophercloud/gophercloud"
   packages = [
     ".",
@@ -214,7 +214,7 @@
     "pagination",
   ]
   pruneopts = "NT"
-  revision = "2d40ff7a3874e2b376ff245d549e7757048cd65c"
+  revision = "0ef034d8b463562a083e72c2a6627e8d0debd936"
 
 [[projects]]
   branch = "master"
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -47,7 +47,7 @@ required = [
 
 [[constraint]]
   name = "github.com/gophercloud/gophercloud"
-  revision = "2d40ff7a3874e2b376ff245d549e7757048cd65c"
+  revision = "0ef034d8b463562a083e72c2a6627e8d0debd936"
 
 [[constraint]]
   name = "github.com/operator-framework/operator-sdk"
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -502,6 +502,7 @@ func getStorageDetails(diskdata []introspection.RootDiskType) []metal3v1alpha1.S
 	for i, disk := range diskdata {
 		storage[i] = metal3v1alpha1.Storage{
 			Name:               disk.Name,
+			ByPath:             disk.ByPath,
 			Rotational:         disk.Rotational,
 			SizeBytes:          metal3v1alpha1.Capacity(disk.Size),
 			Vendor:             disk.Vendor,
--- a/vendor/github.com/gophercloud/gophercloud/acceptance/clients/conditions.go
+++ b/vendor/github.com/gophercloud/gophercloud/acceptance/clients/conditions.go
@@ -19,6 +19,14 @@ func RequireNonAdmin(t *testing.T) {
 	}
 }
 
+// RequirePortForwarding will restrict a test to only be run in environments
+// that support port forwarding
+func RequirePortForwarding(t *testing.T) {
+	if os.Getenv("OS_PORTFORWARDING_ENVIRONMENT") == "" {
+		t.Skip("this test requires support for port forwarding")
+	}
+}
+
 // RequireDNS will restrict a test to only be run in environments
 // that support DNSaaS.
 func RequireDNS(t *testing.T) {
--- a/vendor/github.com/gophercloud/gophercloud/acceptance/openstack/networking/v2/extensions/layer3/layer3.go
+++ b/vendor/github.com/gophercloud/gophercloud/acceptance/openstack/networking/v2/extensions/layer3/layer3.go
@@ -9,6 +9,8 @@ import (
 	"github.com/gophercloud/gophercloud/acceptance/clients"
 	"github.com/gophercloud/gophercloud/acceptance/tools"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/layer3/floatingips"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/layer3/portforwarding"
+
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/layer3/routers"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/ports"
 	th "github.com/gophercloud/gophercloud/testhelper"
@@ -64,6 +66,46 @@ func CreateFloatingIPWithFixedIP(t *testing.T, client *gophercloud.ServiceClient
 	return floatingIP, err
 }
 
+// CreatePortForwarding creates a port forwarding for a given floating IP
+// and port. An error will be returned if the creation failed.
+func CreatePortForwarding(t *testing.T, client *gophercloud.ServiceClient, fipID string, portID string, portFixedIPs []ports.IP) (*portforwarding.PortForwarding, error) {
+	t.Logf("Attempting to create Port forwarding for floating IP with ID: %s", fipID)
+
+	fixedIP := portFixedIPs[0]
+	internalIP := fixedIP.IPAddress
+	createOpts := &portforwarding.CreateOpts{
+		Protocol:          "tcp",
+		InternalPort:      25,
+		ExternalPort:      2230,
+		InternalIPAddress: internalIP,
+		InternalPortID:    portID,
+	}
+
+	pf, err := portforwarding.Create(client, fipID, createOpts).Extract()
+	if err != nil {
+		return pf, err
+	}
+
+	t.Logf("Created Port Forwarding.")
+
+	th.AssertEquals(t, pf.Protocol, "tcp")
+
+	return pf, err
+}
+
+// DeletePortForwarding deletes a Port Forwarding with a given ID and a given floating IP ID.
+// A fatal error is returned if the deletion fails. Works best as a deferred function
+func DeletePortForwarding(t *testing.T, client *gophercloud.ServiceClient, fipID string, pfID string) {
+	t.Logf("Attempting to delete the port forwarding with ID %s for floating IP with ID %s", pfID, fipID)
+
+	err := portforwarding.Delete(client, fipID, pfID).ExtractErr()
+	if err != nil {
+		t.Fatalf("Failed to delete Port forwarding with ID %s for floating IP with ID %s", pfID, fipID)
+	}
+	t.Logf("Successfully deleted the port forwarding with ID %s for floating IP with ID %s", pfID, fipID)
+
+}
+
 // CreateExternalRouter creates a router on the external network. This requires
 // the OS_EXTGW_ID environment variable to be set. An error is returned if the
 // creation failed.
@@ -80,10 +122,8 @@ func CreateExternalRouter(t *testing.T, client *gophercloud.ServiceClient) (*rou
 	t.Logf("Attempting to create external router: %s", routerName)
 
 	adminStateUp := true
-	enableSNAT := false
 	gatewayInfo := routers.GatewayInfo{
-		NetworkID:  choices.ExternalNetworkID,
-		EnableSNAT: &enableSNAT,
+		NetworkID: choices.ExternalNetworkID,
 	}
 
 	createOpts := routers.CreateOpts{
--- a/vendor/github.com/gophercloud/gophercloud/openstack/baremetalintrospection/v1/introspection/results.go
+++ b/vendor/github.com/gophercloud/gophercloud/openstack/baremetalintrospection/v1/introspection/results.go
@@ -234,6 +234,7 @@ type RootDiskType struct {
 	Hctl               string `json:"hctl"`
 	Model              string `json:"model"`
 	Name               string `json:"name"`
+	ByPath             string `json:"by_path"`
 	Rotational         bool   `json:"rotational"`
 	Serial             string `json:"serial"`
 	Size               int    `json:"size"`
--- a/vendor/github.com/gophercloud/gophercloud/openstack/blockstorage/v1/volumes/requests.go
+++ b/vendor/github.com/gophercloud/gophercloud/openstack/blockstorage/v1/volumes/requests.go
@@ -42,7 +42,7 @@ func Create(client *gophercloud.ServiceClient, opts CreateOptsBuilder) (r Create
 		return
 	}
 	_, r.Err = client.Post(createURL(client), b, &r.Body, &gophercloud.RequestOpts{
-		OkCodes: []int{200, 201},
+		OkCodes: []int{200, 201, 202},
 	})
 	return
 }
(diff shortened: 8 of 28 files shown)


pkg/apis/metal3/v1alpha1/baremetalhost_types.go
```diff
@@ -205,6 +205,9 @@ type Storage struct {
 	// A name for the disk, e.g. "disk 1 (boot)"
 	Name string `json:"name"`
 
+	// Disk path, e.g. /dev/disk/by-path/pci-0000:00:07.0
+	ByPath string `json:"by_path"`
```
dhellmann: If this is always the linux device path, maybe the name should be DevicePath? Or even just Path?

I'm interested in hearing from other reviewers about the name before you change it, though.

zaneb: IIUC the "by" part in "by_path" means it is referring to the PCI path. So what it's trying to convey is that this is a stable name that locates the disk by its physical connection to the PCI bus. Not sure how to convey that in the name, but I agree that "by_path" is not really doing it.

dhellmann: Can that path be constructed from a PCI identifier?

How persistent/canonical is the value? Is it the same for a given device for all flavors of Linux? Does it depend in any way on the kernel version or what order modules are loaded or something like that?

zaneb: They're symlinks set up by udev, so I guess in theory you could mess it up by writing custom udev rules. In practice I'd don't imagine that anyone would.
Note that this refers to a particular location in the PCI tree - so if you pull a disk and replace it then the same link would now point to a different disk. If that's not what you want then you should use e.g. `/dev/disk/by-uuid/` instead. But that's kind of the point - when you want to reference disks by where they're physically attached to the controller, this is the device you use to do it.
This is the best reference I've found: https://wiki.archlinux.org/index.php/Persistent_block_device_naming#by-id_and_by-path

redixin: @dhellmann It is called `disk/by_path` in linux `dev` directory, it called `by_path` in `ipa` introspection results, it called `ByPath` in `gophercloud` why it should be named differently in `metal3`?

@dhellmann this is just another introspection result that is refers to a particular location in the PCI tree. It is useful for those who want to see this information in introspection results.

dhellmann: > @dhellmann It is called `disk/by_path` in linux `dev` directory, it called `by_path` in `ipa` introspection results, it called `ByPath` in `gophercloud` why it should be named differently in `metal3`?
> 
> @dhellmann this is just another introspection result that is refers to a particular location in the PCI tree. It is useful for those who want to see this information in introspection results.

I don't find the name ByPath expressive at all. In the other contexts you mention, there is more information available. The /dev/.../by_path/... name for example is clearly a path on the filesystem, but even there it isn't clear what the "path" in the name really means unless you know there is also a /dev/.../by-uuid/... pointing to the same device.

If the value is the PCI path, then we could call it something that expresses that.

If the value can be constructed from other data, such as the PCI device ID, then maybe what this API should expose is just that other data.

What sort of value would be relevant if the image being deployed on the host was Windows instead of Linux?

redixin: @dhellmann this data is part of introspection results, so the user must be familiar with ironic introspection results. There is nothing to do with this data in `metal3`.

> What sort of value would be relevant if the image being deployed on the host was Windows instead of Linux?

It depends on ironic python agent windows implementation.

dhellmann: > @dhellmann this data is part of introspection results, so the user must be familiar with ironic introspection results. There is nothing to do with this data in `metal3`.
> 
> > What sort of value would be relevant if the image being deployed on the host was Windows instead of Linux?
> 
> It depends on ironic python agent windows implementation.

That's unlikely to ever exist, but it shouldn't need to.

If I'm a Windows admin looking at this discovered data, am I going to understand the path value currently being reported? Is there some other representation for the information that is not platform-specific?

redixin: @dhellmann

>Is there some other representation for the information that is not platform-specific?

Ironic python agent reports only this one.

redixin: @dhellmann I understand that `ipa` is implementation detail, but in fact only linux/ironic/ipa will be ever used, and everybody know what `/dev/by_path/...` is.

dhellmann: > @dhellmann I understand that `ipa` is implementation detail, but in fact only linux/ironic/ipa will be ever used, and everybody know what `/dev/by_path/...` is.

It is not safe to assume that IPA and Ironic will be the only backend. From the very beginning I considered designs that worked differently, and for which I would not expect to use Ironic.

Regardless of whether we do change, though, the metal3 API should not be tied to the Ironic API. A metal3 user should not have to learn how Ironic's API works in order to understand metal3.

zaneb: > Ironic python agent reports only this one.

You could easily extract the relevant data and report it as e.g.:

    pciLocation: 0000:00:07.0

and the information would be self-explanatory, usable by people wanting to provision any operating system, and independent of the implementation details of Ironic.

I don't understand why you would rather give up than make such an obviously-beneficial change in presentation.

s3rj1k: @zaneb Acrualy no, not in every case `by-path` value can be reconstructed from PCI device location.

Please look at this two articles:
- https://utcc.utoronto.ca/~cks/space/blog/linux/LinuxDiskNamesChange
- https://unix.stackexchange.com/questions/86764/understanding-dev-disk-by-folders

In recent distros with systemd `by-path` is generated by https://cgit.freedesktop.org/systemd/systemd/tree/src/udev/udev-builtin-path_id.c.

Adding `by-path` in some manner to the codebase would be nice.

Can we continue this discussion?

zaneb: > @zaneb Acrualy no, not in every case `by-path` value can be reconstructed from PCI device location.
> 
> Please look at this two articles:
> 
>     * https://utcc.utoronto.ca/~cks/space/blog/linux/LinuxDiskNamesChange
>     * https://unix.stackexchange.com/questions/86764/understanding-dev-disk-by-folders

I don't understand the point you're trying to make here?

But I think the second link is a great illustration of the reaction of the average person encountering the name `by-path` for the first time:

> * by-path → **?!**
  
> In recent distros with systemd `by-path` is generated by https://cgit.freedesktop.org/systemd/systemd/tree/src/udev/udev-builtin-path_id.c.

https://github.com/metal3-io/baremetal-operator/pull/317#discussion_r332800589

> Adding `by-path` in some manner to the codebase would be nice.

I agree. I don't know why the author of this PR closed it.

> Can we continue this discussion?

I would love to continue, it wasn't my idea to stop :)

s3rj1k: As I see it there are 2 questions that need to answered:

1. Name of the field.
2. Value of the filed.

The most concerning to me is the actual value and the way how it could be reconstructed to original value from generic representation.

I propose we start with the second question and then solved, first one would solve itself automatically.

zaneb: Maybe we should start by asking what this is used for? We can't use it as a root device hint, so it's not like it has to be round-trippable.

Anyway, in my suggestion above you would reconstruct it by prepending `/dev/disk/by-path/pci-`.

s3rj1k: Ok, so the actual value would be for example '0000:00:17.0-ata-2' and reconstructed would be '/dev/disk/by-path/pci-0000:00:17.0-ata-2'. And this is only correct for Linux platform.

As per link above (and other GPT friends):

- by-id
- by-label
- by-uuid
- ...

describe the actual physical storage device or partition.

On the other hand `by-path` describes not the actual physical/logical ID of a device but a place in a 'Server Platform' so to say.

By using `by-path` we actually do not care about a drive and do care of it's location.

Consider use case where you have lots of drives in 3U storage rack and some of this drive are used for system partitions in slots 0-3 for example.

You actually do not care what kind of drive is in slot, only that those specific slots are used for root partition.

When comes time to replace faulty drive you can just swap it without worrying that ID/UUID of a drive will change.

zaneb: I think I understand now. Previously I had assumed that the main use of this was for determining when a disk has been hot-swapped, or for tracking down the physical location of a particular disk.

I think what you're saying is you would pass the full `/dev/disk/by-path/...` string as the Name [root device hint](https://docs.openstack.org/ironic/latest/install/advanced.html#specifying-the-disk-for-deployment-root-device-hints) instead of something like `/dev/sda` (which is basically always a bad idea since it's not stable). So for that you want the whole path, and I'd suggest a name that implies it can be used in the Name field of the root device hints (maybe something like `LocationName`?).

Presumably there are circumstances where it's necessary to do it this way rather than using HCTL (which is already supported both here and in root device hints)? (I'm pretty fuzzy on what does and does not count as a SCSI device these days.)

s3rj1k: I was thinking on some generic field name like DeviceLocation or Location (committing Device prefix as we know that this is a device). This also can be LocationName of course.

This field also can contain any `/dev/disk/by-*` (UUID, Label, ...) to be more generic, `/dev/disk/by-path/` is just a specific use case for this.

This field will complement HCTL for some cases when HCTL is not enough to identify a device.

So how do we do this, new PR?

zaneb: > This field will complement HCTL for some cases when HCTL is not enough to identify a device.

If you could give an example of such cases, that would help everybody I think.

> So how do we do this, new PR?

+1. In theory since this is an API change we should do a design doc in the metal3-docs repo as well. As we've seen, none of the reasons behind this are obvious to someone who hasn't encountered a situation where they need it.

s3rj1k: One example would be NVMe disks.

from `lsscsi -h`
```
    --wwn|-w          output WWN for disks (from /dev/disk/by-id/wwn*)
    <h:c:t:l>         filter output list (def: '*:*:*:*' (all)). Meaning:
                      <host_num:controller:target:lun> or for NVMe:
                      <'N':ctl_num:cntlid:namespace_id>
```

One particular case with 2 NVMe disks:

```
lsscsi -w
[N:0:0:1]    disk    nvme.8086-42545046383135343033485634353052474e-494e54454c205353445045324d583435304737-00000001  /dev/nvme0n1
[N:1:0:1]    disk    nvme.8086-42545046383234353030364734353052474e-494e54454c205353445045324d583435304737-00000001  /dev/nvme1n1
```

```
lsblk -o NAME,HCTL,TRAN,MODEL
NAME        HCTL TRAN   MODEL
nvme1n1          nvme   INTEL SSDPE2MX450G7
├─nvme1n1p1      nvme   
├─nvme1n1p2      nvme   
│ └─md127               
└─nvme1n1p3      nvme   
nvme0n1          nvme   INTEL SSDPE2MX450G7
├─nvme0n1p1      nvme   
├─nvme0n1p2      nvme   
│ └─md127               
├─nvme0n1p3      nvme   
└─nvme0n1p4      nvme
```

Notice that lsblk does not have HCTL, that is because for NVMe disks HCTL is not the same thing as for regular rotational disks.

Also [here](https://ipfs.io/ipfs/QmdA5WkDNALetBn4iFeSepHjdLGJdxPBwZyY47ir1bZGAK/comp/linux/scsi_ids.html) is a bit of ranting from Linus on this matter.

I feel that `/dev/disk/by-uuid/` and `/dev/disk/by-path/` are far more stable and generic way of specifying a device.

As discussed above, the actual field can contain any type of location description (/dev/disk/by-*)

s3rj1k: And another real-world example, this are same platforms that have similar components install, only difference is serial numbers and kernel patch level version. Disks are connected directly to backplain (no raid devices).

```
a5) lsblk -o NAME,HCTL -r
    NAME HCTL
    sda 6:0:0:0
    sdb 6:0:1:0
    sdc 6:0:2:0
    sdd 6:0:3:0
    sde 7:0:0:0
    sdf 7:0:1:0
```
```
a9) lsblk -o NAME,HCTL -r
    NAME HCTL
    sda 0:0:0:0
    sdb 1:0:0:0
    sdc 2:0:0:0
    sdd 3:0:0:0
    sde 4:0:0:0
    sdf 5:0:0:0
```

zaneb: Thanks, that's really helpful. I also have a PR (#657) to correct the egregiously wrong description of the `Name` field, which I think was contributing to the confusion.

s3rj1k: @zaneb Yea, #657 helps, but those name fields are not reboot stable sadly :(

dhellmann: Does Ironic support specifying the device in this way? I don't see anything about by-path in [the documentation](https://docs.openstack.org/ironic/latest/install/advanced.html#specifying-the-disk-for-deployment-root-device-hints).

s3rj1k: @dhellmann Well, this is quite interesting trip into python world, here it goes:

- Class [RootDiskSelectionHook](https://github.com/openstack/ironic-inspector/blob/master/ironic_inspector/plugins/standard.py#L33 ) in ironic_inspector is in charge of getting root device from provided hints.
-  RootDiskSelectionHook calls to [match_root_device_hints](https://github.com/openstack/ironic-lib/blob/master/ironic_lib/utils.py#L444) function in ironic library.
- match_root_device_hints calls to [find_devices_by_hints](https://github.com/openstack/ironic-lib/blob/master/ironic_lib/utils.py#L349)

As you can see, in comments on two last functions we have:
```
by_path: (String): The alternative device name,
                  e.g. /dev/disk/by-path/pci-0000:00
```

After grepping through ironic-inspector code again, looking for what [inventory['disks']](https://github.com/openstack/ironic-inspector/blob/master/ironic_inspector/plugins/standard.py#L50) contains I stumbled upon [this article](https://docs.openstack.org/ironic-python-agent/latest/admin/how_it_works.html#hardware-inventory).

As can be seen from that inventory description for disk:
```
disks

    list of disk block devices with fields: name, model, size (in bytes), rotational (boolean), wwn, serial, vendor, wwn_with_extension, wwn_vendor_extension, hctl and by_path (the full disk path, in the form /dev/disk/by-path/<rest-of-path>).
```

```by_path``` field is actually present.

zaneb: > Does Ironic support specifying the device in this way? I don't see anything about by-path in [the documentation](https://docs.openstack.org/ironic/latest/install/advanced.html#specifying-the-disk-for-deployment-root-device-hints).

Yeah, this is the part it took me a while to get: the `name` field in the hints is just a device file. So it could be `/dev/md0` as in the example there, but it could equally be `/dev/disk/by-path/pci-0000:00`.

s3rj1k: @zaneb Well actually Ironic exposes `by_path` separatly from `name`. So two fields can be present simultaneously.

The actual function that discovers drive hints is [here](https://github.com/openstack/ironic-python-agent/blob/master/ironic_python_agent/hardware.py#L365).

I think that OpenStack guys just forgat to add `by_path` description to docs in [here](https://docs.openstack.org/ironic/latest/install/advanced.html#specifying-the-disk-for-deployment-root-device-hints) but this description is present in [here](https://docs.openstack.org/ironic-python-agent/latest/admin/how_it_works.html#hardware-inventory).

s3rj1k: Should we make `Name` field a string slice?

zaneb: Oh, thanks, I _still_ had the wrong end of the stick because I had been relying on that same doc that Doug linked. I submitted a patch to fix it: https://review.opendev.org/758481
So this changes things in terms of naming. We need to add a field both here and in the root device hints, and the names should match. Maybe `DeviceByPath`? (I wish we'd called the `Name` field something like `DeviceName`, but that ship has sailed.)

s3rj1k: `DeviceByPath` sounds nice considering that someone could actually add a patch to Ironic that enables support for UUID.

dhellmann: We can do API changes now! We can fix things!

Does gophercloud expose the by_path field, or is that another task to put on the implementation list?

s3rj1k: @dhellmann https://github.com/gophercloud/gophercloud/blob/master/openstack/baremetalintrospection/v1/introspection/results.go#L238

dhellmann: I guess there's not a special type for passing the root device hints, is there? That goes in a capability or something, so it's just a generic JSON blob.

s3rj1k: `The inventory is a dictionary (JSON object)` that struct from gophercloud describes `disk` part of the IPA inventory.

zaneb: It's [just JSON](https://github.com/metal3-io/baremetal-operator/blob/master/pkg/provisioner/ironic/devicehints/devicehints.go) to pass the root device hints, so thankfully nothing else to do in gophercloud.

PR #319, opened 2019-10-10
Configure RAID Interface for iRMC driver 

longkb:
Currently, RAID interface is not set when we create Ironic node(#320). This PR aims to make RAID interface to be set by vendor driver.

### DIFF
--- a/pkg/bmc/irmc.go
+++ b/pkg/bmc/irmc.go
@@ -68,7 +68,7 @@ func (a *iRMCAccessDetails) PowerInterface() string {
 }
 
 func (a *iRMCAccessDetails) RAIDInterface() string {
-	return ""
+	return "irmc"
 }
 
 func (a *iRMCAccessDetails) VendorInterface() string {


longkb:
Hi @dhellmann , @dtantsur 
This PR aims to add `bootInterface` into `accessDetail`, that needs for RAID configuratuon. Please help me to review it if you have a time :)

pkg/bmc/idrac.go
```diff
@@ -70,3 +70,7 @@ func (a *iDracAccessDetails) DriverInfo(bmcCreds Credentials) map[string]interfa
 func (a *iDracAccessDetails) BootInterface() string {
 	return "ipxe"
 }
+
+func (a *iDracAccessDetails) RAIDInterface() string {
+	return "no-raid"
```
zaneb: Is there a reason not to set it to "agent"?

RAID configuration is going to be added in a driver-independent way, so I'd have thought we'd want to ensure that all drivers can use it somehow.

longkb: > Is there a reason not to set it to "agent"?

I would like to set it as `no-raid` as default interface

> RAID configuration is going to be added in a driver-independent way, so I'd have thought we'd want to ensure that all drivers can use it somehow.

If vendor driver want to support RAID configuration, they can change `no-raid` to their driver

pkg/bmc/access.go
```diff
@@ -46,6 +46,9 @@ type AccessDetails interface {
 
 	// Boot interface to set
 	BootInterface() string
+
```
arkadykanevsky: This is fine for a single RAID. But in many cases you have 2 RAIDs.
One for 2 or small # of disks for Boot/Operating system.
Another one, including NORAID for remaining disks.

arkadykanevsky: This applies to the whole patch.
It will be too hard to retrofit later.

dhellmann: Do you have a specific proposal for a different API? As far as I can tell, the gopher cloud client for Ironic only accepts a single string. Are you proposing that it should take multiple values? Or are you proposing that the user needs to somehow specify the value?

arkadykanevsky: Doug, 
I am thinking that a variable length array will be sufficient.
First, parameter the size of array. Default 1.
The second is array of RAIDinterfaces. Each raid interface will be a tuple. First, part of tuple will be RAID type, second is list (or array) of disks in that RAID.

I recall that Ironic has a way to specify more than one RAID per node. Would be good to check how they specify it. Maybe Digmabar can comment on it.

dhellmann: This is an internal API, for statically defined types that let us pass different values to ironic based on the BMC type the user gives us. I think the value being returned here is telling ironic which RAID driver to use within itself (note the values returned by the methods below), rather than specifying individual controllers on a host. Are you saying we may need to tell ironic multiple values in that case?

pkg/bmc/idrac.go
dtantsur: Note that software RAID is vendor-independent.

longkb: > 
> 
> Note that software RAID is vendor-independent.

Yep. But sofware RAID is out of scope in this version of RAID configuration

zaneb: The default RAID interface for the iDrac driver in Ironic is `idrac-wsman` (unlike the iRMC driver, where the default is `no-raid`). So this change actually disables RAID support in iDrac where before it worked fine, while enabling it in iRMC where before it was disabled.

I think that for all of the non-iRMC drivers you should return an empty string so that it continues to use the driver default instead of overriding it to `no-raid` in every case.

pkg/bmc/access.go
zaneb: Yes, this is just telling the Ironic driver which RAIDInterface to use to configure RAID on this particular machine. So e.g. for the [iDRAC driver ](https://docs.openstack.org/ironic/latest/admin/drivers/idrac.html#enabling) the options are `idrac-wsman`/`idrac`, or `no-raid` to disable RAID support altogether (or `agent`, to configure it in software). Perhaps one day there'll be an `idrac-redfish` too. But this has nothing to do with what RAID arrays are to be created, that's in #292. All arrays are configured, or not configured, through the same interface.

pkg/bmc/idrac.go
longkb: Thanks @zaneb 
You're right.  I have just update my PR to set RAID interface of non-iRMC drivers to empty string as you said
Please keep going on :)

pkg/provisioner/ironic/ironic.go
```diff
@@ -248,6 +248,13 @@ func (p *ironicProvisioner) ValidateManagementAccess(credentialsChanged bool) (r
 		}
 		p.publisher("Registered", "Registered new host")
 
+		raidInterface := p.bmcAccess.RAIDInterface()
+		if raidInterface != "" {
```
zaneb: The type of the RAIDInterface member in nodes.CreateOpts is string, so it will be initialised to an empty string on lines 239-244 anyway. There's no harm in just passing

    RAIDInterface: p.bmcAccess.RAIDInterface(),

after line 241. You don't need to do anything special to check if the BMC plugin is returning an empty string.

longkb: Thanks @zaneb . You're right.
But it seems that these Interfaces already added by [1]
So I would like to configure the RAID interface to `irmc` in this PR :)

[1] https://github.com/metal3-io/baremetal-operator/commit/c8fdf0483ec30bd1fb9154ba8cc291e340d10f06#diff-c3f8fb66d64901a306b6faf6f7019c9f

pkg/bmc/access.go
digambar15: Yes, I agree with Arkady. IRonic support multiple Raid configuration for RAID. We should allow use to create multiple RAID which would be helpful for configuring root disk (say RAID1) and remaining disks (RAID 0 or non-raid) depending on the RAID controller. This will certainly good feature for us as Ironic is already supporting it.
@zaneb  Yes. You understanding is correct. Above implementation is irrespective of which driver you are using whether it's idrac-wsman/idrac/idrac-redfish.

zaneb: That's exactly what's proposed in the current PR: https://github.com/metal3-io/baremetal-operator/pull/292/files#diff-2336c3885f739a6a5c66bc71e72497b3R133-R140

PR #335, opened 2019-10-31
Add metal3 metadata for deployment from baremetalhost

maelk:
to be consumed by CAPBM to refer to the node

### DIFF
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -291,13 +291,10 @@ func (p *ironicProvisioner) ValidateManagementAccess(credentialsChanged bool) (r
 					Path:  "/instance_info/image_checksum",
 					Value: checksum,
 				},
-				// NOTE(dhellmann): We must fill in *some* value so that
-				// Ironic will monitor the host. We don't have a nova
-				// instance at all, so just give the node it's UUID again.
 				nodes.UpdateOperation{
 					Op:    nodes.ReplaceOp,
 					Path:  "/instance_uuid",
-					Value: p.host.Status.Provisioning.ID,
+					Value: string(p.host.ObjectMeta.UID),
 				},
 			}
 			_, err = nodes.Update(p.client, ironicNode.UUID, updates).Extract()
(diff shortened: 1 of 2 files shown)


zaneb:
Can you elaborate on why this is needed by CAPBM?
Ironic is supposed to be an implementation detail, so exposing this makes me hesitant though it's not automatically out of the question (if it's treated as an opaque string and only compared for equality that's not a disaster). I'm just not clear on why that's needed at all.
Another concern is that this treats the Ironic UUID as stable, but it's not - the Ironic database can disappear at any time and the data in it will get recreated with new IDs.

maelk:
@zaneb thank you for the input. here is the reasoning, but it is a work in progress, I am still trying to figure out the best solution to do this. In order to deploy properly, CAPI requires the providerID to be set on the target Kubernetes node by the cloud-provider of the target cluster. Aside of the fact that on baremetal, we currently do not have a cloud provider, we need to generate a provider id that will be unique for the vm, and that will be set on the vm from the metadata given by ironic in cloud-init, and by CAPBM as baremetalMachine.Spec.ProviderID. It is used only for comparison between the kubernetes node and the CAPI machines to identify which node is which machine. So it could be any random string as long as it is the same on both sides. Using the ironic node id is supposed to make it easier, when you are working with the target cluster, to know which ironic node it is.

My current implementation uses the instance id from ironic : https://github.com/metal3-io/cluster-api-provider-baremetal/pull/153 . This one changes everytime the node is deployed. We could use the ironic node id to have something more stable. However, as you pointed it, it does not survive a database clearing. A stable identifier would be the name, but we cannot guarantee its uniqueness. 

I anyways realized the the Status.Provisioning.ID contains the ironic node UUID when the node is in ready state, so for CAPBM we could get it this way. However we would still need to pass it through the metadata to the host. Maybe we should stick with the instance ID for now ?

dhellmann:
> @zaneb thank you for the input. here is the reasoning, but it is a work in progress, I am still trying to figure out the best solution to do this. In order to deploy properly, CAPI requires the providerID to be set on the target Kubernetes node by the cloud-provider of the target cluster. Aside of the fact that on baremetal, we currently do not have a cloud provider, we need to generate a provider id that will be unique for the vm, and that will be set on the vm from the metadata given by ironic in cloud-init, and by CAPBM as baremetalMachine.Spec.ProviderID. It is used only for comparison between the kubernetes node and the CAPI machines to identify which node is which machine. So it could be any random string as long as it is the same on both sides. Using the ironic node id is supposed to make it easier, when you are working with the target cluster, to know which ironic node it is.
> 
> My current implementation uses the instance id from ironic : [metal3-io/cluster-api-provider-baremetal#153](https://github.com/metal3-io/cluster-api-provider-baremetal/pull/153) . However, as you pointed it, it does not survive a database clearing. A stable identifier would be the name, but we cannot guarantee its uniqueness.

The namespace + name of the BareMetalHost will be unique.

maelk:
Hi @zaneb and @dhellmann Thank you for the feedback. Based on those, I changed it to just pass some info about the baremetal host as metadata, so that we can use them in the cloud-init templating to create the node-name and the providerID (done by CAPBM). How does this look for you ?

maelk:
Hi @zaneb @dhellmann ! Do you have any comments on this new commit ?

PR #338, opened 2019-11-08
Updates BMO dependencies and dependency management tools in Makefile

jan-est:
This PR upgrades BMO dependencies and makes needed changes to BMO to run properly after the upgrade. Following dependencies are upgrades: 

- k8s.io/api
- k8s.io/apiextensions-apiserver
- k8s.io/apimachinery
- k8s.io/client-go
- github.com/coreos/prometheus-operator
- sigs.k8s.io/controller-runtime
- github.com/operator-framework/operator-sdk

This PR also changes dep-check to dep-status command, adds dep-prune command and adds -verbose to dep command. Dep prune command has been used to remove unused packages from vendor folder. 

Previous content of the deploy/crds folder is moved to newly created examples folder.


pkg/apis/metal3/v1alpha1/zz_generated.openapi.go
```diff
@@ -1,4 +1,4 @@
-// +build !
+// +build !ignore_autogenerated
```
zaneb: Is this comment what is causing lint to start failing? Strange if it is... even stranger if not.

jan-est: I'm not sure. These zz_ files are autogenerated by operator-sdk. Now that my pull request is updating version of the operator-sdk something breaking the lint. I don't have any idea what might cause this lint to fail?

deploy/crds/metal3.io_baremetalhosts_crd.yaml
```diff
@@ -0,0 +1,470 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
maelk: This file is supposed to replace https://github.com/metal3-io/baremetal-operator/pull/338/files#diff-c3596490c667247182f3661aa8f5e84f (Github did not let me comment on that file) . I don't think we need to move it, as it is obsolete.

PR #340, opened 2019-11-12
Rename Ready state to Available

### LINKED ISSUE #315: Rename "Ready" state to "Available"
`Ready` is a status also used by the Node resource, but with a different meaning. This could cause unnecessary confusion. The GUI currently uses the term `Available` instead. By switching to this we could eliminate the confusion and ensure consistency between the graphical and command line interfaces.

jtomasek:
Ready is a status also used by the Node resource, but with a different meaning
which causes unnecessary confusion especially on the UI side where Host and Node
status is combined

Fixes #315 
~~Depends on https://github.com/metal3-io/metal3-dev-env/pull/137~~


jtomasek:
Related Console UI change https://github.com/openshift/console/pull/3348

maelk:
This would break the metal3-dev-env verifications. The verify script checks for "ready" state.

jtomasek:
> This would break the metal3-dev-env verifications. The verify script checks for "ready" state.

I've created a PR to accommodate the change in metal3-dev-env https://github.com/metal3-io/metal3-dev-env/pull/137

hardys:
@maelk I merged metal3-io/metal3-dev-env#137 - are you happy to see this go in now, or are there any other places which could be broken by this rename?

hardys:
@jtomasek I think there's another fix needed somewhere in the metal3-dev-env test, failure looks like:

```
FAIL - centos-0 baremetal host in correct state : provisioned
       expected provisioned, got provisioning
```

Can you check that then rebase this when we've got a fix landed please?

maelk:
it looks like it is not from his PR, master tests are breaking the same way

jtomasek:
> it looks like it is not from his PR, master tests are breaking the same way

Right, it looks like a timeout when waiting for the host to become provisioned (provisioning -> provisioned) which is past the 'available' phase.

maelk:
I found the problem with the tests, the BMO crashes at some point, and is not restarted because it runs locally. We should investigate the reason of the crash. Meanwhile, another option could be to run the baremetal operator in cluster with an image generated from the patchset. We would then need to change the pull policy for the deployment. If the operator crashes, it will be restarted.

maelk:
The problem is anyways not related to this patchset, so I would suggest we go forward and merge it

aneeshkp:
Tested manually, works for a new deployment. But if there is already a host with provision state set to ready, then running the operators throws an error. failed: No handler found for state \"ready\"","errorVerbose":"No handler found for state \"ready\"\naction \"ready\" 
Do we need to handle this? Otherwise, this  PR looks good.

sadasu:
> Tested manually, works for a new deployment. But if there is already a host with provision state set to ready, then running the operators throws an error. failed: No handler found for state "ready"","errorVerbose":"No handler found for state "ready"\naction "ready"
> Do we need to handle this? Otherwise, this PR looks good.

Good point. I think the fix should take care of a host that is already in "ready" state.

pkg/controller/baremetalhost/host_state_machine.go
```diff
@@ -41,7 +41,7 @@ func (hsm *hostStateMachine) handlers() map[metal3v1alpha1.ProvisioningState]sta
 		metal3v1alpha1.StateInspecting:            hsm.handleInspecting,
 		metal3v1alpha1.StateExternallyProvisioned: hsm.handleExternallyProvisioned,
 		metal3v1alpha1.StateMatchProfile:          hsm.handleMatchProfile,
-		metal3v1alpha1.StateReady:                 hsm.handleReady,
+		metal3v1alpha1.StateAvailable:             hsm.handleAvailable,
```
zaneb: We can just add another line like:

    metal3v1alpha1.ProvisioningState("ready"): hsm.handleAvailable,

That won't change the current state, but will stop anything from breaking.

jtomasek:
Updated according to @aneeshkp's guidance.

pkg/controller/baremetalhost/host_state_machine.go
aneeshkp: Fixed

maelk:
What was discussed in slack still applies here. The tests won't pass as long as CAPBM is not modified to accept the new status instead of ready. And this cannot be done as long as the status declaration is not in BMO. So this PR needs to be split in two. A first one declaring the new status, then a second one doing the actual changes and removing the old status, that would be merged after the CAPBM changes.

aneeshkp:
run test-integration after PR https://github.com/metal3-io/cluster-api-provider-baremetal/pull/257  is merged

zaneb:
I created #636 to make sure if we see a resource modified by a newer release with StateAvailable set, we will be able to handle it. We should ensure that a release is cut after that merges and before this merges, so that things still work after a downgrade.
/hold

fmuyassarov:
@zaneb @jtomasek Is there still interest in this PR?

zaneb:
I think it's an improvement. From our perspective I think this could go in at any time now without breaking anything. I forget the status of CAPM3. Has there been a release cut since #636 merged in September?

fmuyassarov:
> I think it's an improvement. From our perspective I think this could go in at any time now without breaking anything. I forget the status of CAPM3. Has there been a release cut since #636 merged in September?

Yes, the latest release [v0.4.1](https://github.com/metal3-io/cluster-api-provider-metal3/releases/tag/v0.4.1) was out on 14th of October, so after https://github.com/metal3-io/baremetal-operator/pull/636.
Metal3-dev-env also needs to be modified based on this changes, because there are a couple of places where we are expecting bmh to be in `Ready` state.

/remove-lifecycle stale

zaneb:
Requires metal3-io/metal3-dev-env#810

fmuyassarov:
> Requires [metal3-io/metal3-dev-env#810](https://github.com/metal3-io/metal3-dev-env/pull/810)

and https://github.com/metal3-io/cluster-api-provider-metal3/pull/306

fmuyassarov:
@zaneb this seems to be good to go

Issue #329, opened 2019-10-28
Passing metadata to ironic


maelk:
Currently, the only available metadata on the host when provisioning is the following : 
```json
{
 "base64_encoded_keys": [], 
 "ds": {
  "_doc": "EXPERIMENTAL: The structure and format of content scoped under the 'ds' key may change in subsequent releases of cloud-init.", 
  "ec2_metadata": {}, 
  "meta_data": {
   "instance-id": "cab62e38-b502-43f0-a3ef-08e7377d7d9a", 
   "name": "master-0", 
   "uuid": "cab62e38-b502-43f0-a3ef-08e7377d7d9a"
  }, 
  "network_json": null
 }, 
 "sensitive_keys": [], 
 "v1": {
  "_beta_keys": [
   "subplatform"
  ], 
  "availability-zone": null, 
  "availability_zone": null, 
  "cloud-name": "configdrive", 
  "cloud_name": "configdrive", 
  "instance-id": "cab62e38-b502-43f0-a3ef-08e7377d7d9a", 
  "instance_id": "cab62e38-b502-43f0-a3ef-08e7377d7d9a", 
  "local-hostname": "centos7-bf287dcf0e", 
  "local_hostname": "centos7-bf287dcf0e", 
  "platform": "configdrive", 
  "public_ssh_keys": [], 
  "region": null, 
  "subplatform": "config-disk (/dev/sda2)"
 }
}
```
The metadata field of ds does not contain sufficient information to properly deploy the kubernetes nodes. Metadatas could be provided as a Spec from the BaremetalHost object, allowing it to be configurable from the BaremetalMachine controller that will set sufficient metadata to deploy the kubernetes node.

maelk:
Implemented

PR #348, opened 2019-11-27
Introduce network_data for host

dukov:
Network data is expected under networkData key of a secret referenced
in BaremetalHostSpec.


pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -518,10 +520,15 @@ func (r *ReconcileBareMetalHost) actionProvisioning(prov provisioner.Provisioner
...
 		err = r.client.Get(context.TODO(), key, userDataSecret)
 		if err != nil {
-			return "", errors.Wrap(err,
+			return "", nil, errors.Wrap(err,
 				"failed to fetch user data from secret reference")
 		}
-		return string(userDataSecret.Data["userData"]), nil
+		var networkData map[string]interface{}
+		if err = yaml.Unmarshal(userDataSecret.Data["networkData"], &networkData); err != nil {
+			return "", nil, errors.Wrap(err, "failed to econde network_data.json from secret")
```
hardys: typo s/econde/encode/

Also what happens in the case where networkData isn't defined in the secret - will this break?  Having a test to prove that would be nice :)

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -518,10 +520,15 @@ func (r *ReconcileBareMetalHost) actionProvisioning(prov provisioner.Provisioner
...
+		}
+
+		return string(userDataSecret.Data["userData"]), networkData, nil
```
dhellmann: We should probably also harden this in case the userData key isn't present, but that can be a separate PR.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -505,10 +507,10 @@ func (r *ReconcileBareMetalHost) actionMatchProfile(prov provisioner.Provisioner
 func (r *ReconcileBareMetalHost) actionProvisioning(prov provisioner.Provisioner, info *reconcileInfo) (result reconcile.Result, err error) {
 	var provResult provisioner.Result
 
-	getUserData := func() (string, error) {
+	getUserData := func() (string, map[string]interface{}, error) {
```
dhellmann: This function is now returning more than the user data, so we should rename it.

dhellmann: Why is userData a string but networkData an interface{}?

pkg/controller/baremetalhost/baremetalhost_controller.go
dhellmann: Given the context, it should probably be "decode" or "unmarshal" rather than "encode".

pkg/provisioner/provisioner.go
```diff
@@ -20,7 +20,7 @@ type Factory func(host *metal3v1alpha1.BareMetalHost, bmcCreds bmc.Credentials,
 
 // UserDataSource is the interface for a function to retrieve user
 // data for a host being provisioned.
-type UserDataSource func() (string, error)
+type UserDataSource func() (string, map[string]interface{}, error)
```
dhellmann: Combining the user data and network data in 1 secret isn't going to work in OpenShift. The user data secrets there are managed by a part of the system completely separate from metal3. So, I think we'll need to add a new field to the Host CRD to allow a separate network data secret. It will still be possible for both the user data secret and network data secret to refer to the same secret CR, but in cases where they can't be the same we will be able to use different CRs.

Doing it that way will make it more obvious that we need a new method in this interface instead of changing UserDataSource. The new NetworkDataSource will look a lot like UserDataSource but return a map instead of a string.

pkg/controller/baremetalhost/baremetalhost_controller.go
dhellmann: Should we return a new type instead of expanding the argument list?

pkg/controller/baremetalhost/baremetalhost_controller.go
dukov: reading nonexistent key from map[string][]byte returns nil pointer to []byte, yaml.Unmarshal handles it properly so networkData will be nil as well which is ok for gophercloud module.

That being said this needs to be covered by test. Thanks for pointing this out :)

hardys:
One possible concern is this reflects an OpenStack specific metadata interface:

https://docs.openstack.org/nova/latest/user/metadata.html#openstack-format-metadata

I wonder how that would work in the event we have multiple implementations (not only Ironic), perhaps we should just have a generic "metadata" interface (in addition to user_data), since both those interfaces exist on common cloud platforms, but not necessarily the specific network_data?

Then I guess we could pass network_data as part of the "metadata" payload, and handle it correctly for the ironic case, but that metadata may reflect some different capability in the event a future alternate backend is implemented?

It's difficult to keep these interfaces completely agnostic of the implementation or underlying agents (e.g cloud-init, ignition, ...) but we should probably try :)

dhellmann:
> One possible concern is this reflects an OpenStack specific metadata interface:
> 
> https://docs.openstack.org/nova/latest/user/metadata.html#openstack-format-metadata
> 
> I wonder how that would work in the event we have multiple implementations (not only Ironic), perhaps we should just have a generic "metadata" interface (in addition to user_data), since both those interfaces exist on common cloud platforms, but not necessarily the specific network_data?
>
> Then I guess we could pass network_data as part of the "metadata" payload, and handle it correctly for the ironic case, but that metadata may reflect some different capability in the event a future alternate backend is implemented?
> 
> It's difficult to keep these interfaces completely agnostic of the implementation or underlying agents (e.g cloud-init, ignition, ...) but we should probably try :)


I'm more concerned with ensuring that a user doesn't have to understand Ironic in order to use the metal3 API than with having multiple backends, but I think that leads us to the same point. If a "network data" field is special to OpenStack, and not a more general cloud-init thing, then I want to be careful about how we add it.

How would someone know what to put into a generic metadata field? Is that passed to the host in a standard way, and the contents are up to the image to interpret? Or is it just exposing an Ironic feature without calling it out as such?

What sort of information goes into the network data field? Can we describe an API that supports that more explicitly, and then assemble the data structure that we pass to Ironic as network data ourselves?

hardys:
> How would someone know what to put into a generic metadata field? 

It's always going to depend on the agent consuming the data, which is why I said it's hard to keep these interfaces completely implementation agnostic.

We do know OpenStack, AWS, Azure, GKE, ... all have some concept of "user data" and per-instance metadata.  

The latter is often partly populated by the cloud platform (with details of automatically configured things like IP, hostname, network, etc) and some user-provided data (which is likely to be dependent on the agent in whatever image you happen to run).

In this case network_data as a metadata interface is something invented by OpenStack for Nova, but the general concept of network related metadata is probably common.

Also it's worth noting that adding network_data to the config-drive metadata will only work if the images getting deployed support OpenStack metadata format, so at some point there is always going to be coupling between the data provided, the backend, the images, and the agent in the image.

dukov:
It's a valid point that metadata has to be supported by image but we are using gophercloud under the hood which supports only OpenStack way of providing configdrive which limits us to 3 files:
1. user-data
2. network_data.json
3. meta-data

If we want to use generic approach we have to replace gohercloud with generic cloud client (I'm not aware of such though)

dhellmann:
What information goes into the network data?

dukov:
This was introduced in a liberty release and described in a spec for nova (see
https://specs.openstack.org/openstack/nova-specs/specs/liberty/implemented/metadata-service-network-info.html#rest-api-impact )

It's basically allows  you to perform network configuration of the host and it's supported by cloud init

hardys:
> we are using gophercloud under the hood which supports only OpenStack way of providing configdrive

Yes but the idea is for the backends to be pluggable - that gophercloud usage is internal to the ironic backend for BMO, but in theory others could be added for alternative provisioning tools.

It may be that the concepts of user-data, meta-data and network-data are sufficiently common that most backends could work with such abstractions, but I just wanted to discuss it to explore if there are any alternatives.

zaneb:
There's no standard for this data, so I think our only real choice is going to be to pick something that is particular to a type of cloud (e.g. OpenStack), or to pick something that is particular to a tool (e.g. cloud-init). Given this unpalatable choice, picking something OpenStack-specific seems like the least-worst option, since most tools should support OpenStack.

dhellmann:
> There's no standard for this data, so I think our only real choice is going to be to pick something that is particular to a type of cloud (e.g. OpenStack), or to pick something that is particular to a tool (e.g. cloud-init). Given this unpalatable choice, picking something OpenStack-specific seems like the least-worst option, since most tools should support OpenStack.

I would have said focus on cloud-init, because most images support that. :-)

hardys:
> I would have said focus on cloud-init, because most images support that. :-)

As discussed in https://github.com/metal3-io/metal3-docs/pull/56 we have to be careful when saying cloud-init, because it supports multiple datasources, and multiple config interfaces, including custom user-defined ones (not only cloud-config which is what most folks are probably thinking of)

IMHO it would be a mistake to expose any interface or behavior specific to a configuration tool (however popular it is), because we know for sure users can and will change the images used (including things like custom part-handlers and special agents), and also that there are already at least two active groups using different tools (cloud-init and ignition).

Having thought more about it I've reached agreement with @zaneb - probably adding this is the least-worst option at this point, and we'd have to map network_data as best as possible to any alternative provisioner implementations in future.

pkg/controller/baremetalhost/baremetalhost_controller.go
dukov: done

pkg/controller/baremetalhost/baremetalhost_controller.go
dukov: done

dukov: done switched to []byte

zaneb:
> I would have said focus on cloud-init, because most images support that. :-)

Note that cloud-init already supports the OpenStack way, so that already works with the "most images" that contain cloud-init.

We can tell e.g. Ignition that it should have better support for OpenStack (if it doesn't already - I don't actually know), and if they implement it everybody wins. It's open source so eventually somebody will care enough to do it. But we can't tell Ignition that it should be cloud-init; there's no chance of that ever happening.

hardys:
> We can tell e.g. Ignition that it should have better support for OpenStack (if it doesn't already - I don't actually know)

There is already an OpenStack provider, but it looks like we'll need to add network_data support - currently it only cares about user_data which contains the ignition json config.

https://github.com/coreos/ignition/blob/master/internal/providers/openstack/openstack.go#L43

jan-est:
We might need other metadata in cloud-init as well. Like Mael issued in:
https://github.com/metal3-io/baremetal-operator/issues/329

Personally I don't have opinion where to get this data, but all cloud-init specific data user-data, network-data and meta-data should be handled similar way. I have been working little bit different approach with meta-data only, but I am open for discussion. Check out my branch in Nordix: 

https://github.com/metal3-io/baremetal-operator/compare/master...Nordix:meta-data-jan

dukov:
> We might need other metadata in cloud-init as well. Like Mael issued in:
> #329
> 
> Personally I don't have opinion where to get this data, but all cloud-init specific data user-data, network-data and meta-data should be handled similar way. I have been working little bit different approach with meta-data only, but I am open for discussion. Check out my branch in Nordix:
> 
> [master...Nordix:meta-data-jan](https://github.com/metal3-io/baremetal-operator/compare/master...Nordix:meta-data-jan)

Nice solution. What I've got from conversation above configuration data should be provisioner agnostic so we can replace Ironic with some thing else (e.g. Ubuntu MAAS or cobbler). That being said we can stick to concept of 3 types of data: user-data, network-data and meta-data. We can extend host configuration interface with meta-data getter in upcoming PRs

pkg/controller/baremetalhost/baremetalhost_controller_test.go
```diff
@@ -618,6 +620,166 @@ func TestProvision(t *testing.T) {
...
+						Address:         "ipmi://192.168.122.1:6233",
+						CredentialsName: defaultSecretName,
+					},
+					UserData: &corev1.SecretReference{
+						Name:      "user-data",
+						Namespace: namespace,
+					},
+				}),
+			UserDataSecret: newSecret("user-data", map[string]string{"userData": "somedata"}),
+			AssertF: func(t *testing.T, hcd *hostConfigData) {
```
dhellmann: Shouldn't this be testing to ensure the data values returned from UserData() and NetworkData() are correct?

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -133,6 +133,67 @@ func (info *reconcileInfo) publishEvent(reason, message string) {
...
+// Generic method for data extraction from a Secret. Function uses dataKey
+// parameter to detirmine which data to return in case secret contins multiple
+// keys
+func (hcd *hostConfigData) getSecretData(name, namespace, dataKey string) ([]byte, error) {
+	secret := &corev1.Secret{}
+	key := types.NamespacedName{
+		Name:      name,
+		Namespace: namespace,
+	}
+	if err := hcd.bmReconciler.client.Get(context.TODO(), key, secret); err != nil {
```
dhellmann: This is the only place the reconciler is used. It would be cleaner to pass the client to hostConfigData instead of the reconciler, so that hostConfigData does not need to know about the implementation details of the reconciler.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -133,6 +133,67 @@ func (info *reconcileInfo) publishEvent(reason, message string) {
 	info.events = append(info.events, info.host.NewEvent(reason, message))
 }
 
+// hostConfigData imaplementation of cons configuration data interface
```
dhellmann:
```suggestion
// hostConfigData implementation of cons configuration data interface
```

I'm not sure what "cons" means in that sentence. It would be good to refer to the provisioner.HostConfigData interface by name.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -133,6 +133,67 @@ func (info *reconcileInfo) publishEvent(reason, message string) {
...
+// keys
+func (hcd *hostConfigData) getSecretData(name, namespace, dataKey string) ([]byte, error) {
+	secret := &corev1.Secret{}
+	key := types.NamespacedName{
+		Name:      name,
+		Namespace: namespace,
+	}
+	if err := hcd.bmReconciler.client.Get(context.TODO(), key, secret); err != nil {
+		return nil, errors.Wrap(err,
+			"failed to fetch user data from secret reference")
```
dhellmann: Now that there are multiple secrets and values, this error message should include the name of the actual secret we're trying to fetch.

dhellmann:
This looks like it's going in the right direction. I just had a few comments about ways to tighten up the code a bit and improve the testing.

pkg/controller/baremetalhost/baremetalhost_controller.go
dukov: Just typo :)

pkg/controller/baremetalhost/baremetalhost_controller.go
dukov: Fair enough. Fixed

pkg/controller/baremetalhost/baremetalhost_controller.go
dukov: Fixed

pkg/controller/baremetalhost/baremetalhost_controller_test.go
dukov: unit testing improved

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -133,6 +133,67 @@ func (info *reconcileInfo) publishEvent(reason, message string) {
...
+		Namespace: namespace,
+	}
+	if err := hcd.client.Get(context.TODO(), key, secret); err != nil {
+		errMsg := fmt.Sprintf("failed to fetch user data from secret %s defined in namespace %s", name, namespace)
+		return nil, errors.Wrap(err, errMsg)
+	}
+
+	data, ok := secret.Data[dataKey]
+	if !ok {
+		return nil, NoDataInSecretError{secret: name, key: dataKey}
```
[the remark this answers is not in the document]
dukov: done

pkg/controller/baremetalhost/baremetalhost_controller_test.go
```diff
@@ -90,7 +92,8 @@ func newTestReconciler(initObjs ...runtime.Object) *ReconcileBareMetalHost {
 	c := fakeclient.NewFakeClient(initObjs...)
 
 	// Add a default secret that can be used by most hosts.
-	c.Create(goctx.TODO(), newSecret(defaultSecretName, "User", "Pass"))
+	bmcSecret := newSecret(defaultSecretName, map[string]string{"username": "User", "password": "Pass"})
```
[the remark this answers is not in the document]
dukov: done

pkg/apis/metal3/v1alpha1/baremetalhost_types.go
```diff
@@ -145,6 +145,10 @@ type BareMetalHostSpec struct {
 	// data to be passed to the host before it boots.
 	UserData *corev1.SecretReference `json:"userData,omitempty"`
 
+	// NetworkData holds the reference to the Secret containing content
+	// of network_data.json which is plased to Config Drive
```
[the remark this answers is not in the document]
dukov: done

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -133,6 +133,68 @@ func (info *reconcileInfo) publishEvent(reason, message string) {
...
+	}
+	return string(userData), nil
+
+}
+
+// NetworkData get network configuration
+func (hcd *hostConfigData) NetworkData() ([]byte, error) {
+	if hcd.host.Spec.NetworkData == nil {
+		hcd.log.Info("NetworkData is not set returning epmty(nil) data")
+		return nil, nil
```
dhellmann: Why nil here but an empty string for user data?

pkg/provisioner/provisioner.go
```diff
@@ -18,9 +18,18 @@ type EventPublisher func(reason, message string)
...
-type UserDataSource func() (string, error)
+// HostConfigData retrieves host configuration data
+type HostConfigData interface {
+	// UserData is the interface for a function to retrieve user
+	// data for a host being provisioned.
+	UserData() (string, error)
+
+	// NetworkData is the interface for a function to retrieve netwok
+	// configuration for a host.
+	NetworkData() ([]byte, error)
```
dhellmann: Why do we expect a byte array here but a string for user data?

pkg/controller/baremetalhost/baremetalhost_controller.go
dukov: Network data now returns 'string' so this is updated in latest commit

pkg/provisioner/provisioner.go
dukov: ConfigDrive object from gophercloud library interprets NetworkData as map[string]interface{} which requires 'unmarshalling' data from secret therefore network data has to be []byte. That being said we can convert everything to string (for the price of yet another string -> []byte conversion)  fir the sake of consistency

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -133,6 +133,64 @@ func (info *reconcileInfo) publishEvent(reason, message string) {
 	info.events = append(info.events, info.host.NewEvent(reason, message))
 }
 
+// hostConfigData is an implementation of host configuration data interface.
```
zaneb: It wouldn't hurt to put this in a separate file. This one is huge already.

dukov: Done

dukov:
@dhellmann @zaneb @maelk 
Could you please review this PR one more time. It looks like PR is iterated over a quite long review process. Do we need something else to make it merged?

maelk:
/lgtm
That's a great feature and it will be very easy to expand it for the metadata. @dukov you said you will take care of that one too, right ?

dukov:
> /lgtm
> That's a great feature and it will be very easy to expand it for the metadata. @dukov you said you will take care of that one too, right ?

Thanks a lot :) Yes. I'll take care of metadata implementation

PR #349, opened 2019-11-27
Modify Makefile to install operator-sdk v0.12.0 if not installed

jan-est:
Adds install.sh script for operator-sdk installation to hack folder

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,6 +81,10 @@ $GOPATH/bin/gosec:
 $GOPATH/bin/golint:
 	go get -u golang.org/x/lint/golint
 
+.PHONY: install-operator-sdk
+install-operator-sdk:
+	@sh hack/install.sh
+
 .PHONY: docs
 docs: $(patsubst %.dot,%.png,$(wildcard docs/*.dot))
 
@@ -88,7 +92,7 @@ docs: $(patsubst %.dot,%.png,$(wildcard docs/*.dot))
 	dot -Tpng $< >$@
 
 .PHONY: e2e-local
-e2e-local:
+e2e-local: install-operator-sdk
 	operator-sdk test local ./test/e2e \
 		--namespace $(TEST_NAMESPACE) \
 		--up-local $(SETUP) \
@@ -99,14 +103,14 @@ dep:
 	dep ensure
 
 .PHONY: run
-run:
+run: install-operator-sdk
 	operator-sdk up local \
 		--go-ldflags=$(LDFLAGS) \
 		--namespace=$(RUN_NAMESPACE) \
 		--operator-flags="-dev"
 
 .PHONY: demo
-demo:
+demo: install-operator-sdk
 	operator-sdk up local \
 		--go-ldflags=$(LDFLAGS) \
 		--namespace=$(RUN_NAMESPACE) \
--- a/hack/install.sh
+++ b/hack/install.sh
@@ -1,0 +1,20 @@
+#!/bin/sh
+
+set -eux
+
+# Required operator-sdk version
+REQUIRED_OPERATOR_SDK_VERSION="v0.12.0"
+
+if [ ! "$(command -v operator-sdk)" ]; then 
+    LOCAL_OPERATOR_SDK_VERSION=""
+else 
+    LOCAL_OPERATOR_SDK_VERSION="$(operator-sdk version | head -n1 | cut -d" " -f3 | tr -d '",')"
+fi
+
+# If local operator-sdk version is not required version
+# or operator-sdk is not installed, install operator-sdk v0.12.0
+if [ "${LOCAL_OPERATOR_SDK_VERSION}" != "${REQUIRED_OPERATOR_SDK_VERSION}" ] || [ -z "${LOCAL_OPERATOR_SDK_VERSION}" ]; then
+    mkdir -p ./bin
+    curl -L https://github.com/operator-framework/operator-sdk/releases/download/${REQUIRED_OPERATOR_SDK_VERSION}/operator-sdk-${REQUIRED_OPERATOR_SDK_VERSION}-x86_64-linux-gnu -o ./bin/operator-sdk
+    chmod +x ./bin/operator-sdk
+fi


Makefile
```diff
@@ -4,6 +4,10 @@ GO_TEST_FLAGS = $(VERBOSE)
 DEBUG = --debug
 SETUP = --no-setup
 
+# Local Operator SDK version and required version
+LOCAL_VERSION = $(shell operator-sdk version | head -n1 | cut -d" " -f3 | tr -d '",')
+REQUIRED_VERSION = v0.12.0
```
maelk: this would need to be more specific naming, to show that it is about operator-sdk, not BMO

Makefile
```diff
@@ -81,14 +85,20 @@ $GOPATH/bin/gosec:
 $GOPATH/bin/golint:
 	go get -u golang.org/x/lint/golint
 
+.PHONY: install-operator-sdk
+install-operator-sdk:
+ifneq ($(LOCAL_VERSION), $(REQUIRED_VERSION))
+	@sh hack/install.sh
+endif
```
maelk: it might be cleaner to do that check in the install script

hack/install.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+set -eux
+
```
maelk: you should check here if the command exists, and if it does which version it is. if the version match the expectation, just exit, otherwise proceed

hack/install.sh
```diff
@@ -0,0 +1,11 @@
...
+# Set the release version variable
+OSDK_RELEASE_VERSION=v0.12.0
```
maelk: this is also defined in the Makefile. it should be defined only once. So bringing the check in this file would solve this issue

Makefile
maelk: wouldn't this fail if operator-sdk is not installed ?

Makefile
```diff
@@ -4,6 +4,10 @@ GO_TEST_FLAGS = $(VERBOSE)
 DEBUG = --debug
 SETUP = --no-setup
 
+# Local Operator SDK version and required version
+LOCAL_OPERATOR_SDK_VERSION = $(shell operator-sdk version | head -n1 | cut -d" " -f3 | tr -d '",')
+REQUIRED_OPERATOR_SDK_VERSION = v0.12.0
+
```
maelk: where is this used ?

hack/install.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+
```
maelk: it could be good to set some flags, like set -uex

hack/install.sh
```diff
@@ -0,0 +1,16 @@
...
+# Local operator-sdk version
+LOCAL_OPERATOR_SDK_VERSION="$(operator-sdk version | head -n1 | cut -d" " -f3 | tr -d '",')"
```
maelk: if you set the flags, then you should first check if operator-sdk exists (using which for example) then make sure that you don't have unset variables, so for example default the LOCAL_OPERATOR_SDK_VERSION to an empty string

hack/install.sh
```diff
@@ -0,0 +1,22 @@
...
+    LOCAL_OPERATOR_SDK_VERSION="$(operator-sdk version | head -n1 | cut -d" " -f3 | tr -d '",')"
+fi
+
+# If local operator-sdk version is not required version
+# or operator-sdk is not installed, install operator-sdk v0.12.0
+if [ "${LOCAL_OPERATOR_SDK_VERSION}" != "${REQUIRED_OPERATOR_SDK_VERSION}" ] || [ -z "${LOCAL_OPERATOR_SDK_VERSION}" ]; then
+    curl -LO https://github.com/operator-framework/operator-sdk/releases/download/${REQUIRED_OPERATOR_SDK_VERSION}/operator-sdk-${REQUIRED_OPERATOR_SDK_VERSION}-x86_64-linux-gnu
+    chmod +x operator-sdk-${REQUIRED_OPERATOR_SDK_VERSION}-x86_64-linux-gnu
+    sudo mkdir -p /usr/local/bin/
+    sudo cp operator-sdk-${REQUIRED_OPERATOR_SDK_VERSION}-x86_64-linux-gnu /usr/local/bin/operator-sdk
```
zaneb: Is it necessary to use sudo here? You could just install in `${HOME}/bin/`. Personally I installed the sdk using `go get` so it's in `~/go/bin/`; that would be another way to do it (and is not specific to a particular operating system and CPU arch like downloading the tarball is, so it would work for e.g. Mac users as well).

dhellmann: Yes, please use "go get" like we do for kustomize over in the cluster-api-provider repo https://github.com/metal3-io/cluster-api-provider-baremetal/blob/master/Makefile#L19

jan-est:
@zaneb and @dhellmann sorry about my slow response with PR. I think this patch and right operator-sdk version should fix "make generate" problems issue mentioned in https://github.com/metal3-io/baremetal-operator/issues/360

hack/install.sh
jan-est: I modified install.sh to install operator-sdk to ./bin folder

hack/install.sh
```diff
@@ -0,0 +1,20 @@
...
+if [ ! "$(command -v operator-sdk)" ]; then 
+    LOCAL_OPERATOR_SDK_VERSION=""
+else 
+    LOCAL_OPERATOR_SDK_VERSION="$(operator-sdk version | head -n1 | cut -d" " -f3 | tr -d '",')"
+fi
+
+# If local operator-sdk version is not required version
+# or operator-sdk is not installed, install operator-sdk v0.12.0
+if [ "${LOCAL_OPERATOR_SDK_VERSION}" != "${REQUIRED_OPERATOR_SDK_VERSION}" ] || [ -z "${LOCAL_OPERATOR_SDK_VERSION}" ]; then
+    mkdir -p ./bin
```
longkb: Should we create `.bin` folder here? With this command, the new folder will be added into `git status` command after that.
How about  `$GOPATH/bin` folder?

zaneb:
Is it necessary to require 0.12? This requires golang 1.13, which was only released 3 months ago and isn't in plenty of relatively recent distros (e.g. Fedora 30, Ubuntu 19.10).
(It's also extremely non-obvious that this would be the case, which may or may not be related to the problems people are having with it.)

jan-est:
@zaneb you might be right that operator-sdk version might not be the problem. It is better not to require golang 1.13 at this point. If anyone is not facing any problems with current operator-sdk we can drop this PR.

zaneb:
I mean, I'm definitely having problems (I used `go get` to install 0.8 or something and I can't get it to update to 0.11 due to dependency hell).
I just wonder if the real solution is to vendor the operator-sdk code like we do with other dependencies. I'd rather run code that I compiled myself, and I'm sure folks who are developing on non-Linux systems would appreciate that too.
Maybe switching to modules will make all of this easier... that reminds me to go review #367

PR #350, opened 2019-11-27
Limit BMO access to only BMH in all namespaces

Xenwar:
This commit addresses two issues.
1. Currently, the BMO has access to all resources in the 'metal3.io' api-groups. Considering principle of least privilege as a good practice, It limits it to only the BareMetalHost resource.
2. It also allows it to see BareMetalHosts in all namespaces, much like the CAPBM controller does.

### DIFF
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -1,5 +1,5 @@
 apiVersion: rbac.authorization.k8s.io/v1
-kind: Role
+kind: ClusterRole
 metadata:
   creationTimestamp: null
   name: metal3-baremetal-operator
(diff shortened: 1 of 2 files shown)


zaneb:
This would be a lot easier to approve if it had a commit message describing what the effect of each change is.

Xenwar:
> This would be a lot easier to approve if it had a commit message describing what the effect of each change is.

Thanks for the feedback. 
Added the reasoning.

deploy/role_binding.yaml
```diff
@@ -1,14 +1,14 @@
...
 subjects:
 - kind: ServiceAccount
   name: metal3-baremetal-operator
+  namespace: metal3
 - kind: User
   name: developer
 roleRef:
-  kind: Role
+  kind: ClusterRole
   name: metal3-baremetal-operator
```
maelk: This is still defined as a Role, not a clusterRole

deploy/role.yaml
```diff
@@ -41,6 +41,20 @@ rules:
 - apiGroups:
   - metal3.io
   resources:
-  - '*'
+  - baremetalhosts
```
maelk: If you are trying to apply the least privilege approach, you should also restrict the access to other resources defined above, to the minimum needed. for example restrict the verb to get and list for secrets and configmaps, and restrict the resources accessed to secrets and configmaps + other needed things like events etc.

deploy/role_binding.yaml
Xenwar: fixed

PR #353, opened 2019-11-27
Return "redfish" as type

derekhiggins:
Ironic expects the type to be "redfish"(excluding the application protocol).

Closes: #352

### DIFF
--- a/pkg/bmc/redfish.go
+++ b/pkg/bmc/redfish.go
@@ -28,7 +28,7 @@ type redfishAccessDetails struct {
 const redfishDefaultScheme = "https"
 
 func (a *redfishAccessDetails) Type() string {
-	return a.bmcType
+	return "redfish"
 }
 
 // NeedsMAC returns true when the host is going to need a separate


zaneb:
Other plugins return the unmodified URL scheme from Type(), so this makes it inconsistent.
The error message `No conductor service registered which supports driver redfish+http` suggests to me that somewhere we are passing `Type()` when we should be calling `Driver()`. Grepping through the code, the only place I see `Type()` being called is to be output in an error message...

derekhiggins:
> Other plugins return the unmodified URL scheme from Type(), so this makes it inconsistent.
> The error message `No conductor service registered which supports driver redfish+http` suggests to me that somewhere we are passing `Type()` when we should be calling `Driver()`. Grepping through the code, the only place I see `Type()` being called is to be output in an error message...

Thanks Zan, makes sense, I suspect is in the installer, here
https://github.com/openshift/installer/blob/d9955c2/pkg/tfvars/baremetal/baremetal.go#L69

I'll verify tomorrow and submit a new patch.

zaneb:
Ah, yep, that'll be the place to fix it. We split the `Type()` function into separate `Type()` and `Driver()` funcs at some point, but I didn't realise it was being consumed directly by the installer.

derekhiggins:
Closing in favour of https://github.com/openshift/installer/pull/2731

PR #357, opened 2019-12-03
Add ilo5-virtual-media bmc support

honza:
This is a follow up to #355

### DIFF
--- a/pkg/bmc/access_test.go
+++ b/pkg/bmc/access_test.go
@@ -338,7 +338,7 @@ func TestStaticDriverInfo(t *testing.T) {
 
 		{
 			Scenario: "redfish virtual media",
-			input:    "redfish+virtualmedia://192.168.122.1",
+			input:    "redfish-virtualmedia://192.168.122.1",
 			needsMac: true,
 			driver:   "redfish",
 			boot:     "redfish-virtual-media",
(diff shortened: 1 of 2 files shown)


pkg/bmc/ilo5.go
```diff
@@ -0,0 +1,47 @@
...
+
+func newiLO5VirtualMediaAccessDetails(parsedURL *url.URL) (AccessDetails, error) {
+	return &iLO5AccessDetails{
+		hostname: parsedURL.Hostname(),
+		path:     parsedURL.Path,
+	}, nil
+}
+
+func (a *iLO5AccessDetails) Type() string {
+	return "ilo5"
```
zaneb: This method is used for debugging, so for consistency with other BMC plugins it should return the full URL scheme from the parsedURL.

pkg/bmc/ilo5.go
```diff
@@ -0,0 +1,47 @@
...
+
+func (a *iLO5AccessDetails) Driver() string {
+	return "redfish"
+}
+
+func (a *iLO5AccessDetails) DriverInfo(bmcCreds Credentials) map[string]interface{} {
+	return map[string]interface{}{
+		"ilo_username": bmcCreds.Username,
+		"ilo_password": bmcCreds.Password,
+		"ilo_address":  a.hostname,
```
zaneb: The redfish driver doesn't accept these options: https://docs.openstack.org/ironic/latest/admin/drivers/redfish.html#registering-a-node-with-the-redfish-driver

pkg/bmc/ilo5.go
```diff
@@ -0,0 +1,47 @@
...
+
+type iLO5AccessDetails struct {
+	hostname string
+	path     string
+}
+
+const iLO5DefaultScheme = "https"
+
+func init() {
+	registerFactory("ilo5+virtualmedia", newiLO5VirtualMediaAccessDetails)
```
zaneb: It's not clear to me why we're creating another BMC plugin rather than just registering the existing redfish one to also handle this URL scheme?

PR #367, opened 2019-12-16
Migrate to Go modules

honza:
Steps taken:

* Run `dep ensure` to make sure that the vendor directory is up-to-date
* Run `go mod init ...` to create Go module files
* Run tests, build, etc to make sure everything works
* Upgrade local `operator-sdk` cli tool to the latest version
* Update requires by hand in `go.mod`
* Tests, lint, etc all pass
* Remove vendor directory
* Remove references to `dep` tool from Makefile

Official [migrating to Go modules](https://blog.golang.org/migrating-to-go-modules) documentation


zaneb:
What's the thinking behind re-adding the vendor directory? From what I read this is only for compatibility with non-module-aware versions of Go?

honza:
> What's the thinking behind re-adding the vendor directory? 

The idea is that CI doesn't have to download all the files.

zaneb:
OK, that sounds reasonable.
/lgtm
/test-v1a1-integration

honza:
Do you want me to rebase?

honza:
Updated to use operator-sdk `0.15.1`. 

Currently blocked by https://github.com/metal3-io/project-infra/pull/50

honza:
More context:

The current Prow job uses Go 1.12; this version doesn't automatically detect go modules. The next version, 1.13 does. All jobs which use the `go` binary will fail because it can't find the necessary dependencies (it's blind to the go.mod and go.sum file). With 1.13, the `go` command will see the `go.mod` file, and download the necessary packages.

We are migrating to use go modules because the operator-sdk forces us to.

zaneb:
> The current Prow job uses Go 1.12; this version doesn't automatically detect go modules.

Surely that's just a matter of setting the environment variable `GO111MODULE=on`?

honza:
> Surely that's just a matter of setting the environment variable GO111MODULE=on?

Yes.

But 1.13.x is the latest stable?

honza:
It seems that the canned jobs don't use our `Makefile` so I'm not sure how I can set the envvar in them.

zaneb:
This looks good.
I guess the ideal in a situation like this would be to separate out the change from dep->go mod from any version changes in dependencies. It seems like that wasn't possible in this case, and I don't want to ask you to go back and start again because it's hard enough to get anything passing all of CI!
In lieu of that, can I suggest that we document somewhere all of the dependencies that actually changed? golang and the operator-sdk seem to be two of them.
Are there any changes to checked-in generated code resulting from updating the operator-sdk?

honza:
> and I don't want to ask you to go back and start again

I am a git wizard :sunglasses: I can move some commits around.

I tried to make everything explicit in #444 

* Downgrade operator-sdk to 0.11
* Migrate to go modules
* Remove vendor and Gopkg
* Upgrade operator-sdk to 0.15
* Update build settings (docker, makefile, etc)

The commit messages in the new PR are much more explicit about what's going on.

honza:
Closing in favor of #444

PR #368, opened 2019-12-18
Fix incorrect file name

vincent-pli:
Should be `metal3.io_baremetalhosts_crd.yaml`


longkb:
This PR is overlap with #366 :)

zaneb:
Thanks for opening this! I just approved #366 so this should be fixed shortly.

PR #369, opened 2019-12-18
Expose error types through the CRD Status

zaneb:
Previously, external consumers (such as the OpenShift UI) could only rely on the provisioning status to determine the cause of an error in a structured way (i.e. without having a lookup table of the possible error strings). This requires us to perform a lot of gymnastics with the provisioning status, and still did not always yield the correct results. This PR adds a well-defined field for determining the type of an error that systems such as UIs can use to direct the user to debugging the relevant thing.


zaneb:
Thanks, I had failed to pick up on that for some reason. I had to rebase this after merging the metrics PR anyway, so I switched the order of the arguments to make it consistent.

Issue #327, opened 2019-10-23
enable pruning in openapi settings


dhellmann:
We need to ensure that pruning is turned on in the openapi settings.

https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#pruning-versus-preserving-unknown-fields describes the field

https://github.com/openshift/api/pull/478/commits/db65909e9583109098bc5189fad304f82ceac667 is an example of setting it

dhellmann:
I think this depends on #326

zaneb:
We'll [get this](https://kubernetes.io/blog/2019/06/20/crd-structural-schema/#pruning-don-t-preserve-unknown-fields) by default when we switch to apiVersion `apiextensions.k8s.io/v1`. Supposedly there's a way to opt in with `apiextensions.k8s.io/v1beta1` (the apiVersion in the current CRD), but it's not obvious that there's a way to make kubebuilder autogenerate the opt-in.

zaneb:
We updated to k8s 1.16, so I think we should now be able to switch to v1 (currently still at v1beta1 though).

zaneb:
/remove-lifecycle stale

Since the CRD is generated by the operator-sdk, this is dependent on the operator-sdk generating it with version `apiextensions.k8s.io/v1`. As of v0.13.0 it was not.

zaneb:
Good news: kubebuilder [now supports](https://book.kubebuilder.io/reference/markers/crd-processing.html) generating the CRD with the `x-kubernetes-preserve-unknown-fields` option.
Bad news: it only supports setting it to true :facepalm:

So the solution remains switching to `apiextensions.k8s.io/v1`.

PR #371, opened 2019-12-19
Change CRD apiversion to apiextensions.k8s.io/v1

kirankt:
This change will update the CRD apiversion from apiextensions.k8s.io/v1beta1 to v1, which is the default since k8s 1.16.
Address #327


zaneb:
Is the bot alive? Let's find out...
/approve
/lgtm

russellb:
> Is the bot alive? Let's find out...
> /approve
> /lgtm

no, you will have to merge manually for now

PR #374, opened 2019-12-20
Change BMO CRD apiversion to v1

kirankt:
This change moves the BMO CRD from apiversion v1beta1 to v1 and also makes changes to the CRD to pass validation.

Tested with metal3-dev-scripts and K8s 1.17
Address #327

### DIFF
--- a/deploy/crds/metal3.io_baremetalhosts_crd.yaml
+++ b/deploy/crds/metal3.io_baremetalhosts_crd.yaml
@@ -3,35 +3,6 @@ kind: CustomResourceDefinition
 metadata:
   name: baremetalhosts.metal3.io
 spec:
-  additionalPrinterColumns:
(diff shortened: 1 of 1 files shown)


zaneb:
Was this generated using `make generate`?

kirankt:
> Was this generated using `make generate`?

No it wasn't. It was handcrafted. :-)

make generate still seems to create the v1beta1 version..

kirankt:
Can confirm that as of operator-sdk v0.1.13 the CRDs created by it are still v1beta1. Perhaps, we should leave it at that and close this PR?

zaneb:
You mean v0.13.0?
Tests won't pass unless the CRD is identical to the one that is generated by the operator-sdk using `make generate`. I'm not interested in maintaining this file manually, so I think the answer is going to be to wait for this to make its way into the operator-sdk.

PR #376, opened 2019-12-27
Restrict BMO access to only required RBAC verbs

Xenwar:
This PR applies the principle of least privilege on the BMO by limiting verbs it can invoke.
Closes metal3-io#356

### DIFF
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -9,14 +9,24 @@ rules:
   resources:
   - configmaps
   verbs:
-  - "*"
+  - list
+  - get
+  - create
+  - update
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  verbs:
+  - watch
+  - list
+  - update
 - apiGroups:
   - ""
   resources:
   - events
-  - secrets
   verbs:
-  - "*"
+  - create
 - apiGroups:
   - ""
   resources:


hardys:
@Xenwar can you provide any further details as to how you decided this set of verbs, as IIRC we had problems ref #350 and #354 with this kind of change previously?

Xenwar:
> @Xenwar can you provide any further details as to how you decided this set of verbs, as IIRC we had problems ref #350 and #354 with this kind of change previously?

I have added the verbs incrementally this time. 
I do provision and provision and get feedback from the failing pod logs and add the right required verbs. Followed the procedure repeatedly and arrived at those verbs. I did run the tests and do provision/de-provision and I did not see any error. 

If you concern is if this could break the environment, yest it could, but less likely. 
Should it fail, then we can just add those missing verbs in a new patch.

deploy/role.yaml
```diff
@@ -9,14 +9,24 @@ rules:
   resources:
   - configmaps
```
dhellmann: I suspect that the deployment instructions need these ConfigMap permissions, but that the operator itself doesn't.

Xenwar: I added the verbs for the config map by monitoring the logs of BMO. 
My conclusion is that BMO needs needs the permission on the ConfigMap(s).

Xenwar:
Note: Previous CI run of both ubuntu and centos had passed. However, the run was showing pending on this interface. 
Re-running the test again.

PR #385, opened 2020-01-10
Disable fast track temporarily

maelk:
Until the ironic bug is fixed upstream, we need to disable
fasttrack for the provisioning to work properly

/cc @elfosardo


alexz-kh:
could someone please provide link to upstream issue\bug ?

Issue #377, opened 2019-12-27
BMO Crashing when deleting a control plane node


Xenwar:
When running the following scripts from Metal-dev environment, the BMO is crashing and stays down for a while. 
```
./scripts/v1alpha2/create_controlplane.sh 
# Wait for a while
./scripts/v1alpha2/delete_controlplane.sh 
```
BMO logs shown below
```
{"level":"info","ts":1577430729.4292197,"logger":"cmd","msg":"Go Version: go1.10.8"}
{"level":"info","ts":1577430729.4307537,"logger":"cmd","msg":"Go OS/Arch: linux/amd64"}
{"level":"info","ts":1577430729.431537,"logger":"cmd","msg":"Version of operator-sdk: v0.12.0"}
{"level":"info","ts":1577430729.4318814,"logger":"cmd","msg":"Component version: metal3-io/baremetal-operator was not built with version info"}
{"level":"info","ts":1577430729.928501,"logger":"controller-runtime.metrics","msg":"metrics server is starting to listen","addr":"127.0.0.1:8085"}
{"level":"info","ts":1577430729.928687,"logger":"cmd","msg":"Registering Components."}
{"level":"info","ts":1577430729.9287965,"logger":"baremetalhost_ironic","msg":"ironic settings","endpoint":"http://172.22.0.2:6385/v1/","inspectorEndpoint":"http://172.22.0.2:5050/v1/","deployKernelURL":"http://172.22.0.2:6180/images/ironic-python-agent.kernel","deployRamdiskURL":"http://172.22.0.2:6180/images/ironic-python-agent.initramfs"}
{"level":"info","ts":1577430729.9296534,"logger":"cmd","msg":"Starting the Cmd."}
I1227 07:12:09.929819       1 leaderelection.go:241] attempting to acquire leader lease  metal3/baremetal-operator...
{"level":"info","ts":1577430729.9307973,"logger":"controller-runtime.manager","msg":"starting metrics server","path":"/metrics"}
I1227 07:12:27.386550       1 leaderelection.go:251] successfully acquired lease metal3/baremetal-operator
{"level":"info","ts":1577430747.3877852,"logger":"controller-runtime.controller","msg":"Starting EventSource","controller":"metal3-baremetalhost-controller","source":"kind source: /, Kind="}
[...]

PR #387, opened 2020-01-13
Fix crash when cancelling provisioning

zaneb:
If you cancel provisioning before it has finished, both NeedsProvisioning() and NeedsDeprovisioning() return false because the latter assumes that provisioning has to be complete before the host needs to be deprovisioned.

Because the state machine remembers the state, it tries to continue provisioning (if it ever finished it would move on to deprovisioning, but it doesn't). This results in a crash because the Ironic provisioner assumes (quite reasonably) that the Image is non-null when we are trying to provision.

### DIFF
--- a/pkg/apis/metal3/v1alpha1/baremetalhost_types.go
+++ b/pkg/apis/metal3/v1alpha1/baremetalhost_types.go
@@ -644,11 +644,14 @@ func (host *BareMetalHost) NeedsDeprovisioning() bool {
 	if host.Spec.ExternallyProvisioned {
 		return false
 	}
-	if host.Status.Provisioning.Image.URL == "" {
-		return false
-	}
 	if host.Spec.Image == nil {
 		return true
+	}
+	if host.Spec.Image.URL == "" {
+		return true
+	}
+	if host.Status.Provisioning.Image.URL == "" {
+		return false
 	}
 	if host.Spec.Image.URL != host.Status.Provisioning.Image.URL {
 		return true
--- a/pkg/apis/metal3/v1alpha1/baremetalhost_types.go
+++ b/pkg/apis/metal3/v1alpha1/baremetalhost_types.go
@@ -603,9 +603,6 @@ func (host *BareMetalHost) NeedsHardwareInspection() bool {
 // status and returns true when more work is needed or false
 // otherwise.
 func (host *BareMetalHost) NeedsProvisioning() bool {
-	if host.Spec.ExternallyProvisioned {
-		return false
-	}
 	if !host.Spec.Online {
 		// The host is not supposed to be powered on.
 		return false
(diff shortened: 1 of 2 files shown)


Xenwar:
This fix closes  issue #377

maelk:
The CI failed because this is still using fasttrack. Rebasing on current master will remove the fasttrack. We can then test again

zaneb:
Ugh, I am too used to Zuul where it merges for you before testing.

maelk:
> Ugh, I am too used to Zuul where it merges for you before testing.

We could add something in the CI that would actually do the merge and then run the tests. That would also return error in case of merge conflict

pkg/apis/metal3/v1alpha1/baremetalhost_types.go
```diff
@@ -641,15 +638,15 @@ func (host *BareMetalHost) WasProvisioned() bool {
 // NeedsDeprovisioning compares the settings with the provisioning
 // status and returns true when the host should be deprovisioned.
 func (host *BareMetalHost) NeedsDeprovisioning() bool {
-	if host.Spec.ExternallyProvisioned {
-		return false
+	if host.Spec.Image == nil {
```
dhellmann: This changes the meaning of this method. Before we had the higher-level state machine, this used to be trying to indicate when a transition needed to happen to clean a host. Now it means something closer to "should this host not have an image on it" and leaving it up to the caller to decide if that means a transition is needed. Maybe we should rename the method?

pkg/apis/metal3/v1alpha1/baremetalhost_types_test.go
```diff
@@ -376,24 +376,6 @@ func TestHostNeedsDeprovisioning(t *testing.T) {
 			Expected: false,
 		},
 
-		{
-			Scenario: "externally provisioned",
```
dhellmann: Don't we still need this test case?

pkg/apis/metal3/v1alpha1/baremetalhost_types.go
zaneb: Yeah, it still has that meaning but only when called from certain states. I'm open to suggestions for a better name... `IsProvisionedOrProvisioningAndNeedsDeprovisioning()` didn't quite trip off the tongue ;)
Another option is to check the state inside this function so that it doesn't return true if provisioning hasn't at least started. I was trying to avoid making these functions state-aware and keep the state logic in the state machine. Probably better that it be explicit rather than you have to remember where it's safe to call it from though.
Maybe this function should just be moved to the state machine and made private instead of living here and inviting people to call it for more general purposes?

pkg/apis/metal3/v1alpha1/baremetalhost_types_test.go
zaneb: No, because we don't call this from any states where transitioning to ExternallyProvisioned is allowed (see the second commit message that GitHub does a very good job of hiding from reviewers).
I know that's an unsatisfying answer, due to the problems identified in your other comment.

pkg/apis/metal3/v1alpha1/baremetalhost_types.go
zaneb: I opened a separate PR #397 to address this, since this one was blocking people and got merged already.

PR #388, opened 2020-01-14
Handle registration independently of provisioning state

zaneb:
We may need to re-register the Host at any point in time (whenever the credentials secret changes, for a start). Previously we did this by switching the provisioning state back to registering and then back to the current state when finished.
Since registering an provisioning are in effect orthogonal, separate the registration phase out from the provisioning state. Always ensure the Host registration is current before running the rest of the state machine. To enable this the RegistrationError state is removed, along with the other error provisioning states. The OperationalState and ErrorType fields provide a more reliable reporting of errors than what those provisioning states could.


maelk:
This PR does not include the fasttrack fix. Please rebase to fix the current problem in the CI

pkg/controller/baremetalhost/host_state_machine.go
```diff
@@ -134,21 +129,39 @@ func (hsm *hostStateMachine) checkInitiateDelete() bool {
...
+		// part because we are also willing to pass empty credentials
+		// after this point), but this means that we may not be able to
+		// deprovision a Host if the Ironic database pod is rescheduled
+		// during deprovisioning, or if the credentials need to be
+		// modified.
+		return
 	}
-	return changeState
+
+	if !hsm.Host.Status.GoodCredentials.Match(*info.bmcCredsSecret) {
```
dhellmann: I think I would write this the opposite way, and use an early return here if the credentials do match. Then the bulk of the logic for the method is outside of any conditional block.

zaneb: You're right, that is definitely better. Fixed.

maelk:
Please rebase the PR to fix the CI issue

zaneb:
It would be really nice if we could either merge this or decide that we're not going to merge it, so that I can stop carrying two different state machine implementations around in my head :)

pkg/controller/baremetalhost/host_state_machine.go
```diff
@@ -135,21 +130,43 @@ func (hsm *hostStateMachine) checkInitiateDelete() bool {
...
+		// if we find it missing once a delete has been requested (in
+		// part because we are also willing to pass empty credentials
+		// after this point), but this means that we may not be able to
+		// deprovision a Host if the Ironic database pod is rescheduled
+		// during deprovisioning, or if the credentials need to be
+		// modified.
+		return
+	}
+
+	if hsm.Host.Status.GoodCredentials.Match(*info.bmcCredsSecret) {
```
dhellmann, line 151: Would this mean that a host that was successfully registered once is not re-registered if the database is wiped after a pod restart? Maybe that's handled the next time the provisioner is used to try to do something with the host?

Elsewhere we say that after a host passes through the registration state it never goes back. I wonder if we need this check at all any more?

pkg/controller/baremetalhost/host_state_machine.go
```diff
@@ -135,21 +130,43 @@ func (hsm *hostStateMachine) checkInitiateDelete() bool {
...
+		// Credentials are unchanged since we verified them.
+		return
+	}
+
+	recordStateBegin(hsm.Host, metal3v1alpha1.StateRegistering, metav1.Now())
+	if hsm.Host.Status.ErrorType == metal3v1alpha1.RegistrationError {
+		if hsm.Host.Status.TriedCredentials.Match(*info.bmcCredsSecret) {
```
dhellmann, line 158: Maybe we don't need this test any more? Or do we need to wait for #610?

pkg/controller/baremetalhost/host_state_machine.go
zaneb: > Would this mean that a host that was successfully registered once is not re-registered if the database is wiped after a pod restart?

It means that we won't call actionRegistering below.
In the previous code, it meant we wouldn't flip back to the Registering state (which calls actionRegistering).
So the effect is the same, which is **not** to say that there isn't a bug. I suspect there are a lot of states where we don't re-register the host if the DB is dropped. (The ones that call Adopt() are the only ones where we _do_ AFAIK.)

> Elsewhere we say that after a host passes through the registration state it never goes back.

This is more of an accounting thing. We do re-register below if the creds change, by doing the exact stuff that the Registering state does, but we don't change the provisioning state while we do it any more.

pkg/controller/baremetalhost/host_state_machine.go
zaneb: Yeah, I suspect the existence of this code is a mistake - registration could have failed for a reason other than the creds being wrong, so we probably shouldn't force the user to update them before we look again. #610 provides an exponential backoff on failure, which is what we really want.
This is bug-for-bug compatible with how it works now, so any changes should go in a separate PR.

dhellmann:

/hold

pkg/controller/baremetalhost/host_state_machine.go
kashifest: What will happen incase we had a BMC connectivity issue for a brief moment, and the BMH went to `RegistrationError`, and then after a while the connectivity is reinstated. The workflow you are proposing here won't try to register again since the change in scenario was not related to BMC credentials but related to BMC connectivity. Am I understanding it right?

zaneb: Yes, but with a caveat: it likely wouldn't have gone into RegistrationError before either - it could easily have been ProvisioningError or PowerManagementError, because we don't know the cause of the error, only what we were trying to do when it happened.

In those scenarios it's fine anyway, because we don't need to re-register (i.e. send new credentials to Ironic) to reconnect; a retry is enough. The theory behind this code is probably that the only circumstances we need to re-register is if the password is changed on the BMC, and in that case it won't work until the creds are updated in k8s anyway so there's no point retrying until they are.

The problem with that logic is that if we do change the credentials and connectivity is interrupted *while we are updating the credentials* then we erroneously assume it was because the credentials are bad and don't try again. (The same would presumably apply if you change them in the 'wrong' order - in the k8s Secret first and later on the BMC.) If that proves to be the case, it's a (pre-existing) bug.

dhellmann:
#655 has merged

/hold cancel

pkg/controller/baremetalhost/host_state_machine.go
furkatgofurov7: > In those scenarios it's fine anyway, because we don't need to re-register (i.e. send new credentials to Ironic) to reconnect; a retry is enough. The theory behind this code is probably that the only circumstances we need to re-register is if the password is changed on the BMC, and in that case it won't work until the creds are updated in k8s anyway so there's no point retrying until they are.
> 

Does this mean that in case of BMC connectivity issue, no extra logic needs to be added to host state machine and retry is already implemented? I am interested in this specific case where only bmc connectivity is lost and nothing wrong happened with the credentials. I would like to know your opinion on that specific case, and what needs to be done to get BMH out of `RegistrationError` @zaneb ? Thanks!

dhellmann:
This is going to need to be rebased, but the logic in this version looks good.

pkg/controller/baremetalhost/host_state_machine.go
zaneb: > Does this mean that in case of BMC connectivity issue, no extra logic needs to be added to host state machine and retry is already implemented?

Yes, as of #610 retry is implemented.
(Prior to that we in theory stopped reconciling after a registration error, and waited for some change to the Host or Secret. In practice I suspect that in many cases writes to the status would have triggered an immediate re-reconcile and resultant retry anyway.)

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -455,17 +455,14 @@ func (r *BareMetalHostReconciler) actionInspecting(prov provisioner.Provisioner,
 		return recordActionFailure(info, metal3v1alpha1.InspectionError, provResult.ErrorMessage)
 	}
 
-	if details != nil {
-		info.host.Status.HardwareDetails = details
-		return actionComplete{}
-	}
+	info.host.ClearError()
```
dhellmann, line 458: It looks like sometimes we clear the error in the controller and sometimes in the state machine. That may be an area to clean up in a future PR.

controllers/metal3.io/host_state_machine.go
```diff
@@ -203,39 +220,17 @@ func (hsm *hostStateMachine) handleUnmanaged(info *reconcileInfo) actionResult {
...
-func (hsm *hostStateMachine) handleRegistrationError(info *reconcileInfo) actionResult {
-	if !hsm.Host.Status.TriedCredentials.Match(*info.bmcCredsSecret) {
-		info.log.Info("Modified credentials detected; will retry registration")
-		hsm.NextState = metal3v1alpha1.StateRegistering
-		return actionComplete{}
+	// Getting to the state handler at all means we have successfully
+	// registered using the current BMC credentials, so we can move to the
+	// next state. We will not return to the Registering state, even
+	// if the credentials change and the Host must be re-registered.
+	hsm.Host.ClearError()
```
dhellmann, line 227: See earlier comment about clearing errors.

controllers/metal3.io/host_state_machine_test.go
```diff
@@ -221,6 +217,20 @@ func TestErrorCountIncreasedWhenProvisionerFails(t *testing.T) {
 	}
 }
 
+func TestErrorCountIncreasedWhenRegistrationFails(t *testing.T) {
```
dhellmann, line 220: I'm a little surprised at how few changes there are to the tests. Maybe that says more about our test coverage than this change, though.

dhellmann (approved):
This looks good to me. I don't want to approve it with 1 reviewer late on a Friday, so I'll leave it for others to review next week.

/approve
/cc @andfasano

controllers/metal3.io/host_state_machine.go
zaneb: This is a bit of a special case because we don't call an action*() function from this state any more. I'm not sure it's even necessary, but I wrote this patch a year ago so I might have known the reason then :D

furkatgofurov7:
@zaneb @dhellmann I have a question regarding a specific case, even I asked it earlier in this PR, I will do it once again. 
We have a situation where Ironic can't recover BMHs when connectivity with BMCs fails for a some time. After, Ironic will set a baremetal node as `failed` when it can't connect to it's BMC for a short period of time. Once a bmh is in `error` state, it cannot be recovered, breaking the whole cluster. I have tested it locally, to imitate the situation of BMC connectivity loss, I have deleted VBMC container and checked the logs, it is throwing registration_error (see logs below):

```
`Status:
  Error Message:  Failed to get power state for node 56055790-ddda-427c-a16a-f10261e52382. Error: Redfish connection failed for node 56055790-ddda-427c-a16a-f10261e52382: Unable to connect to https://10.33.131.4/redfish/v1/. Error: HTTPSConnectionPool(host='10.33.131.4', port=443): Max retries exceeded with url: /redfish/v1/ (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7f6982f7a080>: Failed to establish a new connection: [Errno 110] ETIMEDOUT',))
  Error Type:     registration error
  Good Credentials:
  Hardware Profile:  
  Last Updated:      2020-08-18T23:34:15Z
  Operation History:
    Deprovision:
      End:    <nil>
      Start:  <nil>
    Inspect:
      End:    <nil>
      Start:  <nil>
    Provision:
      End:    <nil>
      Start:  <nil>
    Register:
      End:             2020-08-18T23:34:15Z
      Start:           2020-08-18T23:20:47Z
  Operational Status:  error
  Powered On:          false
  Provisioning:
    ID:  56055790-ddda-427c-a16a-f10261e52382
    Image:
      Checksum:  
      URL:       
    State:       registration error
  Tried Credentials:
    Credentials:
      Name:               ephemeral
      Namespace:          dc325-vpod1-capi
    Credentials Version:  9857
Events:                   <none>` 
```

I was just wondering, as I checked the code last time, we do not consider BMC losing connectivity and going to registration error state in host_state_machine, maybe we could add that case too? Although, @zaneb answered this why we do not need to re-register in that case and simple retry is enough, would you mind explain more what happens if even retry fails or simply put this, will nodes get out of error state with current implemented behaviour in BMO? I just would like to know your opinions on that. Thanks in advance!
Sorry, I am still new to the repo and trying to get more familiar with different parts of it.

zaneb:
@furkatgofurov7 let's open a separate issue to discuss it. The goal of this PR is to **not** change any behaviour other than the reported provisioning state, so it's really nothing to do with this and any discussion here is going to get lost.

controllers/metal3.io/baremetalhost_controller.go
andfasano: The `ClearError` is invoked alongside to `actionComplete{}` or `actionContinue` (but not the viceversa). Could make sense to have a utility method to group those commands together, like the `recordActionFailure`?

zaneb: IIRC I looked into doing the `ClearError()` in a single place in `Reconcile()` when we get the actionResult. However, it's not as consistent as you would hope so we have to keep doing it on a case-by-case basis.

The changes in this last commit are just refactoring to make the control flow of this action look like the others, so I wouldn't get too hung up on this particular part. If I'd known this PR was going to take a year I'd have just submitted this commit separately.

dhellmann:
/lgtm

Let's get this in and iterate if we need to.

PR #395, opened 2020-01-20
Fix error generating CRDs with operator-sdk v0.12.0

kirankt:
Using operator-sdk v0.12.0, CRD generation part fails with an unspecified error. Digging further reveals a slight format change in validation pragma for enums (';' replaces ',').

Address #360

### DIFF
--- a/deploy/crds/metal3.io_baremetalhosts_crd.yaml
+++ b/deploy/crds/metal3.io_baremetalhosts_crd.yaml
@@ -470,6 +470,7 @@ spec:
             operationalStatus:
               description: OperationalStatus holds the status of the host
               enum:
+              - ""
               - OK
               - discovered
               - error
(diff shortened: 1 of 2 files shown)


deploy/crds/metal3.io_baremetalhosts_crd.yaml
```diff
@@ -470,7 +470,6 @@ spec:
             operationalStatus:
               description: OperationalStatus holds the status of the host
               enum:
-              - ""
```
zaneb: This is very much required otherwise the operator can never update the CR (I discovered this the hard way).

kirankt: OK. Just added it. Please note that format is slight different now with ; as separators.

PR #397, opened 2020-01-21
Rename NeedsDeprovisioning() method

zaneb:
Since 3452e801c9704bfbf07351c36925fb74b1039ea5 this method is no
longer well-named, since it works only under assumptions about the state
the Host is in when it is called. Rename it and make it a private method
of the HostStateMachine.


pkg/controller/baremetalhost/host_state_machine.go
```diff
@@ -285,7 +301,7 @@ func (hsm *hostStateMachine) handleProvisioningError(info *reconcileInfo) action
 }
 
 func (hsm *hostStateMachine) handleProvisioned(info *reconcileInfo) actionResult {
-	if hsm.Host.NeedsDeprovisioning() {
+	if hsm.provisioningCancelled() {
```
dhellmann: Would a host that has already completed provisioning end up here?

zaneb: Yes. It's not a perfect name.

maelk:
The previous integration test run was giving 404. Re-running to check the current state

dhellmann:
Do we still want this change?

zaneb:
> Do we still want this change?

It still feels like an improvement.

dhellmann (commented):
/lgtm

dhellmann:
I merged it by hand.

Issue #148, opened 2019-04-18
expand image checksum type support


mhrivnak:
The `BareMetalHost.Spec` currently supports [providing a checksum](https://github.com/metalkube/baremetal-operator/blob/8e13289/pkg/apis/metalkube/v1alpha1/baremetalhost_types.go#L150) for the image to be provisioned. The checksum can be provided directly as a string, or as a URL which will be fetched.

These questions need to be answered:
* Is the checksum assumed to be md5sum?
* Should we allow the checksum type to be specified somehow, either as a separate field, or as part of the same field? example of the latter: "sha256:abcdef..."

Whatever answers emerge, the documentation for the Image struct should be improved with more detail.

dhellmann:
cc @juliakreger @dtantsur

juliakreger:
Ff translated directly to ironic's image_checksum it is assumed to be md5. Two different fields are used for other checksums: 'image_os_hash_algo' and 'image_os_hash_value'. FWIW, it requires a stein IPA ramdisk.

hardys:
It would be nice to revive this as in combination with https://storyboard.openstack.org/#!/story/2006936 we could deploy the OKD images directly (they are shipped as xz compressed qcow2 images, with an sha256sum)

I think specifying the type of checksum in the existing field e.g `sha256:xxxx` seems reasonable and keeps the interface terse but still flexible - any objections to that approach?

We could retain the existing behavior by default by assuming md5 when no prefix is given?

@dtantsur can you confirm if `image_os_hash_value` works with the URL support we now rely on?

kirankt:
Question. If the checksum is passed, say, as: "sha256:https://...", does it mean that it needs to converted into md5hash before passing it on to ironic?

kirankt:
Hi, I have a question. Is the plan to still keep "image_checsum" in addition to a combo of "image_os_hash_algo:image_os_hash_value" or completely replace "image_checksum" with the latter?

hardys:
> Hi, I have a question. Is the plan to still keep "image_checsum" in addition to a combo of "image_os_hash_algo:image_os_hash_value" or completely replace "image_checksum" with the latter?

Perhaps @juliakreger can answer that, but I'd assume we continue using the existing interface for md5 checksums (e.g the default case) unless we're advised otherwise.

kirankt:
Fixed in #401 
/close.

PR #401, opened 2020-01-23
Expand image checksum support

kirankt:
The BareMetalHost.Spec currently supports providing a checksum for the image to be provisioned as either a string or URL. But at the moment, the checksum defaults to md5 only. This PR supports other types of checksums such as sha256, etc.

The BareMetalHost.Spec's checksum string value is expanded to specify other checksum algorithms and its value in the form of: "sha256:abcdef...", "md5:xyzabc...", etc

Address #148

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -73,6 +73,9 @@ mainly, but not only, provisioning details.
   * *url* -- The URL of an image to deploy to the host.
   * *checksum* -- The actual checksum or a URL to a file containing
     the checksum for the image at *image.url*.
+  * *checksumType* -- Checksum algorithms can be specified. Currently
+    only `md5`, `sha256`, `sha512` are recognized. If nothing is specified
+    `md5` is assumed.
 
 * *userData* -- A reference to the Secret containing the cloudinit user data
   and its namespace, so it can be attached to the host before it boots
--- a/pkg/apis/metal3/v1alpha1/baremetalhost_types.go
+++ b/pkg/apis/metal3/v1alpha1/baremetalhost_types.go
@@ -201,6 +201,21 @@ type BareMetalHostSpec struct {
 	ExternallyProvisioned bool `json:"externallyProvisioned,omitempty"`
 }
 
+// Checksum Algorithm Type
+// +kubebuilder:validation:Enum=md5;sha256;sha512
+type ChecksumType string
+
+const (
+	// MD5 checksum type
+	MD5 ChecksumType = "md5"
+
+	// SHA256 checksum type
+	SHA256 ChecksumType = "sha256"
+
+	// SHA512 checksum type
+	SHA512 ChecksumType = "sha512"
+)
+
 // Image holds the details of an image either to provisioned or that
 // has been provisioned.
 type Image struct {
@@ -209,6 +224,10 @@ type Image struct {
 
 	// Checksum is the checksum for the image.
 	Checksum string `json:"checksum"`
+
+	// ChecksumType is the checksum algorithm for the image.
+	// e.g md5, sha256, sha512
+	ChecksumType ChecksumType `json:"checksumType,omitempty"`
 }
 
 // FIXME(dhellmann): We probably want some other module to own these
@@ -745,6 +764,27 @@ func (host *BareMetalHost) OperationMetricForState(operation ProvisioningState) 
 		metric = &history.Deprovision
 	}
 	return
+}
+
+// GetImageChecksum returns the hash value and its algo.
+func (host *BareMetalHost) GetImageChecksum() (string, string, bool) {
+	checksum := host.Spec.Image.Checksum
+	checksumType := host.Spec.Image.ChecksumType
+
+	if checksum == "" {
+		// Return empty if checksum is not provided
+		return "", "", false
+	}
+	if checksumType == "" {
+		// If only checksum is specified. Assume type is md5
+		return checksum, string(MD5), true
+	}
+	switch checksumType {
+	case MD5, SHA256, SHA512:
+		return checksum, string(checksumType), true
+	default:
+		return "", "", false
+	}
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
--- a/pkg/apis/metal3/v1alpha1/baremetalhost_types_test.go
+++ b/pkg/apis/metal3/v1alpha1/baremetalhost_types_test.go
@@ -529,3 +529,128 @@ func TestCredentialStatusMatch(t *testing.T) {
 		})
 	}
 }
+
+func TestGetImageChecksum(t *testing.T) {
+	for _, tc := range []struct {
+		Scenario string
+		Host     BareMetalHost
+		Expected bool
+	}{
+		{
+			Scenario: "both checksum value and type specified",
+			Host: BareMetalHost{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "myhost",
+					Namespace: "myns",
+				},
+				Spec: BareMetalHostSpec{
+					Image: &Image{
+						Checksum:     "md5hash",
+						ChecksumType: MD5,
+					},
+				},
+			},
+			Expected: true,
+		},
+		{
+			Scenario: "checksum value specified but not type",
+			Host: BareMetalHost{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "myhost",
+					Namespace: "myns",
+				},
+				Spec: BareMetalHostSpec{
+					Image: &Image{
+						Checksum: "md5hash",
+					},
+				},
+			},
+			Expected: true,
+		},
+		{
+			Scenario: "sha256 checksum value and type specified",
+			Host: BareMetalHost{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "myhost",
+					Namespace: "myns",
+				},
+				Spec: BareMetalHostSpec{
+					Image: &Image{
+						Checksum:     "sha256hash",
+						ChecksumType: SHA256,
+					},
+				},
+			},
+			Expected: true,
+		},
+		{
+			Scenario: "sha512 checksum value and type specified",
+			Host: BareMetalHost{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "myhost",
+					Namespace: "myns",
+				},
+				Spec: BareMetalHostSpec{
+					Image: &Image{
+						Checksum:     "sha512hash",
+						ChecksumType: SHA512,
+					},
+				},
+			},
+			Expected: true,
+		},
+		{
+			Scenario: "checksum value not specified",
+			Host: BareMetalHost{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "myhost",
+					Namespace: "myns",
+				},
+				Spec: BareMetalHostSpec{
+					Image: &Image{
+						ChecksumType: SHA512,
+					},
+				},
+			},
+			Expected: false,
+		},
+		{
+			Scenario: "neither checksum value nor hash specified",
+			Host: BareMetalHost{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "myhost",
+					Namespace: "myns",
+				},
+				Spec: BareMetalHostSpec{
+					Image: &Image{
+						URL: "someurl",
+					},
+				},
+			},
+			Expected: false,
+		},
+		{
+			Scenario: "wrong checksum hash specified",
+			Host: BareMetalHost{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "myhost",
+					Namespace: "myns",
+				},
+				Spec: BareMetalHostSpec{
+					Image: &Image{
+						Checksum:     "somehash",
+						ChecksumType: "boondoggle",
+					},
+				},
+			},
+			Expected: false,
+		},
+	} {
+		t.Run(tc.Scenario, func(t *testing.T) {
+			_, _, actual := tc.Host.GetImageChecksum()
+			if actual != tc.Expected {
+				t.Errorf("expected %v but got %v", tc.Expected, actual)
+			}
+		})
+	}
+}
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -279,12 +279,13 @@ func (p *ironicProvisioner) ValidateManagementAccess(credentialsChanged bool) (r
 			}
 		}
 
-		if p.host.Spec.Image != nil && p.host.Spec.Image.URL != "" {
-			checksum := p.host.Spec.Image.Checksum
+		checksum, checksumType, ok := p.host.GetImageChecksum()
 
+		if ok {
 			p.log.Info("setting instance info",
 				"image_source", p.host.Spec.Image.URL,
-				"checksum", checksum,
+				"image_os_hash_value", checksum,
+				"image_os_hash_algo", checksumType,
 			)
 
 			updates := nodes.UpdateOpts{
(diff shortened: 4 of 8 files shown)


pkg/provisioner/ironic/ironic.go
```diff
@@ -1295,3 +1324,23 @@ func (p *ironicProvisioner) PowerOff() (result provisioner.Result, err error) {
 
 	return result, nil
 }
+
+// getImageChecksum parses the checksum; splits and returns image_hash and image_value
```
dhellmann: It would be useful to see some example values in the comments here to understand what this function is trying to do when it parses its input. This function could also use some unit tests.

kirankt: > It would be useful to see some example values in the comments here to understand what this function is trying to do when it parses its input. This function could also use some unit tests.

+1. Will add examples and unit tests. Thanks for the feedback.

docs/api.md
```diff
@@ -50,8 +50,10 @@ mainly, but not only, provisioning details.
 
 * *image* -- Holds details for the image to be deployed on a given host.
   * *url* -- The URL of an image to deploy to the host.
-  * *checksum* -- The actual checksum or a URL to a file containing
-    the checksum for the image at *image.url*.
+  * *checksum* -- The actual md5 checksum or a URL to a file containing
+    the md5 checksum for the image at *image.url*. Other checksum algorithms
+    can be specified using the format: hash_algo:hash_value
+    e.g. `sha256:6623803...`, `sha256:https://...`, etc
```
zaneb: Can we add a list of the allowed algorithms? Is it just md5 and sha256?

pkg/provisioner/ironic/ironic.go
```diff
@@ -1295,3 +1323,28 @@ func (p *ironicProvisioner) PowerOff() (result provisioner.Result, err error) {
...
+// "458950b8cbebad6d437bd35d4d2bf66a" returns a default algo: "md5", and "458950b8cbebad6d437bd35d4d2bf66a"
+// "sha256:http://172.16.1.100/images/myOSv1/myOS.qcow2.sha256sum" returns "sha256" and "http://172.16.1.100/images/myOSv1/myOS.qcow2.sha256sum"
+// "http://172.16.1.100/images/myOSv1/myOS.qcow2.md5sum" returns "md5" and "http://172.16.1.100/images/myOSv1/myOS.qcow2.md5sum"
+
+func getImageChecksum(checksum string) (string, string) {
+	if checksum != "" {
+		// Check to see if the string is an URL. If so, just return
+		// the value presented
+		uri, err := url.Parse(checksum)
+		if err == nil && uri.Scheme != "" && uri.Host != "" {
```
zaneb: It's not obvious that this does the Right Thing in the case where data is of the form: `sha256:http://...` - obviously it works because there's a test, but mostly by coincidence.
If we had a list of allowable algorithms then we could just check for those prefixes.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1295,3 +1323,28 @@ func (p *ironicProvisioner) PowerOff() (result provisioner.Result, err error) {
 
 	return result, nil
 }
+
+// getImageChecksum parses the checksum and returns image_hash and image_value
```
zaneb: The Ironic fields are `image_hash_algo` and `image_hash_value`; shortening these (in different ways) is confusing and ambiguous.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1295,3 +1323,28 @@ func (p *ironicProvisioner) PowerOff() (result provisioner.Result, err error) {
...
+// e.g. "md5:458950b8cbebad6d437bd35d4d2bf66a" returns "md5" and "458950b8cbebad6d437bd35d4d2bf66a"
+// "458950b8cbebad6d437bd35d4d2bf66a" returns a default algo: "md5", and "458950b8cbebad6d437bd35d4d2bf66a"
+// "sha256:http://172.16.1.100/images/myOSv1/myOS.qcow2.sha256sum" returns "sha256" and "http://172.16.1.100/images/myOSv1/myOS.qcow2.sha256sum"
+// "http://172.16.1.100/images/myOSv1/myOS.qcow2.md5sum" returns "md5" and "http://172.16.1.100/images/myOSv1/myOS.qcow2.md5sum"
+
+func getImageChecksum(checksum string) (string, string) {
```
zaneb: I suddenly wonder if putting this in the Ironic provisioner is a mistake. It effectively makes this field a pass-through for Ironic, which is the thing we're trying to avoid by having a Provisioner interface.
Potentially this could be a method of v1alpha1.Image, that returns the algorithm from a defined list and either the hash value or a URL (and it should probably be explicit about which, since it's likely only Ironic that doesn't care).

docs/api.md
dhellmann: In other places we have tried to hiding behavior behind URL query arguments. In this case, the change would make the value into something that isn't a valid URL. Is there a reason to not simply add a new field to the API to hold the checksum algorithms? That would let us validate the checksum value and would make it easy to declare the allowed values using an enum.

pkg/provisioner/ironic/ironic.go
kirankt: +1. Will update.

docs/api.md
kirankt: It was suggested earlier by @hardys to re-use the existing API. Adding a new field would make things simpler and we don't have to worry about parsing the checkusm field.

pkg/provisioner/ironic/ironic.go
kirankt: If we add an extra filed to specify image_hash_algo along with image_hash_value, this will become moot.

kirankt:
I need your input for the following items:

1. What are your thoughts regarding adding a dedicated checkum algo field in spec.image. Having image.url, image.hash_algo and image.hash_value makes validation simpler while using image.checksum is tricky.

2. Is the value of image.checksum is just a hash or an URL or both. Ironic docs don't mention URL but the BMO doc does.

zaneb:
1. I am in favour of a separate field for the algorithm type.
2. It can be either a checksum or a URL to a checksum. The latter capability was added in Ironic only relatively recently; prior to that we were downloading the URL here in the BMO.

docs/api.md
kirankt: Since adding a dedicated filed under image, this point is now moot.

kirankt: I've added an unit test and have restricted the hash algos.

kirankt:
@dhellmann Would you mind taking another look at this PR, please?

pkg/apis/metal3/v1alpha1/baremetalhost_types.go
```diff
@@ -201,6 +202,10 @@ type Image struct {
 
 	// Checksum is the checksum for the image.
 	Checksum string `json:"checksum"`
+
+	// ChecksumType is the checksum algorithm for the image.
+	// e.g md5, sha256, sha512
+	ChecksumType string `json:"checksumType,omitempty"`
```
dhellmann, line 208: We should set up the kubebuilder enum validation rules for this field. The OperationalStatus and ErrorType fields show an example of how to do that. We should also declare the default for cases where the value is not provided.

pkg/apis/metal3/v1alpha1/baremetalhost_types.go
```diff
@@ -740,6 +745,26 @@ func (host *BareMetalHost) OperationMetricForState(operation ProvisioningState)
...
+		// Return empty if checksum is not provided
+		return "", ""
+	}
+	if hashAlgo == "" {
+		// If only checksum is specified. Assume type is md5
+		return hashValue, "md5"
+	}
+	switch hashAlgo {
+	case "md5", "sha256", "sha512":
+		return hashValue, hashAlgo
```
dhellmann, line 763: We should report an explicit error here instead of appearing to ignore the inputs. I think that means this method needs to return an error in addition to the 2 strings.

pkg/apis/metal3/v1alpha1/baremetalhost_types.go
```diff
@@ -740,6 +745,26 @@ func (host *BareMetalHost) OperationMetricForState(operation ProvisioningState)
 	return
 }
 
+// GetImageChecksum returns the hash value and its algo.
+func (host *BareMetalHost) GetImageChecksum() (string, string) {
```
dhellmann, line 749: If you name the return arguments that will make it clearer to a caller which value is which.

pkg/apis/metal3/v1alpha1/baremetalhost_types_test.go
```diff
@@ -529,3 +529,128 @@ func TestCredentialStatusMatch(t *testing.T) {
...
+			Scenario: "both checksum value and type specified",
+			Host: BareMetalHost{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "myhost",
+					Namespace: "myns",
+				},
+				Spec: BareMetalHostSpec{
+					Image: &Image{
+						Checksum:     "md5hash",
+						ChecksumType: "md5",
```
zaneb, line 549: Prefer to use the constants here.

pkg/apis/metal3/v1alpha1/baremetalhost_types.go
kirankt: Fixed in the new updates.

pkg/apis/metal3/v1alpha1/baremetalhost_types_test.go
kirankt: Thanks. Making the change. It does look nicer as a constant.

pkg/apis/metal3/v1alpha1/baremetalhost_types.go
```diff
@@ -201,6 +201,17 @@ type BareMetalHostSpec struct {
 	ExternallyProvisioned bool `json:"externallyProvisioned,omitempty"`
 }
 
+const (
+	// MD5 checksum type
+	MD5 string = "md5"
```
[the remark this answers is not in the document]
kirankt: So... I did have it the way you suggested but I ran into issues with this option. Unit tests (ironic_test) were failing unless I cast the ChecksumType back to string explicitly...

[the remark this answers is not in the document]
zaneb: I think it makes sense to have casts to string on line 967 and 971 of `startProvisioning()`, which I assume is where the problem is.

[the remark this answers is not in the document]
kirankt: I can cast it in the GetImageChecksum function but the problem still persists in the ironic unit tests:
`
go test  ./cmd/... ./pkg/...
?   	github.com/metal3-io/baremetal-operator/cmd/make-bm-worker	[no test files]
ok  	github.com/metal3-io/baremetal-operator/cmd/make-bm-worker/templates	(cached)
?   	github.com/metal3-io/baremetal-operator/cmd/make-virt-host	[no test files]
?   	github.com/metal3-io/baremetal-operator/cmd/manager	[no test files]
?   	github.com/metal3-io/baremetal-operator/pkg/apis	[no test files]
ok  	github.com/metal3-io/baremetal-operator/pkg/apis/metal3/v1alpha1	(cached)
ok  	github.com/metal3-io/baremetal-operator/pkg/bmc	(cached)
?   	github.com/metal3-io/baremetal-operator/pkg/controller	[no test files]
ok  	github.com/metal3-io/baremetal-operator/pkg/controller/baremetalhost	(cached)
?   	github.com/metal3-io/baremetal-operator/pkg/hardware	[no test files]
?   	github.com/metal3-io/baremetal-operator/pkg/provisioner	[no test files]
?   	github.com/metal3-io/baremetal-operator/pkg/provisioner/demo	[no test files]
?   	github.com/metal3-io/baremetal-operator/pkg/provisioner/fixture	[no test files]
2020-05-05T15:02:53.244-0500	INFO	baremetalhost_ironic	adding image_source	{"host": "myhost"}
2020-05-05T15:02:53.244-0500	INFO	baremetalhost_ironic	adding image_os_hash_algo	{"host": "myhost"}
2020-05-05T15:02:53.244-0500	INFO	baremetalhost_ironic	adding image_os_hash_value	{"host": "myhost"}
2020-05-05T15:02:53.244-0500	INFO	baremetalhost_ironic	setting instance_uuid	{"host": "myhost"}
2020-05-05T15:02:53.244-0500	INFO	baremetalhost_ironic	adding root_gb	{"host": "myhost"}
2020-05-05T15:02:53.244-0500	INFO	baremetalhost_ironic	adding root_device	{"host": "myhost"}
2020-05-05T15:02:53.244-0500	INFO	baremetalhost_ironic	using root device	{"host": "myhost", "hints": {"name":"/dev/vda"}}
2020-05-05T15:02:53.244-0500	INFO	baremetalhost_ironic	adding cpu_arch	{"host": "myhost"}
2020-05-05T15:02:53.244-0500	INFO	baremetalhost_ironic	adding local_gb	{"host": "myhost"}
--- FAIL: TestGetUpdateOptsForNodeVirtual (0.00s)
    ironic_test.go:54: patches: [{add /instance_info/image_source not-empty} {add /instance_info/image_os_hash_algo md5} {add /instance_info/image_os_hash_value checksum} {replace /instance_uuid 27720611-e5d1-45d3-ba3a-222dcfaa4ca2} {add /instance_info/root_gb 10} {add /properties/root_device map[name:/dev/vda]} {add /properties/cpu_arch x86_64} {add /properties/local_gb 50}]
    ironic_test.go:107: expected /instance_info/image_os_hash_algo="md5" got /instance_info/image_os_hash_algo="md5"
2020-05-05T15:02:53.244-0500	INFO	baremetalhost_ironic	adding image_source	{"host": "myhost"}
2020-05-05T15:02:53.244-0500	INFO	baremetalhost_ironic	adding image_os_hash_algo	{"host": "myhost"}
2020-05-05T15:02:53.244-0500	INFO	baremetalhost_ironic	adding image_os_hash_value	{"host": "myhost"}
2020-05-05T15:02:53.244-0500	INFO	baremetalhost_ironic	setting instance_uuid	{"host": "myhost"}
2020-05-05T15:02:53.244-0500	INFO	baremetalhost_ironic	adding root_gb	{"host": "myhost"}
2020-05-05T15:02:53.244-0500	INFO	baremetalhost_ironic	adding root_device	{"host": "myhost"}
2020-05-05T15:02:53.244-0500	INFO	baremetalhost_ironic	using root device	{"host": "myhost", "hints": {"hctl":"0:0:0:0"}}
2020-05-05T15:02:53.244-0500	INFO	baremetalhost_ironic	adding cpu_arch	{"host": "myhost"}
2020-05-05T15:02:53.244-0500	INFO	baremetalhost_ironic	adding local_gb	{"host": "myhost"}
--- FAIL: TestGetUpdateOptsForNodeDell (0.00s)
    ironic_test.go:146: patches: [{add /instance_info/image_source not-empty} {add /instance_info/image_os_hash_algo md5} {add /instance_info/image_os_hash_value checksum} {replace /instance_uuid 27720611-e5d1-45d3-ba3a-222dcfaa4ca2} {add /instance_info/root_gb 10} {add /properties/root_device map[hctl:0:0:0:0]} {add /properties/cpu_arch x86_64} {add /properties/local_gb 50}]
    ironic_test.go:199: expected /instance_info/image_os_hash_algo="md5" got /instance_info/image_os_hash_algo="md5"
FAIL
FAIL	github.com/metal3-io/baremetal-operator/pkg/provisioner/ironic	0.016s
?   	github.com/metal3-io/baremetal-operator/pkg/utils	[no test files]
?   	github.com/metal3-io/baremetal-operator/pkg/version	[no test files]
FAIL
make: *** [Makefile:59: unit] Error 1

`

PR #405, opened 2020-01-23
Makefile updates for docker build and push, and release manifests generation.

Arvinderpal:
Changes introduced: 
 * Docker build and push commands that align with CAPI model.
 * Release manifests generation.

This is the first PR that brings us closer to having release model similar to that of CAPI. See Issue: https://github.com/metal3-io/baremetal-operator/issues/404

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,19 @@ export DEPLOY_RAMDISK_URL=http://172.22.0.1:6180/images/ironic-python-agent.init
 export IRONIC_ENDPOINT=http://localhost:6385/v1/
 export IRONIC_INSPECTOR_ENDPOINT=http://localhost:5050/v1/
 
+# Define Docker related variables. Releases should modify and double check these vars.
+# REGISTRY ?= gcr.io/$(shell gcloud config get-value project)
+REGISTRY ?= quay.io/metal3-io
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -134,3 +147,63 @@ dep-status:
...
+
+.PHONY: docker-build
+docker-build: ## Build the docker image for controller-manager
+	docker build --pull --build-arg ARCH=$(ARCH) -f build/Dockerfile -t $(CONTROLLER_IMG)-$(ARCH):$(TAG) .
+	MANIFEST_IMG=$(CONTROLLER_IMG)-$(ARCH) MANIFEST_TAG=$(TAG) $(MAKE) set-manifest-image
+	$(MAKE) set-manifest-pull-policy
+
+.PHONY: docker-push
+docker-push: ## Push the docker image
+	docker push $(CONTROLLER_IMG)-$(ARCH):$(TAG)
```
maelk, line 163: for this to work with quay.io, we would need to create a new repository on quay.io since the image would be renamed from baremetal-operator to baremetal-operator-amd64. If that is the only architecture supported for now, I don't really see the point of renaming though. That could be done when other architectures are introduced.

PR #407, opened 2020-01-30
Add CurrentBootMode parameter in metal3.io_baremetalhosts_crd.yaml

sharajat:
This implementation is based on PR - https://github.com/metal3-io/metal3-docs/pull/66


dhellmann:
I'm not comfortable taking any more API changes until we have #434 fixed. We shouldn't have to edit the CRD file by hand, and we should be able to enable the new generator test job to validate API changes.

dhellmann:
Is this still useful now that we have the field to set the boot mode? Anything discovered by inspection would be reset to that value, which is also being reflected in the status.

zaneb:
> Is this still useful now that we have the field to set the boot mode?

That's what we [always needed](https://github.com/metal3-io/metal3-docs/pull/66#discussion_r383637558) IMHO.

The design proposal PR has been closed for a while, and it has been obsoleted by other work, so I think we can close this.
/close

PR #412, opened 2020-02-03
Attempt to power off BareMetalHost before delete

### LINKED ISSUE #410: Host should be powered down once deleted

sadasu:
When a BareMetalHost is deleted, power it off before performing
the delete operation.
Fixes #410

### DIFF
--- a/pkg/provisioner/ironic/delete_test.go
+++ b/pkg/provisioner/ironic/delete_test.go
@@ -22,6 +22,7 @@ func TestDelete(t *testing.T) {
 		ironic    *testserver.IronicMock
 		inspector *testserver.InspectorMock
 		hostName  string
+		powerOn   bool
 
 		expectedDirty        bool
 		expectedRequestAfter time.Duration
@@ -143,6 +144,27 @@ func TestDelete(t *testing.T) {
 				Value: true,
 			},
 		},
+		{
+			name: "test-power-on",
+			ironic: testserver.NewIronic(t).Node(
+				nodes.Node{
+					UUID:           nodeUUID,
+					ProvisionState: "active",
+					Maintenance:    false,
+					PowerState:     powerOn,
+				},
+			).NodeUpdate(nodes.Node{
+				UUID: nodeUUID,
+			}),
+			powerOn:              true,
+			expectedDirty:        true,
+			expectedRequestAfter: 0,
+			expectedUpdate: &nodes.UpdateOperation{
+				Op:    "replace",
+				Path:  "/maintenance",
+				Value: true,
+			},
+		},
 	}
 
 	for _, tc := range cases {
@@ -170,6 +192,13 @@ func TestDelete(t *testing.T) {
 			)
 			if err != nil {
 				t.Fatalf("could not create provisioner: %s", err)
+			}
+			if tc.powerOn {
+				prov.status.ID = nodeUUID
+				_, err = prov.PowerOn()
+				if err != nil {
+					t.Errorf("could not power on server: %s", err)
+				}
 			}
 
 			result, err := prov.Delete()
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -55,6 +55,10 @@ const (
 	powerOff     = "power off"
 	softPowerOff = "soft power off"
 	powerNone    = "None"
+
+	// Maximum number of times node Power off would be retried before
+	// it would be deleted.
+	maxPowerOffRetryCount = 3
 )
 
 var bootModeCapabilities = map[metal3v1alpha1.BootMode]string{
@@ -1611,6 +1615,20 @@ func (p *ironicProvisioner) Delete() (result provisioner.Result, err error) {
 		// Nova.
 		p.log.Info("setting host maintenance flag to force image delete")
 		return p.setMaintenanceFlag(ironicNode, true)
+	}
+
+	if ironicNode.PowerState == powerOn && p.host.Status.ErrorCount <= maxPowerOffRetryCount {
+		p.log.Info("host ready to be powered off")
+		_, err := p.hardPowerOff()
+		if err != nil {
+			switch err.(type) {
+			case HostLockedError:
+				p.log.Info("could not power off host, busy")
+				return retryAfterDelay(powerRequeueDelay)
+			default:
+				return operationFailed("failed to power off host")
+			}
+		}
 	}
 
 	p.log.Info("host ready to be removed")
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -446,6 +446,9 @@ func (r *BareMetalHostReconciler) actionDeleting(prov provisioner.Provisioner, i
 	if provResult.Dirty {
 		return actionContinue{provResult.RequeueAfter}
 	}
+	if provResult.ErrorMessage != "" {
+		return recordActionFailure(info, metal3v1alpha1.PowerManagementError, provResult.ErrorMessage)
+	}
 
 	// Remove finalizer to allow deletion
 	info.host.Finalizers = utils.FilterStringFromList(


pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -367,6 +367,14 @@ func (r *ReconcileBareMetalHost) actionDeleting(prov provisioner.Provisioner, in
 		}
 		return actionContinue{provResult.RequeueAfter}
 	}
+	// try to power off host after delete
+	provResult, err = prov.PowerOff()
```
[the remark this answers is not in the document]
sadasu: Done. 
Note : My impression was that the prov.Delete() was a "logical" operation on the node and the prov.PowerOff() was a "physical" operation and the correct order would be to perform the logical Delete() before the "physical" PowerOff().

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -367,6 +367,14 @@ func (r *ReconcileBareMetalHost) actionDeleting(prov provisioner.Provisioner, in
...
+	if err != nil {
+		return actionError{errors.Wrap(err, "failed to power off host before delete")}
+	}
+	if provResult.ErrorMessage != "" {
+		return recordActionFailure(info, metal3v1alpha1.PowerManagementError, provResult.ErrorMessage)
```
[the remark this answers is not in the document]
sadasu: I have modified logic here to use the "Dirty" flag to keep trying to PowerOff the host. Also, I am not sure if we need to save status of the node when the Dirty flag is set. 
Error handling during the PowerOff stage is also probably not in its final state. So, leaving this in WIP state.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -355,8 +355,21 @@ func (r *ReconcileBareMetalHost) actionDeleting(prov provisioner.Provisioner, in
 		info.log.Info("ready to be deleted")
 		return deleteComplete{}
 	}
+	// try to power off host before delete
+	provResult, err := prov.PowerOff()
+	if err != nil {
+		info.log.Info("failed to power off host before delete")
+	}
+	if provResult.Dirty {
+		err = r.saveStatus(info.host)
```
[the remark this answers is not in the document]
sadasu: > There's actually no need to call `saveStatus()` here; it gets saved automatically as a result of returning actionContinue.
> I should have removed the call on line 377 in or around [c625c52](https://github.com/metal3-io/baremetal-operator/commit/c625c521648fba50577096ba58f86fbd19a48901) but I overlooked it.

Do you want to remove the saveStatus() after the Delete() in this PR?

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -355,8 +355,21 @@ func (r *ReconcileBareMetalHost) actionDeleting(prov provisioner.Provisioner, in
 		info.log.Info("ready to be deleted")
 		return deleteComplete{}
 	}
+	// try to power off host before delete
+	provResult, err := prov.PowerOff()
+	if err != nil {
+		info.log.Info("failed to power off host before delete")
```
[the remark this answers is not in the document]
sadasu: Done. Thanks for highlighting the difference between actionError and actionFailed.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -384,13 +384,24 @@ func (r *ReconcileBareMetalHost) actionDeleting(prov provisioner.Provisioner, in
 		info.log.Info("ready to be deleted")
 		return deleteComplete{}
 	}
+	// try to power off host before delete
+	provResult, err := prov.PowerOff()
+	if err != nil {
+		return actionError{errors.Wrap(err, "failed to power off host before delete")}
+	}
+	if provResult.Dirty {
+		if err != nil {
```
zaneb, line 393: Code inside this if statement is unreachable.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -384,13 +384,24 @@ func (r *ReconcileBareMetalHost) actionDeleting(prov provisioner.Provisioner, in
...
 
-	provResult, err := prov.Delete()
+	// proceed with the delete
+	provResult, err = prov.Delete()
 	if err != nil {
 		return actionError{errors.Wrap(err, "failed to delete")}
 	}
 	if provResult.Dirty {
-		err = r.saveStatus(info.host)
 		if err != nil {
```
zaneb, line 405: This is now unreachable also, since it was only here to check the result of saveStatus().

zaneb:
I opened a separate PR (#514) to clean up the existing problem with `saveStatus`, because it turns out it might be related to a bug so that's probably a good reason to keep it separate.

dhellmann:
@sadasu are you still working on this PR?

sadasu:
> @sadasu are you still working on this PR?

Yes, trying to understand PR (#514) and modify implementation here.

dhellmann:
> > @sadasu are you still working on this PR?
> 
> Yes, trying to understand PR (#514) and modify implementation here.

Sounds good. There's no particular rush, I was looking at some PRs that I might be able to help with rebasing or something to move forward. Little stuff for a Friday afternoon. :-)

sadasu:
@dhellmann could you please take a look? I need an approve label.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -432,8 +432,16 @@ func (r *ReconcileBareMetalHost) actionDeleting(prov provisioner.Provisioner, in
 		info.log.Info("ready to be deleted")
 		return deleteComplete{}
 	}
-
-	provResult, err := prov.Delete()
+	// try to power off host before delete
+	provResult, err := prov.PowerOff()
```
dhellmann, line 436: What happens here if the BMC credentials are invalid? Do we sit in a reconcile loop forever until they're fixed, and then the host is deleted?

sadasu: If we were able to provision the host earlier, does that not imply that we have the correct BMC credentials?

dhellmann: They could have been changed out from under us, on the device itself. That would mean Ironic would report an error trying to turn off the power, and here we now always treat that error as something that needs a retry. I wonder if we want to limit the number of retries and at some point just ignore the error? Or maybe move the power off operation in the Delete() method of the provisioner, where it can behave more intelligently based on the specific error .

zaneb: We could call `ValidateManagementAccess()` and only try the power off if it succeeds. There's a risk that a temporary failure could prevent the power being turned off, but I think we're mainly doing this to make testing easier anyway so maybe it's not a big deal.

dhellmann: That might work.

zaneb:
On closer inspection, it looks like the provisioner code doesn't actually check for a failed attempt to power off (it doesn't look at any error messages from Ironic), so I suspect we will just end up retrying forever if the credentials don't work.
In addition, if we can find the host in Ironic (possibly due to a previous call to `actionDelete()` removing it) then there'll be an error that sends us into an infinite retry loop.
There's no current way to distinguish this error from other errors (other than looking at the text), and in any event it might be wise to just (log +) ignore all errors from this call anyway.

/lgtm cancel

pkg/controller/baremetalhost/baremetalhost_controller.go
sadasu: I am concerned about any unintended consequences from https://github.com/metal3-io/baremetal-operator/blob/master/pkg/provisioner/ironic/ironic.go#L384 being called multiple times in the delete path via ValidateManagementAccess().
If https://github.com/metal3-io/baremetal-operator/blob/master/pkg/provisioner/ironic/ironic.go#L228 is the only one that is actually verifying BMC credentials, the why not call just that from PowerOff() before attempting a poweroff and give a better error message about what exactly failed.

sadasu: To make the code more efficient, I would like to propose this:
1. Call poweroff()
2. if previous call returns error, then check for credentials
3. If credentials are bad, ignore all poweroff() errors, report issue with bad credentials, proceed to Delete()
4. If credentials are good [and poweroff() still failed], we could call actionError() [which will result in retry]
5. Delete()

Assuming that credentials changing underneath us is a fairly infrequent occurance, this way we make the expensive call to the BMC to verify credentials only in the case of an error.

zaneb: Checking the credentials doesn't necessarily happen in a single call, it's something that could occur over the course of several requeues. Despite that, this might still work though because PowerOff() doesn't change the state.
However, we'd still have to deal with the case where we get up to calling Delete(), and it succeeds but we fail when trying to remove the finaliser. If that happens, on the next reconcile we will call PowerOff() and it will fail to load the node from Ironic, and from that point we will be in an infinite loop.

I think we need to:
* Set an ErrorMessage in PowerOff() when a previous attempt to power off the host has failed.
* Ignore both the Dirty flag and the ErrorMessage if we get an ErrorMessage from PowerOff().
* Ignore any err returned from PowerOff().

sadasu: Reading this comment https://github.com/metal3-io/baremetal-operator/blob/master/pkg/provisioner/ironic/ironic.go#L1117-L1120 has spawned a related question. Do we have to make sure the node is not externally provisioned before powering it off?

dhellmann: Excellent question, @sadasu. What behavior would you expect?

sadasu: We should not be powering off that node. But, what I am unclear about is whether the responsibility for checking if the node is externally provisioned lies with the  baremetal controller with actionDeleting() or with the powerOff() logic in the provisioner (Ironic). Since the provisioner is already aware if a node is externally provisioned, I think that it should be the provisioner that makes sure it doesn't power off a node that it did not provision.

dhellmann: Another good question. I think the context matters for the answer to that one.

The provisioner does know whether a host is externally provisioned. However, it doesn't know inside PowerOff() what the context for that request is. For example, it doesn't know if the host is being deprovisioned or put into maintenance mode. I could see us wanting to support maintenance mode for externally provisioned hosts, so I think we want to make the power control decision outside of the provisioner.

zaneb: Great catch Sandhya.
I agree with Doug, IMHO it should be the controller that decides not to power off an externally-provisioned node. If we had another provisioner implementation we wouldn't want it to have different behaviour, so it shouldn't be left up to the provisioner. And to the extent possible the provisioner methods shouldn't behave differently depending on the host state; it's the job of the controller to manage state.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -432,7 +432,11 @@ func (r *ReconcileBareMetalHost) actionDeleting(prov provisioner.Provisioner, in
 		info.log.Info("ready to be deleted")
 		return deleteComplete{}
 	}
-
+	// Try to power off host before delete except when it is externally provisioned.
+	if info.host.Status.Provisioning.State != metal3v1alpha1.StateExternallyProvisioned {
+		_, _ = prov.PowerOff()
```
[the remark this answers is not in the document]
zaneb: Oh, and we need to change the provisioner code to set an ErrorMessage when we've attempted to power off and it has failed/timed out, so we don't just try again.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -432,7 +432,11 @@ func (r *ReconcileBareMetalHost) actionDeleting(prov provisioner.Provisioner, in
 		info.log.Info("ready to be deleted")
 		return deleteComplete{}
 	}
-
+	// Try to power off host before delete except when it is externally provisioned.
+	if info.host.Status.Provisioning.State != metal3v1alpha1.StateExternallyProvisioned {
```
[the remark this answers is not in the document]
sadasu: Done.

pkg/controller/baremetalhost/baremetalhost_controller.go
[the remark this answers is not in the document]
sadasu: I think the comments here are taken care of.

pkg/controller/baremetalhost/baremetalhost_controller.go
sadasu: I think this is done too.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -447,7 +447,21 @@ func (r *ReconcileBareMetalHost) actionDeleting(prov provisioner.Provisioner, in
...
+		provResult, err := prov.PowerOff()
+		switch {
+		case err != nil:
+			log.Error(err, fmt.Sprintf("Likely transient error while trying to Power Off Baremetal Host"))
+		case provResult.ErrorMessage != "":
+			log.Error(err, fmt.Sprintf("Error encountered while trying to Power Off Baremetal Host"))
+		case provResult.Dirty:
+			return actionContinue{provResult.RequeueAfter}
+		default:
+			info.host.Status.PoweredOn = false
```
dhellmann, line 461: Should we return here to store the status update? Or are we relying on `prov.Delete()` returning dirty?

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -447,7 +447,21 @@ func (r *ReconcileBareMetalHost) actionDeleting(prov provisioner.Provisioner, in
 		info.log.Info("ready to be deleted")
 		return deleteComplete{}
 	}
-
+	// Try to power off host before delete except when it is externally provisioned.
+	if !info.host.Spec.ExternallyProvisioned {
+		provResult, err := prov.PowerOff()
```
dhellmann, line 452: Do we need to check to see if the host is already powered off by looking at our status field?

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -447,7 +447,21 @@ func (r *ReconcileBareMetalHost) actionDeleting(prov provisioner.Provisioner, in
...
+		switch {
+		case err != nil:
+			log.Error(err, fmt.Sprintf("Likely transient error while trying to Power Off Baremetal Host"))
+		case provResult.ErrorMessage != "":
+			log.Error(err, fmt.Sprintf("Error encountered while trying to Power Off Baremetal Host"))
+		case provResult.Dirty:
+			return actionContinue{provResult.RequeueAfter}
```
dhellmann, line 459: If we're going to ignore the errors, it would be good for the messages to say so. And maybe not to log at error level since the user won't have a chance to do anything about the problem.

pkg/controller/baremetalhost/baremetalhost_controller.go
sadasu: Done.

pkg/controller/baremetalhost/baremetalhost_controller.go
sadasu: We are going to ignore the errors but maybe they are useful to the customer and also might prove to be valuable brea crumbs when we are trying to debug a more complex failure?

pkg/controller/baremetalhost/baremetalhost_controller.go
sadasu: Not sure. But, my thinking is that we should not return before attemping a prov.Delete() unless it is a general failure.

sadasu:
> After talking this through today, I think we want to move the power off call into the provisioner's `Delete()` method. At that point we can use the `hardPowerOff()` method of the `IronicProvisioner` to avoid delays from the soft power off timing out. That also avoids having to update the `Provisioner` API to have `PowerOff()` take an argument to control whether to skip that soft power off.

Changed implementation to match result of discussion mentioned above.

pkg/controller/baremetalhost/baremetalhost_controller.go
sadasu: Setting Dirty to true and returning here to save power status change.

pkg/controller/baremetalhost/baremetalhost_controller.go
sadasu: Done.

pkg/controller/baremetalhost/baremetalhost_controller.go
sadasu: Done.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1265,6 +1265,25 @@ func (p *ironicProvisioner) Delete() (result provisioner.Result, err error) {
 		return p.setMaintenanceFlag(ironicNode, true)
 	}
 
+	if p.host.Status.PoweredOn {
+		p.log.Info("host is being powered off before removal")
+		provResult, err := p.hardPowerOff()
+		switch {
+		case err != nil:
```
dhellmann, line 1272: Do we care about 409 errors telling us Ironic is busy doing something with the host?

pkg/provisioner/ironic/ironic.go
```diff
@@ -1399,6 +1418,13 @@ func (p *ironicProvisioner) hardPowerOff() (result provisioner.Result, err error
 			result.Dirty = true
 			return result, nil
 		}
+		// If the target state is unset while the last error is set,
+		// then the last execution of hard power off has failed.
```
zaneb, line 1422: Won't this also be the case if the last target state was soft power off and execution of that failed? If so then this will break the fallback from soft power off to hard power off in regular operation.

pkg/provisioner/ironic/ironic.go
dhellmann: I wonder if Ironic will let us delete a host if it thinks it's busy? That error would happen (and could be handled) on the delete call, so I guess the worst thing that might happen is we do successfully delete a host without powering it off.

dhellmann:

/hold

dhellmann:
#655 has merged

/hold cancel

pkg/provisioner/ironic/ironic.go
sadasu: hardPowerOff() internally uses changePower() which already takes care of 409 errors.

pkg/provisioner/ironic/ironic.go
sadasu: Agreed. This is probably unnecessary.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1336,6 +1336,25 @@ func (p *ironicProvisioner) Delete() (result provisioner.Result, err error) {
...
+	if p.host.Status.PoweredOn {
+		p.log.Info("host is being powered off before removal")
+		provResult, err := p.hardPowerOff()
+		switch {
+		case err == nil:
+			p.host.Status.PoweredOn = false
+			// Save power status change by marking result as Dirty
+			result.Dirty = true
+			return result, nil
+		case provResult.Dirty:
```
dhellmann, line 1348: I don't think we'll ever get here, because if there is an error (failing the first case) then Dirty is unlikely to be true.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1336,6 +1336,25 @@ func (p *ironicProvisioner) Delete() (result provisioner.Result, err error) {
 		return p.setMaintenanceFlag(ironicNode, true)
 	}
 
+	if p.host.Status.PoweredOn {
+		p.log.Info("host is being powered off before removal")
+		provResult, err := p.hardPowerOff()
+		switch {
+		case err == nil:
+			p.host.Status.PoweredOn = false
```
dhellmann, line 1344: Don't we want to wait for ironic to actually power the host off? That might take a few seconds, during which we need to keep trying. If we set PoweredOn to false here, then the check on 1339 will fail and we'll delete the host on the next pass through.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1336,6 +1336,25 @@ func (p *ironicProvisioner) Delete() (result provisioner.Result, err error) {
 		return p.setMaintenanceFlag(ironicNode, true)
 	}
 
+	if p.host.Status.PoweredOn {
```
dhellmann, line 1339: Should this be looking at the `ironicNode.PowerState` value instead of the version of that copied to the host?

pkg/provisioner/ironic/ironic.go
sadasu: https://github.com/metal3-io/baremetal-operator/blob/master/pkg/provisioner/ironic/ironic.go#L1471 seems to be one place where err == nil and Dirty == true.

dhellmann: Right, but if there is an error then err won't be nil.

pkg/provisioner/ironic/ironic.go
sadasu: Done.

pkg/provisioner/ironic/ironic.go
sadasu: I think this is fixed also.

pkg/provisioner/ironic/ironic.go
sadasu: This code has been restructured.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1336,6 +1340,34 @@ func (p *ironicProvisioner) Delete() (result provisioner.Result, err error) {
...
+				result.RequeueAfter = powerRequeueDelay
+				result.Dirty = true
+				return result, err
+			default:
+				if result.ErrorCount < maxErrorRetryCount {
+					result.ErrorCount++
+					result.Dirty = true
+					return result, err
+				}
+				result.Dirty = false
```
dhellmann, line 1358: This is the default value. Are you resetting it because it was changed somewhere else?

pkg/provisioner/ironic/ironic.go
```diff
@@ -1336,6 +1340,34 @@ func (p *ironicProvisioner) Delete() (result provisioner.Result, err error) {
...
+			}
+
+		} else {
+			// No errors returned by hardPowerOff
+			if provResult.Dirty {
+				result.RequeueAfter = powerRequeueDelay
+				result.Dirty = true
+				return result, err
+			}
+			result.Dirty = false
```
dhellmann, line 1368: I think this is probably not needed, either.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1336,6 +1340,34 @@ func (p *ironicProvisioner) Delete() (result provisioner.Result, err error) {
...
+		p.log.Info("host is being powered off before removal")
+		provResult, err := p.hardPowerOff()
+		if err != nil {
+			switch err.(type) {
+			case HostLockedError:
+				result.RequeueAfter = powerRequeueDelay
+				result.Dirty = true
+				return result, err
+			default:
+				if result.ErrorCount < maxErrorRetryCount {
```
dhellmann, line 1353: I don't think this is going to do what you expect. This data structure is not persisted anywhere, so the count will never grow higher than 1, and it will never be more than 0 at this point because we will always have a fresh result structure.

I think instead of adding an ErrorCount value to the result structure, you want to look at the host.Status.ErrorCount. That is maintained by the controller when it sees an error from one of the methods while it is reconciling the host.

Maybe @andfasano can confirm that would work.

s3rj1k:
Any updates on this?

pkg/provisioner/provisioner.go
```diff
@@ -99,4 +99,7 @@ type Result struct {
 	RequeueAfter time.Duration
 	// Any error message produced by the provisioner.
 	ErrorMessage string
+	// ErrorCount to keep track of the number of times an operation
+	// has encountered errors.
+	ErrorCount int
```
sadasu, line 104: So, based on this https://github.com/metal3-io/baremetal-operator/pull/412/commits/e87a2e8063b2ccd4e60a393d7684457389139c51#r522255159 comment, it is no longer necessary to add this ErrorCount here.

pkg/provisioner/ironic/ironic.go
andfasano: Sorry for the late answer, but @dhellmann is correct, as the business logic should be based on the `host.Status.ErrorCount`. The provisioner then can decide how to proceed by using one of the semantic methods introduced by @zaneb in the [result.go](https://github.com/metal3-io/baremetal-operator/blob/ea587bb61a93547cebd1483be5852d1104b35d33/pkg/provisioner/ironic/result.go) - they improve also the code readability.

sadasu: Yeah, I had wrongly understood result to be preserved across subsequent calls to the same method. I also see that the code around error handling has evolved considerably. So, starting afresh to align with these changes.

pkg/provisioner/ironic/ironic.go
sadasu: Again, this line to reset the dirty flag was added because I thought we were preserving the result structure during subsequent calls to the same method. Removing it since a new result structure is used for each call.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1613,6 +1617,33 @@ func (p *ironicProvisioner) Delete() (result provisioner.Result, err error) {
...
 	}
 
+	if ironicNode.PowerState == powerOn {
+		_, err := p.hardPowerOff()
+		if err != nil {
+			result.Dirty = true
+			switch err.(type) {
+			case HostLockedError:
+				p.log.Info("could not power off host, busy")
+				return retryAfterDelay(powerRequeueDelay)
```
andfasano, line 1627: The `retryAfterDelay` is currently equivalent to an operation continue. Is it ok to manage this case as not an error that could increase the error count (see below)?

pkg/provisioner/ironic/ironic.go
```diff
@@ -1613,6 +1617,33 @@ func (p *ironicProvisioner) Delete() (result provisioner.Result, err error) {
...
+			default:
+				// For other errors continue with the delete
+				if p.host.Status.ErrorCount > maxErrorRetryCount {
+					result.Dirty = false
+					p.host.Status.ErrorCount = 0
+					p.host.Status.ErrorMessage = ""
+				} else {
+					p.host.Status.ErrorCount++
+					p.host.Status.ErrorMessage = "failed to power off host"
+					return transientError(errors.Wrap(err, "failed to power off host"))
```
andfasano, line 1637: `transientError` does not activate the retry mechanism with backoff, but `operationFailed()` is required instead. In addition to that, `BareMetalHostReconciler.actionDeleting` will require to be changed in order to manage properly the case when `provResult.ErrorMessage` is not empty so that `recordActionFailure` could be invoked.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1613,6 +1617,33 @@ func (p *ironicProvisioner) Delete() (result provisioner.Result, err error) {
...
+			switch err.(type) {
+			case HostLockedError:
+				p.log.Info("could not power off host, busy")
+				return retryAfterDelay(powerRequeueDelay)
+			default:
+				// For other errors continue with the delete
+				if p.host.Status.ErrorCount > maxErrorRetryCount {
+					result.Dirty = false
+					p.host.Status.ErrorCount = 0
+					p.host.Status.ErrorMessage = ""
```
andfasano, line 1633: Is not anymore allowed to modify an host from within the provisioner (a copy is passed in the Provisioner factory). If the expected behavior in this case it's to give an `operationComplete()` should be returned and a proper log message traced

pkg/provisioner/ironic/ironic.go
```diff
@@ -1613,6 +1617,33 @@ func (p *ironicProvisioner) Delete() (result provisioner.Result, err error) {
...
+				} else {
+					p.host.Status.ErrorCount++
+					p.host.Status.ErrorMessage = "failed to power off host"
+					return transientError(errors.Wrap(err, "failed to power off host"))
+				}
+			}
+		} else {
+			p.log.Info("host powered off before removal")
+			p.host.Status.ErrorCount = 0
+			p.host.Status.ErrorMessage = ""
```
andfasano, line 1643: See above comments

pkg/provisioner/ironic/ironic.go
sadasu: Yes, my understanding is that we would not want to classify the scenario where the node is busy and cannot currently be powered off as an error.

sadasu: Fixed.

pkg/provisioner/ironic/ironic.go
```diff
@@ -55,6 +55,10 @@ const (
 	powerOff     = "power off"
 	softPowerOff = "soft power off"
 	powerNone    = "None"
+
+	// Maximum number of retries an operation returning error
+	// would be retried.
+	maxErrorRetryCount = 3
```
andfasano, line 61: The name doesn't seem correct, as it is a threshold currently adopted for limiting the number of retries just for powering off the host

pkg/provisioner/ironic/ironic.go
```diff
@@ -1613,6 +1617,30 @@ func (p *ironicProvisioner) Delete() (result provisioner.Result, err error) {
...
 	}
 
+	if ironicNode.PowerState == powerOn {
+		_, err := p.hardPowerOff()
+		if err != nil {
+			result.Dirty = true
+			switch err.(type) {
+			case HostLockedError:
+				p.log.Info("could not power off host, busy")
+				return operationContinuing(powerRequeueDelay)
```
andfasano, line 1627: On a second thought, it's probably better to use `retryAfterDelay` here, since it's currently used when a retry is required after an error. `operationContinuing` has technically the same behavior, but it is currently adopted in normal scenarios where an error is not triggered (ie waiting for some event)

pkg/provisioner/ironic/ironic.go
```diff
@@ -1613,6 +1617,30 @@ func (p *ironicProvisioner) Delete() (result provisioner.Result, err error) {
...
+			default:
+				// For other errors continue with the delete
+				if p.host.Status.ErrorCount > maxErrorRetryCount {
+					result.Dirty = false
```
andfasano, line 1631: not required

pkg/provisioner/ironic/ironic.go
```diff
@@ -1613,6 +1617,30 @@ func (p *ironicProvisioner) Delete() (result provisioner.Result, err error) {
...
+					p.log.Info("could not power off host before removal")
+					operationComplete()
```
andfasano, line 1633: missing `return`

pkg/provisioner/ironic/ironic.go
```diff
@@ -1613,6 +1617,30 @@ func (p *ironicProvisioner) Delete() (result provisioner.Result, err error) {
...
+				} else {
+					return operationFailed("failed to power off host")
+				}
+			}
+		} else {
+			p.log.Info("host powered off before removal")
+			operationComplete()
```
andfasano, line 1640: missing `return`

pkg/provisioner/ironic/ironic.go
sadasu: True but it could be used for other retries too. I'll try to come up with a better name.

pkg/provisioner/ironic/ironic.go
sadasu: Dirty flag is set to true as soon as we detect there is an error (here https://github.com/metal3-io/baremetal-operator/pull/412/files#diff-f05fc4a0c9679815bc34bcef100c2a8d525688f2804a1128c5b9e1e575b76c29R1623). When we don't want to try again, we need to reset it, right?

pkg/provisioner/ironic/ironic.go
sadasu: returning here would mean we would not proceed to nodes.Delete(), right?

pkg/provisioner/ironic/ironic.go
andfasano: At the end of the flow having the `dirty` set to true indicates to update the current host status and to continue. In this case the intention, IIUC, is to give up and stop with the reconciliation. `operationComplete` it's used for that (and in fact it does not set the dirty flag)

pkg/provisioner/ironic/ironic.go
andfasano: Sorry I was probably too concise, my point is that the semantic functions like `operationComplete` are meant as a wrapper to package a `provisioner.Result` and an `error`, so its return value must be used somehow

sadasu: Removed the operationComplete() here because it is premature, imo. We want to proceed to the delete of the node.

pkg/provisioner/ironic/ironic.go
sadasu: same as https://github.com/metal3-io/baremetal-operator/pull/412#discussion_r587741704

pkg/provisioner/ironic/ironic.go
sadasu: Ok. Understand. I removed setting of the dirty flag here.

pkg/provisioner/ironic/ironic.go
sadasu: I agree. So, going back to retryAfterDelay.

pkg/provisioner/ironic/ironic.go
sadasu: Renamed it to maxPowerOffRetryCount. Hopefully, that is a better name.

pkg/provisioner/ironic/ironic.go
sadasu: Thanks for pointing that out. I have updated this code to use the new methods defined in result.go.

pkg/provisioner/ironic/ironic.go
sadasu: https://github.com/metal3-io/baremetal-operator/pull/412#discussion_r587744498

pkg/provisioner/ironic/ironic.go
sadasu: Done.

pkg/provisioner/ironic/ironic.go
andfasano: I think that the threshold check is not in the right position. In case of `hardPowerOff` error another reconcile will be invoked (thanks to the `operationFailed`), so the (increased) `host.Status.ErrorCount` could be evaluated only in the next loop. This means that the check must be performed _before_ `hardPowerOff`, ie something like:

`if ironicNode.PowerState == powerOn && p.host.Status.ErrorCount < maxPowerOffRetryCount {
...
}`

pkg/provisioner/ironic/ironic.go
```diff
@@ -1613,6 +1617,29 @@ func (p *ironicProvisioner) Delete() (result provisioner.Result, err error) {
...
+			default:
+				// For other errors continue with the delete
+				if p.host.Status.ErrorCount > maxPowerOffRetryCount {
+					p.log.Info("could not power off host before removal")
+				} else {
+					return operationFailed("failed to power off host")
+				}
+			}
+		} else {
+			p.log.Info("host powered off before removal")
```
andfasano, line 1637: There's already a log trace before powering off

pkg/provisioner/ironic/ironic.go
```diff
@@ -1613,6 +1617,29 @@ func (p *ironicProvisioner) Delete() (result provisioner.Result, err error) {
...
+		}
+	} else {
+		p.log.Info("host not powered on")
```
andfasano, line 1640: I would remove it

pkg/provisioner/ironic/ironic.go
sadasu: Done.

pkg/provisioner/ironic/ironic.go
sadasu: Ah! Got it! Fixing it in my next version.

pkg/provisioner/ironic/delete_test.go
```diff
@@ -143,6 +144,27 @@ func TestDelete(t *testing.T) {
...
+			}),
+			powerOn:              true,
+			expectedDirty:        true,
+			expectedRequestAfter: 0,
+			expectedUpdate: &nodes.UpdateOperation{
+				Op:    "replace",
+				Path:  "/maintenance",
+				Value: true,
+			},
+		},
```
andfasano, line 167: I think it's missing the test case verifying what happens when the number of retries are greater than `maxPowerOffRetryCount`? (it could be reasonable to have a dedicated separate test for that probably)

sadasu: Please refer to https://github.com/metal3-io/baremetal-operator/pull/816.

sadasu:
Implementation has moved to https://github.com/metal3-io/baremetal-operator/pull/816.

PR #417, opened 2020-02-05
Add log-level flag

### LINKED ISSUE #345: [RFE] Allow customizing the loglevel
Currently the loglevels are configured as debug. While I agree this is really helpful for debugging issues, it would be helpful to expose this as a knob that can be configured.

ian-howell:
Closes #345

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -48,7 +48,7 @@ func setupLogger(devLogging bool, logLevel string) {
 	case "error":
 		level = zap.NewAtomicLevelAt(zap.ErrorLevel)
 	default:
-		level = zap.NewAtomicLevelAt(zap.DebugLevel)
+		level = zap.NewAtomicLevelAt(zap.InfoLevel)
 	}
 	logOptions := []zaplog.Opts{
 		zaplog.UseDevMode(devLogging),
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -12,12 +12,15 @@ import (
 	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
 	sdkVersion "github.com/operator-framework/operator-sdk/version"
 
+	"go.uber.org/zap"
+
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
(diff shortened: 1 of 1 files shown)


maelk:
Should it default to debug ? Maybe info would be better ?

ian-howell:
> Should it default to debug ? Maybe info would be better ?

Current behavior is to use Debug, but I agree that Info should be the default.

zaneb:
AFAIK we don't yet have a way to log anything at any level other than INFO, so this is a bit of an academic exercise. Is that something you'd be up for fixing also?

cmd/manager/main.go
```diff
@@ -29,16 +32,38 @@ func printVersion() {
...
+	case "debug":
+		level = zap.NewAtomicLevelAt(zap.DebugLevel)
+	case "info":
+		level = zap.NewAtomicLevelAt(zap.InfoLevel)
+	case "warn":
+		level = zap.NewAtomicLevelAt(zap.WarnLevel)
+	case "error":
+		level = zap.NewAtomicLevelAt(zap.ErrorLevel)
+	default:
+		level = zap.NewAtomicLevelAt(zap.InfoLevel)
```
dhellmann, line 51: Should we log a warning that the log level given wasn't understood so we're defaulting to "info"?

zaneb:
So apparently the logging interface used in k8s has a [philosophy](https://github.com/go-logr/logr#readme) that all logs should be at INFO or ERROR level, and thus those are the only ones allowed by the API. Apparently this is considered slightly edgy because it is a departure from the golang [philosophy](https://dave.cheney.net/2015/11/05/lets-talk-about-logging) that there should not be an ERROR level either.

Anyhoo, apparently what we should be doing is adding a verbosity indicator as a structured field in the INFO-level logs.

I'm going to refrain from saying what I really think about this philosophy.

/hold

dhellmann:
I guess if we assume all of the logs are being collected for later processing anyway there is less need for traditional "levels" to limit the output.

It seems like the implementation here should change to control the verbosity level as a single integer option, and we need to update all of our log messages with verbosity. Does anyone actually use that feature anywhere else in kubernetes? I haven't really noticed it, but I haven't gone looking for it.

Is it possible that we actually want to change nothing?

PR #418, opened 2020-02-06
Change ironic endpoint keepalived image

kashifest:
Change the image for ironic endpoint keepalived

### DIFF
--- a/deploy/ironic-keepalived-config/image_patch.yaml
+++ b/deploy/ironic-keepalived-config/image_patch.yaml
@@ -7,5 +7,5 @@ spec:
     spec:
       containers:
       # Change the value of image field below to your controller image URL
-      - image: quay.io/metal3-io/ironic-endpoint-keepalived:latest
+      - image: quay.io/metal3-io/keepalived
         name: ironic-endpoint-keepalived


deploy/ironic-keepalived-config/image_patch.yaml
```diff
@@ -7,5 +7,5 @@ spec:
     spec:
       containers:
       # Change the value of image field below to your controller image URL
-      - image: quay.io/metal3-io/ironic-endpoint-keepalived:latest
+      - image: quay.io/metal3-io/keepalived
```
dhellmann: Did the image name change? Or did we have the entirely wrong image here?

maelk: it was wrong. we did not have a quay.io repo

dhellmann: OK. It would be good to have that sort of detail in the PR description or commit message.

kashifest: @dhellmann, sorry I didn't realize it was miss-leading, I thought the directory itself and the commit message and PR description was self-explanatory.

dhellmann: @kashifest I didn't know the history. The change looked fine, but if I was looking at the git logs later I wouldn't have been able to explain why this change was needed. Even replacing "change" with "fix" in the description would have hinted that what was there was broken somehow. An extra line or two in the description would have clarified it completely for me.

It's not a big deal, but I thought I'd mention it since it can be easy to overlook that sort of opportunity for knowledge transfer when a project is moving quickly.

Thanks for fixing the problem!

kashifest: @dhellmann I agree

PR #411, opened 2020-02-03
add ilo5-redfish alias

dhellmann:
The "redfish" type implies generic support for the Redfish
protocol. If a hardware vendor ever needs customization, we will want
an alias in place. This patch adds such an alias for iLO 5 gear,
without adding any customization.

PR #419, opened 2020-02-06
Add idrac-redfish url type

juliakreger:
The idrac BMCs have partiular handling of boot mode as configuration
change requests which need to be honored only at power state changes
as opposed to immediately, because this can conflict and over-write
existing configuration jobs.

While we shouldn't be in this state without more advanced bmc
features being leveraged, it is best to use the vendor supported
hardware type and interfaces for using redfish with their hardware.

This boot interface defaults to using iPXE, as opposed to virtual
media. A separate idrac specific redfish virtual media interface
already existed in the baremetal-operator.


pkg/bmc/redfish.go
```diff
@@ -116,6 +129,30 @@ func (a *redfishAccessDetails) RAIDInterface() string {
...
 	return ""
 }
+// iDrac Redfish Overrides
+
+func (a *redfishAccessDetails) Driver() string {
+	return "idrac"
+}
+
+func (a *redfishiDracAccessDetails) BootInterface() string {
+        return "idrac-redfish-virtual-media"
```
zaneb: This doesn't look right.

juliakreger: nope! Fixing!

docs/api.md
```diff
@@ -23,6 +23,11 @@ mainly, but not only, provisioning details.
         and the port is optional, if using the default one (623).
     * Dell iDRAC
       * `idrac://` (or `idrac+http://` to disable TLS).
+      * `idrac-redfish://``may be used as a prefix to select use of an iDRAC
+        controller with the Redfish protocol as configuration requets  require
+        specific handling with Dell BMCs. The URL must also contain a path to
```
```suggestion
        The URL must also contain a path to
```

docs/api.md
```diff
@@ -23,6 +23,11 @@ mainly, but not only, provisioning details.
...
+        the system ID, for example
+        `idrac-redfish://myhost.example/redfish/v1/Systems/MySystemExample`
```
dhellmann: It would be good to be consistent with #413 here

```suggestion
        `idrac-redfish://myhost.example/redfish/v1/Systems/System.Embedded.1`
        or `idrac-redfish://myhost.example/redfish/v1/Systems/1`
```

pkg/bmc/redfish.go
```diff
@@ -116,6 +129,30 @@ func (a *redfishAccessDetails) RAIDInterface() string {
 func (a *redfishAccessDetails) VendorInterface() string {
 	return ""
 }
+// iDrac Redfish Overrides
+
+func (a *redfishAccessDetails) Driver() string {
```
dhellmann:
```suggestion
func (a *redfishiDracAccessDetails) Driver() string {
```

pkg/bmc/redfish.go
```diff
@@ -116,6 +129,30 @@ func (a *redfishAccessDetails) RAIDInterface() string {
...
+func (a *redfishiDracAccessDetails) PowerInterface() string {
+	return "idrac-redfish"
+}
+
+func (a *redfishiDracAccessDetails) RAIDInterface() string {
+	return "no-raid"
+}
+
+func (a *redfishiDracAccessDetails) VendorInterface() string {
+	return "no-vendor"
```
dhellmann:
```suggestion
	return "no-vendor"
}
```

docs/api.md
```diff
@@ -23,6 +23,11 @@ mainly, but not only, provisioning details.
         and the port is optional, if using the default one (623).
     * Dell iDRAC
       * `idrac://` (or `idrac+http://` to disable TLS).
```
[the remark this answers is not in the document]
juliakreger: I suspect we only want to support https use, tbh.

docs/api.md
juliakreger: I've got the actual default location in another window, I suspect that would be best....

docs/api.md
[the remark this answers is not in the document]
dhellmann: It would still be helpful to explain to the user why to use "idrac" vs. "idrac-redfish".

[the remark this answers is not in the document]
juliakreger: Hopefully the new version clarifies that.

pkg/bmc/redfish.go
```diff
@@ -38,6 +39,12 @@ func newRedfishiDracVirtualMediaAccessDetails(parsedURL *url.URL) (AccessDetails
 	}, nil
 }
 
+func newRedfishiDracAccessDetails(parsedURL *url.URL) (AccessDetails, error) {
```
zaneb: Tests are failing because #396 merged, so this also needs to take a bool for `disableCertificateVerification`.

pkg/bmc/access_test.go
```diff
@@ -372,6 +372,18 @@ func TestStaticDriverInfo(t *testing.T) {
 			vendor: "",
 		},
 
+		{
+			Scenario: "idrac redfish",
+			input:    "idrac+redfish://192.168.122.1",
```
zaneb: It's `idrac-redfish` in the docs (not `idrac+redfish`). I think `-` is correct for consistency with #411. So far we've only used `+` for `+http` and `+https`.

pkg/bmc/redfish.go
```diff
@@ -12,6 +12,7 @@ func init() {
 	registerFactory("redfish-virtualmedia", newRedfishVirtualMediaAccessDetails)
 	registerFactory("ilo5-virtualmedia", newRedfishVirtualMediaAccessDetails)
 	registerFactory("idrac-virtualmedia", newRedfishiDracVirtualMediaAccessDetails)
+	registerFactory("idrac+redfish", newRedfishiDracAccessDetails)
```
zaneb: Should be `idrac-redfish`.

juliakreger: was in an earlier patchset, but changed it trying to figure out go was throwing weird errors. According to stepehen it seems to be a rebase issue.

dhellmann:
@juliakreger it looks like this needs to be run through `go fmt`. Do you want to do that and update the PR?

dhellmann:
This will need to be rebased.

dhellmann:
I think #633 replaces this.

PR #424, opened 2020-02-12
Reboot API Implementation

n1r1:
Allow other controllers to request a reboot for a host by adding reboot annotation, as described in 
https://github.com/metal3-io/metal3-docs/pull/48

Signed-off-by: Nir <niry@redhat.com>

### DIFF
--- a/pkg/apis/metal3/v1alpha1/baremetalhost_types.go
+++ b/pkg/apis/metal3/v1alpha1/baremetalhost_types.go
@@ -457,6 +457,15 @@ type ProvisionStatus struct {
 	// Image holds the details of the last image successfully
 	// provisioned to the host.
 	Image Image `json:"image,omitempty"`
+
+	// LastPoweredOn is the time that the server was last powered on
+	// with the specified image
+	// +optional
+	LastPoweredOn *metav1.Time `json:"lastPoweredOn,omitempty"`
+
+	// PendingRebootSince is the time after which the server must be rebooted
+	// +optional
+	PendingRebootSince *metav1.Time `json:"pendingRebootSince,omitempty"`
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
@@ -526,6 +535,12 @@ func (host *BareMetalHost) ClearError() (dirty bool) {
 		dirty = true
 	}
 	return dirty
+}
+
+// RecordPoweredOn records the current time as the LastPoweredOn time.
+func (host *BareMetalHost) RecordPoweredOn() {
+	now := metav1.Now()
+	host.Status.Provisioning.LastPoweredOn = &now
 }
 
 // setLabel updates the given label when necessary and returns true
@@ -739,6 +754,7 @@ func (host *BareMetalHost) OperationMetricForState(operation ProvisioningState) 
 	}
 	return
 }
+
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
--- a/pkg/apis/metal3/v1alpha1/baremetalhost_types.go
+++ b/pkg/apis/metal3/v1alpha1/baremetalhost_types.go
@@ -458,10 +458,10 @@ type ProvisionStatus struct {
 	// provisioned to the host.
 	Image Image `json:"image,omitempty"`
 
-	// LastPoweredOn is the time that the server was last powered on
+	// PoweredOnAt is the time that the server was powered on
 	// with the specified image
 	// +optional
-	LastPoweredOn *metav1.Time `json:"lastPoweredOn,omitempty"`
+	PoweredOnAt *metav1.Time `json:"poweredOnAt,omitempty"`
 
 	// PendingRebootSince is the time after which the server must be rebooted
 	// +optional
@@ -537,10 +537,12 @@ func (host *BareMetalHost) ClearError() (dirty bool) {
 	return dirty
 }
 
-// RecordPoweredOn records the current time as the LastPoweredOn time.
+// RecordPoweredOn records the current time as the PoweredOnAt time if there's no existing value
 func (host *BareMetalHost) RecordPoweredOn() {
-	now := metav1.Now()
-	host.Status.Provisioning.LastPoweredOn = &now
+	if host.Status.Provisioning.PoweredOnAt.IsZero() {
+		now := metav1.Now()
+		host.Status.Provisioning.PoweredOnAt = &now
+	}
 }
 
 // setLabel updates the given label when necessary and returns true
--- a/pkg/apis/metal3/v1alpha1/baremetalhost_types.go
+++ b/pkg/apis/metal3/v1alpha1/baremetalhost_types.go
@@ -755,7 +755,6 @@ func (host *BareMetalHost) OperationMetricForState(operation ProvisioningState) 
 	return
 }
 
-
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
 // BareMetalHostList contains a list of BareMetalHost
--- a/pkg/apis/metal3/v1alpha1/baremetalhost_types.go
+++ b/pkg/apis/metal3/v1alpha1/baremetalhost_types.go
@@ -457,15 +457,6 @@ type ProvisionStatus struct {
 	// Image holds the details of the last image successfully
 	// provisioned to the host.
 	Image Image `json:"image,omitempty"`
-
-	// PoweredOnAt is the time that the server was powered on
-	// with the specified image
-	// +optional
-	PoweredOnAt *metav1.Time `json:"poweredOnAt,omitempty"`
-
-	// PendingRebootSince is the time after which the server must be rebooted
-	// +optional
-	PendingRebootSince *metav1.Time `json:"pendingRebootSince,omitempty"`
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
@@ -535,14 +526,6 @@ func (host *BareMetalHost) ClearError() (dirty bool) {
 		dirty = true
 	}
 	return dirty
-}
-
-// RecordPoweredOn records the current time as the PoweredOnAt time if there's no existing value
-func (host *BareMetalHost) RecordPoweredOn() {
-	if host.Status.Provisioning.PoweredOnAt.IsZero() {
-		now := metav1.Now()
-		host.Status.Provisioning.PoweredOnAt = &now
-	}
 }
 
 // setLabel updates the given label when necessary and returns true
--- a/pkg/controller/baremetalhost/action_result.go
+++ b/pkg/controller/baremetalhost/action_result.go
@@ -19,6 +19,7 @@ type actionResult interface {
 // state.
 type actionContinue struct {
 	delay time.Duration
+	dirty bool
 }
 
 func (r actionContinue) Result() (result reconcile.Result, err error) {
@@ -29,7 +30,7 @@ func (r actionContinue) Result() (result reconcile.Result, err error) {
 }
 
 func (r actionContinue) Dirty() bool {
-	return true
+	return r.dirty
 }
 
 // actionComplete is a result indicating that the current action has completed,
--- a/pkg/controller/baremetalhost/action_result.go
+++ b/pkg/controller/baremetalhost/action_result.go
@@ -14,12 +14,30 @@ type actionResult interface {
 	Dirty() bool
 }
 
+// actionContinueNoWrite is a result indicating that the current action is still
+// in progress, and that the resource should remain in the same provisioning
+// state without writing the status
+type actionContinueNoWrite struct {
+	delay time.Duration
+}
+
+func (r actionContinueNoWrite) Result() (result reconcile.Result, err error) {
+	result.RequeueAfter = r.delay
+	// Set Requeue true as well as RequeueAfter in case the delay is 0.
+	result.Requeue = true
+	return
+}
+
+func (r actionContinueNoWrite) Dirty() bool {
+	return false
+}
+
+
 // actionContinue is a result indicating that the current action is still
 // in progress, and that the resource should remain in the same provisioning
 // state.
 type actionContinue struct {
 	delay time.Duration
-	dirty bool
 }
 
 func (r actionContinue) Result() (result reconcile.Result, err error) {
@@ -30,7 +48,7 @@ func (r actionContinue) Result() (result reconcile.Result, err error) {
 }
 
 func (r actionContinue) Dirty() bool {
-	return r.dirty
+	return true
 }
 
 // actionComplete is a result indicating that the current action has completed,
--- a/pkg/controller/baremetalhost/action_result.go
+++ b/pkg/controller/baremetalhost/action_result.go
@@ -18,14 +18,7 @@ type actionResult interface {
 // in progress, and that the resource should remain in the same provisioning
 // state without writing the status
 type actionContinueNoWrite struct {
-	delay time.Duration
-}
-
-func (r actionContinueNoWrite) Result() (result reconcile.Result, err error) {
-	result.RequeueAfter = r.delay
-	// Set Requeue true as well as RequeueAfter in case the delay is 0.
-	result.Requeue = true
-	return
+	actionContinue
 }
 
 func (r actionContinueNoWrite) Dirty() bool {
--- a/pkg/controller/baremetalhost/baremetalhost_controller.go
+++ b/pkg/controller/baremetalhost/baremetalhost_controller.go
@@ -685,6 +685,11 @@ func (r *ReconcileBareMetalHost) manageHostPower(prov provisioner.Provisioner, i
 
 		if suffixlessAnnotationExists && !info.host.Status.PoweredOn {
 			delete(info.host.Annotations, rebootAnnotationPrefix)
+
+			if err = r.client.Update(context.TODO(), info.host); err != nil {
+				return actionError{errors.Wrap(err, "failed to remove reboot annotation from host")}
+			}
+
 			return actionContinue{}
 		}
 
--- a/pkg/controller/baremetalhost/baremetalhost_controller.go
+++ b/pkg/controller/baremetalhost/baremetalhost_controller.go
@@ -391,7 +391,7 @@ func getRebootAnnotations(host *metal3v1alpha1.BareMetalHost) (suffixlessExist b
 	 _, suffixlessExist = host.Annotations[rebootAnnotationPrefix]
 
 	for annotation := range host.Annotations{
-		if strings.HasPrefix(annotation, rebootAnnotationPrefix) && annotation != rebootAnnotationPrefix{
+		if strings.HasPrefix(annotation, rebootAnnotationPrefix + "/") {
 			withSuffixExist = true
 			return
 		}
@@ -646,7 +646,7 @@ func (r *ReconcileBareMetalHost) actionDeprovisioning(prov provisioner.Provision
 	info.host.Status.Provisioning.PendingRebootSince = nil
 	if info.host.Annotations != nil {
 		for annotation := range info.host.Annotations {
-			if strings.HasPrefix(annotation, rebootAnnotationPrefix) {
+			if annotation == rebootAnnotationPrefix || strings.HasPrefix(annotation, rebootAnnotationPrefix + "/") {
 				delete(info.host.Annotations, annotation)
 			}
 		}
@@ -678,21 +678,18 @@ func (r *ReconcileBareMetalHost) manageHostPower(prov provisioner.Provisioner, i
 	lastPoweredOn := info.host.Status.Provisioning.LastPoweredOn
 
 	desiredPowerOnState := info.host.Spec.Online
-	isInRebootProcess := lastPoweredOn.Before(pendingRebootSince) || (pendingRebootSince != nil && lastPoweredOn == nil)
+	isInRebootProcess := lastPoweredOn.Before(pendingRebootSince) || (!pendingRebootSince.IsZero() && lastPoweredOn.IsZero())
 
 	if isInRebootProcess {
 		suffixlessAnnotationExists, shouldHoldPowerOff := getRebootAnnotations(info.host)
 
+		if suffixlessAnnotationExists && !info.host.Status.PoweredOn {
+			delete(info.host.Annotations, rebootAnnotationPrefix)
+			return actionContinue{}
+		}
+
 		if info.host.Status.PoweredOn || shouldHoldPowerOff {
 			desiredPowerOnState = false
-		} else {
-			if suffixlessAnnotationExists {
-				delete(info.host.Annotations, rebootAnnotationPrefix)
-
-				if err = r.client.Update(context.TODO(), info.host); err != nil {
-					return actionError{errors.Wrap(err, "failed to remove reboot annotation from host")}
-				}
-			}
 		}
 	}
 
--- a/pkg/controller/baremetalhost/baremetalhost_controller.go
+++ b/pkg/controller/baremetalhost/baremetalhost_controller.go
@@ -390,10 +390,6 @@ func isRebootAnnotation(annotation string) bool{
 
 // clearRebootAnnotations deletes all reboot annotations exist on the provided host
 func clearRebootAnnotations(host *metal3v1alpha1.BareMetalHost) (dirty bool) {
-	if len(host.Annotations) == 0 {
-		return
-	}
-
 	for annotation := range host.Annotations {
 		if isRebootAnnotation(annotation){
 			delete(host.Annotations, annotation)
--- a/pkg/controller/baremetalhost/baremetalhost_controller.go
+++ b/pkg/controller/baremetalhost/baremetalhost_controller.go
@@ -375,10 +375,6 @@ func (r *ReconcileBareMetalHost) credentialsErrorResult(err error, request recon
 
 // hasRebootAnnotation checks for existence of reboot annotations and returns true if at least one exist
 func hasRebootAnnotation(host *metal3v1alpha1.BareMetalHost) bool {
-	if len(host.Annotations) == 0 {
-		return false
-	}
-
 	for annotation := range host.Annotations {
 		if isRebootAnnotation(annotation) {
 			return true
--- a/pkg/controller/baremetalhost/baremetalhost_controller.go
+++ b/pkg/controller/baremetalhost/baremetalhost_controller.go
@@ -381,22 +381,18 @@ func (r *ReconcileBareMetalHost) credentialsErrorResult(err error, request recon
 	}
 }
 
-// check for existence of reboot annotations and returns two booleans indicating
-// the existence of suffix-less reboot annotation and suffixed reboot annotations
-func getRebootAnnotations(host *metal3v1alpha1.BareMetalHost) (suffixlessExist bool, withSuffixExist bool) {
-	if host.Annotations == nil {
-		return
+// hasRebootAnnotation checks for existence of reboot annotations and returns true if at least one exist
+func hasRebootAnnotation(host *metal3v1alpha1.BareMetalHost) bool {
+	if len(host.Annotations) == 0 {
+		return false
 	}
-
-	_, suffixlessExist = host.Annotations[rebootAnnotationPrefix]
 
 	for annotation := range host.Annotations {
-		if strings.HasPrefix(annotation, rebootAnnotationPrefix+"/") {
-			withSuffixExist = true
-			return
+		if strings.HasPrefix(annotation, rebootAnnotationPrefix+"/") || annotation == rebootAnnotationPrefix {
+			return true
 		}
 	}
-	return
+	return false
 }
 
 // checks if reboot is required
@@ -405,9 +401,7 @@ func checkUpdatedRebootRequest(host *metal3v1alpha1.BareMetalHost) (dirty bool) 
 		return
 	}
 
-	suffixlessExist, withSuffixExist := getRebootAnnotations(host)
-
-	if !(suffixlessExist || withSuffixExist) {
+	if !hasRebootAnnotation(host) {
 		return
 	}
 
@@ -678,19 +672,18 @@ func (r *ReconcileBareMetalHost) manageHostPower(prov provisioner.Provisioner, i
 	isInRebootProcess := poweredOnAt.Before(pendingRebootSince) || (!pendingRebootSince.IsZero() && poweredOnAt.IsZero())
 
 	if isInRebootProcess {
-		suffixlessAnnotationExists, shouldHoldPowerOff := getRebootAnnotations(info.host)
+		if !info.host.Status.PoweredOn {
+			if _, hasAnn := info.host.Annotations[rebootAnnotationPrefix]; hasAnn {
+				delete(info.host.Annotations, rebootAnnotationPrefix)
 
-		if suffixlessAnnotationExists && !info.host.Status.PoweredOn {
-			delete(info.host.Annotations, rebootAnnotationPrefix)
-
-			if err = r.client.Update(context.TODO(), info.host); err != nil {
-				return actionError{errors.Wrap(err, "failed to remove reboot annotation from host")}
+				if err = r.client.Update(context.TODO(), info.host); err != nil {
+					return actionError{errors.Wrap(err, "failed to remove reboot annotation from host")}
+				}
+				return actionContinue{}
 			}
-
-			return actionContinue{}
 		}
 
-		if info.host.Status.PoweredOn || shouldHoldPowerOff {
+		if info.host.Status.PoweredOn || hasRebootAnnotation(info.host) {
 			desiredPowerOnState = false
 		}
 	}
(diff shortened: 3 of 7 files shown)


pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -592,19 +659,42 @@ func (r *ReconcileBareMetalHost) manageHostPower(prov provisioner.Provisioner, i
...
+	if isInRebootProcess {
+		suffixlessAnnotationExists, shouldHoldPowerOff := getRebootAnnotations(info.host)
+
+		if info.host.Status.PoweredOn || shouldHoldPowerOff {
+			desiredPowerOnState = false
+		} else {
+			if suffixlessAnnotationExists {
+				delete(info.host.Annotations, rebootAnnotationPrefix)
+
+				if err = r.client.Update(context.TODO(), info.host); err != nil {
```
dhellmann: If we save the resource here and then need to save some other value again later, we'll end up with a conflict error. We usually wait to save the host in Reconcile() based on the type of return value from this method to avoid that race and to allow us to make multiple changes to the resource. This is updating the resource, not it's status, so the existing call to save the resource won't work.

I think to avoid the race, the logic to manage the annotations needs to all happen in this method, before the power is toggled (see related comment about the changes in Reconcile() above).

If the annotation is present and pendingRebootSince is not set, set pendingRebootSince and return actionContinue immediately so Reconcile() stores the status and then loops back to this point to change the power status.

If the annotation is present the lastPoweredOn value is after pendingRebootSince, and the power is actually on, remove the annotation, call Update(), and return actionContinue immediately, without making any other changes. 

Otherwise, figure out what the power state should be and try to enforce it.

Restructuring like that means that we will enter this method a few different times as part of handling each step of the reboot, but a failed attempt to power back on as part of a reboot will keep trying and the annotation will only be removed when the power on call is successful.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -592,19 +659,42 @@ func (r *ReconcileBareMetalHost) manageHostPower(prov provisioner.Provisioner, i
...
+
+	desiredPowerOnState := info.host.Spec.Online
+	isInRebootProcess := lastPoweredOn.Before(pendingRebootSince) || (pendingRebootSince != nil && lastPoweredOn == nil)
+
+	if isInRebootProcess {
+		suffixlessAnnotationExists, shouldHoldPowerOff := getRebootAnnotations(info.host)
+
+		if info.host.Status.PoweredOn || shouldHoldPowerOff {
+			desiredPowerOnState = false
+		} else {
```
zaneb: I think this might be easier to understand as two separate blocks rather than if/else:

```
if isInRebootProcess {
    if desiredPowerOnState && !info.host.Status.PoweredOn {
        if _, hasAnn := info.host.Annotations[rebootAnnotationPrefix]; hasAnn {
            delete(info.host.Annotations, rebootAnnotationPrefix)
            return actionContinue{}
        }
    }
    if info.host.Status.PoweredOn || hasRebootAnnotation(info.host) {
       desiredPowerOnState = false
    }
}
```

(I don't think it's actually necessary, or even desirable, to check the desiredPowerOnState before removing the suffixless annotation, but in the spec we said we would.)

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -569,6 +627,15 @@ func (r *ReconcileBareMetalHost) actionDeprovisioning(prov provisioner.Provision
 	// After the provisioner is done, clear the image settings so we
 	// transition to the next state.
 	info.host.Status.Provisioning.Image = metal3v1alpha1.Image{}
+	info.host.Status.Provisioning.LastPoweredOn = nil
+	info.host.Status.Provisioning.PendingRebootSince = nil
+	if info.host.Annotations != nil {
+		for annotation := range info.host.Annotations {
+			if strings.HasPrefix(annotation, rebootAnnotationPrefix) {
```
zaneb: This is not quite robust enough, see above.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -535,6 +592,7 @@ func (r *ReconcileBareMetalHost) actionProvisioning(prov provisioner.Provisioner
 		// to return false, indicating that it has no more work to
 		// do.
 		info.host.ClearError()
+		info.host.RecordPoweredOn()
```
zaneb: I think you copied this form my earlier PR, but I just noticed two problems with it:

* This will get called multiple times, and update the timestamp each time when we only want it to update the timestamp the first time.
* Really we should write this (and verify that it is written) *before* starting to provision, power on, &c.

This implies that we also need to clear it when we have completed deprovisioning or power off.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -620,14 +710,17 @@ func (r *ReconcileBareMetalHost) manageHostPower(prov provisioner.Provisioner, i
...
+			if desiredPowerOnState {
 				metricLabels[labelPowerOnOff] = "on"
 			} else {
 				metricLabels[labelPowerOnOff] = "off"
 			}
 			powerChangeAttempts.With(metricLabels).Inc()
 		})
 		info.host.ClearError()
+		if desiredPowerOnState {
+			info.host.RecordPoweredOn()
```
zaneb: This will also get called multiple times, see previous comment.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -343,6 +352,50 @@ func (r *ReconcileBareMetalHost) credentialsErrorResult(err error, request recon
 	}
 }
 
+// check for existence of reboot annotations and returns two booleans indicating
+// the existence of suffix-less reboot annotation and suffixed reboot annotations
+func getRebootAnnotations(host *metal3v1alpha1.BareMetalHost) (suffixlessExist bool , withSuffixExist bool){
```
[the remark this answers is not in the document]
n1r1: I think we need it. 
In the suggested code you added below you had:

` if _, hasAnn := info.host.Annotations[rebootAnnotationPrefix]; hasAnn {`

This is actually checking for the existence of the sufflixeless annotation, and later there was  `hasRebootAnnotation()`, which was used after deletion of the sufflixless annotation, so it practically mean `hasSuffixedAnnotation()`

When you used `hasRebootAnnotation()` you had to think on the flow that lead you to this point and if the reboot annotation is suffixed or not. 
When you use `getRebootAnnotations()` you don't need to assume anything, it's more explicit and straightforward.

Let me know what you think

pkg/controller/baremetalhost/baremetalhost_controller_test.go
```diff
@@ -218,6 +219,224 @@ func TestSetLastUpdated(t *testing.T) {
...
+			}
+
+			if host.Status.PoweredOn {
+						return false
+					}
+
+			return true
+		},
+	)
+	time.Sleep(time.Second)
```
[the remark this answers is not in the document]
n1r1: To avoid cases where reboot is so quick that `LastPoweredOn == PendingRebootSince`

pkg/controller/baremetalhost/baremetalhost_controller.go
n1r1: Nice, I'll change the code accordingly.

> I don't think it's actually necessary, or even desirable, to check the desiredPowerOnState before removing the suffixless annotation, but in the spec we said we would

Perhaps we should change the spec?

pkg/controller/baremetalhost/baremetalhost_controller.go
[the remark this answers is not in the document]
zaneb: Yeah, that line is the same code, but I think it's easier to understand if it's local: in the first 'if' statement we're going to clear the suffixless annotation, and if that resulted in an actual change to the Host then we're going to return so it gets written right away.

Everywhere else we just need to check if there are *any* reboot annotations.

pkg/controller/baremetalhost/baremetalhost_controller_test.go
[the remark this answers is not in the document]
zaneb: AFAIK the time has nanosecond resolution.

pkg/controller/baremetalhost/baremetalhost_controller.go
n1r1: But if we clear `LastPoweredOn` after power off, this will break the existing code which relies on `lastPoweredOn.Before(pendingRebootSince)`  to detect reboot process.

pkg/controller/baremetalhost/baremetalhost_controller.go
n1r1: Good point! I replaced the `Update()` with `return actionContinue{}`.
Meanwhile I keep `checkUpdatedRebootRequest()` in it's place until we decide otherwise in the previous comment

Let me know if this covers everything or something is still missing

pkg/controller/baremetalhost/baremetalhost_controller_test.go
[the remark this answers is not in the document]
dhellmann: Yeah, either sleep *way* less or make it so we can send a time function into the code to control it's definition of "now" without sleeping.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -592,19 +659,39 @@ func (r *ReconcileBareMetalHost) manageHostPower(prov provisioner.Provisioner, i
...
+
+	desiredPowerOnState := info.host.Spec.Online
+	isInRebootProcess := lastPoweredOn.Before(pendingRebootSince) || (!pendingRebootSince.IsZero() && lastPoweredOn.IsZero())
+
+	if isInRebootProcess {
+		suffixlessAnnotationExists, shouldHoldPowerOff := getRebootAnnotations(info.host)
+
+		if suffixlessAnnotationExists && !info.host.Status.PoweredOn {
+			delete(info.host.Annotations, rebootAnnotationPrefix)
+			return actionContinue{}
```
dhellmann: This isn't going to do what you expect. The Reconcile() method only saves the status fields, which do not include metadata like the annotations. So you do need to call Update() before returning.

pkg/controller/baremetalhost/baremetalhost_controller_test.go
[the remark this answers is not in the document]
n1r1: Tests are failing when I replace second with ms or ns. I'll check why this is happening and update

pkg/controller/baremetalhost/baremetalhost_controller.go
dhellmann: Just returning isn't going to work, because it won't save the update to the annotation. That's why I tried to explain how to manage the annotations and avoid races from updating the metadata and status in the same pass through the reconcile loop.

pkg/controller/baremetalhost/baremetalhost_controller.go
n1r1: roger, fixed.

pkg/controller/baremetalhost/baremetalhost_controller_test.go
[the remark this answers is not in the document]
n1r1: The time is indeed in nanosecond resolution, but it's truncated to second granularity when storing it in k8s.

pkg/controller/baremetalhost/baremetalhost_controller.go
zaneb: If the host is powered off we don't need to detect whether it needs to be rebooted, do we?

n1r1: We don't need to detect whether it needs to be rebooted but we used it to report that the power on was part of the reboot process. I guess we can give up on this. 
So if we put it before power on/provisioning - don't we take a risk that `LastPowerdOn` will be updated and the power on/provisioning action will fail for a while? This might be confusing

pkg/controller/baremetalhost/baremetalhost_controller_test.go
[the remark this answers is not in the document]
n1r1: Removed `sleep` from tests

pkg/apis/metal3/v1alpha1/baremetalhost_types.go
```diff
@@ -736,6 +751,7 @@ func (host *BareMetalHost) OperationMetricForState(operation ProvisioningState)
 	return
 }
 
+
```
beekhof: nit: stray whitespace

pkg/controller/baremetalhost/baremetalhost_controller.go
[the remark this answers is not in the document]
beekhof: Seems mostly a style issue, in which case the maintainer always wins :)

pkg/controller/baremetalhost/baremetalhost_controller.go
beekhof: @zaneb does the spec become a design artefact at this point? Or do we keep it updated as the implementation changes?

pkg/controller/baremetalhost/baremetalhost_controller.go
beekhof: Clearing `LastPoweredOn` would make things work but feels wrong IMHO.
Maybe its because the `Last` part in `LastPoweredOn` losses its relevance.

What about adding `LastPoweredOff` and only  updating `LastPoweredOn` if it's less than `LastPoweredOff` ?

pkg/apis/metal3/v1alpha1/baremetalhost_types.go
n1r1: thanks, removed

pkg/controller/baremetalhost/baremetalhost_controller.go
zaneb: > Clearing LastPoweredOn would make things work but feels wrong IMHO.

Would renaming it to `PoweredOnTime` or `PoweredOnAt` make it feel less wrong? That would be simpler than adding another field.

> don't we take a risk that LastPowerdOn will be updated and the power on/provisioning action will fail for a while?

This is just the nature of the reconcile loop; you have to verify that you've successfully stored the state saying you're about to do something before you can start to do it. So not a big concern IMHO.

beekhof: I could like with changing it to ``PoweredOnAt``

n1r1: Great, just to be clear on this - `isInRebootProcess` will be true, when the host is powered off, even if we're not in a reboot process.
In other words, if we clear `PoweredOnAt` on power off, we don't have any way to differentiate between powered off machine that was powered off due to reboot or something else.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -584,6 +633,15 @@ func (r *ReconcileBareMetalHost) actionDeprovisioning(prov provisioner.Provision
 	// After the provisioner is done, clear the image settings so we
 	// transition to the next state.
 	info.host.Status.Provisioning.Image = metal3v1alpha1.Image{}
+	info.host.Status.Provisioning.PoweredOnAt = nil
+	info.host.Status.Provisioning.PendingRebootSince = nil
+	if info.host.Annotations != nil {
```
dhellmann: This block could be a little utility function `clearRebootAnnotation`.

The results of changing the annotations are not being saved after this. Just like around like 679, we need to update the spec portion of the host here to save the annotation change.

Alternatively, if we stored enough state to know that the reboot was done, we could just leave the annotation in place and let the controller that added it remove it again. That would simplify the logic here in this controller, since we wouldn't need to use 2 different ways to store the resource.

@zaneb what do you think about that?

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -607,19 +665,43 @@ func (r *ReconcileBareMetalHost) manageHostPower(prov provisioner.Provisioner, i
...
 
+	pendingRebootSince := info.host.Status.Provisioning.PendingRebootSince
+	poweredOnAt := info.host.Status.Provisioning.PoweredOnAt
+
+	desiredPowerOnState := info.host.Spec.Online
+	isInRebootProcess := poweredOnAt.Before(pendingRebootSince) || (!pendingRebootSince.IsZero() && poweredOnAt.IsZero())
+
+	if isInRebootProcess {
+		if !info.host.Status.PoweredOn {
+			if _, hasAnn := info.host.Annotations[rebootAnnotationPrefix]; hasAnn {
```
dhellmann: This isn't checking for both forms of the annotation. Should it? See earlier comment about the utility function.

pkg/controller/baremetalhost/baremetalhost_controller_test.go
```diff
@@ -243,6 +246,224 @@ func TestSetLastUpdated(t *testing.T) {
...
+	tryReconcile(t, r, host,
+		func(host *metal3v1alpha1.BareMetalHost, result reconcile.Result) bool {
+			if !host.Status.Provisioning.PendingRebootSince.Before(host.Status.Provisioning.PoweredOnAt) {
+				return false
+			}
+
+			if !host.Status.PoweredOn {
+				return false
+			}
+
```
dhellmann: If we're going to remove the annotation, we should test for that here, too.

pkg/controller/baremetalhost/baremetalhost_controller_test.go
```diff
@@ -243,6 +246,224 @@ func TestSetLastUpdated(t *testing.T) {
...
```
dhellmann: If we're going to remove the annotation, we should test for that here, too.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -189,6 +190,14 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (result re
 		return reconcile.Result{Requeue: true}, nil
 	}
 
+	if checkUpdatedRebootRequest(host) {
```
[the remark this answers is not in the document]
dhellmann: Your argument about timestamps makes sense.

Seeing this here, I thought we were going to handle reboots "out of band" of the rest of the state machine. I'm not sure that's even possible, though, given Ironic's state management. So I think it's fine to keep it as is for now.

pkg/controller/baremetalhost/baremetalhost_controller.go
dhellmann: The host has an online flag in the spec field to tell us if something is trying to force the power to be off. We aren't using that as part of the reboot, right?

pkg/controller/baremetalhost/baremetalhost_controller.go
n1r1: It first checks for the suffixless. if exists - removes it, and later we check again if any reboot annotation exists. Since we removed the suffixless, we know that the only possible annotation is a suffixed one. This code is different from the original code of the PR. I think it might be confusing. see [this discussion](https://github.com/metal3-io/baremetal-operator/pull/424#discussion_r380390961) for more details.

pkg/controller/baremetalhost/baremetalhost_controller.go
n1r1: Can you please clarify the question? 
The reboot code never **modifies** the `Online` flag, but it can obviously power off the host as part of the reboot, even though that `Online` is `true`.

dhellmann: Right, my point is that we can tell the difference between the states "off because we are rebooting" and "off because the user turned it off" by looking at spec.online.

pkg/controller/baremetalhost/baremetalhost_controller.go
dhellmann: I'm not following. Where is only one of the annotations removed but the other is left behind? I think I see it removing either form on line 640, but I admit I'm having a little trouble following the logic across all of the different places this change touches the code.

pkg/controller/baremetalhost/baremetalhost_controller.go
n1r1: Correct. But we can't tell "on because we just ended a reboot" we just can say "on because user wants it on". Not a big deal maybe.

dhellmann: If we can't tell that the host is on because just finished rebooting it, how do we know when to stop trying to reboot it?

pkg/controller/baremetalhost/baremetalhost_controller.go
n1r1: in line 677 the suffixless annotation being removed, if existed. When we reach line 686 we use `hasRebootAnnotation()` which returns true if either suffixless or suffixed annotation exist. Since we reached this line after we removed the suffixless, or more correctly - we reached this line after we checked that there's no suffixless annotation, so we know that if `hasRebootAnnotation()` returned `true` it means that a suffixed annotation exists.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -607,19 +665,43 @@ func (r *ReconcileBareMetalHost) manageHostPower(prov provisioner.Provisioner, i
...
+
+	if isInRebootProcess {
+		if !info.host.Status.PoweredOn {
+			if _, hasAnn := info.host.Annotations[rebootAnnotationPrefix]; hasAnn {
+				delete(info.host.Annotations, rebootAnnotationPrefix)
+
+				if err = r.client.Update(context.TODO(), info.host); err != nil {
+					return actionError{errors.Wrap(err, "failed to remove reboot annotation from host")}
+				}
+				return actionContinue{}
```
zaneb: To Doug's point, `actionContinue` always returns `Dirty() == true`, so it always results in us attempting to write to the Host. Which in this case will always generate an error (because we've written the metadata here, so the Host is stale). That won't actually break anything, but it will probably result in ugly-looking log messages every time.
I'm not sure of the best way to handle this. Maybe a flag to `actionContinue` to indicate that the annotations need to be written as well/instead?

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -635,14 +717,19 @@ func (r *ReconcileBareMetalHost) manageHostPower(prov provisioner.Provisioner, i
...
+			if desiredPowerOnState {
 				metricLabels[labelPowerOnOff] = "on"
 			} else {
 				metricLabels[labelPowerOnOff] = "off"
 			}
 			powerChangeAttempts.With(metricLabels).Inc()
 		})
 		info.host.ClearError()
+		if desiredPowerOnState {
+			info.host.RecordPoweredOn()
```
zaneb: This should be recorded and written to the datastore before we ever call `prov.PowerOn()`.
Also, as an added wrinkle, this function is called from both the Ready and Provisioned states, but we only want to record the boot time in the Provisioned state.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -635,14 +717,19 @@ func (r *ReconcileBareMetalHost) manageHostPower(prov provisioner.Provisioner, i
...
+		} else {
+			info.host.Status.Provisioning.PoweredOnAt = nil
```
zaneb: I'd expect we'd want to do this once `prov.UpdateHardwareState()` tells us that the power state is off. Here we're doing it as soon as we've asked Ironic to turn the power off, without waiting to see whether it is successful.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -607,19 +665,43 @@ func (r *ReconcileBareMetalHost) manageHostPower(prov provisioner.Provisioner, i
...
 	}
 
+	pendingRebootSince := info.host.Status.Provisioning.PendingRebootSince
+	poweredOnAt := info.host.Status.Provisioning.PoweredOnAt
+
+	desiredPowerOnState := info.host.Spec.Online
+	isInRebootProcess := poweredOnAt.Before(pendingRebootSince) || (!pendingRebootSince.IsZero() && poweredOnAt.IsZero())
+
+	if isInRebootProcess {
+		if !info.host.Status.PoweredOn {
```
zaneb: It's not clear to me why this is nested inside `if isInRebootProcess`. If we have a suffixless annotation present and the power is off, we should always just clear it I think?

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -607,19 +665,43 @@ func (r *ReconcileBareMetalHost) manageHostPower(prov provisioner.Provisioner, i
...
+				delete(info.host.Annotations, rebootAnnotationPrefix)
+
+				if err = r.client.Update(context.TODO(), info.host); err != nil {
+					return actionError{errors.Wrap(err, "failed to remove reboot annotation from host")}
+				}
+				return actionContinue{}
+			}
+		}
+
+		if info.host.Status.PoweredOn || hasRebootAnnotation(info.host) {
```
zaneb: I eventually convinced myself that this condition is correct, but it was hard work :D

* If a reboot annotation is still present, we want to force the power off.
* If not, it could be because someone added a suffixless annotation, we saw it and set `pendingRebootSince`, but then someone else removed the annotation. We don't want reboots to be cancellable, so we need to go through with it: as long as the power is still on we want to force it off.

No suggestion, just leaving this here so I don't have to work it out again next time.

dhellmann: Maybe an inline comment with the reasoning would be good.

beekhof: That sounds like a very good idea

pkg/controller/baremetalhost/baremetalhost_controller.go
n1r1: I've added `dirty` flag per your suggestion.

pkg/controller/baremetalhost/baremetalhost_controller_test.go
n1r1: In the suffixed case, the controller doesn't remove the annotation, and in the suffixless I have a check for that. let me know if you still think that something is missing.

pkg/controller/baremetalhost/baremetalhost_controller.go
n1r1: Since `PoweredOnAt` has been removed, this is outdated.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -41,6 +41,7 @@ import (
 const (
 	hostErrorRetryDelay = time.Second * 10
 	pauseRetryDelay     = time.Second * 30
+	rebootAnnotationPrefix    = "reboot.metal3.io"
```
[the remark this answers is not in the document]
n1r1: fixed

pkg/controller/baremetalhost/baremetalhost_controller.go
n1r1: > This block could be a little utility function clearRebootAnnotation.
done

> It is a problem that the annotation changes here won't get written out to the datastore when we return to Reconcile() though.
fixed

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -584,6 +633,15 @@ func (r *ReconcileBareMetalHost) actionDeprovisioning(prov provisioner.Provision
 	// After the provisioner is done, clear the image settings so we
 	// transition to the next state.
 	info.host.Status.Provisioning.Image = metal3v1alpha1.Image{}
+	info.host.Status.Provisioning.PoweredOnAt = nil
+	info.host.Status.Provisioning.PendingRebootSince = nil
+	if info.host.Annotations != nil {
+		for annotation := range info.host.Annotations {
+			if annotation == rebootAnnotationPrefix || strings.HasPrefix(annotation, rebootAnnotationPrefix+"/") {
```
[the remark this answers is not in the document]
n1r1: Makes sense. added `isRebootAnnotation()`

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -604,22 +646,41 @@ func (r *ReconcileBareMetalHost) manageHostPower(prov provisioner.Provisioner, i
 
 	if provResult.Dirty {
 		info.host.ClearError()
-		return actionContinue{provResult.RequeueAfter}
+		return actionContinue{delay: provResult.RequeueAfter, dirty: true}
+	}
+
+	desiredPowerOnState := info.host.Spec.Online
+
+	if hasRebootAnnotation(info.host) {
```
zaneb: We originally said that we would only pay attention to reboot requests when the server was provisioned, so I think we should also check that we are being called from the Provisioned state (and not the Ready/Available state).
Now that I think about it, this was related to the whole LastBootedAt thing - that enabled us to ensure we didn't bother to reboot a just-provisioned server due to a reboot annotation that predated the provisioning. I think we can handle that by just clearing the annotations again at the start of provisioning (just like we do at the end of deprovisioning).

pkg/controller/baremetalhost/baremetalhost_controller.go
zaneb: If we check the provisioning state on line 654 then we will still need to move this outside that block.

pkg/controller/baremetalhost/baremetalhost_controller.go
n1r1: Do you want me to check that the host is in `ExternallyProvisioned || Provisioned` state or add additional argument to `manageHostPower()` ?

pkg/controller/baremetalhost/action_result.go
```diff
@@ -19,6 +19,7 @@ type actionResult interface {
 // state.
 type actionContinue struct {
 	delay time.Duration
+	dirty bool
```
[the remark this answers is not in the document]
beekhof: I like `actionContinueNoWrite`

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -372,6 +373,41 @@ func (r *ReconcileBareMetalHost) credentialsErrorResult(err error, request recon
 	}
 }
 
+// hasRebootAnnotation checks for existence of reboot annotations and returns true if at least one exist
+func hasRebootAnnotation(host *metal3v1alpha1.BareMetalHost) bool {
+	if len(host.Annotations) == 0 {
+		return false
+	}
```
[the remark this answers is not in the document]
beekhof: I vote to kill it

pkg/controller/baremetalhost/action_result.go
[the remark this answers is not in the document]
n1r1: added `actionContinueNoWrite`

pkg/controller/baremetalhost/baremetalhost_controller.go
n1r1: @zaneb for now I implemented the first option

pkg/controller/baremetalhost/action_result.go
```diff
@@ -14,6 +14,25 @@ type actionResult interface {
 	Dirty() bool
 }
 
+// actionContinueNoWrite is a result indicating that the current action is still
+// in progress, and that the resource should remain in the same provisioning
+// state without writing the status
+type actionContinueNoWrite struct {
+	delay time.Duration
```
zaneb: Instead of duplicating the content of actionContinue (and the `Result()` method), we can just compose it:
```
type actionContinueNoWrite struct {
    actionContinue
}
```
Then you only need to override the `Dirty()` method.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -372,6 +373,37 @@ func (r *ReconcileBareMetalHost) credentialsErrorResult(err error, request recon
...
+// isRebootAnnotation returns true if the provided annotation is a reboot annotation (either suffixed or not)
+func isRebootAnnotation(annotation string) bool{
+	return strings.HasPrefix(annotation, rebootAnnotationPrefix+"/") || annotation == rebootAnnotationPrefix
+}
+
+// clearRebootAnnotations deletes all reboot annotations exist on the provided host
+func clearRebootAnnotations(host *metal3v1alpha1.BareMetalHost) (dirty bool) {
+	if len(host.Annotations) == 0 {
+		return
+	}
```
zaneb: Also redundant.

n1r1: Removd

pkg/controller/baremetalhost/action_result.go
n1r1: nice, done.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -535,6 +563,13 @@ func (r *ReconcileBareMetalHost) actionProvisioning(prov provisioner.Provisioner
 	}
 	info.log.Info("provisioning")
 
+	if clearRebootAnnotations(info.host) {
+		if err := r.client.Update(context.TODO(), info.host); err != nil {
+			return actionError{errors.Wrap(err, "failed to remove reboot annotations from host")}
+		}
+		return actionContinueNoWrite{}
+	}
```
zaneb: I forgot to mention, there is a slight race here in that we will continue to clear reboot annotations up until the provisioning is complete. So if the server completes booting, starts up, and needs remediation before Ironic notices that provisioning is complete, the reboot annotations will be cleared without actually rebooting.
In practice this doesn't seem like a plausible scenario, and it's difficult to know what to do about it, so I'm not too worried.

PR #429, opened 2020-02-13
Fix typo in e2e testing in certificate test

yrobla:
There is a typo for DisableCertificateVerification that
causes the e2e tests to fail.


hardys:
@maelk - looking at https://github.com/metal3-io/baremetal-operator/pull/396 we can see that CI passed - seems we don't run the e2e tests in CI atm?

maelk:
@hardys We are indeed not running  the e2e tests to the best of my knowledge. If they are functional (they were not for quite a long time) we could add a prow job for them.

PR #430, opened 2020-02-14
Re-enable Ironic Fast Track

kashifest:
This PR re-enables  Ironic Fast track.  It was disabled previously due to bug.

### DIFF
--- a/deploy/default/ironic_bmo_configmap.env
+++ b/deploy/default/ironic_bmo_configmap.env
@@ -6,4 +6,4 @@ DEPLOY_RAMDISK_URL=http://172.22.0.2:6180/images/ironic-python-agent.initramfs
 IRONIC_ENDPOINT=http://172.22.0.2:6385/v1/
 IRONIC_INSPECTOR_ENDPOINT=http://172.22.0.2:5050/v1/
 CACHEURL=http://172.22.0.1/images
-IRONIC_FAST_TRACK=false
+IRONIC_FAST_TRACK=true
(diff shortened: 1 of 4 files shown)


maelk:
Since we moved to ironic master, we are hitting again the agent_url bug. We can not re-enable fast-track for now

maelk:
It looks like the fast track problem is still not fixed. We should investigate this deeper

maelk:
The test ran on an env file that was not modified. Please update the PR to change the fast track in all the environment files or run a test that uses on of the modified ones (like test-centos-integration) to check whether fast track works now

maelk:
what about deploy/ironic_ci.env ?

maelk:
This passed the centos tests ??

maelk:
probably a false positive

maelk:
hum, it passed centos again!

maelk:
btw, when deploying in Minikube, this always override the fast track option : https://github.com/metal3-io/metal3-dev-env/blob/master/03_launch_mgmt_cluster.sh#L157

fmuyassarov:
This run is for testing Minikube based CI test
/test-centos-integration

maelk:
btw, for Centos jobs, the fast track is overriden in metal3-dev-env, set to false. It needs to be changed in Metal3-dev-env in that case.

hardys:
I did some testing on fast-track recently and https://github.com/metal3-io/ironic-image/pull/163 was required to make it work, also depending on the timing https://review.opendev.org/#/c/739120 may be needed.  

It wasn't immediately obvious these were required because the deployment still succeeds, but there's an additional reboot which fast-track is supposed to avoid.

maelk:
@hardys We are seeing additional issues in the CI. We need to investigate it, we just did not have the time to focus on this issue yet. Hopefully we'll have some time during the summer

dtantsur (approved):
I'm very much in favour of this.

PR #436, opened 2020-02-20
Look for userData or value in the userdata secret

maelk:
This is to comply with CAPI secrets content in v1alpha3


dhellmann:
This could use a unit test.

maelk:
True, I'll add it in a new PR. thanks

maelk:
@dhellmann I didn't find any unit test existing for this function. Did I miss something ?

dhellmann:
> @dhellmann I didn't find any unit test existing for this function. Did I miss something ?

Ah, no, you're right. We could move the logic into a little function so we could test it, but I'm not sure it's worth that.

PR #438, opened 2020-02-21
Add a Pause annotation for pivoting

maelk:
When pivoting, we need a way to temporarily disable the reconciliation of objects while we are moving them. The BMH status needs to be moved along with the spec, and that is a two-step operation (create the BMH and update the status). If the reconciliation was happening in the meantime, this could trigger an introspection of a deployed host, hence messing up the host. Adding the annotation on the BMH would prevent such issues.

### DIFF
--- a/pkg/controller/baremetalhost/baremetalhost_controller_test.go
+++ b/pkg/controller/baremetalhost/baremetalhost_controller_test.go
@@ -184,6 +184,25 @@ func waitForProvisioningState(t *testing.T, r *ReconcileBareMetalHost, host *met
 	)
 }
 
+// TestPause ensures that the requeue happens when the pause annotation is there.
+func TestPause(t *testing.T) {
+	host := newDefaultHost(t)
+	host.Annotations = map[string]string{
+		metal3v1alpha1.PausedAnnotation: "true",
+	}
+	r := newTestReconciler(host)
+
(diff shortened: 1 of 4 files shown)


pkg/controller/baremetalhost/baremetalhost_controller_test.go
```diff
@@ -184,6 +184,24 @@ func waitForProvisioningState(t *testing.T, r *ReconcileBareMetalHost, host *met
...
+func TestPause(t *testing.T) {
+	host := newDefaultHost(t)
+	host.Annotations = map[string]string{
+		metal3v1alpha1.PausedAnnotation: "true",
+	}
+	r := newTestReconciler(host)
+
+	tryReconcile(t, r, host,
+		func(host *metal3v1alpha1.BareMetalHost, result reconcile.Result) bool {
+			if result.Requeue && result.RequeueAfter == pauseRetryDelay {
```
dhellmann: It would be good to test that some action taken early by reconcile has not been taken for this host, to prove not just that Reconcile() returned the right RequeueAfter value but that it is in fact paused.

maelk: I have added a check on the finalizers that are the next step. There should not be any finalizer.

Issue #439, opened 2020-02-24
Provisioning failing due to missing local-hostname in ironic metadata


maelk:
In some of our deployments, the cloud-init fails due to a missing hostname configuration. the issue is that the metadata does not contain a "local-hostname" key. I would propose to add a local-hostname key with the BMH name as value for now as a quick fix, before allowing for full configuration of the metadata, similar to what was done for the network in #348

maelk:
fixed in #440

PR #440, opened 2020-02-24
add local-hostname in ironic metadata

maelk:
This is needed as a temporary fix for #439 

This adds a metadata key "local-hostname" set to the BMH name passed as metadata by Ironic.


maelk:
/hold
to give some time for reviews

PR #441, opened 2020-02-24
Fix dep check in Makefile

kirankt:
There is no `dep-check` target in the Makefile. This PR changes the target to `dep`


kirankt:
This PR will be obsolete with #367

PR #444, opened 2020-02-27
Go modules, take 2

honza:
This time the go modules migration is separate from upgrading operator-sdk.

This time with more commentary in the commit messages.

cc #367


zaneb:
This is excellent, thanks @honza. If there's any problems this will make it much easier to e.g. bisect and find out the cause.
I'm happy to either merge this or force push to #367.

/lgtm

PR #457, opened 2020-03-20
Don't assume IPv4 in introspection data

hardys:
This can be one of two keys, depending on whether the host
got an ipv4 or ipv6 address, so use the ipv6 key in the case
where the ipv4 one is unset.

Currently in a single-stack ipv6 deployment the IP ends up
empty like:

```
Hostname:     worker-0.ostest.test.metalkube.org
    Nics:
      Ip:
```

But with this change Ip should be correctly populated.

### DIFF
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -477,13 +477,16 @@ func getNICDetails(ifdata []introspection.InterfaceType,
 	for i, intf := range ifdata {
 		baseIntf := basedata[intf.Name]
 		vlans, vlanid := getVLANs(baseIntf)
-
+		ip := intf.IPV4Address
+		if ip == "" {
+			ip = intf.IPV6Address
+		}
 		nics[i] = metal3v1alpha1.NIC{
 			Name: intf.Name,
 			Model: strings.TrimLeft(fmt.Sprintf("%s %s",
 				intf.Vendor, intf.Product), " "),
 			MAC:       intf.MACAddress,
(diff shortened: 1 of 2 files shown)


pkg/provisioner/ironic/ironic.go
```diff
@@ -477,13 +477,16 @@ func getNICDetails(ifdata []introspection.InterfaceType,
 	for i, intf := range ifdata {
 		baseIntf := basedata[intf.Name]
 		vlans, vlanid := getVLANs(baseIntf)
-
+		ip := intf.IPV4Address
+		if ip == "" {
+			ip = intf.IPV6Address
+		}
```
russellb: This seems fine, but it actually highlights another bug that will remain.

Once we try this with a dual-stack IPv4+IPv6 environment, we will now be missing the IPv4 address for an interface.

Maybe we should actually change the API here a bit to be able to list more than one IP address?

russellb: I'm also OK with filing a bug to track fixing dual-stack / multiple IPs in this API.

The PR as-is is an improvement

hardys: Yeah I did wonder about dual stack but as you say it's probably an API change unless we append a duplicate nic with the same name but a different IP?

I'll raise a bug so we can track resolving that, this should at least fix the single stack case so we can make progress.

hardys: https://github.com/metal3-io/baremetal-operator/issues/458

PR #459, opened 2020-03-22
Update pause annotation description

kashifest:
This PR updates the api.md and adds detailed description about how pause annotation for bmh should be handled.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -10,8 +10,14 @@ and its current status.
 ### Pausing reconciliation
 
 It is possible to pause the reconciliation of a BareMetalHost object by adding
-an annotation `baremetalhost.metal3.io/paused`. The value of the annotation does
-not matter. Removing the annotation will enable the reconciliation again.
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -10,8 +10,14 @@ and its current status.
...
 
 It is possible to pause the reconciliation of a BareMetalHost object by adding
-an annotation `baremetalhost.metal3.io/paused`. The value of the annotation does
-not matter. Removing the annotation will enable the reconciliation again.
+an annotation `baremetalhost.metal3.io/paused`. **Metal³**  provider sets the
+value of this annotation as `metal3.io/capM3` when the cluster to which the
+**BareMetalHost** belongs, is paused and removes it when the cluster is
+not paused. If you want to pause the reconciliation of **BareMetalHost** you can
+put any value on this annotation **other than `metal3.io/capM3`**. Please make
+sure that you remove the annotation  **only if the value of the annotation is
```
Xenwar: **only if the value of the annotation is not `metal3.io/capM3, but another value that you have provided`**

PR #463, opened 2020-03-24
Default user and network data namespace to BMH namespace

### LINKED ISSUE #462: Empty namespace in UserData or NetworkData part of the BMH spec defaults to default namespace
If the UserData or NetworkData object references does not have a namespace specified, the Get operation is performed on the default namespace. Should it rather default to the namespace of the BMH ?

maelk:
Empty namespaces in userdata and networkdata object references
now default to the baremetalhost namespace

Fixes #462


dukov:
> /cc @dhellmann
> /cc @mhrivnak
> 
> /cc @dukov This should also be done for the metadata PR. Will you take care of it ?

PR has been updated

PR #469, opened 2020-03-28
Select boot mode automatically

yrobla:
Create a new method for AccessDetails interface, that
will hold the boot_mode property. This property will
be automatically set, depending on bmc type.
Then use this method on ironic provisioner, to
automatically set the boot mode capability for
the host.

### DIFF
--- a/deploy/crds/metal3.io_baremetalhosts_crd.yaml
+++ b/deploy/crds/metal3.io_baremetalhosts_crd.yaml
@@ -505,6 +505,13 @@ spec:
                   description: The machine's UUID from the underlying provisioning
                     tool
                   type: string
+                bootMode:
+                  description: BootMode indicates the boot mode used to provision
+                    the node
+                  enum:
+                  - UEFI
+                  - legacy
+                  type: string
                 image:
                   description: Image holds the details of the last image successfully
                     provisioned to the host.
--- a/pkg/bmc/access.go
+++ b/pkg/bmc/access.go
@@ -44,6 +44,10 @@ type AccessDetails interface {
 	// (such as the kernel and ramdisk locations).
 	DriverInfo(bmcCreds Credentials) map[string]interface{}
 
+	// NodeProperties returns properties of a host, including the boot
+	// mode. This will depend on each interface
+	NodeProperties() map[string]interface{}
+
 	// Boot interface to set
 	BootInterface() string
 
--- a/pkg/bmc/ipmi.go
+++ b/pkg/bmc/ipmi.go
@@ -2,6 +2,8 @@ package bmc
 
 import (
 	"net/url"
+
+	metal3v1alpha1 "github.com/metal3-io/baremetal-operator/pkg/apis/metal3/v1alpha1"
 )
 
 func init() {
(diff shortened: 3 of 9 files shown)


pkg/bmc/ipmi.go
```diff
@@ -73,6 +75,16 @@ func (a *ipmiAccessDetails) DriverInfo(bmcCreds Credentials) map[string]interfac
 	return result
 }
 
+// NodeProperties returns a data structure to return details of
+// the host, including the boot mode. This will be used later to
+// instruct ironic to use specific boot mode
+func (a *ipmiAccessDetails) NodeProperties() map[string]interface{} {
```
yrobla: I think we should have an option here, to discriminate between ipv4/ipv6. And ipv6 should be uefi. Please send feedback on this question.

deploy/crds/metal3.io_baremetalhosts_crd.yaml
```diff
@@ -505,6 +505,13 @@ spec:
                   description: The machine's UUID from the underlying provisioning
                     tool
                   type: string
+                bootMode:
+                  description: BootMode indicates the boot mode used to provision
+                    the node
+                  enum:
+                  - UEFI
```
juliakreger, line 512: So the industry seems to be set on "Uefi", but I suspect this is also okay.

deploy/crds/metal3.io_baremetalhosts_crd.yaml
```diff
@@ -505,6 +505,13 @@ spec:
...
+                  - legacy
```
juliakreger, line 513: "Legacy" as well, in most cases, but this may be fine.

pkg/bmc/ipmi.go
juliakreger: OH! Just saw this comment. I think in practicality terms, yes IPv6 seems to always imply UEFI. I _suspect_ in reality, some hardware supports bios mode with IPv6 and I kind of worry about abstracting additional logic to do things automatically into particular configurations. I guess I'm unsure. @dtantsur @derekhiggins Do you guys have any thoughts on this?

dhellmann: Which address matters, the one the BMC uses or the one the host is going to receive? How do we know the latter when we're going to call this method?

pkg/bmc/access.go
```diff
@@ -44,6 +44,10 @@ type AccessDetails interface {
 	// (such as the kernel and ramdisk locations).
 	DriverInfo(bmcCreds Credentials) map[string]interface{}
 
+	// NodeProperties returns properties of a host, including the boot
+	// mode. This will depend on each interface
+	NodeProperties() map[string]interface{}
```
dhellmann, line 49: I had originally thought we would return a map[string]interface{} because the driver would return data to be passed to Ironic directly. If we're not going to do that, let's define a real type to return so that the caller doesn't have to do the type assertion/conversion themselves.

yrobla: you are right, having an ipv6 on the bmc doesn't mean that the provision is going to be done through ipv6. How do you think we can sort that? Again we come to the need of exposing the parameter?

juliakreger:
I guess I'm wondering why. Why in that we're trying to take a path to automatically determine the desired boot mode? Is BIOS mode just that attractive to people to continue to use?

I have strongly stated UEFI should be our preference moving forward, but I know I've gotten some push back from those that move machines between the two modes often. I guess I'm still wondering, what is the business case for supporting BIOS mode?

yrobla:
I guess the main argument is to don't exclude people with hardware needing legacy, or which UEFI don't work properly.

hardys:
I believe this can be closed as it's been replaced by https://github.com/metal3-io/baremetal-operator/pull/586

PR #470, opened 2020-03-30
Put BMH status in annotation while reconciliation

kashifest:
This PR is to make sure the status is put in annotation and it is retrieved also from annotation during reconciliation.

Fixes#431


pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -198,6 +200,22 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (result re
...
+	// Convert the yaml into a typed object
+	obj := &metal3v1alpha1.BareMetalHost{}
+	codecFactory := serializer.NewCodecFactory(r.scheme)
+
+	if err := runtime.DecodeInto(codecFactory.UniversalDecoder(), content, obj); err != nil {
+		reqLogger.Info("failed to decode annotation %v", err)
+	}
+
+	// Manually restore data.
+	if !reflect.DeepEqual(host.Status, obj.Status) {
```
maelk: This should be done only if the status is empty and the annotation is not empty

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -768,10 +802,10 @@ func (r *ReconcileBareMetalHost) actionManageReady(prov provisioner.Provisioner,
 	return r.manageHostPower(prov, info)
 }
 
-func (r *ReconcileBareMetalHost) saveStatus(host *metal3v1alpha1.BareMetalHost) error {
+func (r *ReconcileBareMetalHost) saveHost(host *metal3v1alpha1.BareMetalHost) error {
 	t := metav1.Now()
 	host.Status.LastUpdated = &t
-	return r.client.Status().Update(context.TODO(), host)
+	return r.client.Update(context.TODO(), host)
```
maelk: this probably will not update the status

pkg/controller/baremetalhost/baremetalhost_controller.go
kashifest: I am not sure if I understood you correctly. what about when both are not empty and they have different values

maelk: discard the annotation then. The annotation should only be used in case we loose the status IMO

kashifest: How will you make sure then that after pivoting status is grabbed from annotation automatically?

pkg/controller/baremetalhost/baremetalhost_controller.go
kashifest: changed it.

pkg/controller/baremetalhost/baremetalhost_controller.go
maelk: Check if the status is empty. If so, and you have an annotation, you can use it to populate the status again. Populating the status should not be something done on regular basis.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -259,10 +277,30 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (result re
...
+		infoNew, ok := runtime.SerializerInfoForMediaType(codecFactory.SupportedMediaTypes(), mediaType)
+		if !ok {
+			info.log.Info("Failed to Serialize")
+		} else {
+			encoder := codecFactory.EncoderForVersion(infoNew.Serializer, metal3v1alpha1.SchemeGroupVersion)
+			newAnnotation, _ := runtime.Encode(encoder, runtime.Object(host))
+			if host.Annotations == nil {
+				host.Annotations = make(map[string]string)
+			}
+			host.Annotations["kubectl.kubernetes.io/last-applied-configuration"] = string(newAnnotation)
```
maelk: That is not the correct annotation to use. you will conflict with Kubectl. Add a new one that is metal3 specific to store the status.

pkg/controller/baremetalhost/baremetalhost_controller.go
kashifest: You are right, we must do it only when needed, I am rechecking the code.

pkg/controller/baremetalhost/baremetalhost_controller.go
kashifest: Yes. I have a plan to do that. I am using it as a trial basis

kashifest:
@maelk what about putting the status back to annotation?

maelk:
I don't get what you mean, sorry.

dhellmann:
Is there a design document somewhere explaining how this is used? I'm surprised to see operator code doing anything other than reading and clearing the annotation one time at the top of the reconcile loop.

If we have this operator copying data back and forth between the annotation and the status field, we have to figure out which is the source of truth. If, instead, we say that the operator will initialize its status from an annotation, then it is possible for an outside entity to construct a host resource with the spec field in and the annotation present and end up with a fully populated host resource in the new cluster.

kashifest:
> Is there a design document somewhere explaining how this is used?

There is no design document as per say. The idea is to populate the status field from annotation if it is empty. But as you see , the annotation and status needs to be in sync, thats why the annotation is being updated whenever the status is updated. If we do this here, tools like clusterctl would be able to move the bmh to target cluster, without the need of an outside entity to construct the object from spec and annotation.

dhellmann:
> > Is there a design document somewhere explaining how this is used?
> 
> There is no design document as per say. The idea is to populate the status field from annotation if it is empty. But as you see , the annotation and status needs to be in sync, thats why the annotation is being updated whenever the status is updated. If we do this here, tools like clusterctl would be able to move the bmh to target cluster, without the need of an outside entity to construct the object from spec and annotation.

OK. I think you're going to have to do enough work to compare the status contents and annotation contents to avoid causing an infinite loop in the reconciliation logic.

I wonder if we could have 1 controller copy the status values for all metal3 resources into an annotation on the CR. Then the controllers that own each resource type would only have to copy the annotation content into the status fields if the status is empty.

kashifest:
> OK. I think you're going to have to do enough work to compare the status contents and annotation contents to avoid causing an infinite loop in the reconciliation logic.

Do you think the current way of copying the status to annotation would create an infinite loop in reconciliation? I am also thinking how to solve this issue. One way could be to use the pause annotation or passing some flag value to indicate the reconciler to do the copy.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -198,6 +200,23 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (result re
...
+
+	content := []byte(annotations[metal3v1alpha1.StatusAnnotation])
+	// Convert the yaml into a typed object
+	obj := &metal3v1alpha1.BareMetalHost{}
+	codecFactory := serializer.NewCodecFactory(r.scheme)
+
+	if err := runtime.DecodeInto(codecFactory.UniversalDecoder(), content, obj); err != nil {
+		reqLogger.Info("failed to decode annotation %v", err)
+	}
+
```
maelk, line 213: this should be probably in the condition. We don't want to do that if we know we will not use the output.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -768,10 +789,37 @@ func (r *ReconcileBareMetalHost) actionManageReady(prov provisioner.Provisioner,
...
+func (r *ReconcileBareMetalHost) saveHostStatus(host *metal3v1alpha1.BareMetalHost) error {
 	t := metav1.Now()
 	host.Status.LastUpdated = &t
-	return r.client.Status().Update(context.TODO(), host)
+
+	errStatus := r.client.Status().Update(context.TODO(), host)
+
+	if err := r.saveHostAnnotation(host); err != nil {
+		return err
+	}
```
maelk, line 800: I would do it the other way around, first updating the annotation, then the status, so that there is no point when the annotation would be behind the status.

kashifest: This is not viable anymore .... we cannot afford to sync status and status annotation all the time because, as soon as you modify an object status the lastUpdated field changes and triggers a change in annotation and that his goes again back and forth. We have to somehow pin the changes when there is a change in bmh state or change in error message

kashifest:
@maelk @dhellmann I would request one more round of reviews from you. The idea is now:
1. Set the status from annotation only if status is empty and status annotation is not.
2. re-populate the status annotation only when there is a significant change is status, which is either a change in Provisioning.State or Error.Message. If we try to keep it synced all the time, it is making an infinite changes because of the lastupdated field being updated by one of the updates.

I would appreciate comments, suggestions.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -198,6 +200,15 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (result re
 		}
 	}
 
+	// re-apply host status from annotation
+
+	obj := r.getHostFromAnnotation(host)
+	// Manually restore data.
```
maelk, line 206: this should be done only if you know that you do not have a status. It should be in line 209

pkg/controller/baremetalhost/baremetalhost_controller.go
maelk: I got that, it is in the next function. My point is that you should change the order of the calls line 787 and 789, so that when there is a change that needs to be saved, we first update the annotation, then the status.

kashifest: Would you please shed some more light what is the motivation to change the order? There will always be an operation lagging behind the other at a specific point and the function call doesn't return before the end of both the operation. I am failing to see the benefit somehow.

pkg/controller/baremetalhost/baremetalhost_controller.go
kashifest: True. Changed it.

pkg/controller/baremetalhost/baremetalhost_controller.go
maelk: It is not a problem to have the status lagging behind the annotation, because it will be updated at the next reconciliation, that's how it is supposed to work. However, if the annotation is lagging behind, and the move happens right then, you will loose the diff, since you won't be able to update the annotation the next time from the ironic data, as it will be gone, hence, you might recreate the node in ironic with wrong state.

It is mostly a theoretical point, since we should not attempt anything if the nodes are not in a stable state (i.e. an operation is ongoing). For example if we pause the reconciliation while something is happening on the ironic side, we will not get the new status, and we will discard the ironic database during the move. In practice, this is not likely to happen considering that during the move. There are two cases. 

The first one is that when we did init on the target cluster, the initial ironic kept the VIP. In that case, even after the move, the initial database will still be present. In that case we are somehow resilient to problems since after moving the BMH, they will be reconciled with the initial ironic database still present. In order to verify that, you could try to discard the status of a BMH in metal3-dev-env to see what happens.

 In the second case, the target ironic took the VIP. that's when we hit problems. If something is happening on the nodes, then they will contact an ironic that is empty. this will probably cause issues in itself. But in addition, the annotation being behind, the node recreated in ironic db afterwards will not be in the correct state wrt the physical node. That would probably end badly. This problem is likely to also happen if the ironic containers crash without using a PV while some nodes are undergoing operations like provisioning, deprovisioning etc.
 
So out of this:
- we should update the annotation first
- Ironic should be managed independently and deployed with PV if possible on the target cluster. The nice point being that if we handle the annotation properly, and do the move when the objects are stable, ironic could be deployed with a simple kustomization.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -198,6 +200,17 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (result re
...
 	}
 
+	// Check if Status is empty and status annotation is present
+	// Manually restore data.
+	if !r.hostHasStatus(host) {
+		obj := r.getHostFromAnnotation(host)
+		if obj != nil {
+			host.Status = obj.Status
+		} else {
+			reqLogger.Info("Failed to find status annotation. Couldn't store status data manually")
```
maelk, line 210: manually ? what does this mean?

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -768,10 +783,73 @@ func (r *ReconcileBareMetalHost) actionManageReady(prov provisioner.Provisioner,
...
+	}
+
+	//Refetch host again
+	err := r.client.Get(context.TODO(),
+		client.ObjectKey{
+			Name:      host.Name,
+			Namespace: host.Namespace,
+		},
+		host,
+	)
```
maelk, line 798: can you please explain why this is needed ?

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -768,10 +783,73 @@ func (r *ReconcileBareMetalHost) actionManageReady(prov provisioner.Provisioner,
...
+func (r *ReconcileBareMetalHost) saveHostAnnotation(host *metal3v1alpha1.BareMetalHost) error {
+	//Repopulate annotation again
+	obj := r.getHostFromAnnotation(host)
+	if obj == nil {
+		return errors.New("Failed to get Host from Status annotation")
+	}
+
+	obj.Status.LastUpdated = host.Status.LastUpdated
+
+	if host.Status != obj.Status {
```
maelk, line 820: you might need a deepEqual kind of check here

kashifest: Done

pkg/controller/baremetalhost/baremetalhost_controller.go
kashifest: This is needed to get rid of the object not found error. To remove this we need a code re-design and the use of capi's PatchHelper type functionality as you suggested.

pkg/controller/baremetalhost/baremetalhost_controller.go
kashifest: I have rephrased it a bit. Since we are putting the status back manually not the reconciler thats why I am using 'manually'

pkg/controller/baremetalhost/baremetalhost_controller.go
kashifest: I have changed the order . Thanks @maelk for the detailed explanation. The current code now puts the bmh in the annotation. I have tried to use other means like json.UnMarshal as you suggested but it didn't work out of the box. It requires redesign of the code. The current version is able to restore the status correctly. I will look into how to get it working only with putting status in annotation in future.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -768,10 +784,72 @@ func (r *ReconcileBareMetalHost) actionManageReady(prov provisioner.Provisioner,
...
 	host.Status.LastUpdated = &t
-	return r.client.Status().Update(context.TODO(), host)
+
+	if err := r.saveHostAnnotation(host); err != nil {
+		return err
+	}
+
+	//Refetch host again
+	obj := host.Status.DeepCopy()
+	err := r.client.Get(context.TODO(),
```
dhellmann, line 797: Does this work? I thought the client went to the cache and the cache was updated right before Reconcile() is called.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -198,6 +201,17 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (result re
 		}
 	}
 
+	// Check if Status is empty and status annotation is present
+	// Manually restore data.
+	if !r.hostHasStatus(host) {
+		obj := r.getHostFromAnnotation(host)
+		if obj != nil {
+			host.Status = obj.Status
```
dhellmann, line 209: What causes the host to be stored after this? Shouldn't we be looking at the value of the annotation and the host value, updating host only if there are changes we need, then storing and returning immediately?

kashifest:
> It seems like there should be more complexity in restoring values from the annotation and less in putting data into the annotation. Would this be more robust and clearer if the annotation is updated every time the host is marked as dirty and the values are only copied from the annotation to the status block if the annotation content is different?

Thanks for the the review Doug @dhellmann . Actually now the restoring of status from annotation is very restrictive. This happens only when the bmh has not status but there is annotation present with the status. These two cases are checked [here](https://github.com/metal3-io/baremetal-operator/blob/2995b6ad984be4a3779df6bde5d878ad78ad1e13/pkg/controller/baremetalhost/baremetalhost_controller.go#L206) and [here](https://github.com/metal3-io/baremetal-operator/blob/2995b6ad984be4a3779df6bde5d878ad78ad1e13/pkg/controller/baremetalhost/baremetalhost_controller.go#L208). And as forthe annotation update it is called inside the saveHostStatus() function which essentially makes sure that the annotation is only updated when there is a change in status.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -198,6 +201,17 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (result re
...
+		} else {
+			reqLogger.Info("Failed to find status annotation. Couldn't restore status data manually")
```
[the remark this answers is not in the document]
kashifest: Since we are placing the status on the BMH when the status is missing and the annotation is present, I thought manually seems a logical word, but I will change it since it is creating confusion. And also the "Failed" message will be removed.

pkg/controller/baremetalhost/baremetalhost_controller.go
kashifest: Yes, because the cache was updated in saveHostAnnotation() function once to save the annotation is called in line 791.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -768,10 +784,72 @@ func (r *ReconcileBareMetalHost) actionManageReady(prov provisioner.Provisioner,
...
+		codecFactory := serializer.NewCodecFactory(r.scheme)
+		mediaType := "application/json"
+		infoNew, ok := runtime.SerializerInfoForMediaType(codecFactory.SupportedMediaTypes(), mediaType)
+
+		if !ok {
+			return errors.New("Failed to update Status annotation")
+		}
+
+		encoder := codecFactory.EncoderForVersion(infoNew.Serializer, metal3v1alpha1.SchemeGroupVersion)
+		newAnnotation, _ := runtime.Encode(encoder, runtime.Object(host))
```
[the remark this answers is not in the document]
kashifest: Yes. Rightly pointed out. That was the motivation. But the runtime.Encode() function requires the obj to have a DeepCopyObject() method and bmhStatus doesn't have the DeepCopyObject() method. So as a workaround the whole bmh is put now. I am planning to update it once I have a solution.

pkg/controller/baremetalhost/baremetalhost_controller.go
kashifest: As, pointed in one of my previous reply, this point is only reached when the status is empty on bmh but the annotation isn't. And storing is caused by the normal reconciliation workflow which checks whether the bmh status is dirty or not.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -768,10 +784,72 @@ func (r *ReconcileBareMetalHost) actionManageReady(prov provisioner.Provisioner,
...
+		if host.Annotations == nil {
+			host.Annotations = make(map[string]string)
+		}
+		host.Annotations[metal3v1alpha1.StatusAnnotation] = string(newAnnotation)
+		return r.client.Update(context.TODO(), host.DeepCopy())
```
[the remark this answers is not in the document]
kashifest: If you are asking about the DeepCopy(), the reason is if we do not use DeepCopy() we will loose the status.

pkg/controller/baremetalhost/baremetalhost_controller.go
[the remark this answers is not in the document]
kashifest: Changed it

pkg/controller/baremetalhost/baremetalhost_controller.go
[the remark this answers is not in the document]
kashifest: Changed it. Now its only storing the status

maelk:
@kashifest is this still WIP ?

kashifest:
> @kashifest is this still WIP ?

No.

kashifest:
Apparently there is still a bug. I will put a hold on it again, until that is fixed. 
/hold

kashifest:
The bug has been fixed, a test case has also been added.
/hold cancel

pkg/controller/baremetalhost/baremetalhost_controller_test.go
```diff
@@ -192,6 +192,30 @@ func waitForProvisioningState(t *testing.T, r *ReconcileBareMetalHost, host *met
...
...",\"fsgsbase\",\"fxsr\",\"hypervisor\",\"lahf_lm\",\"lm\",\"mca\",\"mce\",\"mmx\",\"msr\",\"mtrr\",\"nopl\",\"nx\",\"pae\",\"pat\",\"pclmulqdq\",\"pge\",\"pni\",\"popcnt\",\"pse\",\"pse36\",\"rdrand\",\"rdtscp\",\"rep_good\",\"sep\",\"smep\",\"sse\",\"sse2\",\"sse4_1\",\"sse4_2\",\"ssse3\",\"syscall\",\"tpr_shadow\",\"tsc\",\"tsc_adjust\",\"tsc_deadline_timer\",\"vme\",\"vmx\",\"vnmi\",\"vpid\",\"x2apic\",\"xsave\",\"xsaveopt\",\"xtopology\"],\"count\":4},\"hostname\":\"node-0\"},\"provisioning\":{\"state\":\"provisioned\",\"ID\":\"8a0ede17-7b87-44ac-9293-5b7d50b94b08\",\"image\":{\"url\":\"bar\",\"checksum\":\"\"}},\"goodCredentials\":{\"credentials\":{\"name\":\"node-0-bmc-secret\",\"namespace\":\"metal3\"},\"credentialsVersion\":\"879\"},\"triedCredentials\":{\"credentials\":{\"name\":\"node-0-bmc-secret\",\"namespace\":\"metal3\"},\"credentialsVersion\":\"879\"},\"errorMessage\":\"\",\"poweredOn\":true,\"operationHistory\":{\"register\":{\"start\":\"2020-04-15T12:06:26Z\",\"end\":\"2020-04-15T12:07:12Z\"},\"inspect\":{\"start\":\"2020-04-15T12:07:12Z\",\"end\":\"2020-04-15T12:09:29Z\"},\"provision\":{\"start\":null,\"end\":null},\"deprovision\":{\"start\":null,\"end\":null}}}",
```
dhellmann, line 200: This would be easier to read if the literal string used backticks so we could avoid escaping all of the quotes.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -198,6 +200,22 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (result re
...
 
+	// Check if Status is empty and status annotation is present
+	// Manually restore data.
+	if r.hostHasStatus(host) {
+		reqLogger.Info("Fetching Status from Annotation")
+		objStatus := r.getHostStatusFromAnnotation(host)
+		if objStatus != nil {
+			host.Status = *objStatus
+			errStatus := r.client.Status().Update(context.TODO(), host)
+			if errStatus == nil {
```
dhellmann, line 211: Why not return if there is an error? Shouldn't we loop and try again? Should the error at least be logged?

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -198,6 +200,22 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (result re
...
+				return reconcile.Result{Requeue: true, RequeueAfter: 0}, nil
```
dhellmann, line 212: We don't need to explicitly requeue here because the status update will trigger that.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -768,10 +788,62 @@ func (r *ReconcileBareMetalHost) actionManageReady(prov provisioner.Provisioner,
...
+			Namespace: host.Namespace,
+		},
+		host,
+	)
+	if err != nil {
+		return errors.Wrap(err, "Failed to update Status annotation")
+	}
+	host.Status = *obj
+	errStatus := r.client.Status().Update(context.TODO(), host)
+	return errStatus
```
dhellmann, line 813: Should this error be wrapped like the one on line 809? Why use a separate error variable here instead of reusing err?

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -768,10 +788,62 @@ func (r *ReconcileBareMetalHost) actionManageReady(prov provisioner.Provisioner,
...
+}
+
+func (r *ReconcileBareMetalHost) saveHostAnnotation(host *metal3v1alpha1.BareMetalHost) error {
+	//Repopulate annotation again
+	objStatus := r.getHostStatusFromAnnotation(host)
+	if objStatus != nil {
+		objStatus.LastUpdated = host.Status.LastUpdated
```
dhellmann, line 820: A comment explaining that we're copying this value to avoid continually updating the annotation would be good.

This condition also could be reversed so that if objStatus == nil this function just returns.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -768,10 +788,62 @@ func (r *ReconcileBareMetalHost) actionManageReady(prov provisioner.Provisioner,
...
+	}
+
+	if !reflect.DeepEqual(host.Status, objStatus) {
```
dhellmann, line 823: The usual pattern in go is to prefer early exits so that the "happy path" through a function is not inside a conditional. So, if you reverse this test then when the values are equal this function should return nil. Then lines 824-833 can move outside of a conditional and be the main body of the function.

kashifest: Done

pkg/controller/baremetalhost/baremetalhost_controller.go
kashifest: Done

pkg/controller/baremetalhost/baremetalhost_controller.go
kashifest: Done

pkg/controller/baremetalhost/baremetalhost_controller.go
kashifest: Done

pkg/controller/baremetalhost/baremetalhost_controller.go
kashifest: Done

pkg/controller/baremetalhost/baremetalhost_controller_test.go
kashifest: Done

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -768,10 +788,66 @@ func (r *ReconcileBareMetalHost) actionManageReady(prov provisioner.Provisioner,
...
+	err = r.client.Status().Update(context.TODO(), host)
+	return err
+}
+
+func (r *ReconcileBareMetalHost) saveHostAnnotation(host *metal3v1alpha1.BareMetalHost) error {
+	//Repopulate annotation again
+	objStatus := r.getHostStatusFromAnnotation(host)
+	if objStatus == nil {
+		return nil
+	}
```
maelk, line 821: Shouldn't this return an error ? maybe you should modify the getHostStatusFromAnnotation to forward the error upwards.

kashifest: This function is called in two different places and a check of nil is enough IMO. Returning error will unnecessarily complicate the logic and the tests. Cases like no status-annotation initially would throw an error and we should not stop at that error but continue.

maelk: but what about a malformed annotation? It should block the processing IMO

kashifest: What do mean by malformed annotation? Can you elaborate?

kashifest: Got your point @maelk. I have updated the code accordingly

pkg/controller/baremetalhost/baremetalhost_controller_test.go
```diff
@@ -192,6 +192,29 @@ func waitForProvisioningState(t *testing.T, r *ReconcileBareMetalHost, host *met
 	)
 }
 
+// TestStatusAnnotation ensures that status is manually populated when status annotation is there.
+func TestStatusAnnotation(t *testing.T) {
```
dhellmann, line 196: It would be good to have some logic to test the other cases, like to ensure the annotation is populated correctly after a reconcile and to ensure that the status isn't overwritten by the annotation if the status isn't empty.

kashifest: @dhellmann  I have added test cases for the logics you suggested. Please review

pkg/controller/baremetalhost/baremetalhost_controller_test.go
```diff
@@ -192,6 +195,82 @@ func waitForProvisioningState(t *testing.T, r *ReconcileBareMetalHost, host *met
...
+	r := newTestReconciler(host, bmcSecret)
+
+	tryReconcile(t, r, host,
+		func(host *metal3v1alpha1.BareMetalHost, result reconcile.Result) bool {
+			if utils.StringInList(host.Finalizers, metal3v1alpha1.BareMetalHostFinalizer) {
+				return true
+			}
+			return false
+		},
+	)
```
maelk, line 258: why are you testing the finalizers ? Is it just so that you go through the loop twice because the first time it only sets the finalizer ? If so, shouldn't this be the case also for the other function, i.e. are you sure that the previous function is actually testing the function you intend to test ? I have a suspicion that it is not the case as you have two different ways to do things in two functions that should be essentially the same.
Also, you are not actually checking that the status was not modified. what this does is testing that the annotation matches the status, independently of the starting point of the annotation.

Issue #464, opened 2020-03-25
--namespace cli argument needed for BMO


maelk:
In order to fulfill the provider contract, BMO must accept a --namespace cli argument instead of using a WATCH_NAMESPACE environment variable : https://cluster-api.sigs.k8s.io/clusterctl/provider-contract.html

We would propose to not fail if the environment variable WATCH_NAMESPACE is unset, and override it with --namespace if provided in the CLI.

fmuyassarov:
isn't --namespace option already there?

```
Flags:
      --enable-delve            Start the operator using the delve debugger
      --go-ldflags string       Set Go linker options
  -h, --help                    help for local
      --kubeconfig string       The file path to kubernetes configuration file; defaults to location specified by $KUBECONFIG with a fallback to $HOME/.kube/config if not set
      --namespace string        The namespace where the operator watches for changes.
      --operator-flags string   The flags that the operator needs. Example: "--flag1 value1 --flag2=value2"
```

maelk:
Those are the operator-sdk options, not the manager options

dhellmann:
The namespace to use is determined in the main program for the controller at https://github.com/metal3-io/baremetal-operator/blob/master/cmd/manager/main.go#L45

fmuyassarov:
Fix in  #471

PR #471, opened 2020-03-30
Add --namespace cli argument support for BMO

fmuyassarov:
Fix for #464

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -20,7 +20,10 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 )
 
-var log = logf.Log.WithName("cmd")
+var (
+	log                     = logf.Log.WithName("cmd")
+	watchNamespace          string
+)
 
 func printVersion() {
 	log.Info(fmt.Sprintf("Go Version: %s", runtime.Version()))
@@ -30,8 +33,10 @@ func printVersion() {
 }
 
 func main() {
+	var err error
 	devLogging := flag.Bool("dev", false, "enable dev logging")
 	metricsAddr := flag.String("metrics-addr", "127.0.0.1:8085", "The address the metric endpoint binds to.")
+	flag.StringVar(&watchNamespace, "namespace", "", "Namespace that the controller watches to reconcile BMO objects.")
 	flag.Parse()
 
 	// The logger instantiated here can be changed to any logger
@@ -42,10 +47,15 @@ func main() {
 
 	printVersion()
 
-	namespace, err := k8sutil.GetWatchNamespace()
-	if err != nil {
-		log.Error(err, "Failed to get watch namespace")
-		os.Exit(1)
+	// From CAPI point of view, BMO should be able to watch all namespaces
+	// in case of a deployment that is not multi-tenant. If the deployment
+	// is for multi-tenancy, then the BMO should watch only the provided
+	// namespace.
+	if watchNamespace == "" {
+		watchNamespace, err = k8sutil.GetWatchNamespace()
+		if err != nil {
+			watchNamespace = ""
+		}
 	}
 
 	// Get a config to talk to the apiserver
@@ -59,8 +69,8 @@ func main() {
 	opts := manager.Options{
 		LeaderElection:          true,
 		LeaderElectionID:        "baremetal-operator",
-		LeaderElectionNamespace: namespace,
-		Namespace:               namespace,
+		LeaderElectionNamespace: watchNamespace,
+		Namespace:               watchNamespace,
 		MetricsBindAddress:      *metricsAddr,
 	}
 


cmd/manager/main.go
```diff
@@ -42,10 +46,12 @@ func main() {
 
 	printVersion()
 
-	namespace, err := k8sutil.GetWatchNamespace()
-	if err != nil {
-		log.Error(err, "Failed to get watch namespace")
-		os.Exit(1)
+	// override WATCH_NAMESPACE with --namespace value if provided by CLI
+	if watchNamespace == "" {
+		watchNamespace, err = k8sutil.GetWatchNamespace()
```
maelk: err is undefined, you need to declare it before, or use := (that might be problematic here for the linters)

cmd/manager/main.go
```diff
@@ -59,8 +65,8 @@ func main() {
 	opts := manager.Options{
 		LeaderElection:          true,
 		LeaderElectionID:        "baremetal-operator",
-		LeaderElectionNamespace: namespace,
-		Namespace:               namespace,
+		LeaderElectionNamespace: 	watchNamespace,
+		Namespace:               	watchNamespace,
```
maelk: indentation

cmd/manager/main.go
fmuyassarov: yes, forgot that to add. Thanks

cmd/manager/main.go
fmuyassarov: resolved

cmd/manager/main.go
maelk: maybe at the beginning of the main function would be better, since it is used throughout the function

dhellmann: Why remove the error handling here? If no namespace is provided, don't we want to exit with an error?

fmuyassarov: Because if no namespace is provided then we want to watch cluster-wise (across all the namespaces).

dhellmann: I thought we agreed to add the flag but still require a namespace. Why do we want the operator to ignore namespaces?

fmuyassarov: If I'm not missing anything, we first check for the --namespace value in CLI and if given then override WATCH_NAMESPACE value. If --namespace not passed we check for WATCH_NAMESPACE and if not provided we watch across all namespace, no?

dhellmann: You did correctly describe the behavior, as I understand it. I am surprised that we are allowing that fallback of watching all namespaces, though. I thought we said we would let the command line option override the environment variable, but exit with an error if neither were provided. Perhaps I misunderstood?

What situation do you have that makes it useful to watch all namespaces?

fmuyassarov: I believe we want it to able to watch across all the namespaces in case if a user wants to create BMH in a any other namespace apart from the one where the operator itself is running. As I understood, BMO is going to be part of CAPM3, meaning that it will need to able to look for objects to reconcile in any other namespace provided by `–namespace,`  in order to fulfill the [provider contract](https://cluster-api.sigs.k8s.io/clusterctl/provider-contract.html).

And, I'm not sure if we want to exit with an error if namespace is not provided in neither places but instead I believe we need to watch across all the namespace.

maelk: From CAPI point of view, BMO should be able to watch all namespaces in case of a deployment that is not multi-tenant. If the deployment is for multi-tenancy, then the BMO should watch only the provided namespace. Both should be supported.

dhellmann: OK, that's good information. We should capture it as part of this commit. Could you update the comment on line 50 to explain the error handling?

fmuyassarov: Done.

README.md
```diff
@@ -1,7 +1,5 @@
 # Bare Metal Operator
 
-[![Ubuntu V1alpha2 build status](https://jenkins.nordix.org/view/Airship/job/airship_master_v1a2_integration_test_ubuntu/badge/icon?subject=Ubuntu%20E2E%20V1alpha2)](https://jenkins.nordix.org/view/Airship/job/airship_master_v1a2_integration_test_ubuntu)
-[![CentOS V1alpha2 build status](https://jenkins.nordix.org/view/Airship/job/airship_master_v1a2_integration_test_centos/badge/icon?subject=CentOS%20E2E%20V1alpha2)](https://jenkins.nordix.org/view/Airship/job/airship_master_v1a2_integration_test_centos)
```
dhellmann: Is this change related?

fmuyassarov: It is totally different thing, not related to the actual PR.  
It is just that we have dropped v1alpha2 CI now, so there will not be any CI status reported for v1alpha3 integration tests. 
So, I just wanted to fix it with the same PR as mentioned in the commit message above.

fmuyassarov: Do you want to do it on a separate PR ?

fmuyassarov: I have removed unrelated changes. Will open a separate PR for that.

Issue #472, opened 2020-03-30
BMO should have kustomize config for test setup with BMO standalone


mikkosest:
When using kind cluster in metal3-dev-env, Ironic needs to run outside cluster. It would be handy to have similar kustomize config for BMO standalone than there is for baremetal-operator/deploy/ironic-keepalived-config. Needed by  metal3-io/metal3-dev-env#277

maelk:
Do you mean https://github.com/metal3-io/metal3-dev-env/issues/277 ?

mikkosest:
> Do you mean [metal3-io/metal3-dev-env#277](https://github.com/metal3-io/metal3-dev-env/pull/277) ?

Corrected

maelk:
Is this closed by #473  ? @mikkosest

dhellmann:
See #500 and #501

maelk:
I would think this is resolved. Please re-open if not.

PR #473, opened 2020-03-30
New kustomize config for BMO standalone, needed with kind when ironic…

mikkosest:
… outside cluster. This commit contains config that can be used when kind is wanted to be used as option for minikube

Completes #472

### DIFF
--- a/deploy/ironic-outside-config/ironic_bmo_configmap.env
+++ b/deploy/ironic-outside-config/ironic_bmo_configmap.env
@@ -1,0 +1,9 @@
+HTTP_PORT=6180
+PROVISIONING_INTERFACE=provisioning
+DHCP_RANGE=172.22.0.10,172.22.0.100
+DEPLOY_KERNEL_URL=http://172.22.0.1:6180/images/ironic-python-agent.kernel
+DEPLOY_RAMDISK_URL=http://172.22.0.1:6180/images/ironic-python-agent.initramfs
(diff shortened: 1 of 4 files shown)


maelk:
You probably don't need to create this ironic_outside folder. There is already a no_ironic folder in the operator folder. I believe this would be enough

mikkosest:
> You probably don't need to create this ironic_outside folder. There is already a no_ironic folder in the operator folder. I believe this would be enough
Fixed

mikkosest:
> looks good.
> btw, what is the effect of having IRONIC_FAST_TRACK set to false (was true)?
IRONIC_FAST_TRACK is currently broken so it should be set false everywhere

deploy/ironic_ci.env
```diff
@@ -6,4 +6,4 @@ DEPLOY_RAMDISK_URL=http://172.22.0.1:6180/images/ironic-python-agent.initramfs
 IRONIC_ENDPOINT=http://172.22.0.1:6385/v1/
 IRONIC_INSPECTOR_ENDPOINT=http://172.22.0.1:5050/v1/
 CACHEURL=http://172.22.0.1/images
-IRONIC_FAST_TRACK=true
+IRONIC_FAST_TRACK=false
```
maelk: Could the script that starts ironic outside of the cluster rather use deploy/ironic-outside-config/ironic_bmo_configmap.env ? that would avoid duplication

mikkosest: Fixed

PR #474, opened 2020-03-31
Fix make test phony to remove dep-check call

kashifest:
The `make test`  still calls for dep-check which was removed previously. The call to dep-check in Makefile was accidentally left-behind.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,7 @@ help:
 	@echo "  DEBUG            -- debug flag, if any ($(DEBUG))"
 
 .PHONY: test
-test: generate unit lint dep-check
+test: generate unit lint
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -37,7 +37,7 @@ help:
 	@echo "  DEBUG            -- debug flag, if any ($(DEBUG))"
 
 .PHONY: test
-test: generate unit lint dep-check
+test: generate unit lint
```
maelk: Maybe you should also remove the other dep * targets

jan-est: Agree all dep targets should be removed as unused.

kashifest: done

PR #478, opened 2020-04-02
Default missing variables

kashifest:
This PR adds and defaults the missing variables in `ironic_bmo_configmap.env`. The missing variables are `PROVISIONING_IP` and `PROVISIONING_CIDR`. The keepalived container depends on these two variables. Defaulting it makes sure that, if someone wants to use bmo kustomized  manually, not through metal3-dev-env, then also keepalived container is able to attach the default ironic endpoint ip to provisioning interface.


deploy/ironic-keepalived-config/ironic_bmo_configmap.env
```diff
@@ -1,4 +1,6 @@
 HTTP_PORT=6180
+PROVISIONING_IP=172.22.0.2
```
fmuyassarov, line 2: Should PROVISIONING_IP be 172.22.0.2 or 172.22.0.0/24 ?

kashifest: no it is correct. The cidr is added later.

fmuyassarov: ok, one more question, why is the IP not 172.22.0.1, because if I check [CI logs ](https://jenkins.nordix.org/view/Airship/job/airship_master_v1a3_integration_test_ubuntu/99/consoleFull)I see 
`+++ PROVISIONING_IP=172.22.0.1`

kashifest: Unfortunately the names are same, we should fix the names also. The variable that I am setting is set here, https://github.com/metal3-io/metal3-dev-env/blob/master/03_launch_mgmt_cluster.sh#L108 which in turn is set here https://github.com/metal3-io/metal3-dev-env/blob/c197d5773c32a2c946fd1a062b64a17aa388f30d/lib/network.sh#L45 and I guess the one you are showing is here, https://github.com/metal3-io/metal3-dev-env/blob/c197d5773c32a2c946fd1a062b64a17aa388f30d/lib/network.sh#L44. The reason it was not a problem because the one I am changing is only used in the configmap and as an env variable inside bmo pod. The one you pinpointed is for metal3-dev-env

fmuyassarov: OK. Thanks

PR #484, opened 2020-04-14
Makefile: update operator-sdk command

q384566678:
operator-sdk  `up local` is now  `run --local` since v0.15.0.

Signed-off-by: Zhou Hao <zhouhao@cn.fujitsu.com>

PR #487, opened 2020-04-22
Update operator-sdk

honza:
* Upgrade to `v0.17.0`
* Use `openapi-gen` directly because the original command is deprecated


honza:
> What's up with hack/boilerplate.go.txt? Is that meant to be empty?

Good point; I have set it to `/dev/null`.

> Is #484 a prerequisite for this?

No, they are independent and both necessary.

Issue #477, opened 2020-04-02
Need Clarification for Cluster Creation and Node Provisioning with Real Hardware


sumitjadhav1:
[1] We are currently testing the Metal3/BMO and Cluster Provisioning on Dell PowerEdge Servers. We have successfully completed till Node Introspection (BM Nodes in Ready state) step.
Now next step is BM Node provisioning and Cluster Creation. We are using two approaches as follows :

1. First Node Provisioning using provision_host.sh script and create cluster using cluster scripts :
-> Here BM Nodes state changes from Ready to Provisioned successfully (**Workaround** : We need to configure one boot device parameter in Ironic using CLI before starting Node Provisioning)
-> After this we start with Cluster Scripts available under ~/metal3-dev-env/scripts/v1alphaX
-> We're facing issue here that none of available BM Nodes is picked for Cluster initiation process (to be made Master node)
-> Our doubt : Is this the right approach for Cluster Creation and Provisioning

2. BM Nodes in Ready state and use Cluster scripts :
-> Here BM Nodes are in Ready state post successful introspection. We also apply Workaround as stated above in approach 1 for Ironic for all available Nodes.
-> Now we use Cluster scripts for cluster creation and it picks one of the available nodes for Node Provisioning (to change from Ready to Provisioned state)
-> This operation gets stuck at Ironic with error for which we have already applied the **Workaround** at the beginning.
-> Our doubt : Node Provisioning works in approach-1 whereas here it fails with error at Ironic even though we have Workaround in place.

[2] Another doubt regarding Baremetal Network (192.168.111.0/24), which is used for vBMC nodes created in Metal3 setup.
-> Do we need to change this IP/Range (Cluster API Endpoint is defined in this range) in lib/common.sh file to use iDRAC network available for Dell hardware. Observed below error in our logs :
#####
"failed to create remote cluster client: failed to create client for workload cluster metal3/test1: Get https://192.168.111.249:6443/api?timeout=30s: dial tcp 192.168.111.249:6443: connect: no route to host" 
#####

-> Need help on this front.

jan-est:
Cluster need to be provisioned before creating any machines. I am not sure does this fix your problem but workflow is: 

1. create cluster
2. create control plane
3. create machines

Not sure which version you are working on. But this is the way it is done in v1alpha3

juliakreger:
[...]

PR #493, opened 2020-04-29
Explicitly persistent boot mode setting with Redfish

juliakreger:
Later versions of Redfish corresponding to the later 2019 Schema
documents and Redfish 1.9 have a breaking behavior change from
earlier Redfish versions where it was possible to explicitly
assert a continuously overridden boot device with UEFI.

The changes to the standard functionally removes the ability
for any reasonable override logic to assist until vendors possibly
implement the Redfish 1.10 Boot alias field and logic.

This is further compounded by API and schema version indicators
appearing to be an unreliable indicator as the implementations
are exhibiting different schema and API version numbers from
earlier versions from before the standard change was introduced.

The only clear option at this time seems to be disable assertion
of persistent overrides as vendors adapt their BMC implementations
to the newer Redfish schema/api, at least until we're able to find
a cross vendor or even set of vendor specific happy paths that can
be leveraged and implemented to reliably

The underlying ironic driver_info field
``force_persistent_boot_device`` enables us to disable the
persistent future boot device setting.

This change also adds contextual information to the dev-setup
documentation in order to help provide a complete context.

rhbz# 1828885

### DIFF
--- a/docs/dev-setup.md
+++ b/docs/dev-setup.md
@@ -71,6 +71,31 @@ Note that this script may need customizations to some of the `podman run`
 commands, to include environment variables that configure the containers for
 your environment.
 
+## Use with Redfish BMCs and UEFI
+
+Newer firmware versions being released by hardware vendors, starting in
+late 2019, are beginning to introducing changes which breaks the ability
+for a Redfish API consumer to assert a persistent boot behavior overrides
+utilizing a singular vendor-independnet field, at least when UEFI is also
+in use.
+
+Some, but not all, vendors have UEFI specific setting fields which are
+available, however the standard allows for it to be internally referential
+information. The end result being if these newer fields are even present,
+we need an extensive amount of vendor and hardware specific context, or even
+host specific internal information to be able to effectively leverage the
+UEFI specific fields.
+
+As a result, only one-time boot overrides can realistically be sent until
+vendors implement the latest fields and behavior specified in the early
+2020 Redfish schema documentation which calls for a easily relatable alias
+field to be implemented.
+
+The end result being that hardware *must* be set to persistently boot to
+a Hard Disk or other persistent storage medium prior to being deployed upon,
+if the desired end state is such and persistent settings are not asseted
+before the first reboot of the node after intiial deployment.
+
 ## Using libvirt VMs with Ironic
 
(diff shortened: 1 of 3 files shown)


docs/dev-setup.md
```diff
@@ -71,6 +71,31 @@ Note that this script may need customizations to some of the `podman run`
 commands, to include environment variables that configure the containers for
 your environment.
 
+## Use with Redfish BMCs and UEFI
```
dhellmann, line 74: This doesn't feel like "developer setup" documentation. I don't really see another obviously good place to put it. Maybe it's time to start a user guide? We can do that in a follow-up PR, unless someone points out an existing file that's a better home.

juliakreger: I had the exact same thought, but proposed it as this to see what you guys thought.

juliakreger:
Dmitry, I do agree with you, but I think that independent of this change. I think the easiest thing is to let this merge into upstream master, let someone give that a spin and begin working to get this ported over and backported where applicable while we're working on the other path concurrently. Only time will tell how users/operators break, and at least getting it in the hands of downstream QA would give us an idea.

docs/dev-setup.md
dtantsur: Not sure. There is a lot of background technical information here, which is useful to keep. I do agree on a user guide, but I think the background information can stay here, and the user guide should only mention the behavior we have to stick to.

dtantsur:
PoC of what I mean: https://review.opendev.org/#/c/725239/

dtantsur:
@juliakreger could you maybe close this in favour of https://review.opendev.org/#/c/725239/ ?

dhellmann:
@juliakreger @dtantsur Should we close this since the change in Ironic landed?

sumitjadhav1:
This fix would be helpful for us to verify the behavior with latest iDRAC FW on Dell PowerEdge Servers. 
Here is the earlier reference to it : [BMO-Bug#477](https://github.com/metal3-io/baremetal-operator/issues/477#issuecomment-608374968)

Waiting the merge 👍

juliakreger:
Given fixed firmware and changes in Ironic, I see no reason to keep this PR open. If there is disagreement I guess we can re-open.

PR #491, opened 2020-04-29
make gofmt job strict

dhellmann:

PR #494, opened 2020-04-29
remove lint job from travis

dhellmann:
There's no sense in running the same job on both prow and travis.


dhellmann:
This appears to be a pre-req for #491 (for some reason that PR is still running the travis job).

dhellmann:
Maybe since this change doesn't include any code we can bypass the integration test job?

russellb:
Yeah, I'll merge manually ...

PR #496, opened 2020-04-29
fix kubebuilder validation for vlanid

dhellmann:
The validation rules need to be set where the type is defined, not
where it is used.

Addresses #434


russellb:
Should we add something to CI that checks this?  if not part of this PR, an issue to track follow-up?

russellb:
Nevermind, now I see that in #497 

/lgtm
/check-integration

PR #497, opened 2020-04-29
set up check to ensure operator-sdk generation is done

dhellmann:
Add a make target 'generate-check' to ensure that both the code and
CRDs have been regenerated for any change.

Add hack/Dockerfile.operator-sdk to build an image containing the
operator-sdk and other dependencies for the generator check.

Add hack/generate.sh to run the check in a container using the image
built by the new Dockerfile.

Addresses #434

### DIFF
--- a/hack/generate.sh
+++ b/hack/generate.sh
@@ -1,0 +1,39 @@
+#!/bin/sh
+
+# Ignore the rule that says we should always quote variables, because
+# in this script we *do* want globbing.
+# shellcheck disable=SC2086
+
+set -eux
+
+IS_CONTAINER=${IS_CONTAINER:-false}
+ARTIFACTS=${ARTIFACTS:-/tmp}
+CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-podman}"
+
+if [ "${IS_CONTAINER}" != "false" ]; then
+  eval "$(go env)"
+  cd "${GOPATH}"/src/github.com/metal3-io/baremetal-operator
+  export XDG_CACHE_HOME="/tmp/.cache"
+
+  operator-sdk version
+
+  INPUT_FILES="deploy/crds/*.yaml pkg/apis/metal3/v1alpha1/zz_generated.*.go"
+  cksum $INPUT_FILES > "$ARTIFACTS/lint.cksums.before"
+  export VERBOSE="--verbose"
+  make generate
+  cksum $INPUT_FILES > "$ARTIFACTS/lint.cksums.after"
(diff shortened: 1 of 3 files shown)


dhellmann:
I have manually built an image using the new docker file from this PR and pushed it to quay.io/metal3-io/operator-sdk. After this PR merges, I will update that quay repo to have a trigger to automatically rebuild the image when the docker file changes.

dhellmann:
And after that's done, we can update the project infra to run the new check against this repo.

russellb:
Thanks for adding this!  I just wonder if we can find an operator-sdk image that already exists.  I didn't see an obvious one in a quick search, but I'm asking around ...

russellb:
> Thanks for adding this! I just wonder if we can find an operator-sdk image that already exists. I didn't see an obvious one in a quick search, but I'm asking around ...

The feedback I got was a +1 to what you've already done :-)

dhellmann:
> I have manually built an image using the new docker file from this PR and pushed it to quay.io/metal3-io/operator-sdk. After this PR merges, I will update that quay repo to have a trigger to automatically rebuild the image when the docker file changes.

I added the automatic trigger.

hack/generate.sh
```diff
@@ -0,0 +1,39 @@
...
+  cd "${GOPATH}"/src/github.com/metal3-io/baremetal-operator
+  export XDG_CACHE_HOME="/tmp/.cache"
+
+  operator-sdk version
+
+  INPUT_FILES="deploy/crds/*.yaml pkg/apis/metal3/v1alpha1/zz_generated.*.go"
+  cksum $INPUT_FILES > "$ARTIFACTS/lint.cksums.before"
+  export VERBOSE="--verbose"
+  make generate
+  cksum $INPUT_FILES > "$ARTIFACTS/lint.cksums.after"
```
zaneb, line 24: One downside of only storing checksums is that if you *had* made local changes then `make generate-check` or `make lint` will make them go bye-bye, which could be a bit surprising. That was the reason for my hacky gymnastics in the Makefile previously.

dhellmann: What is the use case for editing the YAML by hand?

zaneb: Nothing valid, but if you didn't realise you weren't supposed to and only found out by using the linter then the info you wanted to copy-and-paste into the right place was just permanently deleted.
It also means `make lint` is not idempotent; running it a second time will pass. You might end up pushing a commit that actually fails if you don't notice the uncommitted changes (although at least there's no data loss in that case).

PR #498, opened 2020-04-30
WIP: change clock speed data type to fix crd generation

dhellmann:
kubebuilder does not support float types for fields. This patch
changes the ClockSpeed type from float64 to json.Number, which is
serialized as a string.

This is not a backwards-compatible API change.

Addresses #434


dhellmann:
See #510 instead

PR #504, opened 2020-05-01
fix crd generation

dhellmann:
Add a copy of operator-sdk v0.17.0 to the local tools directory and
modify it to support float fields. This gives us a version of the tool
that we can use in CI to verify future changes and add a new version
that drops the float field. At that point, we should remove this copy
of the tools and resuming using the published releases.

Addresses #434


maelk:
In the CI, the BMH are in "inspecting" state, while if running the `openstack baremetal node list` they are both manageable.

dhellmann:
> In the CI, the BMH are in "inspecting" state, while if running the `openstack baremetal node list` they are both manageable.

I can't imagine why the changes in this PR would cause that. Could that be related to the recent ironic image changes?

maelk:
we did not see this problem anywhere else until now

dhellmann:
OK, I'll look at it more closely then.

hack/shellcheck.sh
```diff
@@ -7,7 +7,10 @@ CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-podman}"
 
 if [ "${IS_CONTAINER}" != "false" ]; then
   TOP_DIR="${1:-.}"
-  find "${TOP_DIR}" -path ./vendor -prune -o -name '*.sh' -exec shellcheck -s bash {} \+
+  find "${TOP_DIR}" \
+       \( -path ./vendor -prune -o \
+       -path ./tools/operator-sdk -prune \) \
+       -o -name '*.sh' -exec shellcheck -s bash {} \+
```
zaneb, line 13: `find` is the worst unix command.

PR #506, opened 2020-05-05
Postpone inspector start when running out of cluster

maelk:
To give some time for Ironic to start properly and avoid
inspector failing to start due to ironic not being ready


zaneb:
@dtantsur this seems like a bug in Ironic; startup order should not matter, and in a non-pod deployment services can be expected to be restarted independently and they should always come up.

dtantsur:
We've fixed similar issues before, I'd be curious to know the details.

dtantsur:
Okay, I think https://review.opendev.org/#/c/725688/ is the fix (will finish it tomorrow).

PR #509, opened 2020-05-06
Revert "Postpone inspector start when running out of cluster"

maelk:
Reverts metal3-io/baremetal-operator#506 as it should not be needed once Ironic is fixed


maelk:
/close
Proper PR in https://github.com/metal3-io/baremetal-operator/pull/534

PR #512, opened 2020-05-11
WIP Create hosts powered off by default

maelk:


dhellmann:
I could use a little more background for why this change is desirable.

We ran into a bug last week because the default for `online` is false and it wasn't set `true` in one of the components we have generating host CRs. I was thinking of proposing changing `online` to `offline` and changing the default behavior as part of the v1beta1 update.

dhellmann:
See #513

maelk:
/close
This is for some tests with metal3-dev-env integration jobs, regarding an issue we were facing in the CI. It is not meant to be merged

PR #514, opened 2020-05-12
Get rid of extra resource write during delete

zaneb:
Now that 'deleting' is treated as its own provisioning status, returning
actionContinue is sufficient to ensure that the state gets written at
the conclusion of reconciliation. Doing an early write will cause an
error to be logged when we go to do the final write, since the object
generation will no longer match.


dhellmann:
It looks like tide got stuck?

/lgtm
/approve

dhellmann:
All of the tests passed, so I'll merge this by hand.

PR #516, opened 2020-05-13
Requeue after writing status from annotation

zaneb:
If we are copying the Status in from a baremetalhost.metal3.io/status
annotation, return and requeue immediately after writing it rather than
try to continue to do other things, which may result in a write
conflict.

Also, invert the output of the hostHasStatus() method, so that it no
longer returns the exact opposite of what it says on the tin.


maelk:
lgtm

PR #518, opened 2020-05-13
Fix IP address in ironic_ci.env

fmuyassarov:
Fix Ironic IP address.

### DIFF
--- a/deploy/ironic_ci.env
+++ b/deploy/ironic_ci.env
@@ -1,9 +1,11 @@
 HTTP_PORT=6180
+PROVISIONING_IP=172.22.0.2
+PROVISIONING_CIDR=24
 PROVISIONING_INTERFACE=provisioning
 DHCP_RANGE=172.22.0.10,172.22.0.100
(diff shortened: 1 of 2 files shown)


maelk:
internally ironic still uses 172.22.0.1 instead of .2. There might be an issue with ironic-common.sh in ironic and inspector images.

kashifest:
@maelk  the issue is not in  ironic-common.sh in ironic and inspector images. The CI is failing because although ironic is listening on .2, bmo thinks ironic is listening on.1 . This happens because bmo pod is initialized using a different configmap which is overrided in [metal3-dev-env](https://github.com/metal3-io/metal3-dev-env/blob/master/03_launch_mgmt_cluster.sh#L57) . The `${PROVISIONING_URL_HOST}` is .1

maelk:
@kashifest I don't think this is the problem. If it was, your BMH would not even go pass the registering state. Instead if you look into the logs of inspector, you can see that it fails to start properly and that's why the inspection fails. That issue should be solved with the ironic upgrade. Another problem that you can see is that even if it is given the .2 IP address , it still auto-detects .1 in ironic-common.sh.

PR #481, opened 2020-04-09
Restructure operator folder and create kustomizations for different scenarios 

furkatgofurov7:
These changes are made in context of making BMO as part of CAPM3 deployment. 
1. We remove `WATCH_NAMESPACE` so that BMO watches across all namespaces 
2. `operator` folder was renamed to `deployments` and divided into logical units
3. `kustomizations` folder was added containing kustomizations for different case scenarios

This PR is redesigned version of this [PR#479](https://github.com/metal3-io/baremetal-operator/pull/479)

PR #522, opened 2020-05-13
Arranging BMO & Ironic deployments with separate folders 

furkatgofurov7:
**What this PR does / why we need it:**
This PR breaks [metal3-io/baremetal-operator#481](https://github.com/metal3-io/baremetal-operator/pull/481) into two parts and this is a **first** part which introduces what is needed to be added in that PR. It structures `deploy` folder as simple as possible (only for BMO) and fits for operator-sdk while all ironic part is moved to its own folder.

Related to #540 
Follows the proposal in [metal3-docs/pull/91](https://github.com/metal3-io/metal3-docs/pull/91)
Documented in #541


dhellmann:
Ironic and the operator used to be in a single deployment so that the operator could use localhost to talk to ironic without any authentication. Why do we want them to be in separate deployments now?

furkatgofurov7:
The reason is that we want to deploy BMO on its own in CAPM3 and Ironic independently. The use case explaining what we are trying to achieve is a pivoting work and these changes will make it possible to make BMO as part of metal3 in clusterctl.

furkatgofurov7:
> this looks good. We should probably however, add a lot of documentation, and either add a top-level kustomization that will deploy BMO + ironic or add instructions with command examples on how to achieve this.

Yes sure, I will add instructions with commands in `docs` folder, if it is right place to do so?

maelk:
you should probably still put something in the README, to refer to the docs, or a simpler version of the docs, like the tldr of the docs

furkatgofurov7:
> this looks good. We should probably however, add a lot of documentation, and either add a top-level kustomization that will deploy BMO + ironic or add instructions with command examples on how to achieve this.

I have added instructions with command examples in this [PR#541](https://github.com/metal3-io/baremetal-operator/pull/541)

furkatgofurov7:
> you should probably still put something in the README, to refer to the docs, or a simpler version of the docs, like the tldr of the docs

I have modified the previous configuration file and referral link to configuration document inside it

furkatgofurov7:
@dhellmann if it looks good to you as well, it needs your approval to be merged.

deploy/default/kustomization.yaml
```diff
@@ -5,13 +5,8 @@ resources:
 - ../namespace
 - ../rbac
 - ../crds
-- ../operator/ironic
+- ../operator
```
maelk, line 8: we should add ironic here to not modify the current behaviour

furkatgofurov7: Ah true, added. Thanks!

PR #527, opened 2020-05-14
clean up Makefile

dhellmann:
Try to make the targets simpler and single-purpose. Also update the make target to use the same command as the gosec.sh script.

Addresses #505

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,7 +71,7 @@ unit-verbose: ## Run unit tests with verbose output
 	VERBOSE=-v make unit
 
 .PHONY: linters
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -148,6 +148,10 @@ docker-operator:
 docker-sdk:
 	docker build . -f hack/Dockerfile.operator-sdk
 
+.PHONY: docker-golint
+docker-sdk:
```
zaneb, line 152: wrong target name

dhellmann: paste-o. Fixed.

PR #531, opened 2020-05-18
Add missing delay in ValidateManagementAccess()

zaneb:
We discovered in #519 that it is possible for the Ironic node to remain in a `inspect wait` state even after the operator has received the inspection results and changed the state to Ready.
    
When calling ValidateManagementAccess() from the Ready state, we should wait some time before re-reconciling if the host is not in the `manageable` state. Previously, it was requeuing immediately with the effect that the operator would poll the Ironic state in essentially a tight loop.

Also fix up some redundant code that makes it confusing to read.


maelk:
@dhellmann have you been able to successfully trigger the tests somewhere and this is a temporary glitch or have you always failed ? Because I recall multiple times seeing the tests did not start when you triggered, so I'm trying to figure out if this is a bug

PR #532, opened 2020-05-19
Register host config data error metric

zaneb:
The prometheus metric metal3_host_config_data_error_total added in
004fc7a72f95ba7130c5ddf658de309c11e9bafa (#348) was not registered.


maelk:
The travis failure was due to quay.io. Can we retrigger it somehow ?

PR #533, opened 2020-05-20
Add Image disk format in api and ironic updates

maelk:
In order for ironic to use raw image streaming, it needs to be aware that the image is in raw format. This gives the user the possibility to specify that the image is raw. If left empty, the previous behaviour (copy in RAM) is maintained, for qcow2 images etc.


deploy/crds/metal3.io_baremetalhosts_crd.yaml
```diff
@@ -157,6 +157,18 @@ spec:
...
+                format:
+                  description: DiskFormat contains the format of the image (raw, qcow2,
+                    ...) Needs to be set to raw for raw images streaming
+                  enum:
+                  - raw
+                  - qcow2
+                  - qed
+                  - vdi
+                  - vmdk
+                  - vpc
```
zaneb, line 169: Are VPC and VHD the same thing? Let's drop VPC from the list if so.

deploy/crds/metal3.io_baremetalhosts_crd.yaml
```diff
@@ -157,6 +157,18 @@ spec:
                   - sha256
                   - sha512
                   type: string
+                format:
+                  description: DiskFormat contains the format of the image (raw, qcow2,
+                    ...) Needs to be set to raw for raw images streaming
+                  enum:
+                  - raw
+                  - qcow2
+                  - qed
```
zaneb, line 166: "[It is not recommended to use QED for any new images.](https://wiki.qemu.org/Features/QED)"
Let's not make this an option.

maelk:
This requires a change from ironic side to support raw image streaming : https://review.opendev.org/#/c/730317/
I will update the PR according to your comments. I was trying to be as exhaustive as possible, but we can probably limit the list to a minimum set and add new formats when needed.

maelk:
/hold cancel
The ironic patch was merged

PR #536, opened 2020-05-22
Add tryChangeNodeProvisionState() method in ironic provisioner

zaneb:
Sometimes a caller actually needs to know whether we succeeded in
changing the provisioning state or if we're just retrying because of a
conflict.

Mostly it doesn't though, so implement it as a separate method to avoid
adding complexity everywhere.

This is needed by #292.

### DIFF
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -422,7 +422,7 @@ func (p *ironicProvisioner) ValidateManagementAccess(credentialsChanged bool) (r
 	}
 }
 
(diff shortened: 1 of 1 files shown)


pkg/provisioner/ironic/ironic.go
```diff
@@ -432,16 +432,23 @@ func (p *ironicProvisioner) changeNodeProvisionState(ironicNode *nodes.Node, opt
 	changeResult := nodes.ChangeProvisionState(p.client, ironicNode.UUID, opts)
 	switch changeResult.Err.(type) {
 	case nil:
+		success = true
 	case gophercloud.ErrDefault409:
```
dhellmann, line 436: If the reason for failure (or at least "not success") is that we need to retry, why wouldn't we need to check that on every call?

demonCoder95:
What's the holdup on this? Are we waiting on reviews from other cores?

PR #538, opened 2020-05-24
Unify the image reference in operator deployments

maelk:
Kustomizing the image in metal3-dev-env requires the image to
be specific. This unifies the BMO image across different
deployments so that the image kustomization works in all cases


maelk:
not having a tag means latest. I need to verify that latest tracks master though

maelk:
yes, in quay.io, latest = master so specifying master, latest or no tag is equivalent

fmuyassarov:
> not having a tag means latest. I need to verify that latest tracks master though

My point was that, was there specific need to use master instead of latest. But apparently they are the same.

PR #539, opened 2020-05-27
fix typos in state diagram

dhellmann:
Rename 2 states that were "deleted" instead of "deleting".


zaneb:
This fix is also included in the current proposal for #292, but it doesn't hurt to go ahead and merge it.
/lgtm

PR #540, opened 2020-05-28
⚠️ Remove old BMO & Ironic deployments 

furkatgofurov7:
What this PR does / why we need it:
This PR is a **second** part of breaking #481  into two parts, which removes unnecessary folders/files.

First part of breaking #481 which adds necessary folders/files:  #522
Follows the proposal in: [metal3-docs/pull/91](https://github.com/metal3-io/metal3-docs/pull/91)
Documented in: #541


deploy/default/kustomization.yaml
```diff
@@ -1,17 +0,0 @@
-apiVersion: kustomize.config.k8s.io/v1beta1
```
maelk, line 1: actually, this one we probably should keep, as an easy way to deploy BMO standalone, you should change the ironic here for BMO standalone

furkatgofurov7: I think I should have it and also with same content as I have it in [PR#522](https://github.com/metal3-io/baremetal-operator/blob/d881359c2920f1b992cd06667fa990f0177e2030/deploy/default/kustomization.yaml) where I am modifying this file. Because that one will be merged first, and when we will merge this it will actually delete it, right?

maelk: Well you would most probably get a conflict, so let's focus on the other one first

maelk:
this needs to wait for the Metal3-dev-env PR to be in

PR #542, opened 2020-05-28
Update doc link and operator-sdk command

demonCoder95:
Updated the doc to point to the correct installation page for operator-sdk, since the SDK documentation has significantly changed to the website version now.

Also, updated the command in `Deploy Operator Locally` section to reflect newer operator-sdk version v0.17.1

### DIFF
--- a/docs/dev-setup.md
+++ b/docs/dev-setup.md
@@ -2,9 +2,8 @@
 
 ## Install the operator-sdk
 
-Follow the instructions in the Quick Start section of
-<https://github.com/operator-framework/operator-sdk> to check out and
-install the operator-sdk tools.
+Follow the instructions on the [Operator SDK website](https://sdk.operatorframework.io/docs/install-operator-sdk/)
+to check out and install the operator-sdk tools.
 
 ## Install openapi-gen
 
@@ -48,7 +47,7 @@ go get k8s.io/kube-openapi/cmd/openapi-gen
     export DEPLOY_RAMDISK_URL=http://172.22.0.1/images/ironic-python-agent.initramfs
     export IRONIC_ENDPOINT=http://localhost:6385/v1/
     export IRONIC_INSPECTOR_ENDPOINT=http://localhost:5050/v1
-    operator-sdk up local --namespace=metal3
+    operator-sdk run --local --watch-namespace=metal3
     ```
 
 1. Create the CR


demonCoder95:
I think we're good to merge this one.

demonCoder95:
@dhellmann Why is the PR still stuck? Shouldn't this be merged?

dhellmann:
> @dhellmann Why is the PR still stuck? Shouldn't this be merged?

It's waiting for the test-integration job, which has to be triggered manually.

/test-integration

dhellmann:
@maelk is there some way to tell if the test-integration trigger worked for this PR? Or to bypass that test, since this is just a doc change?

honza:
Can we clean up the commits? It looks like we have two commits with duplicate messages, and a merge commit.

demonCoder95:
@honza Can we squash the commits upon merge?

dhellmann:
> @honza Can we squash the commits upon merge?

Does Prow do that when it merges?

maelk:
I think it does if we put a specific label, the `tide/merge-method-squash` . For the ci testing I don't know of a way to override it. We would need to disable the branch protection in Github, and the config in Prox, that also requires the job to pass. The easiest way might be to just run the tests, or merge manually. @dhellmann Jenkins is actually a bit picky to start the tests. you need the `/test-integration` to be in its own comment. I can look into loosening the regex to match wider cases, like `.*/test-integration .*`

dhellmann:
Thanks @maelk 

@demonCoder95 I think it would be safest if you just took a few minutes to clean up the commits by hand. That way we don't have to rely on a behavior of Prow we're not sure about. If you do that, I'll shepherd the patch through the rest of the way and ensure we get it landed quickly.

maelk:
@dhellmann I would probably go for `^((.|\n)*\n)?\/test-integration(( |\n)(.|\n)*)?$` does that seem to catch all cases for you ?

dhellmann:
> @dhellmann I would probably go for `^((.|\n)*\n)?\/test-integration(( |\n)(.|\n)*)?$` does that seem to catch all cases for you ?

I would be OK with requiring the command on a line by itself. I think that's what that expression is trying to do? This might be a little simpler `'(^|.*\n)\/test-integration( *\n|$)'`

maelk:
I have simplified it so that any "/test-integration" surrounded by white space will trigger the job. It makes it easier. I'll create the patchset for it.

Issue #41, opened 2019-02-28
Integrate Ironic discovery workflow


russellb:
Ironic is capable of doing discovery of new hosts.  This should be an optional capability of the baremetal-operator.  When we discover a new host, it should result in automatically creating a corresponding BareMetalHost object.

dhellmann:
I have an early design for this in https://docs.google.com/document/d/1hlMSdim53CQtqlYOnx1ptq3qKHKhP9LhDrRxzCutGoM/edit and will move that to metal3-docs after we have had time to collaborate more there.

dhellmann:
> I have an early design for this in https://docs.google.com/document/d/1hlMSdim53CQtqlYOnx1ptq3qKHKhP9LhDrRxzCutGoM/edit and will move that to metal3-docs after we have had time to collaborate more there.

That design is now available in https://github.com/metal3-io/metal3-docs/pull/98

PR #545, opened 2020-06-02
WIP: implement host discovery

dhellmann:
Addresses #41 

See https://github.com/metal3-io/metal3-docs/pull/98 as well


dhellmann:
The design is going in a different direction, so we aren't going to want this.

PR #546, opened 2020-06-04
remove infinite reconcile loop when fetching status annotation

dhellmann:
The problem won't be caused by the operator's version of the annotation, but we're using the annotation to load just the hardware details portion of the status and that is causing reconciliation to continuously copy the status from the annotation to the host and never let the host move past that point.


maelk:
looks good to me!

PR #547, opened 2020-06-04
Update docs with new Available BMH status

andybraren:
I'm more of a designer than a developer, but [#2](https://github.com/metal3-io/cluster-api-provider-metal3/pull/2) seems to have added the `Available` status as described in #315 to replace (or provide an alternative to? 🤷‍♂️) the `Ready` status for BMHs.

This PR updates the documentation to refer to `Available` instead of `Ready` if it makes sense to do so.


maelk:
Thank you for the PR. This was part of https://github.com/metal3-io/baremetal-operator/pull/340 but that PR is still open. Maybe we should rather try to complete #340 considering that "available" is not in use yet.

andybraren:
Ah I missed that PR, I'll close this one then, thanks @maelk!

PR #550, opened 2020-06-09
New BMH unhealthy annotation

jan-est:
CAPM3 node remediation needs the simple way of coordinate the provisioning and remediation of the unhealthy BMH. We want the user to be able to mark an unhealthy BMH with an ```unhealthy``` annotation. That annotation is used to control provisioning and remediation at the CAPM3 level.

For example, use of current pause annotation prevents the deletion of the machine objects and blocks the remediation of the unhealthy nodes.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -19,6 +19,15 @@ sure that you remove the annotation  **only if the value of the annotation is
 not `metal3.io/capm3`, but another value that you have provided**. Removing the
 annotation will enable the reconciliation again.
 
+### Unhealthy annotation
+
+It is possible to mark BareMetalHost object as unhealthy by adding an
+annotation `baremetalhost.metal3.io/unhealthy`. This annotation does not
+stop the reconciliation of the BMH. This annotation is used in the upper layers
+for coordination. For example **Metal³** provider should not provision BMH
+set as unhealthy. Removing the annotation will enable the normal operations
+on the provider layers.
+
 ### BareMetalHost spec
 
 The *BareMetalHost's* *spec* defines the desire state of the host. It contains
--- a/pkg/apis/metal3/v1alpha1/baremetalhost_types.go
+++ b/pkg/apis/metal3/v1alpha1/baremetalhost_types.go
@@ -25,6 +25,9 @@ const (
 
 	// StatusAnnotation is the annotation that holds the Status of BMH
 	StatusAnnotation = "baremetalhost.metal3.io/status"
+
+	// UnhealthyAnnotation is the annotation that sets unhealthy status of BMH
+	UnhealthyAnnotation = "baremetalhost.metal3.io/unhealthy"
 )
 
 // OperationalStatus represents the state of the host


dhellmann:
Will the BMO do anything special with a host that has this annotation? Or is it used to manage a CAPM3 behavior?

jan-est:
> Will the BMO do anything special with a host that has this annotation? Or is it used to manage a CAPM3 behavior?

BMO does not do anything. This annotation is intended entirely for CAPM3 use.

dhellmann:
> > Will the BMO do anything special with a host that has this annotation? Or is it used to manage a CAPM3 behavior?
> 
> BMO does not do anything. This annotation is intended entirely for CAPM3 use.

I wonder if it should really be defined there instead, then? Unless multiple providers might reuse the annotation? Or the BMO might look for it in the future?

jan-est:
> > > Will the BMO do anything special with a host that has this annotation? Or is it used to manage a CAPM3 behavior?
> > 
> > 
> > BMO does not do anything. This annotation is intended entirely for CAPM3 use.
> 
> I wonder if it should really be defined there instead, then? Unless multiple providers might reuse the annotation? Or the BMO might look for it in the future?

I think annotation would make sense here in BMO and other providers might reuse the annotation. 

At this point we would use this annotation in CAPM3 to prevent annotated BMH to be selected as host for the metal3machine. Pause annotation is bit problematic for our use cases because it pause the whole reconciliation of a BareMetalHost. Use of pause prevent us to delete the machine object for example. 

So we need some mechanism to say "do not use this BMH" without stop the reconciliation.

dhellmann:
> > > > Will the BMO do anything special with a host that has this annotation? Or is it used to manage a CAPM3 behavior?
> > > 
> > > 
> > > BMO does not do anything. This annotation is intended entirely for CAPM3 use.
> > 
> > 
> > I wonder if it should really be defined there instead, then? Unless multiple providers might reuse the annotation? Or the BMO might look for it in the future?
> 
> I think annotation would make sense here in BMO and other providers might reuse the annotation.
> 
> At this point we would use this annotation in CAPM3 to prevent annotated BMH to be selected as host for the metal3machine. Pause annotation is bit problematic for our use cases because it pause the whole reconciliation of a BareMetalHost. Use of pause prevent us to delete the machine object for example.
> 
> So we need some mechanism to say "do not use this BMH" without stopping the reconciliation.

Sure, that makes sense.

kashifest:
lgtm , waiting for @dhellmann  to agree :smile:

kashifest:
ow he did, good, /lgtm

dhellmann:
> lgtm , waiting for @dhellmann to agree 😄

This still feels like an API fo the CAPM3, rather than the BMO. Only the CAPM3 will pay attention to the presence of the annotation.

I understand that multiple controllers could reuse the annotation, but will they? @beekhof what do we use in cluster-api-provider-baremetal to indicate that a host should not be used? Is it just based on the label selectors and state?

kashifest:
> > lgtm , waiting for @dhellmann to agree smile
> 
> This still feels like an API fo the CAPM3, rather than the BMO. Only the CAPM3 will pay attention to the presence of the annotation.
> 
> I understand that multiple controllers could reuse the annotation, but will they? @beekhof what do we use in cluster-api-provider-baremetal to indicate that a host should not be used? Is it just based on the label selectors and state?

ow, was I too fast !!!

jan-est:
> > lgtm , waiting for @dhellmann to agree smile
> 
> This still feels like an API fo the CAPM3, rather than the BMO. Only the CAPM3 will pay attention to the presence of the annotation.
> 
> I understand that multiple controllers could reuse the annotation, but will they? @beekhof what do we use in cluster-api-provider-baremetal to indicate that a host should not be used? Is it just based on the label selectors and state?

If I understood right the state and label selectors are used, but also if the pause annotation is in place host will not be used.

zaneb:
I agree with @dhellmann here; it seems like the purpose of the annotation is for CAPM3 to remember not to use the host, so it's part of the CAPM3 API not the BMH API, and should be namespaced accordingly.

kashifest:
> > > > > Will the BMO do anything special with a host that has this annotation? Or is it used to manage a CAPM3 behavior?
> > > > 
> > > > 
> > > > BMO does not do anything. This annotation is intended entirely for CAPM3 use.
> > > 
> > > 
> > > I wonder if it should really be defined there instead, then? Unless multiple providers might reuse the annotation? Or the BMO might look for it in the future?
> > 
> > 
> > I think annotation would make sense here in BMO and other providers might reuse the annotation.
> > At this point we would use this annotation in CAPM3 to prevent annotated BMH to be selected as host for the metal3machine. Pause annotation is bit problematic for our use cases because it pause the whole reconciliation of a BareMetalHost. Use of pause prevent us to delete the machine object for example.
> > So we need some mechanism to say "do not use this BMH" without stopping the reconciliation.
> 
> Sure, that makes sense.

Seeing this comment I thought an agreement was reached and IMO I don't have any issues of keeping it here or CAPM3, but since then I can see there is increasing consensus of keeping it in CAPM3, so we can revert the commit if deemed so and discuss more on this here.

dhellmann:
> > > > > > Will the BMO do anything special with a host that has this annotation? Or is it used to manage a CAPM3 behavior?
> > > > > 
> > > > > 
> > > > > BMO does not do anything. This annotation is intended entirely for CAPM3 use.
> > > > 
> > > > 
> > > > I wonder if it should really be defined there instead, then? Unless multiple providers might reuse the annotation? Or the BMO might look for it in the future?
> > > 
> > > 
> > > I think annotation would make sense here in BMO and other providers might reuse the annotation.
> > > At this point we would use this annotation in CAPM3 to prevent annotated BMH to be selected as host for the metal3machine. Pause annotation is bit problematic for our use cases because it pause the whole reconciliation of a BareMetalHost. Use of pause prevent us to delete the machine object for example.
> > > So we need some mechanism to say "do not use this BMH" without stopping the reconciliation.
> > 
> > 
> > Sure, that makes sense.
> 
> Seeing this comment I thought an agreement was reached and IMO I don't have any issues of keeping it here or CAPM3, but since then I can see there is increasing consensus of keeping it in CAPM3, so we can revert the commit if deemed so and discuss more on this here.

Yeah, I'm not worried too much. I think this is a good example of an API change that would benefit from a design discussion in metal3-docs. It's a small change, which makes that extra work but that makes it easy to write and easy to approve quickly, too. I'm not asking that we do that retroactively, just pointing it out for next time.

I would prefer the definition of the annotation go in CAPM3 near where it will be used, but I will accept the group consensus if others feel that it's better here.

I would like to hear from @beekhof before we settle things, since I know his team has been working on lots of issues related to remediation.

maelk:
I would think that this kind of annotation belong rather in BMO. They are specific to the BaremetalHost and should be understood by the upper layers that are using it. For example, we have a command line client, independent of the CAPM3 project, that we use to create all the CRs (e.g. Airshipctl), where a command could be to set a BMH node as unhealthy, meaning that it would add this annotation, preventing any other layers, such as CAPM3 or machine API or anything built on top of BMO, to use that specific host. IMO, this should not be in CAPM3, because this type of behaviour should be in the contract around the BMH API, even though for now BMO does not actually act on this annotation. But one could very well imagine that at some point BMO would do something about it, for example report as a metric the number of hosts with unhealthy annotation. 

If we put this in CAPM3, it makes it a CAPM3 API extending to BMO and mixes the two, while it could be defined as part of the BMO API for all higher layer to consume it the same way.

zaneb:
> I would like to hear from @beekhof before we settle things, since I know his team has been working on lots of issues related to remediation.

I think I can answer that. Remediation is concerned with failed Machines, not Hosts. For Hosts not associated with a Machine, it doesn't care about whether a new Machine picks up that Host or not.

> preventing any other layers, such as CAPM3 or machine API or anything built on top of BMO, to use that specific host.

Yeah, that is the key question. Is this a message from the BMO telling everything to stay away? Or is it a message to just CAPM3?
IMHO it's the latter. It's a message from Airshipctl (in this case) to CAPM3. BMO is not involved at either end.

(Incidentally, I can see this problem being solved in a more generic way by the proposed BareMetalAllocation resource type in a future API version.)

> If we put this in CAPM3, it makes it a CAPM3 API extending to BMO and mixes the two

I don't really agree that it's mixing the two. Generally annotations are namespaced to the thing that is supposed to act on them, which in this case is CAPM3, regardless of what resource type they are applied to.

jan-est:
@zaneb @dhellmann  thanks for the discussion in yesterdays' community meeting. We will define this annotation in capm3 and I have opened a PR to reverse the changes made in this one.

PR #553, opened 2020-06-11
Remove Unhealthy annotation changes

jan-est:
This PR will remove API changes made in [550](https://github.com/metal3-io/baremetal-operator/pull/550).

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -19,15 +19,6 @@ sure that you remove the annotation  **only if the value of the annotation is
 not `metal3.io/capm3`, but another value that you have provided**. Removing the
 annotation will enable the reconciliation again.
 
-### Unhealthy annotation
-
-It is possible to mark BareMetalHost object as unhealthy by adding an
-annotation `baremetalhost.metal3.io/unhealthy`. This annotation does not
-stop the reconciliation of the BMH. This annotation is used in the upper layers
-for coordination. For example **Metal³** provider should not provision BMH
-set as unhealthy. Removing the annotation will enable the normal operations
-on the provider layers.
-
 ### BareMetalHost spec
 
 The *BareMetalHost's* *spec* defines the desire state of the host. It contains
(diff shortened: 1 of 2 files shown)


maelk:
/hold until the discussion comes to an agreement
I'm ok to merge this.

zaneb:
Thanks for pushing this up.

Thinking about in the context of the proposed future major API change... a possible implementation in that case would be that you delete the HostAllocation when you want to take a host out of circulation. The effect of that would be deprovision the Host if it is provisioned, and prevent further provisioning. Does that fit the use case you have for this annotation? Or would you e.g. want an already-provisioned host to remain provisioned so that an admin can log in and poke around or something?

If it makes sense to switch to that implementation in a future API (i.e. we're fine with deprovisioning an unhealthy host) then it would make sense to keep this annotation defined in the bmh controller, and go ahead and implement the deprovisioning part.

If not then I think it makes sense to go ahead with this revert and make it a CAPM3 thing. (Honestly, you probably don't even need an annotation - you could just use a label and have CAPM3 filter out that label... no need to even change any code.)

jan-est:
> Thinking about in the context of the proposed future major API change... a possible implementation in that case would be that you delete the HostAllocation when you want to take a host out of circulation. The effect of that would be deprovision the Host if it is provisioned, and prevent further provisioning. Does that fit the use case you have for this annotation? 

Yes, that might work?

> Or would you e.g. want an already-provisioned host to remain provisioned so that an admin can log in and poke around or something?

Actually I don't have the straight answer to this right now. But I would be good to keep discussing about the future API change. I've been working with CAPI healthcheck and idea of CAPM3 remediation controller and I keep thinking what will happen when automated remediation fails after reboot of the node. I cases of hardware failure and node needs to take out of circulation for some time, what would be the best way of doing this. Annotation might be good enough and i'm not sure does admin benefit log?

> 
> If it makes sense to switch to that implementation in a future API (i.e. we're fine with deprovisioning an unhealthy host) then it would make sense to keep this annotation defined in the bmh controller, and go ahead and implement the deprovisioning part.

> If not then I think it makes sense to go ahead with this revert and make it a CAPM3 thing. (Honestly, you probably don't even need an annotation - you could just use a label and have CAPM3 filter out that label... no need to even change any code.)

Now our plan is to annotate BMH with capm3 annotation and remove corresponding machine object that will start the deprovisioning of the BMH.

jan-est:
Could we merge this? We implemented unhealthy annotation in CAPM3.

dhellmann:
I can never remember this command

/hold cancel

PR #557, opened 2020-06-16
Upgrade module to Golang 1.14, and tidy modules

honza:


dhellmann:
This change looks fine, but a little more detail in the commit message about why the change is needed would be good.

honza:
> This change looks fine, but a little more detail in the commit message about why the change is needed would be good.

OK, I can certainly add some. I was hesitant because it was driven by a downstream requirement.

dhellmann:
> > This change looks fine, but a little more detail in the commit message about why the change is needed would be good.
> 
> OK, I can certainly add some. I was hesitant because it was driven by a downstream requirement.

That's OK, just explaining the benefit in general terms would cover it.

honza:
> That's OK, just explaining the benefit in general terms would cover it.

Done :boom:

PR #559, opened 2020-06-22
Update Ironic Keepalived container documentation

kashifest:
The kustomization workflow  has changed significantly and the different deployment scenarios have also been updated. This PR removes the obsolete info from  Ironic Keepalived container documentation and also updates the documentation with the way it should be deployed now.

### DIFF
--- a/docs/ironic-endpoint-keepalived-configuration.md
+++ b/docs/ironic-endpoint-keepalived-configuration.md
@@ -7,71 +7,11 @@ cluster is taken down, target cluster controlplane can re-claim the ironic
 endpoint IP through keepalived. The end goal is to make ironic endpoint
 reachable in the target cluster.
 
-To this end, we have restructred the `baremetal-operator/deploy` folder
-structure to better reflect the kustomization workflow as we now have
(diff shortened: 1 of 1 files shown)


docs/ironic-endpoint-keepalived-configuration.md
```diff
@@ -95,20 +35,20 @@ tree resources/keepalived-docker/
 ```
 
 It is assumed that the docker image is uploaded in some registry and the image
-URL is used with `ironic-keepalived-config/image_patch.yaml` to replace the
-default image URL with the correct URL through kustomization.
+URL is used with `ironic-deployment/keepalived/keepalived_patch.yaml` to replace
+the default image URL with the correct URL through kustomization.
 
 **Important Note**
```
furkatgofurov7, line 41: I forgot to add, **Important Note** part is already included in [deploying.md](https://github.com/metal3-io/baremetal-operator/blob/master/docs/deploying.md), do we need to keep this here as well?

kashifest: I don't see any harm. I would rather keep it here specially since the variables are expected for this container.

PR #560, opened 2020-06-22
Fix: Delete ironic-endpoint-keepalived container also while deleting other ironic containers

kashifest:
Since `tools/run_local_ironic.sh `  installs  `ironic-endpoint-keepalived` , it should also delete this container while it deleted the other ironic containers, otherwise while doing iterations on this file, docker finds `ironic-endpoint-keepalived` already installed and throws an error.


maelk:
/lgtm
this is a quick fix and can go in

PR #562, opened 2020-06-23
Make BMCDetails optional

beekhof:
Hosts with StateExternallyProvisioned may not have BMC details

Signed-off-by: Andrew Beekhof <andrew@beekhof.net>

### DIFF
--- a/pkg/apis/metal3/v1alpha1/baremetalhost_types.go
+++ b/pkg/apis/metal3/v1alpha1/baremetalhost_types.go
@@ -204,6 +204,7 @@ type BareMetalHostSpec struct {
 	Taints []corev1.Taint `json:"taints,omitempty"`
 
 	// How do we connect to the BMC?
+	// +optional
 	BMC BMCDetails `json:"bmc,omitempty"`
 
 	// What is the name of the hardware profile for this host? It


maelk:
/ok-to-test
I'm surprised that this does not result in CRD changes. Either it is unneeded (since we have omitempty) or the CRDs were not re-generated ?

beekhof:
> /ok-to-test
> 
> I'm surprised that this does not result in CRD changes. Either it is unneeded (since we have omitempty) or the CRDs were not re-generated ?

I was tweaking from the web interface. I’ll need to regenerate when I’m back at my desk

zaneb:
I tried doing this and regenerating the CRDs, and it has no effect. The only Spec field that is required in the CRD is `Online`. So I don't think this change does anything.

maelk:
I think the omitempty has the same effect in practice as the optional flag

zaneb:
Yeah, I'm pretty sure omitempty makes it optional.
We could make the fields within the BMC optional, but in practice I think you do actually want all or none of them.

I think the issue @beekhof was trying to fix here is actually openshift/enhancements#388, so I'm going to close this.

/close

PR #563, opened 2020-06-25
Implement an Unmanaged (discovered) state

zaneb:
If no BMC credentials are provided, move the host to a new Unmanaged state and change the operational status to discovered, rather than remaining without a provisioning state and in an operational error.

This has always been envisaged in the state machine diagram (with the name Discovered rather than Unmanaged), but was not previously implemented.


pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -247,14 +247,21 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (result re
...
-		} else {
-			return r.credentialsErrorResult(err, request, host)
+	var bmcCreds *bmc.Credentials
+	var bmcCredsSecret *corev1.Secret
+	switch host.Status.Provisioning.State {
+	case metal3v1alpha1.StateNone, metal3v1alpha1.StateUnmanaged:
+		bmcCreds = &bmc.Credentials{}
+	default:
+		bmcCreds, bmcCredsSecret, err = r.buildAndValidateBMCCredentials(request, host)
+		if err != nil || bmcCreds == nil {
```
ohadlevy, line 257: could we have a default username without a password? does it make sense to check if there are no credentials or if user or password are empty?

zaneb: That's what `buildAndValidateBMCCredentials()` does.

docs/baremetalhost-states.md
```diff
@@ -11,10 +11,10 @@ Newly created hosts move immediately to Discovered or Registering. No
 host stays in the Created state while the operator is working
 properly.
 
-## Discovered
+## Unmanaged
 
-A Discovered host is missing either the BMC address or credentials
-secret name, and does not have enough information to access the BMC
+An Unmanaged host is missing both the BMC address and credentials
```
dhellmann, line 16: Does it need to be both? Could it be either?

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -247,14 +247,21 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (result re
...
-			// If we are in the process of deletion, try with empty credentials
-			bmcCreds = &bmc.Credentials{}
-			bmcCredsSecret = &corev1.Secret{}
-		} else {
-			return r.credentialsErrorResult(err, request, host)
+	var bmcCreds *bmc.Credentials
+	var bmcCredsSecret *corev1.Secret
+	switch host.Status.Provisioning.State {
+	case metal3v1alpha1.StateNone, metal3v1alpha1.StateUnmanaged:
+		bmcCreds = &bmc.Credentials{}
```
dhellmann, line 254: Would it make sense to report some status here to indicate what missing data is preventing the host from moving to the next state?

docs/baremetalhost-states.md
zaneb: Working assumption:
* If you didn't provide either, you meant to create an unmanaged host
* If you provided one but not the other, linked to a missing credential secret, or linked to an incorrectly formatted secret, you screwed up and should see big error messages in the UI.

pkg/controller/baremetalhost/baremetalhost_controller.go
zaneb: That's a good thought, maybe we should record an event when we go into the unmanaged state?

pkg/controller/baremetalhost/host_state_machine.go
```diff
@@ -152,13 +153,24 @@ func (hsm *hostStateMachine) shouldInitiateRegister(info *reconcileInfo) bool {
...
+	}
 	return actionComplete{}
 }
 
+func (hsm *hostStateMachine) handleUnmanaged(info *reconcileInfo) actionResult {
+	if hsm.Host.HasBMCDetails() {
+		hsm.NextState = metal3v1alpha1.StateRegistering
+		return actionComplete{}
+	}
+	return actionContinue{}
```
zaneb, line 171: This is wrong, we don't actually need to reschedule at all here (we won't leave this state unless the CR is modified) - or at the very least not until after a long delay. This will spin in tight loop which is very bad.

docs/baremetalhost-states.md
zaneb: Come to think of it, it's academic anyway because both are required fields within the BMC struct. So you can only specify both or neither.

pkg/apis/metal3/v1alpha1/baremetalhost_types.go
```diff
@@ -637,6 +641,11 @@ func (host *BareMetalHost) getLabel(name string) string {
 	return host.Labels[name]
 }
 
+// HasBMCDetails returns true if the BMC details are set
+func (host *BareMetalHost) HasBMCDetails() bool {
+	return host.Spec.BMC.Address != "" || host.Spec.BMC.CredentialsName != ""
```
n1r1, line 646: looks like this assume that `host.spec.BMC` exists. is this a valid assumption?

n1r1:
In some way, this is a breaking change, and [api.md](https://github.com/metal3-io/baremetal-operator/blob/master/docs/api.md) should be updated to reflect that. 
Few things that caught my eyes there - the `online` flag, which will trigger power change only if there are BMC credentials (or if not in `unmanaged` state); 
The [triggering-provisioning](https://github.com/metal3-io/baremetal-operator/blob/master/docs/api.md#triggering-provisioning) section should relate to the new state;
and probably some other fields like `poweredOn`, etc.

We also need to consider the consequences on other components that already have some assumptions on the existing API

beekhof:
Have we considered a webhook at all?
That seems like a useful addition to what is being proposed here.

We could use one to, for example, look to see if the BMH is unmanaged and reject any changes to `spec.Online` or prevent anything from setting the reboot annotation.  That way the code can assume that some of these problematic states can't occur, even from the `oc` command line.

dhellmann:
> In some way, this is a breaking change, and [api.md](https://github.com/metal3-io/baremetal-operator/blob/master/docs/api.md) should be updated to reflect that.
> Few things that caught my eyes there - the `online` flag, which will trigger power change only if there are BMC credentials (or if not in `unmanaged` state);
> The [triggering-provisioning](https://github.com/metal3-io/baremetal-operator/blob/master/docs/api.md#triggering-provisioning) section should relate to the new state;
> and probably some other fields like `poweredOn`, etc.

I will propose a follow-up to add some comments about requiring BMC credentials for those features.

> We also need to consider the consequences on other components that already have some assumptions on the existing API

@kirankt this is the comment I mentioned this morning.

@n1r1 are you thinking of something specific? The cluster-api-provider-* clients?

dhellmann:
I have verified locally that this PR works.

Another approach we could take is to allow hosts to be in the externally provisioned state without any BMC credentials. We wouldn't need any new states for that, but we would rearrange the existing states and hosts would stay in the "None" state if they weren't externally provisioned and didn't have credentials. Or maybe they would go Created->Registering->Error?

If we did that, we would have to add a status field to reflect whether the host could be managed or not, based on whether the BMC credentials are present and valid.

I'm not sure that approach gives us a better user experience, over adding the new steady state early. And it would complicate the rest of the controller code, because we would have to check any time we wanted to do something that required power control to see if we could.

We should also consider how this applies to [the discovery proposal](https://github.com/metal3-io/metal3-docs/pull/98). As it stands, this change doesn't allow a host to move from Unmanaged to Provisioned without adding the BMC details, but that's a goal of that other proposal.

dhellmann:
Zane's PR needed to be rebased and I don't have access to his GitHub repo to push there, so I opened #569 instead.

PR #565, opened 2020-06-25
Fix SaveHostStatus infinite loop on update conflicts

maelk:
In saveHostStatus, we get the BMH right after updating the 
annotation. In some cases the cache is not updated yet, causing
an error in the status update. This could lead to a loop in   
some cases where the cache was slow to update as the annotations 
were always changed since operationHistory was changing every 
time (it uses the current time as end time for an operation).
To solve this :

- Discard operationHistory when comparing status annotation to  
  current status to prevent unneeded updates of the annotation 
  leading to additional conflicts
- Do not throw an error upon conflict in saveHostStatus, but  
  requeue after an info message

### DIFF
--- a/pkg/controller/baremetalhost/baremetalhost_controller.go
+++ b/pkg/controller/baremetalhost/baremetalhost_controller.go
@@ -289,10 +289,15 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (result re
 		info.log.Info("saving host status",
 			"operational status", host.OperationalStatus(),
 			"provisioning state", host.Status.Provisioning.State)
-		if err = r.saveHostStatus(host); err != nil {
+		requeueNeeded, err := r.saveHostStatus(host)
+		if err != nil {
 			return reconcile.Result{}, errors.Wrap(err,
 				fmt.Sprintf("failed to save host status after %q", initialState))
 		}
+		if requeueNeeded {
+			return reconcile.Result{Requeue: true}, nil
+		}
+		info.log.Info("Updated Status")
 
 		for _, cb := range info.postSaveCallbacks {
 			cb()
@@ -354,8 +359,8 @@ func (r *ReconcileBareMetalHost) credentialsErrorResult(err error, request recon
 			// overwrites our discovered state
 			host.Status.ErrorMessage = err.Error()
 			host.Status.ErrorType = ""
-			saveErr := r.saveHostStatus(host)
-			if saveErr != nil {
+			requeueNeeded, saveErr := r.saveHostStatus(host)
+			if saveErr != nil || requeueNeeded {
 				return reconcile.Result{Requeue: true}, saveErr
 			}
 			// Only publish the event if we do not have an error
@@ -369,16 +374,18 @@ func (r *ReconcileBareMetalHost) credentialsErrorResult(err error, request recon
 	// at some point in the future.
 	case *ResolveBMCSecretRefError:
 		credentialsMissing.Inc()
-		changed, saveErr := r.setErrorCondition(request, host, metal3v1alpha1.RegistrationError, err.Error())
+		changed, requeue, saveErr := r.setErrorCondition(request, host, metal3v1alpha1.RegistrationError, err.Error())
 		if saveErr != nil {
 			return reconcile.Result{Requeue: true}, saveErr
+		} else if !changed {
+			return reconcile.Result{Requeue: true, RequeueAfter: hostErrorRetryDelay}, nil
+		} else if requeue {
+			return reconcile.Result{Requeue: true}, nil
 		}
-		if changed {
-			// Only publish the event if we do not have an error
-			// after saving so that we only publish one time.
-			r.publishEvent(request, host.NewEvent("BMCCredentialError", err.Error()))
-		}
-		return reconcile.Result{Requeue: true, RequeueAfter: hostErrorRetryDelay}, nil
+		// Only publish the event if we do not have an error
+		// after saving the first time so that we only publish one time, requeue immediately to save the status
+		r.publishEvent(request, host.NewEvent("BMCCredentialError", err.Error()))
+		return reconcile.Result{Requeue: true}, nil
 	// If we have found the secret but it is missing the required fields
 	// or the BMC address is defined but malformed we set the
 	// host into an error state but we do not Requeue it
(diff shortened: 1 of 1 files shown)


pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -148,7 +150,6 @@ type reconcileInfo struct {
 	request           reconcile.Request
 	bmcCredsSecret    *corev1.Secret
 	events            []corev1.Event
-	errorMessage      string
```
dhellmann, line 151: Why is this removed?

maelk: it is unused, so it was automatically removed by my editor. I can put it back in the next iteration

dhellmann: If it's not used, removing it is fine. Maybe it could be a separate patch in this PR, though, so the history is clearer.

maelk:
The way the Update function is called in many different places will lead to this kind of problems if the reconciliation is not directly exited. So doing something like this for the whole reconciliation to centralize the update might be helpful to avoid conflicts on the update

dhellmann:
> The way the Update function is called in many different places will lead to this kind of problems if the reconciliation is not directly exited. So doing something like this for the whole reconciliation to centralize the update might be helpful to avoid conflicts on the update

OK, we should fix those structural problems. At one point we had only one place (maybe 2?) calling Update(). If we're seeing conflicts, they represent potentially real issues in real clusters. In CI or local dev clusters it's likely that the cache just isn't updated because the BMO is running in a fast loop, but that cache issue exactly the condition we would see in a real cluster where something else was also updating a host resource. We could catch that specific error and log more quietly, but I don't think we want to be trying to work around the problem using patch.

zaneb:
Wait, what does this code even do? We are storing two copies of the status, in the Status subresource *and* an annotation, on every write now? That seems insane.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -867,12 +872,12 @@ func (r *ReconcileBareMetalHost) saveHostAnnotation(host *metal3v1alpha1.BareMet
 	if objStatus != nil {
 		// This value is copied to avoid continually updating the annotation
```
dhellmann, line 873:
```suggestion
		// These values are copied to avoid continually updating the annotation
```

dhellmann:
> Wait, what does this code even do? We are storing two copies of the status, in the Status subresource _and_ an annotation, on every write now? That seems insane.

The annotation needs to be kept up to date so when the upstream CAPI code pivots the host into the new cluster it just provisioned all of the state goes with it.

maelk:
This change does two things :
* take operationHistory out of the comparison of the status annotation, to not loop after a failure to update a status (i.e. annotation and status operationHistory are out of sync)
* do not throw an error on a conflict when updating the status or the annotation and instead merely requeue

maelk:
@zaneb It is not done at every write, it is only done at a phase change (for example from inspecting to ready). All other changes (that are not required to keep it working through the pivoting) are not updated individually in the annotation.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -865,14 +873,14 @@ func (r *ReconcileBareMetalHost) saveHostAnnotation(host *metal3v1alpha1.BareMet
...
-		// This value is copied to avoid continually updating the annotation
+		// These values are copied to avoid continually updating the annotation
 		objStatus.LastUpdated = host.Status.LastUpdated
+		objStatus.OperationHistory = host.Status.OperationHistory
 		if reflect.DeepEqual(host.Status, *objStatus) {
 			return nil
 		}
 	}
 
-	delete(host.Annotations, metal3v1alpha1.StatusAnnotation)
```
kashifest, line 875: shouldn't we delete the annotation first?

maelk: no, it's replaced by the new one

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -369,31 +374,37 @@ func (r *ReconcileBareMetalHost) credentialsErrorResult(err error, request recon
 	// at some point in the future.
 	case *ResolveBMCSecretRefError:
 		credentialsMissing.Inc()
-		changed, saveErr := r.setErrorCondition(request, host, metal3v1alpha1.RegistrationError, err.Error())
+		changed, requeue, saveErr := r.setErrorCondition(request, host, metal3v1alpha1.RegistrationError, err.Error())
```
dhellmann, line 377: What if instead of returning 3 values from this function so they have to be checked everywhere, we return a reconcile.Result{} and an error? If either is not nil, return both, otherwise drop through.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -832,29 +843,26 @@ func saveHostProvisioningSettings(host *metal3v1alpha1.BareMetalHost) (dirty boo
 	return
 }
 
-func (r *ReconcileBareMetalHost) saveHostStatus(host *metal3v1alpha1.BareMetalHost) error {
+func (r *ReconcileBareMetalHost) saveHostStatus(host *metal3v1alpha1.BareMetalHost) (bool, error) {
```
dhellmann, line 846: With 2 return values, it would be nice to name them so users of the function understand what they're for.

pkg/controller/baremetalhost/baremetalhost_controller.go
dhellmann: We shouldn't have to look inside it, only check if it's nil or not, right?

zaneb: If it returned `*reconcile.Result` I guess so, yeah. Elsewhere we only pass/return `Result`s by value.

dhellmann: Ah, yeah, good point.

maelk:
@zaneb and @dhellmann, thank you for the very valuable feedback. Following your suggestions, I have an alternative to propose : https://github.com/metal3-io/baremetal-operator/pull/566 . This removes all code setting this annotation in BMO. Instead, we set it in CAPM3, when we set the pause annotation before pivoting : https://github.com/metal3-io/cluster-api-provider-metal3/pull/105 . What do you think ? does it sound acceptable ?

kashifest:
I am not so in favor of moving the status annotation from BMO without having a proper discussion since the annotation PR hanged for almost a month and went in after quite many rounds of discussion. Are we just deciding this because we are not eager to find a proper solution to this conflict error? The reason we wanted to put it here in the first place was that we didn't want a separate peace of code to come and populate the status in BMH in the first place, I was in favor of may be updating the annotation less frequently like a change in status or if an error is set, but then we all decided it should be always follow the status

maelk:
to be clear, it is still BMO that is in charge to read the annotation and populate the status from the annotation if the status does not exist. The change I propose is how we create this annotation. Instead of BMO creating it and updating it everytime, it would only be created right before pivoting by CAPM3. CAPM3 knows when the cluster will be pivoted (more or less equivalent to when the cluster is paused) so it can add the status annotation at the same time it sets the pause annotation. The BMH won't then be reconciled again until the pause annotation is removed, so the status and annotation will be in sync until then, and the pause annotation will only be removed in the target cluster, once pivoted, when BMO will then repopulate the status based on the annotation.

One thing I am not clear about is whether there will be enough time to do that operation for all BMHs after clusterctl pauses the cluster and before clusterctl starts pivoting the resources, especially for large clusters.

kashifest:
> Instead of BMO creating it and updating it everytime, it would only be created right before pivoting by CAPM3. CAPM3 knows when the cluster will be pivoted (more or less equivalent to when the cluster is paused) so it can add the status annotation at the same time it sets the pause annotation.

That is not an ideal scenario since with cluster all the associated objects should be paused and while updating N number of BMHs status annotation and waiting to put  those in place, BMO will keep reconciling BMHs. And then ofcourse as you said, the time needed to do those in a large clusters will not be adequate.

PR #566, opened 2020-06-28
Do not set the status annotation

maelk:
This commit removes the code setting the status annotation and
requiring to reconcile the object twice. This is now done in
CAPM3.

### DIFF
--- a/pkg/controller/baremetalhost/baremetalhost_controller.go
+++ b/pkg/controller/baremetalhost/baremetalhost_controller.go
@@ -6,7 +6,6 @@ import (
 	"flag"
 	"fmt"
 	"os"
-	"reflect"
 	"strconv"
 	"strings"
 	"time"
@@ -221,6 +220,17 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (result re
 			return reconcile.Result{Requeue: true}, nil
 		}
 		reqLogger.Info("No status cache found")
+	} else {
+		//The status annotation is unneeded, as the status is present, and it will get outdated, so removing it
+		objStatus, err := r.getHostStatusFromAnnotation(host)
+		if err == nil && objStatus != nil {
+			delete(annotations, metal3v1alpha1.StatusAnnotation)
+			errStatus := r.client.Update(context.TODO(), host)
+			if errStatus != nil {
+				return reconcile.Result{}, errors.Wrap(err, "Could not delete status annotation")
+			}
+			return reconcile.Result{Requeue: true}, nil
+		}
 	}
 
 	// NOTE(dhellmann): Handle a few steps outside of the phase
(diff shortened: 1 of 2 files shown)


maelk:
This also requires https://github.com/metal3-io/cluster-api-provider-metal3/pull/106 to be merged beforehand

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -221,6 +220,17 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (result re
 			return reconcile.Result{Requeue: true}, nil
 		}
 		reqLogger.Info("No status cache found")
+	} else {
+		//The status annotation is unneeded, as the status is present, and it will get outdated, so removing it
```
dhellmann, line 224: So the status annotation will only be sticky when the host is paused? I guess that makes sense.

maelk: I don't think so, because if we were unable to save the status, then the annotation will still be there the next reconciliation loop and will only be removed once the status has been repopulated properly

maelk: the idea is to not have an outdated annotation sticking around

maelk:
@zaneb would this be ok for you ?

dhellmann:
> /hold This requires [metal3-io/cluster-api-provider-metal3#105](https://github.com/metal3-io/cluster-api-provider-metal3/pull/105) and a v1a3 equivalent to be merged first.

105 has merged. Is there a v1a3 version of the change?

maelk:
yes, we can merge this 
/hold cancel

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -221,6 +220,17 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (result re
...
+		objStatus, err := r.getHostStatusFromAnnotation(host)
+		if err == nil && objStatus != nil {
+			delete(annotations, metal3v1alpha1.StatusAnnotation)
+			errStatus := r.client.Update(context.TODO(), host)
+			if errStatus != nil {
+				return reconcile.Result{}, errors.Wrap(err, "Could not delete status annotation")
```
zaneb, line 230: Should be `errStatus`, not `err`.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -221,6 +220,17 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (result re
 			return reconcile.Result{Requeue: true}, nil
 		}
 		reqLogger.Info("No status cache found")
+	} else {
+		//The status annotation is unneeded, as the status is present, and it will get outdated, so removing it
+		objStatus, err := r.getHostStatusFromAnnotation(host)
+		if err == nil && objStatus != nil {
```
zaneb, line 226: I think this should be

    if err != nil || objStatus != nil {

A non-nil `err` can only occur if the annotation is present but cannot be unmarshalled. In this case we still want to delete it.

Actually, even better would be to just do:

    if _, ok := annotations[metal3v1alpha1.StatusAnnotation]; ok {

since that's the thing we actually care about, which saves us from unmarshalling the data for no reason.

PR #569, opened 2020-06-30
Implement an Unmanaged (discovered) state (2)

dhellmann:
If no BMC credentials are provided, move the host to a new Unmanaged
state and change the operational status to discovered, rather than
remaining without a provisioning state and in an operational error.

This has always been envisaged in the state machine diagram (with the name
Discovered rather than Unmanaged), but was not previously implemented.

This is a rebased version of #563


n1r1:
copying the comments from previous PR:
>> We also need to consider the consequences on other components that already have some assumptions on the existing API
> 
> @n1r1 are you thinking of something specific? The cluster-api-provider-* clients?

Yes. CAPx clients and downstream in OCP the UI might need some changes as well (e.g. gray out the reboot button if host has no BMC details)

pkg/apis/metal3/v1alpha1/baremetalhost_types.go
```diff
@@ -634,6 +638,11 @@ func (host *BareMetalHost) getLabel(name string) string {
 	return host.Labels[name]
 }
 
+// HasBMCDetails returns true if the BMC details are set
+func (host *BareMetalHost) HasBMCDetails() bool {
+	return host.Spec.BMC.Address != "" || host.Spec.BMC.CredentialsName != ""
```
n1r1, line 643: looks like this assumes that host.spec.BMC exists (not `nil`). is this a valid assumption?

dhellmann: That field isn't a pointer, so it should be OK.

dhellmann:
yum mirror access failed

/test-integration

dhellmann:
Since this is technically an API change and I keep asking people to write up docs for those, I've put https://github.com/metal3-io/metal3-docs/pull/120 up for review.

dhellmann:
This will need to be rebased on top of #571

dhellmann:
Rebased after #571 merged

maelk:
/test-integration
Even if this is not yet on CAPM3 side, the integration tests should pass

dhellmann:
/hold clear
/cc @zaneb 

This is ready for review and merging.

PR #571, opened 2020-07-02
define the Unmanaged state

dhellmann:
Define a new state for newly-created Hosts with no BMC details.

We need the state definition so we can add it to CAPM3 so that does
not break when #569 merges.

See https://github.com/metal3-io/metal3-docs/pull/120 for more
details.


kirankt:
Looks good to me.
/approve

maelk:
/lgtm
I'll take care of uplifting in CAPM3 once this merges

maelk:
@dhellmann Can you please review the two PRs ? Then it should be ready on CAPM3 side.

PR #573, opened 2020-07-03
WIP: Add support for mutual TLS with ironic

felixest:
Mutual TLS will be used to authenticate against ironic and
ironic-inspector. When the environment variables IRONIC_CERT_FILE,
IRONIC_KEY_FILE and IRONIC_TRUSTED_CA_FILE are all set, mutual TLS will
be used to communicate with the ironic endpoints.

The http.Client with the mTLS configuration will load the client
certificate from the disk every time a request is made. Therefore, it
allows to rotate the client certificate without a restart. The CA file
is only loaded when a new provisioner is created. This happens during
every reconciliation. Therefore, a rotation of the CA certificate
without restart is also supported.

**WIP:** Needs testing with all other mTLS components in place. Deployment needs to be adapted to include cert-manager.

### DIFF
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -2,10 +2,12 @@ package ironic
 
 import (
 	"fmt"
+	"net/http"
 	"os"
 	"strings"
 	"time"
 
+	"go.etcd.io/etcd/pkg/transport"
 	"sigs.k8s.io/yaml"
 
 	"github.com/gophercloud/gophercloud"
@@ -34,10 +36,16 @@ var provisionRequeueDelay = time.Second * 10
 var powerRequeueDelay = time.Second * 10
 var introspectionRequeueDelay = time.Second * 15
 var softPowerOffTimeout = time.Second * 180
+var tlsConnectionTimeout = time.Second * 30
 var deployKernelURL string
 var deployRamdiskURL string
 var ironicEndpoint string
 var inspectorEndpoint string
+var ironicCertFile string
+var ironicKeyFile string
+var ironicTrustedCAFile string
+var ironicUseTLS bool
+var ironicUseMutualTLS bool
 
 const (
 	// See nodes.Node.PowerState for details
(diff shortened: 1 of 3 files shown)


pkg/provisioner/ironic/ironic.go
```diff
@@ -70,6 +77,19 @@ func init() {
...
 		os.Exit(1)
 	}
+	ironicCertFile = os.Getenv("IRONIC_CERT_FILE")
+	ironicKeyFile = os.Getenv("IRONIC_KEY_FILE")
+	ironicTrustedCAFile = os.Getenv("IRONIC_TRUSTED_CA_FILE")
+	if ironicCertFile != "" || ironicKeyFile != "" || ironicTrustedCAFile != "" {
+		if ironicCertFile != "" && ironicKeyFile != "" && ironicTrustedCAFile != "" {
+			ironicUseMutualTLS = true
+		} else {
+			fmt.Fprintf(os.Stderr, "Cannot start: IRONIC_CERT_FILE, IRONIC_KEY_FILE and IRONIC_TRUSTED_CA_FILE variables all must be set to use mTLS")
```
dhellmann, line 87: This error message would be more helpful if it showed which variables are not set.

pkg/provisioner/ironic/ironic.go
```diff
@@ -124,6 +160,19 @@ func newProvisioner(host *metal3v1alpha1.BareMetalHost, bmcCreds bmc.Credentials
 	if err != nil {
 		return nil, err
 	}
+	if ironicUseMutualTLS {
```
dhellmann, line 163: It feels a little odd to instantiate a NoAuth client and then modify its HTTPClient. Is that the way it's supposed to be used?

I see that gophercloud has a different client implementation for basic auth in https://github.com/gophercloud/gophercloud/blob/master/openstack/baremetal/httpbasic/requests.go. Would it make sense to add this new client type to gophercloud? Or do we expect this use pattern to be unique to metal3?

If this is something we only want in metal3, maybe we should move the ironic client creation code into its own function that can be invoked here to separate the details of setting up those clients from the provisioner.

maelk: I would be reluctant to modify gophercloud considering that this is not something supported by Ironic, but an Apache-based setup that we are doing. I'd say it's specific to Metal3. The whole clients creation and adaptation should be indeed moved to its own function. We could implement the exact same thing in BMO, creating our own type of clients like https://github.com/gophercloud/gophercloud/blob/master/openstack/baremetal/httpbasic/requests.go

maelk: the example followed is in the FAQ : https://github.com/gophercloud/gophercloud/blob/master/docs/FAQ.md#implementing-default-logging-and-re-authentication-attempts

zaneb: Keystone supports auth using client certificates, and python-openstackclient supports using client certificates. So I don't actually think it would be unreasonable to add this in gophercloud. (Probably no need to block on that though.)

pkg/provisioner/ironic/ironic.go
```diff
@@ -70,6 +78,29 @@ func init() {
...
+		if ironicTrustedCAFile == "" {
+			fmt.Fprintf(os.Stderr, "Cannot start: No IRONIC_TRUSTED_CA_FILE variable set (required for mTLS)")
+			os.Exit(1)
+		}
+		ironicUseTLS = true
+
+		if ironicCertFile == "" && ironicKeyFile != "" {
+			fmt.Fprintf(os.Stderr, "Cannot start: No IRONIC_CERT_FILE variable set (required for mTLS)")
+			os.Exit(1)
+		} else if ironicCertFile == "" && ironicKeyFile != "" {
```
zaneb, line 94: copy-paste error

zaneb:
I moved all the code for configuring auth and creating clients to a [separate module](https://github.com/metal3-io/baremetal-operator/tree/master/pkg/provisioner/ironic/clients), which should help keep things tidy as we add more code like this.

One suggestion I have after implementing the HTTP basic auth support (in #601): in that case I ended up creating a single environment variable `METAL3_AUTH_ROOT_DIR` (which defaults to `/opt/metal3/auth`), and within that directory the layout is fixed. Ironic creds go in the `ironic` subdirectory and inspector creds go in the `inspector` subdirectory, and the expected content of those directories is hardcoded. We detect whether auth is configured based on whether the relevant files exist. Basically this means that all we have to do when deploying is bind-mount the secrets into the container, not bind-mount them *and* tell the operator via environment variables where we put them. It would be nice if we could control all forms of authentication through the same `METAL3_AUTH_ROOT_DIR` environment variable.

(Note that files in the same directory must correspond to keys in the same k8s Secret, so take that into account when you are designing the layout.)

maelk:
I'm breaking down this PR. A first part to support TLS is in https://github.com/metal3-io/baremetal-operator/pull/631

Issue #575, opened 2020-07-06
Update the testing documentation with test tool installation guidance


demonCoder95:
**Problem Description**
Running ``make lint`` in the parent dir returns the following failure:
```bash
which golint 2>&1 >/dev/null || make OPATH/bin/golint
find ./pkg ./cmd -type f -name \*.go  |grep -v zz_ | xargs -L1 golint -set_exit_status
./pkg/apis/metal3/v1alpha1/baremetalhost_types.go:603:31: should drop = "" from declaration of var emptyErrType; it is the zero value
Found 1 lint suggestions; failing.
./pkg/controller/baremetalhost/baremetalhost_controller.go:50:29: should omit type int from declaration of var maxConcurrentReconciles; it will be inferred from the right-hand side
Found 1 lint suggestions; failing.
Makefile:82: recipe for target 'lint' failed
make: *** [lint] Error 123
```

**More details**
I've tested on the latest commit on master which is ``c1a792c0a287aea870dd8449bd49f3217e670ed1``.

**Suggested Solution**
This should be simple enough, just fixing the var declaration in baremetalhost_types.go file. I can send a patch for this.

Edit: This was due to an outdated linter. Testing documentation is being patched to address this issue now.

PR #577, opened 2020-07-06
Update testing documentation with tool installation directions

demonCoder95:
Fixing a couple of variable declarations that were being
complained about by ``make lint``.

Edit: This was because of an outdated version of ``golint``. This patch will now focus on updating the testing documentation to add directions for installing latest testing tools.

Addresses: #575

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -36,3 +36,66 @@ Run linters test before pushing your commit
 ```bash
 make lint
 ```
+
+## Using the Hack scripts
+
+The repository contains a ``hack`` directory which has some very useful scripts
+for setting up containerized testing environments. It also has Dockerfiles to
+allow you to generate images for your own use.
+
+All the scripts accept the ``CONTAINER_RUNTIME`` environment variable with default
+of ``podman``. This you can edit to ``docker`` as well.
+
+### Example of running ``golint`` in a Docker container
+
+1. Start by setting the environment variable appropriately
+
+    ```bash
+    export CONTAINER_RUNTIME=docker
+    ```
+
+2. From the operator parent dir, you can invoke the hack scripts
+
+    ```bash
+    ./hack/golint.sh
+    ```
+
+    **Note** It's important to be in the operator parent dir because it
+    contains the ``Makefile`` that is used in running the tests for the
+    operator. Otherwise, you'll see the following error
+
+    ```bash
+    sh: 0: Can't open /go/src/github.com/metal3-io/baremetal-operator/hack/golint.sh
+    ```
+
+3. Upon successful execution, you should see the following output. I already
+    had all the images available from a previous run, you might see the images
+    getting downloaded if you're running for the very first time.
+
+    ```bash
+    + IS_CONTAINER=false
+    + CONTAINER_RUNTIME=docker
+    + [ false != false ]
+    + docker run --rm --env IS_CONTAINER=TRUE --volume /home/noor/go/src/github.com/metal3-io/baremetal-operator:/go/src/github.com/metal3-io/baremetal-operator:ro,z\
+    --entrypoint sh --workdir /go/src/github.com/metal3-io/baremetal-operator\
+    quay.io/metal3-io/golint:latest /go/src/github.com/metal3-io/baremetal-operator/hack/golint.sh
+    + IS_CONTAINER=TRUE
+    + CONTAINER_RUNTIME=podman
+    + [ TRUE != false ]
+    + export XDG_CACHE_HOME=/tmp/.cache
+    + make lint
+    which golint 2>&1 >/dev/null || make OPATH/bin/golint
+    find ./pkg ./cmd -type f -name \*.go  |grep -v zz_ | xargs -L1 golint -set_exit_status
+    ```
+
+## Getting the latest tooling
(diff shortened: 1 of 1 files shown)


dhellmann:
Which version of lint shows these errors? The one in our CI job doesn't report them, so I wonder if we're out of date?

demonCoder95:
I'm not sure how to find the version of lint, I've tried various flags on the golint binary. I installed golint with APT which shows the following package version``Version: 0.0+git20161013.3390df4-1``

dhellmann:
> I'm not sure how to find the version of lint, I've tried various flags on the golint binary. I installed golint with APT which shows the following package version`Version: 0.0+git20161013.3390df4-1`

I usually install the developer tools with `go get` so I can have the newer versions. In this case, `go get golang.org/x/lint/golint`.

The change you're making isn't wrong, I would just hate for us to all be using different linters. That's the point of running the lint job in CI.

demonCoder95:
Ah! I see. Thanks for the pointer. I've just updated my linter and after running the linter again on master, there are no problems. 

I think it might be worthwhile to add this to the [testing documentation](https://github.com/metal3-io/baremetal-operator/blob/master/docs/testing.md), so that somebody just starting up might follow the same?

dhellmann:
> Ah! I see. Thanks for the pointer. I've just updated my linter and after running the linter again on master, there are no problems.
> 
> I think it might be worthwhile to add this to the [testing documentation](https://github.com/metal3-io/baremetal-operator/blob/master/docs/testing.md), so that somebody just starting up might follow the same?

Updating the docs is a great idea. Do you want to do that instead of applying this PR? If so, you could mention the scripts in the `hack` directory as alternatives for installing the tools locally.

demonCoder95:
That sounds great! I can revert these small changes within the same PR and add the doc update instead. The scripts in hack folder are very helpful, and it'll be great to have them indexed in the documentation!

dhellmann:
> That sounds great! I can revert these small changes within the same PR and add the doc update instead. The scripts in hack folder are very helpful, and it'll be great to have them indexed in the documentation!

Perfect, thanks!

demonCoder95:
@dhellmann had some issues with commit history. It's ready now. PTAL.

maelk:
lgtm, but maybe the PR and the commit could be renamed to be a bit more descriptive of what the change actually does ? thank you

demonCoder95:
Oh yes, I'll have to do that, since the reason the change started has changed, and it turned into a documentation patch. I'll update the descriptions accordingly. Thanks you @maelk

maelk:
thank you. Please amend your commit to rename it too.

demonCoder95:
Done. Thanks for pointing it out @maelk

maelk:
Sure. And I was just thinking that maybe adding an explanation about the CONTAINER_RUNTIME variable could be nice, so that people that have docker and not podman installed can easily run the tests in containers too. Maybe together with an example ?

demonCoder95:
That sounds good. Working on it.

demonCoder95:
I just noticed that the IS_CONTAINER variable is inversed in logic. IS_CONTAINER=true actually ends up NOT running the tests in the container, and vice versa. Is it supposed to happen? or am I missing something? Just noticed this when making the example for docs.

demonCoder95:
Waiting to put in the command output until I resolve the IS_CONTAINER variable issue. It seems to be it is incorrectly placed in all the scripts.

maelk:
The IS_CONTAINER is correctly placed. There : https://github.com/metal3-io/baremetal-operator/blob/master/hack/golint.sh#L8 it checks that if the variable is not false, then it runs the make target. It is false by default, so it will by default go to run docker, that will run a container with that variable set to true, so then it will run the tests. So it is correct as it is now afaik.

docs/testing.md
```diff
@@ -36,3 +36,46 @@ Run linters test before pushing your commit
...
+2. ``CONTAINER_RUNTIME`` default: podman; specify the runtime you want to use
+
+As you can see, by default, the tests will not run in a container and if chosen to
+be containerized, will do so with podman, you can tweak these variables to your liking.
+
+### Example of running golint in a docker container
+Start by setting the environment variables appropriately
+
+```bash
+export IS_CONTAINER=true
```
maelk, line 57: you should not set that. If you set it, it will not run as a container, it will run locally. It is just a flag so that docker or podman would run the same script, but identify that it is running as a container already.

dhellmann: I think we should prefer people use `make` if they have the tools installed and don't need the containerized versions. Otherwise they should just run the script, and optionally set the `CONTAINER_RUNTIME`, but never set `IS_CONTAINER`.

demonCoder95: Oh, got it. Thanks! Sending a patch now.

demonCoder95:
@maelk @dhellmann Just patched with the complete example and some improvements. Let me know what you think!

demonCoder95:
@maelk @dhellmann fixed the linter problems, should be good to go

dhellmann:
/lgtm
/approve

Thanks @demonCoder95!

demonCoder95:
Happy to help!

PR #581, opened 2020-07-07
Add ValidateForAdmission function

honza:
This function determines if a new BareMetalHost can be added to the
cluster.  It's suitable to be run from a `ValidateCreate` admission
webhook.

Currently, it checks for BMC address uniquess, and BootMACAddress
uniqueness.

### DIFF
--- a/pkg/validation/admission.go
+++ b/pkg/validation/admission.go
@@ -1,0 +1,70 @@
+package validation
+
+import (
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+
+	metal3v1alpha1 "github.com/metal3-io/baremetal-operator/pkg/apis/metal3/v1alpha1"
+)
+
+// ValidateHostForAdmission takes a list of existing hosts, and a host we are
+// trying to add to the cluster.  It returns an error if the host fails
+// uniqueness validation, and cannot be added to the cluster.  The optional
+// third argument `old` denotes the previous version of `host` in the case of an
+// update.
+//
+// This is designed to be run by an admission webhook, `ValidateCreate`, and
+// `ValidateUpdate`.
+func ValidateHostForAdmission(existingHosts []metal3v1alpha1.BareMetalHost, host metal3v1alpha1.BareMetalHost, old *metal3v1alpha1.BareMetalHost) error {
+	if host.Spec.BMC.Address == "" && host.Spec.BootMACAddress == "" && host.Spec.BMC.CredentialsName == "" {
+		return nil
+	}
+
+	if len(existingHosts) == 0 {
+		return nil
+	}
+
+	var allErrs field.ErrorList
+
+	for _, existingHost := range existingHosts {
+		if old != nil && old.Namespace == existingHost.Namespace && old.Name == existingHost.Name {
+			continue
+		}
+
+		// The BMC address is optional so only validate it if it's set
+		if host.Spec.BMC.Address != "" && host.Spec.BMC.Address == existingHost.Spec.BMC.Address {
+			allErrs = append(allErrs, field.Invalid(
+				field.NewPath("spec", "bmc", "address"),
+				host.Spec.BMC.Address,
+				"is not unique",
+			))
+		}
+
+		// The BootMACAddress is optional so only validate it if it's set
+		if host.Spec.BootMACAddress != "" && host.Spec.BootMACAddress == existingHost.Spec.BootMACAddress {
+			allErrs = append(allErrs, field.Invalid(
+				field.NewPath("spec", "bootMACAddress"),
+				host.Spec.BootMACAddress,
+				"is not unique",
+			))
+		}
+
+		// The BMC credentials is optional so only validate it if it's set
+		if host.Spec.BMC.CredentialsName != "" && host.Spec.BMC.CredentialsName == existingHost.Spec.BMC.CredentialsName {
+			allErrs = append(allErrs, field.Invalid(
+				field.NewPath("spec", "bmc", "credentialsName"),
+				host.Spec.BMC.CredentialsName,
+				"is not unique",
+			))
+		}
+	}
+
+	if len(allErrs) == 0 {
+		return nil
+	}
+
+	return apierrors.NewInvalid(
+		metal3v1alpha1.SchemeGroupVersion.WithKind("BareMetalHost").GroupKind(),
+		host.Name,
+		allErrs)
+}
--- a/pkg/validation/admission.go
+++ b/pkg/validation/admission.go
@@ -5,6 +5,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
 	metal3v1alpha1 "github.com/metal3-io/baremetal-operator/pkg/apis/metal3/v1alpha1"
+	"github.com/metal3-io/baremetal-operator/pkg/bmc"
 )
 
 // ValidateHostForAdmission takes a list of existing hosts, and a host we are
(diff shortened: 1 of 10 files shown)


pkg/apis/metal3/v1alpha1/baremetalhost_types.go
```diff
@@ -853,6 +857,50 @@ func (host *BareMetalHost) GetImageChecksum() (string, string, bool) {
...
+	err := c.List(ctx, &hostList, opts)
+
+	if err != nil {
+		return err
+	}
+
+	var allErrs field.ErrorList
+
+	for _, existingHost := range hostList.Items {
+		if host.Spec.BMC.Address == existingHost.Spec.BMC.Address {
```
stbenjam, line 879: Do we want to check for `.Address == ""` earlier? For unmanaged hosts, the BMC address will be optional. If this lands first perhaps we defer adding that check until the unmanaged PR lands https://github.com/metal3-io/baremetal-operator/pull/569

pkg/apis/metal3/v1alpha1/baremetalhost_types.go
```diff
@@ -853,6 +857,50 @@ func (host *BareMetalHost) GetImageChecksum() (string, string, bool) {
...
+// and cannot be added to the cluster.  This is ordinarily run by an
+// admission webhook.
+func (host *BareMetalHost) ValidateForAdmission(c client.Client) error {
+	ctx := context.Background()
+	hostList := BareMetalHostList{}
+	opts := &client.ListOptions{
+		Namespace: host.Namespace,
+	}
+
+	err := c.List(ctx, &hostList, opts)
```
stbenjam, line 870: Can you use field selectors to do a query for existing BMH with the same MAC or BMC address instead of listing all the hosts (of which there could be thousands) e.g. https://godoc.org/sigs.k8s.io/controller-runtime/pkg/client#ListOptions

Not sure if the API tools work that way to allow us to do such a query (or if it's even better)

dhellmann:
Is it possible to do this as a standalone helper function in its own package or does the web hook framework assume it is a method of the type? Adding it this way is going to cause anything consuming the host resource to pull in all of the new dependencies as well, even if they are not implementing the web hook.

honza:
> Is it possible to do this as a standalone helper function in its own package or does the web hook framework assume it is a method of the type? Adding it this way is going to cause anything consuming the host resource to pull in all of the new dependencies as well, even if they are not implementing the web hook.

Yes. The pattern that I have seen is that the webook handler contains a one-liner which is a method call on the resource type. But it's easy to move it to a new package to avoid the dependency problem. I added a new package called `validation`.

pkg/apis/metal3/v1alpha1/baremetalhost_types.go
honza: I'm not sure how think about uniqueness in this context given that multiple fields are nullable. Do we have a list of hard requirements/rules?

pkg/apis/metal3/v1alpha1/baremetalhost_types.go
honza: > Can you use field selectors to do a query for existing BMH with the same MAC or BMC address instead of listing all the hosts (of which there could be thousands)?

No. The field selector only operates on a small number of fields, think `metadata.name`.

There seem to be many fields we need to check, and doing a query for each one seems slow. This way we grab all of the resources at once, and can quickly run through them to validate them.

pkg/apis/metal3/v1alpha1/baremetalhost_types.go
dhellmann: We want the BMC.Address to be unique, if it is set, to avoid having 2 resources managing the same physical host.

We want the BMC.CredentialsName to be unique, if it is set, to avoid having 2 host resources taking ownership of the same Secret resource.

We want the BootMACAddress to be unique to avoid having 2 resources trying to claim the same data for a discovered node. The current draft of the code says the field can be empty, but I don't think we want to support that given that we've seen bugs recently.

pkg/apis/metal3/v1alpha1/baremetalhost_types.go
dhellmann: Is the client used the same type of caching client that a controller uses?

pkg/apis/metal3/v1alpha1/baremetalhost_types.go
honza: I have added a check for the credentials name field.

pkg/validation/admission_test.go
```diff
@@ -0,0 +1,128 @@
...
+						defaultBMCCredentialsName,
+						"is not unique",
+					),
+				}),
+		},
+		{
+			Scenario: "BMC credentials name is unique",
+			Host:     newHost("", "unique", ""),
+			Expected: nil,
+		},
```
dhellmann, line 117: It looks like we could use a test for multiple hosts with none of the optional fields set.

pkg/validation/admission_test.go
```diff
@@ -0,0 +1,139 @@
...
+		t.Run(tc.Scenario, func(t *testing.T) {
+			existingHost := newDefaultHost()
+			c := fakeclient.NewFakeClient(&existingHost)
+			err := ValidateHostForAdmission(c, tc.Host)
+			assert.Equal(t, tc.Expected, err)
+		})
+	}
+}
+
+func TestCanBeAdmittedMultiple(t *testing.T) {
```
dhellmann, line 131: This should also test admission for a host with no values set. Passing them to the fake client won't do that, right?

honza: Done

pkg/validation/admission.go
```diff
@@ -0,0 +1,72 @@
...
+// admission webhook.
+func ValidateHostForAdmission(c client.Client, host metal3v1alpha1.BareMetalHost) error {
+	if host.Spec.BMC.Address == "" && host.Spec.BootMACAddress == "" && host.Spec.BMC.CredentialsName == "" {
+		return nil
+	}
+
+	ctx := context.Background()
+	hostList := metal3v1alpha1.BareMetalHostList{}
+	opts := &client.ListOptions{
+		Namespace: host.Namespace,
```
zaneb, line 24: We need to search all namespaces here because nothing about the Host being in another namespace makes it better when they're both fighting for control over Ironic. In fact, that's probably worse.

pkg/validation/admission.go
```diff
@@ -0,0 +1,72 @@
...
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	metal3v1alpha1 "github.com/metal3-io/baremetal-operator/pkg/apis/metal3/v1alpha1"
+)
+
+// ValidateHostForAdmission returns an error if the host fails uniqueness
+// validation, and cannot be added to the cluster.  This is ordinarily run by an
+// admission webhook.
+func ValidateHostForAdmission(c client.Client, host metal3v1alpha1.BareMetalHost) error {
```
zaneb, line 16: Once we have a validating webhook we'll almost certainly want to use it for other things that don't involve uniqueness across the whole cluster. It might pay to group this stuff that does in a function with a more specific name.

pkg/validation/admission.go
```diff
@@ -0,0 +1,72 @@
...
+		return nil
+	}
+
+	ctx := context.Background()
+	hostList := metal3v1alpha1.BareMetalHostList{}
+	opts := &client.ListOptions{
+		Namespace: host.Namespace,
+	}
+
+	err := c.List(ctx, &hostList, opts)
```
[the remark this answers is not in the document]
honza: By [default](https://github.com/kubernetes-sigs/controller-runtime/blob/f98ea979b60d5d830620f54044b9913ea744b32c/pkg/manager/manager.go#L347), the runtime creates a caching client. 

Do we have any means of performing atomic operations on a set of resources?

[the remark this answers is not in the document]
zaneb: Only using the kubernetes API. So we'd have to have some sort of CRD in which MAC addresses and BMC addresses could be claimed. There's no way that I know of to make that atomic with creating the BareMetalHost though, so I'm not even sure how that would work.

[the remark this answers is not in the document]
honza: I removed the client code from the function, and changed it to accept a list of existing resources; we'll leave it up to the caller for now to deal with caching/atomic semantics.

pkg/validation/admission.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	if len(existingHosts) == 0 {
+		return nil
+	}
+
+	var allErrs field.ErrorList
+
+	for _, existingHost := range existingHosts {
+		if old != nil && old.Name == existingHost.Name {
```
zaneb, line 30: Need to check namespaces here also.

pkg/validation/admission.go
```diff
@@ -0,0 +1,70 @@
...
+
+	var allErrs field.ErrorList
+
+	for _, existingHost := range existingHosts {
+		if old != nil && old.Namespace == existingHost.Namespace && old.Name == existingHost.Name {
+			continue
+		}
+
+		// The BMC address is optional so only validate it if it's set
+		if host.Spec.BMC.Address != "" && host.Spec.BMC.Address == existingHost.Spec.BMC.Address {
```
zaneb, line 35: I think we might want to dig down a bit further rather than just comparing strings. e.g. redfish://192.168.122.1 and redfish-ilo5://192.168.122.1 are the same BMC.
I'm not sure how the path components come into it. Is it possible with e.g. redfish that multiple hosts share a BMC with the same IP address but different paths?

dhellmann: Yes, chassis can have multiple hosts with 1 controller (think blade servers). Each host would have a different path, with (for example) the `/Systems/1` part changing based on which host is being used.

zaneb: Makes sense, so we should compare the hostname, path (for redfish), and presumably port (for IPMI/iDRAC/iRMC), but not the scheme.

dhellmann: How much knowledge of that sort of thing do we want to embed here, vs. using the `AccessDetails` implementations to do the comparison?

honza:
I just reimplemented the logic that performs the uniqueness check. I added a new function to the `AccessDetails` interface which asks implementers to indicate how to compare nodes for uniqueness. 

The unit tests could use some work but I wanted to see first if this was in the general direction we wanted to go in.

pkg/validation/admission.go
```diff
@@ -0,0 +1,67 @@
...
+		}
+
+		existingBmc, err := bmc.NewAccessDetails(existingHost.Spec.BMC.Address, false)
+
+		// If access details aren't set, there is nothing to validate.
+		if err != nil {
+			continue
+		}
+
+		// If the two hosts don't share drivers, we can assume that they don't conflict
```
dhellmann, line 45: I don't think that's a safe assumption. BMCs can support multiple protocols.

zaneb: A safer assumption might be:

* same IP but different driver -> reject
* same IP and same driver -> compare unique access path
* different IP -> accept

This has the advantage of blocking anyone trying to use different drivers for different blades in a blade server :D

We can resolve any hostnames to IP addresses to avoid aliasing at that level. Then the only thing we won't catch is things with multiple IP addresses (dual stack IPv4/IPv6 is probably the only likely one to happen in the real world, I'd expect).

dhellmann:

/hold

PR #587, opened 2020-07-12
Adding support for hardware type iLO

ankit-kumar5:
This commit add support for iLO hardware type to
deploy ilo based systems using BMO.

I haven't added the changes related to hardware type ilo in ironic.conf. Since it doesn't support boot_interface ipxe or pxe. So it was giving error **Boot interface ipxe not supported** when i was adding ilo in **enabled_hardware_types** and restarting the ironic service.

So i had to remove all the other hardware type to make ilo work.

### DIFF
--- a/pkg/bmc/access_test.go
+++ b/pkg/bmc/access_test.go
@@ -311,6 +311,36 @@ func TestParse(t *testing.T) {
 			Hostname: "192.168.122.1",
 			Path:     "",
 		},
+
+		{
+                        Scenario: "ilo url",
+                        Address:  "ilo://192.168.122.1",
+                        Type:     "ilo",
+                        Port:     "",
+                        Host:     "192.168.122.1",
+                        Hostname: "192.168.122.1",
+                        Path:     "",
+                },
+
+		{
+                        Scenario: "ilo url, ipv6",
+                        Address:  "ilo://[fe80::fc33:62ff:fe83:8a76]",
+                        Type:     "ilo",
+                        Port:     "",
+                        Host:     "fe80::fc33:62ff:fe83:8a76",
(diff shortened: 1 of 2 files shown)


dhellmann:
How would this driver be different from the `ilo5-redfish` driver added in #411?

ankit-kumar5:
ilo5-redfish internally uses redfish driver(**hardware type redfish**) and its for redfish compliant systems. It does not take the ilo path. But my changes are for ilo4 based systems which are not redfish compliant and it uses hardware type **ilo**.

ankit-kumar5:
> How would this driver be different from the `ilo5-redfish` driver added in #411?

ilo5-redfish internally uses redfish driver(hardware type redfish) and its for redfish compliant systems. It does not take the ilo path. But my changes are for ilo4 based systems which are not redfish compliant and it uses hardware type ilo.

dhellmann:
> > How would this driver be different from the `ilo5-redfish` driver added in #411?
> 
> ilo5-redfish internally uses redfish driver(**hardware type redfish**) and its for redfish compliant systems. It does not take the ilo path. But my changes are for ilo4 based systems which are not redfish compliant and it uses hardware type **ilo**.

I see. If this is meant for ilo4 systems, let's include the "4" in the name to make that clear. Even if Ironic's driver is called "ilo" it's OK for us to use a different name to help the user make the right choice.

ankit-kumar5:
> Could you update the list of drivers in https://github.com/metal3-io/baremetal-operator/blob/master/docs/api.md#spec-fields to add this new one with enough of a description for a user to understand when they should choose it over the existing ilo5 variants?

Sorry about the mess. By mistake i closed this pull request.  I have created a new pull request.
https://github.com/metal3-io/baremetal-operator/pull/595

PR #589, opened 2020-07-13
do not log for paused hosts

dhellmann:
Do not log reconcile attempts for paused hosts. This keeps the logs
quieter for debugging.


zaneb:
We're currently requeuing every 30s when paused... why do we need to requeue at all? By definition nothing can happen that will make us want to reconcile without removing the annotation, which should trigger the host to be reconciled without anything having been requeued. Wouldn't it be better to do that and log quite prominently that the host is paused?
Having it silent could be very confusing when you're trying to debug. And moving the logging code below the `Get()` call means we no longer know which host we were reconciling if there is an error in getting the host.

dhellmann:
Great point, I will make that change.

PR #591, opened 2020-07-14
stop reconciling paused hosts

dhellmann:
If a host is paused, there is no need to reconcile it
periodically. Any change that removes the pause annotation will
trigger normal reconciliation.

See comments on #589 for more context

### DIFF
--- a/pkg/controller/baremetalhost/baremetalhost_controller.go
+++ b/pkg/controller/baremetalhost/baremetalhost_controller.go
@@ -41,7 +41,6 @@ import (
 
 const (
 	hostErrorRetryDelay    = time.Second * 10
-	pauseRetryDelay        = time.Second * 30
 	unmanagedRetryDelay    = time.Minute * 10
 	rebootAnnotationPrefix = "reboot.metal3.io"
 )
@@ -173,6 +172,10 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (result re
 		}
 	}()
 
+	reqLogger := log.WithValues("Request.Namespace",
+		request.Namespace, "Request.Name", request.Name)
+	reqLogger.Info("Reconciling BareMetalHost")
+
 	// Fetch the BareMetalHost
 	host := &metal3v1alpha1.BareMetalHost{}
 	err = r.client.Get(context.TODO(), request.NamespacedName, host)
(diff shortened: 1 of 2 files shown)


pkg/controller/baremetalhost/baremetalhost_controller_test.go
```diff
@@ -290,8 +290,7 @@ func TestPause(t *testing.T) {
 
 	tryReconcile(t, r, host,
 		func(host *metal3v1alpha1.BareMetalHost, result reconcile.Result) bool {
-			if result.Requeue && result.RequeueAfter == pauseRetryDelay &&
-				len(host.Finalizers) == 0 {
+			if !result.Requeue {
```
zaneb, line 293: Should still check that the finalizer is not set here, shouldn't we?

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -192,12 +195,11 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (result re
...
 		if _, ok := annotations[metal3v1alpha1.PausedAnnotation]; ok {
-			return reconcile.Result{Requeue: true, RequeueAfter: pauseRetryDelay}, nil
+			reqLogger.Info("host is paused, no work to do")
+			return reconcile.Result{Requeue: false}, nil
 		}
 	}
 
-	reqLogger := log.WithValues("Request.Namespace",
-		request.Namespace, "Request.Name", request.Name)
 	reqLogger.Info("Reconciling BareMetalHost")
```
zaneb, line 203: Is it worth moving this back so that it is logged before any error from line 191?

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -192,12 +195,11 @@ func (r *ReconcileBareMetalHost) Reconcile(request reconcile.Request) (result re
 	annotations := host.GetAnnotations()
 	if annotations != nil {
 		if _, ok := annotations[metal3v1alpha1.PausedAnnotation]; ok {
-			return reconcile.Result{Requeue: true, RequeueAfter: pauseRetryDelay}, nil
```
zaneb, line 195: Can delete the definition of `pauseRetryDelay` also now I think.

pkg/controller/baremetalhost/baremetalhost_controller.go
dhellmann: done

pkg/controller/baremetalhost/baremetalhost_controller.go
dhellmann: done

pkg/controller/baremetalhost/baremetalhost_controller_test.go
dhellmann: Done

dhellmann:
The prow cluster is borked so 2 of the jobs here are stuck. The other jobs passed, including the integration job, so even if landing this means future PRs are broken in this repo it shouldn't break CAPM3 or anything else. I'm going to manually merge this PR.

PR #592, opened 2020-07-15
Avoid extra requeues when credentials missing

zaneb:
The extra requeues added in 871926d407c4758f96a7fc011f254a57b2c0eaa1 to
ensure that the Status got written immediately after updating the status
annotation were incompletely reverted by
f1b81a06527cb41b2b7e0212e76fc86976b32324, after which they are no longer
required as we no longer write a status annotation and just write the
Status immediately. This reverts the remaining changes to prevent
unnecessary requeueing.


maelk:
lgtm

PR #593, opened 2020-07-15
:running: add release docs and tools

dhellmann:

We should start tagging formal releases of the baremetal-operator, like
we do for other metal3 components. This PR includes docs and tools for
releasing, based on the ones in CAPM3.


docs/releasing.md
```diff
@@ -0,0 +1,73 @@
...
+## Output
+
+### Expected artifacts
+
+1. A container image of the shared baremetal-operator
+1. A git tag
+1. A release on Github containing:
+    - A manifest file - `infrastructure-components.yaml`
+    - A metadata file - `metadata.yaml`
+    - A cluster template - `cluster-template.yaml`
```
maelk, line 18: I don't think we should have a metadata and cluster template files for BMO, since they are not part of capm3. Having a manifest file that allows a user to easily deploy BMO (i.e. the kustomization output) could be nice to have.

docs/releasing.md
```diff
@@ -0,0 +1,73 @@
...
+
+Multiple additional actions are required in the Metal3 project
+
+### Update the Jenkins jobs
+
+For each minor or major release, two jobs need to be created :
+
+- a master job that runs on a regular basis
+- a PR verification job that is triggered by a keyword on a PR targeted for that
+  release branch.
```
maelk, line 61: we will need to discuss how we deal with the BMO versions in the tests, if we will need multiple jobs etc. and the compatibility matrix with CAPM3 to cover it with the tests.

docs/releasing.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Update Metal3-dev-env
+
+Metal3-dev-env variables need to be modified. After a major or minor release,
+the new minor version (that follows CAPI versioning) should point to master for
+BMO and the released version should point to the release branch.
```
maelk, line 67: Metal3-dev-env will need a bit of rework to be able to give a BMO version as parameter (for the branch and tag)

hack/markdownlint.sh
```diff
@@ -9,7 +9,9 @@ if [ "${IS_CONTAINER}" != "false" ]; then
   TOP_DIR="${1:-.}"
   find "${TOP_DIR}" \
        \( -path ./vendor -prune -o \
-       -path ./tools/operator-sdk -prune \) \
+       -path ./tools/operator-sdk -prune -o \
+       -path ./.github -prune \
```
maelk, line 13: is -prune duplicated ? or is it expected after each path ?

hack/tools/go.mod
```diff
@@ -0,0 +1,27 @@
...
+	github.com/go-critic/go-critic v0.0.0-20181204210945-1df300866540 => github.com/go-critic/go-critic v0.0.0-20190526074819-1df300866540
+	github.com/golangci/errcheck v0.0.0-20181003203344-ef45e06d44b6 => github.com/golangci/errcheck v0.0.0-20181223084120-ef45e06d44b6
+	github.com/golangci/go-tools v0.0.0-20180109140146-af6baa5dc196 => github.com/golangci/go-tools v0.0.0-20190318060251-af6baa5dc196
+	github.com/golangci/gofmt v0.0.0-20181105071733-0b8337e80d98 => github.com/golangci/gofmt v0.0.0-20181222123516-0b8337e80d98
+	github.com/golangci/gosec v0.0.0-20180901114220-66fb7fc33547 => github.com/golangci/gosec v0.0.0-20190211064107-66fb7fc33547
+	github.com/golangci/ineffassign v0.0.0-20180808204949-42439a7714cc => github.com/golangci/ineffassign v0.0.0-20190609212857-42439a7714cc
+	github.com/golangci/lint-1 v0.0.0-20180610141402-ee948d087217 => github.com/golangci/lint-1 v0.0.0-20190420132249-ee948d087217
+	github.com/timakin/bodyclose => github.com/golangci/bodyclose v0.0.0-20190714144026-65da19158fa2
+	mvdan.cc/unparam v0.0.0-20190124213536-fbb59629db34 => mvdan.cc/unparam v0.0.0-20190209190245-fbb59629db34
+)
```
maelk, line 27: maybe we could get rid of this now. the issue was resolved afaik. It is just that we did not revisit it on CAPM3 side.

maelk:
About the synchronization with CAPM3, I would propose to go the same way as for the IPAM controller. CAPM3 merely consumes the releases, so we would ship a CAPM3 release with the latest BMO release. In case CAPM3 requires something from BMO, we would need to simply release BMO first, then update CAPM3 to consume the latest release, and then release CAPM3. That way a specific release of CAPM3 would always be deployed with a specific release of BMO.

The disadvantage of this approach is that it might force us to do a minor release on BMO before a CAPM3 release, if something is needed from BMO that is not yet in the last release. Bug fixes in BMO would also require a new release of CAPM3. But those should not be big issues. That way BMO can be released at its own pace, and uplifted when needed in CAPM3.

docs/releasing.md
dhellmann: What builds that manifest file?

hack/markdownlint.sh
dhellmann: It's how I've seen it used elsewhere (including an earlier version of this file based on the diff).

hack/tools/go.mod
dhellmann: This lets us have different versions of dependencies for the tools than for the main code, right? That seems like something we may want to keep.

docs/releasing.md
dhellmann: It seems like CAPM3 can control that. We would want test jobs for each CAPM3 branch against the head of the branch for the series from BMO that is supported. We really only need to support one version of BMO per version of CAPM3, right? So I don't think that we would end up with a lot more jobs than we have now. The BMO is at the bottom of the dependency stack, so if we cross-test, so that changes to BMO can't break CAPM3 jobs, then we shouldn't need CAPM3 jobs using pinned versions of BMO.

docs/releasing.md
dhellmann: I'll make a note of that. I assume we will need a tagged version of BMO before adding the variables would make sense.

docs/releasing.md
maelk: I guess that would depend on how we consider BMOfrom CAPM3 point of view, if it is just a dependency updated from time to time, allowing complete split of development cycles, then we'd need separate testing for BMO, considering that CAPM3 might not support the new version / feature yet. But if we go for synced development cycles, then a version or feature change in BMO would mean a sync change in CAPM3. That's how we have been doing it until now and probably the easiest with regards to the testing, allowing for full integration tests at any point in BMO development cycles. However, going this way probably means that we need to release BMO for every CAPM3 release.

docs/releasing.md
maelk: After giving it a bit more thought, We should probably rework metal3-dev-env more extensively to allow it to deploy existing (latest?) releases (to verify nothing got broken meanwhile), tests from the master and release branches for CAPM3 and for BMO. Defining which BMO version should be used when testing CAPM3 is straightforward (though there is still the question whether it should be tested with the latest release or the tip of the release branch of BMO). To test the tip of the BMO release branches, however, we would need to clearly identify which release branch of CAPM3 is supposed to support it, for example a bug backport to an older release of BMO should not be tested with the latest master of CAPM3, but with the branch of the release that was using this version of BMO.

hack/tools/go.mod
maelk: as far as I understood it, it was only there as a workaround to get golangci-lint to work properly. now that issue was solved, we don't need it anymore. CAPI removed it : https://github.com/kubernetes-sigs/cluster-api/blob/master/hack/tools/go.mod

maelk:
/test-integration
that one should not have failed

hack/tools/go.mod
dhellmann: Oh, sorry, you were just talking about the replace directive. I thought you meant the whole file. Fixed.

docs/releasing.md
maelk: we can add a Makefile target running kustomize on the config (or deploy for now) folder

docs/releasing.md
dhellmann: I think we're saying similar things.

I envision a branch in BMO for each CAPM3 release. That BMO branch would be tested with the corresponding CAPM3 branch, so that the "series" of releases from the branches stay compatible.

We would do the same for the master branch of each component, to ensure that upcoming releases are compatible. 

Because the BMO is at the bottom of the dependency stack, there may be releases of BMO on a branch that do not correspond to a new CAPM3 release. CAPM3 could choose to consume those explicitly, or not, as long as future changes on each BMO branch continue to be compatible with the same branch of CAPM3. That doesn't really seem very different from what we're doing today, where CAPM3 always consumes the image created from the tip of the branch implicitly.

That leaves the issue of upgrade testing between branches. I think we want to reach a point of stability where we ensure compatibility between series for upgrades, and that would require another test job to install a cluster with one version, update the components, and then ensure that the upgraded cluster stabilizes and there are no errors or data loss. The design and implementation of that test job will be influenced by the release plan, but I think we should come up with the plan for that separately over in metal3-docs.

maelk: I agree. I am thinking about what would happen when there would need to be a new minor or major release of BMO with breaking change. I guess that would force us to do a similar minor/major release on CAPM3. It might be what we want , because such an API change would need to be visible for the consumer of CAPM3 that still create the BMHs. So this synchronization approach could work as long as we ensure that major changes in BMO are reflected in CAPM3, and we don't hide a breaking change of BMO in CAPM3.

For the upgrade tests, I guess it would be done anyways as we are doing the CAPM3 upgrade tests, since an upgrade of BMO would mean an upgrade of CAPM3.

So it looks like the best option for us is to keep the two release cycles somewhat in sync.

dhellmann: Yes, I think if we're going to have a breaking API change in BMO we would want to have a release of both components before that happens. Ideally we will be able to make API changes compatible by using web hooks, which would let us make the change in BMO first and then update CAPM3 to require the newer API later.

Issue #580, opened 2020-07-07
Typo in bootMACAddress causes non-obvious failures


hardys:
If someone accidentally writes `bootMacAddress` instead of `bootMACAddress` (or any other similar typo) we skip creating the ironic port for that node and silently ignore the wrong key containing the MAC.

This then leads to one of several non-obvious failure modes:

- Testing in metal3-dev-env we see the BMH stuck forever in `inspecting` state, IPA reports an error like `Failed looking up node with addresses <mac> ...`

- In downstream OpenShift testing deployment/inspection actually succeeds, but then if something causes the pod containing the ironic services to get rescheduled (such as a controlplane outage), Ironic fails on adoption due to the missing port.

I'm not yet certain why the behavior differs in each case (need to compare the ironic configs), but it seems that we should validate the BMH more strictly to avoid this?

There is also no obvious way to recover from this - modifying the yaml to correct the bootMACAddress key isn't enough, the host is still stuck inspecting (or in adopt failed), I suspect this is because we don't modify the BMC secret, so we don't re-attempt the validate path where the port can now be created.

hardys:
Thanks to @dtantsur for confirming that when using baremetal with IPMI, inspector can guess which data belongs to which node by matching IPMI addresses - so the difference with the downstream results above was that it's a baremetal env and not vbmc as when testing with metal3-dev-env

For some additional context see this downstream issue ref https://bugzilla.redhat.com/show_bug.cgi?id=1854142

hardys:
So it looks like @zaneb added validations for the BMH fields in https://github.com/metal3-io/baremetal-operator/pull/256 but that doesn't account for the case where there's a key that is present but doesn't match the schema for the resource.

I'm not sure on the preferred pattern for adding that, but it seems that we could avoid this problem if we rejected BMH resources that did not conform to the required schema (e.g contain keys we didn't expect in this case).

maelk:
Is there an equivalent of pruning for kubebuilder with Operator SDK ? By default it is enabled in kubebuilder (you can turn it off with specific annotations https://book.kubebuilder.io/reference/markers/crd-processing.html)

dhellmann:
> Is there an equivalent of pruning for kubebuilder with Operator SDK ? By default it is enabled in kubebuilder (you can turn it off with specific annotations https://book.kubebuilder.io/reference/markers/crd-processing.html)

[...]

PR #594, opened 2020-07-16
Enable Pruning for BMH CRD

hardys:
This causes stricter client-side validation against the schema
so that problems like typos in the user-provided BMH yaml can
be caught earlier and report a clear error message.

Fixes: #580

### DIFF
--- a/deploy/crds/metal3.io_baremetalhosts_crd.yaml
+++ b/deploy/crds/metal3.io_baremetalhosts_crd.yaml
@@ -42,6 +42,7 @@ spec:
     - bmhost
     singular: baremetalhost
   scope: Namespaced
+  preserveUnknownFields: false
   subresources:
     status: {}
   validation:


maelk:
would this be overwritten if we regenerate the CRDs ?

zaneb:
Manually editing this file that is generated automatically by operator-sdk isn't a sustainable solution. See #327 for more detailed discussion.
/hold

zaneb:
Should be fixed by #597
/close

hardys:
Thanks @zaneb I'd forgotten about that, additional context also in https://github.com/operator-framework/operator-sdk/issues/2886

I did try with `+kubebuilder:pruning:PreserveUnknownFields=false` in `baremetalhost_types.go` and it doesn't appear to work (`make generate` still removes the preserveUnknownFields in the yaml).

I guess one disadvantage of fixing this via a version bump is we can't backport it, but it seems like the right/only solution.

honza:
> additional context also in operator-framework/operator-sdk#2886

I'm not sure how the additional context helps us. The response provided is somewhat cryptic.

We also need the `preserveUnknownFields: false` under `spec` in order to generate explain information. 

> I did try with +kubebuilder:pruning:PreserveUnknownFields=false in baremetalhost_types.go and it doesn't appear to work (make generate still removes the preserveUnknownFields in the yaml).

Yes, same here. I tried it for both versions.

hardys:
> > additional context also in [operator-framework/operator-sdk#2886](https://github.com/operator-framework/operator-sdk/issues/2886)
> 
> I'm not sure how the additional context helps us. The response provided is somewhat cryptic.

I provided it for reference to other folks hitting the same issues, with the apparent resolution of moving to v1

PR #597, opened 2020-07-16
Switch CRD version to apiextensions.k8s.io/v1

### LINKED ISSUE #327: enable pruning in openapi settings

### LINKED ISSUE #580: Typo in bootMACAddress causes non-obvious failures

zaneb:
The existing apiextensions.k8s.io/v1beta1 version is deprecated and will
be removed in k8s 1.19. v1 is available from k8s 1.16.

Using v1, unknown fields are now pruned by default, which is what we
want.

Fixes #327
Fixes #580

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,7 @@ show_dirs:
 .PHONY: generate
 generate: bin/operator-sdk ## Run the operator-sdk code generator
 	./bin/operator-sdk generate $(VERBOSE) k8s
-	./bin/operator-sdk generate $(VERBOSE) crds
+	./bin/operator-sdk generate $(VERBOSE) crds --crd-version=v1
 	openapi-gen \
 		--input-dirs ./pkg/apis/metal3/v1alpha1 \
 		--output-package ./pkg/apis/metal3/v1alpha1 \
--- a/deploy/crds/metal3.io_baremetalhosts_crd.yaml
+++ b/deploy/crds/metal3.io_baremetalhosts_crd.yaml
@@ -1,37 +1,8 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: baremetalhosts.metal3.io
 spec:
-  additionalPrinterColumns:
-  - JSONPath: .status.operationalStatus
-    description: Operational status
-    name: Status
-    type: string
-  - JSONPath: .status.provisioning.state
-    description: Provisioning status
-    name: Provisioning Status
-    type: string
-  - JSONPath: .spec.consumerRef.name
-    description: Consumer using this host
-    name: Consumer
-    type: string
-  - JSONPath: .spec.bmc.address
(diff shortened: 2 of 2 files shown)


hardys:
Thanks @zaneb - tested locally and can confirm this does address the issues reported in #580

honza:
I was hoping the default pruning would also solve the `explain` bug but that doesn't seem to be the case. Is there some new directive we can use?

dhellmann:
Are there any implications for upgrading existing systems because of this change?

zaneb:
> I was hoping the default pruning would also solve the `explain` bug but that doesn't seem to be the case. Is there some new directive we can use?

It certainly ought to solve it. Explain is not working because it requires 'structural' schemas, and [structural schemas](https://kubernetes.io/blog/2019/06/20/crd-structural-schema/) are also required by the v1 API.

However it says [here](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#preserving-unknown-fields) that:

> CRDs converted from `apiextensions.k8s.io/v1beta1` to `apiextensions.k8s.io/v1` might lack structural schemas, and `spec.preserveUnknownFields` might be true

Not sure what they mean by 'converted' here, but maybe that is the issue?

hardys:
@zaneb @honza how should we proceed with this - should we go ahead and land this, and deal with any remaining issues related to the schema working with `explain` as a follow up?

hardys:
Ok lgtm - @dhellmann did you have specific concerns re upgrades?

dhellmann:
I didn't have any specific concerns, I just wasn't sure what would happen. I think since it's just the version of the CRD type that is changing, there is no implication for users of BareMetalHost resources, so we're OK.

/lgtm

PR #599, opened 2020-07-16
refactor redfish driver hierarchy

dhellmann:
The redfish drivers were mixed together in one file and relied
on inheritance. The point of the BMC types is that they're very
simple, so a little repetition isn't a bad thing. This PR moves the
redfish-virtualmedia and idrac-virtualmedia drivers out to their own
files and removes the inheritance to make them easier to understand.


maelk:
lgtm
/test-integration

PR #600, opened 2020-07-17
WIP: change default hardware profile to not set root device hints

dhellmann:
Change the default behavior to not set an explicit root device hint
based on the hardware profile. This relies on Ironic's default
behavior of choosing the first disk that is bigger than 4GB to hold
the image, which can often bypass issues such as recovery drives or
other devices appearing as `/dev/sda`.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -102,13 +102,20 @@ mainly, but not only, provisioning details.
 
   | **hardwareProfile** | **Root Device** |
   |---------------------|-----------------|
-  | `unknown`           | /dev/sda        |
+  | `unknown`           | *none*          |
   | `libvirt`           | /dev/vda        |
   | `dell`              | HCTL: 0:0:0:0   |
   | `dell-raid`         | HCTL: 0:2:0:0   |
   | `openstack`         | /dev/vdb        |
 
   **NOTE:** These are subject to change.
+
+  When the hardware profile is not specified, and the `unknown`
+  profile is used, no root device hints are provided to the
+  provisioning tool. In that case, the tool tries to use the first
+  drive with more than 4GB of space. The actual device selected may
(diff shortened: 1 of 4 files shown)


docs/api.md
```diff
@@ -102,7 +102,7 @@ mainly, but not only, provisioning details.
 
   | **hardwareProfile** | **Root Device** |
   |---------------------|-----------------|
-  | `unknown`           | /dev/sda        |
+  | `unknown`           | *none*          |
```
zaneb, line 105: Should we document what happens in this case?

dhellmann: I added some text below the table.

shibaPuppy:
@dhellmann 
Does ironic automatically configure root device if there is no value for root device?

dhellmann:
> @dhellmann
> Does ironic automatically configure root device if there is no value for root device?

Yes, it looks for the first device larger than 4GB and uses that.

docs/api.md
```diff
@@ -102,14 +102,21 @@ mainly, but not only, provisioning details.
...
   | `dell`              | HCTL: 0:0:0:0   |
   | `dell-raid`         | HCTL: 0:2:0:0   |
   | `openstack`         | /dev/vdb        |
 
   **NOTE:** These are subject to change.
 
+  When the hardware profile is not specified, and the `unknown`
+  profile is used, no root device hints are provided to the
+  provisioning tool. In that case, the tool tries to use the first
+  drive with more than 4GB of space. The actual device selected may
```
maelk, line 116: tiny nit : that behaviour is specific to Ironic, maybe it would be good to make that clear ?

dhellmann:
I think instead of this, we should update the API to remove the profile entirely.

PR #601, opened 2020-07-20
Add support for http_basic auth for ironic and inspector

iurygregory:
This commit adds the support for BMO to interact with ironic when
using http_basic.

requires gophercloud 0.12.0

### DIFF
--- a/cmd/get-hardware-details/main.go
+++ b/cmd/get-hardware-details/main.go
@@ -12,36 +12,22 @@ import (
 	"github.com/metal3-io/baremetal-operator/pkg/provisioner/ironic/hardwaredetails"
 )
 
-func main() {
-	var nodeID string
-	authConfig := clients.AuthConfig{
-		Type: clients.AuthType(os.Getenv("IRONIC_AUTH_STRATEGY")),
-	}
+type options struct {
+	Endpoint   string
+	AuthConfig clients.AuthConfig
+	NodeID     string
+}
 
-	switch authConfig.Type {
-	case clients.HTTPBasicAuth:
-		if len(os.Args) != 5 {
-			fmt.Println("Usage: get-hardware-details <inspector URI> <inspector User> <inspector Password> <node UUID>")
-			os.Exit(1)
-		}
-		authConfig.Username = os.Args[2]
-		authConfig.Password = os.Args[3]
-		nodeID = os.Args[4]
-	default:
-		if len(os.Args) != 3 {
-			fmt.Println("Usage: get-hardware-details <inspector URI> <node UUID>")
-			os.Exit(1)
-		}
-		nodeID = os.Args[2]
-	}
-	inspectorEndpoint := os.Args[1]
-	inspector, err := clients.InspectorClient(inspectorEndpoint, authConfig)
+func main() {
+	opts := getOptions()
+
+	inspector, err := clients.InspectorClient(opts.Endpoint, opts.AuthConfig)
 	if err != nil {
 		fmt.Printf("could not get inspector client: %s", err)
 		os.Exit(1)
 	}
 
-	introData := introspection.GetIntrospectionData(inspector, nodeID)
+	introData := introspection.GetIntrospectionData(inspector, opts.NodeID)
 	data, err := introData.Extract()
 	if err != nil {
 		fmt.Printf("could not get introspection data: %s", err)
@@ -55,4 +41,27 @@ func main() {
 	}
 
 	fmt.Println(string(json))
+}
+
+func getOptions() (o options) {
+	o.AuthConfig.Type = clients.AuthType(os.Getenv("IRONIC_AUTH_STRATEGY"))
+
+	switch o.AuthConfig.Type {
+	case clients.HTTPBasicAuth:
+		if len(os.Args) != 5 {
+			fmt.Println("Usage: get-hardware-details <inspector URI> <inspector User> <inspector Password> <node UUID>")
+			os.Exit(1)
+		}
+		o.AuthConfig.Username = os.Args[2]
+		o.AuthConfig.Password = os.Args[3]
+		o.NodeID = os.Args[4]
+	default:
+		if len(os.Args) != 3 {
+			fmt.Println("Usage: get-hardware-details <inspector URI> <node UUID>")
+			os.Exit(1)
+		}
+		o.NodeID = os.Args[2]
+	}
+	o.Endpoint = os.Args[1]
+	return
 }
--- a/cmd/get-hardware-details/main.go
+++ b/cmd/get-hardware-details/main.go
@@ -7,52 +7,38 @@ import (
 	"fmt"
 	"os"
 
-	"github.com/gophercloud/gophercloud"
-	"github.com/gophercloud/gophercloud/openstack/baremetalintrospection/httpbasic"
-	"github.com/gophercloud/gophercloud/openstack/baremetalintrospection/noauth"
 	"github.com/gophercloud/gophercloud/openstack/baremetalintrospection/v1/introspection"
+	"github.com/metal3-io/baremetal-operator/pkg/provisioner/ironic/clients"
 	"github.com/metal3-io/baremetal-operator/pkg/provisioner/ironic/hardwaredetails"
 )
 
 func main() {
-	authStrategy := os.Getenv("IRONIC_AUTH_STRATEGY")
-	if authStrategy == "http_basic" {
-		if len(os.Args) != 5 {
-			fmt.Println("Usage: get-hardware-details <inspector URI> <inspector User> <inspector Password> <node UUID>")
-			return
-		}
-	} else {
-		if len(os.Args) != 3 {
-			fmt.Println("Usage: get-hardware-details <inspector URI> <node UUID>")
-			return
-		}
+	var nodeID string
+	authConfig := clients.AuthConfig{
+		Type: clients.AuthType(os.Getenv("IRONIC_AUTH_STRATEGY")),
 	}
 
-	var inspector *gophercloud.ServiceClient
-	var nodeID string
-	if authStrategy == "http_basic" {
-		client, err := httpbasic.NewBareMetalIntrospectionHTTPBasic(httpbasic.EndpointOpts{
-			IronicInspectorEndpoint:     os.Args[1],
-			IronicInspectorUser:         os.Args[2],
-			IronicInspectorUserPassword: os.Args[3],
-		})
-		if err != nil {
-			fmt.Printf("could not get inspector client: %s", err)
+	switch authConfig.Type {
+	case clients.HTTPBasicAuth:
+		if len(os.Args) != 5 {
+			fmt.Println("Usage: get-hardware-details <inspector URI> <inspector User> <inspector Password> <node UUID>")
 			os.Exit(1)
 		}
-		inspector = client
+		authConfig.Username = os.Args[2]
+		authConfig.Password = os.Args[3]
 		nodeID = os.Args[4]
-	} else {
-		client, err := noauth.NewBareMetalIntrospectionNoAuth(
-			noauth.EndpointOpts{
-				IronicInspectorEndpoint: os.Args[1],
-			})
-		if err != nil {
-			fmt.Printf("could not get inspector client: %s", err)
+	default:
+		if len(os.Args) != 3 {
+			fmt.Println("Usage: get-hardware-details <inspector URI> <node UUID>")
 			os.Exit(1)
 		}
-		inspector = client
 		nodeID = os.Args[2]
+	}
+	inspectorEndpoint := os.Args[1]
+	inspector, err := clients.InspectorClient(inspectorEndpoint, authConfig)
+	if err != nil {
+		fmt.Printf("could not get inspector client: %s", err)
+		os.Exit(1)
 	}
 
 	introData := introspection.GetIntrospectionData(inspector, nodeID)
--- a/cmd/get-hardware-details/main.go
+++ b/cmd/get-hardware-details/main.go
@@ -44,24 +44,17 @@ func main() {
 }
 
 func getOptions() (o options) {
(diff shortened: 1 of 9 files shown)


pkg/provisioner/ironic/ironic.go
```diff
@@ -107,37 +148,70 @@ func LogStartup() {
...
 // Provisioner interface) in a consistent way for tests.
 func newProvisioner(host *metal3v1alpha1.BareMetalHost, bmcCreds bmc.Credentials, publisher provisioner.EventPublisher) (*ironicProvisioner, error) {
-	client, err := noauth.NewBareMetalNoAuth(noauth.EndpointOpts{
-		IronicEndpoint: ironicEndpoint,
-	})
-	if err != nil {
-		return nil, err
+	var provisioner ironicProvisioner
+
+	if authStrategy == "http_basic" {
```
dhellmann, line 153: There are some things repeated in this function, too. For example, setting the required micro version of the client should only be done in one place, so we don't miss one if we have to update the value.

If the only difference between the clients is the type used, could we construct the clients in the conditional and then configure them outside the conditional before building the provisioner we're going to return?

cmd/get-hardware-details/main.go
```diff
@@ -7,38 +7,72 @@ import (
...
 )
 
 func main() {
+	authStrategy := os.Getenv("")
+	if authStrategy == "http_basic" && len(os.Args) != 5 {
+		fmt.Println("Usage: get-hardware-details <inspector URI> <inspector User> <inspector Password> <node UUID>")
+		return
+
+	}
 	if len(os.Args) != 3 {
```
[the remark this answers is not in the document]
iurygregory: Yeah, I will  update the checks

cmd/get-hardware-details/main.go
```diff
@@ -7,38 +7,72 @@ import (
...
 	"os"
 
+	"github.com/gophercloud/gophercloud/openstack/baremetalintrospection/httpbasic"
 	"github.com/gophercloud/gophercloud/openstack/baremetalintrospection/noauth"
 	"github.com/gophercloud/gophercloud/openstack/baremetalintrospection/v1/introspection"
 	"github.com/metal3-io/baremetal-operator/pkg/provisioner/ironic/hardwaredetails"
 )
 
 func main() {
+	authStrategy := os.Getenv("")
```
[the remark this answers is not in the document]
iurygregory: I was sure that I've updated, but maybe I forgot to add the change.
I wasn't sure about how this was used so I just added the support for `http_basic` with the necessary parameters.
I will do some search about adding the auth as a flag and update here.

cmd/get-hardware-details/main.go
```diff
@@ -7,38 +7,72 @@ import (
...
+			os.Exit(1)
+		}
 
-	introData := introspection.GetIntrospectionData(inspector, os.Args[2])
-	data, err := introData.Extract()
-	if err != nil {
-		fmt.Printf("could not get introspection data: %s", err)
-		os.Exit(1)
-	}
+		introData := introspection.GetIntrospectionData(inspector, os.Args[4])
```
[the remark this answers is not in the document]
iurygregory: Yeah, we can even keep here using args[2] and username/password could be 3/4. 
I will see how to remove the duplication of the code here

pkg/provisioner/ironic/ironic.go
iurygregory: Will look on how to improve this, thanks!

pkg/provisioner/ironic/ironic.go
```diff
@@ -70,6 +88,29 @@ func init() {
 		fmt.Fprintf(os.Stderr, "Cannot start: No IRONIC_INSPECTOR_ENDPOINT variable set")
 		os.Exit(1)
 	}
+
+	if authStrategy == "http_basic" {
+		ironicUser = os.Getenv("IRONIC_HTTP_BASIC_USERNAME")
+		ironicPassword = os.Getenv("IRONIC_HTTP_BASIC_PASSWORD")
+		inspectorUser = os.Getenv("INSPECTOR_HTTP_BASIC_USERNAME")
+		inspectorPassword = os.Getenv("INSPECTOR_HTTP_BASIC_PASSWORD")
```
zaneb, line 96: It seems [safer](https://twitter.com/IanColdwater/status/1079408591583526913) to me to mount the secrets as a volume rather than pass them as environment variables.

pkg/provisioner/ironic/ironic.go
```diff
@@ -70,6 +88,29 @@ func init() {
 		fmt.Fprintf(os.Stderr, "Cannot start: No IRONIC_INSPECTOR_ENDPOINT variable set")
 		os.Exit(1)
 	}
+
+	if authStrategy == "http_basic" {
+		ironicUser = os.Getenv("IRONIC_HTTP_BASIC_USERNAME")
+		ironicPassword = os.Getenv("IRONIC_HTTP_BASIC_PASSWORD")
```
zaneb, line 94: I wonder if its possible to simplify this.
Do we need to have a non-hard-coded username?
Do we need to have separate passwords for Ironic and Inspector?
Can we just infer the auth strategy from the presence of the data we need?

pkg/provisioner/ironic/ironic.go
```diff
@@ -50,6 +59,15 @@ const (
 func init() {
 	// NOTE(dhellmann): Use Fprintf() to report errors instead of
 	// logging, because logging is not configured yet in init().
+	authStrategy = os.Getenv("IRONIC_AUTH_STRATEGY")
+	if authStrategy == "" {
+		fmt.Fprintf(os.Stderr, "Cannot start: No IRONIC_AUTH_STRATEGY variable set\n")
+		os.Exit(1)
+	}
```
zaneb, line 66: Making this compulsory introduces a dependency, where the deployment (and possibly config map) has to be updated for the BMO to keep working.
I'm not sure this is a good idea, but at the very least we need to update the deploy files in the repo to set it to `noauth` by default.

pkg/provisioner/ironic/ironic.go
```diff
@@ -50,6 +59,15 @@ const (
...
+	if authStrategy != "noauth" && authStrategy != "http_basic" {
```
zaneb, line 67: Minor style thing, this is an obvious candidate for a switch statement:

```
switch authStrategy {
    case "noauth":
    case "http_basic":
        // Read other env vars
    case "":
        fmt.Fprintf(...)
        os.Exit(1)
    default:
        fmt.Fprintf(...)
        os.Exit(1)
}
```

pkg/provisioner/ironic/ironic.go
iurygregory: @zaneb the file would only https://github.com/metal3-io/baremetal-operator/blob/master/deploy/default/ironic_bmo_configmap.env ?

pkg/provisioner/ironic/ironic.go
iurygregory: Sure I will update! I'm just wondering the messages I should use for  `case  ""` and `default` =)

pkg/provisioner/ironic/ironic.go
zaneb: Ah yep, looks like it is [that simple](https://github.com/metal3-io/baremetal-operator/blob/master/deploy/operator/bmo.yaml#L30-L32) here.

In the downstream MAO, we don't pass a whole config-map (which is in a different format) but [select only the keys we need](https://github.com/openshift/machine-api-operator/blob/master/pkg/operator/baremetal_pod.go#L282-L285), so it will require a simultaneous change (if it requires an addition to the config-map then I suspect there's also a change to the installer required, but I think it's probably safe not to do that since the MAO owns the topology of the deployment).

iurygregory: So downstream before we push the BMO changes we would need to update MAO to at least have noauth right?

zaneb: As it stands, yes.

dhellmann:
@zaneb has posted https://github.com/metal3-io/ironic-image/pull/182 to change the way ironic-image configures authentication. It would be nice if we could coordinate this across metal3 so we don't end up doing it different ways in different places. We've had some design documents related to auth, but I don't think we had one that covered those details.

iurygregory:
> @zaneb has posted [metal3-io/ironic-image#182](https://github.com/metal3-io/ironic-image/pull/182) to change the way ironic-image configures authentication. It would be nice if we could coordinate this across metal3 so we don't end up doing it different ways in different places. We've had some design documents related to auth, but I don't think we had one that covered those details.

@dhellmann  agree, I will look at his PR to update things here.

hardys:
/lgtm

@dhellmann can you please revisit this so we can approve unless you see any remaining issues?

pkg/provisioner/ironic/clients/auth.go
```diff
@@ -0,0 +1,95 @@
...
+
+// AuthConfig contains data needed to configure authentication in the client
+type AuthConfig struct {
+	Type     AuthType
+	Username string
+	Password string
+}
+
+func (auth *AuthConfig) load(clientType string) error {
+	authPath := path.Join("/auth", clientType)
```
dhellmann, line 30: What is this doing? Where does "/auth" come from? Maybe this PR just also needs to update documentation?

pkg/provisioner/ironic/clients/auth.go
```diff
@@ -0,0 +1,95 @@
...
+
+const (
+	// NoAuth uses no authentication
+	NoAuth AuthType = "noauth"
+	// HTTPBasicAuth uses HTTP Basic Authentication
+	HTTPBasicAuth AuthType = "http_basic"
+)
+
+// AuthConfig contains data needed to configure authentication in the client
+type AuthConfig struct {
```
dhellmann, line 23: Does this type need to be public? It doesn't look like its fields are used outside of this package.

pkg/provisioner/ironic/clients/auth.go
```diff
@@ -0,0 +1,95 @@
...
+	if _, err := os.Stat(authPath); err != nil {
+		if os.IsNotExist(err) {
+			auth.Type = NoAuth
+			return nil
+		}
+		return err
+	}
+	auth.Type = HTTPBasicAuth
+
+	username, err := ioutil.ReadFile(filepath.Clean(path.Join(authPath, "username")))
```
dhellmann, line 41: Do we need to strip newlines and carriage returns at the end of the file?

pkg/provisioner/ironic/ironic.go
```diff
@@ -50,6 +51,14 @@ const (
 func init() {
 	// NOTE(dhellmann): Use Fprintf() to report errors instead of
 	// logging, because logging is not configured yet in init().
+
+	var authErr error
+	ironicAuth, inspectorAuth, authErr = clients.LoadAuth()
```
dhellmann, line 56: At some point around here, or inside the clients package, we should log the auth type being used, to help with debugging later.

pkg/provisioner/ironic/clients/auth.go
```diff
@@ -0,0 +1,95 @@
...
+)
+
+// AuthConfig contains data needed to configure authentication in the client
+type AuthConfig struct {
+	Type     AuthType
+	Username string
+	Password string
+}
+
+func (auth *AuthConfig) load(clientType string) error {
```
dhellmann, line 29: Making this a method of AuthConfig instead of a factory feels odd. This function could return an error and the user would have a partially configured AuthConfig instance that they could try to use.

pkg/provisioner/ironic/clients/auth.go
```diff
@@ -0,0 +1,95 @@
...
+	username, err := ioutil.ReadFile(filepath.Clean(path.Join(authPath, "username")))
+	if err != nil {
+		return err
+	}
+	if len(username) == 0 {
+		return fmt.Errorf("Empty HTTP Basic Auth username")
+	}
+	auth.Username = string(username)
+
+	password, err := ioutil.ReadFile(filepath.Clean(path.Join(authPath, "password")))
```
dhellmann, line 50: Same question about newlines.

pkg/provisioner/ironic/clients/auth.go
zaneb: It's used in get-hardware-details.

pkg/provisioner/ironic/clients/auth.go
zaneb: Yep, that would be smart.

pkg/provisioner/ironic/ironic.go
zaneb: No logging in `init()` :disappointed:
I really hate using `init()` we should refactor to add an initialisation method to the provisioner interface. But not now ;)

dhellmann: Could we log in `main()`?

pkg/provisioner/ironic/clients/auth.go
zaneb: It's private, but it is a bit weird. I'll change it.

pkg/provisioner/ironic/clients/auth.go
dhellmann: The type is, but not its fields, right? It could be an interface or the fields could be private.

dhellmann:
/approve

This looks OK. I may try to hide the implementation details of the `AuthType` in a separate PR.

zaneb:
I tested this locally and it works.

pkg/provisioner/ironic/clients/auth.go
zaneb: Oh yeah - in earlier commits it used the fields directly, but after 65f3c5fd86bc01d7376bf0fa81b4d3e0365693d3 you're right, they could be private.

PR #605, opened 2020-07-27
add -boot-mode argument to make-bm-worker

dhellmann:
metal3-dev-env uses this program to create host resources. In order to
let metal3-dev-env change the boot mode for hosts, we need to add an
option to pass the value through to the template.

### DIFF
--- a/cmd/make-bm-worker/main.go
+++ b/cmd/make-bm-worker/main.go
@@ -16,6 +16,7 @@ func main() {
 	var disableCertificateVerification = flag.Bool("disableCertificateVerification", false, "will skip certificate validation when true")
 	var hardwareProfile = flag.String("hardwareprofile", "", "hardwareProfile to be used")
 	var macAddress = flag.String("boot-mac", "", "boot-mac for bootMACAddress")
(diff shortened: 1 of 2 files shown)


maelk:
This failed because the boot mode is not specified in the command line :
```
go run /home/****/tested_repo/cmd/make-bm-worker/main.go -address ipmi://192.168.111.1:6230 -password password -user admin -boot-mac 00:6a:7d:7c:4c:98 node-0
Invalid boot mode "", use "UEFI" or "legacy"
exit status 1
```
Should it default to legacy ?

dhellmann:
> This failed because the boot mode is not specified in the command line :
> 
> ```
> go run /home/****/tested_repo/cmd/make-bm-worker/main.go -address ipmi://192.168.111.1:6230 -password password -user admin -boot-mac 00:6a:7d:7c:4c:98 node-0
> Invalid boot mode "", use "UEFI" or "legacy"
> exit status 1
> ```
> 
> Should it default to legacy ?

I assumed the variable would be nil if not specified on the command line. I've fixed the test.

We want the default to be empty here, to allow the baremetal-operator to control the default. The plan there is to use UEFI as the default, which is why metal3-io/metal3-dev-env#395 sets the value to legacy explicitly so our CI will work.

PR #606, opened 2020-07-28
Add support for build variables

andfasano:
This PR adds support for OpenShift build variables in order to set the build version and the correct source git commit reference

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,8 +8,9 @@ PACKAGES = $(foreach dir,$(CODE_DIRS),$(dir)/...)
 COVER_PROFILE = cover.out
 
 # See pkg/version.go for details
-GIT_COMMIT="$(shell git rev-parse --verify 'HEAD^{commit}')"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -8,8 +8,9 @@ PACKAGES = $(foreach dir,$(CODE_DIRS),$(dir)/...)
 COVER_PROFILE = cover.out
 
 # See pkg/version.go for details
-GIT_COMMIT="$(shell git rev-parse --verify 'HEAD^{commit}')"
-export LDFLAGS="-X github.com/metal3-io/baremetal-operator/pkg/version.Raw=$(shell git describe --always --abbrev=40 --dirty) -X github.com/metal3-io/baremetal-operator/pkg/version.Commit=${GIT_COMMIT}"
+SOURCE_GIT_COMMIT ?= $(shell git rev-parse --verify 'HEAD^{commit}')
```
dhellmann, line 11: Is the important change here the names of the variables?

andfasano: Yes, those are the names of the env vars provided by the OpenShift build environment. In this way, if they are not defined, the assigned values fall back to the default case as before.

dhellmann: OK, could you include that information in the commit message? I would hate for someone to come along later and decide they don't like these names and change them.

andfasano: Done

PR #607, opened 2020-07-29
Inspection API implementation

fmuyassarov:
[Inspection API](https://github.com/metal3-io/metal3-docs/blob/master/design/baremetal-operator/re-introspection-interface.md) implementation.


fmuyassarov:
/hold 
Need to add unit tests.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -525,6 +542,21 @@ func (r *ReconcileBareMetalHost) actionInspecting(prov provisioner.Provisioner,
...
+	// Delete inspect annotation if exists
+	if hasInspectAnnotation(info.host) {
+		if info.host.Status.Provisioning.State == metal3v1alpha1.StateInspecting {
+			clearInspectAnnotations(info.host)
+			if err := r.client.Update(context.TODO(), info.host); err != nil {
+				return actionError{errors.Wrap(err, "failed to remove inspect annotations from host")}
+			}
+			return actionContinueNoWrite{}
+		} else {
+			info.host.Status.Provisioning.State = metal3v1alpha1.StateInspecting
```
zaneb, line 554: State changes should only happen in the state machine code.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -796,7 +828,9 @@ func (r *ReconcileBareMetalHost) actionManageReady(prov provisioner.Provisioner,
 		info.host.ClearError()
 		return actionContinue{provResult.RequeueAfter}
 	}
-
+	if hasInspectAnnotation(info.host) {
+		return r.actionInspecting(prov, info)
+	}
```
zaneb, line 833: No, we need to switch the state not just start executing another state. We should do that in the state machine's `handleReady()` method.

pkg/provisioner/ironic/ironic.go
```diff
@@ -508,23 +508,30 @@ func (p *ironicProvisioner) InspectHardware() (result provisioner.Result, detail
...
-				}
-				return
+				inspect = true
 			}
+		} else {
+			err = errors.Wrap(err, "failed to extract hardware inspection status")
+			return
+		}
+	}
+	if reinstrospect && !(nodes.ProvisionState(ironicNode.ProvisionState) == nodes.Inspecting || nodes.ProvisionState(ironicNode.ProvisionState) == nodes.InspectWait) {
```
zaneb, line 521: Typo in 'reintrospect' and also this is not defined with any spelling

A switch on the ProvisionState might be cleaner.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -516,7 +533,7 @@ func (r *ReconcileBareMetalHost) actionRegistering(prov provisioner.Provisioner,
 func (r *ReconcileBareMetalHost) actionInspecting(prov provisioner.Provisioner, info *reconcileInfo) actionResult {
 	info.log.Info("inspecting hardware")
 
-	provResult, details, err := prov.InspectHardware()
+	provResult, details, err := prov.InspectHardware(hasInspectAnnotation(info.host))
```
zaneb, line 536: I don't love that we're relying on the annotation still being present to control behaviour (i.e. whether we always start a new inspection or accept any existing data that happens to still be in the ironic database).

pkg/provisioner/ironic/ironic.go
```diff
@@ -508,23 +508,30 @@ func (p *ironicProvisioner) InspectHardware() (result provisioner.Result, detail
...
+		if success {
+			p.publisher("InspectionStarted", "Hardware inspection started")
 		}
-		err = errors.Wrap(err, "failed to extract hardware inspection status")
 		return
 	}
-	if !status.Finished {
+	if !status.Finished || (nodes.ProvisionState(ironicNode.ProvisionState) == nodes.Inspecting || nodes.ProvisionState(ironicNode.ProvisionState) == nodes.InspectWait) {
+		details = nil
 		p.log.Info("inspection in progress", "started_at", status.StartedAt)
```
zaneb, line 535: As an alternative to the inspect flag... we know the time at which the inspection started (and, from memory, ended if it is complete?), and we also know the time at which we entered the Inspecting state (from Host.Status.OperationHistory.Inspect.Start). So we could assume that if any existing inspection was done before we entered the inspecting state, then we need to start a new inspection.

pkg/provisioner/ironic/ironic.go
```diff
@@ -508,23 +508,30 @@ func (p *ironicProvisioner) InspectHardware() (result provisioner.Result, detail
...
+		)
+		if success {
+			p.publisher("InspectionStarted", "Hardware inspection started")
 		}
-		err = errors.Wrap(err, "failed to extract hardware inspection status")
 		return
 	}
-	if !status.Finished {
+	if !status.Finished || (nodes.ProvisionState(ironicNode.ProvisionState) == nodes.Inspecting || nodes.ProvisionState(ironicNode.ProvisionState) == nodes.InspectWait) {
+		details = nil
```
zaneb, line 534: This is the zero value, it defaults to that anyway.

pkg/provisioner/ironic/ironic.go
```diff
@@ -508,23 +508,30 @@ func (p *ironicProvisioner) InspectHardware() (result provisioner.Result, detail
...
+			nodes.ProvisionStateOpts{Target: nodes.TargetInspect},
+		)
+		if success {
+			p.publisher("InspectionStarted", "Hardware inspection started")
 		}
-		err = errors.Wrap(err, "failed to extract hardware inspection status")
 		return
 	}
-	if !status.Finished {
+	if !status.Finished || (nodes.ProvisionState(ironicNode.ProvisionState) == nodes.Inspecting || nodes.ProvisionState(ironicNode.ProvisionState) == nodes.InspectWait) {
```
zaneb, line 533: In the past we haven't bothered to wait until the node changes state in Ironic if the data is available. Why do you think it is necessary?

pkg/provisioner/ironic/ironic.go
```diff
@@ -508,23 +508,30 @@ func (p *ironicProvisioner) InspectHardware() (result provisioner.Result, detail
 				result.RequeueAfter = introspectionRequeueDelay
 				err = nil
 				return
+			case nodes.Manageable:
```
zaneb, line 511: Is there a reason for mentioning this explicitly?

pkg/provisioner/ironic/ironic.go
dtantsur: Pardon if I misunderstand the goal of the patch, but if we allow re-introspection, the data will be available from the beginning (the old data)

dhellmann: It sounds like we should tell inspector to delete the old data first? Or is waiting for the state change good enough? If we aren't doing that now, it's likely a bug in the original implementation anyway.

dtantsur: I don't think there is a call to remove old data (you could argue we should remove it when starting inspection.. maybe we should). Waiting for the state change will do the trick though.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -424,6 +436,11 @@ func clearRebootAnnotations(host *metal3v1alpha1.BareMetalHost) (dirty bool) {
 	return
 }
 
+// clearInspectAnnotations deletes inspect annotation exist on the provided host
+func clearInspectAnnotations(host *metal3v1alpha1.BareMetalHost) {
```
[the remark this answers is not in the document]
fmuyassarov: Fixed, thanks.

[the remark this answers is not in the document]
fmuyassarov: Fixed, thanks.

pkg/provisioner/ironic/ironic.go
fmuyassarov: Fixed, thanks.

fmuyassarov (commented):
@zaneb Thanks a lot for the review and sorry for addressing the comments late.

pkg/controller/baremetalhost/baremetalhost_controller.go
fmuyassarov: Fixed, thanks.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -525,6 +542,21 @@ func (r *ReconcileBareMetalHost) actionInspecting(prov provisioner.Provisioner,
...
+			clearInspectAnnotations(info.host)
+			if err := r.client.Update(context.TODO(), info.host); err != nil {
+				return actionError{errors.Wrap(err, "failed to remove inspect annotations from host")}
+			}
+			return actionContinueNoWrite{}
+		} else {
+			info.host.Status.Provisioning.State = metal3v1alpha1.StateInspecting
+			return actionContinue{}
+		}
+		provResult.Dirty = true
```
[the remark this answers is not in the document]
fmuyassarov: Fixed.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -525,6 +542,21 @@ func (r *ReconcileBareMetalHost) actionInspecting(prov provisioner.Provisioner,
 		return recordActionFailure(info, metal3v1alpha1.InspectionError, provResult.ErrorMessage)
 	}
 
+	// Delete inspect annotation if exists
+	if hasInspectAnnotation(info.host) {
+		if info.host.Status.Provisioning.State == metal3v1alpha1.StateInspecting {
```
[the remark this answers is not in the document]
fmuyassarov: Fixed. Thanks

pkg/controller/baremetalhost/baremetalhost_controller.go
fmuyassarov: I'm not 100% sure if that's OK to create a new type (`actionInspection`), or if the name of the type is not confusing/overlapping with `actionInspecting`.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -536,6 +553,15 @@ func (r *ReconcileBareMetalHost) actionInspecting(prov provisioner.Provisioner,
 		return recordActionFailure(info, metal3v1alpha1.InspectionError, provResult.ErrorMessage)
 	}
 
+	// Delete inspect annotation if exists
+	if hasInspectAnnotation(info.host) {
+		clearInspectAnnotation(info.host)
+		if err := r.client.Update(context.TODO(), info.host); err != nil {
+			return actionError{errors.Wrap(err, "failed to remove inspect annotations from host")}
+		}
```
zaneb, line 561: Why do we need to write this immediately? It seems like it could wait until the final write at the end of the reconciliation just like ~everything else?

pkg/provisioner/ironic/ironic.go
```diff
@@ -512,7 +512,7 @@ func (p *ironicProvisioner) changeNodeProvisionState(ironicNode *nodes.Node, opt
 // details of devices discovered on the hardware. It may be called
 // multiple times, and should return true for its dirty flag until the
 // inspection is completed.
-func (p *ironicProvisioner) InspectHardware() (result provisioner.Result, details *metal3v1alpha1.HardwareDetails, err error) {
+func (p *ironicProvisioner) InspectHardware(inspect bool) (result provisioner.Result, details *metal3v1alpha1.HardwareDetails, err error) {
```
zaneb, line 515: Can we give this parameter a more meaningful name like `forceNew`?

dhellmann:

/hold

pkg/provisioner/ironic/ironic.go
fmuyassarov: Fixed, thanks

pkg/provisioner/ironic/ironic.go
```diff
@@ -534,23 +534,29 @@ func (p *ironicProvisioner) InspectHardware() (result provisioner.Result, detail
...
-				}
-				return
+				inspect = true
 			}
+		} else {
+			err = errors.Wrap(err, "failed to extract hardware inspection status")
+			return
+		}
+	}
+	if inspect && !(nodes.ProvisionState(ironicNode.ProvisionState) == nodes.Inspecting || nodes.ProvisionState(ironicNode.ProvisionState) == nodes.InspectWait) {
```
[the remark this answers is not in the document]
fmuyassarov: Added it

hardys:
@fmuyassarov what's the status of this PR?

I'm proposing a new disable-inspection API based on this one as part of https://github.com/metal3-io/metal3-docs/pull/155 - so I'd like to understand if I should base it on these changes or just enable the `inspect.metal3.io/disable` interface independently

fmuyassarov:
> @fmuyassarov what's the status of this PR?
> 
> I'm proposing a new disable-inspection API based on this one as part of [metal3-io/metal3-docs#155](https://github.com/metal3-io/metal3-docs/pull/155) - so I'd like to understand if I should base it on these changes or just enable the `inspect.metal3.io/disable` interface independently

Hi @hardys ,

Unfortunately I haven't rebased my changes after migrating to Kubebuilder. So, I would say you don't need to take my changes right now. I will try to finalize this PR by the end of this week. If your changes are merged by then, I'll try to rebase based on your API and put my changes on top.

pkg/controller/baremetalhost/baremetalhost_controller.go
fmuyassarov: fixed it

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -51,6 +51,7 @@ const (
 	unmanagedRetryDelay           = time.Minute * 10
 	provisionerNotReadyRetryDelay = time.Second * 30
 	rebootAnnotationPrefix        = "reboot.metal3.io"
+	inspectAnnotation             = "inspect.metal3.io"
```
fmuyassarov, line 54: @hardys since you are also introducing the inspect annotation, I will rebase once https://github.com/metal3-io/baremetal-operator/pull/787 is merged.

fmuyassarov (commented):
Hi @zaneb , 
I have done the rebase now (after kubebuilder migration), and I think I addressed most of the comments. Would you mind please to review this once more?
/cc @hardys

controllers/metal3.io/baremetalhost_controller.go
fmuyassarov: I have rebased my changes based on #787

controllers/metal3.io/host_state_machine.go
```diff
@@ -301,7 +301,8 @@ func (hsm *hostStateMachine) handleRegistering(info *reconcileInfo) actionResult
 }
 
 func (hsm *hostStateMachine) handleInspecting(info *reconcileInfo) actionResult {
-	actResult := hsm.Reconciler.actionInspecting(hsm.Provisioner, info)
+	forceNew := hasInspectAnnotation(hsm.Host)
```
andfasano, line 304: See below my comments on `forceNew`

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -375,6 +375,25 @@ func inspectionDisabled(host *metal3v1alpha1.BareMetalHost) bool {
 	return false
 }
 
+// clearInspectAnnotation deletes inspect.metal3.io annotation exist on the provided host
+func clearInspectAnnotation(host *metal3v1alpha1.BareMetalHost) (dirty bool) {
+	delete(host.Annotations, inspectAnnotationPrefix)
+	dirty = true
```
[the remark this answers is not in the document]
fmuyassarov: Actually no. Will do the annotation deletion directly, instead of creating `clearInspectAnnotation()`

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -509,17 +528,25 @@ func (r *BareMetalHostReconciler) registerHost(prov provisioner.Provisioner, inf
 }
 
 // Ensure we have the information about the hardware on the host.
-func (r *BareMetalHostReconciler) actionInspecting(prov provisioner.Provisioner, info *reconcileInfo) actionResult {
+func (r *BareMetalHostReconciler) actionInspecting(prov provisioner.Provisioner, info *reconcileInfo, forceNew bool) actionResult {
+	info.log.Info("inspecting hardware")
```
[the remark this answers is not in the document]
fmuyassarov: Okay, removing it

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -509,17 +528,25 @@ func (r *BareMetalHostReconciler) registerHost(prov provisioner.Provisioner, inf
...
 
-	info.log.Info("inspecting hardware")
+	// Delete inspect annotation if exists
+	if hasInspectAnnotation(info.host) {
+		if clearInspectAnnotation(info.host) {
+			if err := r.Update(context.TODO(), info.host); err != nil {
+				return actionError{errors.Wrap(err, "failed to remove inspect annotation from host")}
+			}
+		}
+	}
```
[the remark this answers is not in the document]
fmuyassarov: +1 for removing `clearInspectAnnotation`.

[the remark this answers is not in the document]
fmuyassarov: > The [original proposal](https://github.com/metal3-io/metal3-docs/blob/master/design/baremetal-operator/re-introspection-interface.md) states that the annotation will be removed by the controller once the inspection is completed. Not sure if I'm missing anything, but wouldn't be better to postpone the annotation cleanup _after_ a (succesfull) `InspectHardware` (also to retry in case of inspection error)?
> 
That's right. The reason I had to delete the annotation before calling [prov.InspectHardware ](https://github.com/metal3-io/baremetal-operator/blob/7b8f74a2f2e9d7c6d9fc08fd693d00997a229e73/controllers/metal3.io/baremetalhost_controller.go#L522) is because otherwise `forceNew` would always get `true`,  since the annotation is not deleted, so the host will be stuck in the re-inspection loop.

controllers/metal3.io/host_state_machine.go
```diff
@@ -351,6 +352,11 @@ func (hsm *hostStateMachine) handleReady(info *reconcileInfo) actionResult {
 		return actionComplete{}
 	}
 
+	if hasInspectAnnotation(hsm.Host) {
+		hsm.NextState = metal3v1alpha1.StateInspecting
```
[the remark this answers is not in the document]
fmuyassarov: Sure, will update it. Thanks

pkg/provisioner/ironic/ironic.go
```diff
@@ -629,7 +629,7 @@ func (p *ironicProvisioner) changeNodeProvisionState(ironicNode *nodes.Node, opt
 // details of devices discovered on the hardware. It may be called
 // multiple times, and should return true for its dirty flag until the
 // inspection is completed.
-func (p *ironicProvisioner) InspectHardware(force bool) (result provisioner.Result, details *metal3v1alpha1.HardwareDetails, err error) {
+func (p *ironicProvisioner) InspectHardware(force, forceNew bool) (result provisioner.Result, details *metal3v1alpha1.HardwareDetails, err error) {
```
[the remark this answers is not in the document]
fmuyassarov: I wanted to use `force` parameter, but I thought I would break the existing functionally. But if you think it is okay to use it, then I will change the code to rely on it, instead of adding `forceNew`

fmuyassarov:
@andfasano thanks for the review. I have addressed some of the comments, and answered the rest. PTAL

pkg/provisioner/ironic/ironic.go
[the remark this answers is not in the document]
andfasano: I think it's a possibility worth to be explored to avoid increasing the complexity of the function signature (and moreover the semantic). Enriching unit tests on this portion of code could help to ensure not to break the existing functionality

controllers/metal3.io/baremetalhost_controller.go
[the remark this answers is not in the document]
andfasano: I think that removing the `forceNew` will also simplify the code path, as the condition could be calculated directly here. In addition, removing the annotation in case of error could lead to unexpected results since the inspection will not be repeated

controllers/metal3.io/baremetalhost_controller.go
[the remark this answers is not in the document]
andfasano: Sorry, I was referring just to the return value, since it seems superfluous.

controllers/metal3.io/baremetalhost_controller.go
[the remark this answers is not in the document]
fmuyassarov: Hi @andfasano . I have removed `forceNew` and tried to rely on existing `force`.
Calling `clearInspectAnnotation` just before
```
	clearError(info.host)
	info.host.Status.HardwareDetails = details
	return actionComplete{}
```
is causing endless inspection loop. Because the annotation isn't removed on time. Would you mind please to point out where would be the best to call it if not the current place where I added?

pkg/provisioner/ironic/ironic.go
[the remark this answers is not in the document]
fmuyassarov: Removed `forceNew`.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -445,6 +445,25 @@ func inspectionDisabled(host *metal3v1alpha1.BareMetalHost) bool {
...
+	dirty = true
+	return dirty
+}
+
+// hasInspectAnnotation checks for existence of inspect.metal3.io annotation
+// and returns true if it exist
+func hasInspectAnnotation(host *metal3v1alpha1.BareMetalHost) bool {
+	annotations := host.GetAnnotations()
+	if annotations != nil {
+		if _, ok := annotations[inspectAnnotationPrefix]; ok {
```
dtantsur, line 460: Judging by line 442, the annotation can be `disabled`, in which case we don't want to trigger inspection?

controllers/metal3.io/baremetalhost_controller.go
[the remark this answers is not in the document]
dtantsur: I think you need to move it before line 628, otherwise you're exiting on line 633.

pkg/provisioner/ironic/ironic.go
```diff
@@ -709,7 +709,7 @@ func (p *ironicProvisioner) InspectHardware(force bool) (result provisioner.Resu
 		result, err = operationFailed(status.Error)
 		return
 	}
-	if !status.Finished || nodes.ProvisionState(ironicNode.ProvisionState) == nodes.InspectWait {
+	if !status.Finished || (nodes.ProvisionState(ironicNode.ProvisionState) == nodes.Inspecting || nodes.ProvisionState(ironicNode.ProvisionState) == nodes.InspectWait) {
```
dtantsur, line 712: I wonder if this change should be done separately since (in openshift world we may want it backported since it seems to fix a bug).

dtantsur (changes requested):

controllers/metal3.io/baremetalhost_controller.go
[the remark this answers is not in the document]
andfasano: > Hi @andfasano . I have removed `forceNew` and tried to rely on existing `force`.

Thanks @fmuyassarov but it looks like that collapsing the semantic of the two cases doesn't work properly. The "old" force flag was meant to address just the case where the node reached the `nodes.InspectFail` provisioning state, while the current case it's wider. Given that, it sounds probably a better idea to keep the two features separated for the sake of clarity (even though with a more complex `InspectHardware` signature). I'd like also to have a comment from @zaneb on this point

controllers/metal3.io/baremetalhost_controller.go
fmuyassarov: Good catch! Updating it Thanks.

pkg/provisioner/ironic/ironic.go
fmuyassarov: I can open a separate PR for that. Isn't BMO in Openshift in sync with upstream?

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -579,17 +598,25 @@ func (r *BareMetalHostReconciler) registerHost(prov provisioner.Provisioner, inf
...
+	if hasInspectAnnotation(info.host) {
+		if clearInspectAnnotation(info.host) {
+			if err := r.Update(context.TODO(), info.host); err != nil {
+				return actionError{errors.Wrap(err, "failed to remove inspect annotation from host")}
+			}
+		}
+	}
 
-	provResult, details, err := prov.InspectHardware(info.host.Status.ErrorType == metal3v1alpha1.InspectionError)
+	provResult, details, err := prov.InspectHardware(force)
```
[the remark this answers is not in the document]
fmuyassarov: Right, this is what was trying to avoid initially and why I used `forceNew`.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -445,6 +445,25 @@ func inspectionDisabled(host *metal3v1alpha1.BareMetalHost) bool {
 	return false
 }
 
+// clearInspectAnnotation deletes inspect.metal3.io annotation exist on the provided host
+func clearInspectAnnotation(host *metal3v1alpha1.BareMetalHost) (dirty bool) {
+	delete(host.Annotations, inspectAnnotationPrefix)
+	dirty = true
+	return dirty
+}
```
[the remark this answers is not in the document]
fmuyassarov: Got it, thanks. I think keeping a separate function just to delete the annotation ins't really necessary. I rather delete it in place.

controllers/metal3.io/baremetalhost_controller.go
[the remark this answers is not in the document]
fmuyassarov: Thank you @dtantsur , @andfasano . Updated the deletion.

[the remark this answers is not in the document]
fmuyassarov: > Agree with @dtantsur , move it just before:
> 
> `if provResult.Dirty || details == nil {`

Thank you @dtantsur , @andfasano . Updated.

controllers/metal3.io/baremetalhost_controller.go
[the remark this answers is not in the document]
andfasano: But the current `force` flag is just based on annotation, so the triggering conditions are not the same of before. Looks like another relevant indication to keep the two things separated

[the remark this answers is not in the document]
fmuyassarov: Yes sure. @zaneb do you think it is okay to create another flag, `forceNew` that can server re-inspection only? Because, relying on `force` would break the existing functionality as @andfasano mentioned [here](https://github.com/metal3-io/baremetal-operator/pull/607#discussion_r595418581)

pkg/controller/baremetalhost/baremetalhost_controller.go
fmuyassarov: Fixed.

pkg/provisioner/ironic/ironic.go
fmuyassarov: Fixed it in https://github.com/metal3-io/baremetal-operator/pull/822, will remove it from here.

controllers/metal3.io/baremetalhost_controller.go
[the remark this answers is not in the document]
s3rj1k: @fmuyassarov Very similar issue that I had in https://github.com/metal3-io/baremetal-operator/pull/817/commits/3e860bf5815ca7b0df41fd046fdc328d160034be (PR https://github.com/metal3-io/baremetal-operator/pull/817)
cc: @zaneb @andfasano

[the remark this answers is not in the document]
zaneb: You would at a minimum have to OR the two conditions together.

pkg/provisioner/ironic/ironic.go
```diff
@@ -651,8 +651,8 @@ func (p *ironicProvisioner) InspectHardware(force bool) (result provisioner.Resu
 	}
 
 	status, err := introspection.GetIntrospectionStatus(p.inspector, ironicNode.UUID).Extract()
-	if err != nil {
-		if _, isNotFound := err.(gophercloud.ErrDefault404); isNotFound {
+	if err != nil || force {
```
zaneb, line 654: Given that the existing flag doesn't automatically do what you want, it would make more sense to me to add a new flag (`refresh` or something) rather than try to overload the meaning.

fmuyassarov: Thanks @zaneb . Will add it.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -598,6 +609,14 @@ func (r *BareMetalHostReconciler) actionInspecting(prov provisioner.Provisioner,
 		return recordActionFailure(info, metal3v1alpha1.InspectionError, provResult.ErrorMessage)
 	}
 
+	// Delete inspect annotation if exists
+	if hasInspectAnnotation(info.host) {
+		delete(info.host.Annotations, inspectAnnotationPrefix)
+		if err := r.Update(context.TODO(), info.host); err != nil {
+			return actionError{errors.Wrap(err, "failed to remove inspect annotation from host")}
+		}
```
zaneb, line 617: We might want to just return here, because if we return actionUpdate or actionComplete below the write will fail.

controllers/metal3.io/host_state_machine.go
```diff
@@ -301,7 +301,8 @@ func (hsm *hostStateMachine) handleRegistering(info *reconcileInfo) actionResult
 }
 
 func (hsm *hostStateMachine) handleInspecting(info *reconcileInfo) actionResult {
-	actResult := hsm.Reconciler.actionInspecting(hsm.Provisioner, info)
+	force := hasInspectAnnotation(hsm.Host)
```
andfasano, line 304: * Isn't missing a check on [externally provisioned hosts](https://github.com/metal3-io/metal3-docs/blob/master/design/baremetal-operator/re-introspection-interface.md#proposal)?
* Evaluation of the flag couldn't be moved directly within the `actionInspecting`?

zaneb: By definition here we're in the Inspecting state, not the ExternallyProvisioned state.
I tend to agree with you that it would make sense to do this inside actionInspecting though.

andfasano: Right I was wrong, I misread the ExternallyProvisioned -> Inspecting path in the diagram not considering that couldn't happen when `spec.ExternallyProvisioned` is set.

fmuyassarov: @zaneb @andfasano thanks a lot for your reviews. I have updated the patch now, and added refresh flag. Please take look.

controllers/metal3.io/host_state_machine.go
```diff
@@ -301,7 +301,8 @@ func (hsm *hostStateMachine) handleRegistering(info *reconcileInfo) actionResult
 }
 
 func (hsm *hostStateMachine) handleInspecting(info *reconcileInfo) actionResult {
-	actResult := hsm.Reconciler.actionInspecting(hsm.Provisioner, info)
+	refresh := hasInspectAnnotation(hsm.Host)
```
andfasano, line 304: This can be moved inside `actionInspecting`

fmuyassarov: Done.

fmuyassarov:
@andfasano @zaneb If you think https://github.com/metal3-io/baremetal-operator/pull/607#discussion_r600536374 is also addressed, can we forward and merge the PR?

controllers/metal3.io/baremetalhost_controller.go
zaneb: Yep, that's what I meant.

fmuyassarov: It is added now!

fmuyassarov:
Seems I need to rebase. Are we ok to merge this @zaneb one I do the rebase?

s3rj1k:
@fmuyassarov There is an interesting comment from @zaneb that could probably be fixed in this PR
https://github.com/metal3-io/baremetal-operator/pull/817#issuecomment-809824437
Please take a look, thanks.

fmuyassarov:
> @fmuyassarov There is an interesting comment from @zaneb that could probably be fixed in this PR
> [#817 (comment)](https://github.com/metal3-io/baremetal-operator/pull/817#issuecomment-809824437)
> Please take a look, thanks.

Thanks @s3rj1k. If I understood right, `err = nil` is something that needs to be fixed. I can definitely fix it here, but I was expecting to get this patch in if all the comments have been addressed already.

fmuyassarov:
> > @fmuyassarov There is an interesting comment from @zaneb that could probably be fixed in this PR
> > [#817 (comment)](https://github.com/metal3-io/baremetal-operator/pull/817#issuecomment-809824437)
> > Please take a look, thanks.
> 
> Thanks @s3rj1k. If I understood right, `err = nil` is something that needs to be fixed. I can definitely fix it here, but I was expecting to get this patch in if all the comments have been addressed already.

maybe we could fix it in a follow-up PR or within https://github.com/metal3-io/baremetal-operator/pull/817?

s3rj1k:
> > > @fmuyassarov There is an interesting comment from @zaneb that could probably be fixed in this PR
> > > [#817 (comment)](https://github.com/metal3-io/baremetal-operator/pull/817#issuecomment-809824437)
> > > Please take a look, thanks.
> > 
> > 
> > Thanks @s3rj1k. If I understood right, `err = nil` is something that needs to be fixed. I can definitely fix it here, but I was expecting to get this patch in if all the comments have been addressed already.
> 
> maybe we could fix it in a follow-up PR or within #817?

Yep, we need @zaneb feedback here

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -446,6 +446,18 @@ func inspectionDisabled(host *metal3v1alpha1.BareMetalHost) bool {
 	return false
 }
 
+// hasInspectAnnotation checks for existence of inspect.metal3.io annotation
+// and returns true if it exist
+func hasInspectAnnotation(host *metal3v1alpha1.BareMetalHost) bool {
+	annotations := host.GetAnnotations()
+	if annotations != nil {
+		if _, ok := annotations[inspectAnnotationPrefix]; ok && annotations[inspectAnnotationPrefix] != "disabled" {
```
zaneb, line 454: nit: this is a slightly repetitive way of writing:

```
if expect, ok := annotations[inspectAnnotationPrefix]; ok && expect != "disabled" {
```

zaneb (commented):
/approve

controllers/metal3.io/baremetalhost_controller.go
fmuyassarov: fixed, thanks

fmuyassarov:
/test-integration
@hardys @andfasano Please take a look, I think all the comments are addressed. Thank you

kashifest (commented):
/lgtm
/test-centos-integration

PR #612, opened 2020-07-31
keep boot mode updated

dhellmann:


pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -487,11 +497,8 @@ func (r *ReconcileBareMetalHost) actionRegistering(prov provisioner.Provisioner,
 		info.postSaveCallbacks = append(info.postSaveCallbacks, updatedCredentials.Inc)
 	}
 
-	// Make sure we have saved the current boot mode value.
-	bootMode := info.host.BootMode()
-	if bootMode != info.host.Status.Provisioning.BootMode {
+	if updateBootModeStatus(info.host) {
```
zaneb, line 500: Unfortunately actionRegistering can get run at any time (not just right before moving to the Inspecting state), so even with this change there's no guarantee that the mode reported in the Status is the one that was actually used to provision the host. I think this would work if we did:

```
if info.Host.NeedsHardwareInspection() && !info.Host.Spec.ExternallyProvisioned && updateBootModeStatus(info.host) {
```

but it's pretty hacky :/

We'd also need to add a call to updateBootModeStatus at the transition from ExternallyProvisioned->Inspecting.

pkg/controller/baremetalhost/baremetalhost_controller.go
```diff
@@ -853,6 +860,8 @@ func saveHostProvisioningSettings(host *metal3v1alpha1.BareMetalHost) (dirty boo
 		dirty = true
 	}
 
+	dirty = updateBootModeStatus(host)
```
zaneb, line 863: This can overwrite a change to the root device hints. Should be:

```
if updateBootModeStatus(host) {
    dirty = true
}
```

or perhaps

```
dirty |= updateBootModeStatus(host)
```

pkg/controller/baremetalhost/baremetalhost_controller.go
dhellmann: We only need to update it before inspection and provisioning, so maybe I should move this to `actionInspecting()`? Or even inside the provisioner, since that code knows whether inspection has actually started or not.

pkg/controller/baremetalhost/baremetalhost_controller.go
dhellmann: Done

pkg/controller/baremetalhost/baremetalhost_controller.go
dhellmann: The state machine code could do it, too.

dhellmann: I moved the logic to update the status field into the state machine, since there I could do it before transitioning into the new states for registering and provisioning.

zaneb:
Looks like a good solution. I wonder if we should move all of `saveHostProvisioningSettings()` into the state machine. Either way this is going to conflict with the RAID patch :(
There's another transition to StateInspecting in `handleExternallyProvisioned()` that we should handle as well.

dhellmann:
> Looks like a good solution. I wonder if we should move all of `saveHostProvisioningSettings()` into the state machine. Either way this is going to conflict with the RAID patch :(

I didn't include that change, but I could if you think that's better than a separate PR.

> There's another transition to StateInspecting in `handleExternallyProvisioned()` that we should handle as well.

Added.

kirankt:
LGTM. (don't have actual lgtm priveleges so, I approve :+1: 
/approve

pkg/provisioner/ironic/ironic.go
```diff
@@ -795,9 +793,69 @@ func (p *ironicProvisioner) getUpdateOptsForNode(ironicNode *nodes.Node) (update
...
+		// No boot_mode is set, append and return
+		value = fmt.Sprintf("%s,%s", existingCapabilities, bootModeCapabilities[bootMode])
+		return
+	}
+
+	// The capabilities value has format "var1:val1,var2:val2". We
+	// know that boot_mode is there but not what value it has.  There
+	// are only 2 values, so we can replace the "wrong" string with
+	// the right one without fully parsing the string. We may want to
+	// change this later when we have more boot modes.
```
dtantsur, line 845: I hope we're going to have less boot modes over time, not more :) On the other hand, you cannot fully rule out another capability having "uefi" in its name. Maybe search and replace the whole substring boot_mode:uefi/boot_mode:bios?

Or you can cargo-cult the whole parsing from https://opendev.org/openstack/ironic/src/branch/master/ironic/common/utils.py#L401

dtantsur (approved):

pkg/provisioner/ironic/ironic.go
dhellmann: Yes `bootModeCapabilities[fromMode]` will include `boot_mode:` and then the right mode, so we at least get the full string.

It would be safer to do the full semantic parsing of the input, but if we add another boot mode with `uefi` in the name we're going to need to touch a bunch of this stuff to update the enum anyway, so we can do that work then.

dtantsur: Ah, I didn't realize that.

elfosardo (approved):

dhellmann:
This has two approves but no LGTM. Are we ready to merge it?

zaneb:
Oh, this didn't merge yet. I feel so bad for the poor RAID folks.
/lgtm

PR #613, opened 2020-08-03
Added base64 credentials encoding

FaridHub:
The credentials have to be encoded by base64 before pasting them in YAML.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -346,7 +346,8 @@ status:
     credentialsVersion: "5562"
 ```
 
-And here is the secret `bmo-master-0-bmc-secret` holding the host's BMC credentials, base64 encoded:
+And here is the secret `bmo-master-0-bmc-secret` holding the host's 
+BMC credentials, base64 encoded:
 
 `$echo -n 'admin' | base64`
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -349,13 +349,14 @@ status:
 And here is the secret `bmo-master-0-bmc-secret` holding the host's 
 BMC credentials, base64 encoded:
 
-`$echo -n 'admin' | base64`
+```console
+$echo -n 'admin' | base64
 
-`YWRtaW4=`
+YWRtaW4=
 
-`$echo -n 'password' | base64`
+$echo -n 'password' | base64
 
-`cGFzc3dvcmQ=`
+cGFzc3dvcmQ=
 
 Copy the above base64 encoded username and password pair and 
 paste it into the yaml as mentioned below.
--- a/docs/api.md
+++ b/docs/api.md
@@ -348,6 +348,16 @@ status:
 
 And here it is the secret `bmo-master-0-bmc-secret` holding the host's BMC credentials:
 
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -348,6 +348,16 @@ status:
 
 And here it is the secret `bmo-master-0-bmc-secret` holding the host's BMC credentials:
```
dhellmann, line 349:
```suggestion
And here is the secret `bmo-master-0-bmc-secret` holding the host's BMC credentials, base64 encoded:
```

docs/api.md
```diff
@@ -346,7 +346,17 @@ status:
     credentialsVersion: "5562"
 ```
 
-And here it is the secret `bmo-master-0-bmc-secret` holding the host's BMC credentials:
+And here is the secret `bmo-master-0-bmc-secret` holding the host's BMC credentials, base64 encoded:
```
dhellmann, line 349:
```suggestion
And here is the secret `bmo-master-0-bmc-secret` holding the host's 
BMC credentials, base64 encoded:
```

docs/api.md
```diff
@@ -346,7 +346,17 @@ status:
...
+
+`$echo -n 'admin' | base64`
+
+`YWRtaW4=`
+
+`$echo -n 'password' | base64`
+
+`cGFzc3dvcmQ=`
```
dhellmann, line 357:
```suggestion
```console
$echo -n 'admin' | base64

YWRtaW4=

$echo -n 'password' | base64

cGFzc3dvcmQ=
```
```

File: docs/api.md [hunk 4]
```diff
@@ -346,7 +346,17 @@ status:
...
+
+Copy the above base64 encoded username and password pair and paste it in the yaml as mentioned below.
```
dhellmann, line 359:
```suggestion
Copy the above base64 encoded username and password pair and 
paste it into the yaml as mentioned below.
```

honza:
Needs a rebase, too

furkatgofurov7:
Also, you can run [this script](https://github.com/metal3-io/baremetal-operator/blob/master/hack/markdownlint.sh) inside the hack folder of this repository, to make sure that it will pass markdown linter job.

maelk:
That looks good, but the commits need to be squashed. You can either do it, or we could try to see how tide performs the squashing, however, that might result in a commit message not as meaningful if it just appends all the commit messages. Maybe it would be better if you squashed yourself.

FaridHub:
> That looks good, but the commits need to be squashed. You can either do it, or we could try to see how tide performs the squashing, however, that might result in a commit message not as meaningful if it just appends all the commit messages. Maybe it would be better if you squashed yourself.

How to squash?

zaneb:
> How to squash?

I find the easiest way is `git rebase -i` and then change `pick` to `squash` (or `fixup` if you don't care about the commit message) for any commits that should be squashed into the previous one. Then use the `-f` flag in your usual `git push` command to update the PR.

zaneb:
Looks like this was replaced by #616 instead of being force-pushed.

PR #615, opened 2020-08-12
Fix Keepalived container to stop cleanly

maelk:
Currently the keepalived container, when stopped does not remove the ip
address it sets as VIP. This commit fixes it so that the VIP is removed when the
container is stopped

### DIFF
--- a/resources/keepalived-docker/Dockerfile
+++ b/resources/keepalived-docker/Dockerfile
@@ -1,5 +1,6 @@
 FROM ubuntu
-copy sample.keepalived.conf /etc/keepalived/keepalived.conf
+ENV DEBIAN_FRONTEND=noninteractive
+COPY sample.keepalived.conf /etc/keepalived/keepalived.conf
 COPY  manage-keepalived.sh manage-keepalived.sh
 RUN apt update -y && apt install keepalived -y
-ENTRYPOINT /bin/bash manage-keepalived.sh && tail -f /dev/null
+ENTRYPOINT ["/bin/bash", "manage-keepalived.sh"]
--- a/resources/keepalived-docker/manage-keepalived.sh
+++ b/resources/keepalived-docker/manage-keepalived.sh
@@ -5,12 +5,4 @@ export interface=$PROVISIONING_INTERFACE
 sed  -i "s~INTERFACE~$interface~g" /etc/keepalived/keepalived.conf
 sed  -i "s~CHANGEIP~$assignedIP~g" /etc/keepalived/keepalived.conf
 
-while true;do
-        /etc/init.d/keepalived start
-	if (ip a show "$interface" |grep "inet $assignedIP"); then
-           echo "waiting for VIP to be assigned on interface $interface"
-        else
-           exit 0
-        fi
-        sleep 1
-done
+exec /usr/sbin/keepalived -n -l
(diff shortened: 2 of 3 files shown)


resources/keepalived-docker/manage-keepalived.sh
```diff
@@ -5,12 +5,4 @@ export interface=$PROVISIONING_INTERFACE
...
-while true;do
-        /etc/init.d/keepalived start
-	if (ip a show "$interface" |grep "inet $assignedIP"); then
-           echo "waiting for VIP to be assigned on interface $interface"
-        else
-           exit 0
-        fi
-        sleep 1
-done
+exec /usr/sbin/keepalived -n -l
```
kashifest, line 8: how do you verify whether the IP was assigned to endpoint or not?

maelk: you don't. If keepalived does not start properly, it will exit, then K8S will restart it

kashifest: how should we test it whether its working properly or not? have you tested it locally?

maelk: I'm modifying the CI tests right now

maelk: metal3-io/metal3-dev-env#425 and metal3-io/project-infra#137

resources/keepalived-docker/Dockerfile
```diff
@@ -1,5 +1,6 @@
 FROM ubuntu
-copy sample.keepalived.conf /etc/keepalived/keepalived.conf
+ENV DEBIAN_FRONTEND=noninteractive
+COPY sample.keepalived.conf /etc/keepalived/keepalived.conf
 COPY  manage-keepalived.sh manage-keepalived.sh
 RUN apt update -y && apt install keepalived -y
-ENTRYPOINT /bin/bash manage-keepalived.sh && tail -f /dev/null
+ENTRYPOINT exec /bin/bash manage-keepalived.sh
```
zaneb, line 6: I believe you can write this as `ENTRYPOINT ["/bin/bash", "manage-keepalived.sh"]` to avoid starting two shells.

resources/keepalived-docker/manage-keepalived.sh
zaneb: > you don't. If keepalived does not start properly, it will exit, then K8S will restart it

That sounds fine, but just to be clear... we're not expecting this to e.g. always fail when another host has the VIP are we?  I don't think so, but I don't actually understand what the previous code was trying to do. I do know that CrashLoopBackoff is not the mechanism you want to use for failover ;)

maelk: no, Keepalived will keep running,  in case the master instance fails. It should not exit. I don't recall why the loop was there in the first place, so I just removed it.

resources/keepalived-docker/Dockerfile
maelk: good point, fixed, thank you

maelk:
About using other keepalived images, I don't know why we went with our own, maybe the original authors @kashifest and @Xenwar might know ?

resources/keepalived-docker/manage-keepalived.sh
maelk: I was referring to an actual failure to start, like a misconfiguration or a temporary unavailable interface, where we actually want a failure to start to appear and not be hidden by a loop.

kashifest: The loop was there because it failed to restart itself I believe back then , it was a bug from keepalived if I recall correctly so we went on with the loop

PR #619, opened 2020-09-01
allow mock ironic/inspector service to run without specifying a port

dhellmann:
Rewrite the mock server for the ironic provisioner ready checks so it
does not need a specific port. This will
eventually make it easier to expand the server to support other tests.

### DIFF
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -122,13 +122,13 @@ func LogStartup() {
 
 // A private function to construct an ironicProvisioner (rather than a
 // Provisioner interface) in a consistent way for tests.
-func newProvisionerWithURLs(host *metal3v1alpha1.BareMetalHost, bmcCreds bmc.Credentials, publisher provisioner.EventPublisher, ironicURL, inspectorURL string) (*ironicProvisioner, error) {
-	clientIronic, err := clients.IronicClient(ironicURL, ironicAuth)
+func newProvisionerWithSettings(host *metal3v1alpha1.BareMetalHost, bmcCreds bmc.Credentials, publisher provisioner.EventPublisher, ironicURL string, ironicAuthSettings clients.AuthConfig, inspectorURL string, inspectorAuthSettings clients.AuthConfig) (*ironicProvisioner, error) {
+	clientIronic, err := clients.IronicClient(ironicURL, ironicAuthSettings)
 	if err != nil {
 		return nil, err
 	}
 
-	clientInspector, err := clients.InspectorClient(inspectorURL, inspectorAuth)
+	clientInspector, err := clients.InspectorClient(inspectorURL, inspectorAuthSettings)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


pkg/provisioner/ironic/ironic.go
```diff
@@ -122,13 +122,13 @@ func LogStartup() {
 
 // A private function to construct an ironicProvisioner (rather than a
 // Provisioner interface) in a consistent way for tests.
-func newProvisioner(host *metal3v1alpha1.BareMetalHost, bmcCreds bmc.Credentials, publisher provisioner.EventPublisher) (*ironicProvisioner, error) {
-	clientIronic, err := clients.IronicClient(ironicEndpoint, ironicAuth)
+func newProvisionerWithSettings(host *metal3v1alpha1.BareMetalHost, bmcCreds bmc.Credentials, publisher provisioner.EventPublisher, ironicURL string, ironicAuthSettings clients.AuthConfig, inspectorURL string, inspectorAuthSettings clients.AuthConfig) (*ironicProvisioner, error) {
```
andfasano, line 125: That looks really nice. What about at this point replacing the other occurrences of `newProvisioner` in `ironic_test.go` with `newProvisionerWithSettings`? It will avoid setting explicitly the `IRONIC_ENDPOINT` and `IRONIC_INSPECTOR_ENDPOINT` environment variables to allow the other tests to pass.

dhellmann: I updated the other calls. The variables still have to be set because the init function requires them. Refactoring that is going to be a bit more complicated so I would rather save that for a separate PR.

pkg/provisioner/ironic/ironic_test.go
```diff
@@ -134,8 +134,11 @@ func TestProvisionerIsReady(t *testing.T) {
 func TestGetUpdateOptsForNodeWithRootHints(t *testing.T) {
 
 	eventPublisher := func(reason, message string) {}
+	auth := clients.AuthConfig{Type: clients.NoAuth}
 
-	prov, err := newProvisioner(makeHost(), bmc.Credentials{}, eventPublisher)
+	prov, err := newProvisionerWithSettings(makeHost(), bmc.Credentials{}, eventPublisher,
+		"https://localhost", auth, "http://localhost", auth,
```
zaneb, line 140: The idea here is that these should never be called, right? Shouldn't we avoid using something real like localhost and instead specify e.g. `ironic.test`. (See [RFC 2606](https://tools.ietf.org/html/rfc2606#page-2).)

zaneb (commented):
I was a bit confused at the headline of the PR and commit messages at first; of course the server is still listening on a port and communicating over a socket, but now it uses a system-chosen port instead of requiring us to reserve ports 6385 and 5050 for this.

/approve

pkg/provisioner/ironic/ironic.go
andfasano: Right, that's fine then.

pkg/provisioner/ironic/ironic_test.go
dhellmann: Done

dhellmann:
> I was a bit confused at the headline of the PR and commit messages at first; of course the server is still listening on a port and communicating over a socket, but now it uses a system-chosen port instead of requiring us to reserve ports 6385 and 5050 for this.

I thought based on reading the description of the test server that it was using some sort of in-memory communication instead of a socket, but of course you're right because then we would need a special client, too.

andfasano:
The current implementation looks fine for me

PR #620, opened 2020-09-02
Change base image to distroless

maelk:
The metal3 base image was based on Centos 7 and very outdated.
Using distroless image as base, the size of BMO image went
down from 1.76GB to 45MB.


maelk:
I can change that too, sure. I'll also check why the CI failed

maelk:
BMO seem to require libc (or glibc ?), so I had to use the base image instead of the static one.

fmuyassarov (commented):
Looks good.

PR #622, opened 2020-09-03
reformat api documentation to make linking to field definitions easier

dhellmann:
When explaining the API to a user, it is frequently useful to be able
to refer to the documentation for a specific field. This patch
reformats the documentation to use headers for the top level fields in
the spec and status sections of the host CRD so that anchors are
generated automatically.


maelk:
lgtm
/test-integration

PR #623, opened 2020-09-03
more unit test work for ironic provisioner

dhellmann:
Continue to improve the unit test suite and mock server for the ironic provisioner.


pkg/provisioner/ironic/findhost_test.go
```diff
@@ -0,0 +1,77 @@
...
+			auth := clients.AuthConfig{Type: clients.NoAuth}
+
+			// Update the default host to match the test settings
+			host := makeHost()
+			host.ObjectMeta.Name = tc.hostName
+			host.Status.Provisioning.ID = tc.provisioningID
+
+			prov, err := newProvisionerWithSettings(host, bmc.Credentials{}, nil,
+				tc.ironic.Endpoint(), auth, "https://inspector.test/", auth,
+			)
```
andfasano, line 61: A minor point, I noticed that this setup flow (create the auths, build an host - here with some customizations) and the call `newProvisionerWithSettings` seems to be pretty common, maybe we could have some test scenario builder dedicated for that to remove this duplication

pkg/provisioner/ironic/ready_test.go
```diff
@@ -0,0 +1,144 @@
...
+
+		expectedIronicCalls    string
+		expectedInspectorCalls string
+		expectedIsReady        bool
+		expectedError          string
+	}{
+		{
+			name:                   "IsReady",
+			ironic:                 buildServer(t, "ironic", "{}", driverResponse),
+			inspector:              buildServer(t, "ironic-inspector", "{}", driverResponse),
```
andfasano, line 54: `driverResponse` here is not meaningful

pkg/provisioner/ironic/findhost_test.go
dhellmann: Yes, good point. I noticed that I was doing a lot of copying and pasting as I came up with the tests, but I was focused on the mock API server and not so much at this level.

I expect to keep working on these tests in future PRs, so I will look at ways to consolidate that further as I add more tests.

pkg/provisioner/ironic/ready_test.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/metal3-io/baremetal-operator/pkg/bmc"
+	"github.com/metal3-io/baremetal-operator/pkg/provisioner/ironic/clients"
+	"github.com/metal3-io/baremetal-operator/pkg/provisioner/ironic/testserver"
+	"github.com/stretchr/testify/assert"
+)
+
+func buildServer(t *testing.T, name string, v1 string, drivers string) *testserver.MockServer {
+	return testserver.New(t, name).Response("/v1", v1).
+		Response("/v1/drivers", drivers)
+}
```
andfasano, line 16: IMHO I preferred the previous pattern (based on Fluent Builder and Object Mother), since it was aiming to improve the readability in the test cases definition. I found a little bit misleading have a single helper method with parameters instead of a fluent interface. I think that now that we have a nice middleware helper object (`testserver`) now a fluent interface could be built easily on top of it (with maybe distinct types for distinct servers, to avoid any confusion)

pkg/provisioner/ironic/ready_test.go
dhellmann: done

pkg/provisioner/ironic/testserver/server.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+}
+
+// MockServer is a simple http testing server
+type MockServer struct {
+	t            *testing.T
+	mux          *http.ServeMux
+	name         string
+	Requests     string
+	FullRequests []*http.Request
```
andfasano, line 26: Is this field used somewhere?

pkg/provisioner/ironic/ready_test.go
dhellmann: I'm not sure exactly what you mean here. This function cut down on some rather long lines that repeated a few times in the test setup. Do you mean replace `buildServer` with something like `ironicServer(t, drivers)` and `inspectorServer(t)`? Or something else?

pkg/provisioner/ironic/testserver/server.go
dhellmann: Not yet. I'm anticipating wanting it when we have tests that try to PUT or POST data, so we can see what that data is. I could remove it for now, but it's likely to come back in some form in the next PR anyway.

pkg/provisioner/ironic/testserver/server.go
```diff
@@ -0,0 +1,104 @@
+package testserver
```
andfasano, line 1: I think it's really a good idea to have a dedicated file for a test helper, as it supports reuse in different tests. What about defining a dedicated folder to store test helpers, in case in future some others will be defined? Not sure if there's a consolidated convention for that.

pkg/provisioner/ironic/testserver/server.go
```diff
@@ -0,0 +1,104 @@
...
+	return response
+}
+
+func (m *MockServer) logRequest(r *http.Request, response string) {
+	m.t.Logf("%s: %s %s -> %s", m.name, r.Method, r.URL, response)
+	m.Requests += r.RequestURI + ";"
+	m.FullRequests = append(m.FullRequests, r)
+}
+
+func (m *MockServer) handleNoResponse(w http.ResponseWriter, r *http.Request) {
```
andfasano, line 49: On a second thought, I think we could rename it as `handleDefaultResponse`?

pkg/provisioner/ironic/testserver/server.go
dhellmann: Do you mean to move this to something like `/pkg/fixtures`?

pkg/provisioner/ironic/testserver/server.go
dhellmann: It looks like this method isn't used at all, so I removed it entirely.

pkg/provisioner/ironic/testserver/server.go
andfasano: Yes (and keeping also the convention to have the suffix `test` in the package name - ie like `httptest` - to clearly state that it's not production code)

pkg/provisioner/ironic/testserver/server.go
andfasano: That's ok then

pkg/provisioner/ironic/ready_test.go
andfasano: Added a PR to show better the idea: please see https://github.com/dhellmann/baremetal-operator/pull/1

dhellmann: I've incorporated those changes here. As we write more tests using that approach, I feel like we're going to end up with more and more types based on `MockServer`. I'm not sure how I feel about that. It seems like it could be positive, because we'll have little server types that are used only in part of the test suite. On the other hand, there will be a lot of them.

pkg/provisioner/ironic/testserver/server.go
dhellmann: I could see renaming the package, but it's only going to be used by the ironic provisioner tests so I'm not sure I see the benefit of moving it further away from that test suite.

pkg/provisioner/ironic/ready_test.go
andfasano: Thanks. My general expectation is to enrich the fluent interface - when required - of the defined mock types to add other behaviors rather than making new types.

pkg/provisioner/ironic/testserver/server.go
andfasano: Don't want to anticipate too much the changes, so until they're going to be used just here then I think it's fine to delay moving them into another general folder

andfasano:
Current changes are fine for me

dhellmann:
Rebased and squashed into fewer commits.

/test-integration

PR #625, opened 2020-09-07
update operator-sdk command

zhouhao3:
Signed-off-by: Zhou Hao <zhouhao@cn.fujitsu.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -135,14 +135,14 @@ e2e-local:
 
 .PHONY: run
 run: ## Run the operator outside of a cluster in development mode
-	operator-sdk run --local \
+	operator-sdk run local \
 		--go-ldflags=$(LDFLAGS) \
 		--watch-namespace=$(RUN_NAMESPACE) \
 		--operator-flags="-dev"
 
 .PHONY: demo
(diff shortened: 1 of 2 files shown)


dhellmann:
I think I would rather take the approach in #637 instead of this. The operator-sdk is going to go away in the next few weeks anyway.

Makefile
```diff
@@ -135,14 +135,14 @@ e2e-local:
 
 .PHONY: run
 run: ## Run the operator outside of a cluster in development mode
-	operator-sdk run --local \
+	operator-sdk run local \
```
zaneb, line 138: At a minimum we should use the vendored version of the operator-sdk `./bin/operator-sdk` so that this doesn't (a) require the dev to install their own version, and (b) sometimes break depending on what version that is.

dhellmann:
This won't be needed after we merge https://github.com/metal3-io/baremetal-operator/pull/650

dhellmann:
This is no longer needed after #655, since we do not use operator-sdk any more. Please let me know if `make run` works OK for you or if we need to make any changes.

PR #631, opened 2020-09-10
Add TLS support between Ironic and BMO

maelk:
This PR introduces support for TLS using a custom CA certificate in BMO and modifies the deployment file to be able to deploy Ironic, inspector and BMO with TLS enabled.

This requires https://github.com/metal3-io/ironic-image/pull/198 and https://github.com/metal3-io/ironic-inspector-image/pull/62

### DIFF
--- a/deploy/tls/kustomization.yaml
+++ b/deploy/tls/kustomization.yaml
@@ -1,0 +1,13 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: metal3
+resources:
+- ../default
+
+secretGenerator:
+  - name: ironic-cacert
+    files:
+    - crt=ca.crt
+
+patchesStrategicMerge:
+- tls_ca.yaml
--- a/docs/deploying.md
+++ b/docs/deploying.md
@@ -35,6 +35,9 @@ deploy/
 │   ├── role_binding.yaml
 │   ├── role.yaml
 │   └── service_account.yaml
+├── tls
+│   ├── kustomization.yaml
+│   └── tls_ca.yaml
 └── role.yaml -> rbac/role.yaml
 ```
 
@@ -66,6 +69,9 @@ namely `default`,  `ironic` and `keepalived`. `default` and `ironic` deploy
 only ironic, `keepalived` deploys the ironic with keepalived. As the name
 implies, `keepalived/keepalived_patch.yaml` patches the default image URL
 through kustomization.
+
+## Deployment commands
+
 The user should run following commands to be able to meet requirements of each
 use case as provided below:
 
@@ -89,6 +95,64 @@ kustomize build $BMOPATH/ironic-deployment/default | kubectl apply -f-
 ```
 
 where $BMOPATH points to the baremetal-operator path.
+
+### Deploying with TLS
+
+For this, you need to first copy your Ironic CA Certificate or the ironic
+certificate itself, then build the kustomization for baremetal operator:
+
+```sh
+   cp <path-to-certificate> $BMOPATH/deploy/tls/ca.crt
+   ./tools/bin/kustomize build $BMOPATH/deploy/tls | kubectl apply -f -
+```
+
+Then you need to copy all your certificates and build the kustomization for
+ironic. The Ironic CA certificate is not required
+
+```sh
+   cp <path-to-ca-certificate> $BMOPATH/ironic-deployment/tls/ironic-ca.crt
+   cp <path-to-ca-certificate> $BMOPATH/ironic-deployment/tls/ironic-inspector-ca.crt
+
+   cp <path-to-tls-certificate> $BMOPATH/ironic-deployment/tls/ironic.crt
+   cp <path-to-tls-cert-key> $BMOPATH/ironic-deployment/tls/ironic.key
+
+   cp <path-to-inspector-tls-certificate> \
+   $BMOPATH/ironic-deployment/tls/ironic-inspector.crt
+   cp <path-to-inspector-tls-cert-key> \
+   $BMOPATH/ironic-deployment/tls/ironic-inspector.key
+
+   ./tools/bin/kustomize build $BMOPATH/ironic-deployment/tls | \
+   kubectl apply -f -
+```
+
+### Deploying with Keepalived and TLS
+
+For this, you need to first copy your Ironic CA Certificate, then build the
+kustomization for baremetal operator:
+
+```sh
+   cp <path-to-ca-certificate> $BMOPATH/deploy/tls/ca.crt
+   ./tools/bin/kustomize build $BMOPATH/deploy/tls | kubectl apply -f -
+```
+
+Then you need to copy all your certificates and build the kustomization for
+ironic:
+
+```sh
+   cp <path-to-ca-certificate> $BMOPATH/ironic-deployment/keepalived-tls/ironic-ca.crt
+   cp <path-to-ca-certificate> $BMOPATH/ironic-deployment/keepalived-tls/ironic-inspector-ca.crt
+
+   cp <path-to-tls-certificate> $BMOPATH/ironic-deployment/keepalived-tls/ironic.crt
+   cp <path-to-tls-cert-key> $BMOPATH/ironic-deployment/keepalived-tls/ironic.key
+
+   cp <path-to-inspector-tls-certificate> \
+   $BMOPATH/ironic-deployment/keepalived-tls/ironic-inspector.crt
+   cp <path-to-inspector-tls-cert-key> \
+   $BMOPATH/ironic-deployment/keepalived-tls/ironic-inspector.key
+
+   ./tools/bin/kustomize build $BMOPATH/ironic-deployment/keepalived-tls | \
+   kubectl apply -f -
+```
 
 #### Useful tips
 
--- a/ironic-deployment/keepalived-tls/kustomization.yaml
+++ b/ironic-deployment/keepalived-tls/kustomization.yaml
@@ -1,0 +1,24 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: metal3
+resources:
+- ../default
+
+  secretGenerator:
+  - name: ironic-ca-cert
+    files:
+    - tls.crt=ironic-ca.crt
+  - name: ironic-cert
+    files:
+    - tls.crt=ironic.crt
+    - tls.key=ironic.key
+    type: "kubernetes.io/tls"
+  - name: ironic-inspector-cert
+    files:
+    - tls.crt=ironic-inspector.crt
+    - tls.key=ironic-inspector.key
+    type: "kubernetes.io/tls"
+
+  patchesStrategicMerge:
+  - tls.yaml
+
--- a/pkg/provisioner/ironic/clients/client.go
+++ b/pkg/provisioner/ironic/clients/client.go
@@ -2,16 +2,53 @@ package clients
 
 import (
 	"fmt"
+	"net/http"
+	"os"
+	"time"
 
 	"github.com/gophercloud/gophercloud"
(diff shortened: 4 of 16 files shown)


ironic-deployment/keepalived-tls/kustomization.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: metal3
+resources:
+- ../keepalived
+
+configMapGenerator:
+- behavior: merge
+  literals:
+  - CACERT_FILE=/tls/ca/ca.crt
```
dtantsur, line 10: Do you mean to use client certificate validation? As I mentioned on the ironic-image review, I'd skip it, at least for now. We don't have a way to provide client keys to the ramdisk (nor do you propose one here).

ironic-deployment/keepalived-tls/kustomization.yaml
```diff
@@ -0,0 +1,26 @@
...
+  name: ironic-bmo-configmap
+
+secretGenerator:
+- name: ironic-cacert
+  files:
+  - ca.crt=ca.crt
+- name: ironic-cert
+  files:
+  - tls.crt=tls.crt
+  - tls.key=tls.key
```
dtantsur, line 22: Only this is actually a secret.

pkg/provisioner/ironic/ironic.go
```diff
@@ -84,6 +85,16 @@ func init() {
 		fmt.Fprintf(os.Stderr, "Cannot start: No IRONIC_INSPECTOR_ENDPOINT variable set")
 		os.Exit(1)
 	}
+	ironicTrustedCAFile = os.Getenv("IRONIC_CACERT_FILE")
```
dtantsur, line 88: `IRONIC_CERT_FILE`. You don't need CACERT, at least if I understand your motivation right.

dtantsur (changes requested):
I think there is a confusion around CA cert vs cert. Unless you want to establish a proper chain OR use client certificates, you don't need 3 files. You only need a private key (passed to ironic as a secret) and a certificate (used everywhere else). I suggest we limit at least the initial proposal to this simplest case to avoid confusion. I'd also prefer we generate certificates rather than make an operator do it.

docs/deploying.md
```diff
@@ -90,6 +96,48 @@ kustomize build $BMOPATH/ironic-deployment/default | kubectl apply -f-
 
 where $BMOPATH points to the baremetal-operator path.
 
+### Deploying with TLS
+
+For this, you need to first copy your Ironic CA Certificate, then build the
```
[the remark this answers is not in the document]
dhellmann: Is the goal to allow a deployment script (or helm chart or whatever) to create the files?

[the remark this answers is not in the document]
maelk: In our case, we want to be able to provide the CA for the ironic deployment. Ironic itself would use the certificates generated by cert-manager, but for BMO to accept them all, we need to use the CA, not a specific certificate.
@dhellmann yes, at the end it will be cert-manager creating the certificates and keys, however, it could be given the CA to do so.

ironic-deployment/keepalived-tls/kustomization.yaml
maelk: no, the mTLS part is not supported due to issue with PHA in go. We are going for plain TLS, but we want to be able to accept any certificate genearted by cert-manager using that CA

ironic-deployment/keepalived-tls/kustomization.yaml
maelk: Indeed, Ironic itself probably does not need the CA, only BMO does

ironic-deployment/keepalived-tls/kustomization.yaml
maelk: I'll remove the CA part from Ironic configuration. it is only needed in BMO actually, considering Ironic has the certificates.

pkg/provisioner/ironic/ironic.go
maelk: We need a CA to be able to support the cert-manager certificate rotation. However,the user could also just provide the certificate itself.

maelk (commented):
Let me clarify a bit our scenario. We do not want to do client authentication, that is out of the picture, so as @dtantsur pointed out, we do not need the CA in Ironic, just the certs. However, we want to deploy ironic using cert-manager to generate and rotate the certificates used, based on a CA given by the user (or automatically generated). In that case, BMO should accept any certificate signed by that CA, and not only the current certificate, keeping in mind that there isn't any tight link between ironic and BMO anymore, they could even be deployed in different clusters. Hence, not taking the specific certificate, but the CA, allows for not having to re-deploy BMO when cert-manager rotates the ironic certificate.

maelk:
I have now modified the ironic images. we are providing the certificate and the key for ironic and inspector, and in addition the CA to validate the certificate to both of them so that they can validate each other's certificate . we provide that certificate to BMO to validate the TLS connection with ironic and inspector. @dtantsur and @dhellmann please take a look

docs/deploying.md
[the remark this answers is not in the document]
dtantsur: > we need to use the CA, not a specific certificate

This is not really correct: in the simplest case you can use the certificate. I don't disagree there is a valid case for having a CA, but I would prefer that we also support a simple one-knob configuration "get_me_tls: true" where we'll just generate a pair key-certificate and pass certificate as both CERT_FILE and CACERT_FILE.

dtantsur:
> Hence, not taking the specific certificate, but the CA, allows for not having to re-deploy BMO when cert-manager rotates the ironic certificate.

Got it, thank you for explanation.

ironic-deployment/keepalived-tls/kustomization.yaml
maelk: I added the behaviour of using the certificate as CA for ironic, so now we can remove this.

docs/deploying.md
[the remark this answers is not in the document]
maelk: ok, this is now supported, we can just generate a certificate and a key, and pass it to both ironic and BMO and get it going by passing the same file for IRONIC_CACERT_FILE in BMO.

pkg/provisioner/ironic/clients/client.go
```diff
@@ -26,11 +49,14 @@ func IronicClient(ironicEndpoint string, auth AuthConfig) (client *gophercloud.S
...
-	return
+	if err != nil {
+		return
+	}
+	return updateHTTPClient(client, caFile)
 }
 
 // InspectorClient creates a client for Ironic Inspector
-func InspectorClient(inspectorEndpoint string, auth AuthConfig) (client *gophercloud.ServiceClient, err error) {
+func InspectorClient(inspectorEndpoint string, auth AuthConfig, caFile string) (client *gophercloud.ServiceClient, err error) {
```
zaneb, line 59: Why not make this part of the AuthConfig struct?

pkg/provisioner/ironic/clients/client.go
```diff
@@ -46,5 +72,8 @@ func InspectorClient(inspectorEndpoint string, auth AuthConfig) (client *gopherc
 	default:
 		err = fmt.Errorf("Unknown auth type %s", auth.Type)
 	}
-	return
+	if err != nil {
+		return
+	}
+	return updateHTTPClient(client, caFile)
```
zaneb, line 78: I really like that this is orthogonal to whether we're using basic auth or not :+1:

pkg/provisioner/ironic/clients/client.go
maelk: good point, I will update

maelk: actually, it's handled quite differently, and the path to the CA is unrelated to the authConfig, so I think keeping it separated might be cleaner/clearer. What do you think ?

zaneb: I missed that this is just TLS, not mTLS, so it's only auth if you squint (technically we are authenticating the server!). It probably does make sense to keep it separate if:

* There are no other options likely to be added (e.g. use https but allow self-signed certs); and
* When mTLS support is added, this remains completely independent (they don't e.g. need to share a CA)

maelk: I'm playing with the idea of making it a TLSConfig. I'll modify the PR with that, please take a look.

kashifest:
lgtm

furkatgofurov7:
lgtm

deploy/tls/kustomization.yaml
```diff
@@ -0,0 +1,19 @@
...
+resources:
+- ../default
+
+configMapGenerator:
+- behavior: merge
+  literals:
+    - IRONIC_CACERT_FILE=/tls/ironic_ca.crt
+  name: ironic-bmo-configmap
+
+secretGenerator:
```
dtantsur, line 13: Not sure what this means, but a certificate is not a secret.

docs/deploying.md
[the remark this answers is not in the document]
dtantsur: Let's mention that the CA certificate is no longer required?

docs/deploying.md
```diff
@@ -90,6 +96,48 @@ kustomize build $BMOPATH/ironic-deployment/default | kubectl apply -f-
...
+   cp <path-to-ca-certificate> $BMOPATH/deploy/tls/ca.crt
+   ./tools/bin/kustomize build $BMOPATH/deploy/tls | kubectl apply -f -
+```
+
+Then you need to copy all your certificates and build the kustomization for
+ironic:
+
+```sh
+   cp <path-to-ca-certificate> $BMOPATH/ironic-deployment/tls/ca.crt
+   cp <path-to-tls-certificate> $BMOPATH/ironic-deployment/tls/tls.crt
```
dtantsur, line 114: I'm curious how you're planning for the certificate to be generated by a user if they don't know an IP address and we don't have host names.

maelk: If they don't know the ip address, they can't really do any setup, no ? How would they configure BMO ? What they could maybe do is generate a certificate for a domain name that also covers the 127.0.0.1 IP address with the SAN, give that as PROVISIONING_IP to ironic and inspector, and configure BMO with the domain (for example from k8s service). That could work. But if the user is deploying with the keepalived scenario, then they know the IP address.

deploy/tls/kustomization.yaml
maelk: That's kustomize specific, to create a K8S secret with the certificate and key.

dhellmann:

/hold

ironic-deployment/keepalived-tls/kustomization.yaml
s3rj1k: @maelk "the mTLS part is not supported due to issue with PHA in go"
Can you please be more specific about "PHA in go"?
What this issues are?

maelk: This is the exact issue :https://github.com/golang/go/issues/40521 Go does not support Post Handshake Authentication unfortunately, and our solution design was relying on PHA to separate the IPA endpoints that should not be authenticated from the rest : https://github.com/metal3-io/metal3-docs/blob/master/design/ironic_authentication.md

s3rj1k: @maelk How about just using different DNS records that resolve to same IP?

maelk: We are not using dns, but directly IP addresses. We cannot easily use different endpoints, so we are waiting for the ironic mTLS feature on IPA, that is WIP in Ironic.

maelk:
We will address the change to the config folder, when we work on the metal3-dev-env adaptation. We currently do not have ways to test it in the CI otherwise.

PR #635, opened 2020-09-15
ensure the boot mode is set in ironic before starting inspection

dhellmann:
We were previously only updating the host status copy of the boot mode
when we entered the inspecting state or provisioning states. This change
extends that to set it correctly during the registration process, too,
so that when the node is created in ironic it is given the right mode.

We were only updating the boot mode in ironic during registration and
during provisioning. This change updates the mode in ironic before
starting provisioning.

### DIFF
--- a/pkg/controller/baremetalhost/host_state_machine.go
+++ b/pkg/controller/baremetalhost/host_state_machine.go
@@ -103,7 +103,8 @@ func (hsm *hostStateMachine) updateHostStateFrom(initialState metal3v1alpha1.Pro
 			// it needs error handling logic that we can't support in
 			// this function.
 			if updateBootModeStatus(hsm.Host) {
-				info.log.Info("saving boot mode")
+				info.log.Info("saving boot mode",
+					"new mode", hsm.Host.Status.Provisioning.BootMode)
 			}
 		}
 	}
--- a/pkg/controller/baremetalhost/host_state_machine.go
+++ b/pkg/controller/baremetalhost/host_state_machine.go
@@ -88,6 +88,24 @@ func (hsm *hostStateMachine) updateHostStateFrom(initialState metal3v1alpha1.Pro
 			stateChanges.With(stateChangeMetricLabels(initialState, hsm.NextState)).Inc()
 		})
 		hsm.Host.Status.Provisioning.State = hsm.NextState
+		// Here we assume that if we're being asked to change the
+		// state, the return value of ReconcileState (our caller) is
+		// set up to ensure the change in the host is written back to
+		// the API. That means we can safely update any status fields
+		// along with the state.
+		switch hsm.NextState {
+		case metal3v1alpha1.StateRegistering,
+			metal3v1alpha1.StateInspecting,
+			metal3v1alpha1.StateProvisioning:
+			// TODO: When the user-selectable profile field is
+			// removed, move saveHostProvisioningSettings() from the
+			// controller to this point. We can't move it yet because
+			// it needs error handling logic that we can't support in
+			// this function.
+			if updateBootModeStatus(hsm.Host) {
+				info.log.Info("saving boot mode")
+			}
+		}
 	}
 }
 
@@ -200,9 +218,6 @@ func (hsm *hostStateMachine) handleRegistering(info *reconcileInfo) actionResult
 			hsm.NextState = metal3v1alpha1.StateProvisioned
 		case hsm.Host.NeedsHardwareInspection():
 			hsm.NextState = metal3v1alpha1.StateInspecting
-			if updateBootModeStatus(hsm.Host) {
-				info.log.Info("saving boot mode")
-			}
 		case hsm.Host.NeedsHardwareProfile():
 			hsm.NextState = metal3v1alpha1.StateMatchProfile
 		default:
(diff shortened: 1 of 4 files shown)


pkg/controller/baremetalhost/host_state_machine.go
```diff
@@ -87,6 +87,24 @@ func (hsm *hostStateMachine) updateHostStateFrom(initialState metal3v1alpha1.Pro
...
+		case metal3v1alpha1.StateRegistering,
+			metal3v1alpha1.StateInspecting,
+			metal3v1alpha1.StateProvisioning:
+			// TODO: When the user-selectable profile field is
+			// removed, move saveHostProvisioningSettings() from the
+			// controller to this point. We can't move it yet because
+			// it needs error handling logic that we can't support in
+			// this function.
+			if updateBootModeStatus(hsm.Host) {
+				info.log.Info("saving boot mode")
```
hardys, line 105: Not that we have to fix it in this PR, but this log output would potentially be more useful if it included the bootMode that we saved?

hardys:
lgtm but needs a rebase

pkg/controller/baremetalhost/host_state_machine.go
dhellmann: Sure, that makes sense. Done.

dhellmann:
Rebased

pkg/provisioner/ironic/ironic.go
```diff
@@ -625,102 +641,51 @@ func (p *ironicProvisioner) getUpdateOptsForNode(ironicNode *nodes.Node) (update
...
-	// image_source
-	var op nodes.UpdateOp
-	if _, ok := ironicNode.InstanceInfo["image_source"]; !ok {
-		op = nodes.AddOp
-		p.log.Info("adding image_source")
-	} else {
-		op = nodes.ReplaceOp
-		p.log.Info("updating image_source")
-	}
+	// instance_uuid
```
zaneb, line 644: I'd have assumed we only want to set this when provisioning ('instance' here means Nova instance).

pkg/provisioner/ironic/ironic.go
```diff
@@ -535,6 +535,21 @@ func (p *ironicProvisioner) InspectHardware() (result provisioner.Result, detail
 				err = nil
 				return
 			default:
+				// Update the node settings in ironic to ensure we
+				// have the boot mode configured, at least.
+				var updates nodes.UpdateOpts
+				updates, err = p.getUpdateOptsForNode(ironicNode)
```
zaneb, line 541: This would seem a lot lower-risk if we just refactored the code for setting the boot mode option out of getUpdateOptsForNode() and only set that one here.

pkg/provisioner/ironic/ironic.go
```diff
@@ -625,102 +641,51 @@ func (p *ironicProvisioner) getUpdateOptsForNode(ironicNode *nodes.Node) (update
...
 	)
 
-	checksum, checksumType, _ := p.host.GetImageChecksum()
-
-	// image_os_hash_algo
-	if _, ok := ironicNode.InstanceInfo["image_os_hash_algo"]; !ok {
+	// cpu_arch
+	//
+	// FIXME(dhellmann): This should come from inspecting the
+	// host.
```
zaneb, line 658: We're now calling this func before inspection, which will make it hard to change this.

pkg/controller/baremetalhost/host_state_machine.go
```diff
@@ -88,6 +88,25 @@ func (hsm *hostStateMachine) updateHostStateFrom(initialState metal3v1alpha1.Pro
 			stateChanges.With(stateChangeMetricLabels(initialState, hsm.NextState)).Inc()
 		})
 		hsm.Host.Status.Provisioning.State = hsm.NextState
+		// Here we assume that if we're being asked to change the
+		// state, the return value of ReconcileState (our caller) is
+		// set up to ensure the change in the host is written back to
+		// the API. That means we can safely update any status fields
+		// along with the state.
+		switch hsm.NextState {
+		case metal3v1alpha1.StateRegistering,
```
zaneb, line 97: Is there any point doing this before Registering when it's not copied to Ironic until we reach Inspecting?

I don't think it's safe to do this before #388 is merged, because right now this could end up changing the status of a provisioned Host so that the status would no longer match what we actually provisioned with.

zaneb (commented):
Setting all of the Node options (except the image) prior to inspecting seems like a big change with unpredictable effects that could easily be avoided if our goal is just to fix this bug.

pkg/provisioner/ironic/ironic.go
dhellmann: IIUC, ironic doesn't use any of these values until it needs them, so setting them early shouldn't hurt. They're reset before provisioning, as well.

pkg/provisioner/ironic/ironic.go
dhellmann: It's not clear to me why we need to set this at all if it can come from inspection, but when we do drop the profile we would necessarily not set it here unless we did have inspection data.

pkg/controller/baremetalhost/host_state_machine.go
dhellmann: I can remove registering from the list here.

dhellmann:
/hold

Let's see how #645 looks instead of this one.

dhellmann:
This is no longer needed.

PR #637, opened 2020-09-16
simplify running operator locally

dhellmann:
Different versions of the operator-sdk expect different arguments to
run an operator locally, outside of the cluster. All of the versions
simply build the binary and execute it, though, so we might as well
just do the same to avoid conflicts from version drift on different
developers' systems.

/cc @zaneb @zhouhao3

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,8 @@ SOURCE_GIT_COMMIT ?= $(shell git rev-parse --verify 'HEAD^{commit}')
 BUILD_VERSION ?= $(shell git describe --always --abbrev=40 --dirty)
 export LDFLAGS="-X github.com/metal3-io/baremetal-operator/pkg/version.Raw=${BUILD_VERSION} -X github.com/metal3-io/baremetal-operator/pkg/version.Commit=${SOURCE_GIT_COMMIT}"
 
+OPERATOR_BIN = ./build/_output/bin/baremetal-operator
+
 # Set some variables the operator expects to have in order to work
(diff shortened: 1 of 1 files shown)


maelk:
lgtm

Makefile
```diff
@@ -134,18 +136,12 @@ e2e-local:
...
-run: ## Run the operator outside of a cluster in development mode
-	operator-sdk run --local \
-		--go-ldflags=$(LDFLAGS) \
-		--watch-namespace=$(RUN_NAMESPACE) \
-		--operator-flags="-dev"
+run: build ## Run the operator outside of a cluster in development mode
+	$(OPERATOR_BIN) -dev -namespace=$(RUN_NAMESPACE)
 
 .PHONY: demo
 demo: ## Run the operator outside of a cluster using the demo driver
```
zaneb, line 143: Should depend on `build` also.

dhellmann: Done

zhouhao3:
LGTM


I think other related documents need to be changed accordingly. (e.g docs/dev-setup.md)

dhellmann:
> LGTM
> 
> I think other related documents need to be changed accordingly. (e.g docs/dev-setup.md)

It was always possible to use `make run`. This change only affects what commands are used by the Makefile. All of the previous dependencies are still needed for other parts of the development process. What sorts of changes do you have in mind?

dhellmann:
This will be obsoleted by https://github.com/metal3-io/baremetal-operator/pull/650

PR #639, opened 2020-09-18
Added namespace 'metal3'

FaridHub:
Running both ironic and baremetal-operator in single namespace would be much easier to debug and fetch the statuses of each component by a single 'kubectl get all -n metal3' command instead of running both in different namespaces.

### DIFF
--- a/ironic-deployment/ironic/ironic.yaml
+++ b/ironic-deployment/ironic/ironic.yaml
@@ -2,6 +2,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: metal3-ironic
+  namespace: metal3
 spec:
   replicas: 1
   selector:


maelk:
I don't think we should hardcode this here. you can achieve this by providing the namespace in your command when creating the deployment, or by using a kustomization, to add the namespace. Have a look at how Metal3-dev-env does it. This kind of thing should be left up to the user and not hardcoded IMO.

FaridHub:
> I don't think we should hardcode this here. you can achieve this by providing the namespace in your command when creating the deployment, or by using a kustomization, to add the namespace. Have a look at how Metal3-dev-env does it. This kind of thing should be left up to the user and not hardcoded IMO.

FaridHub:
Thanks for the clarification @maelk .

PR #640, opened 2020-09-18
Doc: Running baremetal-op on On-prem k8 cluster.

FaridHub:
Added a document to run the baremetal-operator on on-prem Kubernetes cluster.

### DIFF
--- a/docs/baremetal-operator-on-kubernetes-cluster.md
+++ b/docs/baremetal-operator-on-kubernetes-cluster.md
@@ -72,10 +72,11 @@ CACHEURL=http://<Ironic-ClusterIP>/images
 IRONIC_FAST_TRACK=false
 ```
 
-* ###Start running Baremetal-operator
+* ### Start running Baremetal-operator
 
   * The below command will apply all
 CRDs, namespace, RBAC and Baremetal-Operator.
+
 
 ```console
 > kubectl apply -k deploy/default
@@ -93,7 +94,9 @@ baremetal-machine.
 
   * Download the desired OS (in this case CentOS 7 qcow) to provision the
 baremetals.
+
   * Using md5sum make a md5 checksum file ending the name with '.md5sum'.
+
   * Copy the OS image and md5sum to the ironic pod -> ironic container
 running in cluster.
 
--- a/docs/baremetal-operator-on-kubernetes-cluster.md
+++ b/docs/baremetal-operator-on-kubernetes-cluster.md
@@ -14,12 +14,14 @@ Fully ready Kubernetes cluster (version >= 1.17)
 
 ## Spinning
 
-Clone the **Baremetal-operator** on your local.
+* ### Clone the **Baremetal-operator**
 
 ```console
 > git clone https://github.com/metal3-io/baremetal-operator.git
 > cd baremetal-operator
 ```
+
+* ### Start **Ironic**
 
 We will be applying the ironic first as we need the cluster IP
 of ironic pod to configure it in the config-maps of Baremetal-operator for
@@ -29,6 +31,8 @@ inter-pod communication.
 > kubectl apply -k ironic-deployement/default
 > kubectl get pods -n default
 ```
+
+* ### Update **config maps**
 
 Take the Cluster-IP Address of ironic pod running in default namespace
 and copy it in the config maps of Baremetal-operator by replace the default
@@ -72,10 +76,10 @@ CACHEURL=http://<Ironic-ClusterIP>/images
 IRONIC_FAST_TRACK=false
 ```
 
-* ### Start running Baremetal-operator
+* ### Start Baremetal-operator
 
-  * The below command will apply all
-CRDs, namespace, RBAC and Baremetal-Operator.
+The below command will apply all CRDs, namespace, RBAC
+and Baremetal-Operator.
 
 
 ```console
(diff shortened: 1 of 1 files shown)


docs/baremetal-operator-on-kubernetes-cluster.md
```diff
@@ -0,0 +1,179 @@
...
+```console
+> git clone https://github.com/metal3-io/baremetal-operator.git
+> cd baremetal-operator
+```
+
+* ### Start **Ironic**
+
+We will be applying the ironic first as we need the cluster IP
+of ironic pod to configure it in the config-maps of Baremetal-operator for
+inter-pod communication.
```
maelk, line 28: this should be done after the configmap update. Otherwise Ironic will start with wrong values and not pick on the update.

maelk (commented):
looks pretty good. I just have one nit. 
Please note however that there are multiple deployment flavours (mostly with or without keepalived for now, but TLS and basic auth are coming) and that could be mentioned ?

docs/baremetal-operator-on-kubernetes-cluster.md
FaridHub: I didn't quite get you, baremetal-op config maps must be updated with ironic ClusterIP; and for that ironic should be running first, isn't it?
I am not much aware of others ways to do it. I found this way (document) to be successfully running and provisioning.
As I couldn't find any documentation for running the metal3-> baremetal-op completely in an on-prem cluster env. 
I will be happy to take suggestions.

maelk: ok, sorry for being unclear. When you deploy ironic, it also uses a configmap that is in ironic-deployment/default. In your case it probably worked because you anyways have a eth2 interface (default there) and would use pretty much the same values. However, if you were not using those default values and did not update that configmap, ironic would not start. Be careful that this configmap is not the Baremetal Operator configmap that you describe later. You need to do the same update for the ironic configmap.

docs/baremetal-operator-on-kubernetes-cluster.md
```diff
@@ -0,0 +1,179 @@
...
+```console
+> kubectl apply -k ironic-deployement/default
+> kubectl get pods -n default
+```
+
+* ### Update **config maps**
+
+Take the Cluster-IP Address of ironic pod running in default namespace
+and copy it in the config maps of Baremetal-operator by replace the default
+provisioning-IP in *deploy/ironic_ci.env* as mentioned below.
```
maelk, line 39: modifying this one is unnecessary if you deploy ironic in your cluster (as you are explaining it above). It is only used when you deplyo with run_local_ironic.sh (not in cluster)

docs/baremetal-operator-on-kubernetes-cluster.md
```diff
@@ -36,8 +36,8 @@ inter-pod communication.
 
 Take the Cluster-IP Address of ironic pod running in default namespace
 and copy it in the config maps of Baremetal-operator by replace the default
-provisioning-IP in *ironic-deployment/default/ironic_bmo_configmap.env* as mentioned below.
-One may also have to change *provisioning-interface*
+provisioning-IP in *ironic-deployment/default/ironic_bmo_configmap.env* as
+mentioned below. One may also have to change *provisioning-interface*
 (by default it is PROVISIONING_INTERFACE=eth2) and its value will change
 as per the OS and hardware specs.
```
FaridHub, line 43: @maelk Do you want me to remove this portion and keep only baremetal-operator/configmap update?

maelk: I think it is still problematic, specifically the order creation of the configmap vs starting ironic and BMO. You need to have the proper config map file before deploying ironic. if you don't, just updating the configmap will not be taken into account by ironic because ironic would need to restart to take it into account.

Regarding your question about clusterIP, ironic should not be using clusterIP. It runs on host networking. You can have it use a Virtual IP with keepalived. This would be the proper way to deploy ironic.

FaridHub: In my case, i had been reapplying the ironic pods after getting an node assigned to it. Mostly, ironic runs everytime in the same worker node, even after reapplying the yaml many times. So, just like static address it can be updated in baremetal-operator. However, one efficient way would be using "NodeSelector" in k8 yaml configuration under specs so that ironic runs only in the specified worker-node everytime. The same can be updated in the baremetal yaml. Do, you want me to update the doc as such?

docs/baremetal-operator-on-kubernetes-cluster.md
```diff
@@ -0,0 +1,179 @@
...
+the cluster; starting with spinning the components
+**Baremetal-operator and Ironic** over a
+**On-Premises Kubernetes Cluster**, till the **provisioning of baremetal
+machines** using only cluster resources.
+
+## Prerequisites
+
+Fully ready Kubernetes cluster (version >= 1.17)
+
+## Spinning
```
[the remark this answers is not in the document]
FaridHub: Sure, will change it as you have suggested above.

docs/baremetal-operator-on-kubernetes-cluster.md
maelk: The idea with the VIP and deploying ironic with keepalived is to remove this constraint of having to always run on the same node. you don't need to set a node selector if you are using a VIP that will automatically be added to the node where Ironic runs. I would very much recommend to use this approach, it gives much more flexibility, for example if that node goes down, Ironic can be restarted somewhere else without issues.

PR #647, opened 2020-09-23
start adding unit tests for ironic.ValidateManagementAccess

dhellmann:


dhellmann:
I would like to freeze non-essential changes for a few days to try to land https://github.com/metal3-io/baremetal-operator/pull/650 without having to rebase it, because rebasing will mean redoing the work from scratch.

/hold

dhellmann:
#655 has merged

/hold cancel

dhellmann:
I've rebased this on the work @andfasano did in #649 

/test-integration

pkg/provisioner/ironic/testserver/ironic.go
```diff
@@ -75,3 +87,45 @@ func (m *IronicMock) WithNodeStatesProvision(nodeUUID string) *IronicMock {
...
+	return m
+}
+
+// CreateNodes configures the server so POSTing to /v1/nodes saves the data
+func (m *IronicMock) CreateNodes() *IronicMock {
+	m.Handler("/v1/nodes", func(w http.ResponseWriter, r *http.Request) {
+		bodyRaw, err := ioutil.ReadAll(r.Body)
+		if err != nil {
+			m.logRequest(r, fmt.Sprintf("ERROR: %s", err))
+			http.Error(w, fmt.Sprintf("%s", err), 500)
```
andfasano, line 103: Minor point, `http.StatusInternalServerError` could be more readable

pkg/provisioner/ironic/testserver/ironic.go
```diff
@@ -75,3 +87,45 @@ func (m *IronicMock) WithNodeStatesProvision(nodeUUID string) *IronicMock {
...
+		m.CreatedNodes = append(m.CreatedNodes, CreatedNode{
+			Body: body,
+			UUID: uuid,
+		})
+
+		// hackily add uuid to the json response by inserting it to the front of the string
+		response := fmt.Sprintf("{\"uuid\": \"%s\", %s", uuid, strings.TrimLeft(body, "{"))
+
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(201)
```
andfasano, line 126: And `http.StatusCreated` here

pkg/provisioner/ironic/validatemanagementaccess_test.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	result, err := prov.ValidateManagementAccess(false)
+	if err != nil {
+		t.Fatalf("error from ValidateManagementAccess: %s", err)
+	}
+	assert.Equal(t, "", result.ErrorMessage)
+	assert.NotEqual(t, "", host.Status.Provisioning.ID)
+	assert.Equal(t, ironic.CreatedNodes[0].UUID, host.Status.Provisioning.ID)
+}
```
andfasano, line 97: (Not blocking) I noticed that the three test cases have been split into three different test methods (`TestValidateManagementAccessNoMAC`, `TestValidateManagementAccessMACOptional` and `TestValidateManagementAccessCreateNode`) sharing a pretty similar setup code.
Maybe it's not worth in this case, but to transform the three methods in a list of cases in a single test method it could be useful to "upgrade" the `makeHost()` factory method to a proper `host builder` object with fluent interface, so that also the host could become a parameter of the cases list (and of course also the expectations code will need to be adjusted).

pkg/provisioner/ironic/ironic_test.go
```diff
@@ -62,3 +70,6 @@ func makeHost() *metal3v1alpha1.BareMetalHost {
 		},
 	}
 }
+
+// Implements provisioner.EventPublisher to swallow events for tests.
+func nullEventPublisher(reason, message string) {}
```
andfasano, line 75: Nice one, we could introduce a set of different hard-coded publishers (ie null, std) to be reused in all the tests to remove some other lines of code

pkg/provisioner/ironic/testserver/ironic.go
```diff
@@ -75,3 +87,45 @@ func (m *IronicMock) WithNodeStatesProvision(nodeUUID string) *IronicMock {
...
+		m.t.Logf("%s: uuid %s", m.name, uuid)
+
+		// Record what we have so the test can examine it later
+		m.CreatedNodes = append(m.CreatedNodes, CreatedNode{
+			Body: body,
+			UUID: uuid,
+		})
+
+		// hackily add uuid to the json response by inserting it to the front of the string
+		response := fmt.Sprintf("{\"uuid\": \"%s\", %s", uuid, strings.TrimLeft(body, "{"))
```
andfasano, line 123: Just as a confirmation, is it fine to return all the fields sent in the  `body` request as a response? From a quick glance on the API [here](https://docs.openstack.org/api-ref/baremetal/?expanded=create-node-detail) should be fine. Maybe in future we could need something more sophisticated, but for the current test seems ok

pkg/provisioner/ironic/validatemanagementaccess_test.go
andfasano: Given that the `ValidateManagementAccess` is a pretty big method to be covered, the existing node and/or credentials change case will be added within this PR?

pkg/provisioner/ironic/findhost_test.go
```diff
@@ -21,22 +21,22 @@ func TestFindExistingHost(t *testing.T) {
 			name:           "no-node",
 			hostName:       "name",
 			provisioningID: "uuid",
-			ironic: testserver.New(t, "ironic").NotFound("/v1/nodes/name").
+			ironic: testserver.NewIronic(t).NotFound("/v1/nodes/name").
```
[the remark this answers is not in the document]
dhellmann: Good point. Done.

pkg/provisioner/ironic/testserver/ironic.go
dhellmann: Fixed

pkg/provisioner/ironic/testserver/ironic.go
dhellmann: Fixed

pkg/provisioner/ironic/validatemanagementaccess_test.go
dhellmann: The setup portions of the functions are similar, but the checks were different enough that it seemed like it would be clearer to just write separate functions this time, instead of having more complicated logic to decide which tests to apply in each case.

In terms of test coverage, I think this PR is big enough and I would like to merge it before finishing the tests for ValidateManagementAccess.

andfasano: Yes, makes sense

andfasano (approved):
Current changes look fine to me

dhellmann:
/test-integration

Rebased to get the `make docker` fix from #647

pkg/provisioner/ironic/testserver/ironic.go
```diff
@@ -66,7 +78,12 @@ func (m *IronicMock) WithDrivers() *IronicMock {
 
 // WithNode configures the server with a valid response for /v1/nodes
 func (m *IronicMock) WithNode(node nodes.Node) *IronicMock {
-	m.ResponseJSON("/v1/nodes/"+node.UUID, node)
+	if node.UUID != "" {
+		m.ResponseJSON("/v1/nodes/"+node.UUID, node)
+	}
+	if node.Name != "" {
```
andfasano, line 84: I have a couple of question on this change: 
- Is it a valid case to have the UUID unset and Name set?
- When both UUID and Name are set, will the ServeMux.Handle panic?

dhellmann: A real ironic service wouldn't have a node without a UUID, but in the tests we sometimes want the node to only be available via one or the other value to simulate the node having settings different from what the provisioner expects.

I'm not sure why the handler would panic? The endpoints are different (one looks up by UUID, the other by name).

andfasano: In case Name == UUID, but I think it's a trivial case

dhellmann: Ah, yes, I didn't consider that. I think it's OK for the test to panic in that situation, since that doesn't seem like something we would expect to encounter in real data. I could add a test to protect against it, if you think it's worth it?

asalkeld (commented):
/lgtm

pkg/provisioner/ironic/testserver/ironic.go
andfasano: I don't think it's worth it, and agree it's fine to panic in such situation

andfasano:
I'm also fine with the changes

pkg/provisioner/ironic/testbmc/testbmc.go
```diff
@@ -0,0 +1,83 @@
...
+import (
+	"net/url"
+
+	"github.com/metal3-io/baremetal-operator/pkg/bmc"
+)
+
+func init() {
+	bmc.RegisterFactory("test", newTestBMCAccessDetails, []string{})
+	bmc.RegisterFactory("test-needs-mac", newTestBMCAccessDetails, []string{})
+}
```
zaneb, line 12: Would it be safer to register these explicitly in tests rather than it happening as a side-effect here?

dhellmann: Probably. On the other hand, all of the rest of the registry stuff expects to be invoked from an init() function like this, the registry is global state without a way to reset right now, and I'm not sure what happens if the same name is registered more than once.

I would be happy to consider making the registry more accepting of dynamic use in a separate PR.

PR #648, opened 2020-09-23
Add gomod script

honza:
This script verifies that go.mod and go.sum are up to date, and that `go mod tidy` can be executed successfully.

This script should be run on every patch in CI.


maelk (commented):
/lgtm

PR #649, opened 2020-09-24
start adding unit tests for Ironic.InspectHardware

andfasano:
Increasing test coverage for Ironic provisioner

### DIFF
--- a/pkg/provisioner/ironic/inspecthardware_test.go
+++ b/pkg/provisioner/ironic/inspecthardware_test.go
@@ -1,0 +1,179 @@
+package ironic
+
+import (
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/metal3-io/baremetal-operator/pkg/bmc"
+	"github.com/metal3-io/baremetal-operator/pkg/provisioner/ironic/clients"
+	"github.com/metal3-io/baremetal-operator/pkg/provisioner/ironic/testserver"
+
+	"github.com/gophercloud/gophercloud/openstack/baremetal/v1/nodes"
+	"github.com/gophercloud/gophercloud/openstack/baremetalintrospection/v1/introspection"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestInspectHardware(t *testing.T) {
+
+	nodeUUID := "33ce8659-7400-4c68-9535-d10766f07a58"
+
+	cases := []struct {
+		name      string
+		ironic    *testserver.IronicMock
+		inspector *testserver.InspectorMock
+
+		expectedDirty        bool
+		expectedRequestAfter int
+		expectedResultError  string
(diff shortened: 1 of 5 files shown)


dhellmann:
I would like to freeze non-essential changes for a few days to try to land https://github.com/metal3-io/baremetal-operator/pull/650 without having to rebase it, because rebasing will mean redoing the work from scratch.

/hold

pkg/provisioner/ironic/testserver/server.go
```diff
@@ -81,6 +82,17 @@ func (m *MockServer) Response(pattern string, payload string) *MockServer {
 	return m
 }
 
+// ResponseJSON marshals the JSON object as payload returned by the response
+// handler
+func (m *MockServer) ResponseJSON(pattern string, payload interface{}) *MockServer {
```
dhellmann, line 87: You were way less lazy here than I was. We should land your change before #647 and then I will rework that one to use this JSON handling that you've added.

pkg/provisioner/ironic/testserver/server.go
```diff
@@ -81,6 +82,17 @@ func (m *MockServer) Response(pattern string, payload string) *MockServer {
...
+	content, err := json.Marshal(payload)
+	if err != nil {
+		panic(err)
```
dhellmann, line 90: The server has a handle to the `testing.T`, so we could just report a fatal error to stop the current test, instead of panicking.

pkg/provisioner/ironic/inspecthardware_test.go
```diff
@@ -0,0 +1,94 @@
...
+			ironic: testserver.NewIronic(t).Ready().WithNode(nodes.Node{
+				UUID: nodeUUID,
+			}),
+			inspector: testserver.NewInspector(t).Ready().
+				WithIntrospection(nodeUUID, introspection.Introspection{
+					Finished: true,
+				}).
+				WithIntrospectionData(nodeUUID, introspection.Data{
+					Inventory: introspection.InventoryType{
+						Hostname: "node-0",
```
dhellmann, line 51: Where does the name come from?

dhellmann (commented):
Nice!

pkg/provisioner/ironic/testserver/server.go
andfasano: Yes, sounds better

pkg/provisioner/ironic/inspecthardware_test.go
andfasano: Just followed the Ironic Inspector API, ie https://docs.openstack.org/ironic-inspector/5.1.0/http-api.html#get-introspection-data, to have roughly a one-to-one relationship

PR #650, opened 2020-09-24
rebuild using kubebuilder 2.3.1

dhellmann:
This implements the work described in
https://github.com/metal3-io/metal3-docs/blob/master/design/baremetal-operator/kubebuilder-migration.md

It deviates a bit from the work list, because it turns out not to be
necessary to fork controller-tools now that they have added a flag to
support "unsafe" types like our float64 field.


main.go
```diff
@@ -0,0 +1,127 @@
...
+		provisionerFactory = fixture.New
+	} else if runInDemoMode {
+		ctrl.Log.Info("USING DEMO MODE")
+		provisionerFactory = demo.New
+	} else {
+		provisionerFactory = ironic.New
+		ironic.LogStartup()
+	}
+
+	if err = (&metal3iocontroller.BareMetalHostReconciler{
```
asalkeld, line 111: personally not a fan of this making the Reconciler struct public, you could just make a metal3iocontroller.New()

Makefile
```diff
@@ -28,163 +27,157 @@ export IRONIC_INSPECTOR_ENDPOINT=http://localhost:5050/v1/
...
+demo: generate fmt vet manifests ## Run the controller in demo mode
+	go run -ldflags $(LDFLAGS) ./main.go -dev -demo-mode -namespace $(RUN_NAMESPACE)
 
-.PHONY: unit-cover
-unit-cover: ## Run unit tests with code coverage
-	go test -coverprofile=$(COVER_PROFILE) $(GO_TEST_FLAGS) $(PACKAGES)
-	go tool cover -func=$(COVER_PROFILE)
+# Install CRDs into a cluster
+install: manifests
+	kustomize build config/crd | kubectl apply -f -
```
asalkeld, line 77: is kustomize assumed to be installed?

Makefile
```diff
@@ -28,163 +27,157 @@ export IRONIC_INSPECTOR_ENDPOINT=http://localhost:5050/v1/
...
+	CONTROLLER_GEN_TMP_DIR=$$(mktemp -d) ;\
+	cd $$CONTROLLER_GEN_TMP_DIR ;\
+	go mod init tmp ;\
+	go get sigs.k8s.io/controller-tools/cmd/controller-gen@$(CONTROLLER_TOOLS_VERSION) ;\
+	rm -rf $$CONTROLLER_GEN_TMP_DIR ;\
+	}
+CONTROLLER_GEN=$(GOBIN)/controller-gen
+else
+CONTROLLER_GEN=$(shell which controller-gen)
+endif
```
asalkeld, line 162: you could put this at the top
```
ifeq (, $(shell which controller-gen))
CONTROLLER_GEN=$(GOBIN)/controller-gen
else
CONTROLLER_GEN=$(shell which controller-gen)
endif
```
then
```
$(CONTROLLER_GEN):
   #install stuff
```
just to avoid the phony target and repeated installs

asalkeld: actually isn't it better to vendor? if the generated code varies across versions that's going to be a pain

config/default/manager_auth_proxy_patch.yaml
```diff
@@ -0,0 +1,25 @@
...
+kind: Deployment
+metadata:
+  name: controller-manager
+  namespace: system
+spec:
+  template:
+    spec:
+      containers:
+      - name: kube-rbac-proxy
+        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.5.0
```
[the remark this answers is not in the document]
dhellmann: This was produced by the code generator. @maelk and I are going to collaborate to fix up all of the kustomize inputs after this PR has landed and we can unblock the other PRs in this repo.

main.go
dhellmann: This is what kubebuilder generates. I don't know if the controller runtime expects these members to be public or not, but didn't want to deviate too far from what the code generator does in this PR.

Makefile
dhellmann: The install logic uses the specific version we need. It's based on what kubebuilder generates. We could enhance it to install into a local directory, like with kustomize, in another PR if we think that's important.

controllers/metal3.io/action_result.go
```diff
@@ -0,0 +1,128 @@
+package controllers
+
+import (
+	"math"
+	"math/rand"
+
+	metal3 "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
+
+	"time"
```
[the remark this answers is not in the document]
dhellmann: I've fixed up the imports in all of the files in this directory to use a consistent order.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -0,0 +1,936 @@
...
+// Reconcile is called when a host changes.
+func (r *BareMetalHostReconciler) Reconcile(request ctrl.Request) (result ctrl.Result, err error) {
+	reconcileCounters.With(hostMetricLabels(request)).Inc()
+	defer func() {
+		if err != nil {
+			reconcileErrorCounter.Inc()
+		}
+	}()
+
+	_ = context.Background()
```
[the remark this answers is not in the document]
dhellmann: No, that's debris from the code generator output. Removed.

Makefile
dhellmann: We used to have a rule to install it. I've restored that in an update to this PR.

dhellmann:
We may need https://github.com/metal3-io/project-infra/pull/151 to make the jobs work.

main.go
```diff
@@ -44,22 +45,37 @@ func init() {
...
+
+	flag.BoolVar(&devLogging, "dev", false, "enable developer logging")
+	flag.StringVar(&metricsAddr, "metrics-addr", "127.0.0.1:8085",
+		"The address the metric endpoint binds to.")
+	// From CAPI point of view, BMO should be able to watch all namespaces
+	// in case of a deployment that is not multi-tenant. If the deployment
+	// is for multi-tenancy, then the BMO should watch only the provided
+	// namespace.
+	flag.StringVar(&watchNamespace, "namespace", "",
+		"Namespace that the controller watches to reconcile BMO objects.")
```
zaneb, line 74: Is it safe to add this without adding the code that we used to have to look up the current namespace if it wasn't supplied?

zaneb (commented):
There's a lot here. The only way I've found to review it is to basically recreate the migration myself and then do diffs against this branch to find out what I missed. I'm only part way through, but I did have some questions so I'll post the ones I've found now.

I also have an idea for how to structure the history so that e.g. git blame is not lost on important files like the bmh controller. It should make rebasing a bit easier as well, so we might be able to unblock other changes. I'm still working my way through all the diffs though (with a goal of ending up in roughly the same place via a different history graph). So I'll have to come back to this on Monday.

Makefile
```diff
@@ -2,7 +2,14 @@
 # Image URL to use all building/pushing image targets
 IMG ?= controller:latest
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
-CRD_OPTIONS ?= "crd:trivialVersions=true"
+# trivialVersions=false means generate the CRD with multiple versions,
+#     eventually allowing conversion webhooks
+# allowDangerousTypes=true lets use the float64 field for clock speeds
+# crdVersions=v1 generates the v1 version of the CRD type
+# preserveUnknownFields=false causes the API server to discard "extra" data instead of storing it
```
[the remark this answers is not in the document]
dhellmann: I think you're right. I figured being explicit wouldn't hurt, though.

main.go
dhellmann: The old lookup code was part of the operator-sdk library that has been dropped. AFAICT, we always set the namespace explicitly. We may want to add some logic to let us use an environment variable to do that more easily than a command line flag, for when we're running inside the cluster.

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -1,46 +1,587 @@
...
+	Image Image `json:"image,omitempty"`
+
+	// The RootDevicehints set by the user
+	RootDeviceHints *RootDeviceHints `json:"rootDeviceHints,omitempty"`
+
+	// BootMode indicates the boot mode used to provision the node
+	BootMode BootMode `json:"bootMode,omitempty"`
+}
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
```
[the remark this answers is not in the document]
dhellmann: Perhaps? I just moved the old file into the new location, and I probably should have done what I did with the controller where I copied the contents over by hand in pieces.

dhellmann:
> There's a lot here. The only way I've found to review it is to basically recreate the migration myself and then do diffs against this branch to find out what I missed. I'm only part way through, but I did have some questions so I'll post the ones I've found now.
> 
> I also have an idea for how to structure the history so that e.g. git blame is not lost on important files like the bmh controller. It should make rebasing a bit easier as well, so we might be able to unblock other changes. I'm still working my way through all the diffs though (with a goal of ending up in roughly the same place via a different history graph). So I'll have to come back to this on Monday.

Perhaps I should remove the holds on other PRs and redo this PR with that new approach. Let's talk Monday.

maelk:
> Perhaps I should remove the holds on other PRs and redo this PR with that new approach. Let's talk Monday.

If so, then we'll need to revert the change of image for the generate job in project-infra. That effectively is blocking all other PRs for now.

dhellmann:
> > Perhaps I should remove the holds on other PRs and redo this PR with that new approach. Let's talk Monday.
> 
> If so, then we'll need to revert the change of image for the generate job in project-infra. That effectively is blocking all other PRs for now.

Let's see if #655 is easier to review. If not, then I'll revert the image change in infra to unblock others while we figure out how to make the migration reviewable.

maelk:
It was not possible to review either this or #655 on Github UI due to the file number restriction (max 3000). But I managed with vscode Github extension. Unfortunately it did not properly show the moved files (they were displayed as added) but overall could be reviewed without major issues.

main.go
maelk: That's fine by me as long as we comply with the CAPI requirements, that the namespace should be providable by command line, and if unset (both in command line and environment variable in that case), default to all namespaces.

dhellmann:
The consensus seems to be that #655 is easier to follow, so I'm going to close this one.

apis/metal3.io/v1alpha1/baremetalhost_types.go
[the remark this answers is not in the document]
zaneb: Further research reveals that they're probably not related, but the deepcopy-gen thing was to cause it to generate a DeepCopyObject() method, which was some [ancient](https://www.openshift.com/blog/kubernetes-deep-dive-code-generation-customresources) workaround that doesn't appear to be needed any more since [we are generating this anyway](https://github.com/metal3-io/baremetal-operator/pull/655/files#diff-682ec4bfea44947253d1b0be36f05931R78-R83).

PR #652, opened 2020-09-25
[doc nit] Fix section header "install operator-sdk"

asalkeld:


asalkeld:
generate error is:
```
/bin/sh: 1: openapi-gen: not found
```

maelk:
yes, that is because we changed the jobs to adapt for #655 that needs to go in first.

asalkeld:
obsolete

PR #658, opened 2020-09-30
Extend BaremetalHost CRD to support BIOS config (simplified version of #302)

asalkeld:
This builds on top of #302, but simplifies the API dramatically.

The idea is to closer match the design example and try and respond to https://github.com/metal3-io/baremetal-operator/pull/302#pullrequestreview-482140928 

/cc @dhellmann @longkb 

@longkb I hope you cool with me trying this out, I really don't want to step on your feet. I just want to help out.
I was hoping that we could get something simpler in and then add what we need bit-by-bit.
(so feel free to add this commit to your PR, and we can close this?)


pkg/bmc/bios.go
```diff
@@ -0,0 +1,149 @@
...
+	"reflect"
+	"strconv"
+
+	"github.com/gophercloud/gophercloud/openstack/baremetal/v1/nodes"
+
+	metal3v1alpha1 "github.com/metal3-io/baremetal-operator/pkg/apis/metal3/v1alpha1"
+)
+
+// A private method to build the clean steps for ILO configuration from BaremetalHost spec
+// TODO(asalkeld) I don't know which driver to associate this with
```
asalkeld, line 13: I can't find a access.Driver() that returns "ilo", so I don't know where to put this

asalkeld:
needs some more thought, field engineers, bigger customers want something different.

PR #666, opened 2020-10-06
Standardize the directory structure

asalkeld:
The current structure looks unusual/non-standard to me - please feel free to close this if I am wrong.
But looking at a handful of operators under github.com/openshift/ and here https://book-v1.book.kubebuilder.io/basics/project_creation_and_structure.html the standard seems to be
 
cmd/..
pkg/..

so I have moved as follows:

main.go -> cmd/manager/main.go
controllers -> pkg/controllers
apis -> pkg/apis


zaneb:
What I discovered after repeatedly misunderstanding it myself, is that https://book-v1.book.kubebuilder.io/ is the docs site for the *old* version of kubebuilder, and it has better SEO than the docs for the current version of kubebuilder (which is actually v2) at https://book.kubebuilder.io/ :angry:
If by luck you stumble across the current docs there's a thing on each page saying "Using Kubebuilder v1? Check the [legacy documentation](https://book-v1.book.kubebuilder.io/)", but there's nothing on the legacy documentation to tell you it's legacy :exploding_head: 

The current locations are actually [where the latest kubebuilder generates them](https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/simplified-scaffolding.md#revised-structure), so we don't need to move them back.
/close

asalkeld:
> What I discovered after repeatedly misunderstanding it myself, is that https://book-v1.book.kubebuilder.io/ is the docs site for the _old_ version of kubebuilder, and it has better SEO than the docs for the current version of kubebuilder (which is actually v2) at https://book.kubebuilder.io/ angry
> If by luck you stumble across the current docs there's a thing on each page saying "Using Kubebuilder v1? Check the [legacy documentation](https://book-v1.book.kubebuilder.io/)", but there's nothing on the legacy documentation to tell you it's legacy exploding_head
> 
> The current locations are actually [where the latest kubebuilder generates them](https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/simplified-scaffolding.md#revised-structure), so we don't need to move them back.
> /close

OK, still not a fan of it :man_shrugging: .. seems to break the *normal* structure of a golang project https://github.com/golang-standards/project-layout
Also I don't think this breaks kubebuilder at all..

zaneb:
Moving it doesn't break anything, but it will complicate using kubebuilder to generate additional code (more controllers, webhooks) in the future. I don't love how they've done it either, but its likely to become the standard for k8s controllers (and CAPM3 uses the same layout), so I think it's best to just go with it.

PR #667, opened 2020-10-07
[WIP] Add kustomize rendering support for build

digambar15:
Still work in progress

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -134,12 +134,12 @@ install: $(KUSTOMIZE) manifests ## Install CRDs into a cluster
 
 .PHONY: uninstall
 uninstall: $(KUSTOMIZE) manifests ## Uninstall CRDs from a cluster
-	kustomize build config/crd | kubectl delete -f -
+	$(KUSTOMIZE) build config/crd | kubectl delete -f -
 
 .PHONY: deploy
 deploy: $(KUSTOMIZE) manifests ## Deploy controller in the configured Kubernetes cluster in ~/.kube/config
-	cd config/manager && kustomize edit set image controller=${IMG}
-	kustomize build config/default | kubectl apply -f -
+	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
+	$(KUSTOMIZE) build config/default &> config/render/kustomization.yaml
 
 .PHONY: manifests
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -134,12 +134,12 @@ install: $(KUSTOMIZE) manifests ## Install CRDs into a cluster
...
 uninstall: $(KUSTOMIZE) manifests ## Uninstall CRDs from a cluster
-	kustomize build config/crd | kubectl delete -f -
+	$(KUSTOMIZE) build config/crd | kubectl delete -f -
 
 .PHONY: deploy
 deploy: $(KUSTOMIZE) manifests ## Deploy controller in the configured Kubernetes cluster in ~/.kube/config
-	cd config/manager && kustomize edit set image controller=${IMG}
-	kustomize build config/default | kubectl apply -f -
+	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
+	$(KUSTOMIZE) build config/default &> config/render/kustomization.yaml
```
maelk, line 142:
```suggestion
	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
	$(KUSTOMIZE) build config/default | kubectl apply -f -
```
You need to add `$(KUSTOMIZE) build config/default &> config/render/kustomization.yaml` on the next target, `manifests` and add a dependency to `$(KUSTOMIZE)` there too.
The name kustomization.yaml might be misleading as one would expect a kustomization. Maybe deployment.yaml or something similar ?

hack/install-controller-gen.sh
```diff
@@ -6,6 +6,7 @@ CONTROLLER_TOOLS_VERSION=${1:-v0.4.0}
 OUTPUT=${2:-${ORIG_DIR}/bin/controller-gen}
 
 WORK_DIR=$(mktemp --tmpdir -d controller-gen-XXX)
+cp -r config/render ${WORK_DIR}/
```
maelk, line 9: This should not be in this file. we should nto touch instal-controller-gen.sh what needs to be modified is generate.sh, to include the new file too : https://github.com/metal3-io/baremetal-operator/blob/master/hack/generate.sh#L18

maelk (changes requested):

maelk:
I think we can close this PR as it was done in #672

maelk:
/close
Please re-open if needed

PR #671, opened 2020-10-08
Fix tls certificate fetch path 

kashifest:
This is a fix to the tls certificate path. It should be `config` instead of `deploy`

### DIFF
--- a/tools/deploy.sh
+++ b/tools/deploy.sh
@@ -51,7 +51,7 @@ IRONIC_AUTH_DIR="${IRONIC_AUTH_DIR:-"${IRONIC_DATA_DIR}auth/"}"
 IRONIC_CERTS_DIR="${IRONIC_CERTS_DIR:-"${IRONIC_DATA_DIR}certs/"}"
 
 sudo mkdir -p "${IRONIC_DATA_DIR}"
-sudo chown -R "${USER}:${USER}" "${IRONIC_DATA_DIR}"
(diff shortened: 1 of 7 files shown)


tools/deploy.sh
```diff
@@ -154,7 +154,7 @@ if [ "${DEPLOY_TLS}" == "true" ]; then
     fi
 
     if [ "${DEPLOY_BMO}" == "true" ]; then
-        cp "${IRONIC_CACERT_FILE}" "${SCRIPTDIR}/deploy/tls/ca.crt"
+        cp "${IRONIC_CACERT_FILE}" "${SCRIPTDIR}/config/tls/ca.crt"
```
maelk, line 157: Looks good. while we are fixing this script, can you please change line 54 this way :
```
sudo chown -R "${USER}:$(id -gn)" "${IRONIC_DATA_DIR}"
```

kashifest: Done

config/manager/manager.yaml
```diff
@@ -3,7 +3,7 @@ kind: Namespace
 metadata:
   labels:
     control-plane: controller-manager
-  name: system
```
maelk, line 6: Don't remove the name!

maelk:
This is in deadlock with metal3-dev-env PR. Merging manually to unlock the metal3-dev-env PR.

PR #674, opened 2020-10-08
Add capm3 specific configurations in config/ dir

kashifest:
We realized that for capm3 having the namespace in `config/manager/manager.yaml` is problematic, fetching it out of there and adding it in its own directory (namespace). 

To allow using namespace and default, a root level kustomization is added. 

Fixing, the command for `config/manager/manager.yaml` from `manager` to `baremetal-operator`

### DIFF
--- a/config/kustomization.yaml
+++ b/config/kustomization.yaml
@@ -1,0 +1,6 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+bases:
+- namespace
+- default
--- a/config/namespace/kustomization.yaml
+++ b/config/namespace/kustomization.yaml
@@ -1,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+- namespace.yaml
--- a/config/namespace/namespace.yaml
+++ b/config/namespace/namespace.yaml
@@ -1,0 +1,6 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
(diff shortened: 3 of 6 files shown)


config/namespace/namespace.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    control-plane: controller-manager
+  name: system
```
maelk, line 6: the name is incorrect

config/kustomization.yaml
```diff
@@ -0,0 +1,3 @@
+bases:
```
maelk, line 1: Can you please add this at the top of the file :
```
kind: Kustomization
```

config/namespace/kustomization.yaml
```diff
@@ -0,0 +1,2 @@
+resources:
```
maelk, line 1: here too ?

maelk (commented):
Please re-generate the output file also

config/kustomization.yaml
kashifest: done

config/namespace/kustomization.yaml
kashifest: done

maelk (approved):

config/namespace/namespace.yaml
kashifest: done

maelk:
fix needed for CAPM3 and metal3-dev-env adjustments
/lgtm

PR #676, opened 2020-10-09
Fix kustomization path

kashifest:
This PR fixes the kustomization path for CAPM3 v1alpha3. It should point to `config/` to include the namespace

### DIFF
--- a/tools/deploy.sh
+++ b/tools/deploy.sh
@@ -36,7 +36,7 @@ fi
 if [ "${DEPLOY_TLS}" == "true" ]; then
     BMO_SCENARIO="${BMO_SCENARIO}/tls"
     IRONIC_SCENARIO="${IRONIC_SCENARIO}/tls"
-else
+elif [ "${DEPLOY_BASIC_AUTH}" == "true" ]; then
     BMO_SCENARIO="${BMO_SCENARIO}/default"
 fi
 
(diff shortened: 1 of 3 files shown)


maelk (approved):

maelk:
We need this in to pass the CI in metal3-dev-env PR.

tools/deploy.sh
```diff
@@ -36,7 +36,7 @@ fi
 if [ "${DEPLOY_TLS}" == "true" ]; then
     BMO_SCENARIO="${BMO_SCENARIO}/tls"
     IRONIC_SCENARIO="${IRONIC_SCENARIO}/tls"
-else
+elif [ "${DEPLOY_BASIC_AUTH}" == "true" ]; then
```
dhellmann, line 39: So BMO_SCENARIO will be unset if we're using no authorization. Is that OK? Maybe we don't use/support that mode any more?

kashifest: Good point. May be we should be more explicit here. What's your suggestion

dhellmann: For now I would be happy with an else statement that just exited with an error. We can add noauth support back if it actually comes up as something we need.

dhellmann: What do you think of something like https://github.com/metal3-io/baremetal-operator/pull/681?

kashifest: Fair enough. I ll put a patch

PR #677, opened 2020-10-09
Fix tls kustomization path

kashifest:
#676 is not properly solving the kustomize path issue. This is a fix for that.


maelk (approved):

maelk:
same here, merging to get the fix in to be able to test the metal3-dev-env PR properly

PR #681, opened 2020-10-09
require an explicit auth method when deploying

dhellmann:
If neither auth method is enabled, exit with an error message.

### DIFF
--- a/tools/deploy.sh
+++ b/tools/deploy.sh
@@ -38,6 +38,9 @@ if [ "${DEPLOY_TLS}" == "true" ]; then
     IRONIC_SCENARIO="${IRONIC_SCENARIO}/tls"
 elif [ "${DEPLOY_BASIC_AUTH}" == "true" ]; then
     BMO_SCENARIO="${BMO_SCENARIO}/default"
+else
+    echo "ERROR: Unsupported deploy scenario. Please set DEPLOY_TLS or DEPLOY_BASIC_AUTH." 1>&2
+    exit 1
 fi
 
 if [ "${DEPLOY_KEEPALIVED}" == "true" ]; then


tools/deploy.sh
```diff
@@ -38,6 +38,9 @@ if [ "${DEPLOY_TLS}" == "true" ]; then
     IRONIC_SCENARIO="${IRONIC_SCENARIO}/tls"
 elif [ "${DEPLOY_BASIC_AUTH}" == "true" ]; then
     BMO_SCENARIO="${BMO_SCENARIO}/default"
+else
```
maelk, line 41: actually this case is acceptable, and is covered by BMO_SCENARIO being set to "${SCRIPTDIR}/config" on line 32. I don't think this should be an error

maelk: if DEPLOY_TLS is false and DEPLOY_BASIC_AUTH too, then we use the /config kustomization

maelk: Maybe we should make it clear in the script instead that we use /config by default ?

dhellmann: OK, great. I always worry when I see if/elif without a final else. :-)

dhellmann: > Maybe we should make it clear in the script instead that we use /config by default ?

That's a good idea. Would we do that by setting a variable here? Maybe consolidating the way we use the auth mode variables to set other variables? Or just printing a message at this point?

maelk: we could set it to "${SCRIPTDIR}/config" again . that would be explicit

kashifest: Thanks @maelk  for the info. I didn't pay attention

dhellmann: The test on lines 28 and 39 are the same, but the outcomes are different. Which is correct?

maelk: I guess this is unclear in the script. We have a kustomization tree as follows :
```
.
├── basic-auth
│   ├── default
│   │   └── kustomization.yaml
│   └── tls
│       └── kustomization.yaml
├── default
│   └── kustomization.yaml
├── kustomization.yaml
└── tls
    └── kustomization.yaml
```
We have multiple scenarios : 

- TLS and basic auth : we need config/basic-auth/tls/kustomization.yaml
- basic auth only: we need config/basic-auth/default/kustomization.yaml
- TLS only : we need config/tls/kustomization.yaml
- neither: we need config/kustomization.yaml

So we first check if we use basic auth, in that case we use config/basic-auth as root, otherwise config/ , then we check if we use TLS. If yes, then we append tls/, if not, we have two cases. if we use basic-auth we have to append default/ , if not, we should not append anything (we are in config/) . 

One way to simplify this logic would be to remove the "default" folder in basic-auth folder and move its content to the basic-auth folder directly, then we would have the same situation with or without basic-auth when not using TLS.

dhellmann: I completely missed that the 2nd set of tests was extending the values set in the first batch. :head-desk:

dhellmann:
We don't need this. The existing logic is fine, and the new error handling case is incorrect.

PR #683, opened 2020-10-14
remove test dependency from docker build

dhellmann:
There doesn't seem to be much point in running all of the lint and
unit tests before building the docker image, and it means we have to
keep the CI system set up to run extra tools in that job. This patch
removes the dependency so `make docker` only builds the image.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -163,7 +163,7 @@ $(CONTROLLER_GEN):
 ## --------------------------------------
 
 .PHONY: docker
-docker: test ## Build the docker image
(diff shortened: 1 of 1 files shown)


zaneb:
Why are we still running Travis at all? All of the jobs except the docker one have duplicates in Prow. Wouldn't it be better to create a Prow job for this (that wouldn't need to go through the Makefile)?
I agree that depending on `test` is overkill, but it kind of makes sense to depend on `generate` so that if one were using it locally, one could be sure that the image was always consistent. We don't want that in CI though.

PR #685, opened 2020-10-15
make logger names more consistent

dhellmann:
Use consistent capitalization and naming conventions for the loggers.

### DIFF
--- a/controllers/metal3.io/baremetalhost_controller_test.go
+++ b/controllers/metal3.io/baremetalhost_controller_test.go
@@ -99,7 +99,7 @@ func newTestReconcilerWithProvisionerFactory(factory provisioner.Factory, initOb
 		Client:             c,
 		Scheme:             scheme.Scheme,
 		ProvisionerFactory: factory,
-		Log:                ctrl.Log.WithName("controller").WithName("baremetalhost").WithName("test"),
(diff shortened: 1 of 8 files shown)


controllers/metal3.io/baremetalhost_controller_test.go
```diff
@@ -99,7 +99,7 @@ func newTestReconcilerWithProvisionerFactory(factory provisioner.Factory, initOb
 		Client:             c,
 		Scheme:             scheme.Scheme,
 		ProvisionerFactory: factory,
-		Log:                ctrl.Log.WithName("controller").WithName("baremetalhost").WithName("test"),
+		Log:                ctrl.Log.WithName("controller").WithName("BareMetalHost").WithName("test"),
```
zaneb, line 102: In main.go it's `controllers` with an s.

Alternatively, on line 752 and in host_config_data_test.go it's 'Test' and then the test name.

No idea what's right, but there is some inconsistency remaining.

dhellmann: OK, I've moved the test log messages to the logger instantiated as part of the `testing.T` instance and I think I'm using actually consistent names for the "production" code loggers.

PR #688, opened 2020-10-19
Add contoller-gen to go.mod

honza:
This way we can avoid the `go get` hack, and track that dependency with
all the other dependencies in the `go.mod` file.

For some unknown reason (to me), this changed some whitespace in one of the generated YAML files. As far as I can tell, the version of the tool that is downloaded is the same (`v0.4.0`), right down to every file.


hack/install-controller-gen.sh
```diff
@@ -2,18 +2,20 @@
 
 set -ex
 
-CONTROLLER_TOOLS_VERSION=${1:-v0.4.0}
-OUTPUT=${2:-${ORIG_DIR}/bin/controller-gen}
+OUTPUT=bin/controller-gen
 
-WORK_DIR=$(mktemp --tmpdir -d controller-gen-XXX)
-trap cleanup EXIT
+if [ -d "vendor" ]
```
dhellmann, line 7: This could use a comment to explain that downstream repos tracking metal3 probably vendor dependencies.

dhellmann (changes requested):
The whitespace changes are a bit odd, but seem innocuous. Perhaps they're caused by a different version of the YAML library being pulled in based on the other dependencies here vs. what would have been pulled in as a direct dependency of controller-gen in a standalone build?

One suggestion about comments inline.

honza:
> Perhaps they're caused by a different version of the YAML library being pulled in based on the other dependencies here vs. what would have been pulled in as a direct dependency of controller-gen in a standalone build?

That's exactly what I thought so I did some digging into the `go mod graph` output for each scenario but came up empty.

honza:
Here are the YAML differences when using `go list -m all` for each scenario. 
```
go modules:

gopkg.in/yaml.v2 v2.3.0
gopkg.in/yaml.v3 v3.0.0-20200313102051-9f266ea9e77c
sigs.k8s.io/yaml v1.2.0

go get:

gopkg.in/yaml.v2 v2.2.8
gopkg.in/yaml.v3 v3.0.0-20190905181640-827449938966
sigs.k8s.io/yaml v1.2.0
```

PR #690, opened 2020-10-21
extend tests for ValidateManagementAccess

dhellmann:
Add a test for when a node already exists and another for when the driver
settings like username and password are different.


pkg/provisioner/ironic/testserver/server.go
```diff
@@ -76,6 +76,20 @@ func (m *MockServer) Response(pattern string, payload string) *MockServer {
 	return m.ResponseWithCode(pattern, payload, http.StatusOK)
 }
 
+// SendJSONResponse marshalls the payload to a JSON object and sends
+// the response using the given writer
+func (m *MockServer) SendJSONResponse(payload interface{}, code int, w http.ResponseWriter, r *http.Request) {
```
andfasano, line 81: Is this a replacement for [ResponseJSON](https://docs.google.com/document/d/1fzHLNz8bLeu0DbO7f20GazbGiqIWKwHzVKaLLdtTsvo/edit#heading=h.s76i9akrmjw9)?

dhellmann: No. ResponseJSON builds a handler with a static response. This formats a response dynamically at the point that it's needed by another handler.

dhellmann: I suppose I could have made both of them use the same code at some layer.

andfasano: Agree, looks like there's a good overlap

pkg/provisioner/ironic/testserver/ironic.go
```diff
@@ -87,6 +90,71 @@ func (m *IronicMock) WithNode(node nodes.Node) *IronicMock {
 	return m
 }
 
+// WithWritableNode configures the server with a valid response for /v1/nodes
+func (m *IronicMock) WithWritableNode(node *nodes.Node) *IronicMock {
```
andfasano, line 94: Probably it would be better to split into different methods the GET/PATCH for /v1/nodes, as they belong to different api calls (and for the get we already have a working implementation)

andfasano: I just realized that I need a similar feature while testing the `Delete` method, let me workout a proposal in a separate PR. For the moment I think it's ok to proceed as it is.

dhellmann:
@andfasano I have included some more refactoring of the test server code, to shorten WithNode() to just Node(). I also changed the new NodeUpdate() to take a before and after argument instead of having the method try to reproduce what Ironic would do to the data structure. The callback argument can be used to verify that the right arguments are being passed to the server, which is really what we want to test.

I thought I understood what you had in mind with registering the path and method together, but I wasn't able to figure out how to apply that to the create or update tests. In the create case, I don't have a static response to provide and I need to see the data that is passed in. In the update case I can provide static responses, but still need to see the input data in the test. I want to avoid turning the test server into a complicated web framework, so it seemed easier to just have 1 handler and check the method. Let me know what you think.

andfasano:
> @andfasano I have included some more refactoring of the test server code, to shorten WithNode() to just Node(). I also changed the new NodeUpdate() to take a before and after argument instead of having the method try to reproduce what Ironic would do to the data structure. The callback argument can be used to verify that the right arguments are being passed to the server, which is really what we want to test.
> 
> I thought I understood what you had in mind with registering the path and method together, but I wasn't able to figure out how to apply that to the create or update tests. In the create case, I don't have a static response to provide and I need to see the data that is passed in. In the update case I can provide static responses, but still need to see the input data in the test. I want to avoid turning the test server into a complicated web framework, so it seemed easier to just have 1 handler and check the method. Let me know what you think.

Setting an expectation on incoming requests it's really a valid and it could be useful in general, and agree that the server must not be complicated (but just recording requests and sending answers if programmed for). My only concern is to avoid duplications, and trying to keep all the infrastructure code as much as possible at the MockServer level.
I've prepared a PR https://github.com/dhellmann/baremetal-operator/pull/2 for describing that, implementing the design so that the test code can easily retrieve a request recorded during the execution (and in this case the IronicMock serves just as a layer to improve the ergonomics for the test author). I think that in such way the test code could be simplified, as we also reuse default responses.

andfasano:
Current version looks definitely improved, changes are ok for me

PR #691, opened 2020-10-22
Add kustomize tool to go.mod

honza:
Just like we did in #688 with controller-gen


config/render/capm3.yaml
```diff
@@ -837,6 +837,16 @@ spec:
         control-plane: controller-manager
     spec:
       containers:
+      - args:
```
dhellmann, line 840: What happened here? Is this a generated file that changed because of a change in kustomize?

dhellmann (commented):
This looks OK to me, with one question inline.

We should get @furkatgofurov7 and @kashifest to take a look, too.

config/render/capm3.yaml
honza: Yes.

dhellmann: OK, thanks

furkatgofurov7: Yes, that's right.

PR #692, opened 2020-10-23
WIP: Fix Kustomize

kashifest:


furkatgofurov7 (commented):
I think we have to put back `tools` in place [like it was before
](https://github.com/metal3-io/baremetal-operator/pull/691/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L24)

Because, CI is failing with:
`make[1]: *** No rule to make target 'tools/bin/kustomize'.  Stop.`

kashifest:
@honza is taking care of this in a separate PR
/close

PR #695, opened 2020-10-23
add health check endpoint

### LINKED ISSUE #656: add health end point
Add options to configure a health endpoint as in https://github.com/metal3-io/cluster-api-provider-metal3/blob/master/main.go#L81

dhellmann:
Add a command line option to specify the address for a health-check
endpoint and use the address to configure that endpoint in the
controller.

Fixes #656

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -39,3 +39,9 @@ spec:
             cpu: 100m
             memory: 20Mi
       terminationGracePeriodSeconds: 10
+      livenessProbe:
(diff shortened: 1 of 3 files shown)


maelk:
Would it make sense to add the liveness / readiness probe in the deployment (config/manager/manager.yaml) in this PR too ?

dhellmann:
> Would it make sense to add the liveness / readiness probe in the deployment (config/manager/manager.yaml) in this PR too ?

Good point, done.

PR #699, opened 2020-10-28
Remove main.go path in tilt-provider.json file

jan-est:
This PR removes the main-go path in tilt-provider.json file. This enable bmo to run properly when [using Tilt](https://github.com/metal3-io/cluster-api-provider-metal3/blob/master/docs/dev-setup.md#tilt-development-environment)

### DIFF
--- a/tilt-provider.json
+++ b/tilt-provider.json
@@ -2,10 +2,9 @@
     "name": "metal3-bmo",
     "config": {
         "kustomize_config": false,
-        "go_main": "cmd/manager/main.go",
         "image": "quay.io/metal3-io/baremetal-operator",
         "live_reload_deps": [
-            "cmd", "pkg", "go.mod", "go.sum"
+            "apis", "cmd", "pkg", "config", "controllers", "go.mod", "go.sum", "main.go"
         ]
     }
 }


Jaakko-Os (commented):
looks good to me

tilt-provider.json
```diff
@@ -2,10 +2,9 @@
     "name": "metal3-bmo",
     "config": {
         "kustomize_config": false,
-        "go_main": "cmd/manager/main.go",
         "image": "quay.io/metal3-io/baremetal-operator",
         "live_reload_deps": [
             "cmd", "pkg", "go.mod", "go.sum"
```
maelk, line 7: this should be updated with the new structure

tilt-provider.json
```diff
@@ -2,10 +2,9 @@
     "name": "metal3-bmo",
     "config": {
         "kustomize_config": false,
-        "go_main": "cmd/manager/main.go",
```
maelk, line 5: maybe change it to main.go only ?

jan-est: I think we should not set "go_main" at all. For example if you are compare it to tilt-provider.json of [capm3](https://github.com/metal3-io/cluster-api-provider-metal3/blob/master/tilt-provider.json) or any other providers, they are not normally define "go_main". And it is not required in [CAPI book tilt-provider.json](https://cluster-api.sigs.k8s.io/developer/tilt.html#tilt-providerjson)

maelk: ok

maelk (approved):

PR #704, opened 2020-10-29
Remove limits and requests from manager manifest

### LINKED ISSUE #703: Limits in the manager manifests are too low
The limits in the manager manifests are too low and cause BMO to crash when it handles a certain number of nodes. We have to remove limits and requests.

kashifest:
Fixes #703 .

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -31,13 +31,6 @@ spec:
           - configMapRef:
               name: ironic
         name: manager
-        resources:
-          limits:
-            cpu: 100m
-            memory: 30Mi
-          requests:
-            cpu: 100m
-            memory: 20Mi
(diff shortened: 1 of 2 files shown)


dhellmann:
I suspect the `generate` job is failing because of a difference in the tool versions between what you have locally and what CI uses. You could try removing tools/bin and running `make generate manifests` again to ensure the right versions of the tools are built.

kashifest:
> I suspect the `generate` job is failing because of a difference in the tool versions between what you have locally and what CI uses. You could try removing tools/bin and running `make generate manifests` again to ensure the right versions of the tools are built.

You are right! Thanks

fmuyassarov (commented):
lgtm

furkatgofurov7 (commented):
Thanks for the patch!

/lgtm

PR #705, opened 2020-10-29
Handle different responses by method on MockServer

andfasano:
This PR introduces the possibility to specify different handlers for different HTTP method on MockServer.

It could help also for #690


pkg/provisioner/ironic/testserver/ironic.go
```diff
@@ -76,23 +76,46 @@ func (m *IronicMock) WithDrivers() *IronicMock {
...
 
-// WithNodeStatesProvision configures the server with a valid response for /v1/nodes/<node>/states/provision
-func (m *IronicMock) WithNodeStatesProvision(nodeUUID string) *IronicMock {
-	m.ResponseWithCode("/v1/nodes/"+nodeUUID+"/states/provision", "{}", http.StatusAccepted)
+// WithNode configures the server with a valid response for [GET] /v1/nodes
+func (m *IronicMock) WithNode(node nodes.Node) *IronicMock {
+	return m.withNode(node, http.MethodGet)
+}
+
+// WithNodeUpdate configures the server with a valid response for [PATCH] /v1/nodes
```
dhellmann, line 99: Is it really valid, though? It's going to return the node passed to the method, not one changed by a patch, right?

pkg/provisioner/ironic/testserver/server.go
```diff
@@ -76,16 +86,59 @@ func (m *MockServer) Response(pattern string, payload string) *MockServer {
 	return m.ResponseWithCode(pattern, payload, http.StatusOK)
 }
 
+func (m *MockServer) buildHandler(pattern string) func(http.ResponseWriter, *http.Request) {
+
+	handler := func(w http.ResponseWriter, r *http.Request) {
+
+		mh, ok := m.responsesByMethod[r.URL.String()]
```
dhellmann, line 93: Does this value of `ok` need to be checked?

andfasano: Fixed, no need by design

pkg/provisioner/ironic/testserver/ironic.go
andfasano: I didn't want to change the previous semantic in this commit, probably we could refine better some of the update methods.

dhellmann: So we need this method because some of the existing API calls are not using GET and now the default for `WithNode()` only supports GET?

andfasano: Exactly. I think in this way we could mimic in a more realistic way the ironic/inspector API where for distinct HTTP methods there could be different behaviors (from our point of view, a different json/code response). I also hit the limits of the previous implementation while developing the tests for the Provisioner.Delete() func.

pkg/provisioner/ironic/testserver/ironic.go
```diff
@@ -76,23 +76,46 @@ func (m *IronicMock) WithDrivers() *IronicMock {
 	return m
 }
 
-// WithNode configures the server with a valid response for /v1/nodes
-func (m *IronicMock) WithNode(node nodes.Node) *IronicMock {
+func (m *IronicMock) buildURL(url string, method string) string {
+	return fmt.Sprintf("%s:%s", url, method)
+}
+
+func (m *IronicMock) withNode(node nodes.Node, method string) *IronicMock {
```
EgorLu, line 83: As this is a private method, maybe we can rename it into something longer but easier to differentiate from `WithNode`?

EgorLu: Unless it's a convention I'm not aware of.

pkg/provisioner/ironic/testserver/ironic.go
dhellmann: OK, the concern I have is that we're not actually testing that the update does anything. That's what I was trying to do in #690 , although I can see there being a concern with re-implementing the patch logic there if it gets much more complicated.

What do you see for the future of this method? Would I rebase 690 on these changes and replace the implementation of this method with something more dynamic?

andfasano: I think it would be really fine. I had to provide a basic backward compatible implementation for it due the infrastructural changes to support the HTTP methods dispatching, otherwise the tests failed due a missing handler (for the PATCH).
So, the current implementation for NodeUpdate does not change the previous behavior, and it will allow to merge this PR and then rework it as needed.

pkg/provisioner/ironic/testserver/ironic.go
andfasano: No particular convention. Do you have any specific idea for the name?

pkg/provisioner/ironic/testserver/ironic.go
dhellmann: Sure, I understand. I'm trying to work out what the next step might be, after we take this patch. I think I see what that would be, now.

pkg/provisioner/ironic/testserver/ironic.go
EgorLu: I think that I would leave the generic method as it is but rename the specific/partial ones to `WithGetNode`, `WithPatchNode` and so on. Though I'm not sure how compatible it is.

I must agree that choosing a name here is a bit difficult, since the functions do HTTP requests internally, but there's an abstraction level for the user.

Or maybe something like `genericWithNode`? What do you think?

pkg/provisioner/ironic/adopt_test.go
```diff
@@ -39,7 +39,7 @@ func TestAdopt(t *testing.T) {
 			ironic: testserver.NewIronic(t).Ready().WithNode(nodes.Node{
 				ProvisionState: string(nodes.Manageable),
 				UUID:           nodeUUID,
-			}).WithNodeStatesProvision(nodeUUID),
+			}).WithNodeStatesProvision(nodeUUID).WithNodeStatesProvisionUpdate(nodeUUID),
```
dhellmann, line 42: Looking through this PR again, I think we want to be careful to not make the mock server always try to act exactly like Ironic. Most of the time we're going to want to keep the setup small and easy, since if we're only testing one part of the provisioner at a time.

So, I think we do want something like what you're adding here to support different methods for a node URL, but I don't think we want that to be the only way to set up the server and I don't think we want it to be the default. We haven't needed separate "update" calls until now. Let's see if we can extend the API in a way that doesn't require us to use them everywhere, or modify existing tests.

Maybe that means changing WithNode() and WithNodeStatesProvision() to register handlers for multiple methods? Tests that need separate methods could then call something other than WithNode() or WithNodeStatesProvision(), keeping the complication local to the test.

andfasano: On a second thought, I considered a mistake the fact that we didn't need the update calls until now: there was just a single handler registered for the same pattern that incidentally worked well for both the get/update cases.

From my point of view, the goal in a test case is to program the Ironic mock to return the desired values given some specific API calls, and performing clearly this declaration of intents in the the case could just make more readable the test itself.

I also agree on your point, and sometimes the underlying interaction operator <-> server can be pretty complex, and maybe in a test we're interested to "tweak" just few of the overall calls, relying on the default behavior for the others.
So, we could define a "default handler" (as we did initially for the catch all case), eventually configurable, to be used when the test author didn't set explicitly an handler (instead of failing as per my initial implementation).
At the risk of some magic, this anyhow will not force the author to define all the mock calls, but those ones just relevant for the given case, removing some of the complication.

dhellmann: I like the idea of the default handler. It would help us balance the need to be explicit when configuring the mock server with the need to be concise in the tests to keep them readable.

andfasano: Ok, I've added the default handler and simplified the test code. I've added a new method `IronicMock.WithDefaultResponses()` so that the test author could enable the feature on a test-by-test case: it's not mandatory to describe every single call - but just those ones that are relevant for the test case.
The default handler accepts patterns with variables, and variables could be used in the payload to craft a little bit better the response (see MockServer.AddDefaultResponse()'.
Looks like that now with this approach we could be in the position to specify easily a default behavior, and override it on a case bases (and even for method) where needed.

pkg/provisioner/ironic/testserver/ironic.go
```diff
@@ -25,12 +25,25 @@ type IronicMock struct {
...
 // NewIronic builds an ironic mock server
 func NewIronic(t *testing.T) *IronicMock {
+
 	return &IronicMock{
 		MockServer:   New(t, "ironic"),
 		CreatedNodes: nil,
 	}
 }
 
+// WithDefaultResponses sets a valid answer for all the API calls
```
dhellmann, line 35: Maybe this could call Ready() to set up those responses as well?

pkg/provisioner/ironic/testserver/server.go
```diff
@@ -115,13 +176,75 @@ func (m *MockServer) ErrorResponse(pattern string, errorCode int) *MockServer {
...
+}
+
+// AddDefaultResponse adds a default response for the specified pattern/method.
+// It is possible to use variables in the pattern using curly braces, ie `/v1/nodes/{id}/power`
+// Pattern variables can be reused in the payload, so that they will be substitued with the actual value when sending the response
+// If httpMethod is empty, the response will be applied for any method
+func (m *MockServer) AddDefaultResponse(patternWithVars string, httpMethod string, code int, payload string) *MockServer {
+
+	pattern := "^" + regexp.MustCompile("{(.[^}]*)}").ReplaceAllString(patternWithVars, "(?P<$1>.[^/]*)") + "$"
+	m.t.Logf("%s: adding default response for %s -> {%d, %s}", m.name, patternWithVars, code, payload)
```
dhellmann, line 204: It would be nice to see that new pattern logged here, too, for debugging.

pkg/provisioner/ironic/testserver/server.go
```diff
@@ -115,13 +176,75 @@ func (m *MockServer) ErrorResponse(pattern string, errorCode int) *MockServer {
...
+}
+
+func (m *MockServer) defaultHandler(w http.ResponseWriter, r *http.Request) {
+
+	url := r.URL.String()
+	method := r.Method
+
+	for _, response := range m.defaultResponses {
+		if response.method == "" || response.method == method {
+			re := regexp.MustCompile(response.pattern)
```
dhellmann, line 226: Why not store the compiled pattern?

pkg/provisioner/ironic/testserver/server.go
```diff
@@ -115,13 +176,75 @@ func (m *MockServer) ErrorResponse(pattern string, errorCode int) *MockServer {
...
+			match := re.FindStringSubmatch(url)
+
+			if match != nil {
```
dhellmann, line 229: Nit: I would have reversed this check and used a `continue`.

dhellmann (approved):
This looks good. I have a few small comments, but those can be addressed in separate PRs.

pkg/provisioner/ironic/testserver/ironic.go
andfasano: maybe something like `withNodeBase`, being a base technical method (but honestly don't have better ideas)

pkg/provisioner/ironic/testserver/ironic.go
andfasano: Sounds reasonable

pkg/provisioner/ironic/testserver/server.go
andfasano: Ok

pkg/provisioner/ironic/testserver/server.go
andfasano: laziness :) will fix

dhellmann:
This improves on what we have. Let's merge it and keep iterating.

/lgtm

PR #706, opened 2020-10-29
Update local ironic documentation with TLS information

kashifest:
This PR updates the documentation to add TLS variables that need to be exported in case the local ironic containers are expected to run with TLS enabled.


maelk (approved):

maelk:
what about adding something about basic auth too ?

kashifest:
> what about adding something about basic auth too ?

Good point. I am adding it.

kashifest:
> what about adding something about basic auth too ?

Done

furkatgofurov7 (commented):
It is good to be more explicit and provide information to users. 

/lgtm

Issue #709, opened 2020-11-04
Too many connections towards ironic left open


kashifest:
Current way of establishing http transport connection from BMO towards ironic(api, inspector etc.) is leaving too many connections open/idle. Here is a typical output from the host

```
netstat -pantu | grep ESTABLISHED
(Not all processes could be identified, non-owned process info
 will not be shown, you would have to be root to see it all.)
tcp        0      0 127.0.0.1:34470         127.0.0.1:3306          ESTABLISHED -                   
tcp        0      0 127.0.0.1:44210         127.0.0.1:3306          ESTABLISHED -                   
tcp        0      0 127.0.0.1:34192         127.0.0.1:3306          ESTABLISHED -                   
tcp        0      0 127.0.0.1:36032         127.0.0.1:3306          ESTABLISHED -                   
tcp        0      0 127.0.0.1:37862         127.0.0.1:3306          ESTABLISHED -                   
tcp        0      0 192.168.111.1:42376     192.168.111.1:22        ESTABLISHED -                   
tcp        0      0 172.22.0.1:37238        172.22.0.2:6385         ESTABLISHED -                   
tcp        0      0 127.0.0.1:47332         127.0.0.1:3306          ESTABLISHED -                   
tcp        0      0 192.168.111.1:42592     192.168.111.1:22        ESTABLISHED -                   
tcp        0      0 192.168.111.1:22        192.168.111.1:42598     ESTABLISHED -                   
tcp        0      0 192.168.111.1:42870     192.168.111.1:22        ESTABLISHED -                   
tcp        0      0 192.168.111.1:22        192.168.111.1:42870     ESTABLISHED -                   
tcp        0      0 127.0.0.1:35430         127.0.0.1:3306          ESTABLISHED -                   
tcp        0      0 192.168.111.1:22        192.168.111.1:42378     ESTABLISHED -                   
tcp        0      0 192.168.111.1:22        192.168.111.1:42592     ESTABLISHED -                   
tcp        0      0 127.0.0.1:35082         127.0.0.1:3306          ESTABLISHED -                   
tcp        0      0 127.0.0.1:50364         127.0.0.1:3306          ESTABLISHED -                   
tcp        0      0 127.0.0.1:43588         127.0.0.1:3306          ESTABLISHED -                   
tcp        0      0 192.168.111.1:42598     192.168.111.1:22        ESTABLISHED -                   
[...]

PR #707, opened 2020-11-03
Fix TLS transport connection establishment

kashifest:
This PR fixes #709 
To be specific, this PR disables keepalive while opening http connections, so that the socket connections don't remain open.

PR #710, opened 2020-11-06
Avoid recreating new client for ironic & inspector

kashifest:
This PR comes as an alternative to #707 and as a solution to #709.
It tries to avoid recreating new client for ironic & inspector as suggested by @zaneb

### DIFF
--- a/pkg/provisioner/ironic/clients/client.go
+++ b/pkg/provisioner/ironic/clients/client.go
@@ -15,6 +15,8 @@ import (
 )
 
 var tlsConnectionTimeout = time.Second * 30
+var inspectorClient *gophercloud.ServiceClient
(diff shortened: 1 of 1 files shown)


kashifest:
/assign @maelk 
/cc @zaneb @dhellmann 
@zaneb how does this PR look as a solution?

maelk:
/test-centos-integration
this looks ok to me, just verifying centos to check when ironic runs as a pod

dhellmann:
/hold

I don't like caching the connection at such a low level, because that hides the cache under something the tests will call. I've proposed #712 as an alternative that caches the clients in the provisioner package instead.

kashifest:
Closing due to #712 
/close

PR #711, opened 2020-11-06
start adding unit tests for Ironic.Delete

andfasano:
Add new test cases for the Ironic provisioner.Delete()

### DIFF
--- a/pkg/provisioner/ironic/delete_test.go
+++ b/pkg/provisioner/ironic/delete_test.go
@@ -1,0 +1,192 @@
+package ironic
+
+import (
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/gophercloud/gophercloud/openstack/baremetal/v1/nodes"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/metal3-io/baremetal-operator/pkg/bmc"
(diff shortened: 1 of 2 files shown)


andfasano:
/hold
Currently based on #690 , waiting for the merge

pkg/provisioner/ironic/delete_test.go
```diff
@@ -0,0 +1,192 @@
...
+		{
+			name: "available-node",
+			ironic: testserver.NewIronic(t).WithDefaultResponses().Node(
+				nodes.Node{
+					UUID:           nodeUUID,
+					ProvisionState: "available",
+				},
+			),
+			expectedDirty:        true,
+			expectedRequestAfter: 10,
```
dhellmann, line 100: Should we refer to the variable that controls that timeout, instead of using a hard-coded value? That way if we adjust the timeout we won't have to modify the tests.

dhellmann (approved):
This looks good to me. I have one small question inline, but not enough to block it.

pkg/provisioner/ironic/delete_test.go
andfasano: Fixed

PR #712, opened 2020-11-06
cache the ironic api clients

dhellmann:
Share the clients client instead of making a new one every time we
create a provisioner (during every call to Reconcile) so we can reuse
the transport and open connections.

This is an alternative to #710


kashifest:
This looks good @dhellmann . I have no issues if the issue #709  is solved like this. I will close the other two alternative PRs #710 and #707. Thanks for taking care of this.
/lgtm

PR #713, opened 2020-11-09
Remove finalizer when deleting a host with a malformed BMC address

kirankt:
Sometimes when a host is created with a malformed BMC address, trying to immediately delete it will cause in an error and the deletion hangs. This is because a valid driver for this URL cannot be found and the code to remove the finalizer is not run. This PR ensures that the finalizer is removed when the URL is malformed and the deletion timestamp is non zero.

### DIFF
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -210,6 +210,18 @@ func (r *BareMetalHostReconciler) Reconcile(request ctrl.Request) (result ctrl.R
 	}
 	prov, err := r.ProvisionerFactory(host, *bmcCreds, info.publishEvent)
 	if err != nil {
+		if !info.host.DeletionTimestamp.IsZero() {
+			// Remove the finalizer, if present
+			if hostHasFinalizer(host) {
+				info.log.Info(
+					"removing finalizer",
+					"timestamp", info.host.DeletionTimestamp,
+				)
+
+				host.Finalizers = utils.FilterStringFromList(
+					host.Finalizers, metal3v1alpha1.BareMetalHostFinalizer)
+			}
+		}
 		return ctrl.Result{}, errors.Wrap(err, "failed to create provisioner")
 	}
 


controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -191,6 +191,12 @@ func (r *BareMetalHostReconciler) Reconcile(request ctrl.Request) (result ctrl.R
 				// If we are in the process of deletion, try with empty credentials
 				bmcCreds = &bmc.Credentials{}
 				bmcCredsSecret = &corev1.Secret{}
+
+				// Remove the finalizer, if present
+				if hostHasFinalizer(host) {
+					host.Finalizers = utils.FilterStringFromList(
+						host.Finalizers, metal3v1alpha1.BareMetalHostFinalizer)
```
dhellmann, line 198: Using empty credentials should have allowed us to make it to `actionDelete()` where this is happening. What does the provisioner return that prevents that from happening?

kirankt: There is a bz for this bug (https://bugzilla.redhat.com/show_bug.cgi?id=1891301). When the user created the host, they specified the URL with special characters e.g. '<redfish:// ....>' and then tried to delete the host, which causes the deletion process to hang. 

Frankly, I'm not sure whether this is the correct place to remove the finalizer. Also, not sure why the malformed URL didn't get flagged.

zaneb: The log message says "failed to create provisioner", so I believe it's line 213 that is failing.

zaneb:
This looks like it should solve the problem now; let's rebase and squash it too.
A log message is essential for this I think.
It would be safer if we checked the the Provisioning state was empty (i.e. we have *never* been able to create a provisioner for this Host, vs. we created a Host and provisioned it and now somebody has changed the URL).
It occurs to me that if all errors coming out of ProvisionerFactory are deterministic then we were doing the wrong thing already by returning the error (if only the user can fix it by updating the Host then we don't really want to reconcile again). If that's not the case, however, then we need to identify errors that are caused by user input specifically, and limit this change to those.

kirankt:
/retitle Remove finalizer when deleting a host with a malformed BMC address

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -210,6 +210,18 @@ func (r *BareMetalHostReconciler) Reconcile(request ctrl.Request) (result ctrl.R
 	}
 	prov, err := r.ProvisionerFactory(host, *bmcCreds, info.publishEvent)
 	if err != nil {
+		if !info.host.DeletionTimestamp.IsZero() {
```
maelk, line 213: That would catch much more than just a malformed BMC address, no ? Basically any errors (even temporary) in the provisioner would lead to BMH deletion, is this a desired behaviour ?

zaneb:
I think the bug is fixed by https://github.com/metal3-io/baremetal-operator/pull/820/commits/beea4d0ead807a8f19b38d538db3502ee3504b97. In any event Andrea has opened a new PR, so we can close this one.
/close

PR #714, opened 2020-11-10
Handle AdoptFail state in Ironic Adopt 

### LINKED ISSUE #697: Retry when node in "Adopt Failed" state
Currently, we are not retrying when ironic fails to adopt and puts the node in 'adopt failed' state:  https://github.com/metal3-io/baremetal-operator/blob/master/pkg/provisioner/ironic/ironic.go#L1001

As [checked](https://docs.openstack.org/ironic/latest/_images/states.svg), we could try to retry to get a node out of "adopt failed" by:

-  sending manage command(via API) ==> manageable
- sending adopt command (via API) ==> adopting(retry adoption)

furkatgofurov7:
Fixes #697

### DIFF
--- a/pkg/provisioner/ironic/adopt_test.go
+++ b/pkg/provisioner/ironic/adopt_test.go
@@ -73,6 +73,7 @@ func TestAdopt(t *testing.T) {
 
 			expectedDirty:        false,
 			expectedRequestAfter: 0,
+			expectedError:        true,
 		},
 	}
 
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -996,7 +996,7 @@ func (p *ironicProvisioner) Adopt() (result provisioner.Result, err error) {
 	var ironicNode *nodes.Node
 
 	if ironicNode, err = p.findExistingHost(); err != nil {
-		err = errors.Wrap(err, "could not find host to adpot")
+		err = errors.Wrap(err, "could not find host to adopt")
 		return
 	}
 	if ironicNode == nil {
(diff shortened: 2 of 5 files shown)


pkg/provisioner/ironic/ironic.go
```diff
@@ -1025,8 +1025,20 @@ func (p *ironicProvisioner) Adopt() (result provisioner.Result, err error) {
 		result.RequeueAfter = provisionRequeueDelay
 		result.Dirty = true
 	case nodes.AdoptFail:
-		result.ErrorMessage = fmt.Sprintf("Host adoption failed: %s",
-			ironicNode.LastError)
+		// The node is in AdoptFail, the way to get node
+		// out of this state is by target manage to go
+		// to manageable. Once the node is in manageable
+		// it will move to target adopt.
+		result, err = p.changeNodeProvisionState(
```
maelk, line 1032: I'm concerned that if this fails we will requeue for the backoff delay, and when coming back here, we will still be in AdoptFail, so we will go to manageable, and requeue again for the backoff time, making it in effect two successive backoffs. In the function that calls this function, can we requeue immediately if err is not nil ?

pkg/provisioner/ironic/adopt_test.go
```diff
@@ -73,6 +73,7 @@ func TestAdopt(t *testing.T) {
 
 			expectedDirty:        false,
 			expectedRequestAfter: 0,
+			expectedError:        true,
 		},
 	}
```
jan-est, line 79: Do we still need a test case for retry and to make sure that Node will move from AdoptFail state back to manageable?

pkg/provisioner/ironic/ironic.go
furkatgofurov7: So, do you mean, we could call requeue [here](https://github.com/metal3-io/baremetal-operator/blob/master/controllers/metal3.io/baremetalhost_controller.go#L691)?

zaneb:
What do you think of https://github.com/zaneb/baremetal-operator/commit/36fd3fb5a533ee9b7d6b5d92a3ab2c7d7ff01d5c?

maelk:
@zaneb your commit looks good. @furkatgofurov7 would you be ok with it ?

furkatgofurov7:
@maelk sure, I am okay with that. @zaneb will you open a PR then with your commit, I will close this one? Thanks!

zaneb:
Yep, it's part of a series that I still need to test, but will open a PR as soon as that's ready.

furkatgofurov7:
> Yep, it's part of a series that I still need to test, but will open a PR as soon as that's ready.

@zaneb thanks! Closing this, as it will be covered by another patch. 

/close

PR #715, opened 2020-11-10
Look up existing nodes by MAC address

honza:
If an existing Ironic node, and a BMH share a MAC address, link them together.

Also includes some new functions in the test server for testing port logic

### DIFF
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -230,6 +230,31 @@ func (p *ironicProvisioner) validateNode(ironicNode *nodes.Node) (errorMessage s
 	return "", nil
 }
 
+func (p *ironicProvisioner) listAllPorts(address string) ([]ports.Port, error) {
+	var allPorts []ports.Port
+
+	opts := ports.ListOpts{}
+
+	if address != "" {
+		opts.Address = address
+	}
+
+	pager := ports.List(p.client, opts)
+
+	if pager.Err != nil {
+		return allPorts, pager.Err
+	}
+
+	allPages, err := pager.AllPages()
+
+	if err != nil {
+		return allPorts, err
+	}
+
+	return ports.ExtractPorts(allPages)
+
+}
+
 // Look for an existing registration for the host in Ironic.
 func (p *ironicProvisioner) findExistingHost() (ironicNode *nodes.Node, err error) {
 	// Try to load the node by UUID
@@ -258,11 +283,48 @@ func (p *ironicProvisioner) findExistingHost() (ironicNode *nodes.Node, err erro
 		p.log.Info("found existing node by name")
 		return ironicNode, nil
 	case gophercloud.ErrDefault404:
-		return nil, nil
+		p.log.Info(
+			fmt.Sprintf("node with name %s doesn't exist", p.host.Name))
 	default:
 		return nil, errors.Wrap(err,
 			fmt.Sprintf("failed to find node by name %s", p.host.Name))
 	}
+
+	// Try to load the node by port address
+	p.log.Info("looking for existing node by MAC", "MAC", p.host.Spec.BootMACAddress)
+	allPorts, err := p.listAllPorts(p.host.Spec.BootMACAddress)
+
+	if err != nil {
+		p.log.Info("failed to find an existing port with address", "MAC", p.host.Spec.BootMACAddress)
+		return nil, nil
+	}
+
+	if len(allPorts) > 0 {
+		nodeUUID := allPorts[0].NodeUUID
+		ironicNode, err = nodes.Get(p.client, nodeUUID).Extract()
+		switch err.(type) {
+		case nil:
+			p.log.Info("found existing node by ID")
+
+			// If the node has a name, this means we didn't find it above.
+			if ironicNode.Name != "" {
+				return nil, errors.New(fmt.Sprint("node found by MAC but has a name: ", ironicNode.Name))
+			}
+
+			return ironicNode, nil
+		case gophercloud.ErrDefault404:
+			return nil, errors.Wrap(err,
+				fmt.Sprintf("port exists but linked node doesn't %s", nodeUUID))
+		default:
+			return nil, errors.Wrap(err,
+				fmt.Sprintf("port exists but failed to find linked node by ID %s", nodeUUID))
+		}
+	} else {
+		p.log.Info("port with address doesn't exist", "MAC", p.host.Spec.BootMACAddress)
+	}
+
+	return nil, nil
+
 }
 
 // ValidateManagementAccess registers the host with the provisioning
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -492,6 +492,29 @@ func (p *ironicProvisioner) ValidateManagementAccess(credentialsChanged bool) (r
 			p.log.Info("setting provisioning id", "ID", p.status.ID)
 		}
 
+		if ironicNode.Name == "" {
+			updates := nodes.UpdateOpts{
+				nodes.UpdateOperation{
+					Op:    nodes.ReplaceOp,
+					Path:  "/name",
+					Value: p.host.Name,
+				},
+			}
+			ironicNode, err = nodes.Update(p.client, ironicNode.UUID, updates).Extract()
+			switch err.(type) {
+			case nil:
+			case gophercloud.ErrDefault409:
+				p.log.Info("could not update ironic node name, busy")
+				result.Dirty = true
+				result.RequeueAfter = provisionRequeueDelay
+				return result, nil
+			default:
+				return result, errors.Wrap(err, "failed to update ironc node name")
+			}
+			p.log.Info("updated ironic node name")
+
+		}
+
 		// Look for the case where we previously enrolled this node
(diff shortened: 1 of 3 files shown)


pkg/provisioner/ironic/testserver/ironic.go
```diff
@@ -198,3 +200,29 @@ func (m *IronicMock) WithNodeValidate(nodeUUID string) *IronicMock {
...
+	m.ResponseJSON(m.buildURL("/v1/ports?address="+address, method), resp)
+
+	return m
+}
+
+// WithPort configures the server with a valid response for
+//    [GET] /v1/nodes/<node uuid>/ports
+//    [GET] /v1/ports
+//    [GET] /v1/ports?=address=<node uuid>
+func (m *IronicMock) WithPort(port ports.Port) *IronicMock {
```
dhellmann, line 226: #690 renames WithNode() to just Node(), so I think this can be just Port().

pkg/provisioner/ironic/validatemanagementaccess_test.go
```diff
@@ -95,3 +96,43 @@ func TestValidateManagementAccessCreateNode(t *testing.T) {
 	assert.NotEqual(t, "", host.Status.Provisioning.ID)
 	assert.Equal(t, ironic.CreatedNodes[0].UUID, host.Status.Provisioning.ID)
 }
+
+func TestValidateManagementAccessLinkExistingIronicNodeByMAC(t *testing.T) {
```
dhellmann, line 100: It would be good to have another test for the failure path by explicitly creating a port with the wrong UUID as well.

pkg/provisioner/ironic/validatemanagementaccess_test.go
```diff
@@ -95,3 +96,79 @@ func TestValidateManagementAccessCreateNode(t *testing.T) {
...
+		t.Fatalf("error from ValidateManagementAccess: %s", err)
+	}
+	assert.Equal(t, "", result.ErrorMessage)
+	assert.NotEqual(t, "", host.Status.Provisioning.ID)
+
+	// No new nodes should have been created
+	assert.Equal(t, len(ironic.CreatedNodes), 0)
+}
+
+func TestValidateManagementAccessExistingPortWithWrongUUID(t *testing.T) {
```
dhellmann, line 140: Sorry, I left a confused comment last time. What we want is to test cases where there is a port, but the port doesn't match what we're using to look it up. I said the UUID, but I should have said MAC.

The case you have here is another good one, but I would expect to have Ironic report an error because there is already a port using the MAC but that is not associated with our node. Therefore when we try to create a new node, it would fail because a port with that MAC already exists.

honza: It looks like the `findExistingNode` method has some overloaded semantics that need to be fixed.

dhellmann: The test calls `ValidateManagementAccess()`, which does both the lookup and the create. I suspect it's working because our mock server implementation is incomplete so it's not reporting the error that ironic would.

dhellmann: Oh, duh, and now I see that on 170 you're calling `assert.Error()` so the test *is* expecting an error. Do we know it's the right error?

pkg/provisioner/ironic/ironic.go
```diff
@@ -230,6 +230,25 @@ func (p *ironicProvisioner) validateNode(ironicNode *nodes.Node) (errorMessage s
...
+func (p *ironicProvisioner) listAllPorts(address string) ([]ports.Port, error) {
+	var allPorts []ports.Port
+
+	opts := ports.ListOpts{}
+
+	if address != "" {
+		opts.Address = address
+	}
+
+	allPages, err := ports.List(p.client, opts).AllPages()
```
andfasano, line 242: List returns a pagination.Pager that could have an error (pager.Err), it could be nice to check it before gettings all the pages

pkg/provisioner/ironic/ironic.go
```diff
@@ -250,6 +269,23 @@ func (p *ironicProvisioner) findExistingHost() (ironicNode *nodes.Node, err erro
 		}
 	}
 
+	// Try to load the node by port
+	p.log.Info("looking for existing node by MAC", "MAC", p.host.Spec.BootMACAddress)
+	allPorts, err := p.listAllPorts(p.host.Spec.BootMACAddress)
```
andfasano, line 274: Minor point, but maybe all this block could be moved into a `checkExistingNodeByMac` method, and maybe `listAllPorts` could be merged within (in any case it's using the `host` member, that could be accessed directly from within the dunc

pkg/provisioner/ironic/testserver/ironic.go
```diff
@@ -198,3 +200,25 @@ func (m *IronicMock) WithNodeValidate(nodeUUID string) *IronicMock {
...
+	}
+
+	resp := map[string][]ports.Port{
+		"ports": {port},
+	}
+
+	address := url.QueryEscape(port.Address)
+
+	m.ResponseJSON(m.buildURL("/v1/nodes/"+port.NodeUUID+"/ports", http.MethodGet), resp)
+	m.ResponseJSON(m.buildURL("/v1/ports", http.MethodGet), resp)
```
andfasano, line 220: A question, also `/v1/ports` without any node parameter must be configured?

pkg/provisioner/ironic/validatemanagementaccess_test.go
honza: I added an explicit check for the error kind.

pkg/provisioner/ironic/ironic.go
honza: Yes, `ports.List()` returns a `pagination.Pager` but it's simply an instance of the type with nothing in it. No work has been done yet. It doesn't touch the network until you try to paginate (or access the results), and thus there are no errors yet. 

https://github.com/gophercloud/gophercloud/blob/eeff63dbdcab503e85178c5faa8a0acd4735f7f7/openstack/baremetal/v1/ports/requests.go#L67

https://github.com/gophercloud/gophercloud/blob/eeff63dbdcab503e85178c5faa8a0acd4735f7f7/pagination/pager.go#L54

Dealing with the `err` returned from `AllPages()` should be enough here.

andfasano: Mmh, I think there could be a case of error (https://github.com/gophercloud/gophercloud/blob/eeff63dbdcab503e85178c5faa8a0acd4735f7f7/openstack/baremetal/v1/ports/requests.go#L63).

I mentioned it because I did the same mistake in another point of the code using a similar pattern, and ignoring the pager error made the troubleshooting really a pain.

honza: Good point, fixed!

pkg/provisioner/ironic/ironic.go
```diff
@@ -258,11 +283,48 @@ func (p *ironicProvisioner) findExistingHost() (ironicNode *nodes.Node, err erro
...
+	if len(allPorts) > 0 {
+		nodeUUID := allPorts[0].NodeUUID
+		ironicNode, err = nodes.Get(p.client, nodeUUID).Extract()
+		switch err.(type) {
+		case nil:
+			p.log.Info("found existing node by ID")
+
+			// If the node has a name, this means we didn't find it above.
+			if ironicNode.Name != "" {
+				return nil, errors.New("node found by MAC but has a name")
```
dhellmann, line 311: Do we want this error to say what name it found?

pkg/provisioner/ironic/validatemanagementaccess_test.go
```diff
@@ -179,3 +181,168 @@ func TestValidateManagementAccessNewCredentials(t *testing.T) {
...
+	existingNode := nodes.Node{
+		UUID: "33ce8659-7400-4c68-9535-d10766f07a58",
+	}
+
+	existingNodePort := ports.Port{
+		NodeUUID: existingNode.UUID,
+		Address:  "11:11:11:11:11:11",
+	}
+
+	createCallback := func(node nodes.Node) {
```
dhellmann, line 198: We see this pattern a few times. Maybe we can add a FailCreateNode() method to the IronicMock in another PR.

dhellmann (commented):
This is looking good. I have a couple of small questions inline, but I'm generally happy with it.

pkg/provisioner/ironic/ironic.go
```diff
@@ -258,11 +283,48 @@ func (p *ironicProvisioner) findExistingHost() (ironicNode *nodes.Node, err erro
...
+	// Try to load the node by port address
+	p.log.Info("looking for existing node by MAC", "MAC", p.host.Spec.BootMACAddress)
+	allPorts, err := p.listAllPorts(p.host.Spec.BootMACAddress)
+
+	if err != nil {
+		p.log.Info("failed to find an existing port with address", "MAC", p.host.Spec.BootMACAddress)
+		return nil, nil
+	}
+
+	if len(allPorts) > 0 {
```
dhellmann, line 302: I would probably have reversed this logic and returned nil early if there are no ports, but it's not worth holding this PR up over that.

dhellmann (commented):
/approve

Issue #708, opened 2020-11-03
BMH goes to non recoverable error state


furkatgofurov7:
There is a case where Ironic can't recover BMHs when connectivity with BMCs fails for some time. 
After that, Ironic will set a bare metal node as failed when it can't connect to its BMC for a short period of time. Once a bmh is in RegistrationError, it cannot be recovered, breaking the whole cluster. I have tested it locally, to imitate the situation of BMC connectivity loss, I have deleted vbmc container and re-created it after some time(10seconds) while BMHs' are in a registering state. 
Below provided are logs:
```
``Status:
  Error Message:  Failed to get power state for node 56055790-ddda-427c-a16a-f10261e52382. Error: Redfish connection failed for node 56055790-ddda-427c-a16a-f10261e52382: Unable to connect to https://10.33.131.4/redfish/v1/. Error: HTTPSConnectionPool(host='10.33.131.4', port=443): Max retries exceeded with url: /redfish/v1/ (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7f6982f7a080>: Failed to establish a new connection: [Errno 110] ETIMEDOUT',))
  Error Type:     registration error
  Good Credentials:
  Hardware Profile:  
  Last Updated:      2020-08-18T23:34:15Z
  Operation History:
    Deprovision:
      End:    <nil>
      Start:  <nil>
    Inspect:
      End:    <nil>
      Start:  <nil>
    Provision:
      End:    <nil>
      Start:  <nil>
    Register:
      End:             2020-08-18T23:34:15Z
      Start:           2020-08-18T23:20:47Z
  Operational Status:  error
  Powered On:          false
  Provisioning:
    ID:  56055790-ddda-427c-a16a-f10261e52382
    Image:
      Checksum:  
      URL:       
    State:       registration error
  Tried Credentials:
    Credentials:
      Name:               ephemeral
      Namespace:          dc325-vpod1-capi
    Credentials Version:  9857
Events:                   <none>` `
```

[...]

Issue #698, opened 2020-10-27
Registration of provisioned Host is never retried


zaneb:
When calling `Adopt()` (in the provisioned or externally provisioned states), we assume that either the Node does not exist in Ironic (in which case we attempt to register it), or that it is in the manageable state. If the Node exists in the enrol state due to Ironic having failed to move it to manageable (e.g. because the BMC is temporarily unreachable), then we go into an infinite reconcile loop where we just complain about the wrong state and never attempt to try again:

```
2020-10-21T18:13:59.447 Reconciling BareMetalHost {Request.Namespace: 'openshift-machine-api', Request.Name: 'openshift-master-0-0'}
2020-10-21T18:13:59.447 verifying ironic provisioner dependencies {host: 'openshift-master-0-0'}
2020-10-21T18:13:59.530 found existing node by ID {host: 'openshift-master-0-0'}
2020-10-21T18:13:59.531 Reconciler error {controller: 'metal3-baremetalhost-controller', request: 'openshift-machine-api/openshift-master-0-0', error: 'action "externally provisioned" failed: Invalid state for adopt: enroll'}
github.com/go-logr/zapr.(*zapLogger).Error
        /go/src/github.com/metal3-io/baremetal-operator/vendor/github.com/go-logr/zapr/zapr.go:128
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
        /go/src/github.com/metal3-io/baremetal-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:258
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
        /go/src/github.com/metal3-io/baremetal-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:232
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).worker
        /go/src/github.com/metal3-io/baremetal-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:211
k8s.io/apimachinery/pkg/util/wait.JitterUntil.func1
        /go/src/github.com/metal3-io/baremetal-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:152
k8s.io/apimachinery/pkg/util/wait.JitterUntil
[...]

PR #716, opened 2020-11-12
Handle re-registration in all states

zaneb:
Ensure that regardless of the current provisioning state, we always re-register the Node in ironic if the ironic database has been lost.
Ensure that if this occurs during deprovisioning, the Node is re-adopted so that we commence deprovisioning again, so that the Host is always left in a clean state.

This also changes the steady state that we use for Ready hosts in Ironic. Previously we would try to maintain the ironic state as `manageable`. Now we maintain it in `available` if possible. This means that we don't end up doing a second automated cleaning on provisioning if we have just done an automated cleaning on deprovisioning.

### DIFF
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -679,14 +679,29 @@ func (r *BareMetalHostReconciler) manageHostPower(prov provisioner.Provisioner, 
 	return steadyStateResult
 }
 
-// A host reaching this action handler should be provisioned or
-// externally provisioned -- a state that it will stay in until the
-// user takes further action. Both of those states mean that it has
-// been registered with the provisioner once, so we use the Adopt()
-// API to ensure that is still true. Then we monitor its power status.
+// A host reaching this action handler should be provisioned or externally
+// provisioned -- a state that it will stay in until the user takes further
+// action. Both of those states mean that it has been registered with the
+// provisioner once, so we use the ValidateManagmentAccess() API to ensure that
+// is still true, then the Adopt() API to make sure that the provisioner is
+// aware of the provisioning details. Then we monitor its power status.
 func (r *BareMetalHostReconciler) actionManageSteadyState(prov provisioner.Provisioner, info *reconcileInfo) actionResult {
+	// We always pass false for credentialsChanged because if they had
+	// changed we would have ended up in actionRegister() instead of
+	// here.
+	provResult, err := prov.ValidateManagementAccess(false)
+	if err != nil {
+		return actionError{err}
+	}
+	if provResult.ErrorMessage != "" {
+		return recordActionFailure(info, metal3v1alpha1.RegistrationError, provResult.ErrorMessage)
+	}
+	if provResult.Dirty {
+		info.host.ClearError()
+		return actionContinue{provResult.RequeueAfter}
+	}
 
-	provResult, err := prov.Adopt()
+	provResult, err = prov.Adopt()
 	if err != nil {
 		return actionError{err}
 	}
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -181,6 +181,7 @@ func (r *BareMetalHostReconciler) Reconcile(request ctrl.Request) (result ctrl.R
 	// management controller.
 	var bmcCreds *bmc.Credentials
 	var bmcCredsSecret *corev1.Secret
+	haveCreds := false
 	switch host.Status.Provisioning.State {
 	case metal3v1alpha1.StateNone, metal3v1alpha1.StateUnmanaged:
 		bmcCreds = &bmc.Credentials{}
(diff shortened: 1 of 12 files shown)


furkatgofurov7 (commented):
Looks great! I'm very much in favor of this. 
@zaneb we could update the PR description maybe, as this patch, will fix at least 2 (if not 3) of the open issues ([697](https://github.com/metal3-io/baremetal-operator/issues/697), [698](https://github.com/metal3-io/baremetal-operator/issues/698) and [708](https://github.com/metal3-io/baremetal-operator/issues/708)) in the repo.

zaneb:
Those issues are referenced in the individual commit messages, so they will get automatically closed when this merges.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -219,7 +222,7 @@ func (r *BareMetalHostReconciler) Reconcile(request ctrl.Request) (result ctrl.R
 		return ctrl.Result{Requeue: true, RequeueAfter: provisionerNotReadyRetryDelay}, nil
 	}
 
-	stateMachine := newHostStateMachine(host, r, prov)
+	stateMachine := newHostStateMachine(host, r, prov, haveCreds)
```
andfasano, line 225: Nit: just to avoid passing another param to the host machine, couldn't this check be done via the Provisioner interface (verifying that `bmcCreds` is not empty)?

pkg/provisioner/ironic/ironic.go
```diff
@@ -981,37 +993,24 @@ func (p *ironicProvisioner) startProvisioning(ironicNode *nodes.Node, hostConf p
...
 	if ironicNode == nil {
-		// The node does not exist, but we were called so the
-		// controller thinks that the node existed at one time. That
-		// likely means data loss from restarting the database, so
-		// pass through the validation process to register the node
-		// again. Pass true to indicate that we need to re-test the
-		// credentials, just in case.
-		p.log.Info("re-registering host")
-		return p.ValidateManagementAccess(true)
+		err = provisioner.NeedsRegistration
```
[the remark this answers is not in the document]
zaneb: Returning an error all the way up through Reconcile() (which this does) always results in a log trace.

controllers/metal3.io/baremetalhost_controller.go
zaneb: We don't know what type of Provisioner we have, so we don't know where (or even if) that data is stored. Also I don't want to have to look at the contents of the creds (which are not passed as a pointer) to see if they're empty or not.

honza:
Amazing commit messages!

The only thing I'd note is the absence of unit tests.

controllers/metal3.io/baremetalhost_controller.go
andfasano: `bmCreds` are already passed to the ProvisionerFactory, a plain wrapper having the same check in the Provisioner interface could avoid to store it directly in the state machine. But it's a minor point.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1225,45 +1224,21 @@ func (p *ironicProvisioner) Deprovision() (result provisioner.Result, err error)
...
 		"deploy step", ironicNode.DeployStep,
+		"instance_info", ironicNode.InstanceInfo,
 	)
 
 	switch nodes.ProvisionState(ironicNode.ProvisionState) {
 
 	case nodes.Error, nodes.CleanFail:
-		if !ironicNode.Maintenance {
-			p.log.Info("setting host maintenance flag to force image delete")
-			return p.setMaintenanceFlag(ironicNode, true)
```
dtantsur, line 1226: Clean failure may set `maintenance` mode, you need to clear it here (instead of setting it to `true` which we used to do for some reason).

dhellmann (approved):
This looks good. I appreciate the effort you took to split it into reviewable commits with detailed commit messages.

Now that we have a framework for adding them, I hope we can prioritize adding tests in follow-up PRs.

/lgtm

dhellmann:
/hold

I didn't notice the comment from @dtantsur before adding LGTM.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1094,12 +1087,22 @@ func (p *ironicProvisioner) Provision(hostConf provisioner.HostConfigData) (resu
 			return result, nil
 		}
 		p.log.Info("recovering from previous failure")
-		return p.startProvisioning(ironicNode, hostConf)
+		if provResult, err := p.startProvisioning(ironicNode, hostConf); err != nil || provResult.Dirty || provResult.ErrorMessage != "" {
+			return provResult, err
+		}
+
+		return p.changeNodeProvisionState(ironicNode,
+			nodes.ProvisionStateOpts{Target: nodes.TargetActive})
```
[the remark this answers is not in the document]
zaneb: This code used to be at the end of `startProvisioning()` but it has moved here because we only wanted to call it in the DeployFail state and not the Available state (which now calls `startProvisioning` - it used to be Manageable).

pkg/provisioner/ironic/ironic.go
zaneb: Thanks, I added another commit to do this.

maelk:
Great to see this work completed! This PR solves quite some issues for us. Thank you @zaneb for taking care of it

PR #717, opened 2020-11-12
Fix utility install scripts

zaneb:
These couldn't actually install anything.

### DIFF
--- a/hack/install-controller-gen.sh
+++ b/hack/install-controller-gen.sh
@@ -2,6 +2,7 @@
 
 set -ex
 
+CONTROLLER_TOOLS_VERSION=${1:-v0.4.0}
 OUTPUT=bin/controller-gen
 
 # Check for a vendor directory if any downstream forks use that dependency
@@ -13,11 +14,11 @@ then
 fi
 
 
-CMDPATH="$GOPATH/pkg/mod/sigs.k8s.io/controller-tools@v0.4.0/cmd/controller-gen"
+CMDPATH="$GOPATH/pkg/mod/sigs.k8s.io/controller-tools@${CONTROLLER_TOOLS_VERSION}/cmd/controller-gen"
 
 if [ ! -f "$CMDPATH" ]
 then
     go mod download
 fi
 
-go build -o "${OUTPUT}" "$CMDPATH"
+go build -o "${OUTPUT}" "sigs.k8s.io/controller-tools/cmd/controller-gen"
--- a/hack/install-golangci-lint.sh
+++ b/hack/install-golangci-lint.sh
@@ -20,4 +20,4 @@ then
     go mod download
 fi
 
-go build -o "${OUTPUT}" "$CMDPATH"
+go build -o "${OUTPUT}" "github.com/golangci/golangci-lint/cmd/golangci-lint"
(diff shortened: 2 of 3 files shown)


dhellmann:
Do we even need these now that the makefile installs everything into a local bin directory?

zaneb:
Aren't these scripts how it installs everything into the local bin directory?

dhellmann:
> Aren't these scripts how it installs everything into the local bin directory?

Heh, yes. I thought that stuff was all inline in the Makefile now.

hack/install-controller-gen.sh
```diff
@@ -13,11 +14,11 @@ then
...
-CMDPATH="$GOPATH/pkg/mod/sigs.k8s.io/controller-tools@v0.4.0/cmd/controller-gen"
+CMDPATH="$GOPATH/pkg/mod/sigs.k8s.io/controller-tools@${CONTROLLER_TOOLS_VERSION}/cmd/controller-gen"
 
 if [ ! -f "$CMDPATH" ]
 then
     go mod download
 fi
 
-go build -o "${OUTPUT}" "$CMDPATH"
+go build -o "${OUTPUT}" "sigs.k8s.io/controller-tools/cmd/controller-gen"
```
dhellmann, line 24: I guess we don't need a version here because we have the version pinned in go.mod?

zaneb: Yeah, that's the theory.
I don't know if it always works (i.e. gets the right version), but you can't specify a version to `go build`, only `go get`, according to the error message, so without this change it never works.

honza:
> Fix utility install scripts

This is kind of vague. Do you mind elaborating in the commit message?

The original scripts worked fine for me (and in CI for that matter), so I'm interested in seeing what kind of errors you ran into.

zaneb:
The error was:

```
can't load package: package /pkg/mod/sigs.k8s.io/controller-tools@v0.4.0/cmd/controller-gen: cannot use path@version syntax in GOPATH mode
```

when run from openshift-metal3/dev-scripts.

dhellmann:
@zaneb which version of go do you have installed? I don't see that error with 1.14.4.

zaneb:
```
$ /usr/local/go/bin/go version
go version go1.14.4 linux/amd64
```

Running the same command on my laptop with go 1.13 yields the same results for me.

The script where it's failing for me is [`metal3-dev/local-bmo.sh`](https://github.com/openshift-metal3/dev-scripts/blob/e9a590419f42a613a18e07ebbb8a39acb50abf1f/metal3-dev/local-bmo.sh)

zaneb:
OK, I misremembered the error. Reproducing on the test box instead of my laptop it's:

```
+(./tools/install_kustomize.sh:22): main(): go build -o bin/kustomize /pkg/mod/sigs.k8s.io/kustomize/kustomize/v3@v3.8.5
can't load package: package /pkg/mod/sigs.k8s.io/kustomize/kustomize/v3@v3.8.5: can only use path@version syntax with 'go get'
```

(similar for each of the tools).

The reason for the error message is that the path `/pkg/mod/...` doesn't exist, so it thinks we're trying to use a path@version rather than a filename. The reason it's looking for that path is that GOPATH is not exported by the script. That is better fixed there than here.

/close

zaneb:
Fixed in openshift-metal3/dev-scripts#1149

PR #722, opened 2020-11-17
Add default value for BMH errorCount field

andfasano:
Adding a default value for the new errorCount field to allow smoother upgrades


zaneb:
This file is generated by kubebuilder, so you can't just edit it directly.

It also looks like you're using a different version of kubebuilder than the one in the tools.

andfasano:
Thanks, I was trying to understand what went wrong, since even running the `hack/generate.sh` locally didn't produce any error. I'll check the kubebuilder version

dhellmann:
How does this help? Isn't the default for an integer always 0?

andfasano:
> How does this help? Isn't the default for an integer always 0?

The scenario tested is an older version of BMO (without the errorCount field) and a newer CRD (with the errorCount fiel defined) applied. Without declaring explicitly the default, when the BMO tries to update the status of a resource then an error is thrown. With the default, the resource is updated without any error (with errorCount = 0)

dhellmann:
> > How does this help? Isn't the default for an integer always 0?
> 
> The scenario tested is an older version of BMO (without the errorCount field) and a newer CRD (with the errorCount fiel defined) applied. Without declaring explicitly the default, when the BMO tries to update the status of a resource then an error is thrown. With the default, the resource is updated without any error (with errorCount = 0)

OK, that makes sense. Thanks for explaining!

furkatgofurov7:
This page https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_20.04/Release.key seems to be down, causing CI failure, while downloading the release keys.

dhellmann:
> This page https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_20.04/Release.key seems to be down, causing CI failure, while downloading the release keys.

We're looking for an Ubuntu key on an OpenSUSE web site?

furkatgofurov7:
> > This page https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_20.04/Release.key seems to be down, causing CI failure, while downloading the release keys.
> 
> We're looking for an Ubuntu key on an OpenSUSE web site?

@dhellmann apparently for adding xUbuntu_20.04 image release key, as part of running [ubuntu install requirements script](https://github.com/metal3-io/metal3-dev-env/blob/bee0c12efb8c1f38dc2c99178d094420040386a5/ubuntu_install_requirements.sh#L43) in metal3-dev-env. Seems to be back available now, re-triggering the test. 

/test-integration

maelk:
@dhellmann It's actually part of the podman installation : https://podman.io/getting-started/installation . A bit counter-intuitive...

PR #723, opened 2020-11-17
unittests: Add a mechanism for simulating a lost ironic db

honza:
Calling the new `ClearDatabase()` method will change all registered non-POST handlers to reply with a 404.  Attempts to to create new resources via POST requests will still succeed.

### DIFF
--- a/pkg/provisioner/ironic/testserver/ironic.go
+++ b/pkg/provisioner/ironic/testserver/ironic.go
@@ -270,3 +270,40 @@ func (m *IronicMock) Nodes(allNodes []nodes.Node) *IronicMock {
 	m.ResponseJSON(m.buildURL("/v1/nodes", http.MethodGet), resp)
 	return m
 }
+
+// ClearDatabase simulates the loss of the Ironic database
+//
+// Database clearing is simulated by setting all registered non-POST handlers to
+// return a 404 error.  Both regular handlers, and default handlers are
+// affected.
+func (m *IronicMock) ClearDatabase() {
+	m.t.Log("clearing ironic database")
+
+	// First, set all default handlers to reply with 404
+	for i, resp := range m.defaultResponses {
+		// POST requests should still succeed when the database is empty
+		if resp.method == "POST" {
+			continue
+		}
+
+		m.defaultResponses[i].code = http.StatusNotFound
+		m.defaultResponses[i].payload = ""
+	}
+
+	// Second, set named handlers to reply with 404
+	for url, methodMap := range m.responsesByMethod {
+
+		for method := range methodMap {
+			// POST requests should still succeed when the database is empty
+			if method == "POST" {
+				continue
+			}
+
+			m.responsesByMethod[url][method] = response{
+				code:    http.StatusNotFound,
+				payload: "",
+			}
+		}
+	}
+
+}
--- a/pkg/provisioner/ironic/testserver/ironic_test.go
+++ b/pkg/provisioner/ironic/testserver/ironic_test.go
@@ -1,0 +1,98 @@
+package testserver
+
+import (
+	"net/http"
+	"testing"
+
+	"github.com/gophercloud/gophercloud/openstack/baremetal/v1/nodes"
+)
+
+func TestIronicDatabaseClearing(t *testing.T) {
+	ironic := NewIronic(t).WithDefaultResponses()
+	ironic.AddDefaultResponse("/v1/nodes", "POST", http.StatusCreated, "{}")
+	ironic.Start()
+	defer ironic.Stop()
+
+	endpoint := ironic.Endpoint()
+
+	url := endpoint + "nodes/uuid"
+
+	resp, err := http.Get(url) // #nosec
+
+	if err != nil {
+		t.Error(err)
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		t.Fail()
+	}
+
+	ironic.ClearDatabase()
+
+	resp, err = http.Get(url) // #nosec
+
+	if err != nil {
+		t.Error(err)
+	}
+
+	if resp.StatusCode != http.StatusNotFound {
+		t.Fail()
+	}
+
+	// After clearing the db, POSTs should still work
+	resp, err = http.PostForm(endpoint+"nodes", nil)
+
+	if err != nil {
+		t.Error(err)
+	}
+
+	if resp.StatusCode != http.StatusCreated {
+		t.Fail()
+	}
+}
+
+func TestIronicDatabaseClearingNode(t *testing.T) {
+	ironic := NewIronic(t).WithDefaultResponses().Node(nodes.Node{
+		UUID: "abc",
+	})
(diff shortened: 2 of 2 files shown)


andfasano:
Hi @honza,
I think that in general the feature sounds interesting, and I have a couple of general points

1) By design we tried to keep `MockServer` as much as possible agnostic and generic, and to move any specific "business" trait into the specific implementations (`IronicMock` and `InspectorMocks`), so I'd recommend to push only generic features in `MockServer`.
2) Regarding implementing a global unique answer for any requests: I think it could be worth adding the possibility to configure the default response code. Now it's implicitly an `http.StatusOk` with an empty payload here[](https://github.com/metal3-io/baremetal-operator/blob/c72e2e0d880359a9434b484ba73c7a5990c2f976/pkg/provisioner/ironic/testserver/server.go#L252). With such mechanism, it would be enough to simply clear out `responsesByMethod` and `defaultResponses` (maybe a `Reset()` method?) and set the default response code in the event trigger.
3) Following 1), the even trigger could be implemented by invoking a generic `onSendData` func from `sendData`, implemented by `IronicMock`. Such handler could implement the required counting logic and implement the actions specified in 2).

honza:
I agree with 1 and 3.

For 2, I think `POST` requests should still succeed. I meant to do that but forgot. So, a sweeping reset probably isn't appropriate.

pkg/provisioner/ironic/testserver/ironic.go
```diff
@@ -230,3 +243,70 @@ func (m *IronicMock) WithNodeValidate(nodeUUID string) *IronicMock {
 	m.ResponseWithCode("/v1/nodes/"+nodeUUID+"/validate", "{}", http.StatusOK)
 	return m
 }
+
+// ClearIronicDatabase simulates the loss of the Ironic database
+//
+// Database clearing is simulated by setting all registered non-POST handlers to
+// return a 404 error.  Both regular handlers, and default handlers are
+// affected.
+func (m *IronicMock) ClearIronicDatabase() {
```
andfasano, line 252: Maybe this one could be private?

pkg/provisioner/ironic/testserver/ironic.go
```diff
@@ -230,3 +243,70 @@ func (m *IronicMock) WithNodeValidate(nodeUUID string) *IronicMock {
...
+
+		newDefaultResponse := defaultResponse{
+			method: resp.method,
+			re:     resp.re,
+			response: response{
+				code:    http.StatusNotFound,
+				payload: "",
+			},
+		}
+		newDefaultResponses = append(newDefaultResponses, newDefaultResponse)
```
andfasano, line 278: What about simplifying the code by directly modifying the slice elements? Ie:

```
for i := 0; i < len(m.defaultResponses); i++ {
  resp := &m.defaultResponses[i]
  // POST requests should still succeed when the database is empty
  if resp.method != http.MethodPost {
    resp.code = http.StatusNotFound
  }
}
```

pkg/provisioner/ironic/testserver/ironic.go
```diff
@@ -230,3 +243,70 @@ func (m *IronicMock) WithNodeValidate(nodeUUID string) *IronicMock {
...
+				newResponsesByMethod[url][method] = resp
+				continue
+			}
+
+			newResponsesByMethod[url][method] = response{
+				code:    http.StatusNotFound,
+				payload: "",
+			}
+		}
+	}
```
andfasano, line 301: Also this one could be simplified, in this case by simply reassigning a modified resp

pkg/provisioner/ironic/testserver/server.go
```diff
@@ -257,6 +260,10 @@ func (m *MockServer) sendData(w http.ResponseWriter, r *http.Request, code int,
 	w.Header().Set("Content-Type", "application/json")
 	w.WriteHeader(code)
 	fmt.Fprint(w, payload)
+
+	if m.onSendDataCallback != nil {
+		m.onSendDataCallback(r.URL, code, payload)
```
andfasano, line 265: :+1:

honza:
Note that this now also disables rule G107: https://securego.io/docs/rules/g107.html

andfasano:
Nice addition, changes are ok for me

dhellmann:
This looks OK, but I'm not sure what sorts of test scenarios we have that are enabled by this change. Can you elaborate a bit on that?

honza:
> This looks OK, but I'm not sure what sorts of test scenarios we have that are enabled by this change. Can you elaborate a bit on that?

The idea is that you can simulate losing the Ironic database while you're in the middle of, say, provisioning. After a certain amount of requests, Ironic will forget everything. We can then check that the error handling is correct at various points in the execution. This was previously impossible (to my knowledge).

 I'm already working on writing more tests to cover those scenarios.

dhellmann:
> > This looks OK, but I'm not sure what sorts of test scenarios we have that are enabled by this change. Can you elaborate a bit on that?
> 
> The idea is that you can simulate losing the Ironic database while you're in the middle of, say, provisioning. After a certain amount of requests, Ironic will forget everything. We can then check that the error handling is correct at various points in the execution. This was previously impossible (to my knowledge).
> 
> I'm already working on writing more tests to cover those scenarios.

OK, I see.

I'm worried that basing this on the number of API calls made will mean that we have to update tests as the calling pattern of the provisioner changes over time, which it is likely to do as we add features like node cleaning or fix bugs like ensuring that the driver for a node in ironic matches what we have in the BMC URL if the URL is changed after the host is first registered.

Can we make the "discard the database" step a public method of the server so it can be invoked by the test? That way the test could, for example, run Reconcile() once, then throw the database away, then run Reconcile() again. Or it could invoke one method of the Provisioner, then discard the database, then invoke another method of the Provisioner.

honza:
Fair enough---I was overthinking it. I removed the request counting logic, and left the `ClearDatabase()` method, and made it public.

zaneb:
This looks great. How does one un-clear the database again? Does it automatically get repopulated from the POST requests? Can we maybe check that in the unit tests?

s3rj1k:
@honza Seems that this PR needs a rebase

s3rj1k:
I am very interested in this PR, maybe we could land this one?

honza:
> This looks great. How does one un-clear the database again? Does it automatically get repopulated from the POST requests? Can we maybe check that in the unit tests?

Currently, the server is pretty dumb. It doesn't actually maintain much of a state, let alone a database. When you're writing a test, and you need the database to start returning real data after clearing it, you can, of course, manually register new handlers for the resources that you need. If we want an *ad hoc, informally-specified, bug-ridden, slow implementation* of Ironic, we can have that conversation but it's beyond the scope of this tiny PR. :joy: 

The initial impulse for this was the scenario where you set up some state in Ironic, and then simulate the database going away. You can then run the first reconcile step to check that you get back the expected action result.

zaneb:
OK, let's get this merged then and I guess we'll find out when people start using it whether there's anything more to do.
/lgtm

PR #725, opened 2020-11-19
Limit the number of hosts simultaneously provisioned

andfasano:
This PR introduces a soft limit for the number of hosts being simultaneously being provisioned.

Design doc: https://github.com/metal3-io/metal3-docs/blob/master/design/baremetal-operator/limit-hosts-provisioning.md. 

A new configuration variable is provided, called `PROVISIONING_LIMIT` (defaulted to 20), to specify the desired batch size that the operator will try to honor while provisioning new hosts. If there are no available slots, the current action for an host will be rescheduled using a fixed backoff with jittering to spread evenly the delayed requests.
Due the asynchronous behaviour of the Ironic subsystem, and the configurable level of concurrency of the operator, it's not possible to guarantee that such limit will be always respected, and occasional overflows could happen especially during the initial burst; nevertheless the operator will try to enforce such limit as much as possible, and for such reason it's recommended to keep a configuration where PROVISIONING_LIMIT is greater than BMO_CONCURRENCY.

----
__Demo__

Here's a quick demo of the current implementation, using a limit set to 8 and a BMO concurrency level of 3 (demo speeded up). A batch of additional 15 nodes is added to the cluster. It is possible to note the overflow on the initial burst, then the limit is maintained until the provisioning was completed.

[asciicast](https://asciinema.org/a/9baJxFlOqEh5XlRXCYHrPKQfC)

### DIFF
--- a/apis/metal3.io/v1alpha1/baremetalhost_types.go
+++ b/apis/metal3.io/v1alpha1/baremetalhost_types.go
@@ -116,6 +116,8 @@ const (
 	// OperationalStatusError is the status value for when the host
 	// has any sort of error.
 	OperationalStatusError OperationalStatus = "error"
+
+	OperationalStatusDelayed = "delayed"
 )
 
 // ErrorType indicates the class of problem that has caused the Host resource
@@ -523,7 +525,7 @@ type BareMetalHostStatus struct {
 	// after modifying this file
 
 	// OperationalStatus holds the status of the host
-	// +kubebuilder:validation:Enum="";OK;discovered;error
+	// +kubebuilder:validation:Enum="";OK;discovered;error;delayed
 	OperationalStatus OperationalStatus `json:"operationalStatus"`
 
 	// ErrorType indicates the type of failure encountered when the
--- a/controllers/metal3.io/action_result.go
+++ b/controllers/metal3.io/action_result.go
@@ -12,7 +12,9 @@ import (
 	"github.com/metal3-io/baremetal-operator/pkg/provisioner"
 )
 
-const maxBackOffCount = 10
+// This is an upper limit for the ErrorCount, so that the max backoff
+// timeout will not exceed (roughly) 8 hours
+const maxBackOffCount = 9
 
 func init() {
 	rand.Seed(time.Now().UTC().UnixNano())
@@ -52,6 +54,18 @@ type actionUpdate struct {
 
 func (r actionUpdate) Dirty() bool {
 	return true
+}
+
+// actionDelayed it's the same of an actionUpdate, but the requeue time
+// is calculated using a fixed backoff with jitter
+type actionDelayed struct {
+	actionUpdate
+}
+
+func (r actionDelayed) Result() (result reconcile.Result, err error) {
+	result.RequeueAfter = calculateBackoff(1)
+	result.Requeue = true
+	return
 }
 
 // actionComplete is a result indicating that the current action has completed,
@@ -110,6 +124,16 @@ type actionFailed struct {
 	errorCount int
 }
 
+// Distribution sample for errorCount values:
+// 1  [1m, 2m]
+// 2  [2m, 4m]
+// 3  [4m, 8m]
+// 4  [8m, 16m]
+// 5  [16m, 32m]
+// 6  [32m, 1h4m]
+// 7  [1h4m, 2h8m]
+// 8  [2h8m, 4h16m]
+// 9  [4h16m, 8h32m]
 func calculateBackoff(errorCount int) time.Duration {
 
 	if errorCount > maxBackOffCount {
@@ -119,7 +143,7 @@ func calculateBackoff(errorCount int) time.Duration {
 	base := math.Exp2(float64(errorCount))
 	/* #nosec */
 	backOff := base - (rand.Float64() * base * 0.5)
-	backOffDuration := time.Minute * time.Duration(backOff)
+	backOffDuration := time.Duration(float64(time.Minute) * backOff)
 	return backOffDuration
 }
 
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -53,9 +53,6 @@ const (
 	rebootAnnotationPrefix        = "reboot.metal3.io"
 )
 
-func init() {
-}
-
 // BareMetalHostReconciler reconciles a BareMetalHost object
 type BareMetalHostReconciler struct {
 	client.Client
@@ -214,7 +211,6 @@ func (r *BareMetalHostReconciler) Reconcile(request ctrl.Request) (result ctrl.R
 	}
 
 	ready, err := prov.IsReady()
-
 	if err != nil {
 		return ctrl.Result{}, errors.Wrap(err, "failed to check services availability")
 	}
@@ -293,6 +289,15 @@ func recordActionFailure(info *reconcileInfo, errorType metal3v1alpha1.ErrorType
 	info.publishEvent(eventType, errorMessage)
 
 	return actionFailed{dirty: true, ErrorType: errorType, errorCount: info.host.Status.ErrorCount}
+}
+
+func recordActionDelayed(info *reconcileInfo) actionResult {
+
+	counter := delayedProvisioningHostCounters.With(hostMetricLabels(info.request))
+	info.postSaveCallbacks = append(info.postSaveCallbacks, counter.Inc)
+
+	info.host.SetOperationalStatus(metal3v1alpha1.OperationalStatusDelayed)
+	return actionDelayed{}
 }
 
 func (r *BareMetalHostReconciler) credentialsErrorResult(err error, request ctrl.Request, host *metal3v1alpha1.BareMetalHost) (ctrl.Result, error) {
--- a/controllers/metal3.io/host_state_machine.go
+++ b/controllers/metal3.io/host_state_machine.go
@@ -6,6 +6,7 @@ import (
 	metal3v1alpha1 "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
 	"github.com/metal3-io/baremetal-operator/pkg/provisioner"
 
+	"github.com/pkg/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
@@ -77,9 +78,33 @@ func recordStateEnd(info *reconcileInfo, host *metal3v1alpha1.BareMetalHost, sta
 	return
 }
 
+func (hsm *hostStateMachine) ensureProvisioningCapacity(info *reconcileInfo) actionResult {
+	hasCapacity, err := hsm.Provisioner.HasProvisioningCapacity()
+	if err != nil {
+		return actionError{errors.Wrap(err, "failed to get hosts currently being provisioned")}
+	}
+	if !hasCapacity {
+		return recordActionDelayed(info)
+	}
+
+	return nil
+}
+
 func (hsm *hostStateMachine) updateHostStateFrom(initialState metal3v1alpha1.ProvisioningState,
-	info *reconcileInfo) {
+	info *reconcileInfo) actionResult {
 	if hsm.NextState != initialState {
+
+		// Check if there is a free slot available when trying to
+		// provision an host - if not the action will be delayed.
+		// The check is limited to only the provisioning states to
+		// avoid putting an excessive pressure on the provisioner
+		switch hsm.NextState {
+		case metal3v1alpha1.StateInspecting, metal3v1alpha1.StateProvisioning:
+			if actionRes := hsm.ensureProvisioningCapacity(info); actionRes != nil {
+				return actionRes
+			}
+		}
+
 		info.log.Info("changing provisioning state",
 			"old", initialState,
 			"new", hsm.NextState)
@@ -109,11 +134,47 @@ func (hsm *hostStateMachine) updateHostStateFrom(initialState metal3v1alpha1.Pro
 			}
 		}
 	}
+
+	return nil
 }
 
-func (hsm *hostStateMachine) ReconcileState(info *reconcileInfo) actionResult {
+func (hsm *hostStateMachine) checkDelayedHost(info *reconcileInfo) actionResult {
+
+	// Check if there's a free slot for hosts that have been previously delayed
+	if info.host.Status.OperationalStatus == metal3v1alpha1.OperationalStatusDelayed {
+		if actionRes := hsm.ensureProvisioningCapacity(info); actionRes != nil {
+			return actionRes
+		}
+
+		// A slot is available, let's cleanup the status and retry
+		clearError(info.host)
+		return actionUpdate{}
+	}
+
+	// Make sure the check is re-applied when provisioning an
+	// host not yet tracked by the provisioner
+	switch info.host.Status.Provisioning.State {
+	case metal3v1alpha1.StateInspecting, metal3v1alpha1.StateProvisioning:
+		if actionRes := hsm.ensureProvisioningCapacity(info); actionRes != nil {
+			return actionRes
+		}
+	}
+
+	return nil
+}
+
+func (hsm *hostStateMachine) ReconcileState(info *reconcileInfo) (actionRes actionResult) {
 	initialState := hsm.Host.Status.Provisioning.State
-	defer hsm.updateHostStateFrom(initialState, info)
+
+	defer func() {
+		if overrideAction := hsm.updateHostStateFrom(initialState, info); overrideAction != nil {
+			actionRes = overrideAction
+		}
+	}()
+
+	if delayedResult := hsm.checkDelayedHost(info); delayedResult != nil {
+		return delayedResult
+	}
 
 	if hsm.checkInitiateDelete() {
 		info.log.Info("Initiating host deletion")
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -3,6 +3,7 @@ package ironic
 import (
 	"fmt"
 	"os"
+	"strconv"
 	"strings"
 	"time"
 
@@ -39,6 +40,7 @@ var (
 	ironicInsecure            bool
 	ironicAuth                clients.AuthConfig
 	inspectorAuth             clients.AuthConfig
+	maxProvisioningHosts      int = 20
 
 	// Keep pointers to ironic and inspector clients configured with
 	// the global auth settings to reuse the connection between
@@ -98,6 +100,15 @@ func init() {
 	ironicInsecureStr := os.Getenv("IRONIC_INSECURE")
 	if strings.ToLower(ironicInsecureStr) == "true" {
 		ironicInsecure = true
+	}
+
+	if maxHostsStr := os.Getenv("PROVISIONING_LIMIT"); maxHostsStr != "" {
+		value, err := strconv.Atoi(maxHostsStr)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Cannot start: Invalid value set for variable PROVISIONING_LIMIT=%s", maxHostsStr)
+			os.Exit(1)
+		}
+		maxProvisioningHosts = value
 	}
 }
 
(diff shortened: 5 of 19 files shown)


pkg/provisioner/ironic/ironic.go
```diff
@@ -1523,5 +1536,53 @@ func (p *ironicProvisioner) IsReady() (result bool, err error) {
...
+		return false, err
+	}
+
+	for _, node := range allNodes {
+
+		if _, ok := node.DriverInternalInfo["agent_url"]; !ok {
+			continue
+		}
+
+		switch node.ProvisionState {
```
zaneb, line 1574: Would looking at the target provisioning state instead/as well provide a faster response to changes?

andfasano: Unfortunately it didn't :/. In my tests also TargetState usually points just to `manageable` or `active` states

pkg/provisioner/ironic/ironic.go
```diff
@@ -1525,3 +1538,58 @@ func (p *ironicProvisioner) IsReady() (result bool, err error) {
...
+
+func (p *ironicProvisioner) getCurrentProvisioningHosts() map[string]struct{} {
+	//lazy init
+	if p.provisioningHosts != nil {
+		return p.provisioningHosts
+	}
+
+	p.provisioningHosts = make(map[string]struct{})
+	if err := p.loadProvisioningHosts(p.provisioningHosts); err != nil {
+		p.log.Error(err, "Unable to get hosts currently being provisioned")
```
dhellmann, line 1559: Shouldn't this error propagate up through the IsProvisioning() and CanBeProvisioned() APIs?

controllers/metal3.io/host_state_machine.go
```diff
@@ -76,8 +76,17 @@ func recordStateEnd(info *reconcileInfo, host *metal3v1alpha1.BareMetalHost, sta
 }
 
 func (hsm *hostStateMachine) updateHostStateFrom(initialState metal3v1alpha1.ProvisioningState,
-	info *reconcileInfo) {
+	info *reconcileInfo) bool {
 	if hsm.NextState != initialState {
+
+		// Check if there is a free slot for provisioning the host
+		switch hsm.NextState {
+		case metal3v1alpha1.StateInspecting, metal3v1alpha1.StateProvisioning:
```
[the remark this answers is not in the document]
andfasano: Ok

pkg/provisioner/ironic/ironic.go
andfasano: Well, initially I took that approach, just to note that the code in the state machine was slightly more complicated for the standard checks, and still the error handling wasn't very clear (refuse at all the operation?), so I opted for a simplified interface / error management giving priority to state handler (let's process anyway the request).
I also considered that in most of the cases an erroneous situation should be detected by the previous IsReady(), so the chances to get a problem here were fewer (but still non-zero).

controllers/metal3.io/host_state_machine.go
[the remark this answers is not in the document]
andfasano: Reviewed the comments (also in updateHostStateFrom)

controllers/metal3.io/host_state_machine.go
```diff
@@ -107,11 +119,50 @@ func (hsm *hostStateMachine) updateHostStateFrom(initialState metal3v1alpha1.Pro
 			}
 		}
 	}
+
+	return true
 }
 
-func (hsm *hostStateMachine) ReconcileState(info *reconcileInfo) actionResult {
+func (hsm *hostStateMachine) checkDelayedHost(info *reconcileInfo) (res bool, actionRes actionResult) {
```
zaneb, line 126: Just from reading this function I have no idea what the first return value means.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -293,6 +289,18 @@ func recordActionFailure(info *reconcileInfo, errorType metal3v1alpha1.ErrorType
 	return actionFailed{dirty: true, ErrorType: errorType, errorCount: info.host.Status.ErrorCount}
 }
 
+func recordActionFailureForBusyProvisioner(info *reconcileInfo) actionFailed {
```
zaneb, line 292: I'm thinking that we will need to move the state machine file to its own module soon, so it would be better if this were defined in that file and didn't have any dependencies on stuff in baremetalhost_controller.go.

pkg/provisioner/provisioner.go
```diff
@@ -88,6 +88,12 @@ type Provisioner interface {
 	// IsReady checks if the provisioning backend is available to accept
 	// all the incoming requests.
 	IsReady() (result bool, err error)
+
+	// IsProvisioning verifies if the current host is currently being provisioned or not
+	IsProvisioning() bool
+
+	// CanBeProvisioned checks if the backend has a free provisioning slot
+	CanBeProvisioned() bool
```
zaneb, line 96: Couldn't these be collapsed into a single HasProvisioningCapacity() function? They already share state.

controllers/metal3.io/host_state_machine.go
```diff
@@ -107,11 +119,50 @@ func (hsm *hostStateMachine) updateHostStateFrom(initialState metal3v1alpha1.Pro
...
 	initialState := hsm.Host.Status.Provisioning.State
-	defer hsm.updateHostStateFrom(initialState, info)
+
+	defer func() {
+		if !hsm.updateHostStateFrom(initialState, info) {
+			actionRes = recordActionFailureForBusyProvisioner(info)
+		}
+	}()
+
+	if delayed, actionRes := hsm.checkDelayedHost(info); delayed {
```
zaneb, line 163: We could just check for nil instead of having two return values, like we do for ensureRegistered() on line 172.

pkg/provisioner/ironic/ironic.go
zaneb: I agree with Doug here. We should retry otherwise a temporary ironic outage (maybe caused by extreme overload) could cause everything to be allowed to start provisioning.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1623,3 +1636,58 @@ func (p *ironicProvisioner) IsReady() (result bool, err error) {
...
+	allNodes, err := nodes.ExtractNodes(page)
+	if err != nil {
+		return err
+	}
+
+	for _, node := range allNodes {
+
+		switch node.ProvisionState {
+		case "cleaning", "clean wait", "inspecting", "inspect wait",
+			"deploying", "wait call-back":
```
zaneb, line 1686: Use the constants defined in the nodes package.

pkg/provisioner/ironic/ironic.go
zaneb: At the start of provisioning we are usually in manageable state (not always after #716, but always for the first deployment). So if we're seeing e.g. current state is manageable but target state is available, that machine is probably about to be provisioned. Ditto for current state is available and target state is active.
I assume the target state must be updated synchronously, so it really should be available whenever we look.

I suspect the real problem here is that there's a large delay between the point where you're making a go/no-go decision (on entry to the Provisioning state) and when we actually ask Ironic to go to active (not until after cleaning). And now it's even more fun because a node in the available state may _or may not_ represent a Host in the provisioning state.

I think the solution to this is the Preparing state I proposed [on the mailing list](https://groups.google.com/g/metal3-dev/c/N-EpqQn94bw). That way all nodes will be in available before starting provisioning, so we'd only have to look for those with target state active.

zaneb (commented):
OK, so if I'm following correctly here, the main thing you've changed is to check for capacity and block reconciling continuously, rather than just preventing the Host from entering the Provisioning state. (In fact, if we do it this way I think we should stop blocking the host from changing state - better to consistently block only in Provisioning than sometimes in Provisioning and sometimes in Ready.) This is suboptimal, because it may mean that we e.g. let a node start cleaning when another node that has already finished cleaning is waiting to start provisioning, but possibly necessary because of the very large delay between entering the Provisioning state and doing anything that looks like provisioning in Ironic. However, if we find that only a subset of states during provisioning are the ones that are actually placing strain on ironic (sitting there waiting for IPA to clean a node might be easy, for example) then this could still come out as a win.

The following combinations of current and target node provisioning states correspond to Hosts in the Provisioning state with what I'd expect (with one exception) to be a fairly high degree of consistency:

* `Manageable` -> `TargetProvide`
* `Cleaning` -> `TargetProvide`
* `CleanWait` -> `TargetProvide`
* `CleanFail` (could actually be deprovisioning, but it's brief enough that we can live with it)
* `Available` -> `TargetActive`
* `Deploying`
* `DeployWait`

However, there is a large hole there - once the state becomes `Available` but before we request `TargetActive`, we cannot tell from the ironic states alone whether the node is Provisioning or Ready after deprovisioning. (Unless, of course, we were to store this information somewhere in ironic - which is totally possible.) While setting up the data and requesting a move to active is relatively quick, it may take some time for the operator to notice that the node has reached `Available` and in that window it may start allowing other Hosts to start provisioning.

If we implement [my proposal](https://groups.google.com/g/metal3-dev/c/N-EpqQn94bw) to add a `Preparing` state _before_ `Ready`, so that the request for TargetActive always happens right at the start of Provisioning, then I think that gating on state changes only would become viable again. Since we want that state anyway, it's worth at least considering whether our path would look easier if we did that first or merged this and then reworked it later, assuming we prefer to gate only on state changes.

Use of the error message and error type fields is effective (operational status would be another option, but this is more visible). However, after having written #749 I now think it's a mistake to increment the error count even once because it mixes these routine delays with other possible causes for error. I think all we really need here is a random backoff from a fixed interval. It's fine to add a new actionResult type for this - we actually need one for retry-after-delay anyway.

pkg/provisioner/ironic/ironic.go
andfasano: Done

pkg/provisioner/ironic/ironic.go
```diff
@@ -1623,3 +1636,58 @@ func (p *ironicProvisioner) IsReady() (result bool, err error) {
...
+			hosts[node.Name] = struct{}{}
+		}
```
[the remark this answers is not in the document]
andfasano: In all of my tests I wasn't able to detect such scenarios, it looks like that either ironic sets the target when change the provision state or keep it as it is. I also added a check just after changing the provisioner state, to be sure. Maybe @dtantsur can confirm

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -293,6 +289,18 @@ func recordActionFailure(info *reconcileInfo, errorType metal3v1alpha1.ErrorType
...
+
+	// Avoid incrementing the error counter if the host has been already delayed,
+	// to keep a narrow retry window with some jitter
+	switch info.host.Status.ErrorType {
+	case metal3v1alpha1.ProvisionerBusyError:
+		return actionFailed{dirty: false, ErrorType: info.host.Status.ErrorType, errorCount: info.host.Status.ErrorCount}
```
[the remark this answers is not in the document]
andfasano: I didn't want to flatten the timeout to avoid losing the jitter, so I added the limit directly on the max count

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -293,6 +289,18 @@ func recordActionFailure(info *reconcileInfo, errorType metal3v1alpha1.ErrorType
...
+	}
+
+	return recordActionFailure(info, metal3v1alpha1.ProvisionerBusyError, "Delayed, provisioner busy")
```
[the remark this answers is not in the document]
andfasano: See above

controllers/metal3.io/action_result.go
```diff
@@ -110,6 +110,17 @@ type actionFailed struct {
 	errorCount int
 }
 
+// Distribution sample:
+// 58s
```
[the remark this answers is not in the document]
andfasano: Right, wrong doc, fixed it, thanks

controllers/metal3.io/action_result.go
```diff
@@ -110,6 +110,17 @@ type actionFailed struct {
...
+// 1m57s
+// 3m19s
+// 7m32s
+// 9m11s
+// 29m18s
+// 41m18s
+// 1h19m8s
+// 3h24m58s
+// 5h36m10s
```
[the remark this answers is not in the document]
andfasano: Done

controllers/metal3.io/action_result.go
```diff
@@ -119,7 +130,7 @@ func calculateBackoff(errorCount int) time.Duration {
 	base := math.Exp2(float64(errorCount))
 	/* #nosec */
 	backOff := base - (rand.Float64() * base * 0.5)
-	backOffDuration := time.Minute * time.Duration(backOff)
+	backOffDuration := time.Second * time.Duration(backOff*60)
```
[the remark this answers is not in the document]
andfasano: Done

pkg/provisioner/provisioner.go
andfasano: I reviewed all the checks currently applied, and it seems ok to merge them together in all the three cases (at the beginning of the state machine and when transitioning). Done

controllers/metal3.io/host_state_machine.go
andfasano: Done

pkg/provisioner/ironic/ironic.go
andfasano: Initially I couldn't use the const since some of the states were missing (I've also added them in this PR: https://github.com/gophercloud/gophercloud/pull/2052). But then I moved to a more restricted subset of states, so if we keep such subset (the original proposed was slightly larger) it would be ok to use the const. Done

pkg/provisioner/ironic/ironic.go
andfasano: @dtantsur can you please confirm such behavior? From my local tests the state doesn't seem to be update synchronously, neither the target_provision_state

dtantsur: > current state is manageable but target state is available

To nitpick a bit: this is not possible. So called stable states always have target state None. Stable states are states where a node simply stays without external input (BMO is external input in this context). This includes ``manageable``, ``available``, ``active``. Non-empty target state appears in so called transient states: ``wait call-back``, all states ending with ``..ing`` or ``.. wait``.

Now let me check the code to provide a more targeted answer.

pkg/provisioner/ironic/ironic.go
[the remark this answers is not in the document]
dtantsur: Manageable and available are stable states (see my explanation earlier), they should not be counted.

Well, you could count nodes in ``available`` state with ``instance_uuid`` set but it does not provide you a noticeable advantage.

dtantsur (approved):
LGTM

andfasano:
> 
> The following combinations of current and target node provisioning states correspond to Hosts in the Provisioning state with what I'd expect (with one exception) to be a fairly high degree of consistency:
> 
>     * `Manageable` -> `TargetProvide`
> 
>     * `Cleaning` -> `TargetProvide`
> 
>     * `CleanWait` -> `TargetProvide`
> 
>     * `CleanFail` (could actually be deprovisioning, but it's brief enough that we can live with it)
> 
>     * `Available` -> `TargetActive`
> 
>     * `Deploying`
> 
>     * `DeployWait`
> 
Apart the `manageable` and `available` states that cannot be used as per the @dtantsur comment, the others have been already in use - the overall goal was to roughly define how busy is currently Ironic (in the initial proposal the subset was slightly larger, including the rescuing/unrescuing/deleting states, which probably could make sense to reintroduce now).

> 
> However, there is a large hole there - once the state becomes `Available` but before we request `TargetActive`, we cannot tell from the ironic states alone whether the node is Provisioning or Ready after deprovisioning. (Unless, of course, we were to store this information somewhere in ironic - which is totally possible.) While setting up the data and requesting a move to active is relatively quick, it may take some time for the operator to notice that the node has reached `Available` and in that window it may start allowing other Hosts to start provisioning.
> 
> If we implement [my proposal](https://groups.google.com/g/metal3-dev/c/N-EpqQn94bw) to add a `Preparing` state _before_ `Ready`, so that the request for TargetActive always happens right at the start of Provisioning, then I think that gating on state changes only would become viable again. Since we want that state anyway, it's worth at least considering whether our path would look easier if we did that first or merged this and then reworked it later, assuming we prefer to gate only on state changes.
> 
Currently gating on state changes seems to work reasonably well, without affecting the convergence of the overall provisioning, so I'd prefer to proceed and merge with this approach and review it later once the new `Preparing` state will be available.

> Use of the error message and error type fields is effective (operational status would be another option, but this is more visible). However, after having written #749 I now think it's a mistake to increment the error count even once because it mixes these routine delays with other possible causes for error. I think all we really need here is a random backoff from a fixed interval. It's fine to add a new actionResult type for this - we actually need one for retry-after-delay anyway.

The ProvisionerBusy error is now used to detect that an host have been previously delayed, so an initial check can be performed even before applying the state machine. If we want to move to a plain action result for the backoff - and also see the usefulness of that - then I suppose operational status could be used to track this situation

andfasano:
> > Use of the error message and error type fields is effective (operational status would be another option, but this is more visible). However, after having written #749 I now think it's a mistake to increment the error count even once because it mixes these routine delays with other possible causes for error. I think all we really need here is a random backoff from a fixed interval. It's fine to add a new actionResult type for this - we actually need one for retry-after-delay anyway.
> 
> The ProvisionerBusy error is now used to detect that an host have been previously delayed, so an initial check can be performed even before applying the state machine. If we want to move to a plain action result for the backoff - and also see the usefulness of that - then I suppose operational status could be used to track this situation

Update: I've added a new actionDelayed that randomly backoff in [1m, 2m] and switched to use OperationalStatus to track delayed hosts instead of the error type

pkg/provisioner/ironic/ironic.go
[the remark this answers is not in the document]
zaneb: OK, if it changes from manageable->cleaning and available->deploying synchronously then this is not an issue. Thanks Dmitry for the clarification.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1637,27 +1637,33 @@ func (p *ironicProvisioner) IsReady() (result bool, err error) {
...
+	return len(hosts) < maxProvisioningHosts, nil
 }
 
-func (p *ironicProvisioner) getCurrentProvisioningHosts() map[string]struct{} {
+func (p *ironicProvisioner) getCurrentProvisioningHosts() (map[string]struct{}, error) {
 	//lazy init
 	if p.provisioningHosts != nil {
-		return p.provisioningHosts
+		return p.provisioningHosts, nil
 	}
```
zaneb, line 1658: We don't need this cache any more, do we?

pkg/provisioner/ironic/ironic.go
```diff
@@ -1687,12 +1687,12 @@ func (p *ironicProvisioner) loadProvisioningHosts(hosts map[string]struct{}) err
 
 	for _, node := range allNodes {
 
-		switch node.ProvisionState {
-		case "cleaning", "clean wait", "inspecting", "inspect wait",
-			"deploying", "wait call-back":
+		switch nodes.ProvisionState(node.ProvisionState) {
+		case nodes.Cleaning, nodes.CleanWait,
+			nodes.Inspecting, nodes.InspectWait,
+			nodes.Deploying, nodes.DeployWait:
```
zaneb, line 1693: Gopher Claus <- gopherclouse

controllers/metal3.io/action_result.go
```diff
@@ -36,6 +36,16 @@ func (r actionContinueNoWrite) Dirty() bool {
 	return false
 }
 
+type actionDelayed struct {
+	actionContinue
```
zaneb, line 40: Just a note that after https://github.com/metal3-io/baremetal-operator/pull/761/commits/5663434797fc88bd5a1c2c524baf8400e7ea5589 this will need to be `actionUpdate` (or possibly we will add a `Dirty()` method + flag).
Whichever PR merges first, we'll need to remember to update the other.

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -116,6 +116,8 @@ const (
 	// OperationalStatusError is the status value for when the host
 	// has any sort of error.
 	OperationalStatusError OperationalStatus = "error"
+
+	OperationalStatusDelayed = "delayed"
```
zaneb, line 120: I'm wondering if there is a better name for this, but I can't think of one.

zaneb (commented):
/approve

controllers/metal3.io/host_state_machine.go
```diff
@@ -75,8 +76,20 @@ func recordStateEnd(info *reconcileInfo, host *metal3v1alpha1.BareMetalHost, sta
 	}
 }
 
+func (hsm *hostStateMachine) ensureProvisioninCapacity(info *reconcileInfo) actionResult {
```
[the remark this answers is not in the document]
andfasano: thanks

pkg/provisioner/ironic/ironic.go
andfasano: Right, removing it

andfasano:
Commits squashed and demo cast updated

zaneb (commented):
Rebase required, but otherwise looks good to me. Love the demo!

controllers/metal3.io/action_result.go
andfasano: Thanks, done

zaneb:
I don't see any issues with the rebase. Who wants to LGTM?

PR #726, opened 2020-11-20
Close connections to inspector

derekhiggins:
connections to inspector are being left open for reuse but
not actually getting reused. Set
RequestOpts.KeepResponseBody: true
which disables the connections reuse functioanlity while the
problem is fixed.

Fixes: openshift#112

### DIFF
--- a/pkg/provisioner/ironic/dependencies.go
+++ b/pkg/provisioner/ironic/dependencies.go
@@ -45,9 +45,11 @@ func (i *ironicDependenciesChecker) checkEndpoint(client *gophercloud.ServiceCli
 	// from the client endpoint.
 	endpoint := strings.TrimSuffix(client.Endpoint, "/")
 
-	_, err := client.Get(endpoint, nil, nil)
+	resp, err := client.Get(endpoint, nil, &gophercloud.RequestOpts{KeepResponseBody: true})
 	if err != nil {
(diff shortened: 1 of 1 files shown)


zaneb:
Is this fixed by #712?

maelk:
@kashifest you verified that caching the client was solving the issue, right ? So we should not need this option set afaik.

kashifest:
> @kashifest you verified that caching the client was solving the issue, right ? So we should not need this option set afaik.

I believe so , yes. This should not be needed anymore after #712 . @derekhiggins would you please verify your usecase again with #712 ?

derekhiggins:
> Is this fixed by #712?

looks like it would be, I'll try it out

> > @kashifest you verified that caching the client was solving the issue, right ? So we should not need this option set afaik.
> 
> I believe so , yes. This should not be needed anymore after #712 . @derekhiggins would you please verify your usecase again with #712 ?

will do

derekhiggins:
> > Is this fixed by #712?
> 
> looks like it would be, I'll try it out
> 
> > > @kashifest you verified that caching the client was solving the issue, right ? So we should not need this option set afaik.
> > 
> > 
> > I believe so , yes. This should not be needed anymore after #712 . @derekhiggins would you please verify your usecase again with #712 ?
> 
> will do

#712 appears to have fixed the problem I was having.

PR #729, opened 2020-11-23
Add remove local ironic script

kashifest:
Separating the local container removal from run_local_ironic script, so that it can be used in metal3-dev-env also. This simplifies ironic container handling.

### DIFF
--- a/tools/run_local_ironic.sh
+++ b/tools/run_local_ironic.sh
@@ -130,11 +130,7 @@ sudo mkdir -p "$IRONIC_DATA_DIR/html/images"
 # By default, image directory points to dir having needed images when metal3-dev-env environment in use.
 # In other cases user has to store images beforehand.
 
-for name in ironic ironic-api ironic-conductor ironic-inspector dnsmasq httpd mariadb ipa-downloader \
-    ironic-endpoint-keepalived ironic-log-watch ironic-inspector-log-watch; do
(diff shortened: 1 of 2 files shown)


tools/run_local_ironic.sh
```diff
@@ -132,11 +132,8 @@ sudo mkdir -p "$IRONIC_DATA_DIR/html/images"
 # By default, image directory points to dir having needed images when metal3-dev-env environment in use.
 # In other cases user has to store images beforehand.
 
-for name in ironic ironic-api ironic-conductor ironic-inspector dnsmasq httpd mariadb ipa-downloader \
-    ironic-endpoint-keepalived ironic-log-watch ironic-inspector-log-watch; do
-    sudo "${CONTAINER_RUNTIME}" ps | grep -w "$name$" && sudo "${CONTAINER_RUNTIME}" kill "$name"
-    sudo "${CONTAINER_RUNTIME}" ps --all | grep -w "$name$" && sudo "${CONTAINER_RUNTIME}" rm "$name" -f
-done
+# shellcheck disable=SC1091
+source "tools/remove_local_ironic.sh"
```
maelk, line 136: no need to source, you can just run the script, it would remove the shellcheck issue altogether I think

maelk (approved):

tools/run_local_ironic.sh
kashifest: I was sourcing because I wanted to retain the env vars of parent shell

fmuyassarov:
Lgtm

furkatgofurov7 (commented):
/lgtm

PR #730, opened 2020-11-26
[WIP] Limit the number of hosts simultaneously provisioned (BMO only version)

andfasano:
While testing #725 with an high number of nodes I found it wasn't not working as expected, so I deployed an alternative implementation which adopted a different approach in respect to the one originally discussed.

Adding some notes here to provide as much as possible additional details.

---

The problem encountered was an immediate surge when provisioning simultaneously an higher number of nodes, resulting in a breach to the configured threshold: the BMO reconcile loop processed the provisioning requests very rapidly, way before the Ironic subsystem was able to update its status and communicate it back. And things get worse when the BMO concurrency level is increased.

So, this PR bypasses such limitation by observing directly the BMO states instead of the Ironic ones, and by preventing an host to transition in a provisionable state (_Inspecting_ or _Provisioning_) if there isn't a free slot available. This resulted in a simpler and more reliable check.

A very small critical section has been introduced to cope with the BMO concurrency level to handle properly a synchronized counter (a set) tracking the hosts currently being provisioned.

When an host cannot transition to a provisionable state due the lack of free slots, the action is rescheduled with an exponential backoff (by reusing the existing ErrorCount mechanism). A new ErrorType has been introduced to keep track of such situation. The current host state is left unchanged.

A preliminary check is applied when processing an host _already delayed_: if there are no free slots then the request is immediately rescheduled, to avoid putting unnecessary additional stress to the Ironic subsystem.

### DIFF
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -50,9 +51,21 @@ const (
 	unmanagedRetryDelay           = time.Minute * 10
 	provisionerNotReadyRetryDelay = time.Second * 30
 	rebootAnnotationPrefix        = "reboot.metal3.io"
+	maxProvisioningHostsDefault   = 20
 )
 
+var maxProvisioningHosts int
+
 func init() {
+	maxProvisioningHosts = maxProvisioningHostsDefault
+	if maxHostsStr := os.Getenv("MAX_CONCURRENT_PROVISIONING_HOSTS"); maxHostsStr != "" {
+		value, err := strconv.Atoi(maxHostsStr)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Cannot start: Invalid value set for variable MAX_CONCURRENT_PROVISIONING_HOSTS=%s", maxHostsStr)
+			os.Exit(1)
+		}
+		maxProvisioningHosts = value
+	}
 }
 
 // BareMetalHostReconciler reconciles a BareMetalHost object
@@ -61,6 +74,22 @@ type BareMetalHostReconciler struct {
 	Log                logr.Logger
 	Scheme             *runtime.Scheme
 	ProvisionerFactory provisioner.Factory
+
+	currentProvisioningHosts map[string]struct{}
+	mu                       sync.Mutex
+}
+
+// NewBareMetalHostReconciler creates a new reconciler instance
+func NewBareMetalHostReconciler(client client.Client, scheme *runtime.Scheme, factory provisioner.Factory) *BareMetalHostReconciler {
+	r := BareMetalHostReconciler{
+		Client:             client,
+		Log:                ctrl.Log.WithName("controllers").WithName("BareMetalHost"),
+		Scheme:             scheme,
+		ProvisionerFactory: factory,
+
+		currentProvisioningHosts: make(map[string]struct{}),
+	}
+	return &r
 }
 
 // Instead of passing a zillion arguments to the action of a phase,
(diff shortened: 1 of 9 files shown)


controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -914,9 +943,62 @@ func hostHasFinalizer(host *metal3v1alpha1.BareMetalHost) bool {
...
+		return true, actionContinue{}
+	}
+
+	return false, nil
+}
+
 // SetupWithManager reigsters the reconciler to be run by the manager
 func (r *BareMetalHostReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
+	r.syncProvisioningHosts(mgr)
```
zaneb, line 1000: I believe there is a race here; we haven't yet started watching resources or obtained the leader lock, so Hosts can be modified after we build our initial state without us seeing the changes (potentially ever, since our updates are edge-triggered).

zaneb (commented):
It makes me nervous that we are relying on an in-memory cache of states as a source of truth with no mechanism to ensure it is in sync after startup.

This is also not on the path toward likely future enhancements like automatically scaling with the number of conductors available. TBH I'd prefer an approach with this error management but the state being read from ironic more like in #725. The target_provisioning_state should be updated more or less immediately, so if we add that into consideration I wouldn't expect any major capacity overflow problems.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -914,9 +943,62 @@ func hostHasFinalizer(host *metal3v1alpha1.BareMetalHost) bool {
 	return utils.StringInList(host.Finalizers, metal3v1alpha1.BareMetalHostFinalizer)
 }
 
+// Resync the set at the startup
+func (r *BareMetalHostReconciler) syncProvisioningHosts(mgr ctrl.Manager) {
+	hosts := &metal3v1alpha1.BareMetalHostList{}
```
dhellmann, line 948: The controller runtime uses a caching client, which means when we ask for resources that we're subscribed to we look at the cache and get a list back quickly. If we eliminate the mutex management and just write a function that returns a number, it should be plenty fast enough to call every time we start to reconcile.

controllers/metal3.io/host_state_machine.go
```diff
@@ -104,11 +109,22 @@ func (hsm *hostStateMachine) updateHostStateFrom(initialState metal3v1alpha1.Pro
...
 
-func (hsm *hostStateMachine) ReconcileState(info *reconcileInfo) actionResult {
+func (hsm *hostStateMachine) ReconcileState(info *reconcileInfo) (actionRes actionResult) {
 	initialState := hsm.Host.Status.Provisioning.State
-	defer hsm.updateHostStateFrom(initialState, info)
+	defer func() {
+		if !hsm.updateHostStateFrom(initialState, info) {
+			actionRes = recordActionFailure(info, metal3v1alpha1.TooManyHostsError, "Delayed, too many hosts")
+		}
+	}()
```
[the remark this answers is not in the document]
dhellmann: I suggested adding a `transition(newState)` method to the state machine to use instead of assigning the new state directly. By doing that, all of the transitions would pass through one place and we could have a switch statement for handling before & after calls consistently.

controllers/metal3.io/baremetalhost_controller.go
dhellmann: This PR uses the same cache to initialize the data structure, right?

andfasano: Not really, I've the APIReader() returned by the manager to access directly the API server

andfasano:
> It makes me nervous that we are relying on an in-memory cache of states as a source of truth with no mechanism to ensure it is in sync after startup.
> 
> This is also not on the path toward likely future enhancements like automatically scaling with the number of conductors available. TBH I'd prefer an approach with this error management but the state being read from ironic more like in #725. The target_provisioning_state should be updated more or less immediately, so if we add that into consideration I wouldn't expect any major capacity overflow problems.

I understand your concerns and the idea to use as much as possible ironic as the only source of truth. Still remains the problem of the requests on the wire, for which ironic couldn't be used effectively, with the risk of a complete overflow - where instead the cached worked properly. I'll try to work out the best of the two approaches, in order to guarantee the respect of the configured limit and the comments mentioned so far.

andfasano:
Feedback addressed in https://github.com/metal3-io/baremetal-operator/pull/725

PR #732, opened 2020-12-02
Make 'get bmh' output concise and column one-word

Insullone:
Change column 'Provisioning Status' to 'Provisioning_Status'
Change column 'Hardware Profile' to 'Hardware_Profile'
Columns 'Status', 'BMC', 'Hardware_Profile' only visible with '-o wide'


fmuyassarov (commented):
lgtm

zaneb:
Provisioning Status is by far the most useful field IMHO, it's pretty unfortunate that we're now hiding it. I think we should rename it to 'State' and show it by default again.

fmuyassarov:
> (Updated this comment because I misread the patch.)
> I think we should rename the Provisioning_Status to 'State'.

+1 to rename. If none it interested in changing it, I can make a quick fix to that.

PR #733, opened 2020-12-02
Add manifest linting script

fmuyassarov:
Add a script for validation of the k8s core objects within the repo. This is the first step to have a mandatory CI job on every future pull request. The second step will be to configure the prow to run the script.
There are some files such as kustomization patches - for which we need to skip the validation because their deployment definitions are not complete. As such, kubeval will through an error for them. However, we can give a pattern of the files in the same directory that need to be skipped from the check. That's why we add the word `patch` in all the kustomization patches so that we can filter them out with a single pattern, which in this case is  `patch`.


hack/manifestlint.sh
```diff
@@ -0,0 +1,33 @@
...
+if [ "${IS_CONTAINER}" != "false" ]; then
+    kubeval --strict --ignore-missing-schemas \
+    -d config,examples -i kustom,patch -o tap
+else
+  "${CONTAINER_RUNTIME}" run --rm \
+    --env IS_CONTAINER=TRUE \
+    --volume "${PWD}:/workdir:ro,z" \
+    --entrypoint sh \
+    --workdir /workdir \
+    garethr/kubeval:latest \
```
dhellmann, line 31: Is this image on docker hub? Is it going to cause us rate limiting problems?

dhellmann (commented):
The general approach looks good. Could you include the detail from the PR description in the git commit message so when we're trying to understand in 6 months why those files were renamed there's an easily accessible explanation?

One other question inline about the specific image being used.

hack/manifestlint.sh
fmuyassarov: Yes, currently it is in Docker hub. After this question, I realized that are more images that we pull from the Docker hub, apart from what we already listed in the [issue](https://github.com/metal3-io/project-infra/issues/166). So, if I check all the other scripts under the hack/, I can see that we are also pulling:
- registry.hub.docker.com/securego/gosec:latest
- registry.hub.docker.com/pipelinecomponents/markdownlint:latest
- registry.hub.docker.com/koalaman/shellcheck-alpine:stable

Since I haven't seen yet a rate limit issue for the markdown, shellcheck, gosec and etc jobs, I would assume it is going to be the same for this script too. But, it could be that we didn't have yet too many PRs at a time, and that's why we didn't face that issue yet. So, probably we will need to mirror/copy those images too if we are going with that path.

fmuyassarov:
@dhellmann updated the commit message.

/test-integration

hack/manifestlint.sh
dhellmann: I saw you added those to the list in the tracking issue, so I think it's OK to say we will handle mirroring all of those images as one project. I suspect we're not seeing rate limiting errors on those jobs because the CI cluster has the relevant images cached.

dhellmann:
Oh, it would be good to add this to one of the Makefile targets, too.

fmuyassarov:
> Oh, it would be good to add this to one of the Makefile targets, too.

Sure, I will do it.

hack/manifestlint.sh
fmuyassarov: Ah, true. Cashing could be the reason.

fmuyassarov:
PR https://github.com/metal3-io/baremetal-operator/pull/734 is opened to add a Makefile target.

PR #736, opened 2020-12-04
Add console=ttyS0 to kernel params

derekhiggins:
Set console=ttyS0 so that the IPA console can
be logged for debugging.

To be used along with
metal3-io/ironic-image#226


maelk (commented):
/lgtm

PR #737, opened 2020-12-04
Avoid getting stuck in ValidateManagementAccess due to node state

zaneb:
It is not safe to assume in ValidateManagementAccess() that the ironic
state will eventually converge to manageable, available, or active. The
state may in fact be something (like e.g. adopt failed) that requires
action on the part of the actual provisioning function for the current
state to resolve. We previously weren't allowing that, because
ValidateManagementAccess() was always forcing a requeue whenever it
encountered a node in another state.

Since b7b10d79a86bb8e5f04fc86986c7b84fb1a34871 we call
ValidateManagementAccess() in substantially every provisioning state, so
there are even more node states (e.g. inspect failed) that we could get
stuck in.

Move the code for handling the verifying state so that it runs only in
that state, and allow all other states from manageable on to be handled
by the host state machine.

### DIFF
--- a/pkg/provisioner/ironic/validatemanagementaccess_test.go
+++ b/pkg/provisioner/ironic/validatemanagementaccess_test.go
@@ -136,7 +136,7 @@ func TestValidateManagementAccessExistingNode(t *testing.T) {
 	assert.Equal(t, "uuid", host.Status.Provisioning.ID)
 }
 
-func TestValidateManagementAccessExistingNodeStatus(t *testing.T) {
+func TestValidateManagementAccessExistingNodeContinue(t *testing.T) {
 	statuses := []nodes.ProvisionState{
 		nodes.Manageable,
 		nodes.Available,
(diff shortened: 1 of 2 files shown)


pkg/provisioner/ironic/validatemanagementaccess_test.go
```diff
@@ -136,6 +136,71 @@ func TestValidateManagementAccessExistingNode(t *testing.T) {
 	assert.Equal(t, "uuid", host.Status.Provisioning.ID)
 }
 
+func TestValidateManagementAccessExistingNodeStatus(t *testing.T) {
+	statuses := []nodes.ProvisionState{
```
dhellmann, line 140: `Verifying` isn't in this list, and it's going to have a different expected behavior. I don't see another test that covers that case. How many other states are missing here?

zaneb: This is testing all the ones from Manageable on to verify that they don't return Dirty. So only Enroll and Verifying are missing.

zaneb: Added tests for the other 2 states.

dhellmann (approved):
/approve
/lgtm

PR #738, opened 2020-12-05
Add kubebuilder validation for ClockSpeed

s3rj1k:
This small patch adds kubebuilder validation format for ClockSpeed as double.


maelk:
/test-integration
It looks like we have an issue in the integration test regex, it only matches if there is no other keyword before :/ I'll try to fix it when I have a bit of time

s3rj1k:
@maelk  test-integration failed on calico docker image pull

should we retrigger?

dhellmann:
/approve

Thanks!

PR #742, opened 2020-12-07
Check if MaintenanceFlag was already set.

s3rj1k:

### DIFF
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -1198,7 +1198,7 @@ func (p *ironicProvisioner) Provision(hostConf provisioner.HostConfigData) (resu
 	case nodes.CleanFail:
 		if ironicNode.Maintenance {
 			p.log.Info("clearing maintenance flag")
-			return p.setMaintenanceFlag(ironicNode, false)
+			return result, p.setMaintenanceFlag(ironicNode, false)
 		}
 		return p.changeNodeProvisionState(
 			ironicNode,
@@ -1288,7 +1288,15 @@ func (p *ironicProvisioner) Provision(hostConf provisioner.HostConfigData) (resu
 	}
 }
 
-func (p *ironicProvisioner) setMaintenanceFlag(ironicNode *nodes.Node, value bool) (result provisioner.Result, err error) {
+func (p *ironicProvisioner) setMaintenanceFlag(ironicNode *nodes.Node, value bool) (err error) {
+	if ironicNode == nil {
+		panic("ironic node object must by defined")
+	}
+	if ironicNode.Maintenance == value {
+		p.log.Info("maintenance flag already set", "MaintenanceFlag", value)
+		return nil
+	}
+
 	_, err = nodes.Update(
 		p.client,
 		ironicNode.UUID,
@@ -1304,11 +1312,11 @@ func (p *ironicProvisioner) setMaintenanceFlag(ironicNode *nodes.Node, value boo
 	case nil:
 	case gophercloud.ErrDefault409:
 		p.log.Info("could not set host maintenance flag, busy")
+		return HostLockedError{Address: p.host.Spec.BMC.Address}
 	default:
-		return result, errors.Wrap(err, "failed to set host maintenance flag")
+		return errors.Wrap(err, "failed to set host maintenance flag")
 	}
-	result.Dirty = true
-	return result, nil
+	return nil
 }
 
 // Deprovision removes the host from the image. It may be called
@@ -1344,7 +1352,9 @@ func (p *ironicProvisioner) Deprovision() (result provisioner.Result, err error)
 	case nodes.CleanFail:
 		if ironicNode.Maintenance {
 			p.log.Info("clearing maintenance flag")
-			return p.setMaintenanceFlag(ironicNode, false)
+			if err := p.setMaintenanceFlag(ironicNode, false); err != nil {
+				return result, err
+			}
 		}
 		return p.changeNodeProvisionState(
 			ironicNode,
@@ -1431,7 +1441,9 @@ func (p *ironicProvisioner) Delete() (result provisioner.Result, err error) {
 		// delete while bypassing Ironic's internal checks related to
 		// Nova.
 		p.log.Info("setting host maintenance flag to force image delete")
-		return p.setMaintenanceFlag(ironicNode, true)
+		if err := p.setMaintenanceFlag(ironicNode, true); err != nil {
+			return result, err
+		}
 	}
 
 	p.log.Info("host ready to be removed")
--- a/pkg/provisioner/ironic/delete_test.go
+++ b/pkg/provisioner/ironic/delete_test.go
@@ -121,8 +121,7 @@ func TestDelete(t *testing.T) {
 				},
 			).NodeUpdateError(nodeUUID, http.StatusConflict),
 
-			expectedDirty:        true,
-			expectedRequestAfter: 0,
+			expectedError: "BMC test://test.bmc/ host is locked",
 		},
 		{
 			name: "not-in-maintenance-update",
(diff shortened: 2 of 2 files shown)


pkg/provisioner/ironic/ironic.go
```diff
@@ -1289,6 +1289,15 @@ func (p *ironicProvisioner) Provision(hostConf provisioner.HostConfigData) (resu
 }
 
 func (p *ironicProvisioner) setMaintenanceFlag(ironicNode *nodes.Node, value bool) (result provisioner.Result, err error) {
+	if ironicNode == nil {
```
zaneb, line 1292: There no conceivable scenario in which we'd get to this point without having an actual ironic node.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1289,6 +1289,15 @@ func (p *ironicProvisioner) Provision(hostConf provisioner.HostConfigData) (resu
...
+		p.log.Info("no node found, already deleted")
+		return result, nil
+	}
+	if ironicNode.Maintenance == value {
+		p.log.Info("maintenance flag already set", "MaintenanceFlag", value)
+		return result, nil
```
zaneb, line 1298: It's not clear that an empty result is always the right thing to return in these circumstances. That's one reason this is better checked in the caller.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1289,6 +1289,15 @@ func (p *ironicProvisioner) Provision(hostConf provisioner.HostConfigData) (resu
 }
 
 func (p *ironicProvisioner) setMaintenanceFlag(ironicNode *nodes.Node, value bool) (result provisioner.Result, err error) {
+	if ironicNode == nil {
+		p.log.Info("no node found, already deleted")
+		return result, nil
+	}
+	if ironicNode.Maintenance == value {
```
zaneb, line 1296: This method isn't being called anywhere without checking for this first, so what's the problem?

zaneb (commented):
/ok-to-test

pkg/provisioner/ironic/ironic.go
s3rj1k: Code uses ironicNode.UUID in call just below, so if something goes wrong and there is no check for `nil` application would panic.

pkg/provisioner/ironic/ironic.go
s3rj1k: currently in case of error, empty result is also returned, see 
https://github.com/metal3-io/baremetal-operator/blob/master/pkg/provisioner/ironic/ironic.go#L1308
(well it could be empty)

pkg/provisioner/ironic/ironic.go
s3rj1k: Seems so, as of current code, but using this check in local scope of this function ensures that this check will happen, avoids human error and avoids unneeded call to Ironic.

I should probably remove this checks from other places and leave only one here.

What do you think?

pkg/provisioner/ironic/ironic.go
s3rj1k: Could also help in tests

s3rj1k:
@zaneb So what do we do with this, do we continue the effort or abandon it?

pkg/provisioner/ironic/ironic.go
zaneb: It probably _should_ panic, because that's always going to be a bug.

pkg/provisioner/ironic/ironic.go
zaneb: The result is effectively ignored if the error is non-nil, but here you're returning nil for `err` so it matters.

In the places where this is being called, we do something [completely](https://github.com/metal3-io/baremetal-operator/blob/master/pkg/provisioner/ironic/ironic.go#L1203-L1206) [different](https://github.com/metal3-io/baremetal-operator/blob/master/pkg/provisioner/ironic/ironic.go#L1349-L1352) if Maintenance mode is not set.

pkg/provisioner/ironic/ironic.go
zaneb: I think that you won't be able to distinguish this case from others using the return value in the calling code and hence you will find there is no correct way to remove the current checks without adding another return value.

pkg/provisioner/ironic/ironic.go
s3rj1k: I would like to avoid calling `panic`, the whole reason of this line was actually to avoid panic in runtime when  `ironicNode.UUID` accessed and the `ironicNode` is undefined.

pkg/provisioner/ironic/ironic.go
zaneb: > My thoughts where that a case when `ironicNode.Maintenance == value` is a no-op.

If the call is synchronous (is it?) then you could treat it that way if you remove the `result` return value altogether and just return an `error`.

> so just setting `result.Dirty = true` and skipping `nodes.Update(` could be enough?

That would result in an infinite loop if you removed the checks in the callers.

s3rj1k: hmm, so to sum up, any changes to current behavior would only complicate things even more :)

pkg/provisioner/ironic/ironic.go
zaneb: Think of it as an assertion, so if somebody writes some obviously wrong code we will hear about it as soon as possible (preferably in unit tests).

pkg/provisioner/ironic/ironic.go
zaneb: I dug through a week's worth of Slack threads and rediscovered that @dhellmann told me it _is_ synchronous. So ditching the result altogether and just returning an error is an option.

s3rj1k: @zaneb Refactored PR taking into account your latest comment

s3rj1k:
@zaneb Can you please re-review this PR, code changes are finished.
Thanks.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1305,10 +1313,9 @@ func (p *ironicProvisioner) setMaintenanceFlag(ironicNode *nodes.Node, value boo
 	case gophercloud.ErrDefault409:
 		p.log.Info("could not set host maintenance flag, busy")
```
zaneb, line 1314: In this case nothing has been done and we need to retry after a delay, so we can't just return nil.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1344,7 +1351,7 @@ func (p *ironicProvisioner) Deprovision() (result provisioner.Result, err error)
 	case nodes.CleanFail:
 		if ironicNode.Maintenance {
 			p.log.Info("clearing maintenance flag")
-			return p.setMaintenanceFlag(ironicNode, false)
+			return result, p.setMaintenanceFlag(ironicNode, false)
```
zaneb, line 1354: This is wrong, you need to *not* return here unless there is an error, so we can carry on and call `changeNodeProvisionState()`. Since we don't set Dirty=true, we'll never get called again after returning from here.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1431,7 +1438,7 @@ func (p *ironicProvisioner) Delete() (result provisioner.Result, err error) {
 		// delete while bypassing Ironic's internal checks related to
 		// Nova.
 		p.log.Info("setting host maintenance flag to force image delete")
-		return p.setMaintenanceFlag(ironicNode, true)
+		return result, p.setMaintenanceFlag(ironicNode, true)
```
zaneb, line 1441: ditto here

zaneb (commented):
This feels like a lot of effort to expend on a completely imaginary problem.

s3rj1k:
@zaneb Done, please review when possible.

s3rj1k:
@zaneb @dhellmann Any news on this?

PR #745, opened 2020-12-08
Fix issues with deprovisioning after an error/failover

zaneb:
If provisioning fails and the node ends up in the `deploy failed` state,
we try to recover by deprovisioning. That would work better if we
actually handled `deploy failed` in `Deprovision()`.

Only call `Adopt()` if the previous provisioning completed and recorded
the requisite image data.

### DIFF
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -612,26 +612,28 @@ func clearHostProvisioningSettings(host *metal3v1alpha1.BareMetalHost) {
 }
 
 func (r *BareMetalHostReconciler) actionDeprovisioning(prov provisioner.Provisioner, info *reconcileInfo) actionResult {
-	// Adopt the host in case it has been re-registered during the
-	// deprovisioning process before it completed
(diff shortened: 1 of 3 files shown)


controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -563,23 +563,25 @@ func clearHostProvisioningSettings(host *metal3v1alpha1.BareMetalHost) {
...
-	if err != nil {
-		return actionError{err}
-	}
-	if provResult.ErrorMessage != "" {
-		return recordActionFailure(info, metal3v1alpha1.RegistrationError, provResult.ErrorMessage)
-	}
-	if provResult.Dirty {
-		info.host.ClearError()
-		return actionContinue{provResult.RequeueAfter}
+	if info.host.Status.Provisioning.Image.URL != "" {
```
dhellmann, line 566: This is going to have implications for the work @hardys is doing to add live ISO image support, since that PR adds a second URL field to check. I wonder if that means we should have gone with an image mode flag instead of separate fields.

dhellmann (approved):
This looks good.

PR #747, opened 2020-12-09
Filter out status updates from the reconcile loop

andfasano:
Everytime the BMH status is saved the `LastUpdated` field is updated with the current time, and a new update is generated if the marshalled string results different from the previous version (up to the seconds), producing more reconcile loops than needed.
This PR filters out all the updates on the Status by ignoring all those ones that do not increase the Generation field (except for the finalizers and annotations) [1]

[1] https://www.openshift.com/blog/kubernetes-operators-best-practices

### DIFF
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"reflect"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


dhellmann (approved):

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -947,10 +948,11 @@ func (r *BareMetalHostReconciler) SetupWithManager(mgr ctrl.Manager) error {
 						return true
 					}
 
-					if oldHost.Status.ErrorCount != newHost.Status.ErrorCount {
-						//skip reconcile loop
+					//Discard updates that do not increase the resource Generation (such as on Status.LastUpdated), except for the finalizer
+					if e.MetaNew.GetGeneration() == e.MetaOld.GetGeneration() && reflect.DeepEqual(e.MetaNew.GetFinalizers(), e.MetaOld.GetFinalizers()) {
```
zaneb, line 952: We need to make sure we react to annotations changing as well. AIUI that's part of the metadata so it doesn't affect the generation.

zaneb (commented):
Great that you found that doc! I've been wondering about how this was supposed to work since forever.
In BMO I think we are pretty good at always returning Requeue=true when we want to requeue, instead of relying on the status write to trigger it (like some other controllers, e.g. CAPBM, do), so this ought to work. One comment inline.

controllers/metal3.io/baremetalhost_controller.go
andfasano: Thanks, will do

zaneb:
I checked all of the places that Reconcile() returns, and confirmed that the only ones where we're not explicitly requesting a requeue are where we really don't want it.
/lgtm
/test-integration

Issue #739, opened 2020-12-07
BMH stucks in RegistrationError when vbmc is not reachable for short time 


furkatgofurov7:
We have the following scenario:
While BMHs' are in `registering` state, we are running a small script to simulate the scenario, where connection to vbmc will be lost, by simply stoping vbmc container, sleeping for 10 seconds, and starting vbmc again. When simulated, one of the BMHs will be stuck in registering which in turn results in `Registrationerror`. Interestingly, BMO tries to retry and handle the error several times(as `ErrorCount` keeps increasing),  but BMH still is not able to get out of RegistrationError. 
Below provided are truncated version of logs for:
- [Get bmh output](http://paste.openstack.org/show/800786/)
- [Spec of the node in RegistrationError](http://paste.openstack.org/show/800785/)
- [BMO](http://paste.openstack.org/show/800806/)
- [Ironic-api](http://paste.openstack.org/show/800796/)
- [Ironic-conductor](http://paste.openstack.org/show/800797/)
- [Ironic-inspector](http://paste.openstack.org/show/800798/)
- [Ironic-endpoint-keepalived](http://paste.openstack.org/show/800799/)

Any direction/help is much appreciated. Thanks!

Upd: [node_0-serial0.log](http://paste.openstack.org/show/800809/)
Upd2: [baremtal node list](http://paste.openstack.org/show/800925/)

furkatgofurov7:
This issue is the continuation of #708, after testing BMO with changes included #716.

zaneb:
The problem is in the ironic provisioner. Looking at the Node data from ironic, there's no way to tell if there is a `last_error` showing because our attempt to move to manageable just timed out, or if we have seen it before, have waited the requisite time, and want to try again. There's no timestamp in the ironic data (and trusting timestamps from other services is questionable anyway.) Currently we only try again if the credentials have changed, so in all other cases we'll just keep reporting the same error and incrementing the error count.

I think there's something similar going with inspection failures. In `Adopt()` I '[fixed](https://github.com/metal3-io/baremetal-operator/pull/716/commits/324f7b108ce035f32cc879f29b2c7d7749beb07b)' the equivalent problem (#697) by forcing a retry whenever we had an existing error of that type. This doesn't really solve the problem though:

[...]

PR #749, opened 2020-12-11
Do retries with backoff in ValidateManagementAccess, Inspect, and Deprovision

zaneb:
As [discussed](https://github.com/metal3-io/baremetal-operator/issues/739#issuecomment-742218022) in #739, in order to decide whether an error we see from Ironic is one that we have seen and dealt with before (in which case we should retry the present operation) or not (in which case we should deal with the error) we need to store some sort of state.

This PR uses a differential between the ErrorType/ErrorMessage and the ErrorCount to encode this differential. The actual error is cleared whenever we successfully start or continue a new operation, but the error count is preserved until the operation is fully complete. This allows us to both determine when an error is new (when no error is currently recorded in the Status) and yet still do exponential backoff when multiple consecutive errors occur.

This fixes issues with registering, adopting, inspecting, and deprovisioning. The issues were different in each case due to a patchwork of implementations, which are now more consistent.

Some deprovisioning errors that were previously unrecoverable can now be retried. On deletion of the Host, we will retry deprovisioning up to 3 times before giving up and allowing the Host to disappear (previously there was no way to delete it in some states, other than manually removing the finalizer).

One known issue with this is that if a request to change state in Ironic results in a 409 Conflict response, this just sets the Dirty flag and is indistinguishable from success when viewed outside the Ironic provisioner. If such a conflict occurs, we will effectively skip a retry attempt, increment the ErrorCount again, and sleep for even longer before the next attempt.

### DIFF
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -370,10 +370,6 @@ func clearError(host *metal3v1alpha1.BareMetalHost) (dirty bool) {
 		host.Status.ErrorMessage = ""
 		dirty = true
 	}
-	if host.Status.ErrorCount != 0 {
-		host.Status.ErrorCount = 0
-		dirty = true
-	}
 	return dirty
 }
 
@@ -720,6 +716,7 @@ func (r *BareMetalHostReconciler) manageHostPower(prov provisioner.Provisioner, 
 	// state and there were no errors, so reflect the new state in the
 	// host status field.
 	info.host.Status.PoweredOn = info.host.Spec.Online
+	info.host.Status.ErrorCount = 0
 	return steadyStateResult
 }
 
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -604,7 +604,7 @@ func (r *BareMetalHostReconciler) actionDeprovisioning(prov provisioner.Provisio
 
 	info.log.Info("deprovisioning")
 
-	provResult, err = prov.Deprovision()
+	provResult, err = prov.Deprovision(info.host.Status.ErrorType == metal3v1alpha1.ProvisioningError)
 	if err != nil {
 		return actionError{errors.Wrap(err, "failed to deprovision")}
 	}
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -472,7 +472,7 @@ func (r *BareMetalHostReconciler) actionRegistering(prov provisioner.Provisioner
 func (r *BareMetalHostReconciler) actionInspecting(prov provisioner.Provisioner, info *reconcileInfo) actionResult {
 	info.log.Info("inspecting hardware")
 
-	provResult, details, err := prov.InspectHardware()
+	provResult, details, err := prov.InspectHardware(info.host.Status.ErrorType == metal3v1alpha1.InspectionError)
 	if err != nil {
 		return actionError{errors.Wrap(err, "hardware inspection failed")}
 	}
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -434,7 +434,7 @@ func (r *BareMetalHostReconciler) actionRegistering(prov provisioner.Provisioner
 		info.postSaveCallbacks = append(info.postSaveCallbacks, updatedCredentials.Inc)
 	}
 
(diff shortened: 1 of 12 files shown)


furkatgofurov7:
@zaneb thanks a lot for the patch. I will test the patch with the scenario I was simulating with vbmc problem and come back report it here.

andfasano:
> @zaneb thanks a lot for the patch. I will test the patch with the scenario I was simulating with vbmc problem and come back report it here.

hey @furkatgofurov7 could you please share some additional details about the tested scenario? I'd like also to try reproducing it if possible

furkatgofurov7:
Hi @andfasano! sure thing, I have explained it in the [issue](https://github.com/metal3-io/baremetal-operator/issues/739) itself in short, but I will add detailed steps for reproducing it as well.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -409,7 +409,7 @@ func (r *BareMetalHostReconciler) actionRegistering(prov provisioner.Provisioner
 		info.postSaveCallbacks = append(info.postSaveCallbacks, updatedCredentials.Inc)
 	}
 
-	provResult, err := prov.ValidateManagementAccess(credsChanged)
+	provResult, err := prov.ValidateManagementAccess(credsChanged, info.host.Status.ErrorType == metal3v1alpha1.RegistrationError)
```
andfasano, line 412: Minor: what about pushing this check directly within the related provisioner method (ValidateManagementAccess in this case)? It will help in keeping such logic within the provisioner code, and at the same time will minimize the impacts on the interface

controllers/metal3.io/host_state_machine.go
```diff
@@ -240,13 +240,15 @@ func (hsm *hostStateMachine) handleRegistering(info *reconcileInfo) actionResult
 	} else {
 		hsm.NextState = metal3v1alpha1.StateInspecting
 	}
+	hsm.Host.Status.ErrorCount = 0
```
andfasano, line 243: Probably it could make sense to bind the error count reset to the `actionComplete` (maybe with an utility method like `recordActionComplete`), otherwise my feeling is that it could be easily missed next time some new code will be added handling a completed action.

controllers/metal3.io/baremetalhost_controller.go
zaneb: The controller (not the provisioner) owns setting the errors and knowing the types, so conceptually I don't feel like this belongs in the provisioner.

controllers/metal3.io/host_state_machine.go
zaneb: It was nice having it be magic, but there's no current place to hang it - actionComplete doesn't work for the steady states.
The fact that we can't move the code into the state machine for the steady states indicates a design problem, and maybe once that is resolved we'll have a convenient place to tie it.

andfasano: Agree that there's probably something to be reviewed in the steady states, as in the current implementation the ErrorCount is cleared when:
1. When completing an action and moving from/to: 
  * Registering -> Inspecting/Ext.Prov
  * Inspecting -> Match Profile
  * Provisioning -> Provisioned
  * Deprovisioning -> Deleting / Ready
2. Due the power management within the steady states (Ext.Prov.//Provisioned ) and Ready

But at least for point 1 having an utility function that set the actionComplete and resets the ErrorCount could help in reducing the scattering.

controllers/metal3.io/baremetalhost_controller.go
andfasano: The current implementation seems pretty fixed for every case, ie `Adopt` checks always for `RegistrationError`

furkatgofurov7:
> @zaneb thanks a lot for the patch. I will test the patch with the scenario I was simulating with vbmc problem and come back report it here.

Nice work, I have tested the patch locally to test the scenario described in the issue, and it works flawlessly!

furkatgofurov7 (commented):
Looks good to me.

/cc @dhellmann

andfasano:
Looks also good to me

furkatgofurov7:
This needs a rebase though. 

/cc @zaneb

andfasano:
@zaneb @furkatgofurov7 I think it could be useful in this scope also to land #757 as it contains a fix for ErrorCount

zaneb:
Rebased.
/test-integration

furkatgofurov7:
LGTM, leaving it up to @andfasano to give a final lgtm

andfasano:
Thanks @furkatgofurov7, even if late I was fine with them

PR #750, opened 2020-12-14
ironic provisioner de-duplicate image options

hardys:
This aims to de-duplicate the code where ironic image settings are managed, so that new code planned for the LiveImage enhancement don't need to be added in two places

### DIFF
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -420,75 +420,23 @@ func (p *ironicProvisioner) ValidateManagementAccess(credentialsChanged bool) (r
 			imageData = p.host.Spec.Image
 		}
 
-		checksum, checksumType, ok := imageData.GetChecksum()
-
-		if ok {
-			p.log.Info("setting instance info",
-				"image_source", imageData.URL,
-				"image_os_hash_value", checksum,
-				"image_os_hash_algo", checksumType,
-			)
-
-			updates := nodes.UpdateOpts{
-				nodes.UpdateOperation{
-					Op:    nodes.AddOp,
-					Path:  "/instance_info/image_source",
-					Value: imageData.URL,
-				},
-				nodes.UpdateOperation{
-					Op:    nodes.AddOp,
-					Path:  "/instance_info/image_os_hash_value",
-					Value: checksum,
-				},
-				nodes.UpdateOperation{
-					Op:    nodes.AddOp,
-					Path:  "/instance_info/image_os_hash_algo",
-					Value: checksumType,
-				},
-				nodes.UpdateOperation{
-					Op:    nodes.ReplaceOp,
-					Path:  "/instance_uuid",
-					Value: string(p.host.ObjectMeta.UID),
-				},
+		if imageData != nil {
+			updates, err := p.getImageUpdateOptsForNode(ironicNode, imageData)
+			if err != nil {
+				return result, errors.Wrap(err, "Could not get Image options for node")
 			}
-
-			// image_checksum
-			//
-			// FIXME: For older versions of ironic that do not have
-			// https://review.opendev.org/#/c/711816/ failing to
-			// include the 'image_checksum' causes ironic to refuse to
-			// provision the image, even if the other hash value
-			// parameters are given. We only want to do that for MD5,
-			// however, because those versions of ironic only support
-			// MD5 checksums.
-			if checksumType == string(metal3v1alpha1.MD5) {
-				updates = append(
-					updates,
-					nodes.UpdateOperation{
-						Op:    nodes.AddOp,
-						Path:  "/instance_info/image_checksum",
-						Value: checksum,
-					},
-				)
-			}
-
-			if imageData.DiskFormat != nil {
-				updates = append(updates, nodes.UpdateOperation{
-					Op:    nodes.AddOp,
-					Path:  "/instance_info/image_disk_format",
-					Value: *imageData.DiskFormat,
-				})
-			}
-			_, err = nodes.Update(p.client, ironicNode.UUID, updates).Extract()
-			switch err.(type) {
-			case nil:
-			case gophercloud.ErrDefault409:
-				p.log.Info("could not update host settings in ironic, busy")
-				result.Dirty = true
-				result.RequeueAfter = provisionRequeueDelay
-				return result, nil
-			default:
-				return result, errors.Wrap(err, "failed to update host settings in ironic")
+			if len(updates) != 0 {
+				_, err = nodes.Update(p.client, ironicNode.UUID, updates).Extract()
+				switch err.(type) {
+				case nil:
+				case gophercloud.ErrDefault409:
+					p.log.Info("could not update host settings in ironic, busy")
+					result.Dirty = true
+					result.RequeueAfter = provisionRequeueDelay
+					return result, nil
+				default:
+					return result, errors.Wrap(err, "failed to update host settings in ironic")
+				}
 			}
 		}
 	} else {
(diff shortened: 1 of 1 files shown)


pkg/provisioner/ironic/ironic.go
```diff
@@ -775,16 +723,7 @@ func (p *ironicProvisioner) UpdateHardwareState() (result provisioner.Result, er
...
-	if err != nil {
-		return updates, errors.Wrap(err,
-			fmt.Sprintf("Could not start provisioning with bad hardware profile %s",
-				p.host.HardwareProfile()))
-	}
-
+func (p *ironicProvisioner) getImageUpdateOptsForNode(ironicNode *nodes.Node, imageData *metal3v1alpha1.Image) (updates nodes.UpdateOpts, err error) {
 	// image_source
 	var op nodes.UpdateOp
 	if _, ok := ironicNode.InstanceInfo["image_source"]; !ok {
```
dhellmann, line 729: At some point we probably want this check to compare the value, not just see if there is one. That should be another PR, though, since it has bigger implications.

dhellmann (commented):
This looks right to me. I don't know what sort of unit test coverage there is for `getUpdateOptsForNode()` that might need to be changed or expanded, but these changes look fine.

hardys:
> This looks right to me. I don't know what sort of unit test coverage there is for `getUpdateOptsForNode()` that might need to be changed or expanded, but these changes look fine.

Thanks for the quick review - the [test coverage](https://github.com/metal3-io/baremetal-operator/blob/master/pkg/provisioner/ironic/updateopts_test.go) looks OK to me, I could change that to refer to the new function explicitly - but I was thinking the fact that the existing tests all pass was a reasonable way to show that I hadn't broken things with the refactoring :)

hardys:
@dhellmann any more updates needed before we can merge this one?

/cc @zaneb

dhellmann:
I'm happy with these changes. I'll give @zaneb a chance to review them, too.

/approve

pkg/provisioner/ironic/ironic.go
```diff
@@ -420,65 +420,13 @@ func (p *ironicProvisioner) ValidateManagementAccess(credentialsChanged bool) (r
 			imageData = p.host.Spec.Image
 		}
 
-		checksum, checksumType, ok := imageData.GetChecksum()
-
+		_, _, ok := imageData.GetChecksum()
```
zaneb, line 423: I could image moving this into `getImageUpdateOptsForNode()` (where we're already calling it to get the _other_ two return values) at some point, and only checking here whether `updates` is empty.

pkg/provisioner/ironic/ironic.go
```diff
@@ -420,65 +420,13 @@ func (p *ironicProvisioner) ValidateManagementAccess(credentialsChanged bool) (r
...
-				nodes.UpdateOperation{
-					Op:    nodes.AddOp,
-					Path:  "/instance_info/image_os_hash_algo",
-					Value: checksumType,
-				},
-				nodes.UpdateOperation{
-					Op:    nodes.ReplaceOp,
-					Path:  "/instance_uuid",
-					Value: string(p.host.ObjectMeta.UID),
-				},
```
zaneb, line 452: So previously we were setting the instance_uuid here, but now we're not.
I can't recall if that was *required* for adoption (I thought it was?) but it sure makes debugging simpler.

dhellmann: I feel like  it was required for ironic to believe the node was in use.

hardys: Ah thanks, I missed that it got lost here, will fix!

hardys: That's already set in getUpdateOptsForNode - I wonder if we should just call that here - it sets a few more things but we probably want all-the instance_info set on adoption?

zaneb: The rest are only required for actual provisioning. It might not hurt to set them all on adoption, but it might need more work to check that we have all of the info required - currently we set this stuff if there is an image in either the status or the spec, but an image in the spec doesn't guarantee that provisioning has actually started. e.g. it's entirely possible to have an image in the spec but no hardware profile set.
Come to think of it, we probably should be checking that the provision state has at least reached Provisioning before putting the image stuff in there.

hardys: I see, so since `instance_uuid` isn't dependent on the hardware profile, perhaps I should have moved that into `getImageUpdateOptsForNode()`

I see now that in the current code we only set `instance_uuid` when `imageData.GetChecksum()` returns `ok` so moving the `instance_uuid` to `getImageUpdateOptsForNode()` would match the current behavior

pkg/provisioner/ironic/ironic.go
hardys: Yeah I actually did that initially but it breaks a bunch of tests - I could perhaps make that change in a follow-up PR to avoid combining refactoring with test updates and logical changes in the same PR?

pkg/provisioner/ironic/ironic.go
hardys: reinstated `instance_uuid`

hardys:
Ok I fixed the `instance_uuid` issue - from comments here and on #759 there are two potential further de-duplications:

* Avoid duplicate GetChecksum() call
* Potentially just use AddOp for both Add and Replace operations ref https://github.com/metal3-io/baremetal-operator/pull/759#discussion_r555865833 comment from @dtantsur 

Unless folks feel strongly otherwise I'll push follow-up PRs for those, since it's not strictly related to the image instance_info de-duplication I wanted to do before adding the live-iso code?

pkg/provisioner/ironic/ironic.go
zaneb: I have a patch for this but it needs tests: https://github.com/zaneb/baremetal-operator/commit/ed324acfcfb4d3efa77bba7527a575f16f8ca3a8
Feel free to borrow.

pkg/provisioner/ironic/ironic.go
```diff
@@ -420,65 +420,21 @@ func (p *ironicProvisioner) ValidateManagementAccess(credentialsChanged bool) (r
...
+			if err != nil {
+				return result, errors.Wrap(err, "Could not get Image options for node")
+			}
+			updates = append(
+				updates,
 				nodes.UpdateOperation{
 					Op:    nodes.ReplaceOp,
 					Path:  "/instance_uuid",
 					Value: string(p.host.ObjectMeta.UID),
 				},
```
zaneb, line 435: I was picturing that this would be set inside `getImageUpdateOptsForNode()`. It's a thing that always needs to be set at the same time as the `/instance_info` (i.e. before provisioning or adopting), so there's no reason not to keep all the related code together.
Now that I write that, it's suspicious that RootGB is the only `/instance_info` key that is not set in `getImageUpdateOptsForNode()` but left behind in `getUpdateOptsForNode()` (which possibly also suggests another name for this function). I suspect that the reason it hasn't broken adoption is that the [issue in Ironic](https://storyboard.openstack.org/#!/story/2005165) referenced in the comments has loooooong since been fixed, and we should just delete it altogether as thoroughly obsolete (pretty sure we already rely on other Ironic behaviour that is less than 2 years old).

zaneb (commented):
/approve

pkg/provisioner/ironic/ironic.go
hardys: Yeah I was going to add this to `getImageUpdateOptsForNode()` but then noticed is that it's already set in `getImageUpdateOptsForNode()` and we only pass the `p.host.Spec.Image` into that function - if you prefer I can rework the parameters so we can set instance_uuid there, and add a commit to remove the `root_gb`

hardys: Thanks, I moved `instance_info` to `getImageUpdateOptsForNode` - I'll propose a separate PR with the removal of `root_gb`

pkg/provisioner/ironic/ironic.go
hardys: I had a closer look and found a way to avoid breaking the tests, so this is addressed in the latest version

PR #754, opened 2020-12-16
Add LiveImage support

hardys:
First-pass of implementation related to metal3-io/metal3-docs#150


hardys:
This is still WIP as currently the BMH controller isn't triggering provisioning when liveImage is added, and I think we need to handle changing the DeployInterface after inspection as currently it only gets set on nodes.Create, feedback welcome though :)

/cc @sacharya 

Also note this depends on https://github.com/metal3-io/baremetal-operator/pull/750

pkg/provisioner/ironic/ironic.go
```diff
@@ -362,13 +362,19 @@ func (p *ironicProvisioner) ValidateManagementAccess(credentialsChanged bool) (r
 	if ironicNode == nil {
 		p.log.Info("registering host in ironic")
 
+		deployInterface := "direct"
+		if p.host.Spec.LiveImage != nil {
```
dhellmann, line 366: This should check for an empty string, too.

The precedence for the 2 image sources is a little confusing. In `NeedsProvisioning()` the `Image` is checked first so the `LiveImage` is only used if there is no regular image. Here, the `LiveImage` may override the `Image`. Similarly in `imageURL` on line 1076, the live image settings take precedence. However, if both `Image` and `LiveImage` are set, the `switch` statement on line 423 will apply the `Image` (because that rule comes first in the switch statement), and it would incorrectly use the "ramdisk" deploy interface set here.

It might help if we defined a function or two to give us the deploy settings for a host, so we only have to manage the precedence in one place.

pkg/provisioner/ironic/ironic.go
```diff
@@ -775,104 +734,133 @@ func (p *ironicProvisioner) UpdateHardwareState() (result provisioner.Result, er
...
-	hwProf, err := hardware.GetProfile(p.host.HardwareProfile())
-
-	if err != nil {
-		return updates, errors.Wrap(err,
-			fmt.Sprintf("Could not start provisioning with bad hardware profile %s",
-				p.host.HardwareProfile()))
-	}
-
-	// image_source
+func (p *ironicProvisioner) getImageUpdateOptsForNode(ironicNode *nodes.Node, imageData *metal3v1alpha1.Image, liveImageData *metal3v1alpha1.LiveImage) (updates nodes.UpdateOpts, err error) {
```
dhellmann, line 737: Inside this method we have the information we need to derive imageData and liveImageData, so we shouldn't need to pass them as arguments.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1129,6 +1126,30 @@ func (p *ironicProvisioner) Adopt(force bool) (result provisioner.Result, err er
 	return
 }
 
+func (p *ironicProvisioner) ironicHasSameImage(ironicNode *nodes.Node) (sameImage bool) {
+	// To make it easier to test if ironic is configured with
+	// the same image we are trying to provision to the host.
+	if p.host.Spec.LiveImage != nil {
+		sameImage := (ironicNode.InstanceInfo["boot_iso"] == p.host.Spec.LiveImage.URL)
```
dhellmann, line 1133: This `:=` redefines `sameImage`, which I think shadows the named return value. I think you just want `=`.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1129,6 +1126,30 @@ func (p *ironicProvisioner) Adopt(force bool) (result provisioner.Result, err er
...
+		p.log.Info("checking image settings",
+			"boot_iso", ironicNode.InstanceInfo["boot_iso"],
+			"same", sameImage,
+			"provisionState", ironicNode.ProvisionState)
+	} else {
+		checksum, checksumType, _ := p.host.GetImageChecksum()
+		sameImage := (ironicNode.InstanceInfo["image_source"] == p.host.Spec.Image.URL &&
```
dhellmann, line 1140: Same shadowing.

pkg/provisioner/ironic/ironic.go
hardys: Thanks, that's a good idea, I was thinking they should be mutually exclusive so the order didn't matter, but until we actually implement such validation it will - I'll rework to use a function so it's consistent.

pkg/provisioner/ironic/ironic.go
hardys: The reason I did this is because since https://github.com/metal3-io/baremetal-operator/pull/716/commits/2e4e00b2c387d2b35979ecba61f8fb2036bb2fcf from @zaneb the imageData is potentially different depending on where it's called - in `ValidateManagementAccess` it could use the `Status` image on adopt, but when called via setUpForProvisioning->getUpdateOptsForNode we'll only ever want to use the `Spec`?

I wasn't sure if unconditionally doing that check against the status was safe, so I tried to replicate the existing logic, I was worried  about introducing a new corner case in addition to the one @zaneb was fixing..

pkg/provisioner/ironic/ironic.go
hardys: Oops - that's probably related to why it's not working!  

I wonder if we can modify the `got vet` config so we catch this kind of thing ref https://github.com/metal3-io/baremetal-operator/blob/master/.golangci.yml#L18

hardys:
@dhellmann you asked for details of how I'm testing this:

https://github.com/metal3-io/metal3-dev-env/pull/579/ applied and `export IMAGE_OS="FCOS-ISO"` in `config_$USER.sh`

This should result in a locally cached FCOS iso:

```
$ curl -I http://172.22.0.1/images/fedora-coreos-33.20201201.2.1-live.x86_64.iso
HTTP/1.1 200 OK
Date: Fri, 18 Dec 2020 10:17:24 GMT
Server: Apache
Last-Modified: Fri, 18 Dec 2020 10:11:00 GMT
ETag: "2ee00000-5b6ba534537d3"
Accept-Ranges: bytes
Content-Length: 786432000
Content-Type: application/octet-stream
```

Then I'm running `make` and modifying the `/opt/metal3-dev-env/bmhosts_crs.yaml` to add liveImage to one of the BMH resources, e.g:

```
kind: BareMetalHost
metadata:
  name: node-1
spec:
  online: true
  bootMACAddress: 00:8e:b1:1f:8e:4f
  bootMode: legacy
  bmc:
    address: redfish+http://192.168.111.1:8000/redfish/v1/Systems/b85dd1da-439a-4da0-ba9d-686cc0177447
    credentialsName: node-1-bmc-secret
  liveImage:
    url: http://172.22.0.1/images/fedora-coreos-33.20201201.2.1-live.x86_64.iso
```

Then applying the change e.g:

```
$ kubectl apply -f bmhosts_crs.yaml -n metal3
secret/node-0-bmc-secret unchanged
baremetalhost.metal3.io/node-0 unchanged
secret/node-1-bmc-secret unchanged
baremetalhost.metal3.io/node-1 configured
```

Currently that doesn't result in any state change, the BMH remains in `ready` provisioning status, so I must be missing some Image reference somewhere, will keep digging :)

controllers/metal3.io/host_state_machine.go
```diff
@@ -293,16 +293,25 @@ func (hsm *hostStateMachine) provisioningCancelled() bool {
 	if hsm.Host.HasError() {
 		return true
 	}
-	if hsm.Host.Spec.Image == nil {
+	if hsm.Host.Spec.Image == nil && hsm.Host.Spec.LiveImage == nil {
```
hardys, line 296: This logic is very similar to the BMH NeedsProvisioning() - I wonder if we could use the inverse of that here instead of duplicating nearly the same conditionals?

pkg/provisioner/ironic/ironic.go
```diff
@@ -775,103 +729,140 @@ func (p *ironicProvisioner) UpdateHardwareState() (result provisioner.Result, er
...
-		updates,
-		nodes.UpdateOperation{
-			Op:    op,
-			Path:  "/instance_info/image_source",
-			Value: p.host.Spec.Image.URL,
-		},
-	)
-
-	checksum, checksumType, _ := p.host.GetImageChecksum()
+	if imageData != nil {
```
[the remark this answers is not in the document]
hardys: Sure I'll do that - but FWIW the existing logic hasn't changed at all, it's just been indented, but the way the diff is rendered doesn't make that very clear.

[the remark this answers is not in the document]
dhellmann: Yeah, I did eventually work that out, which is why the comment was about the way the diff looked. :-)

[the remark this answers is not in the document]
hardys: Ok I reworked this as described, and it did clarify the diff somewhat, thanks!

hardys:
Ok testing locally this seems to be working now - so I removed the WIP but we need to decide if the case of switching from Image->LiveImage or LiveImage->Image needs to be handled here or in a followup (I've not tested that yet, just adding LiveImage where there is no existing Image

dhellmann:
> Ok testing locally this seems to be working now - so I removed the WIP but we need to decide if the case of switching from Image->LiveImage or LiveImage->Image needs to be handled here or in a followup (I've not tested that yet, just adding LiveImage where there is no existing Image

I'd be happy to have that work in a follow-up PR, to make things easier to review.

pkg/provisioner/ironic/ironic.go
```diff
@@ -775,18 +729,35 @@ func (p *ironicProvisioner) UpdateHardwareState() (result provisioner.Result, er
...
+func (p *ironicProvisioner) getImageUpdateOptsForNode(ironicNode *nodes.Node, imageData *metal3v1alpha1.Image, liveImageData *metal3v1alpha1.LiveImage) (updates nodes.UpdateOpts, err error) {
+	var op nodes.UpdateOp
+	if liveImageData != nil {
+		if _, ok := ironicNode.InstanceInfo["boot_iso"]; !ok {
+			op = nodes.AddOp
+			p.log.Info("adding boot_iso")
+		} else {
+			op = nodes.ReplaceOp
+			p.log.Info("updating boot_iso")
+		}
```
zaneb, line 741: Don't we also need to remove any options corresponding to the other type of image?

controllers/metal3.io/host_state_machine.go
```diff
@@ -293,16 +293,25 @@ func (hsm *hostStateMachine) provisioningCancelled() bool {
...
 		return true
 	}
-	if hsm.Host.Spec.Image.URL == "" {
+	if hsm.Host.Spec.LiveImage != nil && hsm.Host.Spec.LiveImage.URL == "" {
 		return true
 	}
-	if hsm.Host.Status.Provisioning.Image.URL == "" {
+	if hsm.Host.Spec.LiveImage != nil && hsm.Host.Status.Provisioning.LiveImage.URL == "" {
```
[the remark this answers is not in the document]
hardys: Thanks for the review @zaneb - In the docstring I do mention that only Image or LiveImage should be set, but I agree long term they should be validated as mutually exclusive.

Re the ISO flag, that was my original proposal but this approach was preferred since it's possible we might support a live kernel/ramdisk in future, and the checksum argument to the current Image section currently isn't applicable to the LiveImage case: https://github.com/metal3-io/metal3-docs/pull/150#discussion_r520903924

pkg/provisioner/ironic/ironic.go
hardys: Yeah, I think there are several corner cases in the Image->LiveImage and LiveImage->Image case which I've not yet tested or accounted for in the code.

I can attempt to do that in this PR, or as previously mentioned by @dhellmann we might want to land this first iteration then take care of that as a followup, so it's easier to review?

hardys:
Closing in favor of https://github.com/metal3-io/baremetal-operator/pull/759

PR #755, opened 2020-12-17
Improved Provisioner readiness check

andfasano:
This PR introduce a better readiness check for the Ironic provisioner, by checking if the required host driver has been loaded by Ironic or not - instead of just counting the number of drivers loaded. More detail can be found in [1]

[1] https://github.com/metal3-io/ironic-image/pull/233

### DIFF
--- a/pkg/provisioner/ironic/dependencies.go
+++ b/pkg/provisioner/ironic/dependencies.go
@@ -7,7 +7,10 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack/baremetal/v1/drivers"
-	"github.com/gophercloud/gophercloud/pagination"
+	"github.com/pkg/errors"
+
+	metal3v1alpha1 "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
+	"github.com/metal3-io/baremetal-operator/pkg/bmc"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


pkg/bmc/access.go
```diff
@@ -107,21 +109,21 @@ func getParsedURL(address string) (parsedURL *url.URL, err error) {
 
 // NewAccessDetails creates an AccessDetails structure from the URL
 // for a BMC.
-func NewAccessDetails(address string, disableCertificateVerification bool) (AccessDetails, error) {
+func NewAccessDetails(host *metal3v1alpha1.BareMetalHost) (AccessDetails, error) {
```
dhellmann, line 112: We don't want this package to depend on the host API, because this package is used in the OpenShift installer where we don't have a host API.

dhellmann (changes requested):

pkg/bmc/access.go
andfasano: Ok, thanks, will revert the change

pkg/provisioner/ironic/dependencies.go
```diff
@@ -61,30 +67,36 @@ func (i *ironicDependenciesChecker) checkIronic() (ready bool, err error) {
...
+	if pager.Err != nil {
+		return false, pager.Err
+	}
 
+	page, err := pager.AllPages()
 	if err != nil {
-		return ready, err
+		return false, err
+	}
+	drivers, err := drivers.ExtractDrivers(page)
```
derekhiggins, line 88: To avoid confusion I wonder should you rename "drivers" as its also used for the drivers package that is imported

andfasano: Good point, will fix, thanks

andfasano:
@dhellmann could you please recheck this PR? Thanks

dtantsur (changes requested):
You don't need to iterate over drivers, you just need to get the one you expect to see (and handle HTTP 404).

andfasano:
/close 

will require a sligthly different approach to support the case where the provisioner is not required, to be planned into another different PR

Issue #458, opened 2020-03-20
Handle dual-stack configuration for NIC IP


hardys:
https://github.com/metal3-io/baremetal-operator/pull/457 fixes an issue where the IP is not populated for the single-stack IPV6 case, but as mentioned in the review comments it doesn't solve the case where a NIC gets both an ipv4 and ipv6 IP during introspection.

How to handle that from an API perspective may require some discussion, so I won't attempt to fix it in #457 but instead am raising this to track solving the dual-stack case in future.

PR #758, opened 2021-01-05
Handle dual-stack configuration in inspection data

hardys:
Currently we assume only an ipv4 *or* an ipv6 IP exists in the
inspection data, but it's valid for both to exist in a dual-stack
scenario.

To avoid changing the BMH API, in this case we create an additional
nic entry in the list, which will contain each IP but otherwise
the same data.

Fixes: #458


pkg/provisioner/ironic/hardwaredetails/hardwaredetails.go
```diff
@@ -59,24 +59,35 @@ func getNICSpeedGbps(intfExtradata introspection.ExtraHardwareData) (speedGbps i
...
-	nics := make([]metal3v1alpha1.NIC, len(ifdata))
-	for i, intf := range ifdata {
+	var nics []metal3v1alpha1.NIC
+	for _, intf := range ifdata {
 		baseIntf := basedata[intf.Name]
 		vlans, vlanid := getVLANs(baseIntf)
-		ip := intf.IPV4Address
-		if ip == "" {
-			ip = intf.IPV6Address
+		if intf.IPV4Address != "" {
```
dhellmann, line 66: Is it possible for the list of interfaces to include any that do not have addresses? Under the old approach, those would show up in the list without an IP. Now they will not appear in the list at all.

hardys: Ok updated to handle this case, and added a test.

dhellmann:
This version looks good.

/approve

hardys:
> This version looks good.

Thanks - I just pushed a minor update to the test (I forgot to compare the value for the NoIp nic added in the previous revision)

bfournie (commented):
Do you think description here needs to change to indicate a 2nd nic will be used for dual-stack? https://github.com/metal3-io/baremetal-operator/blob/master/config/crd/bases/metal3.io_baremetalhosts.yaml#L343

hardys:
> Do you think description here needs to change to indicate a 2nd nic will be used for dual-stack? https://github.com/metal3-io/baremetal-operator/blob/master/config/crd/bases/metal3.io_baremetalhosts.yaml#L343

Good point, thanks!  I'll push another commit which clarifies that.

bfournie (approved):

PR #759, opened 2021-01-11
Add live-iso support

hardys:
Alternative to #754 

@dhellmann @zaneb PTAL - if you're happy with this interface I'll update the design doc PR

This is working but I'm planning to add some more tests and try to clean up the diff a little before removing WIP

### DIFF
--- a/apis/metal3.io/v1alpha1/baremetalhost_types.go
+++ b/apis/metal3.io/v1alpha1/baremetalhost_types.go
@@ -815,6 +815,12 @@ func (image *Image) GetChecksum() (checksum, checksumType string, ok bool) {
 		return
 	}
 
+	if image.DiskFormat != nil && *image.DiskFormat == "live-iso" {
+		// Checksum is not required for live-iso
+		ok = true
+		return
+	}
+
 	if image.Checksum == "" {
 		// Return empty if checksum is not provided
 		return
--- a/apis/metal3.io/v1alpha1/baremetalhost_types.go
+++ b/apis/metal3.io/v1alpha1/baremetalhost_types.go
@@ -297,15 +297,18 @@ type Image struct {
 	URL string `json:"url"`
 
 	// Checksum is the checksum for the image.
-	Checksum string `json:"checksum"`
+	Checksum string `json:"checksum,omitempty"`
 
 	// ChecksumType is the checksum algorithm for the image.
 	// e.g md5, sha256, sha512
 	ChecksumType ChecksumType `json:"checksumType,omitempty"`
 
-	// DiskFormat contains the format of the image (raw, qcow2, ...)
-	// Needs to be set to raw for raw images streaming
-	// +kubebuilder:validation:Enum=raw;qcow2;vdi;vmdk
+	// DiskFormat contains the format of the image (raw, qcow2, ...).
+	// Needs to be set to raw for raw images streaming.
+	// Note live-iso means an iso referenced by the url will be live-booted
+	// and not deployed to disk, and in this case the checksum options
+	// are not required and if specified will be ignored.
+	// +kubebuilder:validation:Enum=raw;qcow2;vdi;vmdk;live-iso
 	DiskFormat *string `json:"format,omitempty"`
 }
 
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -372,6 +372,7 @@ func (p *ironicProvisioner) ValidateManagementAccess(credentialsChanged, force b
 				BootInterface:       p.bmcAccess.BootInterface(),
 				Name:                p.host.Name,
 				DriverInfo:          driverInfo,
+				DeployInterface:     p.deployInterface(),
 				InspectInterface:    "inspector",
 				ManagementInterface: p.bmcAccess.ManagementInterface(),
 				PowerInterface:      p.bmcAccess.PowerInterface(),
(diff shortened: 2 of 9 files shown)


hardys:
The new interface looks like:

```
  image:
    url: http://172.22.0.1/images/fedora-coreos-33.20201214.2.0-live.x86_64.iso
    format: live-iso
```

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -297,15 +297,18 @@ type Image struct {
...
 	// ChecksumType is the checksum algorithm for the image.
 	// e.g md5, sha256, sha512
 	ChecksumType ChecksumType `json:"checksumType,omitempty"`
 
 	// DiskFormat contains the format of the image (raw, qcow2, ...)
 	// Needs to be set to raw for raw images streaming
-	// +kubebuilder:validation:Enum=raw;qcow2;vdi;vmdk
+	// Note live-iso means an iso referenced by the url will be live-booted
+	// and not deployed to disk, and in this case the checksum options
+	// should be omitted.
```
dhellmann, line 310: How about instead of "should be omitted" saying "will be ignored"?

It would also be nice to clean up the rendered version a bit by adding `.` to the end of the previous 2 sentences.

pkg/provisioner/ironic/ironic.go
```diff
@@ -775,103 +724,153 @@ func (p *ironicProvisioner) UpdateHardwareState() (result provisioner.Result, er
...
+				p.log.Info("removing " + item)
+				updates = append(
+					updates,
+					nodes.UpdateOperation{
+						Op:   op,
+						Path: "/instance_info/" + item,
+					},
+				)
+			}
+		}
```
dhellmann, line 767: I don't see any common code after the if/else block, so I think a `return` here would make the diff easier to follow.

dhellmann (commented):
This implementation certainly looks simpler than the other. I have a few nits inline, but nothing worth holding up the review over.

pkg/provisioner/ironic/ironic.go
```diff
@@ -420,65 +421,13 @@ func (p *ironicProvisioner) ValidateManagementAccess(credentialsChanged bool) (r
...
-					Value: checksum,
-				},
-				nodes.UpdateOperation{
-					Op:    nodes.AddOp,
-					Path:  "/instance_info/image_os_hash_algo",
-					Value: checksumType,
-				},
-				nodes.UpdateOperation{
-					Op:    nodes.ReplaceOp,
-					Path:  "/instance_uuid",
```
dtantsur, line 450: It's not critical, but this item should be kept for every deploy method.

dtantsur, line 450: Not sure where this went, but it should be set for any deployment (although it only matters if we share ironic with something else).

pkg/provisioner/ironic/ironic.go
```diff
@@ -775,103 +724,153 @@ func (p *ironicProvisioner) UpdateHardwareState() (result provisioner.Result, er
...
 	var op nodes.UpdateOp
-	if _, ok := ironicNode.InstanceInfo["image_source"]; !ok {
-		op = nodes.AddOp
-		p.log.Info("adding image_source")
+	if imageData.DiskFormat != nil && *imageData.DiskFormat == "live-iso" {
+		if _, ok := ironicNode.InstanceInfo["boot_iso"]; !ok {
+			op = nodes.AddOp
+			p.log.Info("adding boot_iso")
+		} else {
+			op = nodes.ReplaceOp
```
dtantsur, line 734: A weird property of JSON patch (at least as applied in ironic): add can be used to replace values (but not the other way).

pkg/provisioner/ironic/ironic.go
```diff
@@ -775,103 +724,153 @@ func (p *ironicProvisioner) UpdateHardwareState() (result provisioner.Result, er
...
+				Op:    op,
+				Path:  "/instance_info/boot_iso",
+				Value: imageData.URL,
+			},
+		)
+		updates = append(
+			updates,
+			nodes.UpdateOperation{
+				Op:    nodes.ReplaceOp,
+				Path:  "/deploy_interface",
```
dtantsur, line 749: Assuming nodes are ever reused, you need the opposite action when live ISO is not used (set `deploy_interface` to `direct`).

pkg/provisioner/ironic/updateopts_test.go
```diff
@@ -286,3 +286,73 @@ func TestGetUpdateOptsForNodeDell(t *testing.T) {
...
+			Path:  "/instance_info/boot_iso",
+			Value: "not-empty",
+		},
+		{
+			Path:  "/deploy_interface",
+			Value: "ramdisk",
+		},
+		{
+			Path:  "/instance_uuid",
+			Value: "27720611-e5d1-45d3-ba3a-222dcfaa4ca2",
```
dtantsur, line 324: Ah, so it's not missing. Great.

pkg/provisioner/ironic/ironic.go
hardys: Thanks - yeah that is under discussion on https://github.com/metal3-io/baremetal-operator/pull/750 - hopefully we can work that out then rebase this and the diff will be clearer :)

apis/metal3.io/v1alpha1/baremetalhost_types.go
hardys: Done!

pkg/provisioner/ironic/ironic.go
hardys: Thanks, that sounds like a potential simplification we could address via a followup

pkg/provisioner/ironic/ironic.go
hardys: Ah good point, thanks - done!

pkg/provisioner/ironic/updateopts_test.go
hardys: It's actually set in two places which is why this test didn't catch my removal - I'll look at improving the test coverage as a follow-up.

dhellmann (approved):
This looks good. *MUCH* easier to understand than the version with separate fields, and not harder for the user to consume.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1098,6 +1159,30 @@ func (p *ironicProvisioner) Adopt(force bool) (result provisioner.Result, err er
 	return
 }
 
+func (p *ironicProvisioner) ironicHasSameImage(ironicNode *nodes.Node) (sameImage bool) {
+	// To make it easier to test if ironic is configured with
+	// the same image we are trying to provision to the host.
+	if p.host.Spec.Image != nil && p.host.Spec.Image.DiskFormat != nil && *p.host.Spec.Image.DiskFormat == "live-iso" {
+		sameImage = (ironicNode.InstanceInfo["boot_iso"] == p.host.Spec.Image.URL)
```
zaneb, line 1166: This is not cleared from Ironic when switching back from live-iso to a regular boot, so it may not be operative. I think we should either clear it at the same time as setting `deploy_interface` back to direct, or also verify the `deploy_interface` here, or both.

zaneb (commented):
Wow, this is so much simpler.

imain (approved):
Looks good to me Stephen.

pkg/provisioner/ironic/ironic.go
hardys: Ah thanks, good catch - I updated `getImageUpdateOptsForNode` to account for this, and added tests for both image->iso and iso->image in `updateopts_test.go`

I'm not sure adding an explicit check for deploy_interface here gains us much though, since in that case we expect `*p.host.Spec.Image.DiskFormat` to change from `live-iso` to something else, then we'll hit the `else` and the checksum/image_source values in `ironicNode.InstanceInfo` will always be empty, so sameImage will always evaluate to false AFAICT?

That said, it does raise the question of whether we should check `*p.host.Spec.Image.DiskFormat` since in any other change of disk-format we might want to set sameImage to false, even if the URL is the same?  I guess that's something to potentially address in a different PR though.

zaneb: Agree that doing both is redundant. (Wouldn't hurt, in a belt-and-braces sense, but this is fine too.)

You make a good point though that if provisioning fails because you specified the wrong image format, and you then fix it, it currently won't retry, at least on paper. I'm not actually even 100% sure this code is still needed, since after a failure is first recorded we change the state to Deprovisioning. Definitely a question for elsewhere though.

pkg/provisioner/ironic/updateopts_test.go
hardys: To close the loop on this, it's no longer set in two places after https://github.com/metal3-io/baremetal-operator/commit/177d9efb4d627f6c3d9fae92e0b68c94c06aa713

Issue #721, opened 2020-11-17
User friendly BMH status


Xenwar:
As we have it now, the `kubectl get bmh` output contains information that is too detailed.
The content seems to be the same as what is in the logs. 

It could be improved by changing it to an end user friendly string that is short enough to give relevant information.
more entries are added as comments.

Xenwar:
```
kubectl get bmh -n metal3
NAMESPACE  NAME   STATUS  PROVISIONING STATUS  CONSUMER          BMC
          HARDWARE PROFILE  ONLINE  ERROR
metal3   node-1  error  provisioning     test1-controlplane-mfhtp  redfish+http://192.168.111.1:8000/redfish/v1/Systems/bfdd43ff-3776-495d-b
3f0-fa71c413fa32  unknown      true   Image provisioning failed: Agent returned error for deploy step {'step': 'write_image', 'priority': 80, '
argsinfo': None, 'interface': 'deploy'} on node 14c36cf3-67f6-48c8-b8a8-68a4514dec5c : Error performing deploy_step write_image: Command execution faile
d: Unable to find a valid partition table on the disk after writing the image. Error Unexpected error while running command.
Command: parted -s -m /dev/sda unit MiB print
Exit code: 1
Stdout: 'BYT;\n/dev/sda:51200MiB:scsi:512:512:unknown:QEMU QEMU HARDDISK:;\n'
Stderr: 'Error: /dev/sda: unrecognised disk label\n'.
```

Xenwar:
```
baremetal node show 14c36cf3-67f6-48c8-b8a8-68a4514dec5c -f json | jq
{
  "allocation_uuid": null,
  "automated_clean": null,
  "bios_interface": "redfish",
  "boot_interface": "ipxe",
  "chassis_uuid": null,
  "clean_step": {},
  "conductor": "localhost",
  "conductor_group": "",
  "console_enabled": false,
  "console_interface": "no-console",
  "created_at": "2020-11-17T05:16:22+00:00",
  "deploy_interface": "direct",
  "deploy_step": {},
  "description": null,
  "driver": "redfish",
  "driver_info": {
    "deploy_kernel": "http://172.22.0.2:6180/images/ironic-python-agent.kernel",
    "deploy_ramdisk": "http://172.22.0.2:6180/images/ironic-python-agent.initramfs",
    "redfish_address": "http://192.168.111.1:8000",
    "redfish_password": "******",
[...]

PR #760, opened 2021-01-12
Rename column of 'get bmh' short

Insullone:
Rename column 'Provisioning_Status' to 'State'.


zaneb:
Thanks!

Some other suggestions if you're looking for improvements to make to this output:

* The Status column is pretty near useless. IMHO we could drop it without losing anything.
* If that gave us enough space to restore the Online flag to the default set without overflowing the line, I would be in favour.
* The Error column should just show `.status.errorType` instead of `.status.errorMessage`. See #721.

/approve

PR #761, opened 2021-01-12
Improve communications between controller and provisioner

zaneb:
Use semantic functions (`retryAfterDelay()`, `operationContinuing()`, `operationComplete()`, `operationFailed()`, `transientError()`) to generate return values from ironic provisioner. This is a first step toward changing the `provisioner.Result` type to include richer information (e.g. to be able to distinguish between a request to retry after a delay and the operation simply continuing).

Stop writing to the BareMetalHost Status subresource in the provisioner, and require all changes to be communicated back to the controller via return values from the Provisioner interface.

Only update the BareMetalHost Status subresource when something has changed, instead of on ~every reconcile.

### DIFF
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -434,7 +434,7 @@ func (r *BareMetalHostReconciler) actionRegistering(prov provisioner.Provisioner
 		info.postSaveCallbacks = append(info.postSaveCallbacks, updatedCredentials.Inc)
 	}
 
-	provResult, err := prov.ValidateManagementAccess(credsChanged, info.host.Status.ErrorType == metal3v1alpha1.RegistrationError)
+	provResult, provID, err := prov.ValidateManagementAccess(credsChanged, info.host.Status.ErrorType == metal3v1alpha1.RegistrationError)
 	if err != nil {
 		noManagementAccess.Inc()
 		return actionError{errors.Wrap(err, "failed to validate BMC access")}
@@ -444,6 +444,12 @@ func (r *BareMetalHostReconciler) actionRegistering(prov provisioner.Provisioner
 
 	if provResult.ErrorMessage != "" {
 		return recordActionFailure(info, metal3v1alpha1.RegistrationError, provResult.ErrorMessage)
+	}
+
+	if provID != "" && info.host.Status.Provisioning.ID != provID {
+		info.log.Info("setting provisioning id", "ID", provID)
+		info.host.Status.Provisioning.ID = provID
+		provResult.Dirty = true
 	}
 
 	if provResult.Dirty {
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -644,18 +644,16 @@ func (r *BareMetalHostReconciler) manageHostPower(prov provisioner.Provisioner, 
 	var provResult provisioner.Result
 
 	// Check the current status and save it before trying to update it.
-	provResult, err := prov.UpdateHardwareState()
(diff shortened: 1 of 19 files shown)


controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -419,52 +419,74 @@ func (r *BareMetalHostReconciler) actionUnmanaged(prov provisioner.Provisioner,
...
+		info.log.Info("clearing previous error message")
+		dirty = true
 	}
 
+	if dirty {
+		// Normally returning actionComplete would be sufficient to ensure
+		// changes are written, but when this is called in a state other than
+		// Registering, the state machine will ignore actionComplete and
+		// proceed with the action function for the current state.
+		return actionUpdate{}
```
andfasano, line 488: The comment isn't clear to me, isn't the update driven by the `needsReregister` flag in case of `actionComplete` regardless of the state? Not sure if it could help to return also the `dirty` flag as part of this func

pkg/provisioner/ironic/ironic.go
```diff
@@ -1050,26 +1034,27 @@ func (p *ironicProvisioner) setUpForProvisioning(ironicNode *nodes.Node, hostCon
 	)
 	p.publisher("ProvisioningStarted",
 		fmt.Sprintf("Image provisioning started for %s", p.host.Spec.Image.URL))
-	return
+	// TODO(zaneb): This doesn't make much sense as the overall operation isn't
+	// complete, but it's currently the only return value that the caller can
+	// distinguish from retryAfterDelay().
+	return operationComplete()
```
andfasano, line 1040: Are you referring to `actionProvisioning`?

andfasano (commented):
Apart a couple of minor points/questions, changes look good to me.

I think this approach is much more neat and readable - as the semantic functions express better the intention in the code, and definitely helps in promoting a better separation of the host update logic outside of the Provisioner.

andfasano (approved):

controllers/metal3.io/baremetalhost_controller.go
zaneb: I didn't actually look at the state machine code before writing this comment, so yeah it's more subtle.
The effect is the same though, I think: if this returns actionComplete then ensureRegistered will either return actionContinue or nil (which will proceed with the action function for the current state). In either case nothing gets written to the Status, so we have to return actionUpdate to make sure that writes happen.

pkg/provisioner/ironic/ironic.go
zaneb: I updated this commit to just return here (as it did before) instead of returning operationComplete(), and removed the comment.

controllers/metal3.io/baremetalhost_controller.go
zaneb: I added another commit at the end to rename the method and simplify both it and the caller, since this in no longer an 'action' method proper.

main.go
```diff
@@ -118,21 +118,24 @@ func main() {
 		os.Exit(1)
 	}
 
-	provisionerFactory := func(host *metal3iov1alpha1.BareMetalHost, bmcCreds bmc.Credentials, publish provisioner.EventPublisher) (provisioner.Provisioner, error) {
+	provisionerFactory := func(host metal3iov1alpha1.BareMetalHost, bmcCreds bmc.Credentials, publish provisioner.EventPublisher) (provisioner.Provisioner, error) {
 		isUnmanaged := host.Spec.ExternallyProvisioned && !host.HasBMCDetails()
 
+		hostCopy := host.DeepCopy()
```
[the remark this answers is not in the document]
andfasano: Sounds fine

PR #763, opened 2021-01-14
Add Preparing state to do manual clean

Hellcatlk:
See: https://github.com/metal3-io/baremetal-operator/pull/292#pullrequestreview-566740018

Co-Authored-By: Nguyen Phuong An <annp@vn.fujitsu.com>
Co-Authored-By: Kim Bao Long <longkb@vn.fujitsu.com>
Signed-off-by: zouyu <zouy.fnst@cn.fujitsu.com>


Hellcatlk:
@andfasano @dtantsur @zaneb

pkg/provisioner/ironic/ironic.go
```diff
@@ -1129,6 +1129,79 @@ func (p *ironicProvisioner) Adopt(force bool) (result provisioner.Result, err er
...
+
+	default:
+		// Wait states like cleaning and clean wait
+		p.log.Info("waiting for host to become manageable",
+			"state", ironicNode.ProvisionState,
+			"deploy step", ironicNode.DeployStep)
+		// Avoid start manual cleaning success but set Cleaned flag failed
+		if !p.host.Status.ManualCleaning {
+			p.host.Status.ManualCleaning = true
+		}
```
andfasano, line 1198: This one doesn't make me feel very comfortable, especially because is not performed in a restricted set of states (`cleaning` and `clean wait` as mentioned in the comment). If a cleaning in really in progress, IINW then it could be verified by checking that the `clean_steps` field of the `driver_internal_info` is not empty (see [1]).

[1] https://docs.openstack.org/ironic/latest/admin/cleaning.html

andfasano (commented):
Thanks for this new smaller PR, that is surely more manageable for reviewing.

In general changes looks fine to me, there's only one point requiring a little bit of management in the default case.

In addition, it would be really useful to have a dedicate test for `Prepare` to verify all the various cases (and it will be also a good addition for the review) - hopefully the new mock used to cover other Provisioner parts could be used for that

pkg/provisioner/ironic/ironic.go
```diff
@@ -1129,6 +1129,79 @@ func (p *ironicProvisioner) Adopt(force bool) (result provisioner.Result, err er
...
 
+func (p *ironicProvisioner) startManualCleaning(ironicNode *nodes.Node) (result provisioner.Result, err error) {
+	var cleanSteps []nodes.CleanStep
+
+	// Start manual clean
+	if len(cleanSteps) != 0 {
+		p.log.Info("remove existing configuration and set new configuration", "steps", cleanSteps)
+		opts := nodes.ProvisionStateOpts{Target: nodes.TargetClean, CleanSteps: cleanSteps}
+		var success bool
+		success, result, err = p.tryChangeNodeProvisionState(ironicNode, opts)
```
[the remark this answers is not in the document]
Hellcatlk: Thanks, done.

pkg/provisioner/ironic/ironic.go
Hellcatlk: Thanks, done.

andfasano: `DriverInternalInfo` it's a map that could contain other fields, so very likely that check must be modified.

zaneb (commented):
Thanks for opening this! This really does make it 1000x easier to review.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1129,6 +1129,75 @@ func (p *ironicProvisioner) Adopt(force bool) (result provisioner.Result, err er
...
+		if result.Dirty == true && err == nil {
+			p.host.Status.ManualCleaning = true
+		}
+		return
+
+	case nodes.CleanFail:
+		if p.host.Status.ManualCleaning {
+			p.host.Status.ManualCleaning = false
+			result.Dirty = true
+			return
```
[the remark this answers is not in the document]
andfasano: > AIUI from #725 (comment) the change of Node state from manageable->cleaning is synchronous, so no danger on that front.

AFAIK record update is not synchronous (otherwise the API call will have to wait the db operation before answering). cc @dtantsur

Hellcatlk:
@zaneb PTAL.

demonCoder95:
@Hellcatlk It appears the PR has merge conflicts with master. And thank you for creating this smaller PR, it is definitely easier to handle for the reviewers.

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -557,6 +563,9 @@ type BareMetalHostStatus struct {
 	// ErrorCount records how many times the host has encoutered an error since the last successful operation
 	// +kubebuilder:default:=0
 	ErrorCount int `json:"errorCount"`
+
+	// ManualCleaning records whether manual clean is executed
+	ManualCleaning bool `json:"manualCleaning"`
```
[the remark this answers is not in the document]
demonCoder95: So, you suggest we:

1. Update this flag upon an attempted transition from ``preparing`` to ``ready``, keeping the flag ``true``.
2. We copy over the successfully configured parameters by the current clean step into ``status``. 
3. We then use this ``status`` field to determine if other ``clean_steps`` should be executed (i.e. either to stay in ``preparing`` or move ahead to ``ready`` state.)
4. Upon there being no more ``clean_steps`` to execute, the flag should be set ``false``, which means moving ahead into ``ready`` state.

That sounds reasonable to me. Isn't that what's being implemented right now?

demonCoder95:
/test-integration

@dtantsur @andfasano @zaneb 
Can you please take a last look on this one. We need to get this merged quickly to move onto RAID and firmware configuration functionality we've been meaning to implement for a very long time now. Unfortunately this work have been dragged on too long and it's been very inconvenient especially for the reviewers. 

Thank you!

andfasano (commented):
Apart a couple of minor issues, I think the main point it's about determining the completion of the manual cleanup and avoiding to expose it as a technical/internal detail. I'd agree with @zaneb [comment](https://github.com/metal3-io/baremetal-operator/pull/763#discussion_r558466562) that flag adoption could be problematic - and very likely not be exposed directly to the end user, as it seems more fragile if not carefully managed in all the possible conditions. The suggested approach (working on a Status diff) seems more robust (as probably it could also cover the clean fail) and worth to be explored, especially in the case where it could be reused of other scenarios, like the early cleaning. Given that, I'll leave the final comment to @zaneb

controllers/metal3.io/host_state_machine.go
```diff
@@ -268,17 +269,25 @@ func (hsm *hostStateMachine) handleExternallyProvisioned(info *reconcileInfo) ac
...
 	default:
-		hsm.NextState = metal3v1alpha1.StateReady
+		hsm.NextState = metal3v1alpha1.StatePreparing
 	}
 	return actionComplete{}
 }
 
-func (hsm *hostStateMachine) handleReady(info *reconcileInfo) actionResult {
+func (hsm *hostStateMachine) handlePreparing(info *reconcileInfo) actionResult {
 	if hsm.Host.Spec.ExternallyProvisioned {
```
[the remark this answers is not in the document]
Hellcatlk: .This part comes from `handleReady`, I just moved it to `handPreparing`.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -559,6 +559,28 @@ func (r *BareMetalHostReconciler) actionMatchProfile(prov provisioner.Provisione
...
+	if provResult.ErrorMessage != "" {
+		info.log.Info("handling cleaning error in controller")
+		return recordActionFailure(info, metal3v1alpha1.PreparingError, provResult.ErrorMessage)
+	}
+
+	info.host.Status.ManualCleaning = manualCleaning
+	if provResult.Dirty {
+		clearError(info.host)
+		return actionContinue{provResult.RequeueAfter}
+	}
```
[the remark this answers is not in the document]
Hellcatlk: Thanks, done.

pkg/provisioner/ironic/ironic.go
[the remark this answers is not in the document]
zaneb: > the API call will have to wait the db operation before answering

That's the only sane way to do it, otherwise there's no way to know if your 'accepted' request will ever be acted upon or will just be lost without a trace. Writing the target state to the DB is not a particularly expensive operation.

As far as I can tell this is how it works - the [handler](https://opendev.org/openstack/ironic/src/branch/master/ironic/api/controllers/v1/node.py#L2107) [calls](https://opendev.org/openstack/ironic/src/branch/master/ironic/api/controllers/v1/node.py#L2216-L2218) the RPC [`update_node` method](https://opendev.org/openstack/ironic/src/branch/master/ironic/conductor/rpcapi.py#L233-L234), which [uses synchronous `call`](https://opendev.org/openstack/ironic/src/branch/master/ironic/conductor/rpcapi.py#L257-L258) (not asynchronous `cast`) to get the conductor to [save](https://opendev.org/openstack/ironic/src/branch/master/ironic/conductor/manager.py#L280) the node to the DB. So it's all synchronous.

AFAICT the only part in question is whether the provisioning state gets updated at the same time as we set the target state. I read Dmitry's comment as saying that since we are transitioning from a stable state (manageable) the current state has to be updated as part of that synchronous call.

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -130,6 +130,9 @@ const (
 	// InspectionError is an error condition occurring when an attempt to
 	// obtain hardware details from the Host fails.
 	InspectionError ErrorType = "inspection error"
+	// PreparingError is an error condition occurring when do
+	// cleaning steps failed
+	PreparingError ErrorType = "preparing error"
```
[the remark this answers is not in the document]
zaneb: Yep, comments are updated now but let's change the name to PreparationError for consistency with RegistrationError and InspectionError.

apis/metal3.io/v1alpha1/baremetalhost_types.go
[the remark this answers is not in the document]
zaneb: I think we can do it without adding this flag at all - see comments on actionPrepare below.

zaneb (commented):
Thanks, I think this is getting really close.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -559,6 +559,41 @@ func (r *BareMetalHostReconciler) actionMatchProfile(prov provisioner.Provisione
...
+	if dirty {
+		info.log.Info("updating host provisioning settings")
+		// Restart manual clean by set ManualCleaning flag to false
+		info.host.Status.ManualCleaning = false
+		return actionUpdate{}
+	}
+
+	// Do prepare(manual clean)
+	provResult, manualCleaning, err := prov.Prepare()
```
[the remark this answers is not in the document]
zaneb: The other thing we'd have to do with this is to reset the Status if the node is missing from Ironic (due to it restarting with a fresh DB) in either the Preparing or Ready states.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -792,6 +827,7 @@ func (r *BareMetalHostReconciler) actionManageReady(prov provisioner.Provisioner
 		}
 		if dirty {
 			info.log.Info("updating host provisioning settings")
+			return actionUpdate{}
```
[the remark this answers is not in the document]
Hellcatlk: Because I need different return actionResult to distinguish whether we need to go back to the `Preparing` state.

```go
func (hsm *hostStateMachine) handleReady(info *reconcileInfo) actionResult {
		hsm.NextState = metal3v1alpha1.StateExternallyProvisioned
		return actionComplete{}
	}

	// ErrorCount is cleared when appropriate inside actionManageReady
	actResult := hsm.Reconciler.actionManageReady(hsm.Provisioner, info)
	if _, update := actResult.(actionUpdate); update {
		hsm.NextState = metal3v1alpha1.StatePreparing
	} else if _, complete := actResult.(actionComplete); complete {
		hsm.NextState = metal3v1alpha1.StateProvisioning
	}
	return actResult
}
```

pkg/provisioner/ironic/ironic.go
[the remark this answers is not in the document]
andfasano: There are systems where such behavior could not be acceptable and where the DB tasks are considered expensive. Thanks for pointing out the Ironic's code - IIRC the observed behavior in #725 was that also the current state wasn't update synchronously, but I will recheck it. In any case probably it won't matter anymore if the flag approach will be dropped.

apis/metal3.io/v1alpha1/baremetalhost_types.go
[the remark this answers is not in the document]
demonCoder95: Got it, and I agree with the diff approach as @andfasano also agrees.

pkg/provisioner/ironic/ironic.go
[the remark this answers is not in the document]
dtantsur: A transition to a transient state is synchronous, i.e. after you issue a provisioning request you can expect the node to be in a new state on the next read (with an exception of a potentially quick enroll -> (manage) -> verifying -> (fail) -> enroll process).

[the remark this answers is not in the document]
andfasano: :+1: Thanks for clarifying @dtantsur

zaneb (commented):
Thanks for making these changes. Eliminating the flag in the API makes this much easier to land, because bugs can always be fixed later but API changes are forever :)

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -559,6 +559,42 @@ func (r *BareMetalHostReconciler) actionMatchProfile(prov provisioner.Provisione
...
+		info.log.Info("handling cleaning error in controller")
+		return recordActionFailure(info, metal3v1alpha1.PreparationError, provResult.ErrorMessage)
+	}
+
+	if provResult.Dirty {
+		result := actionContinue{provResult.RequeueAfter}
+		if clearError(info.host) || (dirty && started) {
+			// If clearError return true, but !started, restore provisioningSettings.
+			if dirty && !started {
+				info.host.Status.Provisioning = *provisioningSettings
```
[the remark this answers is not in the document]
Hellcatlk: When clean steps are empty, `started` and `provResult.Dirty ` are false, In this way we can correctly save host provisioning settings and go to `Ready` state.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -559,6 +559,42 @@ func (r *BareMetalHostReconciler) actionMatchProfile(prov provisioner.Provisione
...
+	}
+
+	// Do prepare(manual clean).
+	provResult, started, err := prov.Prepare(dirty)
+	if err != nil {
+		return actionError{errors.Wrap(err, "error preparing host")}
+	}
+
+	if provResult.ErrorMessage != "" {
+		info.log.Info("handling cleaning error in controller")
```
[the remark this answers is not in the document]
Hellcatlk: If I call `clearHostProvisioningSettings()` at the end of `actionDeprovisioning()`, this seems to cause the host's state to be unconditionally changed from "Ready" to "Preparing".

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -528,7 +534,7 @@ type BareMetalHostStatus struct {
 
 	// ErrorType indicates the type of failure encountered when the
 	// OperationalStatus is OperationalStatusError
-	// +kubebuilder:validation:Enum=provisioned registration error;registration error;inspection error;provisioning error;power management error
+	// +kubebuilder:validation:Enum=provisioned registration error;registration error;inspection error;preparing error;provisioning error;power management error
```
zaneb, line 537: `preparation error`

controllers/metal3.io/baremetalhost_controller.go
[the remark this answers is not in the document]
zaneb: > If I call `clearHostProvisioningSettings()` at the end of `actionDeprovisioning()`, this seems to cause the host's state to be unconditionally changed from "Ready" to "Preparing".

Yeah, that's expected and I think fine. We'll always want to redo the RAID config (if any) between provisionings because there's no guarantee that it hasn't been messed up.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1155,6 +1155,82 @@ func (p *ironicProvisioner) ironicHasSameImage(ironicNode *nodes.Node) (sameImag
...
+		if unprepared {
+			result, err = p.changeNodeProvisionState(
+				ironicNode,
+				nodes.ProvisionStateOpts{Target: nodes.TargetManage},
+			)
+		}
+
+	case nodes.Manageable:
+		if unprepared {
+			started, result, err = p.startManualCleaning(ironicNode)
```
zaneb, line 1197: nit: if you changed the order of the return values, you could just `return p.startManualCleaning(ironicNode)`.

controllers/metal3.io/host_state_machine_test.go
```diff
@@ -441,6 +441,10 @@ func (m *mockProvisioner) UpdateHardwareState() (hwState provisioner.HardwareSta
 	return
 }
 
+func (m *mockProvisioner) Prepare(unprepared bool) (result provisioner.Result, started bool, err error) {
+	return m.nextResult, started, err
```
zaneb, line 445: Setting `started = m.nextResult.Dirty` would probably help some future test writer.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1155,6 +1155,80 @@ func (p *ironicProvisioner) ironicHasSameImage(ironicNode *nodes.Node) (sameImag
...
+			ironicNode,
+			nodes.ProvisionStateOpts{Target: nodes.TargetManage},
+		)
+
+	case nodes.Manageable:
+		if unprepared {
+			result, err = p.startManualCleaning(ironicNode)
+			if result.Dirty == true && err == nil {
+				started = true
+			}
```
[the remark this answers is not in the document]
zaneb: > When clean steps are empty, `started` and `provResult.Dirty ` are false, in this case we can correctly save host provisioning settings and go to `Ready` state.

Good point, I forgot that it also saves the settings when Dirty is false.

apis/metal3.io/v1alpha1/baremetalhost_types.go
Hellcatlk: Thanks, done.

zaneb (commented):
This looks basically good to me - just a cosmetic comment inline because one of my earlier suggestions was ambiguous.
I'll approve now and somebody else can review once you've switched that argument order and squashed.
@andfasano this would probably be a good time to take another look over it.
/approve
/label tide/merge-method-squash
/test-integration

Thanks for all your work on this! Not only does this mean that we can hopefully get RAID support in without any more of the constant rebasing, but it's actually a big improvement to the bmo design in general, that opens up a path to make even more improvements in future.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1155,23 +1155,33 @@ func (p *ironicProvisioner) ironicHasSameImage(ironicNode *nodes.Node) (sameImag
...
+		)
 	}
+	result, err = operationComplete()
 	return
 }
 
 // Prepare remove existing configuration and set new configuration.
 // If `started` is true,  it means that we successfully executed `tryChangeNodeProvisionState`.
-func (p *ironicProvisioner) Prepare(unprepared bool) (result provisioner.Result, started bool, err error) {
+func (p *ironicProvisioner) Prepare(unprepared bool) (started bool, result provisioner.Result, err error) {
```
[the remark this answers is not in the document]
Hellcatlk: Thanks, done.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -559,6 +562,44 @@ func (r *BareMetalHostReconciler) actionMatchProfile(prov provisioner.Provisione
 	return actionComplete{}
 }
 
+func (r *BareMetalHostReconciler) actionPreparing(prov provisioner.Provisioner, info *reconcileInfo) actionResult {
+	info.log.Info("preparing")
+
+	// Save provisioning settings.
+	provisioningSettings := info.host.Status.Provisioning.DeepCopy()
+	dirty, err := saveHostProvisioningSettings(info.host)
```
andfasano, line 570: @Hellcatlk, just to clarify, IIUC @zaneb 's [comment](https://docs.google.com/spreadsheets/d/1cyKN434j5FkjuAvSDsBq4IBdGlZNGa6jDAda7bre-rU/edit#gid=188658650) then `saveHostProvisioningSettings` will contains the necessary logic to trigger the manual cleaning in the Prepare. Right now in this method just the `Provisioning.RootDeviceHints` are checked. If it is expected to be enriched in future for the related RAID section, then probably I'd find more clear to add comment for that within such method

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -559,6 +562,44 @@ func (r *BareMetalHostReconciler) actionMatchProfile(prov provisioner.Provisione
...
+		clearHostProvisioningSettings(info.host)
+		return recordActionFailure(info, metal3v1alpha1.PreparationError, provResult.ErrorMessage)
+	}
+
+	if provResult.Dirty {
+		result := actionContinue{provResult.RequeueAfter}
+		if clearError(info.host) || (dirty && started) {
+			// If clearError return true, but started is false, restore provisioningSettings.
+			if dirty && !started {
+				info.host.Status.Provisioning = *provisioningSettings
```
andfasano, line 592: Maybe here a log trace could be useful

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -559,6 +562,44 @@ func (r *BareMetalHostReconciler) actionMatchProfile(prov provisioner.Provisione
...
+	if provResult.ErrorMessage != "" {
+		info.log.Info("handling cleaning error in controller")
+		clearHostProvisioningSettings(info.host)
+		return recordActionFailure(info, metal3v1alpha1.PreparationError, provResult.ErrorMessage)
+	}
+
+	if provResult.Dirty {
+		result := actionContinue{provResult.RequeueAfter}
+		if clearError(info.host) || (dirty && started) {
+			// If clearError return true, but started is false, restore provisioningSettings.
```
andfasano, line 590: This comment is not really clear to me, can you please elaborate it a little bit?

pkg/provisioner/ironic/ironic.go
```diff
@@ -1155,6 +1155,93 @@ func (p *ironicProvisioner) ironicHasSameImage(ironicNode *nodes.Node) (sameImag
...
+			},
+		)
+	}
+	result, err = operationComplete()
+	return
+}
+
+// Prepare remove existing configuration and set new configuration.
+// If `started` is true,  it means that we successfully executed `tryChangeNodeProvisionState`.
+func (p *ironicProvisioner) Prepare(unprepared bool) (result provisioner.Result, started bool, err error) {
```
andfasano, line 1184: What about covering this part using the new mocks? I think that for such relevant feature could be really useful

andfasano (approved):

controllers/metal3.io/host_state_machine.go
```diff
@@ -268,20 +269,31 @@ func (hsm *hostStateMachine) handleExternallyProvisioned(info *reconcileInfo) ac
...
+		hsm.NextState = metal3v1alpha1.StatePreparing
 	}
 	return actionComplete{}
 }
 
+func (hsm *hostStateMachine) handlePreparing(info *reconcileInfo) actionResult {
+	actResult := hsm.Reconciler.actionPreparing(hsm.Provisioner, info)
+	if _, complete := actResult.(actionComplete); complete {
+		hsm.NextState = metal3v1alpha1.StateReady
+	}
```
andfasano, line 281: Here in case of completion it's required to clean the `errorCount`: `hsm.Host.Status.ErrorCount = 0`. This could happen in case previously the clean failed and then later went fine

andfasano (changes requested):

controllers/metal3.io/baremetalhost_controller.go
zaneb: That's correct. But note that this function already existed, and this is the right thing to do regardless of whether RAID is added.

controllers/metal3.io/baremetalhost_controller.go
zaneb: Nothing actually happens here (the provisioning settings remain unchanged from before), so there's nothing interesting to log.

controllers/metal3.io/baremetalhost_controller.go
zaneb: This is just describing how the if statement below can come to be true.

controllers/metal3.io/host_state_machine.go
zaneb: Good catch! Yes.

controllers/metal3.io/baremetalhost_controller.go
andfasano: Right, I was thinking to something more like the `// TODO:` in the `buildManualCleaningSteps`

andfasano (approved):

controllers/metal3.io/host_state_machine.go
andfasano: Note: I switched back my review to approve (as it was originally) since this is the only change I think it was required, my other comments are just suggestions (looks like GitHub collapsed all my comments to the latest review status)

Hellcatlk: Thanks, done.

zaneb: Should be inside the `if ...; complete {` block.

zaneb:
Tests are failing on:
```
errors in package: [/home/prow/go/src/github.com/metal3-io/baremetal-operator/controllers/metal3.io/host_state_machine_test.go:621:11: m.nextResult undefined (type *mockProvisioner has no field or method nextResult) /home/prow/go/src/github.com/metal3-io/baremetal-operator/controllers/metal3.io/host_state_machine_test.go:621:25: m.nextResult undefined (type *mockProvisioner has no field or method nextResult)]"
```

This is due to #725, which just merged this morning, [changing](https://github.com/metal3-io/baremetal-operator/pull/725/commits/2f6b99b97b0423a61300b3ce16233145a8488b20#diff-727164cb2eb71b6771d5d9e2795e400d5619019534ed4949dc321842cd44e867R574) the test code :(

andfasano:
> Tests are failing on:
> 
> ```
> errors in package: [/home/prow/go/src/github.com/metal3-io/baremetal-operator/controllers/metal3.io/host_state_machine_test.go:621:11: m.nextResult undefined (type *mockProvisioner has no field or method nextResult) /home/prow/go/src/github.com/metal3-io/baremetal-operator/controllers/metal3.io/host_state_machine_test.go:621:25: m.nextResult undefined (type *mockProvisioner has no field or method nextResult)]"
> ```
> 
> This is due to #725, which just merged this morning, [changing](https://github.com/metal3-io/baremetal-operator/pull/725/commits/2f6b99b97b0423a61300b3ce16233145a8488b20#diff-727164cb2eb71b6771d5d9e2795e400d5619019534ed4949dc321842cd44e867R574) the test code :(

Hopefully it could be fixed easily by changing the `mockProvisioner.Prepare()` method. There will be anyhow another failure to fix: https://github.com/metal3-io/baremetal-operator/blob/3f79aae54bcead50cee67f8f2b19ecd605e457f2/controllers/metal3.io/host_state_machine_test.go#L382 since it changed to `match-profile -> preparing`, and at this point probably it could be a good idea to add a new case for `preparing -> ready` in the same test.

Hellcatlk:
> > Tests are failing on:
> > ```
> > errors in package: [/home/prow/go/src/github.com/metal3-io/baremetal-operator/controllers/metal3.io/host_state_machine_test.go:621:11: m.nextResult undefined (type *mockProvisioner has no field or method nextResult) /home/prow/go/src/github.com/metal3-io/baremetal-operator/controllers/metal3.io/host_state_machine_test.go:621:25: m.nextResult undefined (type *mockProvisioner has no field or method nextResult)]"
> > ```
> > 
> > 
> > This is due to #725, which just merged this morning, [changing](https://github.com/metal3-io/baremetal-operator/pull/725/commits/2f6b99b97b0423a61300b3ce16233145a8488b20#diff-727164cb2eb71b6771d5d9e2795e400d5619019534ed4949dc321842cd44e867R574) the test code :(
> 
> Hopefully it could be fixed easily by changing the `mockProvisioner.Prepare()` method. There will be anyhow another failure to fix:
> 
> https://github.com/metal3-io/baremetal-operator/blob/3f79aae54bcead50cee67f8f2b19ecd605e457f2/controllers/metal3.io/host_state_machine_test.go#L382
> since it changed to `match-profile -> preparing`, and at this point probably it could be a good idea to add a new case for `preparing -> ready` in the same test.

Done.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1199,6 +1199,93 @@ func (p *ironicProvisioner) ironicHasSameImage(ironicNode *nodes.Node) (sameImag
...
+		)
+
+	case nodes.Cleaning, nodes.CleanWait:
+		p.log.Info("waiting for host to become manageable",
+			"state", ironicNode.ProvisionState,
+			"deploy step", ironicNode.DeployStep)
+		result, err = operationContinuing(provisionRequeueDelay)
+
+	default:
+		result, err = transientError(fmt.Errorf("Have unexpected ironic node state %s", ironicNode.ProvisionState))
```
andfasano, line 1284: Note: during the review I noticed that sometimes some nodes ended up here having the `Provisioning` state still set to `inspectWait`. This is due the fact that in the `Inspection` state BMO decides to move on by looking just at the Inspector result, without waiting for the node to go back to the `manageable` state. We could fix this problem in another PR, since the code here is handling the described scenario without any side effect

andfasano:
/lgtm

Just a final note (@dtantsur @zaneb): this PR modifies the BMO's state machine by adding an extra step for the manual cleaning during the initial phase for provisioning a node. I'm not sure how it will be impacting, but sounds like that this new state could be considered when evaluating the provisioning limit introduced in #725

PR #768, opened 2021-01-18
Get correct image checksum in getImageUpdateOptsForNode()

zaneb:
We always want to get the checksum from the Image passed as imageData.
Due to an inadvertant change during the refactoring that created this
method (177d9efb4d627f6c3d9fae92e0b68c94c06aa713), we were always
getting the checksum from the Image in the Spec, even though imageData
may reference the one in the Status during registration.

This meant that during deprovisioning (with no image in the Spec),
adoption would fail as the node would not have image data added during
registration.


hardys:
Thanks for spotting this, not sure how/why I made that mistake! :thinking: 

/lgtm

PR #769, opened 2021-01-19
Corret shellcheck command

MaXinjian:
There is no vendor dir under this repo, just need to shellcheck all .sh script

Signed-off-by: Ma Xinjian <maxj.fnst@cn.fujitsu.com>


zaneb:
Technically this is correct, as we removed the vendor directory some time back.
That said, there are CI systems that require vendored code to build (i.e. they won't fetch dependencies). We are maintaining a fork of this repo in OpenShift that has this limitation, and it's conceivable that there are other folks in the same boat.

Removing this exception would make it harder for us, and while that's not strictly upstream's problem, as far as I can tell nothing is incorrect and there is no benefit to changing it.

PR #770, opened 2021-01-19
Add test coverage of Spec vs Status Image

hardys:
As seen in the regression CI discovered that prompted
https://github.com/metal3-io/baremetal-operator/pull/768
we're missing some test coverage of the two possible
image sources in ValidateManagementAccess


fmuyassarov (commented):
Thank you!.LGTM

PR #772, opened 2021-01-20
Ironic: Don't adopt after clean failure during deprovisioning

zaneb:
During deprovisioning of a Host, if 'deleting' (i.e. deprovisioning) the
node succeeds (i.e. it doesn't go to the Error state) but the automated
cleaning that follows fails, the only way to recover is to return the
node to the manageable state.

Previously, once in the manageable state we would attempt adoption on
the node so that we could deprovision again. However, in the course of
'deleting' the node, the image information is cleared from it so it
cannot be adopted again. (Adoption continues to be the right thing to do
if the node has just been re-registered due to the Ironic database being
recreated, and in that case the image information is present since it
gets added during the initial registration.)

To work around this, don't attempt to adopt during the Deprovisioning
state if the node is manageable and the image data is not present.
Handle the manageable state in Deprovision() by declaring the
deprovisioning complete.

A node in the manageable state cannot be re-provisioned without first
being cleaned - it must go through cleaning to reach the available state
before it can be provisioned. Provisioning already handles nodes in the
manageable state, as this is how they begin after the initial inspection
of the host before the first provisioning (which does the initial
cleaning).

In the long term it might be better to declare Deprovisioning complete as soon as cleaning _begins_, and move the Host to a new Preparing state which will retry the cleaning until the node can be moved to `available`. In the meantime though, this should resolve the bug.

### DIFF
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -1053,6 +1053,19 @@ func (p *ironicProvisioner) Adopt(force bool) (result provisioner.Result, err er
 		return transientError(fmt.Errorf("Invalid state for adopt: %s",
 			ironicNode.ProvisionState))
 	case nodes.Manageable:
+		_, hasImageSource := ironicNode.InstanceInfo["image_source"]
+		_, hasBootISO := ironicNode.InstanceInfo["boot_iso"]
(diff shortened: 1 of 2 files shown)


furkatgofurov7 (commented):
LGTM

zaneb:
Two reviews is good enough for me :)
/hold cancel

PR #773, opened 2021-01-21
Add BuildTime to manager binary.

s3rj1k:
This PR adds BuildTime variable to manager binary and populates this variable during build via Makefile.

Additional small fixes:
 - use short git commit hash in manager binary
 - use $(OPERATOR_NAME) variable for manager output name (to sync with Dockerfile)
 - do not print empty variables in manager binary version package (Dockerfile does not set them)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,9 +24,12 @@ GOLANGCI_LINT := $(BIN_DIR)/golangci-lint
 KUSTOMIZE := $(BIN_DIR)/kustomize
 
 # See pkg/version.go for details
-SOURCE_GIT_COMMIT ?= $(shell git rev-parse --verify 'HEAD^{commit}')
+SOURCE_GIT_COMMIT ?= $(shell git rev-parse --short HEAD)
 BUILD_VERSION ?= $(shell git describe --always --abbrev=40 --dirty)
-export LDFLAGS="-X github.com/metal3-io/baremetal-operator/pkg/version.Raw=${BUILD_VERSION} -X github.com/metal3-io/baremetal-operator/pkg/version.Commit=${SOURCE_GIT_COMMIT}"
+VERSION_URI = "github.com/metal3-io/baremetal-operator/pkg/version"
+export LDFLAGS="-X $(VERSION_URI).Raw=${BUILD_VERSION} \
+                -X $(VERSION_URI).Commit=${SOURCE_GIT_COMMIT} \
+                -X $(VERSION_URI).BuildTime=$(shell date +%Y-%m-%dT%H:%M:%S%z)"
 
 # Set some variables the operator expects to have in order to work
 # Those need to be the same as in config/default/ironic.env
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -24,9 +24,12 @@ GOLANGCI_LINT := $(BIN_DIR)/golangci-lint
...
 
 # See pkg/version.go for details
-SOURCE_GIT_COMMIT ?= $(shell git rev-parse --verify 'HEAD^{commit}')
+SOURCE_GIT_COMMIT ?= $(shell git rev-parse --short HEAD)
 BUILD_VERSION ?= $(shell git describe --always --abbrev=40 --dirty)
-export LDFLAGS="-X github.com/metal3-io/baremetal-operator/pkg/version.Raw=${BUILD_VERSION} -X github.com/metal3-io/baremetal-operator/pkg/version.Commit=${SOURCE_GIT_COMMIT}"
+VERSION_URI = "github.com/metal3-io/baremetal-operator/pkg/version"
+export LDFLAGS="-X $(VERSION_URI).Raw=${BUILD_VERSION} \
+                -X $(VERSION_URI).Commit=${SOURCE_GIT_COMMIT} \
+                -X $(VERSION_URI).BuildTime=$(shell date +%Y-%m-%dT%H:%M:%S%z)"
```
dhellmann, line 32: Is the important thing the time we ran the compiler, or the time associated with the commit that was built?

main.go
```diff
@@ -57,7 +57,13 @@ func init() {
 func printVersion() {
 	setupLog.Info(fmt.Sprintf("Go Version: %s", runtime.Version()))
 	setupLog.Info(fmt.Sprintf("Go OS/Arch: %s/%s", runtime.GOOS, runtime.GOARCH))
-	setupLog.Info(fmt.Sprintf("baremetal-operator version: %s", version.String))
+	if len(version.Commit) > 0 {
```
dhellmann, line 60: Maybe we can default those strings to something like "unset" so we always have something to print, and it's obvious when a value wasn't provided.

Makefile
s3rj1k: I was thinking about a compile time as you can rebuild same commit multiple times and knowing compile time could differentiate those builds.

main.go
s3rj1k: Sure, would UNSET in caps work? Capitals are for a user to notice that those values where not defined.

s3rj1k: Rebased with `unknown`, looks better in my opinion.

Makefile
dhellmann: That makes some sense. I would expect the results to be essentially the same for the same commit, but these variables are evidence that it might not be.

dhellmann (commented):
/approve

s3rj1k:
@dhellmann I think integration test did not trigger

s3rj1k:
Are we safe to merge?

s3rj1k:
@dhellmann ?

PR #780, opened 2021-01-26
Fail registration when boot MAC conflicts

honza:
If we can't find a node by name, and have to resort to using MAC address
lookup, we can't accept a node that has a name.  In that scenario, we
have detected a MAC address collision, and human intervention is
required.  We signal this by returning an `operationFailed` provisioning
result with the offending MAC address in the error message.

### DIFF
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -60,6 +60,19 @@ var bootModeCapabilities = map[metal3v1alpha1.BootMode]string{
 	metal3v1alpha1.Legacy: "boot_mode:bios",
 }
 
+type macAddressConflictError struct {
+	Address      string
+	ExistingNode string
+}
+
+func (e macAddressConflictError) Error() string {
+	return fmt.Sprintf("MAC address %s conflicts with existing node %s", e.Address, e.ExistingNode)
+}
+
+func NewMacAddressConflictError(address, node string) error {
+	return macAddressConflictError{Address: address, ExistingNode: node}
+}
+
 func init() {
 	// NOTE(dhellmann): Use Fprintf() to report errors instead of
 	// logging, because logging is not configured yet in init().
@@ -310,7 +323,7 @@ func (p *ironicProvisioner) findExistingHost() (ironicNode *nodes.Node, err erro
 
 			// If the node has a name, this means we didn't find it above.
 			if ironicNode.Name != "" {
-				return nil, errors.New(fmt.Sprint("node found by MAC but has a name: ", ironicNode.Name))
+				return nil, NewMacAddressConflictError(p.host.Spec.BootMACAddress, ironicNode.Name)
 			}
 
 			return ironicNode, nil
@@ -342,7 +355,12 @@ func (p *ironicProvisioner) ValidateManagementAccess(credentialsChanged, force b
 
 	ironicNode, err = p.findExistingHost()
 	if err != nil {
-		result, err = transientError(errors.Wrap(err, "failed to find existing host"))
+		switch err.(type) {
+		case macAddressConflictError:
+			result, err = operationFailed(err.Error())
+		default:
+			result, err = transientError(errors.Wrap(err, "failed to find existing host"))
+		}
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/provisioner/ironic/ironic.go
```diff
@@ -310,7 +322,7 @@ func (p *ironicProvisioner) findExistingHost() (ironicNode *nodes.Node, err erro
 
 			// If the node has a name, this means we didn't find it above.
 			if ironicNode.Name != "" {
-				return nil, errors.New(fmt.Sprint("node found by MAC but has a name: ", ironicNode.Name))
+				return nil, NewMacAddressConflict(p.host.Spec.BootMACAddress)
```
dhellmann, line 325: Can we provide enough information to tell which other host has the same MAC?

honza: We can add the name of the node we found, good idea

pkg/provisioner/ironic/ironic.go
```diff
@@ -60,6 +60,19 @@ var bootModeCapabilities = map[metal3v1alpha1.BootMode]string{
 	metal3v1alpha1.Legacy: "boot_mode:bios",
 }
 
+type macAddressConflict struct {
```
andfasano, line 63: nit: macAddressConflictError

pkg/provisioner/ironic/ironic.go
```diff
@@ -342,7 +355,12 @@ func (p *ironicProvisioner) ValidateManagementAccess(credentialsChanged, force b
 
 	ironicNode, err = p.findExistingHost()
 	if err != nil {
-		result, err = transientError(errors.Wrap(err, "failed to find existing host"))
+		switch err.(type) {
+		case *macAddressConflictError:
```
andfasano, line 359: It would be nice here to use the error manipulations functions like `errors.As()` to check the error type

pkg/provisioner/ironic/validatemanagementaccess_test.go
```diff
@@ -574,8 +574,8 @@ func TestValidateManagementAccessExistingPortButHasName(t *testing.T) {
 		t.Fatalf("could not create provisioner: %s", err)
 	}
 
-	_, _, err = prov.ValidateManagementAccess(false, false)
-	assert.EqualError(t, err, "failed to find existing host: node found by MAC but has a name: wrong-name")
+	_, err = prov.findExistingHost()
+	assert.EqualError(t, err, "MAC address 11:11:11:11:11:11 conflicts with existing node wrong-name")
 }
```
andfasano, line 580: It's not clear to me why here the test switched to `findExistingHost`. I would consider better, in this suite, to continue to verify `ValidateManagementAccess` behavior, also because it's part of the public interface, ie something like:
```

res, _, err := prov.ValidateManagementAccess(false, false)
assert.Nil(t, err)
assert.Equal(t, res.ErrorMessage, "MAC address 11:11:11:11:11:11 conflicts with existing node wrong-name")

```

pkg/provisioner/ironic/ironic.go
honza: I think switching on the error type is the common way of doing this in this project.

andfasano: That's right, I suggested that because since 1.13 the new error wrapping/manipulation functions have been introduced (see [1]). That's something anyhow that we could review more generally in another PR in case.

[1] https://golang.org/doc/go1.13#error_wrapping
[2] https://golang.org/pkg/errors/

pkg/provisioner/ironic/ironic.go
```diff
@@ -60,6 +60,19 @@ var bootModeCapabilities = map[metal3v1alpha1.BootMode]string{
 	metal3v1alpha1.Legacy: "boot_mode:bios",
 }
 
+type macAddressConflictError struct {
+	Address      string
+	ExistingNode string
+}
+
+func (e *macAddressConflictError) Error() string {
```
zaneb, line 68: nit: I don't think this has to be a pointer receiver (since you're not modifying it); removing the * would allow you to ditch all of the other * and & stuff (and the NewMacAddressConflictError function, whose only real purpose is adding the &).

PR #781, opened 2021-01-27
Parameterize docker registry

Xenwar:
This PR parameterizes the docker registry for building images.
Which registry is used determined by the following variable, with the give default value.
Exporting the variables per Dockerfile allows this repo to be independent of `metal3-dev-env`
This has an effect when using `Makefile` and  the value is overridden. 

`export OVERRIDE_DOCKER_HUB_REGISTRY=${OVERRIDE_DOCKER_HUB_REGISTRY:-"registry.hub.docker.com"}
`

If you wish to use a different registry, please assign value in a `HOST:PORT/path` format.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,7 @@
 # Build the manager binary
-FROM registry.hub.docker.com/library/golang:1.15.3 AS builder
+ARG OVERRIDE_DOCKER_HUB_REGISTRY=${OVERRIDE_DOCKER_HUB_REGISTRY:-"registry.hub.docker.com"}
+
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM registry.hub.docker.com/library/golang:1.15.3 AS builder
+FROM library/golang:1.15.3 AS builder
```
fmuyassarov, line 2: I think that's the only one that we use in Metal3-dev-env CI tests.

Xenwar: Thanks. Only Dockerfiles are changed now.

smoshiur1237:
Looks good to me.

Xenwar:
replaced by an alternative solution.

Issue #626, opened 2020-09-07
Configurable disk cleaning options


kashifest:
This is a feature request. The user story is as follows: 

**User story:**
As a user, I want to clean up the disks associated with a node in a configurable way. For example, during the initial provisioning I want all the disks of a particular node to be cleaned. However, during an upgrade procedure, I want to clean only the root disk associated with that node and leave the other disks uncleaned as we have data on it and we don't want it to be erased. I want to communicate this information to ironic through the bmh CR. 

This will also require ironic to understand the request coming from BMO to clean disks associated with a node in a configurable way. A similar feature request towards ironic is also being requested. Note that, the default cleaning path will still be the same i.e. cleaning all the disks as it is done today. We only want to make the cleaning configurable runtime. On BMO side, the change foreseen is an api change to communicate the desired configuration to ironic, the configuration being either clean root disk only or clean all the disks.

dtantsur:
An extract from the IRC discussion:
```
13:14 <fmuyassarov> dtantsur, It is coming from upgrade use-case. For example, when we do the k8s cluster upgrade the nodes will be deprovisioned & provisioned. So, we don’t want to lose disk data. However, in normal scenarios we still want to make sure that while provisioning we do cleaning for all the disk. One way to do it is to have two ironic images (one for upgrade, another for normal use cases) with two differen
13:14 <fmuyassarov> t configurations, but I don’t think it is a good idea to have two ironic containers running.
13:15 <dtantsur> fmuyassarov: what's the reason of doing upgrades via reprovisioning?
13:29 <fmuyassarov> dtantsur, Because, we are using the Cluster API KCP (KubeadmControlPlane) upgrade feature.  And while upgrading, KCP deletes a Machine which will cause Metal3Machine deletion and finally BareMetalHost deprovisioning. So, AFAIK Metal3 doesn’t have a way to control that behavior since it happens on a higher level.
13:29 <fmuyassarov> So, it deleted a Machine and recreates a new one.
13:48 <dtantsur> sigh
13:48 <dtantsur> I wonder if we can somehow hold it from actual deprovisioning?
13:49 <dtantsur> in addition to cleaning problems, it's also going to be time-consuming
13:53 <fmuyassarov> TBH I don't think that there is a way to avoid a BMH going through deprovisioning since KCP deletes the associated Machine object.
13:56 <dtantsur> fmuyassarov: yeah, but we could keep BareMetalHost in some new "pending" state until it gets picked again?
[...]

PR #784, opened 2021-02-02
Allow disable automated cleaning

fmuyassarov:
Disabling [disk cleaning](https://github.com/metal3-io/metal3-docs/blob/master/design/cluster-api-provider-metal3/allow_disabling_node_disk_cleaning.md) implementation.
Fixes: https://github.com/metal3-io/baremetal-operator/issues/626
Related changes in CAPM3: https://github.com/metal3-io/cluster-api-provider-metal3/pull/156


config/crd/bases/metal3.io_baremetalhosts.yaml
```diff
@@ -121,6 +121,10 @@ spec:
               description:
                 description: Description is a human-entered text used to help identify the host
                 type: string
+              disableAutomatedClean:
```
hardys, line 124: I know the design for this landed already, but are we OK with using an annotation for disabling inspection ref metal3-io/metal3-docs#155 and #787, but disabling cleaning via the BMH spec?

pkg/provisioner/ironic/ironic.go
```diff
@@ -1212,6 +1212,37 @@ func (p *ironicProvisioner) Provision(hostConf provisioner.HostConfigData) (resu
 		return transientError(provisioner.NeedsRegistration)
 	}
 
+	p.log.Info("WARNING: automated cleaning is disabled for host",
```
hardys, line 1215: Is this actually a warning?  Seems like it's just Info logging, since the behavior is what was requested?

In #787 I did also generate an event though, so it's visible via the BMH `describe` output, we may want the same in this case?

pkg/provisioner/ironic/ironic.go
```diff
@@ -1377,6 +1408,38 @@ func (p *ironicProvisioner) Deprovision(force bool) (result provisioner.Result,
...
```
hardys, line 1411: Seems like we should refactor this into a function, given that the Provision/Deprovision code is the same?

pkg/provisioner/ironic/ironic.go
kashifest: Shouldn't this logging be inside some check?

pkg/provisioner/ironic/ironic.go
kashifest: Same here . The log is coming before the check

kashifest (changes requested):
Looks Good. Minor comments in line. +1 for the idea of moving the code peace into a function

config/crd/bases/metal3.io_baremetalhosts.yaml
hardys: @dhellmann any thoughts on that?  I'm not blocking on it, just wondering if we have any guidelines on what should be in the spec vs annotations?

In practice it seems like annotations are used to modify the baremetalhost_controller behavior, but the spec is used to influence the provisioner configuration, but that seems like an implementation detail rather than an API design pattern :)

pkg/provisioner/ironic/ironic.go
fmuyassarov: @hardys , @kashifest agree with both points. Updating. Thanks.

pkg/provisioner/ironic/ironic.go
furkatgofurov7: +1 for refactoring it into a function.

config/crd/bases/metal3.io_baremetalhosts.yaml
dhellmann: I'm not sure why we used an annotation for disabling inspection. Maybe because it wasn't really a thing we thought most users would (or should) do? Or because the semantics of changing an API field to enable inspection after a host was provisioned were confusing?

An API field to turn off cleaning seems reasonable, and the semantics of changing it later are clearer (at least to me). I could be convinced that the inspection and cleaning features should be treated the same way, as API fields.

I should have provided this feedback on the design review, but I think we don't want to use a boolean field. It's harder to manage changes in behavior. For example, today cleaning is all or none. In the future, it may be possible to clean some disks and leave others alone. We may want a 3rd mode to cover that case. Creating this field as an enum now would let us make that change relatively easily, without an API version change and upgrade process. So, I suggest a name like `automatedCleaningMode` with values `Full` and `None` instead of a bool here.

hardys: > I'm not sure why we used an annotation for disabling inspection.

I did it that way to align with the re-introspection proposal that was already accepted : https://github.com/metal3-io/metal3-docs/blob/master/design/baremetal-operator/re-introspection-interface.md

> Maybe because it wasn't really a thing we thought most users would (or should) do? 

This is probably also true, the main use-case we have for it atm is when booting a live-iso, which likely isn't the most common path, and we might encourage folks to use a visible API in the spec for the wrong reasons (e.g faster deployments, but the same could be said for this API I guess)

dhellmann: > > I'm not sure why we used an annotation for disabling inspection.
> 
> I did it that way to align with the re-introspection proposal that was already accepted : https://github.com/metal3-io/metal3-docs/blob/master/design/baremetal-operator/re-introspection-interface.md

Ah, good point. In that case the annotation is removed after the re-inspection is done, right? I suppose we could have had the operator toggle a flag that's part of the schema instead. But we probably based the re-inspection thing on the reboot API, and that was also based on an annotation.

> > Maybe because it wasn't really a thing we thought most users would (or should) do?
> 
> This is probably also true, the main use-case we have for it atm is when booting a live-iso, which likely isn't the most common path, and we might encourage folks to use a visible API in the spec for the wrong reasons (e.g faster deployments, but the same could be said for this API I guess)

Yeah, true. I have seen recommendations for using annotations for interactions between components (such as an annotation on a host that is consumed by something other than the BMO) or as early implementations of APIs that aren't ready to be locked into a CRD. That last use feels vague and fuzzy, but I think that's how I justified the reboot, re-inspection, and skip-inspection changes in my mind. Eventually, those *should* become part of the spec fields, when we figure out how to do that. This change seems to fit easily enough into the API fields that we can move right to that implementation. But I'll go with the group's consensus, so if others would be more comfortable using an annotation, that's OK with me.

fmuyassarov: > > > I'm not sure why we used an annotation for disabling inspection.
> > 
> > 
> > I did it that way to align with the re-introspection proposal that was already accepted : https://github.com/metal3-io/metal3-docs/blob/master/design/baremetal-operator/re-introspection-interface.md
> 
> Ah, good point. In that case the annotation is removed after the re-inspection is done, right? 

That's right. The implementation PR is not merged yet though https://github.com/metal3-io/baremetal-operator/pull/607

>I suppose we could have had the operator toggle a flag that's part of the schema instead. But we probably based the re-inspection thing on the reboot API, and that was also based on an annotation.
> 
That's true. Re-inspection was designed to be similar interface as reboot API.

fmuyassarov: > I should have provided this feedback on the design review, but I think we don't want to use a boolean field. It's harder to manage changes in behavior. For example, today cleaning is all or none. In the future, it may be possible to clean some disks and leave others alone. We may want a 3rd mode to cover that case. Creating this field as an enum now would let us make that change relatively easily, without an API version change and upgrade process. So, I suggest a name like `automatedCleaningMode` with values `Full` and `None` instead of a bool here. We can update the design doc after the code is approved.

@dhellmann Thanks, I agree with your point here and can change the both code + proposal. But are we okay to have enum here considering Ironic `Node.AutomatedClean` is bool. I know `automatedCleaningMode` doesn't have to be the same type as  `Node.AutomatedClean`, but just asking if there could be any issue with that.

hardys: > But I'll go with the group's consensus, so if others would be more comfortable using an annotation, that's OK with me.

+1 I'm fine with this going into the spec, I just wanted to ensure we're all OK with that approach :)

And I agree that in the case of the inspection API it's fine to use an annotation for consistency (it'd be confusing to have both annotations and spec fields modifying that behavior IMO), and that's definitely an API which might see further changes in future, so it makes sense to avoid committing to a spec API now.

dhellmann: > @dhellmann Thanks, I agree with your point here and can change the both code + proposal. But are we okay to have enum here considering Ironic `Node.AutomatedClean` is bool. I know `automatedCleaningMode` doesn't have to be the same type as `Node.AutomatedClean`, but just asking if there could be any issue with that.

I think it's OK for the APIs to be different. We've never made the assumption that metal3 users are Ironic users, and we try to design our APIs following kubernetes standards and conventions.

fmuyassarov:
Hi @dhellmann @hardys @kashifest 
Thanks for the reviews once more. I have updated the code as requested and opened a PR [#166](https://github.com/metal3-io/metal3-docs/pull/166) to update proposal in metal3-docs. PTAL.

fmuyassarov:
@dhellmann Please have look, updated as per the latest changes in the proposal.

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -278,6 +278,14 @@ type BareMetalHostSpec struct {
...
 	// Image field is filled in, this field is ignored.
 	ExternallyProvisioned bool `json:"externallyProvisioned,omitempty"`
+
+	// When set to None, automated cleaning will be avoided
+	// during provisioning and deprovisioning.
+	// +optional
+	// +kubebuilder:default:=Full
+	// +kubebuilder:validation:Enum:=Full;Disabled
+	// +kubebuilder:validation:Optional
+	AutomatedCleaningMode string `json:"automatedCleaningMode,omitempty"`
```
dhellmann, line 288: For `BootMode` we define a type and set the kubebuilder options on that type. Can we do the same for this enum? That would let us have `const` values for comparisons, too.

pkg/provisioner/ironic/ironic.go
```diff
@@ -353,6 +353,44 @@ func (p *ironicProvisioner) findExistingHost() (ironicNode *nodes.Node, err erro
 
 }
 
+func (p *ironicProvisioner) updateAutomatedClean(ironicNode *nodes.Node) (err error) {
+	var updates nodes.UpdateOpts
+
+	if p.host.Spec.AutomatedCleaningMode == "Disabled" {
```
dhellmann, line 359: Maybe this logic can be collapsed so that we're only creating the `updates` instance in one place and the conditional only sets the `Value` argument?

fmuyassarov: That's a good point. Fixed it.

apis/metal3.io/v1alpha1/baremetalhost_types.go
fmuyassarov: Yes of course. Fixed.

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -278,8 +278,25 @@ type BareMetalHostSpec struct {
 	// the power status and hardware inventory inspection. If the
 	// Image field is filled in, this field is ignored.
 	ExternallyProvisioned bool `json:"externallyProvisioned,omitempty"`
+
+	// When set to Disabled, automated cleaning will be avoided
+	// during provisioning and deprovisioning.
+	// +optional
+	// +kubebuilder:default:=Full
+	// +kubebuilder:validation:Optional
+	AutomatedCleaningMode AutomatedCleaningMode `json:"automatedCleaningMode,omitempty"`
```
maelk, line 287: you could make this a pointer to simplify figuring out if it is omitted

pkg/provisioner/ironic/ironic.go
```diff
@@ -1212,6 +1245,21 @@ func (p *ironicProvisioner) Provision(hostConf provisioner.HostConfigData) (resu
 		return transientError(provisioner.NeedsRegistration)
 	}
 
+	if p.host.Spec.AutomatedCleaningMode == "None" {
```
maelk, line 1248: What is this "None" ? You should also consider the empty string (or nil pointer if you make it a pointer)

pkg/provisioner/ironic/ironic.go
```diff
@@ -1377,6 +1425,21 @@ func (p *ironicProvisioner) Deprovision(force bool) (result provisioner.Result,
...
```
maelk, line 1428: same here

pkg/provisioner/ironic/ironic.go
fmuyassarov: Oh sorry, this should be `Disabled`. Forgot to replace it.

fmuyassarov: Fixed it. Thanks.

apis/metal3.io/v1alpha1/baremetalhost_types.go
fmuyassarov: Added a check for nil pointer rather as you suggested. Thanks.

fmuyassarov:
@dhellmann @maelk I think I addressed all the comments now. Please take a look once more. Thank you.

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -278,8 +278,25 @@ type BareMetalHostSpec struct {
...
 }
 
+// AutomatedCleaningMode is the interface to enable/disable automated cleaning
+// +kubebuilder:validation:Enum:=Full;Disabled
+type AutomatedCleaningMode string
+
+// Allowed automated cleaning modes
+const (
+	Disabled AutomatedCleaningMode = "Disabled"
```
dhellmann, line 296: It would be good to namespace these constants a bit like we do with the state machine states. I know we didn't do that with the checksum types just below, but we probably should have.

```suggestion
	CleaningModeDisabled AutomatedCleaningMode = "Disabled"
```

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -278,8 +278,25 @@ type BareMetalHostSpec struct {
...
+	Full     AutomatedCleaningMode = "Full"
```
dhellmann, line 297:
```suggestion
	CleaningModeFull     AutomatedCleaningMode = "Full"
```

dhellmann (commented):
One nit about names for the constants, since they're public. Otherwise this looks good to me.

apis/metal3.io/v1alpha1/baremetalhost_types.go
fmuyassarov: Good point, fixed it.

fmuyassarov:
/test-integration 
Triggering CI even though it passed. I need to double check the functionality

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -278,8 +278,25 @@ type BareMetalHostSpec struct {
...
 }
 
+// AutomatedCleaningMode is the interface to enable/disable automated cleaning
+// +kubebuilder:validation:Enum:=Full;Disabled
+type AutomatedCleaningMode string
+
+// Allowed automated cleaning modes
+const (
+	CleaningModeDisabled AutomatedCleaningMode = "Disabled"
+	CleaningModeFull     AutomatedCleaningMode = "Full"
```
dtantsur, line 297: Maybe not call it "Full" in case we ever support actually removing all data, not just the partitioning table? We in ironic call it "metadata cleaning".

pkg/provisioner/ironic/ironic.go
```diff
@@ -1212,6 +1245,23 @@ func (p *ironicProvisioner) Provision(hostConf provisioner.HostConfigData) (resu
...
 
+	if p.host.Spec.AutomatedCleaningMode == metal3v1alpha1.CleaningModeDisabled {
+		p.log.Info("automated cleaning is disabled for host",
+			"ID", ironicNode.UUID,
+			"automated_clean", ironicNode.AutomatedClean,
+		)
+		p.publisher("CleaningDisabled", "Automated cleaning disabled")
+	}
+
+	if ironicNode.AutomatedClean != nil {
```
dtantsur, line 1256: It's `nil` by default. Not sure what you meant here, but it seems like it will never be updated?

pkg/provisioner/ironic/ironic.go
```diff
@@ -1212,6 +1245,23 @@ func (p *ironicProvisioner) Provision(hostConf provisioner.HostConfigData) (resu
...
+		if (p.host.Spec.AutomatedCleaningMode == metal3v1alpha1.CleaningModeDisabled && *ironicNode.AutomatedClean == true) || (p.host.Spec.AutomatedCleaningMode != metal3v1alpha1.CleaningModeDisabled && *ironicNode.AutomatedClean == false) {
```
dtantsur, line 1257: Maybe move this check inside `updateAutomatedClean` to avoid duplication of logic?

pkg/provisioner/ironic/ironic.go
```diff
@@ -1377,6 +1427,23 @@ func (p *ironicProvisioner) Deprovision(force bool) (result provisioner.Result,
...
 
+	if p.host.Spec.AutomatedCleaningMode == metal3v1alpha1.CleaningModeDisabled {
+		p.log.Info("automated cleaning is disabled for host",
+			"ID", ironicNode.UUID,
+			"automated_clean", ironicNode.AutomatedClean,
+		)
+		p.publisher("CleaningDisabled", "Automated cleaning disabled")
+	}
+
+	if ironicNode.AutomatedClean != nil {
```
dtantsur, line 1438: Same comments here

dtantsur (changes requested):

pkg/provisioner/ironic/ironic.go
fmuyassarov: @dtantsur This was added here to avoid NPE. Removing `if ironicNode.AutomatedClean != nil` will cause runtime panic. (example [snippet](https://paste.ubuntu.com/p/MKXGcgGDZK/)). Maybe `ironicNode.AutomatedClean` in gophercloud should be defaulted to `True` as you mentioned during the review on the PR in gophercloud, or I need to assign `True` here before my logic comes into play.

fmuyassarov: > It's `nil` by default. Not sure what you meant here, but it seems like it will never be updated?

It seems so.

fmuyassarov: I did some code refactoring to verify if that's the issue, and indeed we are not able to update it just because ironicNode.AutomatedClean is `nil`

fmuyassarov: As discussed on the [slack](https://kubernetes.slack.com/archives/CHD49TLE7/p1615195502083900), I changed the code a bit to enforce always updating the Ironic node.

apis/metal3.io/v1alpha1/baremetalhost_types.go
fmuyassarov: I can definitely rename it if desired. Any thoughts on this @dhellmann ?

fmuyassarov:
/hold cancel
@dtantsur @dhellmann @kashifest Please take look, it is ready from my side, I did a couple of test locally. The only open discussion that is unresolved is https://github.com/metal3-io/baremetal-operator/pull/784#discussion_r589422010, otherwise I think I addressed all the comments.

apis/metal3.io/v1alpha1/baremetalhost_types.go
dhellmann: Future-proofing the name would be good.

If we're only cleaning metadata, but on all devices, should we reflect that? `MetadataAllDevices` seems a bit long but at least it's descriptive. We could then have `MetadataSelectedDevices` at some point and variations for `AllDataAllDevices` and `AllDataSelectedDevices` as well.

@dtantsur Is the switch to control whether we delete metadata or all data exposed in the Ironic API, or just a config setting?

Are we ever likely to support different modes per device?

fmuyassarov: > Future-proofing the name would be good.
> 
> If we're only cleaning metadata, but on all devices, should we reflect that? `MetadataAllDevices` seems a bit long but at least it's descriptive. We could then have `MetadataSelectedDevices` at some point and variations for `AllDataAllDevices` and `AllDataSelectedDevices` as well.
> 
Maybe for now just `Metadata` ? Because, if we are going to extend it in the future, we will anyways have to touch the API. As such, maybe we could leave it to be done in future and for this proposal just stick to the `Metadata` ?

dhellmann: OK, I can go along with `Metadata` applying to all devices, for now. And in the future we can have `MetadataSelectedDevices` or something like that.

fmuyassarov: @dhellmann Updated the PR and the [proposal](https://github.com/metal3-io/metal3-docs/pull/166) as well.

pkg/provisioner/ironic/ironic.go
```diff
@@ -354,6 +354,44 @@ func (p *ironicProvisioner) findExistingHost() (ironicNode *nodes.Node, err erro
...
+	var updates nodes.UpdateOpts
+	var value bool
+
+	if p.host.Spec.AutomatedCleaningMode == metal3v1alpha1.CleaningModeDisabled {
+		p.log.Info("automated cleaning is disabled for host",
+			"ID", ironicNode.UUID,
+			"automated_clean", ironicNode.AutomatedClean,
+		)
+		p.publisher("CleaningDisabled", "Automated cleaning disabled")
+	} else if p.host.Spec.AutomatedCleaningMode != metal3v1alpha1.CleaningModeDisabled {
```
dhellmann, line 367: The condition here is redundant, isn't it?

Is there any reason not to always log what cleaning mode is being used? That would let us do something like the below and avoid the conditional entirely.

```go
p.log.Info("setting cleaning mode", 
    "ID", ironicNode.UUID, 
    "mode", p.host.Spec.AutomatedCleaningMode)
value = p.host.Spec.AutomatedCleaningMode != metal3v1alpha1.CleaningModeDisabled

updates = ...
```

pkg/provisioner/ironic/ironic.go
```diff
@@ -1347,6 +1385,8 @@ func (p *ironicProvisioner) Provision(hostConf provisioner.HostConfigData) (resu
 		return transientError(provisioner.NeedsRegistration)
 	}
 
+	_ = p.updateAutomatedClean(ironicNode)
```
dhellmann, line 1388: If the error was a 409, we need to pause before trying anything else because the host is busy. If it's not a 409, we probably don't want to ignore the error.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1512,6 +1552,8 @@ func (p *ironicProvisioner) Deprovision(force bool) (result provisioner.Result,
...
```
dhellmann, line 1555: We need to handle the error here, too.

fmuyassarov: Yes, fixed it.

pkg/provisioner/ironic/ironic.go
fmuyassarov: Yes it is! That's a good point, thanks.

pkg/provisioner/ironic/ironic.go
fmuyassarov: Fixed it. Thanks.

pkg/provisioner/ironic/provision_test.go
```diff
@@ -66,7 +66,7 @@ func TestProvision(t *testing.T) {
 		},
 		{
 			name: "active state",
-			ironic: testserver.NewIronic(t).Ready().Node(nodes.Node{
+			ironic: testserver.NewIronic(t).WithDefaultResponses().Node(nodes.Node{
 				ProvisionState: string(nodes.Active),
 				UUID:           nodeUUID,
 			}),
```
fmuyassarov, line 72: @dhellmann Please take a look at this change. Because I'm not sure if this is okay.  I had to make change in order to pass unit tests. Once I added error handling as you [suggested](https://github.com/metal3-io/baremetal-operator/pull/784#discussion_r592578762) , unit tests started failing as below
```
$ make unit

   server.go:249: ironic: Cannot find any default response for [PATCH] /v1/nodes/33ce8659-7400-4c68-9535-d10766f07a58
    server.go:72: ironic: PATCH /v1/nodes/33ce8659-7400-4c68-9535-d10766f07a58 -> 
    --- FAIL: TestProvision/active_state (0.00s)
        server.go:17: inspector: new server created
        server.go:134: inspector: adding response for [GET] /v1
        server.go:134: inspector: adding response for [GET] /v1/introspection/33ce8659-7400-4c68-9535-d10766f07a58
        server.go:67: inspector: endpoint: http://127.0.0.1:41231/v1/
        provision_test.go:121: 
            	Error Trace:	provision_test.go:121
            	Error:      	Received unexpected error:
            	            	EOF
            	            	failed to update ironic node automated_clean
            	            	github.com/metal3-io/baremetal-operator/pkg/provisioner/ironic.(*ironicProvisioner).updateAutomatedClean
            	            		/home/fmuyassarov/go/src/github/baremetal-operator/pkg/provisioner/ironic/ironic.go:386
            	            	github.com/metal3-io/baremetal-operator/pkg/provisioner/ironic.(*ironicProvisioner).Provision
            	            		/home/fmuyassarov/go/src/github/baremetal-operator/pkg/provisioner/ironic/ironic.go:1386
            	            	github.com/metal3-io/baremetal-operator/pkg/provisioner/ironic.TestProvision.func1
            	            		/home/fmuyassarov/go/src/github/baremetal-operator/pkg/provisioner/ironic/provision_test.go:111
            	            	testing.tRunner
            	            		/usr/local/go/src/testing/testing.go:1194
            	            	runtime.goexit
            	            		/usr/local/go/src/runtime/asm_amd64.s:1371
            	            	could not update ironic node automated_clean field
            	            	github.com/metal3-io/baremetal-operator/pkg/provisioner/ironic.(*ironicProvisioner).Provision
            	            		/home/fmuyassarov/go/src/github/baremetal-operator/pkg/provisioner/ironic/ironic.go:1387
            	            	github.com/metal3-io/baremetal-operator/pkg/provisioner/ironic.TestProvision.func1
            	            		/home/fmuyassarov/go/src/github/baremetal-operator/pkg/provisioner/ironic/provision_test.go:111
            	            	testing.tRunner
            	            		/usr/local/go/src/testing/testing.go:1194
            	            	runtime.goexit
            	            		/usr/local/go/src/runtime/asm_amd64.s:1371
            	Test:       	TestProvision/active_state
```

```
$ make unit | grep PATCH
...
    server.go:72: ironic: PATCH /v1/nodes/33ce8659-7400-4c68-9535-d10766f07a58 -> {"automated_clean":null,"uuid":"33ce8659-7400-4c68-9535-d10766f07a58","name":"","power_state":"","target_power_state":"","provision_state":"","target_provision_state":"","maintenance":false,"maintenance_reason":"","fault":"","last_error":"","reservation":"","driver":"","driver_info":null,"driver_internal_info":null,"properties":null,"instance_info":null,"instance_uuid":"","chassis_uuid":"","extra":null,"console_enabled":false,"raid_config":null,"target_raid_config":null,"clean_step":null,"deploy_step":null,"resource_class":"","boot_interface":"","console_interface":"","deploy_interface":"","inspect_interface":"","management_interface":"","network_interface":"","power_interface":"","raid_interface":"","rescue_interface":"","storage_interface":"","traits":null,"vendor_interface":"","conductor_group":"","protected":false,"protected_reason":"","owner":""}
    server.go:249: ironic: Cannot find any default response for [PATCH] /v1/nodes/33ce8659-7400-4c68-9535-d10766f07a58
    server.go:72: ironic: PATCH /v1/nodes/33ce8659-7400-4c68-9535-d10766f07a58 -> 
    server.go:249: ironic: Cannot find any default response for [PATCH] /v1/nodes/33ce8659-7400-4c68-9535-d10766f07a58
    server.go:72: ironic: PATCH /v1/nodes/33ce8659-7400-4c68-9535-d10766f07a58 -> 
...
```

Replacing `testserver.NewIronic(t).Ready().Node` -> `testserver.NewIronic(t).WithDefaultResponses().Node` fixed the units tests. But I'm not 100% clear about those unit tests.

fmuyassarov: I don't understand how my patch is breaking the existing unit tests.

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -344,8 +344,25 @@ type BareMetalHostSpec struct {
...
 }
 
+// AutomatedCleaningMode is the interface to enable/disable automated cleaning
+// +kubebuilder:validation:Enum:=Metadata;Disabled
+type AutomatedCleaningMode string
+
+// Allowed automated cleaning modes
+const (
+	CleaningModeDisabled AutomatedCleaningMode = "Disabled"
+	CleaningModeMetadata AutomatedCleaningMode = "Metadata"
```
dtantsur, line 363: Thanks! There is no API knob to tune which cleaning runs, but there may be in the future.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1347,6 +1383,10 @@ func (p *ironicProvisioner) Provision(hostConf provisioner.HostConfigData) (resu
 		return transientError(provisioner.NeedsRegistration)
 	}
 
+	if err = p.updateAutomatedClean(ironicNode); err != nil {
+		return transientError(errors.Wrap(err, "could not update ironic node automated_clean field"))
```
dtantsur, line 1387: nit: I'd prefer this to be phrased in terms of metal3 entities, not ironic fields

pkg/provisioner/ironic/provision_test.go
dtantsur: My guess: previously we would call PATCH only once for all operations. Now you've added another call, but there is no fixture for it.

hardys: Do we actually have to do the PATCH in the Provision/Deprovision functions, or could we just add the update e.g to `getImageUpdateOptsForNode` ?

That would simplify the code changes, and reduce the test impact but perhaps I'm missing some corner cases which won't get handled in that case?

dtantsur: When is it called? If it's only called before the deployment, it's too late to disable cleaning (it happens on the way from `manageable` to `available`).

pkg/provisioner/ironic/ironic.go
fmuyassarov: Sure, will update it.

pkg/provisioner/ironic/provision_test.go
fmuyassarov: I'm not sure about `getImageUpdateOptsForNode`,  and the reason I used it in the Provision() & Deprovision() is because cleaning also can be enabled/disabled for provisioning & deprovisioning. I could not find any better place to call Ironic to update the field.

andfasano: @fmuyassarov in case of problems the mock's output has been designed to help you arrange the required calls in the test setup: the most frequent reason it's a change in the code that stimulates one (or more) API calls.
The usage of `WithDefaultResponses` could be handy when, in the test setup, you're not interested in the "usual" API calls, and so its goal it's to add a valid (mostly generic) answer for such calls. On the other side, it could hide some important interactions that instead could be worth to be mocked. If you remove `WithDefaultResponses` and run the test again, the output should provide enough info to detect which call requires to be mocked (ie by looking at the messages in the format `... Cannot find any default response for ...`).

fmuyassarov: Thank you @andfasano for the explanation! That's what I got now...

```
       server.go:72: ironic: GET /v1/nodes/33ce8659-7400-4c68-9535-d10766f07a58 -> {"automated_clean":null,"uuid":"33ce8659-7400-4c68-9535-d10766f07a58","name":"","power_state":"","target_power_state":"","provision_state":"active","target_provision_state":"","maintenance":false,"maintenance_reason":"","fault":"","last_error":"","reservation":"","driver":"","driver_info":null,"driver_internal_info":null,"properties":null,"instance_info":null,"instance_uuid":"","chassis_uuid":"","extra":null,"console_enabled":false,"raid_config":null,"target_raid_config":null,"clean_step":null,"deploy_step":null,"resource_class":"","boot_interface":"","console_interface":"","deploy_interface":"","inspect_interface":"","management_interface":"","network_interface":"","power_interface":"","raid_interface":"","rescue_interface":"","storage_interface":"","traits":null,"vendor_interface":"","conductor_group":"","protected":false,"protected_reason":"","owner":""}
    server.go:236: ironic: found default response for /v1/nodes/33ce8659-7400-4c68-9535-d10766f07a58: {200, {"automated_clean":null,"uuid":"{id}","name":"","power_state":"","target_power_state":"","provision_state":"","target_provision_state":"","maintenance":false,"maintenance_reason":"","fault":"","last_error":"","reservation":"","driver":"","driver_info":null,"driver_internal_info":null,"properties":null,"instance_info":null,"instance_uuid":"","chassis_uuid":"","extra":null,"console_enabled":false,"raid_config":null,"target_raid_config":null,"clean_step":null,"deploy_step":null,"resource_class":"","boot_interface":"","console_interface":"","deploy_interface":"","inspect_interface":"","management_interface":"","network_interface":"","power_interface":"","raid_interface":"","rescue_interface":"","storage_interface":"","traits":null,"vendor_interface":"","conductor_group":"","protected":false,"protected_reason":"","owner":""}}
    server.go:72: ironic: PATCH /v1/nodes/33ce8659-7400-4c68-9535-d10766f07a58 -> {"automated_clean":null,"uuid":"33ce8659-7400-4c68-9535-d10766f07a58","name":"","power_state":"","target_power_state":"","provision_state":"","target_provision_state":"","maintenance":false,"maintenance_reason":"","fault":"","last_error":"","reservation":"","driver":"","driver_info":null,"driver_internal_info":null,"properties":null,"instance_info":null,"instance_uuid":"","chassis_uuid":"","extra":null,"console_enabled":false,"raid_config":null,"target_raid_config":null,"clean_step":null,"deploy_step":null,"resource_class":"","boot_interface":"","console_interface":"","deploy_interface":"","inspect_interface":"","management_interface":"","network_interface":"","power_interface":"","raid_interface":"","rescue_interface":"","storage_interface":"","traits":null,"vendor_interface":"","conductor_group":"","protected":false,"protected_reason":"","owner":""}
    server.go:67: ironic: endpoint: http://127.0.0.1:39187/v1/
    server.go:72: ironic: GET /v1/nodes/33ce8659-7400-4c68-9535-d10766f07a58 -> {"automated_clean":null,"uuid":"33ce8659-7400-4c68-9535-d10766f07a58","name":"","power_state":"","target_power_state":"","provision_state":"cleaning","target_provision_state":"","maintenance":false,"maintenance_reason":"","fault":"","last_error":"","reservation":"","driver":"","driver_info":null,"driver_internal_info":null,"properties":null,"instance_info":null,"instance_uuid":"","chassis_uuid":"","extra":null,"console_enabled":false,"raid_config":null,"target_raid_config":null,"clean_step":null,"deploy_step":null,"resource_class":"","boot_interface":"","console_interface":"","deploy_interface":"","inspect_interface":"","management_interface":"","network_interface":"","power_interface":"","raid_interface":"","rescue_interface":"","storage_interface":"","traits":null,"vendor_interface":"","conductor_group":"","protected":false,"protected_reason":"","owner":""}
    server.go:249: ironic: Cannot find any default response for [PATCH] /v1/nodes/33ce8659-7400-4c68-9535-d10766f07a58
    server.go:72: ironic: PATCH /v1/nodes/33ce8659-7400-4c68-9535-d10766f07a58 -> 
    --- FAIL: TestProvision/other_state:_Cleaning (0.00s)
        server.go:17: inspector: new server created
        server.go:134: inspector: adding response for [GET] /v1
        server.go:134: inspector: adding response for [GET] /v1/introspection/33ce8659-7400-4c68-9535-d10766f07a58
        server.go:67: inspector: endpoint: http://127.0.0.1:37975/v1/
        provision_test.go:113: 
            	Error Trace:	provision_test.go:113
            	Error:      	Not equal: 
            	            	expected: true
            	            	actual  : false
            	Test:       	TestProvision/other_state:_Cleaning
        provision_test.go:114: 
            	Error Trace:	provision_test.go:114
            	Error:      	Not equal: 
            	            	expected: 10s
            	            	actual  : 0s
            	Test:       	TestProvision/other_state:_Cleaning
        provision_test.go:121: 
            	Error Trace:	provision_test.go:121
            	Error:      	Received unexpected error:
            	            	EOF
            	            	failed to update ironic node automated_clean
            	            	github.com/metal3-io/baremetal-operator/pkg/provisioner/ironic.(*ironicProvisioner).updateAutomatedClean
            	            		/home/fmuyassarov/go/src/github/baremetal-operator/pkg/provisioner/ironic/ironic.go:386
            	            	github.com/metal3-io/baremetal-operator/pkg/provisioner/ironic.(*ironicProvisioner).Provision
            	            		/home/fmuyassarov/go/src/github/baremetal-operator/pkg/provisioner/ironic/ironic.go:1386
            	            	github.com/metal3-io/baremetal-operator/pkg/provisioner/ironic.TestProvision.func1
            	            		/home/fmuyassarov/go/src/github/baremetal-operator/pkg/provisioner/ironic/provision_test.go:111
            	            	testing.tRunner
            	            		/usr/local/go/src/testing/testing.go:1194
            	            	runtime.goexit
            	            		/usr/local/go/src/runtime/asm_amd64.s:1371
            	            	could not update ironic node automated_clean field
            	            	github.com/metal3-io/baremetal-operator/pkg/provisioner/ironic.(*ironicProvisioner).Provision
            	            		/home/fmuyassarov/go/src/github/baremetal-operator/pkg/provisioner/ironic/ironic.go:1387
            	            	github.com/metal3-io/baremetal-operator/pkg/provisioner/ironic.TestProvision.func1
            	            		/home/fmuyassarov/go/src/github/baremetal-operator/pkg/provisioner/ironic/provision_test.go:111
            	            	testing.tRunner
            	            		/usr/local/go/src/testing/testing.go:1194
            	            	runtime.goexit
            	            		/usr/local/go/src/runtime/asm_amd64.s:1371
            	Test:       	TestProvision/other_state:_Cleaning
```

andfasano: To fix it, as also @dtantsur suggested, you need to instruct the Ironic mock to give a valid answer for a node PATCH using `NodeUpdate`, ie:

```
              {
			name: "active state",
			ironic: testserver.NewIronic(t).Ready().Node(nodes.Node{
				ProvisionState: string(nodes.Active),
				UUID:           nodeUUID,
			}).NodeUpdate(nodes.Node{
				UUID: nodeUUID,
			}),
			expectedRequestAfter: 0,
			expectedDirty:        false,
		},
```

fmuyassarov: Great, that indeed fixed my issue! Thank you @andfasano ! Please take a look.

fmuyassarov:
Thank you all for your reviews. I think I have addressed all the comments I had so far. Please take a look.

pkg/provisioner/ironic/ironic.go
```diff
@@ -354,6 +354,42 @@ func (p *ironicProvisioner) findExistingHost() (ironicNode *nodes.Node, err erro
...
+
+	updates = nodes.UpdateOpts{
+		nodes.UpdateOperation{
+			Op:    nodes.ReplaceOp,
+			Path:  "/automated_clean",
+			Value: value,
+		},
+	}
+
+	_, err = nodes.Update(p.client, ironicNode.UUID, updates).Extract()
```
andfasano, line 377: Wouldn't be better to request an update only _if_ required? Otherwise the call will be raised everytime (and the current changes on the unit tests are showing that)

andfasano (changes requested):

pkg/provisioner/ironic/ironic.go
dtantsur: Knowing whether it's required can be tricky. The default value is `nil`, which means "use the default from ironic.conf". So you'll have to assume what's in ironic.conf.

fmuyassarov: > Wouldn't be better to request an update only _if_ required? Otherwise the call will be raised everytime (and the current changes on the unit tests are showing that)

It would be, but `ironicNode.AutomatedClean` has nil by default as @dtantsur pointed out. As such, for example adding this:
```
if p.host.Spec.AutomatedCleaningMode == metal3v1alpha1.CleaningModeDisabled && *ironicNode.AutomatedClean != false || p.host.Spec.AutomatedCleaningMode == metal3v1alpha1.CleaningModeMetadata && *ironicNode.AutomatedClean == false {
    // do the update here
}
```
will cause runtime panic issue as we are using `*ironicNode.AutomatedClean`.
```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1484c44]
```
 We had the [slack chat](https://kubernetes.slack.com/archives/CHD49TLE7/p1615195502083900) about this with @dtantsur, and thought the way forward would be to always update it.

fmuyassarov:
@andfasano @dhellmann @dtantsur @hardys 
I think all the comments have been addressed by now. Also the [update](https://github.com/metal3-io/metal3-docs/pull/166) in the original proposal got merged. Can we merge this one too if looks okay? Because I will still need to update it in [CAPM3 PR ](https://github.com/metal3-io/cluster-api-provider-metal3/pull/156)after this is merged.

pkg/provisioner/ironic/ironic.go
andfasano: Of course dereferencing a pointer without a nil check it's not feasible. And I don't think it's a blocker, considering that the call it's performed just in the provisioning/deprovisioning, but it would be nice to have a way to use that flag.

>The default value is nil, which means "use the default from ironic.conf". So you'll have to assume what's in ironic.conf.
Any way to retrieve the global conf in that case?

hardys:
> @andfasano @dhellmann @dtantsur @hardys
> I think all the comments have been addressed by now. Also the [update](https://github.com/metal3-io/metal3-docs/pull/166) in the original proposal got merged. Can we merge this one too if looks okay? Because I will still need to update it in [CAPM3 PR ](https://github.com/metal3-io/cluster-api-provider-metal3/pull/156)after this is merged.

This version lgtm - I'll trigger the integration tests and do some local testing of the Disabled case before approving (and hopefully we can get some other reviews in the meantime).

pkg/provisioner/ironic/ironic.go
dtantsur: There is no way to retrieve the global conf. What we could do is to avoid updates after the flag has been updated once (and its value is not `nil`).

pkg/provisioner/ironic/ironic.go
```diff
@@ -1347,6 +1383,10 @@ func (p *ironicProvisioner) Provision(hostConf provisioner.HostConfigData) (resu
 		return transientError(provisioner.NeedsRegistration)
 	}
 
+	if err = p.updateAutomatedClean(ironicNode); err != nil {
```
zaneb, line 1386: It seems like a mistake to call this in Provision() (or even Deprovision()), since there are numerous opportunities for the node to get cleaned at other times - notably right after registration, but also at any other time when the ironic database is dropped and we don't call Adopt().
I think we should be setting the cleaning mode in ValidateManagementAccess(), which gets called on every reconcile. That way the node is always in a state with the right setting. Perhaps we could consolidate this along with the 3 calls we are already making to nodes.Update() in that method into a single one.

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -344,8 +344,25 @@ type BareMetalHostSpec struct {
...
+	AutomatedCleaningMode AutomatedCleaningMode `json:"automatedCleaningMode,omitempty"`
 }
 
+// AutomatedCleaningMode is the interface to enable/disable automated cleaning
+// +kubebuilder:validation:Enum:=Metadata;Disabled
+type AutomatedCleaningMode string
+
+// Allowed automated cleaning modes
+const (
+	CleaningModeDisabled AutomatedCleaningMode = "Disabled"
```
zaneb, line 362: Other constants in the API are lower-case, should we do that here too?

docs/api.md
```diff
@@ -172,6 +172,12 @@ The sub-fields are
 * *rotational* -- A boolean indicating whether the device should be
   a rotating disk (`true`) or not (`false`).
 
+#### automatedCleaningMode
+
+An interface to enable/disable automated cleaning during provisioning
+and deprovisioning. When set to `Disabled`, automated cleaning will be
+skipped, where `Metadata`(default value) enables it.
```
zaneb, line 179: How do we expect 'Metadata' to differ from other possible cleaning modes?

pkg/provisioner/ironic/ironic.go
```diff
@@ -354,6 +354,42 @@ func (p *ironicProvisioner) findExistingHost() (ironicNode *nodes.Node, err erro
 
 }
 
+func (p *ironicProvisioner) updateAutomatedClean(ironicNode *nodes.Node) (err error) {
+	var updates nodes.UpdateOpts
+	var value bool
```
zaneb, line 359: nit: these declarations aren't needed, you can just use `:=`

pkg/provisioner/ironic/ironic.go
fmuyassarov: I remember a similar [comment](https://github.com/metal3-io/metal3-docs/pull/151#discussion_r527709842) from @dhellmann on the proposal that we could to do it in `ValidateManagementAccess()`, but I did it within `Provision()` and  `Deprovision()` because I thought these are the only methods from after which Ironic node gets cleaned. From Ironic [state machine](https://docs.openstack.org/ironic/pike/_images/states.svg). I see that Ironic cleaning state will come to cleaning state either after managable (BMH Ready state) or active -> deleting (BMH deprovisioning state).
@zaneb Would you mind to share an example use case where Ironic node gets cleaned without us calling `Provision()` or  `Deprovision()` ?

apis/metal3.io/v1alpha1/baremetalhost_types.go
fmuyassarov: Yes, and this means allowed values will also need to be changed in the baremetalhost types. But your point makes sense. Will update it.

docs/api.md
fmuyassarov: This was decided to be called Metadata to reflect Ironic Metadata cleaning after the discussion [here](https://github.com/metal3-io/baremetal-operator/pull/784#discussion_r589422010). So, in the future there might be on option to select only certain devices to be cleaned.

pkg/provisioner/ironic/ironic.go
zaneb: You're correct, I misremembered the change in 18465130c8f42bda4aa8191b5093656dcc029d90.
I do plan to change the state machine so that the Preparing state will move the state from Manageable->Available (and thus trigger automatic cleaning) as [proposed on the mailing list](https://groups.google.com/g/metal3-dev/c/N-EpqQn94bw/m/nRkM9bmuAAAJ), but currently it does not do so.

fmuyassarov:
@hardys previously integration CI passed, but I had some comments to be fixed from @zaneb, and now it is ready again. I can trigger it now if we are good to go with the current state. Please let me know If I can help in testing the feature.

pkg/provisioner/ironic/ironic.go
hardys: I made a similar [comment](https://github.com/metal3-io/baremetal-operator/pull/784#discussion_r594475109) but the current approach works, so perhaps we can address the consolidation of Updates in a follow-up PR?

hardys:
This lgtm - I tested it locally and it works as expected - I see @zaneb made a comment similar to our previous discussion about when to handle the `nodes.Update` - we probably can improve/consolidate that but perhaps that can be done as a followup or as part of the future cleaning rework mentioned by @zaneb ?

/lgtm

fmuyassarov:
> This lgtm - I tested it locally and it works as expected - I see @zaneb made a comment similar to our previous discussion about when to handle the `nodes.Update` - we probably can improve/consolidate that but perhaps that can be done as a followup or as part of the future cleaning rework mentioned by @zaneb ?
> 
> /lgtm

Thanks @hardys ! If possible, it would be nice if we do it in a follow up PR, and I can sign up to help on that.

fmuyassarov:
@dhellmann would you mind to take a look once more please?

pkg/provisioner/ironic/ironic.go
```diff
@@ -354,6 +354,39 @@ func (p *ironicProvisioner) findExistingHost() (ironicNode *nodes.Node, err erro
...
+		},
+	}
+
+	_, err = nodes.Update(p.client, ironicNode.UUID, updates).Extract()
+
+	switch err.(type) {
+	case nil:
+	case gophercloud.ErrDefault409:
+		p.log.Info("could not update ironic node automated_clean, busy")
+		_, err = retryAfterDelay(provisionRequeueDelay)
```
zaneb, line 380: This line just returns nil and does not retry. This is wrong and will cause the calling code to believe that cleaning has been disabled when it hasn't.

pkg/provisioner/ironic/ironic.go
```diff
@@ -354,6 +354,39 @@ func (p *ironicProvisioner) findExistingHost() (ironicNode *nodes.Node, err erro
...
+		return
+	default:
+		_, err = transientError(errors.Wrap(err, "failed to update ironic node automated_clean"))
```
zaneb, line 383: Using the transientError call here is pointless in a function that only returns an `error`.

zaneb (commented):
OK, if someone is going to follow up with moving it to the registration then I'm OK with doing it later.

That said, there's a bug in the error handling here and it might be quicker to move the code than to fix it :shrug:

dtantsur:
/lgtm cancel

@fmuyassarov could you update per Zane's comments?

dtantsur:
/approve

Approving since the direction and the overall code is correct, just some details need polishing.

pkg/provisioner/ironic/ironic.go
fmuyassarov: @zaneb you are right. Do you think I should not retry at all, and replace L380 with for example something like
`return errors.Wrap(err, "could not update ironic node automated_clean, busy")` or shall I instead add `result provisioner.Result` as second return value for `updateAutomatedClean()` and do retry? Maybe the latter option makes more sense, because I can see that we are retrying in most of the places in ironic.go.

pkg/provisioner/ironic/ironic.go
fmuyassarov: True, will fix it once we decide on the questions above.

pkg/provisioner/ironic/ironic.go
```diff
@@ -354,6 +354,39 @@ func (p *ironicProvisioner) findExistingHost() (ironicNode *nodes.Node, err erro
 
 }
 
+func (p *ironicProvisioner) updateAutomatedClean(ironicNode *nodes.Node) (err error) {
+	p.log.Info("setting cleaning mode",
+		"ID", ironicNode.UUID,
+		"mode", p.host.Spec.AutomatedCleaningMode)
+
+	p.publisher("CleaningMode", string(p.host.Spec.AutomatedCleaningMode))
```
flaper87, line 362: Not sure if there's a problem in my env but this resulted being quite verbose for me:

DeepinScreenshot_select-area_20210324082825

fmuyassarov: Seems that's due the publisher is being called on all the states if Ironic node. 
@hardys do you think it is okay that I add the publisher (for deprovisioning) when Ironic node is either in [`Active or DeployFail`](https://github.com/metal3-io/baremetal-operator/blob/5ebf0ab6e367a854c4177edc6612381b1d65dc01/pkg/provisioner/ironic/ironic.go#L1600)  or maybe when it is [`deleting`](https://github.com/metal3-io/baremetal-operator/blob/5ebf0ab6e367a854c4177edc6612381b1d65dc01/pkg/provisioner/ironic/ironic.go#L1584) ?

And during provisioning, I was thinking maybe to add publisher when Ironic node is in [Active](https://github.com/metal3-io/baremetal-operator/blob/5ebf0ab6e367a854c4177edc6612381b1d65dc01/pkg/provisioner/ironic/ironic.go#L1477)?

hardys: @fmuyassarov do we need an event for this at all?

I'm thinking we don't output events for many other types of configuration, so we could just move this update to `ValidateManagementAccess()` as @zaneb mentioned, and apend it to the [other updates](https://github.com/metal3-io/baremetal-operator/blob/master/pkg/provisioner/ironic/ironic.go#L472..L479)?  Maybe log something but drop the event output?

pkg/provisioner/ironic/ironic.go
zaneb: Returning an `error` will result in an immediate* requeue plus an ugly error in the logs. Generally speaking, responding to ironic getting hammered by hammering it harder is not ideal, which is why we use retryAfterDelay everywhere else we get a 409.

\* subject to throttling in the controller

pkg/provisioner/ironic/ironic.go
fmuyassarov: I added events here as per https://github.com/metal3-io/baremetal-operator/pull/784#discussion_r585064586. I can replace events with simple logging if that's okay.

fmuyassarov: I have removed events, and instead kept only logging that was on top of that. I hope that's ok ?
```
	p.log.Info("setting automated cleaning mode to",
		"ID", ironicNode.UUID,
		"mode", p.host.Spec.AutomatedCleaningMode)
```

pkg/provisioner/ironic/ironic.go
fmuyassarov: Fixed accordingly, thank you!

pkg/provisioner/ironic/ironic.go
fmuyassarov: @hardys  Isn't `ValidateManagementAccess` is used only when we are registering the host for the first time? If yes, that will work when someone enables/disables cleaning before provisioning. But what if we want to enable/disable cleaning before deprovisioning?

fmuyassarov: @zaneb any thoughts on above question?

zaneb: `ValidateManagementAccess()` is called in ~every state since #716.

hardys: > I added events here as per [#784 (comment)](https://github.com/metal3-io/baremetal-operator/pull/784#discussion_r585064586). I can replace events with simple logging if that's okay.

Logging wfm - my previous comment was just to say that I did generate an event in the inspection-disabled case - but if we generate an event (or log output) we want to do that once, not repeatedly

pkg/provisioner/ironic/ironic.go
```diff
@@ -546,6 +546,36 @@ func (p *ironicProvisioner) ValidateManagementAccess(credentialsChanged, force b
...
 	// }
 
+	p.log.Info("setting automated cleaning mode to",
+		"ID", ironicNode.UUID,
+		"mode", p.host.Spec.AutomatedCleaningMode)
+
+	if (p.host.Spec.AutomatedCleaningMode == "disabled" && *ironicNode.AutomatedClean != false) || (p.host.Spec.AutomatedCleaningMode == "enabled" && *ironicNode.AutomatedClean == false) {
+		value := p.host.Spec.AutomatedCleaningMode != metal3v1alpha1.CleaningModeDisabled
+
+		updates := nodes.UpdateOpts{
```
hardys, line 556: Does this need to be an additional update?

I was thinking this could be called immediately after `getImageUpdateOptsForNode` (e.g a new `getCleaningModeOpts ` function perhaps?) and appended to the `updates`?

@zaneb am I right in saying we should hit that code in both the Provisioning and Deprovisioning state, since imageData will be set via either the Spec or Status?

pkg/provisioner/ironic/ironic.go
```diff
@@ -546,6 +546,36 @@ func (p *ironicProvisioner) ValidateManagementAccess(credentialsChanged, force b
 	// 	return result, errors.Wrap(err, "failed to get provisioning state in ironic")
 	// }
 
+	p.log.Info("setting automated cleaning mode to",
+		"ID", ironicNode.UUID,
+		"mode", p.host.Spec.AutomatedCleaningMode)
```
hardys, line 551: We should only log this when actually making a change

pkg/provisioner/ironic/ironic.go
```diff
@@ -391,7 +391,7 @@ func (p *ironicProvisioner) ValidateManagementAccess(credentialsChanged, force b
 	driverInfo["deploy_kernel"] = deployKernelURL
 	driverInfo["deploy_ramdisk"] = deployRamdiskURL
 
-	result, err = operationComplete()
+	result, _ = operationComplete()
```
zaneb, line 394: This change doesn't do anything and no good can come of it.

pkg/provisioner/ironic/ironic.go
zaneb: I [agree](https://github.com/metal3-io/baremetal-operator/pull/784#discussion_r597888586), we should stop copying and pasting the code to Update() and all of its error handling, which we already call in 3 places (now 4) when 1 would suffice, and just add one more thing to the list.

Note, however, that the `getImageUpdateOptsForNode()` is only called if the node doesn't exist. Which is a bug, because if updating it fails then it will never get called again. You are correct that the image data is available in both the Provisioning and Deprovisioning states.

We could refactor and fix the bug in a separate PR if that helps.

fmuyassarov: Do we need a new func for this? or can we just add it within  `getImageUpdateOptsForNode()?`

zaneb: I would just do the Update() once after the if/else, and have the stuff that is currently calling Update() just add stuff to the updates list.
We should definitely not call Update() within `getImageUpdateOptsForNode()` with the same list of updates that we're about to return, only for Update() to be called with it again.

fmuyassarov: @zaneb can you please take a look if the current state is what you meant?

zaneb: I wouldn't put it inside `getImageUpdateOptsForNode()` because:

1. of the bug mentioned above, which means it is not called most of the time when we want it to be called (you'd have to delete the ironic DB to get the cleaning mode to update)
2. even when the bug is fixed, it'll only be called when there is an image specified, but that's not a condition we want to impose on whether the cleaning mode gets updated

fmuyassarov:
@zaneb @hardys do you think it is better now?

pkg/provisioner/ironic/ironic.go
```diff
@@ -508,61 +500,61 @@ func (p *ironicProvisioner) ValidateManagementAccess(credentialsChanged, force b
...
+		"mode", p.host.Spec.AutomatedCleaningMode)
+
+	value := p.host.Spec.AutomatedCleaningMode != metal3v1alpha1.CleaningModeDisabled
+	updates = append(
+		updates,
+		nodes.UpdateOperation{
+			Op:    nodes.ReplaceOp,
+			Path:  "/automated_clean",
+			Value: value,
+		},
```
hardys, line 540: This UpdateOperation is still applied unconditionally, should we be checking `ironicNode.AutomatedClean` and only setting it while needed?

zaneb: Yeah, you also need to check for a non-nil pointer before dereferencing it.

fmuyassarov: > Yeah, you also need to check for a non-nil pointer before dereferencing it.

We will never be able to do the update if we check that, because `automated_clean` = `nil` by default in gophercloud.

zaneb: If it's nil you always update:

```
if ironicNode.AutomatedClean == nil || (p.host.Spec.AutomatedCleaningMode == "disabled" && *ironicNode.AutomatedClean != false) || (p.host.Spec.AutomatedCleaningMode == "enabled" && *ironicNode.AutomatedClean == false) {
```

fmuyassarov: 🤦‍♂️, thank you!

pkg/provisioner/ironic/ironic.go
```diff
@@ -508,61 +500,61 @@ func (p *ironicProvisioner) ValidateManagementAccess(credentialsChanged, force b
...
+	)
+	ironicNode, err = nodes.Update(p.client, ironicNode.UUID, updates).Extract()
```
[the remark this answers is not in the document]
fmuyassarov: I think yes, with the current state.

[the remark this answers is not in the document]
hardys: Ok if we think that can happen we probably need to check len(updates) and only do the nodes.Update if there are updates pending?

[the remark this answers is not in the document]
hardys: nvm, I see the latest version does that :)

hardys:
This version lgtm other than resolving the failing tests, thanks @fmuyassarov!

fmuyassarov:
Thanks everyone for reviews. I think all the comments now addressed!.

PR #786, opened 2021-02-02
Automatic setting of UEFI secure boot

dtantsur:
Design: https://github.com/metal3-io/metal3-docs/pull/161

This depends on https://github.com/metal3-io/baremetal-operator/pull/783

### DIFF
--- a/apis/metal3.io/v1alpha1/baremetalhost_types.go
+++ b/apis/metal3.io/v1alpha1/baremetalhost_types.go
@@ -90,12 +90,13 @@ type RootDeviceHints struct {
 }
 
 // BootMode is the boot mode of the system
-// +kubebuilder:validation:Enum=UEFI;legacy
+// +kubebuilder:validation:Enum=UEFI;UEFISecureBoot;legacy
 type BootMode string
 
 // Allowed boot mode from metal3
 const (
 	UEFI            BootMode = "UEFI"
+	UEFISecureBoot  BootMode = "UEFISecureBoot"
 	Legacy          BootMode = "legacy"
 	DefaultBootMode BootMode = UEFI
 )
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -58,8 +58,9 @@ const (
 )
 
 var bootModeCapabilities = map[metal3v1alpha1.BootMode]string{
-	metal3v1alpha1.UEFI:   "boot_mode:uefi",
-	metal3v1alpha1.Legacy: "boot_mode:bios",
+	metal3v1alpha1.UEFI:           "boot_mode:uefi",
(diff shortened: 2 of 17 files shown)


apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -90,12 +90,13 @@ type RootDeviceHints struct {
...
 
 // BootMode is the boot mode of the system
-// +kubebuilder:validation:Enum=UEFI;legacy
+// +kubebuilder:validation:Enum=UEFI;UEFI secure boot;legacy
 type BootMode string
 
 // Allowed boot mode from metal3
 const (
 	UEFI            BootMode = "UEFI"
+	SecureUEFI      BootMode = "UEFI secure boot"
```
hardys, line 99: nit, but I'm not sure if spaces are typically used in enumerated values like this - I wonder if it should be something like `SecureBootUEFI` or just `SecureBoot` if we know that only works with UEFI?

@dhellmann any thoughts on that?

dhellmann: Let's avoid spaces in the enum. I don't think they're forbidden, but as a style thing I haven't seen them enough to say they're "liked" generally.

In terms of the actual name to use, I don't have an opinion.

dtantsur: `UEFISecureBoot` resembles the canonical name of the feature the most, I'll use it.

dtantsur: The only problem (that led me to this) is that `legacy` is low-cased.

dhellmann: What do kubernetes (or OpenShift?) API guidelines say about enum values? I think they want capital letters, so we got `legacy` wrong.

pkg/provisioner/ironic/ironic.go
```diff
@@ -396,6 +397,13 @@ func (p *ironicProvisioner) ValidateManagementAccess(credentialsChanged, force b
 	if ironicNode == nil {
 		p.log.Info("registering host in ironic")
 
+		if p.host.Spec.BootMode == metal3v1alpha1.UEFISecureBoot && !p.bmcAccess.SupportsSecureBoot() {
+			msg := fmt.Sprintf("BMC driver %s does not support secure boot", p.bmcAccess.Type())
```
hardys, line 401: It'd be nice to have a test for this error path - similar to the existing ones in https://github.com/metal3-io/baremetal-operator/blob/master/pkg/provisioner/ironic/validatemanagementaccess_test.go

bfournie (approved):

pkg/provisioner/ironic/ironic.go
dtantsur: https://github.com/metal3-io/baremetal-operator/pull/831

PR #788, opened 2021-02-05
Fix the CPU clock rate unit in the API description

tomassedovic:
The hardware subfield for the CPU clock rate is called
`clockMegahertz`, but the description uses `GHz`. This changes the
description to use `MHz`.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -234,7 +234,7 @@ The sub-fields are
 * *cpu* -- Details of the CPU(s) in the system.
   * *arch* -- The architecture of the CPU.
   * *model* -- The model string.
-  * *clockMegahertz* -- The speed in GHz of the CPU.
+  * *clockMegahertz* -- The speed in MHz of the CPU.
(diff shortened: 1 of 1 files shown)


tomassedovic:
This assumes that the right unit is indeed MHz, but looking at the sample provided later in the docs, that seems to be the case.

hardys:
I checked and the right unit does appear to be MHz:

```
$ cat /proc/cpuinfo  | grep "^model name" | head -n1
model name	: Intel(R) Xeon(R) Gold 6230 CPU @ 2.10GHz
# Then in the BMH status we see:
      clockMegahertz: 2095
```

/ok-to-test
/approve

iurygregory (approved):

dtantsur (approved):

tomassedovic:
Looks like the last thing necessary to merge this is someone commenting `/lgtm`.

dhellmann:
/lgtm

Thanks!

PR #791, opened 2021-02-11
Change the BMH's scope from Namespaced to Cluster scope

sadasu:


dhellmann:
This is a pretty significant change to the API. I think we'll want a design doc explaining why we need to make it over in metal3-docs before we make the code change.

s3rj1k:
@dhellmann Can this be configurable some how?

s3rj1k:
How about creating BMHClustered? And supporting both cases.
Also it would be hard migrating current namespaced BMHs to cluster scope in production.

config/crd/bases/metal3.io_baremetalhosts.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.4.1
+    controller-gen.kubebuilder.io/version: v0.4.0
```
dtantsur, line 7: Why a version downgrade?

kirankt: My guess is that the local controller-gen version is different. Could be remedied easily.

sadasu:
After further investigation if appears that this change is not needed and the wrong direction for what I was looking for. Thanks for taking a look and providing comments.

PR #793, opened 2021-02-12
Add more logging in ironic provisioner.

s3rj1k:

### DIFF
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -470,6 +470,7 @@ func (p *ironicProvisioner) ValidateManagementAccess(credentialsChanged, force b
 				return
 			}
 			if len(updates) != 0 {
+				p.log.Info("updating host settings in ironic")
 				_, err = nodes.Update(p.client, ironicNode.UUID, updates).Extract()
 				switch err.(type) {
 				case nil:
@@ -481,6 +482,7 @@ func (p *ironicProvisioner) ValidateManagementAccess(credentialsChanged, force b
 					result, err = transientError(errors.Wrap(err, "failed to update host settings in ironic"))
 					return
 				}
+				p.log.Info("updated host settings in ironic")
 			}
 		}
 	} else {
@@ -491,6 +493,7 @@ func (p *ironicProvisioner) ValidateManagementAccess(credentialsChanged, force b
 		provID = ironicNode.UUID
 
 		if ironicNode.Name == "" {
+			p.log.Info("updating ironic node name")
 			updates := nodes.UpdateOpts{
 				nodes.UpdateOperation{
 					Op:    nodes.ReplaceOp,
@@ -516,6 +519,7 @@ func (p *ironicProvisioner) ValidateManagementAccess(credentialsChanged, force b
 		// Look for the case where we previously enrolled this node
 		// and now the credentials have changed.
 		if credentialsChanged {
+			p.log.Info("updating host driver settings")
 			updates := nodes.UpdateOpts{
 				nodes.UpdateOperation{
 					Op:    nodes.ReplaceOp,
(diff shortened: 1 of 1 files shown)


s3rj1k:
@dtantsur test-integration did not trigger, thanks

elfosardo (approved):

s3rj1k:
Is it safe to merge?
@elfosardo @stbenjam

s3rj1k:
Any news on this?

s3rj1k:
@dhellmann Can you please help out with this?

dhellmann:
This seems like it's going to make the logs very verbose. Do the new messages help when debugging a specific problem?

s3rj1k:
@dhellmann Yea, they helped to track a node state and debug broken ironic config.

dhellmann:
> @dhellmann Yea, they helped to track a node state and debug broken ironic config.

Are there ways we can/should add more detail to some specific error handling cases? Because logging before & after steps of every API call into Ironic feels excessive.

s3rj1k:
> > @dhellmann Yea, they helped to track a node state and debug broken ironic config.
> 
> Are there ways we can/should add more detail to some specific error handling cases? Because logging before & after steps of every API call into Ironic feels excessive.

We could add debug level for logs, I saw somewhere PR about debug level logger, seems like a good idea.

dhellmann:
Rather than adding debug log messages that require restarting the container, I really meant could we be more explicit in some of the existing messages or add a few targeted messages in places where we're missing useful feedback.

s3rj1k:
> Rather than adding debug log messages that require restarting the container, I really meant could we be more explicit in some of the existing messages or add a few targeted messages in places where we're missing useful feedback.

Would be super nice to have all RW function calls to ironic logged.

Like 
 - setting Maintenance Flag with previous and current value and possibly a reason why it was set
 - changing deploy_interface with with previous and current value and  a reason why it was changed
 - changing node name, ...

dhellmann:
> > Rather than adding debug log messages that require restarting the container, I really meant could we be more explicit in some of the existing messages or add a few targeted messages in places where we're missing useful feedback.
> 
> Would be super nice to have all RW function calls to ironic logged.
> 
> Like
> 
> * setting Maintenance Flag with previous and current value and possibly a reason why it was set
> * changing deploy_interface with with previous and current value and  a reason why it was changed
> * changing node name, ...

Or general approach to error handling has been to expose the error through the API. Some users won't have access to the logs for debugging, so if we lean too much on logs for debugging the project is harder to use.

We've also tried to avoid exposing the implementation details of using Ironic too much. It's not a secret, but if the k8s user can't figure out what's going on from the API and fix it using the API, then we have a bad bug.

s3rj1k:
> > > Rather than adding debug log messages that require restarting the container, I really meant could we be more explicit in some of the existing messages or add a few targeted messages in places where we're missing useful feedback.
> > 
> > 
> > Would be super nice to have all RW function calls to ironic logged.
> > Like
> > 
> > * setting Maintenance Flag with previous and current value and possibly a reason why it was set
> > * changing deploy_interface with with previous and current value and  a reason why it was changed
> > * changing node name, ...
> 
> Or general approach to error handling has been to expose the error through the API. Some users won't have access to the logs for debugging, so if we lean too much on logs for debugging the project is harder to use.
> 
> We've also tried to avoid exposing the implementation details of using Ironic too much. It's not a secret, but if the k8s user can't figure out what's going on from the API and fix it using the API, then we have a bad bug.

Throwing k8s events maybe?

Recompiling operator with custom patches just to debug ironic also seems a bit of overkill :)

dhellmann:
Yes, I agree that asking the user to add their own debug messages is bad. :-)

You mentioned that adding these log messages helped you discover a problem with the Ironic configuration. What was the nature of the problem? How could we have reported that problem more clearly or exposed it in a more discoverable place?

s3rj1k:
> Yes, I agree that asking the user to add their own debug messages is bad. :-)
> 
> You mentioned that adding these log messages helped you discover a problem with the Ironic configuration. What was the nature of the problem? How could we have reported that problem more clearly or exposed it in a more discoverable place?

Issue was with constantly changing deploy_interface and custom ironic config, this started happening after https://github.com/metal3-io/baremetal-operator/commit/0f12bb2eabdffd734020a9806ed20f481ffe85f5 .

Enabling debug logging in ironic and matching them to logs from baremetal operator (ironic.go state changes) helped to find that deploy_interface was changed unexpectedly between `inspection` and `provisioning`.

Also please note that when using vanilla ironic image from metal3-io everything work as expected.

PR #795, opened 2021-02-15
Implement explicit reboot mode options

rdoxenham:
The default reboot-interface behaviour is to attempt a soft power
off, and if this fails, revert to a hard power off (PR #294). For high
availability use-cases we require the ability to immediately power-off
a node. This PR attempts to address that requirement and is a WIP given
that other changes are going to be required to the provisioner API and
the CAPBM actuator to actually enact these changes.

Also see: https://bugzilla.redhat.com/show_bug.cgi?id=1927678
And design doc: metal3-io/metal3-docs#164


controllers/metal3.io/baremetalhost_controller_test.go
```diff
@@ -363,36 +363,45 @@ func TestHasRebootAnnotation(t *testing.T) {
 	host := newDefaultHost(t)
 	host.Annotations = make(map[string]string)
 
-	if hasRebootAnnotation(host) {
+	// FIXME (rdo) Here we ignore the isHardReboot boolean from
+	// hasRebootAnnotation(). We may want to expand the unit-test
+	// to validate the hard poweroff functionality as we build
+	// out such capabilities.
+	hasReboot, _ := hasRebootAnnotation(host)
```
hardys, line 370: Instead of ignoring the hardMode value, I'd do e.g:

```
hasReboot, hardMode := hasRebootAnnotation(host)
assert.False(t, hasReboot)
assert.False(t, hardMode)
```

Then you can also add some new cases that cover the case where `hardMode=true`

rdoxenham: Thanks @hardys, I've added a new test case for the hard/soft modes and shifted the original test case to utilising assert.

controllers/metal3.io/baremetalhost_controller_test.go
```diff
@@ -363,48 +363,76 @@ func TestHasRebootAnnotation(t *testing.T) {
...
+	host.Annotations = make(map[string]string)
+	suffixedAnnotation := rebootAnnotationPrefix + "/foo/"
+	host.Annotations[suffixedAnnotation] = "{\"mode\": \"hard\"}"
+
+	hasReboot, hardMode = hasRebootAnnotation(host)
+	assert.True(t, hasReboot)
+	assert.True(t, hardMode)
+
+	delete(host.Annotations, suffixedAnnotation)
+	host.Annotations[rebootAnnotationPrefix] = "{\"mode\": \"soft\"}"
```
hardys, line 417: You might also want to add a test for the case where the value isn't valid json, which should also result in hardMode==false

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -339,12 +339,28 @@ func (r *BareMetalHostReconciler) credentialsErrorResult(err error, request ctrl
...
-func hasRebootAnnotation(host *metal3v1alpha1.BareMetalHost) bool {
-	for annotation := range host.Annotations {
+func hasRebootAnnotation(host *metal3v1alpha1.BareMetalHost) (hasReboot, hardMode bool) {
+	for annotation, value := range host.GetAnnotations() {
 		if isRebootAnnotation(annotation) {
-			return true
+			hasReboot = true
+			if isHardReboot(value) {
+				hardMode = true
+			}
```
dhellmann, line 348: A break here would make sense.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -339,12 +339,28 @@ func (r *BareMetalHostReconciler) credentialsErrorResult(err error, request ctrl
...
 	}
+	return
+}
+
+// isHardReboot checks if reboot annotation requests a hard reboot and returns true if so
+func isHardReboot(annotation string) bool {
+	annotation_map := map[string]string{}
+	err := json.Unmarshal([]byte(annotation), &annotation_map)
+	if err != nil {
+		return false
```
dhellmann, line 359: We should log the parse error here so a user can see what's going on if the system isn't behaving as they expect.

dhellmann (commented):
This looks like a good start. I have a few comments inline, and there's the TODO item for making a public hard power off API in the provisioner.

We should also make sure https://github.com/metal3-io/metal3-docs/pull/164 is fully up to date with the new API and that it's approved before this PR.

controllers/metal3.io/baremetalhost_controller.go
rdoxenham: @dhellmann you mean break once it's found the first reboot annotation? If so, I considered that, but then I remembered that there may be multiple clients here that could set a reboot annotation. First one may be soft, the second one hard, in which we'd want the hard one to be honoured.

dhellmann: Oh, good point, yes. Maybe a comment explaining why we don't want to break, to help *cough* "future reviewers" then. :-)

controllers/metal3.io/baremetalhost_controller.go
hardys: Good point - @dhellmann do you think just logging the event is sufficient, or should we generate an event to make it more visible to the user?

dhellmann: If we can produce an event without stopping what we're doing, that would be good as well.

controllers/metal3.io/baremetalhost_controller.go
rdoxenham: Sounds good :-)

controllers/metal3.io/baremetalhost_controller_test.go
rdoxenham: Done, thanks for the suggestion!

controllers/metal3.io/baremetalhost_controller.go
hardys: I was thinking we could log then call `info.publishEvent` similar to some other places in this code

dhellmann: Yeah, that sounds right. I couldn't remember how event publishing worked, but if it's already in use here in the provisioner I agree it makes sense to use in this case.

rdoxenham: Thanks, this is now taken care of.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -339,12 +339,32 @@ func (r *BareMetalHostReconciler) credentialsErrorResult(err error, request ctrl
...
 	}
+	return
+}
+
+// isHardReboot checks if reboot annotation requests a hard reboot and returns true if so
+func isHardReboot(annotation string, info *reconcileInfo) bool {
+	annotation_map := map[string]string{}
+	err := json.Unmarshal([]byte(annotation), &annotation_map)
+	if err != nil {
+		info.publishEvent("InvalidAnnotationValue", "reboot annotation has invalid mode, assuming soft-reboot")
```
dhellmann, line 361: This message would be more helpful if it included the error message from `Unmarshall`, too, like the log message does.

```suggestion
		info.publishEvent("InvalidAnnotationValue", fmt.Sprintf("could not parse reboot annotation (%s), assuming soft reboot mode", err))
```

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -339,12 +339,32 @@ func (r *BareMetalHostReconciler) credentialsErrorResult(err error, request ctrl
...
+	return
+}
+
+// isHardReboot checks if reboot annotation requests a hard reboot and returns true if so
+func isHardReboot(annotation string, info *reconcileInfo) bool {
+	annotation_map := map[string]string{}
+	err := json.Unmarshal([]byte(annotation), &annotation_map)
+	if err != nil {
+		info.publishEvent("InvalidAnnotationValue", fmt.Sprintf("could not parse reboot annotation (%s), assuming soft reboot mode", err))
+		info.log.Info("Could not parse reboot annotation value: " + err.Error())
```
n1r1, line 362: It would be helpful to add more details to the log message - the bmh name and the value that failed to be parsed.
I'd also add the "assuming soft reboot mode" text to the log message as well.

hardys: I think the logger already includes the BMH name, so we shouldn't need to explicitly add it to the event/log string.

Agree adding the value would be helpful in the logs, otherwise we just get something like `invalid character '}' after object key`

Also we're saying `reboot annotation (%s)` but putting err in the string for the event - is that what we want, or should that show the value we found to be invalid, e.g `could not parse reboot annotation value ("{\"thisisabadvalue\"}") - invalid json, assuming soft-reboot`

pkg/provisioner/provisioner.go
```diff
@@ -85,7 +85,7 @@ type Provisioner interface {
 
 	// PowerOff ensures the server is powered off independently of any image
 	// provisioning operation.
```
hardys, line 87: We should add something in the comment about the boolean, something like `the boolean argument may be used to specify if a hard-reboot (force power-off) is required`

pkg/provisioner/ironic/power_test.go
```diff
@@ -199,7 +199,8 @@ func TestPowerOff(t *testing.T) {
 			}
 
 			prov.status.ID = nodeUUID
-			result, err := prov.PowerOff()
+			// We pass false here to default to a soft power off
+			result, err := prov.PowerOff(false)
```
hardys, line 203: I'd add a new boolean variable to the test-case struct above, then this can be e.g `tc.hardMode` - and we can add some new test-cases that exercise the new code path with `hardMode=true`

rdoxenham: Yep, good idea. I've got that covered.

pkg/provisioner/provisioner.go
rdoxenham: Agreed. I've taken care of this.

controllers/metal3.io/baremetalhost_controller.go
rdoxenham: Thanks, taken care of I believe :)

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -780,8 +806,11 @@ func (r *BareMetalHostReconciler) manageHostPower(prov provisioner.Provisioner,
 	info.log.Info("power state change needed",
 		"expected", desiredPowerOnState,
 		"actual", info.host.Status.PoweredOn,
+		"hard reboot", desiredHardReboot,
 		"reboot process", desiredPowerOnState != info.host.Spec.Online)
 
+	// WIP (rdo) The isHardReboot functionality is a WIP as the provisioner
+	// API needs changes to allow us to call hardPowerOff here if requested
```
hardys, line 813: I think you'll need to switch the order of the commits, then you can remove this and pass `desiredHardReboot` into `prov.PowerOff()` below?

rdoxenham: That's correct. The other commit listed in the PR adds the desiredHardReboot into the PowerOff()

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -785,7 +785,7 @@ func (r *BareMetalHostReconciler) manageHostPower(prov provisioner.Provisioner,
 	if desiredPowerOnState {
 		provResult, err = prov.PowerOn()
 	} else {
-		provResult, err = prov.PowerOff()
+		provResult, err = prov.PowerOff(desiredHardReboot)
```
hardys, line 817: This change should be in the second commit - I'd suggest a fixup commit to remove it from this one, then another to add it to the other baremetalhost_controller.go changes (and remove the FIXME comment)

rdoxenham: OK I think I've done it :-)

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -339,12 +339,36 @@ func (r *BareMetalHostReconciler) credentialsErrorResult(err error, request ctrl
...
+				hardMode = true
+			}
 		}
 	}
+	return
+}
+
+// isHardReboot checks if reboot annotation requests a hard reboot and returns true if so
+func isHardReboot(annotation string, info *reconcileInfo) bool {
+	annotation_map := map[string]string{}
```
dhellmann, line 358: Can we use a real type here, based on a `struct`?

rdoxenham: Yep, I've changed this to a struct. Let me know if that's what you were thinking? Thanks!

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -339,11 +339,41 @@ func (r *BareMetalHostReconciler) credentialsErrorResult(err error, request ctrl
...
+			}
+		}
+	}
+	return
+}
+
+// isHardReboot checks if reboot annotation requests a hard reboot and returns true if so
+func isHardReboot(annotation string, info *reconcileInfo) bool {
+
+	type Annotations struct {
```
dhellmann, line 359: API types should really be defined in `apis/metal3.io/v1alpha1/baremetalhost_types.go`. This might have a name like `RebootAnnotationArguments`.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -339,11 +339,41 @@ func (r *BareMetalHostReconciler) credentialsErrorResult(err error, request ctrl
...
+		Mode	string	`json:"mode"`
+		Value	string	`json:"value"`
```
dhellmann, line 361: We don't need the second field, right? We only accept a mode argument, and its value will go into the `Mode` field.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -339,11 +339,41 @@ func (r *BareMetalHostReconciler) credentialsErrorResult(err error, request ctrl
...
+		}
+	}
+	return
+}
+
+// isHardReboot checks if reboot annotation requests a hard reboot and returns true if so
+func isHardReboot(annotation string, info *reconcileInfo) bool {
+
+	type Annotations struct {
+		Mode	string	`json:"mode"`
```
dhellmann, line 360: The type for `Mode` should be specified so it's clear we only expect the two values defined by the API. We do something similar with the status values.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -339,11 +339,41 @@ func (r *BareMetalHostReconciler) credentialsErrorResult(err error, request ctrl
...
+		Value	string	`json:"value"`
+	}
+
+	annotations := Annotations{}
+	err := json.Unmarshal([]byte(annotation), &annotations)
+	if err != nil {
+		if annotation != "" {
```
dhellmann, line 367: Instead of testing this inside the error handler, we could skip trying to parse it at all and return early.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -339,11 +339,41 @@ func (r *BareMetalHostReconciler) credentialsErrorResult(err error, request ctrl
...
+			info.publishEvent("InvalidAnnotationValue", fmt.Sprintf("could not parse reboot annotation (%s) - invalid json, assuming soft-reboot", annotation))
+			info.log.Info(fmt.Sprintf("Could not parse reboot annotation ('%s') - invalid json, assuming soft-reboot", annotation))
+		} else {
+			info.log.Info("No reboot annotation value specified, assuming soft-reboot.")
 		}
+		return false
+	}
+	if annotations.Mode == "hard" {
```
dhellmann, line 375: Instead of checking the string literal, this should check a const defined in the types file with the struct.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -339,11 +339,41 @@ func (r *BareMetalHostReconciler) credentialsErrorResult(err error, request ctrl
...
+		Mode	string	`json:"mode"`
+		Value	string	`json:"value"`
+	}
+
+	annotations := Annotations{}
+	err := json.Unmarshal([]byte(annotation), &annotations)
+	if err != nil {
+		if annotation != "" {
+			info.publishEvent("InvalidAnnotationValue", fmt.Sprintf("could not parse reboot annotation (%s) - invalid json, assuming soft-reboot", annotation))
+			info.log.Info(fmt.Sprintf("Could not parse reboot annotation ('%s') - invalid json, assuming soft-reboot", annotation))
```
dhellmann, line 369:
```suggestion
			info.log.Info(fmt.Sprintf("Could not parse reboot annotation (%q) - invalid json, assuming soft-reboot", annotation))
```

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -339,11 +339,41 @@ func (r *BareMetalHostReconciler) credentialsErrorResult(err error, request ctrl
...
+	type Annotations struct {
+		Mode	string	`json:"mode"`
+		Value	string	`json:"value"`
+	}
+
+	annotations := Annotations{}
+	err := json.Unmarshal([]byte(annotation), &annotations)
+	if err != nil {
+		if annotation != "" {
+			info.publishEvent("InvalidAnnotationValue", fmt.Sprintf("could not parse reboot annotation (%s) - invalid json, assuming soft-reboot", annotation))
```
dhellmann, line 368:
```suggestion
			info.publishEvent("InvalidAnnotationValue", fmt.Sprintf("could not parse reboot annotation (%q) - invalid json, assuming soft-reboot", annotation))
```

pkg/provisioner/provisioner.go
```diff
@@ -84,8 +84,9 @@ type Provisioner interface {
 	PowerOn() (result Result, err error)
 
 	// PowerOff ensures the server is powered off independently of any image
-	// provisioning operation.
-	PowerOff() (result Result, err error)
+	// provisioning operation. The boolean argument may be used to specify
+	// if a hard reboot (force power off) is required - true if so.
+	PowerOff(hardMode bool) (result Result, err error)
```
dhellmann, line 89: Instead of a bool, this argument should be the same type used to define the `Mode` field. That way if we add more modes, we don't have to change the argument.

controllers/metal3.io/baremetalhost_controller.go
rdoxenham: No problem. I'll add it there.

controllers/metal3.io/baremetalhost_controller.go
rdoxenham: Agreed, this was left over from my golang playground test code. Thanks for catching it.

controllers/metal3.io/baremetalhost_controller.go
rdoxenham: Thanks!

controllers/metal3.io/baremetalhost_controller.go
rdoxenham: Sounds good.

pkg/provisioner/provisioner.go
beekhof: Likewise, `hasRebootAnnotation()` would want to return a mode instead of a "hard" bool

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -483,16 +483,17 @@ type CredentialsStatus struct {
...
-type RebootAnnotation string
+// RebootMode defines known variations of reboot modes
+type RebootMode string
 
 const (
-	RebootAnnotationHard RebootAnnotation = "hard"
-	RebootAnnotationSoft RebootAnnotation = "soft"
+	RebootModeHard RebootMode = "hard"
+	RebootModeSoft RebootMode = "soft"
+	RebootModeNone RebootMode = ""
```
hardys, line 492: I don't think you need this specified explicitly, because the zero-value of a string is always an empty string in golang:

https://tour.golang.org/basics/12

rdoxenham: thanks @hardys I was just following precedent, e.g. https://github.com/metal3-io/baremetal-operator/blob/master/apis/metal3.io/v1alpha1/baremetalhost_types.go#L156

I'm using this comparison to return a RebootMode even when there's no reboot requested, as it helps with the tests, but I can revert if you think best.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -339,12 +339,46 @@ func (r *BareMetalHostReconciler) credentialsErrorResult(err error, request ctrl
...
 	}
+	return
+}
+
+// isHardReboot checks if reboot annotation requests a hard reboot and returns true if so
+func isHardReboot(annotation string, info *reconcileInfo) bool {
+
+	type RebootAnnotationArguments struct {
+		Mode	metal3v1alpha1.RebootMode	`json:"mode"`
+	}
```
n1r1, line 365: why this type being declared twice?
can't we just use `RebootAnnotationArguments` that was declared in `baremetalhost_types.go`?

apis/metal3.io/v1alpha1/baremetalhost_types.go
rdoxenham: I've cleaned this up as you're right, it's not required even if other statuses have it.

controllers/metal3.io/baremetalhost_controller.go
rdoxenham: Doh! You're right, this is a carry over from my testing. Fixed.

pkg/provisioner/provisioner.go
rdoxenham: Thanks @beekhof and @dhellmann - I've updated this now.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -765,7 +792,9 @@ func (r *BareMetalHostReconciler) manageHostPower(prov provisioner.Provisioner,
 
 	provState := info.host.Status.Provisioning.State
 	isProvisioned := provState == metal3v1alpha1.StateProvisioned || provState == metal3v1alpha1.StateExternallyProvisioned
-	if hasRebootAnnotation(info.host) && isProvisioned {
+
+	desiredReboot, desiredHardReboot := hasRebootAnnotation(info)
```
dhellmann, line 796:
```suggestion
	desiredReboot, desiredRebootMode := hasRebootAnnotation(info)
```

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -780,12 +809,13 @@ func (r *BareMetalHostReconciler) manageHostPower(prov provisioner.Provisioner,
 	info.log.Info("power state change needed",
 		"expected", desiredPowerOnState,
 		"actual", info.host.Status.PoweredOn,
+		"reboot mode", desiredHardReboot,
```
dhellmann, line 812:
```suggestion
		"reboot mode", desiredRebootMode,
```

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -780,12 +809,13 @@ func (r *BareMetalHostReconciler) manageHostPower(prov provisioner.Provisioner,
...
 		"reboot process", desiredPowerOnState != info.host.Spec.Online)
 
 	if desiredPowerOnState {
 		provResult, err = prov.PowerOn()
 	} else {
-		provResult, err = prov.PowerOff()
+		provResult, err = prov.PowerOff(desiredHardReboot)
```
dhellmann, line 818:
```suggestion
		provResult, err = prov.PowerOff(desiredRebootMode)
```

dhellmann:
This looks really close. The commit messages don't match the code (there's no isHardReboot now) so fixing that up should make this ready to go, as soon as the design is approved.

rdoxenham:
> This looks really close. The commit messages don't match the code (there's no isHardReboot now) so fixing that up should make this ready to go, as soon as the design is approved.

Thanks @dhellmann, commit messages cleaned up and I think they pretty much align with what we've done.

dhellmann:
This looks good, let's see what the tests say.

I don't see this annotation documented in this repo anywhere at all, so I won't hold up the code change over adding documentation. Maybe we can add that in another PR.

dhellmann:
/retitle support hard reboot mode via annotation

dhellmann:
/retitle Implement explicit reboot mode options

That should make the title match the design doc PR in https://github.com/metal3-io/metal3-docs/pull/164 a little better.

dhellmann:
I'm happy with this version, so after the design doc merges next week if there are no other negative comments on this PR it should be able to merge quickly.

/approve

PR #796, opened 2021-02-15
Add inspect.metal3.io/hardwaredetails annotation

hardys:
As described in https://github.com/metal3-io/metal3-docs/pull/155

This enables addition of only the hardware part of the status,
unlike the status annotation which can only be set on the first
reconcile, since it allows all fields of the status to be modified.

### DIFF
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -52,6 +52,7 @@ const (
 	provisionerNotReadyRetryDelay = time.Second * 30
 	rebootAnnotationPrefix        = "reboot.metal3.io"
 	inspectAnnotationPrefix       = "inspect.metal3.io"
+	hardwareDetailsAnnotation     = inspectAnnotationPrefix + "/hardwaredetails"
 )
 
 // BareMetalHostReconciler reconciles a BareMetalHost object
@@ -152,6 +153,14 @@ func (r *BareMetalHostReconciler) Reconcile(request ctrl.Request) (result ctrl.R
 			}
 			return ctrl.Result{Requeue: true}, nil
 		}
+	}
+
+	// Consume hardwaredetails from annotation if present
+	hwdUpdated, err := r.updateHardwareDetails(request, host)
+	if err != nil {
+		return ctrl.Result{}, errors.Wrap(err, "Could not update Hardware Details")
+	} else if hwdUpdated {
+		return ctrl.Result{Requeue: true}, nil
 	}
 
 	// NOTE(dhellmann): Handle a few steps outside of the phase
@@ -257,6 +266,42 @@ func (r *BareMetalHostReconciler) Reconcile(request ctrl.Request) (result ctrl.R
 	logResult(info, result)
 
 	return
+}
+
+// Consume inspect.metal3.io/hardwaredetails when either
+// inspect.metal3.io=disabled or there are no existing HardwareDetails
+func (r *BareMetalHostReconciler) updateHardwareDetails(request ctrl.Request, host *metal3v1alpha1.BareMetalHost) (bool, error) {
+	updated := false
+	if host.Status.HardwareDetails == nil || inspectionDisabled(host) {
+		objHardwareDetails, err := r.getHardwareDetailsFromAnnotation(host)
+		if err != nil {
+			return updated, errors.Wrap(err, "Error getting HardwareDetails from annotation")
+		}
+		if objHardwareDetails != nil {
+			host.Status.HardwareDetails = objHardwareDetails
+			err = r.saveHostStatus(host)
+			if err != nil {
+				return updated, errors.Wrap(err, "Could not update hardwaredetails from annotation")
+			}
+			r.publishEvent(request, host.NewEvent("UpdateHardwareDetails", "Set HardwareDetails from annotation"))
+			updated = true
+		}
+	}
+	// We either just processed the annotation, or the status is already set
+	// so we remove it
+	annotations := host.GetAnnotations()
+	if _, present := annotations[hardwareDetailsAnnotation]; present {
+		delete(host.Annotations, hardwareDetailsAnnotation)
+		err := r.Update(context.TODO(), host)
+		if err != nil {
+			return updated, errors.Wrap(err, "Could not update removing hardwaredetails annotation")
+		}
+		// In the case where the value was not just consumed, generate an event
+		if updated != true {
+			r.publishEvent(request, host.NewEvent("RemoveAnnotation", "HardwareDetails annotation ignored, status already set and inspection is not disabled"))
+		}
+	}
+	return updated, nil
 }
 
 func logResult(info *reconcileInfo, result ctrl.Result) {
@@ -986,6 +1031,20 @@ func (r *BareMetalHostReconciler) getHostStatusFromAnnotation(host *metal3v1alph
 		return nil, err
 	}
 	return objStatus, nil
+}
+
+// extract HardwareDetails from annotation if present
+func (r *BareMetalHostReconciler) getHardwareDetailsFromAnnotation(host *metal3v1alpha1.BareMetalHost) (*metal3v1alpha1.HardwareDetails, error) {
+	annotations := host.GetAnnotations()
+	if annotations[hardwareDetailsAnnotation] == "" {
+		return nil, nil
+	}
+	content := []byte(annotations[hardwareDetailsAnnotation])
+	objHardwareDetails := &metal3v1alpha1.HardwareDetails{}
+	if err := json.Unmarshal(content, objHardwareDetails); err != nil {
+		return nil, err
+	}
+	return objHardwareDetails, nil
 }
 
 func (r *BareMetalHostReconciler) setErrorCondition(request ctrl.Request, host *metal3v1alpha1.BareMetalHost, errType metal3v1alpha1.ErrorType, message string) (err error) {
--- a/docs/inspectAnnotation.md
+++ b/docs/inspectAnnotation.md
@@ -1,0 +1,35 @@
+# Inspect Annotation
+
+Baremetalhost's(BMH) _Status_ sub-resource contains a _hardware_ key
+which contains the result of introspection which is carried out during
+BMH registration.
+
+In some circumstances it may be desirable to disable this inspection process,
+and provide data from external source.  The _Inspect Annotation_ provides some
+interfaces to enable this.
+
+Note the `inspect.metal3.io/hardwaredetails` annotation is consumed:
+
+* At any time when `inspect.metal3.io: disabled` is specified
+* When there is no existing HardwareDetails data in the Status
+
+The `inspect.metal3.io/hardwaredetails` annotation will be removed when
+successfully processed or when the status is already set, generating an
+event in each case.
+
+The structure of the annotation's value should match the hardware status
+field schema, or a subset of that schema, for example:
+
+```yaml
+inspect.metal3.io: disabled
+inspect.metal3.io/hardwaredetails: '{"systemVendor":{"manufacturer":"QEMU",
+"productName":"Standard PC (Q35 + ICH9, 2009)","serialNumber":""},
+"firmware":{"bios":{"date":"","vendor":"","version":""}},"ramMebibytes":4096,
+"nics":[{"name":"eth0","model":"0x1af4 0x0001","mac":"00:b7:8b:bb:3d:f6",
+"ip":"172.22.0.64","speedGbps":0,"vlanId":0,"pxe":true}],
+"storage":[{"name":"/dev/sda","rotational":true,"sizeBytes":53687091200,
+"vendor":"QEMU","model":"QEMU HARDDISK","serialNumber":"drive-scsi0-0-0-0",
+"hctl":"6:0:0:0"}],"cpu":{"arch":"x86_64",
+"model":"Intel Xeon E3-12xx v2 (IvyBridge)","clockMegahertz":2494.224,
+"flags":["foo"],"count":4},"hostname":"hwdAnnotation-0"}'
+```
--- a/controllers/metal3.io/baremetalhost_controller_test.go
+++ b/controllers/metal3.io/baremetalhost_controller_test.go
@@ -32,6 +32,7 @@ const (
 	namespace         string = "test-namespace"
 	defaultSecretName string = "bmc-creds-valid" // #nosec
 	statusAnnotation  string = `{"operationalStatus":"OK","lastUpdated":"2020-04-15T15:00:50Z","hardwareProfile":"StatusProfile","hardware":{"systemVendor":{"manufacturer":"QEMU","productName":"Standard PC (Q35 + ICH9, 2009)","serialNumber":""},"firmware":{"bios":{"date":"","vendor":"","version":""}},"ramMebibytes":4096,"nics":[{"name":"eth0","model":"0x1af4 0x0001","mac":"00:b7:8b:bb:3d:f6","ip":"172.22.0.64","speedGbps":0,"vlanId":0,"pxe":true},{"name":"eth1","model":"0x1af4  0x0001","mac":"00:b7:8b:bb:3d:f8","ip":"192.168.111.20","speedGbps":0,"vlanId":0,"pxe":false}],"storage":[{"name":"/dev/sda","rotational":true,"sizeBytes":53687091200,"vendor":"QEMU","model":"QEMU HARDDISK","serialNumber":"drive-scsi0-0-0-0","hctl":"6:0:0:0"}],"cpu":{"arch":"x86_64","model":"Intel Xeon E3-12xx v2 (IvyBridge)","clockMegahertz":2494.224,"flags":["aes","apic","arat","avx","clflush","cmov","constant_tsc","cx16","cx8","de","eagerfpu","ept","erms","f16c","flexpriority","fpu","fsgsbase","fxsr","hypervisor","lahf_lm","lm","mca","mce","mmx","msr","mtrr","nopl","nx","pae","pat","pclmulqdq","pge","pni","popcnt","pse","pse36","rdrand","rdtscp","rep_good","sep","smep","sse","sse2","sse4_1","sse4_2","ssse3","syscall","tpr_shadow","tsc","tsc_adjust","tsc_deadline_timer","vme","vmx","vnmi","vpid","x2apic","xsave","xsaveopt","xtopology"],"count":4},"hostname":"node-0"},"provisioning":{"state":"provisioned","ID":"8a0ede17-7b87-44ac-9293-5b7d50b94b08","image":{"url":"bar","checksum":""}},"goodCredentials":{"credentials":{"name":"node-0-bmc-secret","namespace":"metal3"},"credentialsVersion":"879"},"triedCredentials":{"credentials":{"name":"node-0-bmc-secret","namespace":"metal3"},"credentialsVersion":"879"},"errorMessage":"","poweredOn":true,"operationHistory":{"register":{"start":"2020-04-15T12:06:26Z","end":"2020-04-15T12:07:12Z"},"inspect":{"start":"2020-04-15T12:07:12Z","end":"2020-04-15T12:09:29Z"},"provision":{"start":null,"end":null},"deprovision":{"start":null,"end":null}}}`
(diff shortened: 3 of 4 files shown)


controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -154,6 +155,33 @@ func (r *BareMetalHostReconciler) Reconcile(request ctrl.Request) (result ctrl.R
...
+		if _, present := annotations[hardwareDetailsAnnotation]; present {
+			delete(annotations, hardwareDetailsAnnotation)
+			errStatus := r.Update(context.TODO(), host)
+			if errStatus != nil {
+				return ctrl.Result{}, errors.Wrap(errStatus, "Could not delete hardwaredetails annotation")
+			}
+			return ctrl.Result{Requeue: true}, nil
+		}
+	}
+
```
andfasano, line 184: The impression is that this new block, along with the previous one, represent an important preliminary step for properly restore a resource status from annotations, so they could be a good candidates to be refactored into a separate func - and this could also improve the readability of the reconcile loop.

In addition, what about coalescing the two update operations and postponing the `saveHostStatus` - if one of the two blocks modifies the status? In this way we could avoid one reconcile loop when both annotations are set

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -52,6 +52,7 @@ const (
 	provisionerNotReadyRetryDelay = time.Second * 30
 	rebootAnnotationPrefix        = "reboot.metal3.io"
 	inspectAnnotationPrefix       = "inspect.metal3.io"
+	hardwareDetailsAnnotation     = inspectAnnotationPrefix + "/hardwaredetails"
```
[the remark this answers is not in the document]
hardys: Thanks for the reminder!  https://github.com/metal3-io/metal3-docs/pull/155 now updated.

controllers/metal3.io/baremetalhost_controller.go
hardys: I think it's unlikely a user will ever want to set both annotations, since the hardwaredetails annotation is a subset of the existing status annotation, also the status annotation will only ever be evaluated on the very first reconcile loop.

I can certainly refactor them into a separate function though, you're right that will make the reconcile function a bit more readable, and it's already pretty big.

hardys: Ok I added a new function for the new code, if it's OK with you @andfasano I'll propose a follow-up for refactoring the existing code, since it's not strictly related to this feature PR?

andfasano: Sure. My feeling is that such kind of refactoring could also help in testing the feature (in more isolation)

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -259,6 +274,33 @@ func (r *BareMetalHostReconciler) Reconcile(request ctrl.Request) (result ctrl.R
 	return
 }
 
+// Consume inspect.metal3.io/hardwaredetails when either
+// inspect.metal3.io=disabled or there are no existing HardwareDetails
+func (r *BareMetalHostReconciler) updateHardwareDetails(request ctrl.Request, host *metal3v1alpha1.BareMetalHost) (bool, error) {
+	reqLogger := r.Log.WithValues("baremetalhost", request.NamespacedName)
```
andfasano, line 280: nit: a `baremetalhost` logger should be already available in the reconcile loop, could be passed as parameter (could save some cycles in all the cases where the logger is not needed)

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -259,6 +274,33 @@ func (r *BareMetalHostReconciler) Reconcile(request ctrl.Request) (result ctrl.R
...
+	updated := false
+	if host.Status.HardwareDetails == nil || inspectionDisabled(host) {
+		objHardwareDetails, err := r.getHardwareDetailsFromAnnotation(host)
+		if err != nil {
+			return false, err
+		}
+		if err == nil && objHardwareDetails != nil {
```
andfasano, line 287: I think `err == nil` could be safely removed?

controllers/metal3.io/baremetalhost_controller.go
hardys: Yeah I did consider that, but wanted to avoid bloating the parameter list, and several other functions already do it this way.

How about I move this inside the conditional, so it's only done when the logger is needed?

controllers/metal3.io/baremetalhost_controller.go
hardys: Yes, good point, will do!

controllers/metal3.io/baremetalhost_controller.go
andfasano: Yep, or in alternative it can be logged outside the function when `hwdUpdated` is set

hardys:
Thanks for the feedback @andfasano I've tested this locally and it now seems to be working OK, so ready for another review when you get time - also perhaps @dhellmann you could take a look please?  Thanks!

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -259,6 +268,41 @@ func (r *BareMetalHostReconciler) Reconcile(request ctrl.Request) (result ctrl.R
...
+// inspect.metal3.io=disabled or there are no existing HardwareDetails
+func (r *BareMetalHostReconciler) updateHardwareDetails(request ctrl.Request, host *metal3v1alpha1.BareMetalHost) (bool, error) {
+	reqLogger := r.Log.WithValues("baremetalhost", request.NamespacedName)
+	updated := false
+	if host.Status.HardwareDetails == nil || inspectionDisabled(host) {
+		objHardwareDetails, err := r.getHardwareDetailsFromAnnotation(host)
+		if err != nil {
+			reqLogger.Info("Error getting HardwareDetails from annotation")
+			return updated, err
+		} else if objHardwareDetails != nil {
```
dhellmann, line 281: nit: the `else` isn't needed here

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -259,6 +268,41 @@ func (r *BareMetalHostReconciler) Reconcile(request ctrl.Request) (result ctrl.R
...
+			reqLogger.Info("Setting HardwareDetails from annotation")
+			host.Status.HardwareDetails = objHardwareDetails
+			err = r.saveHostStatus(host)
+			if err != nil {
+				reqLogger.Info("Could not update hardwaredetails from annotation")
+				return updated, err
```
dhellmann, line 287: Same comment about wrapping the error.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -259,6 +268,41 @@ func (r *BareMetalHostReconciler) Reconcile(request ctrl.Request) (result ctrl.R
...
+	}
+	// We either just processed the annotation, or the status is already set
+	// so we remove it
+	annotations := host.GetAnnotations()
+	if _, present := annotations[hardwareDetailsAnnotation]; present {
+		reqLogger.Info("Removing HardwareDetails annotation")
+		delete(host.Annotations, hardwareDetailsAnnotation)
+		err := r.Update(context.TODO(), host)
+		if err != nil {
+			return updated, err
```
dhellmann, line 300: Same comment about wrapping the error.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -259,6 +268,41 @@ func (r *BareMetalHostReconciler) Reconcile(request ctrl.Request) (result ctrl.R
...
+// Consume inspect.metal3.io/hardwaredetails when either
+// inspect.metal3.io=disabled or there are no existing HardwareDetails
+func (r *BareMetalHostReconciler) updateHardwareDetails(request ctrl.Request, host *metal3v1alpha1.BareMetalHost) (bool, error) {
+	reqLogger := r.Log.WithValues("baremetalhost", request.NamespacedName)
+	updated := false
+	if host.Status.HardwareDetails == nil || inspectionDisabled(host) {
+		objHardwareDetails, err := r.getHardwareDetailsFromAnnotation(host)
+		if err != nil {
+			reqLogger.Info("Error getting HardwareDetails from annotation")
+			return updated, err
```
dhellmann, line 280: Should this error be wrapped with a message like what's logged in the line above to provide context?

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -259,6 +268,41 @@ func (r *BareMetalHostReconciler) Reconcile(request ctrl.Request) (result ctrl.R
...
+			updated = true
+		}
+	}
+	// We either just processed the annotation, or the status is already set
+	// so we remove it
+	annotations := host.GetAnnotations()
+	if _, present := annotations[hardwareDetailsAnnotation]; present {
+		reqLogger.Info("Removing HardwareDetails annotation")
+		delete(host.Annotations, hardwareDetailsAnnotation)
+		err := r.Update(context.TODO(), host)
```
dhellmann, line 298: If we save the status above, will the host be out of date and trigger an error updating it's annotations here? I remember seeing those sorts of errors early on in development, but it may have been from multiple updates to the same part of the resource or some other condition, so maybe it isn't an issue here.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -949,6 +993,28 @@ func (r *BareMetalHostReconciler) getHostStatusFromAnnotation(host *metal3v1alph
...
+}
+
+// extract host from HardwareDetails annotation
+func (r *BareMetalHostReconciler) getHardwareDetailsFromAnnotation(host *metal3v1alpha1.BareMetalHost) (*metal3v1alpha1.HardwareDetails, error) {
+	annotations := host.GetAnnotations()
+	if annotations[hardwareDetailsAnnotation] == "" {
+		return nil, nil
+	}
+	content := []byte(annotations[hardwareDetailsAnnotation])
+	objHardwareDetails, err := unmarshalHardwareDetailsAnnotation(content)
```
sadasu, line 1011: We can just return unmarshalHardwareDetailsAnnotation(content) here. unmarshalHardwareDetailsAnnotation() is already returning the right values whether there is an error or not (nil, err and objHardwareDetails, nil).

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -259,6 +268,41 @@ func (r *BareMetalHostReconciler) Reconcile(request ctrl.Request) (result ctrl.R
 	return
 }
 
+// Consume inspect.metal3.io/hardwaredetails when either
+// inspect.metal3.io=disabled or there are no existing HardwareDetails
+func (r *BareMetalHostReconciler) updateHardwareDetails(request ctrl.Request, host *metal3v1alpha1.BareMetalHost) (bool, error) {
+	reqLogger := r.Log.WithValues("baremetalhost", request.NamespacedName)
+	updated := false
+	if host.Status.HardwareDetails == nil || inspectionDisabled(host) {
```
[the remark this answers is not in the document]
sadasu: Should this condition be an && instead of an ||? Also, we might be able to not delete the annotation later (after consuming it) if this condition is an &&.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -259,6 +268,41 @@ func (r *BareMetalHostReconciler) Reconcile(request ctrl.Request) (result ctrl.R
...
+			err = r.saveHostStatus(host)
+			if err != nil {
+				reqLogger.Info("Could not update hardwaredetails from annotation")
+				return updated, err
+			}
+			updated = true
+		}
+	}
+	// We either just processed the annotation, or the status is already set
+	// so we remove it
```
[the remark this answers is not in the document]
sadasu: Based on https://github.com/metal3-io/baremetal-operator/pull/796/files#r585936511, do we need to remove the annotation at all?

controllers/metal3.io/baremetalhost_controller.go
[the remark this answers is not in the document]
hardys: > Should this condition be an && instead of an ||? That way, when the annotation is processed, then we won't read the HW details again, even if the annotation is not deleted.

No, in the [design doc](https://github.com/metal3-io/metal3-docs/pull/155/files) we agreed the following behavior wrt when to update the status:

- At any time when inspect.metal3.io=disabled
- When there is no existing HardwareDetails data in the Status

So in the logic above, we'll update the status if HardwareDetails aren't yet set, or always in the case where `inspectionDisabled(host)` is true

The annotation is immediately deleted, so we won't re-read the same annotation, but it means a user could potentially update the data in the case where inspection is disabled (which IMO makes sense, because in that case the data is externally managed and so we shouldn't prevent updates)

controllers/metal3.io/baremetalhost_controller.go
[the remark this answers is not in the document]
hardys: I copied this behavior from the status annotation code https://github.com/metal3-io/baremetal-operator/blob/master/controllers/metal3.io/baremetalhost_controller.go#L145..L153

However in that case the rationale for removing is that the status data will become outdated, which isn't the case here.

However, we do need some way to handle the following scenarios:

1. hardwaredetails annotation is set, but we ignore it (inspection is enabled, and status.hardware already set)
2. User updates hardwaredetails annotation in the case where inspection is disabled

For (1) we should perhaps remove the annotation and generate an event, rather than just ignoring it forever?

And for (2) we have to either remove the annotation (so we can detect new data because the annotation got re-applied), or we need to compare old/new status (which isn't impossible, but it'll mean json encoding the existing status and comparing the strings, which seems more complex and error prone than just removing the annotation after we consume it?)

controllers/metal3.io/baremetalhost_controller.go
hardys: I tested locally and this seems to work, but I wasn't sure since the status annotation code does return from the Reconcile() after each update - I could do the same here if we think that's safer?

Part of the problem is the unit tests don't exhibit the same behavior wrt resource updates that we see on the live cluster (we just look at the modified host object, not the stored state), so it's more difficult to be sure we get this kind of thing right.

controllers/metal3.io/baremetalhost_controller.go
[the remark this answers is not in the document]
hardys: > I think I would reverse this logic and return early if none of the conditions for consuming the annotation are met.

@dhellmann The reason I did it this way is the following code to remove the annotation is currently (intentionally) hit unconditionally, so we either remove it after consuming the annotation, or after it's ignored (as you mention in the other thread maybe this second case should generate an event to make it more visible?)

controllers/metal3.io/baremetalhost_controller.go
hardys: Done

controllers/metal3.io/baremetalhost_controller.go
hardys: Thanks, now wrapped

controllers/metal3.io/baremetalhost_controller.go
hardys: Thanks, now wrapped

controllers/metal3.io/baremetalhost_controller.go
hardys: Yeah, I was wrapping the error returned from the function, but including this context makes sense, now updated, thanks!

controllers/metal3.io/baremetalhost_controller.go
hardys: Yeah good point, it can probably be simplified into a single function, please check the latest version - thanks for the review!

controllers/metal3.io/baremetalhost_controller.go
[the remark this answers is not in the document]
dhellmann: OK, that makes sense. It wasn't clear why we always wanted to remove it, but I think if we can signal via an event that we're ignoring the input then it makes sense to remove it.

controllers/metal3.io/baremetalhost_controller.go
dhellmann: Agreed, we need to figure out a way to add end-to-end tests for some of these more complicated behaviors.

controllers/metal3.io/baremetalhost_controller.go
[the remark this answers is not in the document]
hardys: Ok the latest version generates two events, one when consuming the annotation

```
Normal  UpdateHardwareDetails  10m   metal3-baremetal-controller  Setting HardwareDetails from annotation
```
And one when ignoring it:

```
Normal  RemoveAnnotation  11m   metal3-baremetal-controller  HardwareDetails annotation ignored, status already set and inspection is not disabled
```

controllers/metal3.io/baremetalhost_controller.go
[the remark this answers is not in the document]
hardys: I'll resolve this thread, since we covered generating an events in the previous one and I added an update there

hardys:
@dhellmann @andfasano @sadasu thanks for the reviews - I added the events discussed, another test for the case where existing HardwareDetails can be overwritten (when inspection is disabled), and some docs, so I think this is ready for another view pass when you get a moment :)

controllers/metal3.io/baremetalhost_controller.go
andfasano: I think it should be ok, and agreed that we need to find a better approach to such kind of tests

andfasano:
Changes looks fine to me, leaving final comments to the others

docs/inspectAnnotation.md
```diff
@@ -0,0 +1,28 @@
+# Inspect Annotation
+
+Baremetalhost's(BMH) _Status_ sub-resource contain a _hardware_ key
+which contains the result of introspection which is carried out during
+BMH registration.
+
+In some circumstances it may be desirable to disable this inspection process,
+and provide data from external source.  The _Inspect Annotation_ provides some
+interfaces to enabled this.
```
dhellmann, line 9: nit: we should say in here that the structure of the annotation's value should match the `hardware` field's schema. We can do that as an update if there are no other reasons to change this PR.

dhellmann (approved):
I'm happy with this version. I had one suggestion for the docs, but we can fix that separately unless there's another reason to update this PR.

docs/inspectAnnotation.md
hardys: Ack, I'm happy to add it here - we should perhaps also mention it in the status annotation doc, and that could link to this mentioning that it's an alternative API if you only want to update the hardware field of the status?

hardys: I should also perhaps mention that the hardware details can be a subset of the full hardware field schema since https://github.com/metal3-io/baremetal-operator/pull/803

hardys: Ok updated in the latest revision

hardys:
@andfasano @sadasu @dhellmann I think this should be good to go now, PTAL, thanks!

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -259,6 +268,42 @@ func (r *BareMetalHostReconciler) Reconcile(request ctrl.Request) (result ctrl.R
...
+// inspect.metal3.io=disabled or there are no existing HardwareDetails
+func (r *BareMetalHostReconciler) updateHardwareDetails(request ctrl.Request, host *metal3v1alpha1.BareMetalHost) (bool, error) {
+	updated := false
+	if host.Status.HardwareDetails == nil || inspectionDisabled(host) {
+		objHardwareDetails, err := r.getHardwareDetailsFromAnnotation(host)
+		if err != nil {
+			return updated, errors.Wrap(err, "Error getting HardwareDetails from annotation")
+		}
+		if objHardwareDetails != nil {
+			r.publishEvent(request, host.NewEvent("UpdateHardwareDetails", "Setting HardwareDetails from annotation"))
```
andfasano, line 281: Wouldn't be safer to postpone events publishing just after a successful `r.saveHostStatus`?

docs/inspectAnnotation.md
```diff
@@ -0,0 +1,31 @@
...
+
+Note the `inspect.metal3.io/hardwaredetails` annotation is consumed:
```
andfasano, line 11: Could be worth mentioning here also the conditions under which the annotation will be removed (ie when successfully processed or when the status is already set)?

andfasano (approved):

controllers/metal3.io/baremetalhost_controller.go
hardys: Good point, I moved the events to after the status is stored in both cases.

docs/inspectAnnotation.md
hardys: Done, thanks!

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -988,6 +1033,20 @@ func (r *BareMetalHostReconciler) getHostStatusFromAnnotation(host *metal3v1alph
 	return objStatus, nil
 }
 
+// extract host from HardwareDetails annotation
```
honza, line 1036: nit: is this comment out of date?

docs/inspectAnnotation.md
```diff
@@ -0,0 +1,35 @@
+# Inspect Annotation
+
+Baremetalhost's(BMH) _Status_ sub-resource contain a _hardware_ key
```
honza, line 3: s/contain/contains/

docs/inspectAnnotation.md
```diff
@@ -0,0 +1,35 @@
...
+which contains the result of introspection which is carried out during
+BMH registration.
+
+In some circumstances it may be desirable to disable this inspection process,
+and provide data from external source.  The _Inspect Annotation_ provides some
+interfaces to enabled this.
```
honza, line 9: s/enabled/enable/

controllers/metal3.io/baremetalhost_controller.go
hardys: Hmm, not sure how this happened tbh, probably a copy/paste/sed typo, will update to clarify

PR #797, opened 2021-02-16
Prepend the namespace to ironic nodenames to avoid conflicts

asalkeld:
Modify the hostname for ironic to prepend the namespace to the name to avoid naming conflicts.

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -27,6 +27,10 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: metadata.name
+          - name: POD_NAMESPACE
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.namespace
         envFrom:
           - configMapRef:
               name: ironic
--- a/main.go
+++ b/main.go
@@ -103,13 +103,18 @@ func main() {
 
 	printVersion()
 
+	leaderElectionNamespace := os.Getenv("POD_NAMESPACE")
+	if leaderElectionNamespace == "" {
+		leaderElectionNamespace = watchNamespace
+	}
+
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
 		Scheme:                  scheme,
 		MetricsBindAddress:      metricsAddr,
 		Port:                    0, // Add flag with default of 9443 when adding webhooks
 		LeaderElection:          enableLeaderElection,
 		LeaderElectionID:        "baremetal-operator",
-		LeaderElectionNamespace: watchNamespace,
+		LeaderElectionNamespace: leaderElectionNamespace,
 		Namespace:               watchNamespace,
 		HealthProbeBindAddress:  healthAddr,
 	})
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -51,10 +51,11 @@ var (
 
 const (
 	// See nodes.Node.PowerState for details
-	powerOn      = "power on"
-	powerOff     = "power off"
-	softPowerOff = "soft power off"
-	powerNone    = "None"
+	powerOn       = "power on"
+	powerOff      = "power off"
+	softPowerOff  = "soft power off"
+	powerNone     = "None"
+	nameSeparator = "~"
 )
 
 var bootModeCapabilities = map[metal3v1alpha1.BootMode]string{
@@ -183,8 +184,6 @@ func newProvisionerWithSettings(host metal3v1alpha1.BareMetalHost, bmcCreds bmc.
 }
 
 func newProvisionerWithIronicClients(host metal3v1alpha1.BareMetalHost, bmcCreds bmc.Credentials, publisher provisioner.EventPublisher, clientIronic *gophercloud.ServiceClient, clientInspector *gophercloud.ServiceClient) (*ironicProvisioner, error) {
-	provisionerLogger := log.WithValues("host", host.Name)
-
 	// Ensure we have a microversion high enough to get the features
 	// we need.
 	clientIronic.Microversion = "1.56"
@@ -194,10 +193,10 @@ func newProvisionerWithIronicClients(host metal3v1alpha1.BareMetalHost, bmcCreds
 		bmcCreds:  bmcCreds,
 		client:    clientIronic,
 		inspector: clientInspector,
-		log:       provisionerLogger,
-		debugLog:  provisionerLogger.V(1),
 		publisher: publisher,
 	}
+	p.log = log.WithValues("host", p.ironicNodeNameFromHost())
+	p.debugLog = p.log.V(1)
 
 	return p, nil
 }
@@ -308,18 +307,25 @@ func (p *ironicProvisioner) findExistingHost() (ironicNode *nodes.Node, err erro
 	}
 
 	// Try to load the node by name
-	p.log.Info("looking for existing node by name", "name", p.host.Name)
-	ironicNode, err = nodes.Get(p.client, p.host.Name).Extract()
-	switch err.(type) {
-	case nil:
-		p.debugLog.Info("found existing node by name")
-		return ironicNode, nil
-	case gophercloud.ErrDefault404:
-		p.log.Info(
-			fmt.Sprintf("node with name %s doesn't exist", p.host.Name))
-	default:
-		return nil, errors.Wrap(err,
-			fmt.Sprintf("failed to find node by name %s", p.host.Name))
+	nodeSearchList := []string{p.ironicNodeNameFromHost()}
+	if !strings.Contains(p.host.Name, nameSeparator) {
+		nodeSearchList = append(nodeSearchList, p.host.Name)
+	}
+
+	for _, nodeName := range nodeSearchList {
+		p.debugLog.Info("looking for existing node by name", "name", nodeName)
+		ironicNode, err = nodes.Get(p.client, nodeName).Extract()
+		switch err.(type) {
+		case nil:
+			p.debugLog.Info("found existing node by name")
+			return ironicNode, nil
+		case gophercloud.ErrDefault404:
+			p.log.Info(
+				fmt.Sprintf("node with name %s doesn't exist", nodeName))
+		default:
+			return nil, errors.Wrap(err,
+				fmt.Sprintf("failed to find node by name %s", nodeName))
+		}
 	}
 
 	// Try to load the node by port address
@@ -501,12 +507,12 @@ func (p *ironicProvisioner) ValidateManagementAccess(credentialsChanged, force b
 		// if there are differences.
 		provID = ironicNode.UUID
 
-		if ironicNode.Name == "" {
+		if ironicNode.Name != p.ironicNodeNameFromHost() {
 			updates := nodes.UpdateOpts{
 				nodes.UpdateOperation{
 					Op:    nodes.ReplaceOp,
 					Path:  "/name",
-					Value: p.host.Name,
+					Value: p.ironicNodeNameFromHost(),
 				},
 			}
 			ironicNode, err = nodes.Update(p.client, ironicNode.UUID, updates).Extract()
@@ -1840,6 +1846,10 @@ func (p *ironicProvisioner) softPowerOff() (result provisioner.Result, err error
 	return result, nil
 }
 
+func (p *ironicProvisioner) ironicNodeNameFromHost() string {
+	return p.host.Namespace + nameSeparator + p.host.Name
+}
+
 // IsReady checks if the provisioning backend is available
 func (p *ironicProvisioner) IsReady() (result bool, err error) {
 	p.debugLog.Info("verifying ironic provisioner dependencies")
@@ -1857,7 +1867,7 @@ func (p *ironicProvisioner) HasProvisioningCapacity() (result bool, err error) {
 	}
 
 	// If the current host is already under processing then let's skip the test
-	if _, ok := hosts[p.host.Name]; ok {
+	if _, ok := hosts[p.ironicNodeNameFromHost()]; ok {
 		return true, nil
 	}
 
--- a/config/render/capm3.yaml
+++ b/config/render/capm3.yaml
@@ -1039,6 +1039,10 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.name
+        - name: POD_NAMESPACE
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
         envFrom:
         - configMapRef:
             name: baremetal-operator-ironic
--- a/pkg/provisioner/ironic/delete_test.go
+++ b/pkg/provisioner/ironic/delete_test.go
@@ -86,7 +86,7 @@ func TestDelete(t *testing.T) {
 		},
 		{
 			name:   "not-ironic-node",
-			ironic: testserver.NewIronic(t).Ready().NoNode(nodeUUID).NoNode("myhost"),
+			ironic: testserver.NewIronic(t).Ready().NoNode(nodeUUID).NoNode("myns" + nameSeparator + "myhost").NoNode("myhost"),
 		},
 		{
 			name: "available-node",
--- a/pkg/provisioner/ironic/findhost_test.go
+++ b/pkg/provisioner/ironic/findhost_test.go
@@ -23,7 +23,7 @@ func TestFindExistingHost(t *testing.T) {
 			name:           "no-node",
 			hostName:       "name",
 			provisioningID: "uuid",
-			ironic:         testserver.NewIronic(t).NoNode("name").NoNode("uuid"),
+			ironic:         testserver.NewIronic(t).NoNode("myns" + nameSeparator + "name").NoNode("name").NoNode("uuid"),
 		},
 		{
 			name:           "by-name",
@@ -31,21 +31,21 @@ func TestFindExistingHost(t *testing.T) {
 			provisioningID: "uuid",
 			ironic: testserver.NewIronic(t).NoNode("uuid").
 				Node(nodes.Node{
-					Name: "name",
+					Name: "myns" + nameSeparator + "name",
 					UUID: "different-uuid",
 				}),
-			nodeName: "name",
+			nodeName: "myns" + nameSeparator + "name",
 		},
 		{
 			name:           "by-uuid",
 			hostName:       "name",
 			provisioningID: "uuid",
-			ironic: testserver.NewIronic(t).NoNode("name").
+			ironic: testserver.NewIronic(t).NoNode("myns" + nameSeparator + "name").NoNode("name").
 				Node(nodes.Node{
-					Name: "different-name",
+					Name: "myns" + nameSeparator + "different-name",
 					UUID: "uuid",
 				}),
-			nodeName: "different-name",
+			nodeName: "myns" + nameSeparator + "different-name",
 		},
 	}
 
--- a/pkg/provisioner/ironic/provisioncapacity_test.go
+++ b/pkg/provisioner/ironic/provisioncapacity_test.go
@@ -62,7 +62,7 @@ func TestHasProvisioningCapacity(t *testing.T) {
 			allNodes := []nodes.Node{}
 			for n, state := range tc.nodeStates {
 				allNodes = append(allNodes, nodes.Node{
-					Name:           fmt.Sprintf("node-%d", n),
+					Name:           fmt.Sprintf("myns%snode-%d", nameSeparator, n),
 					ProvisionState: string(state),
 				})
 			}
--- a/pkg/provisioner/ironic/updatehardwarestate_test.go
+++ b/pkg/provisioner/ironic/updatehardwarestate_test.go
@@ -90,15 +90,15 @@ func TestUpdateHardwareState(t *testing.T) {
 			name: "node-not-found-by-name",
 
 			hostName: "worker-0",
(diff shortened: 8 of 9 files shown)


pkg/provisioner/ironic/ironic.go
```diff
@@ -198,9 +198,9 @@ func newProvisionerWithIronicClients(host metal3v1alpha1.BareMetalHost, bmcCreds
 		bmcCreds:  bmcCreds,
 		client:    clientIronic,
 		inspector: clientInspector,
-		log:       log.WithValues("host", host.Name),
 		publisher: publisher,
 	}
+	p.log = log.WithValues("host", p.hostName())
```
hardys, line 203: Was there a reason this needed to move from the ironicProvisioner arguments?

pkg/provisioner/ironic/ironic.go
```diff
@@ -1784,6 +1784,13 @@ func (p *ironicProvisioner) softPowerOff() (result provisioner.Result, err error
 	return result, nil
 }
 
+func (p *ironicProvisioner) hostName() string {
+	if p.host.Namespace == "openshift-machine-api" {
```
hardys, line 1788: IMO we shouldn't hardcode this, it's specific to the openshift deployment of the BMO.

Could we perhaps only prepend the namespace in the case where [WATCH_NAMESPACE](https://github.com/metal3-io/baremetal-operator/blob/master/main.go#L89) is not set?

In the Openshift case, can we restart the ironic containers when the single->multi-namespace [reconfiguration via the CBO](https://github.com/openshift/cluster-baremetal-operator/pull/106) happens?? /cc @sadasu @dhellmann 

If not then I guess another option is to look for an existing host (e.g the `findExistingHost()` function - it's not clear to me if we ever do the lookup by name in practice, but that function could be made to look for both the old/new name format?

dhellmann: > IMO we shouldn't hardcode this, it's specific to the openshift deployment of the BMO.

I agree, we don't want OpenShift-specific behavior hardcoded upstream.

> Could we perhaps only prepend the namespace in the case where [WATCH_NAMESPACE](https://github.com/metal3-io/baremetal-operator/blob/master/main.go#L89) is not set?

We've assumed that the name field in ironic needs to be unique. Is that actually true? How does upstream deal with the potential for name conflicts when watching all namespaces?

If the name does need to be unique, I think I would rather always have the name prefix (rather than triggering that behavior based on whether we're only watching a namespace, unless that proves difficult to implement.

> In the Openshift case, can we restart the ironic containers when the single->multi-namespace [reconfiguration via the CBO](https://github.com/openshift/cluster-baremetal-operator/pull/106) happens?? /cc @sadasu @dhellmann

In OpenShift, reconfiguring metal3 will restart the whole pod, yes.

> If not then I guess another option is to look for an existing host (e.g the `findExistingHost()` function - it's not clear to me if we ever do the lookup by name in practice, but that function could be made to look for both the old/new name format?

I honestly don't remember the use case for looking up the node in ironic by its name instead of UUID. Do we need that behavior, or was it one of my belt-and-braces choices?

On upgrade, an existing host should find it's ironic node via the UUID. Can we change the name of an existing node in ironic to match the new pattern?

asalkeld: So it should be better now (not hardcoded)..

hardys: > We've assumed that the name field in ironic needs to be unique. Is that actually true? How does upstream deal with the potential for name conflicts when watching all namespaces?

/cc @dtantsur but I think it does need to be unique, because the API supports lookup via both uuid and name node_ident (and because the BMO code does a lookup by name when it fails to find the UUID...)

I think upstream the `owner` field must be used for each tenant/project ID - so perhaps we should do the same?

https://docs.openstack.org/api-ref/baremetal/?expanded=create-node-detail

> On upgrade, an existing host should find it's ironic node via the UUID. Can we change the name of an existing node in ironic to match the new pattern?

Yes this does appear to work, if we decide not to use the owner field:

```
$ baremetal node list | grep worker-0
| 7c5d5dcb-cd32-4538-90b3-aff8523e16f4 | ostest-worker-0 | 547a2958-b1f6-4904-8c56-5a6516936e62 | None        | active             | True        |

$ baremetal node set --name testns-ostest-worker-0 ostest-worker-0
Management cluster forced to be minikube when container runtime is not docker

$ baremetal node list | grep worker-0
| 7c5d5dcb-cd32-4538-90b3-aff8523e16f4 | testns-ostest-worker-0 | 547a2958-b1f6-4904-8c56-5a6516936e62 | None        | active             | True        |
```

dtantsur: Nodes names are unique in ironic, as you've already checked. `owner` is not a namespace, names have to be globally unique. Otherwise you can use `owner` as you wish, it only has special meaning with keystone.

dhellmann (commented):
It's a little unfortunate that we have to leak the deployment configuration into the provisioner, but I don't see an elegant way to avoid it that doesn't leak the implementation of the provisioner up into `main()`.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1784,6 +1790,14 @@ func (p *ironicProvisioner) softPowerOff() (result provisioner.Result, err error
 	return result, nil
 }
 
+func (p *ironicProvisioner) hostName() string {
+	if !p.watchAllNamespaces || p.host.Namespace == p.podNamespace {
+		return p.host.Name
+	}
+
+	return p.host.Namespace + "-" + p.host.Name
```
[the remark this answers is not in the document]
asalkeld: ok, so would the "." be better? I thought it's a "hostname" and a "." might be seen as a sub-domain - but maybe this is not an issue.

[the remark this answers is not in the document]
dhellmann: What about `/`? Maybe ironic doesn't allow that?

config/manager/manager.yaml
```diff
@@ -27,6 +27,10 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: metadata.name
+          - name: POD_NAMESPACE
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.namespace
```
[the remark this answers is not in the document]
asalkeld: @dhellmann really? AFAIK it is an automatic variable that picks up the namespace from the pod that is running it.
https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/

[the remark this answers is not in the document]
asalkeld: doh, forgot it is manually defined: https://github.com/openshift/cluster-baremetal-operator/blob/dd41e00d29316cfd4a86334f63825e3279fb1715/provisioning/baremetal_pod.go#L283-L302

Here is a PR for that https://github.com/openshift/cluster-baremetal-operator/pull/107

pkg/provisioner/ironic/ironic.go
hardys: > names have to be globally unique

Thanks @dtantsur - how does OpenStack handle this then, in the case where you have Nova instances per-project which don't have globally unique names?

If the name has to be globally unique then I don't think the `owner` field helps us, so the current prepending approach is probably our only option?

dtantsur: Nova instance names are not related to node names, Nova matches ironic nodes by UUID (more precisely, hypervisor UUID which is equal to node UUID for Ironic).

pkg/provisioner/ironic/ironic.go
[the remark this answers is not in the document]
dtantsur: I think we expect it to be a valid hostname, I don't think we allow `/`

pkg/provisioner/ironic/ironic.go
dtantsur: In the ironic world, an instance name is rather mapped to an allocation name (I don't think we use allocation in BMO, but we might). The node name is something descriptive of the node itself (rack-7-node-32).

dhellmann: I don't know what an allocation name is, so we probably don't use it.

The BMO should always be able to look up a node by UUID, although we did consider other options for discovery. The name is used as a fallback, and as a debugging aid.

What if we combine the host CR UUID, namespace, and name? That would give us a unique value that couldn't be spoofed and we can use whatever punctuation separators are valid without concerns about collisions.

config/manager/manager.yaml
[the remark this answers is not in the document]
sadasu: Should Namespace passed in via POD_NAMESPACE be provided for BMO Leader election here: https://github.com/metal3-io/baremetal-operator/blob/master/main.go#L114?

pkg/provisioner/ironic/ironic.go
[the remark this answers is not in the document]
asalkeld: @dtantsur what character would you suggest?

pkg/provisioner/ironic/ironic.go
```diff
@@ -1784,6 +1790,14 @@ func (p *ironicProvisioner) softPowerOff() (result provisioner.Result, err error
 	return result, nil
 }
 
+func (p *ironicProvisioner) hostName() string {
+	if !p.watchAllNamespaces || p.host.Namespace == p.podNamespace {
+		return p.host.Name
+	}
+
+	return p.host.Namespace + "." + p.host.Name
```
maelk, line 1798: Would this break the behaviour during upgrade ? If the nodes are created with an old version of BMO, they only have their name, but then we might be looking for it with namespace + name after the upgrade.

pkg/provisioner/ironic/ironic.go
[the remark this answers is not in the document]
hardys: Note that we pass the BMH (not Ironic) name as metadata to the host:

https://github.com/metal3-io/baremetal-operator/blob/master/pkg/provisioner/ironic/ironic.go#L1399

So although we'll need to align with the ironic validation constraints, I don't think this name actually ever gets used as a hostname on the target OS?

pkg/provisioner/ironic/ironic.go
hardys: I think it depends if we expect to support upgrading from an environment without watchAllNamespaces to one where it's true - there was some discussion about that in the OpenShift case here:

https://github.com/openshift/cluster-baremetal-operator/pull/106#discussion_r576242689

@sadasu @dhellmann did we reach a conclusion on that discussion?

If we expect to support upgrading from a single-namespace to an all-namespaces configuration, I think we'll need to modify the conditional in ValidateManagementAccess where we rename the Ironic host only when `ironicNode.Name == ""` - we could rename whenever the current hostname doesn't match `p.hostName()` there instead?

config/manager/manager.yaml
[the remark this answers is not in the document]
asalkeld: @sadasu I don't think that has to change.

pkg/provisioner/ironic/ironic.go
asalkeld: ok, i'll make that change.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1784,6 +1790,14 @@ func (p *ironicProvisioner) softPowerOff() (result provisioner.Result, err error
 	return result, nil
 }
 
+func (p *ironicProvisioner) hostName() string {
```
andfasano, line 1793: I'd suggest to rename this method to something like ie `getIronicNodeNameFromHost(host metal3v1alpha1.BareMetalHost)` to make it more clear that this is mapping function to calculate the effective Ironic node name to be used given a BMH (since the current name sounds a little bit too vague for me)

asalkeld: I'll meet you half way, I still need p.watchAllNamespaces, so I'll this
```
func (p *ironicProvisioner) IronicNodeNameFromHost() string
```
also golang getters shouldn't start with "get" > https://golang.org/doc/effective_go#Getters

pkg/provisioner/ironic/ironic.go
```diff
@@ -192,15 +196,17 @@ func newProvisionerWithIronicClients(host metal3v1alpha1.BareMetalHost, bmcCreds
...
-	}
+		host:               host,
+		status:             &(host.Status.Provisioning),
+		bmcAccess:          bmcAccess,
+		bmcCreds:           bmcCreds,
+		client:             clientIronic,
+		inspector:          clientInspector,
+		publisher:          publisher,
+		watchAllNamespaces: os.Getenv("WATCH_NAMESPACE") == "",
+		podNamespace:       os.Getenv("POD_NAMESPACE"),
```
andfasano, line 207: Is it possible to pass these two params along with the other params in the factory method? Otherwise this strictly coupling could affect also the unit testing.

pkg/provisioner/ironic/ironic.go
andfasano: Mine was just an example, `IronicNodeNameFromHost` it's fine for me :+1:

sadasu (commented):
/lgtm

pkg/provisioner/ironic/ironic.go
```diff
@@ -1784,6 +1790,14 @@ func (p *ironicProvisioner) softPowerOff() (result provisioner.Result, err error
 	return result, nil
 }
 
+func (p *ironicProvisioner) ironicNodeNameFromHost() string {
+	if !p.watchAllNamespaces || p.host.Namespace == p.podNamespace {
```
sadasu, line 1794: If we are watching all namespaces and a BMH resource is created in pod namespace, will this return just the hostname (without the namespace prefix)? 
Even in that scenario, if I understand this feature correctly, we would want this method to return <namespace>.<hostname> whenever p.watchAllNamespaces.

sadasu (commented):
Do we have to consider the disable case, where BMO goes from watching all namespaces back to just watching 1 namespace? In that case would it be better for the method that generates the node name to always generate the name with the Namespace prefix?

pkg/provisioner/ironic/ironic.go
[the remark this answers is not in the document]
imain: I wonder about using a UUID for the hostname?  Sounds like we don't use it for lookup so it shouldn't affect that.  This would mess with debuggability a bit but could make this all a lot easier.

imain:
lgtm!

imain (approved):
Looks good to me

pkg/provisioner/ironic/ironic.go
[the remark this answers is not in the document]
hardys: > I wonder about using a GUID for the hostname? Sounds like we don't use it for lookup so it shouldn't effect that. This would mess with debuggability a bit but could make this all a lot easier.

Personally I think it would affect debuggability a lot, as a developer there's a lot of value in being able to interact with the Ironic API directly and be able to see at a glance which hosts align with specific BMH resources.

pkg/provisioner/ironic/ironic.go
asalkeld: so the point, as I understand it, is to prevent these names from clashing (because they are different clusters and you could have nodes with the same name). The idea is that if we go from watchAllNamespace to !watchAllNamespace and back at least the current cluster (it's bmh is in POD_NAMESPACE) would always be the same. I think @hardys suggested keeping the current namespaced names without the "." so that when we upgrade the names are consistent.

sadasu: When BMH resources are listed for all namespaces, the output would contain some with a namespace prefix and others without. Just making sure we are ok with it.

hardys: In the thread below I wasn't suggesting we keep the current names, but that we should rename existing ironic nodes in the case where the configuration has gone from single-namespace to watchAllNamespaces.

E.g here https://github.com/metal3-io/baremetal-operator/blob/master/pkg/provisioner/ironic/ironic.go#L493..L498 we'd change the name whenever it doesn't match the expected format.

It'll need some testing to ensure there aren't corner cases to consider, but that would resolve this concern I think?

Obviously the code here would also need to return the prefixed name whenever the watchAllNamespaces mode is enabled.

hardys: > The idea is that if we go from watchAllNamespace to !watchAllNamespace and back 

I was thinking that the only sane path was to switch from a single-namespace to `watchAllNamespaces=true` - that was the focus of my previous comment, I don't see how we can really support the reverse path, since there could be BMH resources in any namespace?

asalkeld: Thanks for the clarifcation @hardys , I'll fixup today.

pkg/provisioner/ironic/ironic.go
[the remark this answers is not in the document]
imain: Yeah good point.

pkg/provisioner/ironic/ironic.go
zaneb: The reason we look up by name is as a fallback if we fail to record the UUID of the node in the Status after creating it; it's necessary to have such a fallback to make the creation idempotent.
We do use the allocation ID, it's set to the UID of the Host. I don't think it's set immediately on creation though.

pkg/provisioner/ironic/ironic.go
```diff
@@ -192,22 +194,23 @@ func newProvisionerWithIronicClients(host metal3v1alpha1.BareMetalHost, bmcCreds
...
-		log:       log.WithValues("host", host.Name),
-		publisher: publisher,
+		host:               host,
+		status:             &(host.Status.Provisioning),
+		bmcAccess:          bmcAccess,
+		bmcCreds:           bmcCreds,
+		client:             clientIronic,
+		inspector:          clientInspector,
+		publisher:          publisher,
+		watchAllNamespaces: watchAllNamespaces,
```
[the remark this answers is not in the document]
asalkeld: sure, changing..

furkatgofurov7:
There was [an issue](https://kubernetes.slack.com/archives/CHD49TLE7/p1615895171009200) with our CI recently, so tests did not run properly in the previous runs. Hopefully should be fine now.

/test-integration

pkg/provisioner/ironic/ironic.go
```diff
@@ -303,18 +304,20 @@ func (p *ironicProvisioner) findExistingHost() (ironicNode *nodes.Node, err erro
...
-	case nil:
-		p.log.Info("found existing node by name")
-		return ironicNode, nil
-	case gophercloud.ErrDefault404:
-		p.log.Info(
-			fmt.Sprintf("node with name %s doesn't exist", p.host.Name))
-	default:
-		return nil, errors.Wrap(err,
-			fmt.Sprintf("failed to find node by name %s", p.host.Name))
+	for _, nodeName := range []string{p.ironicNodeNameFromHost(), p.host.Name} {
```
zaneb, line 307: Thinking a bit more about this, I think I was wrong and we don't reeeeally need the fallback in practice.

* If the Ironic DB is in ephemeral storage (almost everyone) then when you update all of the old nodes will be lost anyway. The only exception would be if you change the config but the Ironic DB is in a different Pod that doesn't get restarted.
* If it's in persistent storage, you won't run into this except for Hosts that were just created in the middle of switching from namespaced->clusterwide and which failed to write their ProvisionerID to the k8s API.

However, there is a real security issue here. If you have RBAC permissions to create a BareMetalHost, even in a namespace where it is not reconciled, then you could create a Host with the name `admin-namespace~admin-host` in advance of the switch from single-namespace to cluster-wide. Then you will end up with a BareMetalHost that has the same ProvisionerID as the one in the admin namespace, and thus with some luck and maybe timing you will be able to control it without having the credentials.

We could eliminate the security issue by not falling back in cases where the host name contains a `~`.

main.go
```diff
@@ -111,7 +111,7 @@ func main() {
 		Port:                    0, // Add flag with default of 9443 when adding webhooks
 		LeaderElection:          enableLeaderElection,
 		LeaderElectionID:        "baremetal-operator",
-		LeaderElectionNamespace: watchNamespace,
+		LeaderElectionNamespace: os.Getenv("POD_NAMESPACE"),
```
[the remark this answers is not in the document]
asalkeld: sure, did that

pkg/provisioner/ironic/ironic.go
asalkeld: done

pkg/provisioner/ironic/ironic.go
```diff
@@ -303,18 +304,25 @@ func (p *ironicProvisioner) findExistingHost() (ironicNode *nodes.Node, err erro
...
-		p.log.Info("found existing node by name")
-		return ironicNode, nil
-	case gophercloud.ErrDefault404:
-		p.log.Info(
-			fmt.Sprintf("node with name %s doesn't exist", p.host.Name))
-	default:
-		return nil, errors.Wrap(err,
-			fmt.Sprintf("failed to find node by name %s", p.host.Name))
+	nodeSearchList := []string{p.ironicNodeNameFromHost()}
+	if !strings.Contains(p.host.Name, "~") {
```
zaneb, line 308: nit: we should really use the constant nameSeparator here.

zaneb (commented):
One tiny comment.
/approve

pkg/provisioner/ironic/ironic.go
asalkeld: doh, fixed

zaneb:
This has been around for a while and has had quite a few eyes on it.
/lgtm

asalkeld:
needed a unit test fix after @honza's PR got merged.

PR #800, opened 2021-02-23
Development: account for unset GOPATH

dtantsur:
This variable has a well known default, just use it.

### DIFF
--- a/hack/install-controller-gen.sh
+++ b/hack/install-controller-gen.sh
@@ -2,6 +2,10 @@
 
 set -ex
 
+if [ -z "${GOPATH:-}" ]; then
+    eval "$(go env | grep GOPATH)"
+fi
+
 OUTPUT=bin/controller-gen
 
 # Check for a vendor directory if any downstream forks use that dependency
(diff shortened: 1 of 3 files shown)


fmuyassarov (commented):
looks good, but maybe for consistency we could have only , `eval "$(go env)"`, without the condition? like we do in other files under hack directory?, example [unit.sh](https://github.com/metal3-io/baremetal-operator/blob/dbcb214044624b6388bd78a072a426f59c51a6ed/hack/unit.sh#L10)

dtantsur:
Not sure. That one is executed in a container, so it has no chances of interfering with user's settings.

fmuyassarov:
> Not sure. That one is executed in a container, so it has no chances of interfering with user's settings.

Make sense.

elfosardo (approved):

maelk (approved):

hardys:
@dtantsur sorry but this now needs a rebase to make the gomod test pass

fmuyassarov:
> @dtantsur sorry but this now needs a rebase to make the gomod test pass

Ow true, it is actually due to bumping Go to 1.16. Sorry.

dtantsur:
Ah, oops, missed the comments above.

PR #799, opened 2021-02-19
Bump the Golang version to 1.16

fmuyassarov:
Bump the Golang version to [1.16](https://blog.golang.org/go1.16)

PR #803, opened 2021-03-01
Make BMH HardwareDetails fields optional

hardys:
Currently we force the status annotation to provide every
field, but when #796 lands it may be preferable to allow
a subset, so it's possible to set only a subset of the
hardware status fields when inspection is disabled.

### DIFF
--- a/apis/metal3.io/v1alpha1/baremetalhost_types.go
+++ b/apis/metal3.io/v1alpha1/baremetalhost_types.go
@@ -355,24 +355,24 @@ const (
 
 // CPU describes one processor on the host.
 type CPU struct {
-	Arch           string     `json:"arch"`
-	Model          string     `json:"model"`
-	ClockMegahertz ClockSpeed `json:"clockMegahertz"`
-	Flags          []string   `json:"flags"`
-	Count          int        `json:"count"`
+	Arch           string     `json:"arch,omitempty"`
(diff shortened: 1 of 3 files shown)


dhellmann:
I'm not sure what's up with the gomod job. Maybe this needs a rebase?

hardys:
> I'm not sure what's up with the gomod job. Maybe this needs a rebase?

Hmm, not sure either tbh but my local repo was a little out of date, now rebased so hopefully that fixes it :)

hardys:
The job says `Please run 'go mod tidy' and commit the changes to go.mod & go.sum.` but running locally with go1.15.8 doesn't change those files.  @honza perhaps you know what's needed here? :)

dhellmann:
> The job says `Please run 'go mod tidy' and commit the changes to go.mod & go.sum.` but running locally with go1.15.8 doesn't change those files. @honza perhaps you know what's needed here? :)

Is the CI job using go 1.16? I saw some patches related to updating, but I didn't keep track of whether they had landed.

hardys:
Looks like we may need to land #799 since https://github.com/metal3-io/project-infra/pull/172 landed and bumped to go 1.16?

asalkeld:
looks good to me..

romfreiman:
@hardys how console ui will look like in case of partial info?

PR #807, opened 2021-03-03
Use PROVISIONING_LIMIT also for deprovisioning.

s3rj1k:

### DIFF
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
 
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
@@ -338,9 +339,16 @@ func recordActionFailure(info *reconcileInfo, errorType metal3v1alpha1.ErrorType
 	return actionFailed{dirty: true, ErrorType: errorType, errorCount: info.host.Status.ErrorCount}
 }
 
-func recordActionDelayed(info *reconcileInfo) actionResult {
+func recordActionDelayed(info *reconcileInfo, state metal3v1alpha1.ProvisioningState) actionResult {
+	var counter prometheus.Counter
 
-	counter := delayedProvisioningHostCounters.With(hostMetricLabels(info.request))
+	switch state {
+	case metal3v1alpha1.StateDeprovisioning, metal3v1alpha1.StateDeleting:
+		counter = delayedDeprovisioningHostCounters.With(hostMetricLabels(info.request))
+	default:
+		counter = delayedProvisioningHostCounters.With(hostMetricLabels(info.request))
+	}
+
 	info.postSaveCallbacks = append(info.postSaveCallbacks, counter.Inc)
 
 	info.host.SetOperationalStatus(metal3v1alpha1.OperationalStatusDelayed)
--- a/controllers/metal3.io/host_state_machine.go
+++ b/controllers/metal3.io/host_state_machine.go
@@ -79,13 +79,14 @@ func recordStateEnd(info *reconcileInfo, host *metal3v1alpha1.BareMetalHost, sta
 	return
 }
 
-func (hsm *hostStateMachine) ensureProvisioningCapacity(info *reconcileInfo) actionResult {
-	hasCapacity, err := hsm.Provisioner.HasProvisioningCapacity()
+func (hsm *hostStateMachine) ensureCapacity(info *reconcileInfo, state metal3v1alpha1.ProvisioningState) actionResult {
+	hasCapacity, err := hsm.Provisioner.HasCapacity()
 	if err != nil {
-		return actionError{errors.Wrap(err, "failed to get hosts currently being provisioned")}
+		return actionError{errors.Wrap(err, "failed to determine current provisioner capacity")}
 	}
+
 	if !hasCapacity {
-		return recordActionDelayed(info)
+		return recordActionDelayed(info, state)
 	}
 
 	return nil
@@ -96,12 +97,13 @@ func (hsm *hostStateMachine) updateHostStateFrom(initialState metal3v1alpha1.Pro
 	if hsm.NextState != initialState {
 
 		// Check if there is a free slot available when trying to
-		// provision an host - if not the action will be delayed.
-		// The check is limited to only the provisioning states to
+		// (de)provision an host - if not the action will be delayed.
+		// The check is limited to only the (de)provisioning states to
 		// avoid putting an excessive pressure on the provisioner
 		switch hsm.NextState {
-		case metal3v1alpha1.StateInspecting, metal3v1alpha1.StateProvisioning:
-			if actionRes := hsm.ensureProvisioningCapacity(info); actionRes != nil {
+		case metal3v1alpha1.StateInspecting, metal3v1alpha1.StateProvisioning,
+			metal3v1alpha1.StateDeprovisioning, metal3v1alpha1.StateDeleting:
+			if actionRes := hsm.ensureCapacity(info, hsm.NextState); actionRes != nil {
 				return actionRes
 			}
 		}
@@ -144,7 +146,7 @@ func (hsm *hostStateMachine) checkDelayedHost(info *reconcileInfo) actionResult 
 
 	// Check if there's a free slot for hosts that have been previously delayed
 	if info.host.Status.OperationalStatus == metal3v1alpha1.OperationalStatusDelayed {
-		if actionRes := hsm.ensureProvisioningCapacity(info); actionRes != nil {
+		if actionRes := hsm.ensureCapacity(info, info.host.Status.Provisioning.State); actionRes != nil {
 			return actionRes
 		}
 
@@ -156,8 +158,9 @@ func (hsm *hostStateMachine) checkDelayedHost(info *reconcileInfo) actionResult 
 	// Make sure the check is re-applied when provisioning an
 	// host not yet tracked by the provisioner
 	switch info.host.Status.Provisioning.State {
-	case metal3v1alpha1.StateInspecting, metal3v1alpha1.StateProvisioning:
-		if actionRes := hsm.ensureProvisioningCapacity(info); actionRes != nil {
+	case metal3v1alpha1.StateInspecting, metal3v1alpha1.StateProvisioning,
+		metal3v1alpha1.StateDeprovisioning, metal3v1alpha1.StateDeleting:
+		if actionRes := hsm.ensureCapacity(info, info.host.Status.Provisioning.State); actionRes != nil {
 			return actionRes
 		}
 	}
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -41,7 +41,7 @@ var (
 	ironicInsecure            bool
 	ironicAuth                clients.AuthConfig
 	inspectorAuth             clients.AuthConfig
-	maxProvisioningHosts      int = 20
+	maxBusyHosts              int = 20
 
 	// Keep pointers to ironic and inspector clients configured with
 	// the global auth settings to reuse the connection between
@@ -125,7 +125,7 @@ func init() {
 			fmt.Fprintf(os.Stderr, "Cannot start: Invalid value set for variable PROVISIONING_LIMIT=%s", maxHostsStr)
 			os.Exit(1)
 		}
-		maxProvisioningHosts = value
+		maxBusyHosts = value
 	}
 }
 
@@ -1844,11 +1844,11 @@ func (p *ironicProvisioner) IsReady() (result bool, err error) {
 	return checker.IsReady()
 }
 
-func (p *ironicProvisioner) HasProvisioningCapacity() (result bool, err error) {
+func (p *ironicProvisioner) HasCapacity() (result bool, err error) {
 
-	hosts, err := p.loadProvisioningHosts()
+	hosts, err := p.loadBusyHosts()
 	if err != nil {
-		p.log.Error(err, "Unable to get hosts currently being provisioned")
+		p.log.Error(err, "Unable to get hosts for determining current provisioner capacity")
 		return false, err
 	}
 
@@ -1857,10 +1857,10 @@ func (p *ironicProvisioner) HasProvisioningCapacity() (result bool, err error) {
 		return true, nil
 	}
 
-	return len(hosts) < maxProvisioningHosts, nil
+	return len(hosts) < maxBusyHosts, nil
 }
 
-func (p *ironicProvisioner) loadProvisioningHosts() (hosts map[string]struct{}, err error) {
+func (p *ironicProvisioner) loadBusyHosts() (hosts map[string]struct{}, err error) {
 
 	hosts = make(map[string]struct{})
 	pager := nodes.List(p.client, nodes.ListOpts{
@@ -1885,7 +1885,8 @@ func (p *ironicProvisioner) loadProvisioningHosts() (hosts map[string]struct{}, 
 		switch nodes.ProvisionState(node.ProvisionState) {
 		case nodes.Cleaning, nodes.CleanWait,
 			nodes.Inspecting, nodes.InspectWait,
-			nodes.Deploying, nodes.DeployWait:
+			nodes.Deploying, nodes.DeployWait,
+			nodes.Deleting:
 			hosts[node.Name] = struct{}{}
 		}
 	}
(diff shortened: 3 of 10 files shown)


dhellmann:
I like the general idea of doing this, but I think Andrea is a better person to approve the implementation, since he's more familiar with how that part of the system works.

/ok-to-test
/unassign @dhellmann
/assign @andfasano

s3rj1k:
This adds limit for deleting/deprovisioning host to PROVISIONING_LIMIT.

Verified manually, works, needs work for unit tests.

@dhellmann Please look at this, if this is feasible solution I will fix unit tests in following commits, thanks.

s3rj1k:
@dhellmann Great! Thanks! That was super quick answer)

s3rj1k:
@andfasano Can you please look at this? Specifically unit-tests

andfasano:
> @andfasano Can you please look at this? Specifically unit-tests

@s3rj1k I'd like first to understand the motivation if possible, because this feature was originally linked to this proposal https://github.com/metal3-io/metal3-docs/blob/master/design/baremetal-operator/limit-hosts-provisioning.md, where we limited the scope just for the provisioning phase - whereas there was no interest in limiting the activity during the deprovisioning. Did you maybe have any specific scenario to be addressed?

s3rj1k:
> > @andfasano Can you please look at this? Specifically unit-tests
> 
> @s3rj1k I'd like first to understand the motivation if possible, because this feature was originally linked to this proposal https://github.com/metal3-io/metal3-docs/blob/master/design/baremetal-operator/limit-hosts-provisioning.md, where we limited the scope just for the provisioning phase - whereas there was no interest in limiting the activity during the deprovisioning. Did you maybe have any specific scenario to be addressed?

When deleting a big cluster (lot's of bmhs) I see similar issues (similar to provisioning lots of hosts simultaneously) with DHCP server during deprovisioning.

s3rj1k:
Also docs have `deleting` in pseudo-code example (https://github.com/metal3-io/metal3-docs/blob/master/design/baremetal-operator/limit-hosts-provisioning.md)
изображение

s3rj1k:
@andfasano ?

pkg/provisioner/ironic/ironic.go
```diff
@@ -1837,7 +1837,8 @@ func (p *ironicProvisioner) loadProvisioningHosts() (hosts map[string]struct{},
 		switch nodes.ProvisionState(node.ProvisionState) {
 		case nodes.Cleaning, nodes.CleanWait,
 			nodes.Inspecting, nodes.InspectWait,
-			nodes.Deploying, nodes.DeployWait:
+			nodes.Deploying, nodes.DeployWait,
+			nodes.Deleting:
```
dtantsur, line 1841: nit: `deleting` is a short state, the actual PXE booting happens in CleanWait.

dtantsur (approved):

s3rj1k:
@dtantsur fixed, thanks
Can we trigger `test-integration` on this PR?

s3rj1k:
@andfasano How about renaming Prometheus counter metal3_delayed__provisioning_total -> metal3_delayed__total similar to first points?

> As a last comment, probably such enhancement could be worth to have a related design proposal - or amend accordingly the existing one?

Better to amend existing one.

dhellmann:
> > @andfasano How about renaming Prometheus counter metal3_delayed__provisioning_total -> metal3_delayed__total similar to first points?
> 
> IMO it would make more sense to have two separate counters to track two different phases (provisioning and deprovisioning)

We should also consider whether we count the metrics as part of our API. If we do, then renaming something becomes non-trivial, and adding a new one is simpler.

s3rj1k:
In addition renaming metric makes PR trivial, adding newer metric would make PR a bit bigger

andfasano:
> Also as a note, @dtantsur pointed out that `nodes.Deleting` is a short phase and actual work is done in CleanWait [#807 (comment)](https://github.com/metal3-io/baremetal-operator/pull/807#discussion_r589530082)
> 
> So the question is, how reliable new deprovision metric would be?
> 
> cc: @dhellmann @andfasano

I think that original discussion focused on including all the `*Wait` and `*ing` states from the [Ironic states](https://docs.openstack.org/ironic/latest/_images/states.svg), so a possibility could be to reintroduce the missing ones.

s3rj1k:
So the main question that needs to be answered to know how we are going to handle this is:

> We should also consider whether we count the metrics as part of our API.

cc: @dhellmann @andfasano @dtantsur

andfasano:
> So the main question that needs to be answered to know how we are going to handle this is:
> 
> > We should also consider whether we count the metrics as part of our API.
> 
> cc: @dhellmann @andfasano @dtantsur

The metric is something that popped out during the development phase, but I think it could make sense to consider it part of the initial API.

s3rj1k:
@andfasano How about we rename metric to maintain simplicity of change in this PR and then update design docs to state that this (renamed) metric is part of API?

That way we could merge this change without refactoring nodes states that @dtantsur mentioned earlier.

cc: @dhellmann

dhellmann:
> @andfasano How about we rename metric to maintain simplicity of change in this PR and then update design docs to state that this (renamed) metric is part of API?
> 
> That way we could merge this change without refactoring nodes states that @dtantsur mentioned earlier.
> 
> cc: @dhellmann

I have customers in production with this version, and while I don't know if any of them look at the metric I do consider it to be part of the API. So, we need to maintain the existing metric as-is, and either add a new more generic one that counts for both cases or add a new one just for deprovisioning.

s3rj1k:
@andfasano Addressed your previous points and refactored this PR, please take a look.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -335,9 +336,16 @@ func recordActionFailure(info *reconcileInfo, errorType metal3v1alpha1.ErrorType
...
 }
 
-func recordActionDelayed(info *reconcileInfo) actionResult {
+func recordActionDelayed(info *reconcileInfo, nextState metal3v1alpha1.ProvisioningState) actionResult {
+	var counter prometheus.Counter
+
+	switch nextState {
+	case metal3v1alpha1.StateDeprovisioning, metal3v1alpha1.StateDeleting:
+		counter = delayedDeprovisioningHostCounters.With(hostMetricLabels(info.request))
+	default:
```
andfasano, line 345: I think I'd prefer to see explicitly here the cases for the provisioning counter, ie `metal3v1alpha1.StateInspecting, metal3v1alpha1.StateProvisioning`

controllers/metal3.io/host_state_machine.go
```diff
@@ -79,13 +79,14 @@ func recordStateEnd(info *reconcileInfo, host *metal3v1alpha1.BareMetalHost, sta
 	return
 }
 
-func (hsm *hostStateMachine) ensureProvisioningCapacity(info *reconcileInfo) actionResult {
-	hasCapacity, err := hsm.Provisioner.HasProvisioningCapacity()
+func (hsm *hostStateMachine) ensureCapacity(info *reconcileInfo) actionResult {
+	hasCapacity, err := hsm.Provisioner.HasCapacity()
 	if err != nil {
-		return actionError{errors.Wrap(err, "failed to get hosts currently being provisioned")}
+		return actionError{errors.Wrap(err, "failed to get hosts currently being (de)provisioned")}
```
andfasano, line 85: Probably now it's better a more generic message, ie something like `failed to determine current provisioner capacity"

controllers/metal3.io/host_state_machine.go
```diff
@@ -79,13 +79,14 @@ func recordStateEnd(info *reconcileInfo, host *metal3v1alpha1.BareMetalHost, sta
...
 	}
+
 	if !hasCapacity {
-		return recordActionDelayed(info)
+		return recordActionDelayed(info, hsm.NextState)
```
andfasano, line 89: This doesn't look fully correct. Only in one case the check is performed against `hsm.NextState`, while in two cases the `info.host.Status.Provisioning.State` it's used (and in one none at all).

controllers/metal3.io/host_state_machine.go
```diff
@@ -156,7 +158,12 @@ func (hsm *hostStateMachine) checkDelayedHost(info *reconcileInfo) actionResult
 	// host not yet tracked by the provisioner
 	switch info.host.Status.Provisioning.State {
 	case metal3v1alpha1.StateInspecting, metal3v1alpha1.StateProvisioning:
-		if actionRes := hsm.ensureProvisioningCapacity(info); actionRes != nil {
+		if actionRes := hsm.ensureCapacity(info); actionRes != nil {
+			return actionRes
+		}
+
+	case metal3v1alpha1.StateDeprovisioning, metal3v1alpha1.StateDeleting:
+		if actionRes := hsm.ensureCapacity(info); actionRes != nil {
```
andfasano, line 166: I guess the two cases could be merged

pkg/provisioner/ironic/ironic.go
```diff
@@ -1835,11 +1835,11 @@ func (p *ironicProvisioner) IsReady() (result bool, err error) {
...
 }
 
-func (p *ironicProvisioner) HasProvisioningCapacity() (result bool, err error) {
+func (p *ironicProvisioner) HasCapacity() (result bool, err error) {
 
-	hosts, err := p.loadProvisioningHosts()
+	hosts, err := p.loadBusyHosts()
 	if err != nil {
-		p.log.Error(err, "Unable to get hosts currently being provisioned")
+		p.log.Error(err, "Unable to get hosts currently being (de)provisioned")
```
andfasano, line 1842: Same message here

controllers/metal3.io/host_state_machine.go
s3rj1k: @andfasano I am not sure how you want to handle this.

Generally passing `hsm.NextState` to `recordActionDelayed()` is ok with you?

Also I did a bit for internal fixing (addressing also previous comment) in `recordActionDelayed()` so your point could be solved by rebased PR

s3rj1k:
@andfasano Rebased PR, please take a look.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -335,9 +336,20 @@ func recordActionFailure(info *reconcileInfo, errorType metal3v1alpha1.ErrorType
...
+
+	switch nextState {
+	case metal3v1alpha1.StateInspecting, metal3v1alpha1.StateProvisioning:
+		counter = delayedProvisioningHostCounters.With(hostMetricLabels(info.request))
+	case metal3v1alpha1.StateDeprovisioning, metal3v1alpha1.StateDeleting:
+		counter = delayedDeprovisioningHostCounters.With(hostMetricLabels(info.request))
+	case metal3v1alpha1.StateReady:
+		counter = delayedProvisioningHostCounters.With(hostMetricLabels(info.request))
+	default:
+		return actionDelayed{}
```
andfasano, line 350: This doesn't sound right to me. The operationalStatus must be set regardless of the provisioning state. I'd suggest just to keep a trace here

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -335,9 +336,20 @@ func recordActionFailure(info *reconcileInfo, errorType metal3v1alpha1.ErrorType
...
-func recordActionDelayed(info *reconcileInfo) actionResult {
+func recordActionDelayed(info *reconcileInfo, nextState metal3v1alpha1.ProvisioningState) actionResult {
+	var counter prometheus.Counter
+
+	switch nextState {
+	case metal3v1alpha1.StateInspecting, metal3v1alpha1.StateProvisioning:
+		counter = delayedProvisioningHostCounters.With(hostMetricLabels(info.request))
+	case metal3v1alpha1.StateDeprovisioning, metal3v1alpha1.StateDeleting:
+		counter = delayedDeprovisioningHostCounters.With(hostMetricLabels(info.request))
+	case metal3v1alpha1.StateReady:
```
andfasano, line 347: Assuming that the state parameter could be also the current one, this can be removed

controllers/metal3.io/host_state_machine.go
andfasano: Sure, let my clarify my point with some additional details. I think that in `checkDelayedHost` the current host provisioning state should be passed to `recordActionDelayed`, while in `updateHostStateFrom` it's ok to use `hsm.NextState`. Even though technically speaking `hsm.NextState` is equivalent to the `currentState` when evaluated in `checkDelayedHost`, I think that for the sake of clarity - and moreover to avoid future issues in case of refactoring - it's better to make it clear and explicit (so that the current temporal dependency will be removed)

controllers/metal3.io/baremetalhost_controller.go
s3rj1k: I am not following, you mean that for `default` case I still need to set `counter` ?
To what value then? How to choose is it a provisioning or deprovisioning counter?
Also I need to define counter because it gets incremented in few lines below, without setting it, nil pointer exception occurs.

controllers/metal3.io/baremetalhost_controller.go
s3rj1k: Similar questing about default value of counter.
https://github.com/metal3-io/baremetal-operator/pull/807#discussion_r601311963

controllers/metal3.io/host_state_machine.go
s3rj1k: So removing `hsm.NextState` and only using `info` to detect state?

> Even though technically speaking hsm.NextState is equivalent to the currentState when evaluated in checkDelayedHost

Hmm, this was not obvious to me, so I used `hsm.NextState`, removing then makes sense, also I would like to add a comment about it in code to make it more clearer for other contributors.

controllers/metal3.io/baremetalhost_controller.go
s3rj1k: @andfasano Your feedback is highly appreciated here, thanks.

andfasano: The `default` case returns immediately with `actionDelayed{}` skipping the operationalStatus. On a second though, I think a more conservative approach (in respect to the previous implementation) here could be preferrable

controllers/metal3.io/host_state_machine.go
andfasano: > So removing `hsm.NextState` and only using `info` to detect state?

Unfortunately I don't think it could be sufficient, you'll still need an extra parameter for the state in the `ensureCapacity` I guess

s3rj1k:
@andfasano Refactored based on your previous points, please take a look again, thanks.

controllers/metal3.io/host_state_machine.go
```diff
@@ -96,12 +97,13 @@ func (hsm *hostStateMachine) updateHostStateFrom(initialState metal3v1alpha1.Pro
...
-		// The check is limited to only the provisioning states to
+		// (de)provision an host - if not the action will be delayed.
+		// The check is limited to only the (de)provisioning states to
 		// avoid putting an excessive pressure on the provisioner
 		switch hsm.NextState {
-		case metal3v1alpha1.StateInspecting, metal3v1alpha1.StateProvisioning:
-			if actionRes := hsm.ensureProvisioningCapacity(info); actionRes != nil {
+		case metal3v1alpha1.StateInspecting, metal3v1alpha1.StateProvisioning,
+			metal3v1alpha1.StateDeprovisioning, metal3v1alpha1.StateDeleting:
+			if actionRes := hsm.ensureCapacity(info, hsm.NextState); actionRes != nil {
```
andfasano, line 106: Not a blocking point, but I was noticing that since this list is now getting bigger, it would be nice to remove the code duplication of this block with the one in[ checkDelayedHost](https://github.com/metal3-io/baremetal-operator/pull/807/files#diff-eeed3703d04e4c23a7d7af8cd0b7931b6b7990f23d826c49bdbc31c5f0a50291R160) - I guess this could also helps in passing the provisioning state

andfasano (approved):

controllers/metal3.io/host_state_machine.go
s3rj1k: @andfasano Thanks for approve, lets do code deduplication as a separate PR.

andfasano: Changes look fine to me, but I'd like to give @dhellmann and @dtantsur the opportunity for a final review

dhellmann: Don't block this work on me. I'm not likely to have time to come back and review it closely for a couple of weeks.

dtantsur: Same as Doug, unfortunately.

andfasano:
@s3rj1k looks like some conflicts needs to be resolved to proceed

s3rj1k:
@andfasano Fixed, all tests are green

andfasano:
@s3rj1k Can you please squash the commits?

s3rj1k:
@andfasano Rebased PR, all tests are green.

s3rj1k:
@andfasano Thanks!

PR #811, opened 2021-03-10
Add support for using pre-configured boot settings

rhjanders:
This change adds support for using Ironic noop management_interface in Metal3
by adding new BMC types: redfish-noop:// and ipmi-noop://.
This is a useful feature while running Metal3 on hardware with issues around
boot settings management, or in cases where the operator requires a very
specific boot configuration.

### DIFF
--- a/pkg/bmc/ipmi_noop.go
+++ b/pkg/bmc/ipmi_noop.go
@@ -1,0 +1,98 @@
+package bmc
+
+import (
+	"net/url"
+)
+
+func init() {
+	RegisterFactory("ipmi-noop", newIPMINoopAccessDetails, []string{})
+	RegisterFactory("libvirt-noop", newIPMINoopAccessDetails, []string{})
+}
+
+func newIPMINoopAccessDetails(parsedURL *url.URL, disableCertificateVerification bool) (AccessDetails, error) {
+	return &ipmiNoopAccessDetails{
+		bmcType:                        parsedURL.Scheme,
+		portNum:                        parsedURL.Port(),
+		hostname:                       parsedURL.Hostname(),
+		disableCertificateVerification: disableCertificateVerification,
+	}, nil
+}
+
+type ipmiNoopAccessDetails struct {
+	bmcType                        string
+	portNum                        string
+	hostname                       string
+	disableCertificateVerification bool
+}
+
+const ipmiDefaultPort = "623"
+
+func (a *ipmiNoopAccessDetails) Type() string {
+	return a.bmcType
+}
+
+// NeedsMAC returns true when the host is going to need a separate
+// port created rather than having it discovered.
+func (a *ipmiNoopAccessDetails) NeedsMAC() bool {
+	// libvirt-based hosts used for dev and testing require a MAC
+	// address, specified as part of the host, but we don't want the
+	// provisioner to have to know the rules about which drivers
+	// require what so we hide that detail inside this class and just
+	// let the provisioner know that "some" drivers require a MAC and
+	// it should ask.
+	return a.bmcType == "libvirt"
+}
+
+func (a *ipmiNoopAccessDetails) Driver() string {
+	return "ipmi"
+}
+
+func (a *ipmiNoopAccessDetails) DisableCertificateVerification() bool {
+	return a.disableCertificateVerification
+}
+
+// DriverInfo returns a data structure to pass as the DriverInfo
+// parameter when creating a node in Ironic. The structure is
+// pre-populated with the access information, and the caller is
+// expected to add any other information that might be needed (such as
+// the kernel and ramdisk locations).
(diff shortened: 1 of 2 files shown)


rhjanders:
Hi there. This is my first OpenShift PR - I appreciate your hints and guidance. Thanks!

hardys:
Hi @rhjanders - thanks for the PR, I'm not entirely sure I understand the use-case here from the description, can you elaborate?

It may be worth pushing a small proposal to https://github.com/metal3-io/metal3-docs/pulls which explains the use-case in more detail, which hardware this is useful for, and why we think it's justified to have a new top-level BMC type for this?

Also in https://github.com/metal3-io/baremetal-operator/blob/master/docs/api.md#bmc we'll need some docs to describe this new API.

I'm wondering if this is something we expect end-users to want, or a workaround for dev-testing on buggy hardware/firwmare versions?

rhjanders:
Thank you for your response @hardys . I will create a proposal document as you suggested and will aim to provide the information you requested there.

hardys:
Ok I found https://bugzilla.redhat.com/show_bug.cgi?id=1872656 which has some more context but I still think a metal3-docs PR would be helpful.

The bz conflates two issues AFAICS:
-  Boot mode switches from BIOS to UEFI - this can be controlled via the spec.bootMode filed added by @dhellmann via #602
- Host tries to PXE boot from the wrong NIC - as @derekhiggins mentions I think this is a manual step, and the reporter probably had a misconfigured host boot order?

I'm not opposed to this new interface, and I see @dhellmann mentioned something similar on the bz, but I think discussing the use-case in a metal3-docs PR may help us ensure we're adding a generally useful API vs a workaround for potential user error ;)

pkg/bmc/ipmi_noop.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+type ipmiNoopAccessDetails struct {
+	bmcType                        string
+	portNum                        string
+	hostname                       string
+	disableCertificateVerification bool
+}
+
+const ipmiDefaultPort = "623"
```
dtantsur, line 28: Can we move it to a common location? maybe somewhere under pkg/provisioner/ironic?

pkg/bmc/ipmi_noop.go
```diff
@@ -0,0 +1,98 @@
+package bmc
+
+import (
+	"net/url"
+)
+
+func init() {
+	RegisterFactory("ipmi-noop", newIPMINoopAccessDetails, []string{})
+	RegisterFactory("libvirt-noop", newIPMINoopAccessDetails, []string{})
```
dtantsur, line 9: I wonder if we need libvirt supports for that at all

pkg/bmc/redfish_noop.go
```diff
@@ -0,0 +1,112 @@
...
+
+import (
+	"net/url"
+	"strings"
+)
+
+func init() {
+	schemes := []string{"http", "https"}
+	RegisterFactory("redfish-noop", newRedfishNoopAccessDetails, schemes)
+	RegisterFactory("ilo5-redfish-noop", newRedfishNoopAccessDetails, schemes)
```
dtantsur, line 11: I'm not sure we need ilo5 support for it, ilo5 should work with boot mode (I hope).

pkg/bmc/redfish_noop.go
```diff
@@ -0,0 +1,112 @@
...
+
+func (a *redfishNoopAccessDetails) Driver() string {
+	return "redfish"
+}
+
+func (a *redfishNoopAccessDetails) DisableCertificateVerification() bool {
+	return a.disableCertificateVerification
+}
+
+func getRedfishNoopAddress(bmcType, host string) string {
```
dtantsur, line 56: Could we move this to a common location? It's the same for all redfish implementations.

dtantsur (commented):
A few minor things inline

dhellmann:
> Ok I found https://bugzilla.redhat.com/show_bug.cgi?id=1872656 which has some more context but I still think a metal3-docs PR would be helpful.
> 
> The bz conflates two issues AFAICS:
> 
> * Boot mode switches from BIOS to UEFI - this can be controlled via the spec.bootMode filed added by @dhellmann via #602
> * Host tries to PXE boot from the wrong NIC - as @derekhiggins mentions I think this is a manual step, and the reporter probably had a misconfigured host boot order?
> 
> I'm not opposed to this new interface, and I see @dhellmann mentioned something similar on the bz, but I think discussing the use-case in a metal3-docs PR may help us ensure we're adding a generally useful API vs a workaround for potential user error ;)

I agree, let's write down a clear description of the problem. This feels like an awful lot of code to pass a "noop" string via the management interface field in the API.

rhjanders:
I created https://github.com/metal3-io/metal3-docs/pull/170 with the design proposal as requested. I look forward to your feedback, we can move the discussion there.

PR #814, opened 2021-03-12
Add a new field to bmh.Status "ProvisionCount"

asalkeld:
This is to show how many times we’ve tried to provision a host.


controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -709,6 +709,7 @@ func (r *BareMetalHostReconciler) actionProvisioning(prov provisioner.Provisione
 		return actionContinue{}
 	}
 
+	info.host.Status.ProvisionCount++
```
andfasano, line 712: From the description, looks like the counter should be increased only in case of successfull provisioning (ie near `actionComplete{}`)

PR #816, opened 2021-03-16
Attempt to hard power off node before it is deleted

### LINKED ISSUE #410: Host should be powered down once deleted

sadasu:
When a BareMetalHost is deleted, power it off before performing
the delete operation.
Fixes #410

### DIFF
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -52,6 +52,10 @@ const (
 	rebootAnnotationPrefix        = "reboot.metal3.io"
 	inspectAnnotationPrefix       = "inspect.metal3.io"
 	hardwareDetailsAnnotation     = inspectAnnotationPrefix + "/hardwaredetails"
+
+	// Maximum number of times node Power off would be retried before
+	// it would be deleted.
+	maxPowerOffRetryCount = 3
 )
 
 // BareMetalHostReconciler reconciles a BareMetalHost object
@@ -483,7 +487,7 @@ func (r *BareMetalHostReconciler) actionDeleting(prov provisioner.Provisioner, i
 		return deleteComplete{}
 	}
 
-	provResult, err := prov.Delete()
+	provResult, err := prov.Delete(info.host.Status.ErrorCount <= maxPowerOffRetryCount)
 	if err != nil {
 		return actionError{errors.Wrap(err, "failed to delete")}
 	}
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -490,6 +490,9 @@ func (r *BareMetalHostReconciler) actionDeleting(prov provisioner.Provisioner, i
 	if provResult.Dirty {
 		return actionContinue{provResult.RequeueAfter}
 	}
+	if provResult.ErrorMessage != "" {
+		return recordActionFailure(info, metal3v1alpha1.PowerManagementError, provResult.ErrorMessage)
+	}
 
 	// Remove finalizer to allow deletion
 	info.host.Finalizers = utils.FilterStringFromList(
--- a/pkg/provisioner/ironic/delete_test.go
+++ b/pkg/provisioner/ironic/delete_test.go
@@ -143,6 +143,19 @@ func TestDelete(t *testing.T) {
 				Value: true,
 			},
 		},
+		{
+			name: "power-change-with-locked-host",
+			ironic: testserver.NewIronic(t).Node(
+				nodes.Node{
+					UUID:           nodeUUID,
+					ProvisionState: "active",
+					Maintenance:    true,
+					PowerState:     powerOn,
+				},
+			).WithNodeStatesPower(nodeUUID, http.StatusConflict).WithNodeStatesPowerUpdate(nodeUUID, http.StatusConflict),
+			expectedDirty:        false,
+			expectedRequestAfter: 0,
+		},
 	}
 
 	for _, tc := range cases {
@@ -205,7 +218,7 @@ func TestDeleteWithPowerOff(t *testing.T) {
 		expectedError               string
 	}{
 		{
-			name: "test-power-change-without-force",
+			name: "power-change-without-retry",
 			ironic: testserver.NewIronic(t).Node(
 				nodes.Node{
 					UUID:           nodeUUID,
@@ -219,7 +232,7 @@ func TestDeleteWithPowerOff(t *testing.T) {
 			expectedError:               "",
 		},
 		{
-			name: "test-power-change-with-force",
+			name: "power-change-with-retry",
 			ironic: testserver.NewIronic(t).Node(
 				nodes.Node{
 					UUID:           nodeUUID,
--- a/pkg/provisioner/ironic/delete_test.go
+++ b/pkg/provisioner/ironic/delete_test.go
@@ -172,7 +172,7 @@ func TestDelete(t *testing.T) {
 				t.Fatalf("could not create provisioner: %s", err)
 			}
 
-			result, err := prov.Delete()
+			result, err := prov.Delete(true)
 
 			assert.Equal(t, tc.expectedDirty, result.Dirty)
 			assert.Equal(t, tc.expectedRequestAfter, result.RequeueAfter)
@@ -196,44 +196,16 @@ func TestDeleteWithPowerOff(t *testing.T) {
 	nodeUUID := "33ce8659-7400-4c68-9535-d10766f07a58"
 
 	cases := []struct {
-		name        string
-		ironic      *testserver.IronicMock
-		inspector   *testserver.InspectorMock
-		priorErrors int
+		name      string
+		ironic    *testserver.IronicMock
+		inspector *testserver.InspectorMock
+		retryFlag bool
 
 		expectedPowerStateInRequest string
 		expectedError               string
 	}{
 		{
-			name: "test-power-change-without-errors",
-			ironic: testserver.NewIronic(t).Node(
-				nodes.Node{
-					UUID:           nodeUUID,
-					ProvisionState: "active",
-					Maintenance:    true,
-					PowerState:     powerOn,
-				},
-			),
-			priorErrors:                 0,
-			expectedPowerStateInRequest: "power off",
-			expectedError:               "",
-		},
-		{
-			name: "test-power-change-with-errors",
-			ironic: testserver.NewIronic(t).Node(
-				nodes.Node{
-					UUID:           nodeUUID,
-					ProvisionState: "active",
-					Maintenance:    true,
-					PowerState:     powerOn,
-				},
-			),
-			priorErrors:                 maxPowerOffRetryCount - 1,
-			expectedPowerStateInRequest: "power off",
-			expectedError:               "",
-		},
-		{
-			name: "test-power-change-with-max-errors",
+			name: "test-power-change-without-force",
 			ironic: testserver.NewIronic(t).Node(
 				nodes.Node{
 					UUID:           nodeUUID,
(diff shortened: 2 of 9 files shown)


pkg/provisioner/ironic/delete_test.go
```diff
@@ -190,3 +219,100 @@ func TestDelete(t *testing.T) {
...
+				nodes.Node{
+					UUID:           nodeUUID,
+					ProvisionState: "active",
+					Maintenance:    true,
+					PowerState:     powerOn,
+				},
+			).WithNodeStatesPowerUpdate(nodeUUID, http.StatusAccepted),
+			priorErrors:        0,
+			expectedPowerState: "power off",
+			expectedError:      "failed to remove host",
```
andfasano, line 248: Since this test does not focus on delete action itself, this kind of expectedError sounds a little bit strange. It's better to mock the call to Delete to make it going on without errors.

pkg/provisioner/ironic/delete_test.go
```diff
@@ -190,3 +219,100 @@ func TestDelete(t *testing.T) {
...
+			ironic: testserver.NewIronic(t).Node(
+				nodes.Node{
+					UUID:           nodeUUID,
+					ProvisionState: "active",
+					Maintenance:    true,
+					PowerState:     powerOn,
+				},
+			).WithNodeStatesPowerUpdate(nodeUUID, http.StatusAccepted),
+			priorErrors:        0,
+			expectedPowerState: "power off",
```
andfasano, line 247: This is more a `expectedPowerStateRequestChangeTo` (or something less verbose, the point here is that this is a request more than an expected result)

pkg/provisioner/ironic/delete_test.go
```diff
@@ -190,3 +219,100 @@ func TestDelete(t *testing.T) {
...
+					ProvisionState: "active",
+					Maintenance:    true,
+					PowerState:     powerOn,
+				},
+			),
+			priorErrors:        4,
+			expectedPowerState: "",
+			expectedError:      "failed to remove host",
+		},
+	}
```
andfasano, line 278: A couple of another cases are required to manage when `p.hardPowerOff()` returns an error

pkg/provisioner/ironic/delete_test.go
```diff
@@ -190,3 +219,100 @@ func TestDelete(t *testing.T) {
...
+				nodes.Node{
+					UUID:           nodeUUID,
+					ProvisionState: "active",
+					Maintenance:    true,
+					PowerState:     powerOn,
+				},
+			),
+			priorErrors:        4,
+			expectedPowerState: "",
+			expectedError:      "failed to remove host",
```
andfasano, line 276: See previous comment

pkg/provisioner/ironic/delete_test.go
```diff
@@ -190,3 +219,100 @@ func TestDelete(t *testing.T) {
...
+				nodes.Node{
+					UUID:           nodeUUID,
+					ProvisionState: "active",
+					Maintenance:    true,
+					PowerState:     powerOn,
+				},
+			),
+			priorErrors:        2,
+			expectedPowerState: "power off",
+			expectedError:      "",
```
andfasano, line 262: See previous comment

pkg/provisioner/ironic/delete_test.go
```diff
@@ -190,3 +219,100 @@ func TestDelete(t *testing.T) {
...
+			name: "test-power-change-with-errors",
+			ironic: testserver.NewIronic(t).Node(
+				nodes.Node{
+					UUID:           nodeUUID,
+					ProvisionState: "active",
+					Maintenance:    true,
+					PowerState:     powerOn,
+				},
+			),
+			priorErrors:        2,
```
andfasano, line 260: nit: `maxPowerOffRetryCount-1`

pkg/provisioner/ironic/delete_test.go
```diff
@@ -190,3 +219,100 @@ func TestDelete(t *testing.T) {
...
+			name: "test-power-change-with-max-errors",
+			ironic: testserver.NewIronic(t).Node(
+				nodes.Node{
+					UUID:           nodeUUID,
+					ProvisionState: "active",
+					Maintenance:    true,
+					PowerState:     powerOn,
+				},
+			),
+			priorErrors:        4,
```
andfasano, line 274: nit: `maxPowerOffRetryCount+1`

pkg/provisioner/ironic/delete_test.go
```diff
@@ -171,6 +193,13 @@ func TestDelete(t *testing.T) {
 			if err != nil {
 				t.Fatalf("could not create provisioner: %s", err)
 			}
+			if tc.powerOn {
```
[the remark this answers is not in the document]
sadasu: Agreed.

pkg/provisioner/ironic/delete_test.go
sadasu: Done.

pkg/provisioner/ironic/delete_test.go
sadasu: Done.

pkg/provisioner/ironic/delete_test.go
sadasu: Done.

pkg/provisioner/ironic/delete_test.go
sadasu: I wasn't able to control the return value of hardPowerOff() using  WithNodeStatesPowerUpdate() in https://github.com/metal3-io/baremetal-operator/blob/master/pkg/provisioner/ironic/testserver/ironic.go#L227 .

Any ideas. Only the mock for Delete() seems to be affecting the outcome.

sadasu: Using https://github.com/metal3-io/baremetal-operator/blob/master/pkg/provisioner/ironic/testserver/ironic.go#L227, I wasn't able to mock the out come for hardPowerOff() which is called from within the Delete().

andfasano: That's a real interesting case and it's worth some additional explanation.

Adding the mocking function to test case should be ok, ie: 
```
ironic: testserver.NewIronic(t).Node(
				nodes.Node{
					UUID:           nodeUUID,
					ProvisionState: "active",
					Maintenance:    true,
					PowerState:     powerOn,
				},
			).WithNodeStatesPowerUpdate(nodeUUID, http.StatusConflict).Delete(nodeUUID),
```
In this case the expected error is returned, anyhow this type assertion https://github.com/sadasu/baremetal-operator/blob/4d0f31405af55c03516533cd6e624aa6c32afaa2/pkg/provisioner/ironic/ironic.go#L1624  is going to fail because the real error is wrapped (twice).

As per the golang recommended best practices on errors (https://blog.golang.org/go1.13-errors), type check must be changed using the new errors functions like: 

```
var hostErr *HostLockedError
if errors.As(err, &hostErr) {
  p.log.Info("could not power off host, busy")
  return retryAfterDelay(powerRequeueDelay)
} else {
  return operationFailed("failed to power off host")
}
```

But that's not _yet_ sufficient, because the underlying `changePower` does not return a pointer receiver for the error: https://github.com/sadasu/baremetal-operator/blob/4d0f31405af55c03516533cd6e624aa6c32afaa2/pkg/provisioner/ironic/ironic.go#L1681

So that code as well must be changed to:

```
return result, &HostLockedError{Address: p.host.Spec.BMC.Address}
```
This means of course that other points in the code (ie `PowerOn`) needs to be reviewed for properly managing the error

pkg/provisioner/ironic/delete_test.go
```diff
@@ -143,6 +144,27 @@ func TestDelete(t *testing.T) {
 				Value: true,
 			},
 		},
+		{
+			name: "test-power-on",
```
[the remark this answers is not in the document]
sadasu: All power related tests are now being handled in TestDeleteWithPowerOff() so removing this test case.

zaneb (commented):
There's one other problem we'll have to address, which is that hardPowerOff() doesn't set (and Delete doesn't check for) an ErrorMessage when there is a LastError from ironic. So if all the ironic calls succeed but ironic cannot actually change the power state, then we will keep retrying forever.

Fun fact: this also means that we currently never report a power management error. I raised #828 to record this. Basically I don't think we will be able to complete this until that issue is fixed.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1613,6 +1617,20 @@ func (p *ironicProvisioner) Delete() (result provisioner.Result, err error) {
...
+	if ironicNode.PowerState == powerOn && p.host.Status.ErrorCount <= maxPowerOffRetryCount {
+		p.log.Info("host ready to be powered off")
+		_, err := p.hardPowerOff()
+		if err != nil {
+			switch err.(type) {
+			case HostLockedError:
+				p.log.Info("could not power off host, busy")
+				return retryAfterDelay(powerRequeueDelay)
+			default:
+				return operationFailed("failed to power off host")
```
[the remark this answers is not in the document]
sadasu: Done in https://github.com/metal3-io/baremetal-operator/pull/816/commits/2649e51141d3c3d493452eb393eea8f44ab4f775

pkg/provisioner/ironic/ironic.go
```diff
@@ -1613,6 +1617,20 @@ func (p *ironicProvisioner) Delete() (result provisioner.Result, err error) {
 		return p.setMaintenanceFlag(ironicNode, true)
 	}
 
+	if ironicNode.PowerState == powerOn && p.host.Status.ErrorCount <= maxPowerOffRetryCount {
```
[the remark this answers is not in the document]
sadasu: Done in https://github.com/metal3-io/baremetal-operator/pull/816/commits/2649e51141d3c3d493452eb393eea8f44ab4f775

pkg/provisioner/ironic/ironic.go
```diff
@@ -1612,7 +1612,7 @@ func (p *ironicProvisioner) Deprovision(force bool) (result provisioner.Result,
 // Delete removes the host from the provisioning system. It may be
 // called multiple times, and should return true for its dirty flag
 // until the deprovisioning operation is completed.
-func (p *ironicProvisioner) Delete() (result provisioner.Result, err error) {
+func (p *ironicProvisioner) Delete(force bool) (result provisioner.Result, err error) {
```
andfasano, line 1615: That's just a `forceHardPowerOff` I guess, otherwise the meaning looks too much generic in respect of the current implementation

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -483,13 +487,16 @@ func (r *BareMetalHostReconciler) actionDeleting(prov provisioner.Provisioner, i
 		return deleteComplete{}
 	}
 
-	provResult, err := prov.Delete()
+	provResult, err := prov.Delete(info.host.Status.ErrorCount > maxPowerOffRetryCount)
```
andfasano, line 490: It seems the opposite, as per the current implementation?

sadasu:
The baremetal host controller is checking for the host's error count before performing some but not all actions. A host's error count should inform whether the controller would perform *any* action on the host. Does the controller have a higher error threshold for the overall number of errors and not call *any* action on that host?  And should this checked before the action is attempted and not within the logic for each action?

pkg/provisioner/ironic/delete_test.go
sadasu: I have attempt to do this in https://github.com/metal3-io/baremetal-operator/pull/816/commits/c941bc0ddd6324299b68272855578b1b38c879ba. This commit might need more work.

pkg/provisioner/ironic/delete_test.go
```diff
@@ -143,6 +143,19 @@ func TestDelete(t *testing.T) {
...
+			name: "power-change-with-locked-host",
+			ironic: testserver.NewIronic(t).Node(
+				nodes.Node{
+					UUID:           nodeUUID,
+					ProvisionState: "active",
+					Maintenance:    true,
+					PowerState:     powerOn,
+				},
+			).WithNodeStatesPower(nodeUUID, http.StatusConflict).WithNodeStatesPowerUpdate(nodeUUID, http.StatusConflict),
+			expectedDirty:        false,
```
sadasu, line 156: These expected values are consistent with what is returned by transientError(). http.StatusConflict is not resulting in HostLockedError.

maelk:
Hello! What is the status with this PR ? Is it still blocked ?

sadasu:
> Hello! What is the status with this PR ? Is it still blocked ?

@maelk During the discussion in this PR it was determined that https://github.com/metal3-io/baremetal-operator/pull/841 needs to be solved first. I am working on that right now. Comments welcome.

andfasano:
@sadasu looks like the PR needs a rebase

furkatgofurov7:
@sadasu hi! Would you have time to re-open this PR and fix the conflicts? This should be stlll valid to fix the #410

PR #817, opened 2021-03-16
Allocate Ironic port when node exists and has no port allocated.

s3rj1k:
This PR fixes issues below:

 - While node is created in Ironic DB and some misfortune happens to Ironic pod or DB pod, Ironic does not allocate port created node, this in turn breaks inspection with Ironic erroring on absence of allocated port, operator does not handle this case and stuck in erroring reconcile loop until back-off timer is triggered.
 
When node has no allocated port to it, Ironic returns this error
```
No lookup attributes were found, inspector won't be able to find it after introspection, consider creating ironic ports or providing an IPMI address
```

_IPMI address part is just wrong, BMC creds are checked in Registering state and are validated, also they where validated manually just in case._

After checking sources that return this error in: https://github.com/openstack/ironic-inspector/blob/stable/victoria/ironic_inspector/introspect.py#L102-L119
We can clearly see that this error is indeed because of not allocated port to node (no MAC), this also is verified by this unix pipe magic for specified host.

```
kubectl get bmh worker-1 -o jsonpath='{.spec.bootMACAddress}' | xargs -rI{} --verbose openstack baremetal port show --address '{}' -f value --fields node_uuid | xargs -rI{} --verbose openstack baremetal node show {} -f value --fields name
```

### DIFF
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -287,10 +287,6 @@ func (p *ironicProvisioner) listAllPorts(address string) ([]ports.Port, error) {
 
 	pager := ports.List(p.client, opts)
 
-	if pager.Err != nil {
-		return allPorts, pager.Err
-	}
-
 	allPages, err := pager.AllPages()
 
 	if err != nil {
@@ -329,9 +325,6 @@ func (p *ironicProvisioner) nodeHasAssignedPort(ironicNode *nodes.Node) (bool, e
 	}
 
 	pager := ports.List(p.client, opts)
-	if pager.Err != nil {
-		return false, errors.Wrap(pager.Err, "failed to list ports")
-	}
 
 	allPages, err := pager.AllPages()
 	if err != nil {
@@ -1696,9 +1689,6 @@ func (p *ironicProvisioner) loadBusyHosts() (hosts map[string]struct{}, err erro
 	pager := nodes.List(p.client, nodes.ListOpts{
 		Fields: []string{"uuid,name,provision_state,driver_internal_info,target_provision_state"},
 	})
-	if pager.Err != nil {
-		return nil, pager.Err
-	}
 
 	page, err := pager.AllPages()
 	if err != nil {
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -321,6 +321,53 @@ func (p *ironicProvisioner) getNode() (*nodes.Node, error) {
 	}
 }
 
+// Verifies that node has port assigned by Ironic.
+func (p *ironicProvisioner) nodeHasAssignedPort(ironicNode *nodes.Node) (bool, error) {
+	opts := ports.ListOpts{
+		Fields:   []string{"node_uuid"},
+		NodeUUID: ironicNode.UUID,
+	}
+
+	pager := ports.List(p.client, opts)
+	if pager.Err != nil {
+		return false, errors.Wrap(pager.Err, "failed to list ports")
+	}
+
+	allPages, err := pager.AllPages()
+	if err != nil {
+		return false, errors.Wrap(err, "failed to page over list of ports")
+	}
+
+	empty, err := allPages.IsEmpty()
+	if err != nil {
+		return false, errors.Wrap(err, "failed to check port list status")
+	}
+
+	if empty {
+		p.debugLog.Info("node has no assigned port")
+		return false, nil
+	}
+
+	p.debugLog.Info("node has assigned port")
+	return true, nil
+}
+
+// Verify that MAC is already allocated to some node port.
+func (p *ironicProvisioner) isAddressAllocatedToPort(address string) (bool, error) {
+	allPorts, err := p.listAllPorts(address)
+	if err != nil {
+		return false, errors.Wrap(err, fmt.Sprintf("failed to list ports for %s", address))
+	}
+
+	if len(allPorts) == 0 {
+		p.debugLog.Info("address does not have allocated ports", "address", address)
+		return false, nil
+	}
+
+	p.debugLog.Info("address is allocated to port", "address", address)
+	return true, nil
+}
+
 // Look for an existing registration for the host in Ironic.
 func (p *ironicProvisioner) findExistingHost(bootMACAddress string) (ironicNode *nodes.Node, err error) {
 	// Try to load the node by UUID
@@ -387,6 +434,25 @@ func (p *ironicProvisioner) findExistingHost(bootMACAddress string) (ironicNode 
 	// Either the node was never created or the Ironic database has
 	// been dropped.
 	return nil, nil
+}
+
+func (p *ironicProvisioner) createPXEEnabledNodePort(uuid, macAddress string) error {
+	p.log.Info("creating PXE enabled ironic port for node", "NodeUUID", uuid, "MAC", macAddress)
+
+	enable := true
+
+	_, err := ports.Create(
+		p.client,
+		ports.CreateOpts{
+			NodeUUID:   uuid,
+			Address:    macAddress,
+			PXEEnabled: &enable,
+		}).Extract()
+	if err != nil {
+		return errors.Wrap(err, fmt.Sprintf("failed to create ironic port for node: %s, MAC: %s", uuid, macAddress))
+	}
+
+	return nil
 }
 
 // ValidateManagementAccess registers the host with the provisioning
@@ -475,18 +541,9 @@ func (p *ironicProvisioner) ValidateManagementAccess(data provisioner.Management
 		// If we know the MAC, create a port. Otherwise we will have
 		// to do this after we run the introspection step.
 		if p.bootMACAddress != "" {
-			enable := true
-			p.log.Info("creating port for node in ironic", "MAC",
-				p.bootMACAddress)
-			_, err = ports.Create(
-				p.client,
-				ports.CreateOpts{
-					NodeUUID:   ironicNode.UUID,
-					Address:    p.bootMACAddress,
-					PXEEnabled: &enable,
-				}).Extract()
+			err = p.createPXEEnabledNodePort(ironicNode.UUID, p.bootMACAddress)
 			if err != nil {
-				result, err = transientError(errors.Wrap(err, "failed to create port in ironic"))
+				result, err = transientError(err)
 				return
 			}
 		}
@@ -498,6 +555,34 @@ func (p *ironicProvisioner) ValidateManagementAccess(data provisioner.Management
 		provID = ironicNode.UUID
 
 		updater.SetTopLevelOpt("name", ironicNodeName(p.objectMeta), ironicNode.Name)
+
+		// When node exists but has no assigned port to it by Ironic and actuall address (MAC) is present
+		// in host config and is not allocated to different node lets try to create port for this node.
+		if p.bootMACAddress != "" {
+			var nodeHasAssignedPort, addressIsAllocatedToPort bool
+
+			nodeHasAssignedPort, err = p.nodeHasAssignedPort(ironicNode)
+			if err != nil {
+				result, err = transientError(err)
+				return
+			}
+
+			if !nodeHasAssignedPort {
+				addressIsAllocatedToPort, err = p.isAddressAllocatedToPort(p.bootMACAddress)
+				if err != nil {
+					result, err = transientError(err)
+					return
+				}
+
+				if !addressIsAllocatedToPort {
+					err = p.createPXEEnabledNodePort(ironicNode.UUID, p.bootMACAddress)
+					if err != nil {
+						result, err = transientError(err)
+						return
+					}
+				}
+			}
+		}
 
 		// Look for the case where we previously enrolled this node
 		// and now the credentials have changed.


s3rj1k:
cc: @dhellmann Can you please also look at this, thanks.

dtantsur:
I was under impression that we always create a port before inspection, otherwise e.g. Redfish inspection would not work.

s3rj1k:
@dtantsur We allocate port only when node was not created, for created node we expect that port was created.
This PR fixes this exact edge case when node was created but for some unknown reason port was not allocated for that node.

s3rj1k:
@dtantsur All tests are green :)

s3rj1k:
@stbenjam @dhellmann @dtantsur Are we safe to merge?

zaneb:
@s3rj1k can you describe what the actual change is in the first patch? Git makes the diff really hard to interpret but after going through it as closely as I can using a visual diff tool, I was unable it find any change in logic. It looks like a straight refactor (although arguably an improvement in readability). The commit message only says that something was broken and is now fixed but I can't figure out what. What's the actual code path that is different?

s3rj1k:
@zaneb Yea, its a diff mess

So basically I am introducing `var introspectionStatusNotFound bool` flag and refactoring an `if` `case` to make it more readable

doing this 

```
	status, err := introspection.GetIntrospectionStatus(p.inspector, ironicNode.UUID).Extract()
	if err != nil {
		if _, isNotFound := err.(gophercloud.ErrDefault404); isNotFound {
			introspectionStatusNotFound = true
			p.log.Info("inspection result not found")
		} else {
			result, err = transientError(errors.Wrap(err, "failed to extract hardware inspection status"))
			return
		}
	}

	if introspectionStatusNotFound { ... }
``` 

Old case code is mostly unchanged in { ... }, this fixes an issue with forced introspection, before, because of complicated flow logic it was hard to spot.

s3rj1k:
@zaneb Did you get a chance to look at this again?

zaneb:
I'm not following how that code fails to get executed:
If `GetIntrospectionStatus()` on line 653 returns a 404 Not Found then
`err` will be non-`nil` on line 654
`isNotFound` will be true on line 655
If the node's provision state is InspectFail, we'll hit the default case of the switch on line 661
The block starting at line 662 will execute or not depending on whether the `force` flag is set.

AFAICT the logic is the same after the patch, only the line numbers change:
If `GetIntrospectionStatus()` on line 655 returns a 404 Not Found then
`isNotFound` will be true on line 657
`introspectionStatusNotFound` will be true on line 666
If the node's provision state is InspectFail, we'll hit the default case of the switch on line 672
The block starting at line 673 will execute or not depending on whether the `force` flag is set.

There's no difference.

On closer inspection (no pun intended) there _does_ appear to be a bug here though: I'm pretty sure I meant to put a return statement after `err = nil` on line 669. There'd be no point setting it to nil when we're going to fall through and overwrite it anyway, so that seems like a mistake on my part. Your patch doesn't change that though.

s3rj1k:
@zaneb, so I've removed first part of PR, currently it only deals with Ironic port creation, hoping this would help in getting this merged.

s3rj1k:
@dtantsur Can you please take a look at this PR, hopefully approve it :)

zaneb:
> I'm pretty sure I meant to put a return statement after `err = nil` on line 669.

I opened #840 to fix this.

> So that do you think? Leaving only a second part of a patch sounds good?

Sounds good to me, but this leaves only the part I don't understand well enough to review.

s3rj1k:
@zaneb Are there someone else beside Dmitry who can review openstack part of PR?

Dmitry says his busy with other work and can't help currently.

pkg/provisioner/ironic/ironic.go
```diff
@@ -510,6 +556,25 @@ func (p *ironicProvisioner) ValidateManagementAccess(data provisioner.Management
...
+				ports.CreateOpts{
+					NodeUUID:   ironicNode.UUID,
+					Address:    p.bootMACAddress,
+					PXEEnabled: &enable,
+				}).Extract()
+			if err != nil {
+				result, err = transientError(errors.Wrap(err, "failed to create port in ironic"))
+				return
+			}
+		}
```
zaneb, line 576: Since this whole block is copied and pasted from the one at line 525, can we split it out into a private method? Or delete that one and move this after the if/else.

pkg/provisioner/ironic/ironic.go
```diff
@@ -321,6 +321,52 @@ func (p *ironicProvisioner) getNode() (*nodes.Node, error) {
...
+func (p *ironicProvisioner) nodeHasAssignedPort(ironicNode *nodes.Node) bool {
+	opts := ports.ListOpts{
+		Fields: []string{"node_uuid"},
+	}
+
+	opts.NodeUUID = ironicNode.UUID
+
+	pager := ports.List(p.client, opts)
+	if pager.Err != nil {
+		p.log.Info("node has no assigned port", "error", pager.Err)
```
zaneb, line 334: Let's use the debugLog for these.

pkg/provisioner/ironic/ironic.go
```diff
@@ -321,6 +321,52 @@ func (p *ironicProvisioner) getNode() (*nodes.Node, error) {
...
+		return false
+	}
+
+	allPages, err := pager.AllPages()
+	if err != nil {
+		p.log.Info("node has no assigned port", "error", err)
+		return false
```
zaneb, line 341: Is this the right way to handle errors? I'd have thought we'd want to return the error and retry.

pkg/provisioner/ironic/ironic.go
```diff
@@ -321,6 +321,52 @@ func (p *ironicProvisioner) getNode() (*nodes.Node, error) {
 	}
 }
 
+// Verifies that node has port assigned by Ironic.
+func (p *ironicProvisioner) nodeHasAssignedPort(ironicNode *nodes.Node) bool {
+	opts := ports.ListOpts{
+		Fields: []string{"node_uuid"},
+	}
+
+	opts.NodeUUID = ironicNode.UUID
```
zaneb, line 330: nit: why is this not being set in the initialiser above?

pkg/provisioner/ironic/ironic.go
```diff
@@ -321,6 +321,52 @@ func (p *ironicProvisioner) getNode() (*nodes.Node, error) {
...
+
+	p.log.Info("node has assigned port")
+	return true
+}
+
+// Verify that MAC is already allocated to some node port.
+func (p *ironicProvisioner) isPortAllocatedToAddress(address string) bool {
+	if allPorts, err := p.listAllPorts(address); err == nil && len(allPorts) > 0 {
+		p.log.Info("address is allocated to port", "address", address)
+		return true
```
zaneb, line 363: Why not check which node the port is assigned to here, instead of listing the ports again in nodeHasAssignedPort?

zaneb (commented):
So the issue is presumably that if we fail creating the Port the first time, we never retry because port creation only happens when we've just created a new Node. Very similar to #839.

pkg/provisioner/ironic/ironic.go
s3rj1k: moved to private method, seems more readable that way

pkg/provisioner/ironic/ironic.go
s3rj1k: rebased to use versioned logger with level 6

zaneb: Please just use `p.debugLog()`.

pkg/provisioner/ironic/ironic.go
s3rj1k: I'd like to keep signature of parent method as simple as possible, with only `bool` in returned values
( `(r *receiver) f(a args) bool` ) so it can be used if `if` conditions as simple as possible.

For using retries in that case one would have to decide how many times to retry and implement some kind of backoff.

pkg/provisioner/ironic/ironic.go
s3rj1k: yea, missed that, thanks, fixed

pkg/provisioner/ironic/ironic.go
s3rj1k: From `ports.*` functions signatures one can assume that Ironic can assign multiple ports per single node.

So to avoid hard codding logic that in our case only sigle port can be allocated to node, I've split this check into two methods.

 - `isPortAllocatedToAddress` to see if MAC is used in any of available ports
 - `nodeHasAssignedPort` to see if node has some port assigned to it, in theory it can be some other none PXE port
 
 If we are sure that one-on-one mapping of MAC to Node would not potentially break something in that future, I could refactor this PR, but this refactor is basically, `redo PR from scratch`.

pkg/provisioner/ironic/ironic.go
s3rj1k: Sorry, did not notice we have explicit debug level logger there, thought you were talking about versioned one, fixed.

pkg/provisioner/ironic/ironic.go
zaneb: > I'd like to keep signature of parent method as simple as possible

Ignoring errors makes it simpler than is possible :)

> For using retries in that case one would have to decide how many times to retry and implement some kind of backoff.

Just `return transientError(err)`. That's what it's for.

s3rj1k: returning error would require to change logic in 
```
if p.bootMACAddress != "" && !p.nodeHasAssignedPort(ironicNode) && !p.isPortAllocatedToAddress(p.bootMACAddress) {
```

to some nested `if`

We will lose readability because we need to differentiate `error` from `bool`, so returned values from function will have to be `(error, bool)`

Will do if this (`(error, bool)`) sound as acceptable solution.

zaneb: > We will lose readability because we need to differentiate `error` from `bool`,

correctness > readability

> so returned values from function will have to be `(error, bool)`
> 
> Will do if this (`(error, bool)`) sound as acceptable solution.

`bool, error` is idiomatic in go, but yes.

pkg/provisioner/ironic/ironic.go
zaneb: > in theory it can be some other none PXE port

This is the part I disagree with. The only port we add manually on creation of the Node is the PXE port. Therefore if that iniitial port creation fails we always need to try it again to get consistent results, regardless of the existence of any other ports.

The only scenario I can think of in which other ports exist but the PXE port does not is if the MAC address provided is wrong and inspection consequently [removes](https://github.com/metal3-io/ironic-inspector-image/blob/master/ironic-inspector.conf.j2#L36) it while adding the other ports. This doesn't seem to be related to the failure you described. With drivers where NeedsMAC() is true (i.e. the ones where the missing port is a problem) this scenario can't happen, because inspector can't find the Node to update without the PXE port existing.

s3rj1k:
@zaneb Can you please take another look, PR is rebased

s3rj1k:
@zaneb @stbenjam kindly remainder, PR waits review, thanks.

zaneb (commented):
This looks correct, but expensive.

pkg/provisioner/ironic/ironic.go
```diff
@@ -321,6 +321,53 @@ func (p *ironicProvisioner) getNode() (*nodes.Node, error) {
...
 }
 
+// Verifies that node has port assigned by Ironic.
+func (p *ironicProvisioner) nodeHasAssignedPort(ironicNode *nodes.Node) (bool, error) {
+	opts := ports.ListOpts{
+		Fields:   []string{"node_uuid"},
+		NodeUUID: ironicNode.UUID,
+	}
+
+	pager := ports.List(p.client, opts)
```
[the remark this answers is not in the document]
zaneb: Scratch that, looking at the example shows it just contains a link to the API endpoint we can hit to get a list of ports assigned to the node :(

pkg/provisioner/ironic/ironic.go
```diff
@@ -510,6 +567,32 @@ func (p *ironicProvisioner) ValidateManagementAccess(data provisioner.Management
...
+		if p.bootMACAddress != "" {
+			var nodeHasAssignedPort, addressIsAllocatedToPort bool
+
+			nodeHasAssignedPort, err = p.nodeHasAssignedPort(ironicNode)
+			if err != nil {
+				result, err = transientError(err)
+				return
+			}
+
+			addressIsAllocatedToPort, err = p.isPortAllocatedToAddress(p.bootMACAddress)
```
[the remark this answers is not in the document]
s3rj1k: yea, I was intentionally avoiding nested `if`s, to make code more readable, fixed in latest rebase

pkg/provisioner/ironic/ironic.go
```diff
@@ -321,6 +321,53 @@ func (p *ironicProvisioner) getNode() (*nodes.Node, error) {
...
+		p.debugLog.Info("node has no assigned port")
+		return false, nil
+	}
+
+	p.debugLog.Info("node has assigned port")
+	return true, nil
+}
+
+// Verify that MAC is already allocated to some node port.
+func (p *ironicProvisioner) isPortAllocatedToAddress(address string) (bool, error) {
```
zaneb, line 356: Would `isAddressAllocatedToPort()` be a more accurate name?

pkg/provisioner/ironic/ironic.go
```diff
@@ -321,6 +321,53 @@ func (p *ironicProvisioner) getNode() (*nodes.Node, error) {
...
 
+// Verifies that node has port assigned by Ironic.
+func (p *ironicProvisioner) nodeHasAssignedPort(ironicNode *nodes.Node) (bool, error) {
+	opts := ports.ListOpts{
+		Fields:   []string{"node_uuid"},
+		NodeUUID: ironicNode.UUID,
+	}
+
+	pager := ports.List(p.client, opts)
+	if pager.Err != nil {
```
zaneb, line 332: This can't actually happen, because no request has been made yet.

pkg/provisioner/ironic/ironic.go
```diff
@@ -321,6 +321,53 @@ func (p *ironicProvisioner) getNode() (*nodes.Node, error) {
...
+		return false, errors.Wrap(pager.Err, "failed to list ports")
+	}
+
+	allPages, err := pager.AllPages()
+	if err != nil {
+		return false, errors.Wrap(err, "failed to page over list of ports")
```
zaneb, line 338: This is where an error from the request will show up.

pkg/provisioner/ironic/ironic.go
```diff
@@ -321,6 +321,53 @@ func (p *ironicProvisioner) getNode() (*nodes.Node, error) {
...
+	}
+
+	ok, err := allPages.IsEmpty()
```
zaneb, line 341: `empty` would be a more descriptive name than `ok` here.

pkg/provisioner/ironic/ironic.go
```diff
@@ -321,6 +321,53 @@ func (p *ironicProvisioner) getNode() (*nodes.Node, error) {
 	}
 }
 
+// Verifies that node has port assigned by Ironic.
+func (p *ironicProvisioner) nodeHasAssignedPort(ironicNode *nodes.Node) (bool, error) {
+	opts := ports.ListOpts{
+		Fields:   []string{"node_uuid"},
+		NodeUUID: ironicNode.UUID,
+	}
```
zaneb, line 329: It feels like we're copying and pasting all of the logic of listAllPorts() here for the sake of not adding one extra filtering option parameter.

zaneb (commented):
I remain worried about the extra overhead of this on every reconcile, but it doesn't seem avoidable.
/approve

pkg/provisioner/ironic/ironic.go
s3rj1k: code was modeled after `listAllPorts()`
https://github.com/metal3-io/baremetal-operator/blob/2e6cac3facf3ebc913f42084390780954dec70bb/pkg/provisioner/ironic/ironic.go#L290

so I assume that `listAllPorts()` also incorrect?

zaneb: > so I assume that `listAllPorts()` also incorrect?

Indeed.

PR #820, opened 2021-03-18
Eliminate need for "empty" provisioner

### LINKED ISSUE #819: Excessive logging of provisioner details
Since cadeb28e31 we now log details about the provisioner on every reconcile instead of once at startup.

zaneb:
Lazily create the BMCAccess information inside the ironic provisioner when needed for registration, rather than in the factory when creating a Provisioner. This eliminates the need for the empty provisioner workaround introduced in #748 to allow us to update the Host state to Unmanaged without creation of the provisioner failing.

Fixes #819


controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -213,7 +214,13 @@ func (r *BareMetalHostReconciler) Reconcile(ctx context.Context, request ctrl.Re
 		request:        request,
 		bmcCredsSecret: bmcCredsSecret,
 	}
-	prov, err := r.ProvisionerFactory(*host, *bmcCreds, info.publishEvent)
+
+	factory := r.ProvisionerFactory
+	if host.Spec.ExternallyProvisioned && !host.HasBMCDetails() {
+		factory = empty.New
+	}
+
```
andfasano, line 222: The previous implementation was a wrapper function with the main goal of keeping the decisional logic well defined in one place only (for maintenance reasons), and also to reflect the factory having a more dynamic behavior as the new umanaged hosts feature required an evaluation per host (so at every reconcile loop), in contrast with the original implementation where the decision was more static (at the beginning). Definition remained in the main.go I guess for the sake of simplicity - and it could have been moved into a dedicated struct type probably - but the selection still happened in the controller.

The problem of the excessive logging is somehow a side effect of such requirement, since in a standard environment potentially some hosts could use the default provisioner (ironic) while others could use the empty one - considered a completely different provisioner; in such case it could be useful to know which provider has been been used for a given host, and maybe the log traces could be reviewed.

Not sure if I've missed something, but the current proposed implementation instead looks like more as an override (for all the cases - test, demo, default), which is a different approach.

A more sophisticated implementation could be to split between at the startup across the demo/test/production factories, and keep the production one as a factory of factories (deciding between ironic or empty)

zaneb: That's discussed to some extend in the commit message/PR description. The empty provisioner isn't really a provisioner, it's just a piece of runtime logic put in there to stop the real provisioner choking because of being given no BMC details.
Arguably a better approach might be to make the existing provisioners cope gracefully with not being provided BMC details. I guess one way to do that might be to decide between ironic vs empty in the factory for the ironic provisoner.

zaneb: As far as a backportable short-term fix goes though, I still think this is the best approach.

andfasano: > I guess one way to do that might be to decide between ironic vs empty in the factory for the ironic provisoner

My proposal was in that direction (but considering the empty provisioner a valid one, which is the discussion point I guess). For me it would be fine to move it there then, as it could be more clear that it's going to be used to cover the unmanaged case, and the demo/fixture provisioner  previous behaviors will not overridden.

hardys: As [discussed on slack](https://kubernetes.slack.com/archives/CHD49TLE7/p1616436038039800) I think this behavior should be independent of whether the host is `externallyProvisioned` or not

andfasano (approved):
I like the approach, and I'd suggest to transform `p.bmAccess()` in a lazy getter to avoid modifying some Provisioner signatures (and still keeping a max single instantiation per loop as before)

zaneb:
> I like the approach, and I'd suggest to transform `p.bmAccess()` in a lazy getter to avoid modifying some Provisioner signatures (and still keeping a max single instantiation per loop as before)

The only real effect of that would be that you have to copy and paste the error handing to a lot of extra places. `NewAccessDetails()` doesn't make any network calls, it just parses a URL. We can afford to do that twice in one loop in one particular state without caching the result. None of the public Provisioner method signatures are modified.

hardys:
This lgtm but now needs a rebase

andfasano:
> The only real effect of that would be that you have to copy and paste the error handing to a lot of extra places. `NewAccessDetails()` doesn't make any network calls, it just parses a URL. We can afford to do that twice in one loop in one particular state without caching the result. None of the public Provisioner method signatures are modified.

Right, I meant IronicProvisioner private methods that have been changed to accept the bmcCredentials parameters - which make the relationship pretty loosely coupled (surely not a blocking a point, but my preference would be to avoid it).

On a second thought, the current implementation fullfils the requirement expressed in https://github.com/metal3-io/baremetal-operator/pull/824 without triggering any error just thanks to the fact that the BMC credentials check have been postponed later into the `ValidateManagementAccess` (and `Prepare`), a point that will never be reached for an Unmanaged host without credentials, and there's no comment in the code (or even better a test IINW, as it's also missing from `TestMissingBMCParameters`) about that.

hardys:
>  or even better a test IINW, as it's also missing from TestMissingBMCParameters

I did notice there's no test coverage of the `Unmanaged` state when I fixed https://github.com/metal3-io/baremetal-operator/pull/824 but there wasn't an obvious way to solve it at that point.

With this refactor I guess it's a good time to add such coverage - I'm not blocking on it, but it could be added either in this PR or a followup. @andfasano what are your thoughts?

andfasano:
> With this refactor I guess it's a good time to add such coverage - I'm not blocking on it, but it could be added either in this PR or a followup. @andfasano what are your thoughts?

I think either options would be fine. [baremetal_controller_host_test](https://github.com/metal3-io/baremetal-operator/blob/master/controllers/metal3.io/baremetalhost_controller_test.go) and/or [validatemanagementaccess_test](https://github.com/metal3-io/baremetal-operator/blob/master/pkg/provisioner/ironic/validatemanagementaccess_test.go)/[prepare_test](https://github.com/metal3-io/baremetal-operator/blob/master/pkg/provisioner/ironic/prepare_test.go) could be good candidates for capturing that.

zaneb:
Now with more unit tests. Thanks for the suggestion @andfasano, in retrospect that was definitely required.
/test-integration

PR #822, opened 2021-03-22
Bug: fix missing Inspecting state

fmuyassarov:
We are missing ironic node Inspecting state in this particular place. This was initially done in https://github.com/metal3-io/baremetal-operator/pull/607 but it was suggested by @dtantsur to have this fix separately so that it can be also bakcported in OpenShift BMO older versions.


zaneb:
From looking at #785 it seems like the worst thing that can happen is some error logs so I'm not sure a backport is required, but we can by all means fix it here.
/approve
/test-integration

PR #823, opened 2021-03-22
Fix some lint errors

elfosardo:


apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -118,6 +118,8 @@ const (
 	// has any sort of error.
 	OperationalStatusError OperationalStatus = "error"
 
+	// OperationalStatusDelayed is the status value for when the host
+	// deployment needs to be delayed to limit simlultaneous hosts provisioning
```
furkatgofurov7, line 122: tiny nit: 'simultaneous'

dhellmann:
These all look OK. I wonder how the linter job we have didn't catch them. Do we need to change some settings?

apis/metal3.io/v1alpha1/baremetalhost_types.go
elfosardo: d'oh! Me and my typos :/

elfosardo:
> These all look OK. I wonder how the linter job we have didn't catch them. Do we need to change some settings?

apparently we do, most of the linters in golangci-lint are disabled by default
I'll submit a PR with some basic options, we can choose if we want something more advanced later

pkg/provisioner/ironic/ironic.go
```diff
@@ -72,6 +72,7 @@ func (e macAddressConflictError) Error() string {
 	return fmt.Sprintf("MAC address %s conflicts with existing node %s", e.Address, e.ExistingNode)
 }
 
+// NewMacAddressConflictError is a wrap for macAddressConflictError error
```
dtantsur, line 75: This doesn't seem very informative, maybe we should improve it further.

dtantsur (commented):
/lgtm

PR #825, opened 2021-03-23
Enable golint with golangci-lint

elfosardo:
Most linters are disabled by default when using golangci-lint, we need
to explicitely enable them if we want them to run.
This change enables the basic golint support.
For more info please check https://golangci-lint.run/usage/quick-start

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,12 +81,6 @@ unit-verbose: ## Run unit tests with verbose output
 .PHONY: linters
 linters: lint generate-check fmt-check
 
-.PHONY: sec
-sec: lint
-
-.PHONY: fmt
-fmt: lint
-
 .PHONY: vet
 vet: lint
 
@@ -118,7 +112,7 @@ demo: generate lint manifests ## Run in demo mode
 	go run -ldflags $(LDFLAGS) ./main.go -namespace=$(RUN_NAMESPACE) -dev -demo-mode
 
 .PHONY: run-test-mode
-run-test-mode: generate fmt vet manifests ## Run against the configured Kubernetes cluster in ~/.kube/config
+run-test-mode: generate fmt-check lint manifests ## Run against the configured Kubernetes cluster in ~/.kube/config
 	go run -ldflags $(LDFLAGS) ./main.go -namespace=$(RUN_NAMESPACE) -dev -test-mode
 
 .PHONY: install
--- a/pkg/provisioner/demo/demo.go
+++ b/pkg/provisioner/demo/demo.go
@@ -77,7 +77,7 @@ func New(hostData provisioner.HostData, publisher provisioner.EventPublisher) (p
 	return p, nil
 }
 
-func (m *demoProvisioner) HasCapacity() (result bool, err error) {
+func (p *demoProvisioner) HasCapacity() (result bool, err error) {
 	return true, nil
 }
 
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -4,6 +4,7 @@ run:
 linters:
   disable-all: true
   enable: [
+    "golint",
     "govet",
     "goimports",
     "gofmt",
--- a/controllers/metal3.io/action_result.go
+++ b/controllers/metal3.io/action_result.go
@@ -113,7 +113,7 @@ func (r actionError) Dirty() bool {
 }
 
 func (r actionError) NeedsRegistration() bool {
-	return errors.Is(r.err, provisioner.NeedsRegistration)
+	return errors.Is(r.err, provisioner.ErrNeedsRegistration)
 }
 
 // actionFailed is a result indicating that the current action has failed,
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -303,7 +303,7 @@ func (p *ironicProvisioner) listAllPorts(address string) ([]ports.Port, error) {
 
 func (p *ironicProvisioner) getNode() (*nodes.Node, error) {
 	if p.nodeID == "" {
-		return nil, provisioner.NeedsRegistration
+		return nil, provisioner.ErrNeedsRegistration
 	}
 
 	ironicNode, err := nodes.Get(p.client, p.nodeID).Extract()
@@ -314,7 +314,7 @@ func (p *ironicProvisioner) getNode() (*nodes.Node, error) {
 	case gophercloud.ErrDefault404:
 		// Look by ID failed, trying to lookup by hostname in case it was
 		// previously created
-		return nil, provisioner.NeedsRegistration
+		return nil, provisioner.ErrNeedsRegistration
 	default:
 		return nil, errors.Wrap(err,
 			fmt.Sprintf("failed to find node by ID %s", p.nodeID))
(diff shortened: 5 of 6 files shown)


elfosardo:
golint should now fail

elfosardo:
@dhellmann golint and govet now fail, not sure which options we want to keep
if we leave golint, we need to fix the rest besides what's fixed in #823

dhellmann:
I think all of the failures are things we should fix, so I'm OK with the set of rules that are enabled.

The govet and golint jobs are running the same thing. Is govet enabled? We could probably drop one of the jobs if the same tool runs all of the tests in one job.

elfosardo:
> I think all of the failures are things we should fix, so I'm OK with the set of rules that are enabled.
> 
> The govet and golint jobs are running the same thing. Is govet enabled? We could probably drop one of the jobs if the same tool runs all of the tests in one job.

yes, govet is included in golint
I agree we should drop govet and just run golint, I don't have that power though :)
I'm going to fix the rest of the lint errors here

elfosardo:
I will leave vet in makefile until govet job has been removed

fmuyassarov:
> I will leave vet in makefile until govet job has been removed

If the plan is to remove the govet from mandatory CI jobs, that can be done here https://github.com/metal3-io/project-infra/blob/ebe2d90c26bb3bf284fd34c2c171a82058249da6/prow/config/config.yaml#L252-L265

elfosardo:
> > I will leave vet in makefile until govet job has been removed
> 
> If the plan is to remove the govet from mandatory CI jobs, that can be done here https://github.com/metal3-io/project-infra/blob/ebe2d90c26bb3bf284fd34c2c171a82058249da6/prow/config/config.yaml#L252-L265

oh cool, thanks for pointing that out!
I will submit a patch for that, but I think we should merge this first

dtantsur:
What's with this `p` vs `m` in methods? It doesn't seem like a meaningful change to me, and it makes this patch conflict with everything other patch in existence. Maybe skip that?

elfosardo:
> What's with this `p` vs `m` in methods? It doesn't seem like a meaningful change to me, and it makes this patch conflict with everything other patch in existence. Maybe skip that?

@dtantsur it's unfortunately a valid lint error:
pkg/provisioner/demo/demo.go:81:1: receiver name p should be consistent with previous receiver name m for demoProvisioner

we never really ran golint here that's why now we need to make these changes

pkg/provisioner/demo/demo.go
```diff
@@ -78,17 +78,17 @@ func (m *demoProvisioner) HasProvisioningCapacity() (result bool, err error) {
 
 // ValidateManagementAccess tests the connection information for the
 // host to verify that the location and credentials work.
-func (p *demoProvisioner) ValidateManagementAccess(credentialsChanged, force bool) (result provisioner.Result, provID string, err error) {
-	p.log.Info("testing management access")
+func (m *demoProvisioner) ValidateManagementAccess(credentialsChanged, force bool) (result provisioner.Result, provID string, err error) {
```
dhellmann, line 81: This is a lot of code churn to fix a linter issue. What about changing `m` on line 75 to `p`?

dhellmann (changes requested):

Makefile
```diff
@@ -19,7 +19,7 @@ BIN_DIR := bin
 
 CRD_OPTIONS ?= "crd:trivialVersions=false,allowDangerousTypes=true,crdVersions=v1"
 CONTROLLER_GEN ?= go run sigs.k8s.io/controller-tools/cmd/controller-gen
-GOLANGCI_LINT ?= GOLANGCI_LINT_CACHE=$(GOLANGCI_LINT_CACHE) go run github.com/golangci/golangci-lint/cmd/golangci-lint
+GOLANGCI_LINT ?= GOLANGCI_LINT_CACHE=$(GOLANGCI_LINT_CACHE) go run github.com/golangci/golangci-lint/cmd/golangci-lint run
```
asalkeld, line 22: i think leave this as-is

Makefile
```diff
@@ -81,18 +81,13 @@ unit-verbose: ## Run unit tests with verbose output
...
-fmt: lint
-
+# NOTE(elfosardo): leaving this alias until we remove the govet job
 .PHONY: vet
 vet: lint
 
 .PHONY: lint
 lint:
-	$(GOLANGCI_LINT) run
+	$(GOLANGCI_LINT) -E golint
```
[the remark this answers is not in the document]
asalkeld: just add "golint" here https://github.com/metal3-io/baremetal-operator/blob/master/.golangci.yml#L9

asalkeld (changes requested):

Makefile
```diff
@@ -118,7 +112,7 @@ demo: generate lint manifests ## Run in demo mode
 	go run -ldflags $(LDFLAGS) ./main.go -namespace=$(RUN_NAMESPACE) -dev -demo-mode
 
 .PHONY: run-test-mode
-run-test-mode: generate fmt vet manifests ## Run against the configured Kubernetes cluster in ~/.kube/config
+run-test-mode: generate fmt-check lint manifests ## Run against the configured Kubernetes cluster in ~/.kube/config
```
dhellmann, line 115: I think this was using `fmt` instead of `fmt-check` so that code would be fixed before being passed to the compiler.

dhellmann (commented):
One nit, but I'm not sure it really matters.

Thanks for working on this!

/approve

dhellmann (approved):

dhellmann:
oh, now there are merge conflicts

Makefile
elfosardo: I think fmt-check is correct as it's using the gofmt.sh script under /hack
fmt was just an alias for lint
vet is also an alias, but we will remove it later to not break the current jobs

dhellmann:
/approve
/lgtm

Let's get this in so we don't regress the lint job and have to keep fixing issues in this PR.

elfosardo:
@dhellmann not sure why this is "Not mergeable", CI seems ok now :/

dhellmann:
There are some issues with the upstream Prow instance.

@maelk @kashifest @zaneb is it safe for me to merge this PR by hand?

kashifest:
> There are some issues with the upstream Prow instance.
> 
> @maelk @kashifest @zaneb is it safe for me to merge this PR by hand?

OK for me. By the way since tide is still working, we tested IPAM repo with enabling Travis to do a the tests and making prow optional and tide was able to merge it. We configured Travis test as required now. I have a PR to do the same for CAPM3 and Dev-Env. Shall I do the same for BMO? https://github.com/metal3-io/project-infra/pull/179. You have to configure Travis tests as required.

dhellmann:
> > There are some issues with the upstream Prow instance.
> > @maelk @kashifest @zaneb is it safe for me to merge this PR by hand?
> 
> OK for me. By the way since tide is still working, we tested IPAM repo with enabling Travis to do a the tests and making prow optional and tide was able to merge it. We configured Travis test as required now. I have a PR to do the same for CAPM3 and Dev-Env. Shall I do the same for BMO? [metal3-io/project-infra#179](https://github.com/metal3-io/project-infra/pull/179). You have to configure Travis tests as required.

It sounds like that's probably a good idea, but we should talk about it somewhere that more community members will be able to offer their opinion. Maybe the mailing list?

In the mean time, I'll merge this one by hand.

PR #826, opened 2021-03-23
Handle hardwaredetails annotation value error

hardys:
Currently we only handle the case where the json fails to parse
e.g is invalid json, but we silently ignore any value which is
valid data but fails to unpack into the HardwareDetails object.

In this case we should avoid removing the annotation since it's
not been copied into the status.hardware, and return an error
so the user can detect that the json data wasn't in the expected
format.

### DIFF
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -274,7 +274,7 @@ func (r *BareMetalHostReconciler) updateHardwareDetails(request ctrl.Request, ho
 	if host.Status.HardwareDetails == nil || inspectionDisabled(host) {
 		objHardwareDetails, err := r.getHardwareDetailsFromAnnotation(host)
 		if err != nil {
-			return updated, errors.Wrap(err, "Error getting HardwareDetails from annotation")
+			return updated, errors.Wrap(err, "Error parsing HardwareDetails from annotation")
 		}
 		if objHardwareDetails != nil {
 			host.Status.HardwareDetails = objHardwareDetails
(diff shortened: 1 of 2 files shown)


hardys:
@tsorya FYI this should catch the issue you encountered where the json was valid, but the format was not

hardys:
@andfasano @zaneb PTAL - thanks!

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1038,11 +1038,16 @@ func (r *BareMetalHostReconciler) getHardwareDetailsFromAnnotation(host *metal3v
 		return nil, nil
 	}
 	content := []byte(annotations[hardwareDetailsAnnotation])
-	objHardwareDetails := &metal3v1alpha1.HardwareDetails{}
-	if err := json.Unmarshal(content, objHardwareDetails); err != nil {
+	objHardwareDetails := metal3v1alpha1.HardwareDetails{}
+	if err := json.Unmarshal(content, &objHardwareDetails); err != nil {
```
andfasano, line 1042: I think the code could be simplified by using the `DisallowUnknownFields()` of the json Decoder, ie just: 

```
decoder := json.NewDecoder(strings.NewReader(annotations[hardwareDetailsAnnotation]))
decoder.DisallowUnknownFields()

if err := decoder.Decode(&objHardwareDetails); err != nil {
	return nil, err
}
```

In this way if the annotation string contains a non-matching key in the target struct an error should be returned

hardys: Thanks, I switched to this approach in the latest version

zaneb:
It's unfortunate that this error renders the BMH effectively inoperable. But it does seem like garbage data should be treated the same as bad json.
/lgtm

PR #827, opened 2021-03-23
Add support for detached annotation

hardys:
This allows a new annotation as discussed in https://github.com/metal3-io/metal3-docs/pull/168

 ```
baremetalhost.metal3.io/unmanaged
```
*EDIT1* this has now been renamed to "deregistered" for both the annotation and OperationalStatus - I'll update the docs PR and this description when we've reached consensus on that new name.

*EDIT2* ...then renamed again to "detached"

When set this puts the BMH OperationalStatus into a "unmanaged",
which removes the host from the underlying provisioner (without
deprovisioning)

When the annotation removed the BMH can resume management without
any disruptive actions (the previous Provisioning State is maintained)

### DIFF
--- a/apis/metal3.io/v1alpha1/baremetalhost_types.go
+++ b/apis/metal3.io/v1alpha1/baremetalhost_types.go
@@ -155,6 +155,9 @@ const (
 	// PowerManagementError is an error condition occurring when the
 	// controller is unable to modify the power state of the Host.
 	PowerManagementError ErrorType = "power management error"
+	// DetachError is an error condition occurring when the
+	// controller is unable to detatch the host from the provisioner
+	DetachError ErrorType = "detach error"
 )
 
 // ProvisioningState defines the states the provisioner will report
--- a/apis/metal3.io/v1alpha1/baremetalhost_types.go
+++ b/apis/metal3.io/v1alpha1/baremetalhost_types.go
@@ -39,6 +39,10 @@ const (
 	// an immediate requeue)
 	PausedAnnotation = "baremetalhost.metal3.io/paused"
 
+	// Detached is the annotation which stops provisioner management of the host
+	// unlike in the paused case, the host status may be updated
+	DetachedAnnotation = "baremetalhost.metal3.io/detached"
+
 	// StatusAnnotation is the annotation that keeps a copy of the Status of BMH
 	// This is particularly useful when we pivot BMH. If the status
 	// annotation is present and status is empty, BMO will reconstruct BMH Status
@@ -121,6 +125,10 @@ const (
 	// OperationalStatusDelayed is the status value for when the host
 	// deployment needs to be delayed to limit simultaneous hosts provisioning
 	OperationalStatusDelayed = "delayed"
+
+	// OperationalStatusDetached is the status value when the host is
+	// marked unmanaged via the detached annotation
+	OperationalStatusDetached OperationalStatus = "detached"
 )
 
 // ErrorType indicates the class of problem that has caused the Host resource
@@ -625,7 +633,7 @@ type BareMetalHostStatus struct {
 	// after modifying this file
 
 	// OperationalStatus holds the status of the host
-	// +kubebuilder:validation:Enum="";OK;discovered;error;delayed
+	// +kubebuilder:validation:Enum="";OK;discovered;error;delayed;detached
 	OperationalStatus OperationalStatus `json:"operationalStatus"`
 
 	// ErrorType indicates the type of failure encountered when the
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -322,6 +322,7 @@ func recordActionFailure(info *reconcileInfo, errorType metal3v1alpha1.ErrorType
 	setErrorMessage(info.host, errorType, errorMessage)
 
 	eventType := map[metal3v1alpha1.ErrorType]string{
+		metal3v1alpha1.DetachError:                  "DetachError",
 		metal3v1alpha1.ProvisionedRegistrationError: "ProvisionedRegistrationError",
 		metal3v1alpha1.RegistrationError:            "RegistrationError",
 		metal3v1alpha1.InspectionError:              "InspectionError",
@@ -537,6 +538,33 @@ func getCurrentImage(host *metal3v1alpha1.BareMetalHost) *metal3v1alpha1.Image {
 		return host.Spec.Image.DeepCopy()
 	}
 	return nil
+}
+
+// detachHost() detaches the host from the Provisioner
+func (r *BareMetalHostReconciler) detachHost(prov provisioner.Provisioner, info *reconcileInfo) actionResult {
+	provResult, err := prov.Detach()
+	if err != nil {
+		return actionError{errors.Wrap(err, "failed to detach")}
+	}
+	if provResult.ErrorMessage != "" {
+		return recordActionFailure(info, metal3v1alpha1.DetachError, provResult.ErrorMessage)
+	}
+	if provResult.Dirty {
+		if info.host.Status.ErrorType == metal3v1alpha1.DetachError && clearError(info.host) {
+			return actionUpdate{actionContinue{provResult.RequeueAfter}}
+		}
+		return actionContinue{provResult.RequeueAfter}
+	}
+	slowPoll := actionContinue{unmanagedRetryDelay}
+	if info.host.Status.ErrorType == metal3v1alpha1.DetachError {
+		clearError(info.host)
+		info.host.Status.ErrorCount = 0
+	}
+	if info.host.SetOperationalStatus(metal3v1alpha1.OperationalStatusDetached) {
+		info.log.Info("host is detached, removed from provisioner")
+		return actionUpdate{slowPoll}
+	}
+	return slowPoll
 }
 
 // Test the credentials by connecting to the management controller.
--- a/controllers/metal3.io/host_state_machine.go
+++ b/controllers/metal3.io/host_state_machine.go
@@ -183,6 +183,10 @@ func (hsm *hostStateMachine) ReconcileState(info *reconcileInfo) (actionRes acti
 		return actionComplete{}
 	}
 
+	if detachedResult := hsm.checkDetachedHost(info); detachedResult != nil {
+		return detachedResult
+	}
+
 	if registerResult := hsm.ensureRegistered(info); registerResult != nil {
 		hostRegistrationRequired.Inc()
 		return registerResult
@@ -216,7 +220,11 @@ func (hsm *hostStateMachine) checkInitiateDelete() bool {
 	default:
 		hsm.NextState = metal3v1alpha1.StateDeleting
 	case metal3v1alpha1.StateProvisioning, metal3v1alpha1.StateProvisioned:
-		hsm.NextState = metal3v1alpha1.StateDeprovisioning
+		if hsm.Host.OperationalStatus() == metal3v1alpha1.OperationalStatusDetached {
+			hsm.NextState = metal3v1alpha1.StateDeleting
+		} else {
+			hsm.NextState = metal3v1alpha1.StateDeprovisioning
+		}
 	case metal3v1alpha1.StateDeprovisioning:
 		// Allow state machine to run to continue deprovisioning.
 		return false
@@ -225,6 +233,46 @@ func (hsm *hostStateMachine) checkInitiateDelete() bool {
 		return false
 	}
 	return true
+}
+
+// hasInspectAnnotation checks for existence of baremetalhost.metal3.io/detached
+func hasDetachedAnnotation(host *metal3v1alpha1.BareMetalHost) bool {
+	annotations := host.GetAnnotations()
+	if annotations != nil {
+		if _, ok := annotations[metal3v1alpha1.DetachedAnnotation]; ok {
+			return true
+		}
+	}
+	return false
+}
+
+func (hsm *hostStateMachine) checkDetachedHost(info *reconcileInfo) (result actionResult) {
+	// If the detached annotation is set we remove any host from the
+	// provisioner and take no further action
+	// Note this doesn't change the current state, only the OperationalStatus
+	if hasDetachedAnnotation(hsm.Host) {
+		// Only allow detaching hosts in Provisioned/ExternallyProvisioned states
+		switch info.host.Status.Provisioning.State {
+		case metal3v1alpha1.StateProvisioned, metal3v1alpha1.StateExternallyProvisioned:
+			return hsm.Reconciler.detachHost(hsm.Provisioner, info)
+		}
+	}
+	if info.host.Status.ErrorType == metal3v1alpha1.DetachError {
+		clearError(info.host)
+		hsm.Host.Status.ErrorCount = 0
+		info.log.Info("removed detach error")
+		return actionUpdate{}
+	}
+	if info.host.OperationalStatus() == metal3v1alpha1.OperationalStatusDetached {
+		newStatus := metal3v1alpha1.OperationalStatusOK
+		if info.host.Status.ErrorType != "" {
+			newStatus = metal3v1alpha1.OperationalStatusError
+		}
+		info.host.SetOperationalStatus(newStatus)
+		info.log.Info("removed detached status")
+		return actionUpdate{}
+	}
+	return nil
 }
 
 func (hsm *hostStateMachine) ensureRegistered(info *reconcileInfo) (result actionResult) {
--- a/controllers/metal3.io/host_state_machine_test.go
+++ b/controllers/metal3.io/host_state_machine_test.go
@@ -649,6 +649,10 @@ func (m *mockProvisioner) Delete() (result provisioner.Result, err error) {
 	return m.getNextResultByMethod("Delete"), err
 }
 
+func (m *mockProvisioner) Detach() (result provisioner.Result, err error) {
+	return m.getNextResultByMethod("Detach"), err
+}
+
 func (m *mockProvisioner) PowerOn() (result provisioner.Result, err error) {
 	return m.getNextResultByMethod("PowerOn"), err
 }
--- a/controllers/metal3.io/host_state_machine_test.go
+++ b/controllers/metal3.io/host_state_machine_test.go
@@ -146,6 +146,238 @@ func TestProvisioningCapacity(t *testing.T) {
 	}
 }
 
+func TestDetach(t *testing.T) {
+	testCases := []struct {
+		Scenario                  string
+		Host                      *metal3v1alpha1.BareMetalHost
+		HasDetachedAnnotation     bool
+		ExpectedDetach            bool
+		ExpectedDirty             bool
+		ExpectedOperationalStatus metal3v1alpha1.OperationalStatus
+		ExpectedState             metal3v1alpha1.ProvisioningState
+	}{
+		{
+			Scenario:                  "ProvisionedHost",
+			Host:                      host(metal3v1alpha1.StateProvisioned).build(),
+			ExpectedDetach:            false,
+			ExpectedDirty:             false,
+			ExpectedOperationalStatus: metal3v1alpha1.OperationalStatusOK,
+			ExpectedState:             metal3v1alpha1.StateProvisioned,
+		},
+		{
+			Scenario:                  "DetachProvisionedHost",
+			Host:                      host(metal3v1alpha1.StateProvisioned).build(),
+			HasDetachedAnnotation:     true,
+			ExpectedDetach:            true,
+			ExpectedDirty:             true,
+			ExpectedOperationalStatus: metal3v1alpha1.OperationalStatusDetached,
+			ExpectedState:             metal3v1alpha1.StateProvisioned,
+		},
+		{
+			Scenario:                  "DeleteDetachedProvisionedHost",
+			Host:                      host(metal3v1alpha1.StateProvisioned).SetOperationalStatus(metal3v1alpha1.OperationalStatusDetached).setDeletion().build(),
+			HasDetachedAnnotation:     true,
+			ExpectedDetach:            false,
+			ExpectedDirty:             true,
+			ExpectedOperationalStatus: metal3v1alpha1.OperationalStatusDetached,
+			// Should move to Deleting without any Deprovisioning
+			ExpectedState: metal3v1alpha1.StateDeleting,
+		},
+		{
+			Scenario:                  "ExternallyProvisionedHost",
+			Host:                      host(metal3v1alpha1.StateExternallyProvisioned).SetExternallyProvisioned().build(),
+			HasDetachedAnnotation:     false,
+			ExpectedDetach:            false,
+			ExpectedDirty:             false,
+			ExpectedOperationalStatus: metal3v1alpha1.OperationalStatusOK,
+			ExpectedState:             metal3v1alpha1.StateExternallyProvisioned,
+		},
+		{
+			Scenario:                  "DetachExternallyProvisionedHost",
+			Host:                      host(metal3v1alpha1.StateExternallyProvisioned).SetExternallyProvisioned().build(),
+			HasDetachedAnnotation:     true,
+			ExpectedDetach:            true,
+			ExpectedDirty:             true,
+			ExpectedOperationalStatus: metal3v1alpha1.OperationalStatusDetached,
+			ExpectedState:             metal3v1alpha1.StateExternallyProvisioned,
+		},
+		{
+			Scenario:                  "NoneHost",
+			Host:                      host(metal3v1alpha1.StateNone).build(),
+			HasDetachedAnnotation:     true,
+			ExpectedDetach:            false,
+			ExpectedDirty:             true,
+			ExpectedOperationalStatus: metal3v1alpha1.OperationalStatusDiscovered,
+			ExpectedState:             metal3v1alpha1.StateUnmanaged,
+		},
+		{
+			Scenario:                  "UnmanagedHost",
+			Host:                      host(metal3v1alpha1.StateUnmanaged).build(),
+			HasDetachedAnnotation:     true,
+			ExpectedDetach:            false,
+			ExpectedDirty:             false,
+			ExpectedOperationalStatus: metal3v1alpha1.OperationalStatusOK,
+			ExpectedState:             metal3v1alpha1.StateUnmanaged,
+		},
+		{
+			Scenario:                  "RegisteringHost",
+			Host:                      host(metal3v1alpha1.StateRegistering).build(),
+			HasDetachedAnnotation:     true,
+			ExpectedDetach:            false,
+			ExpectedDirty:             true,
+			ExpectedOperationalStatus: metal3v1alpha1.OperationalStatusOK,
+			ExpectedState:             metal3v1alpha1.StateInspecting,
+		},
+		{
+			Scenario:                  "InspectingHost",
+			Host:                      host(metal3v1alpha1.StateInspecting).build(),
+			HasDetachedAnnotation:     true,
+			ExpectedDetach:            false,
+			ExpectedDirty:             true,
+			ExpectedOperationalStatus: metal3v1alpha1.OperationalStatusOK,
+			ExpectedState:             metal3v1alpha1.StateMatchProfile,
+		},
+		{
+			Scenario:                  "MatchProfileHost",
+			Host:                      host(metal3v1alpha1.StateMatchProfile).build(),
+			HasDetachedAnnotation:     true,
+			ExpectedDetach:            false,
+			ExpectedDirty:             true,
+			ExpectedOperationalStatus: metal3v1alpha1.OperationalStatusOK,
+			ExpectedState:             metal3v1alpha1.StatePreparing,
+		},
+		{
+			Scenario:                  "AvailableHost",
+			Host:                      host(metal3v1alpha1.StateAvailable).build(),
+			HasDetachedAnnotation:     true,
+			ExpectedDetach:            false,
+			ExpectedDirty:             true,
+			ExpectedOperationalStatus: metal3v1alpha1.OperationalStatusOK,
+			ExpectedState:             metal3v1alpha1.StatePreparing,
+		},
+		{
+			Scenario:                  "PreparingHost",
+			Host:                      host(metal3v1alpha1.StatePreparing).build(),
+			HasDetachedAnnotation:     true,
+			ExpectedDetach:            false,
+			ExpectedDirty:             true,
+			ExpectedOperationalStatus: metal3v1alpha1.OperationalStatusOK,
+			ExpectedState:             metal3v1alpha1.StateReady,
+		},
+		{
+			Scenario:                  "ReadyHost",
+			Host:                      host(metal3v1alpha1.StateReady).build(),
+			HasDetachedAnnotation:     true,
+			ExpectedDetach:            false,
+			ExpectedDirty:             true,
+			ExpectedOperationalStatus: metal3v1alpha1.OperationalStatusOK,
+			ExpectedState:             metal3v1alpha1.StatePreparing,
+		},
+		{
+			Scenario:                  "ProvisioningHost",
+			Host:                      host(metal3v1alpha1.StateProvisioning).build(),
+			HasDetachedAnnotation:     true,
+			ExpectedDetach:            false,
+			ExpectedDirty:             true,
+			ExpectedOperationalStatus: metal3v1alpha1.OperationalStatusOK,
+			ExpectedState:             metal3v1alpha1.StateProvisioned,
+		},
+		{
+			Scenario:                  "DeprovisioningHost",
+			Host:                      host(metal3v1alpha1.StateDeprovisioning).build(),
+			HasDetachedAnnotation:     true,
+			ExpectedDetach:            false,
+			ExpectedDirty:             true,
+			ExpectedOperationalStatus: metal3v1alpha1.OperationalStatusOK,
+			ExpectedState:             metal3v1alpha1.StateReady,
+		},
+		{
+			Scenario:                  "DeletingHost",
+			Host:                      host(metal3v1alpha1.StateDeleting).setDeletion().build(),
+			HasDetachedAnnotation:     true,
+			ExpectedDetach:            false,
+			ExpectedDirty:             false,
+			ExpectedOperationalStatus: metal3v1alpha1.OperationalStatusOK,
+			ExpectedState:             metal3v1alpha1.StateDeleting,
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.Scenario, func(t *testing.T) {
+			if tc.HasDetachedAnnotation {
+				tc.Host.Annotations = map[string]string{
+					metal3v1alpha1.DetachedAnnotation: "true",
+				}
+			}
+			prov := newMockProvisioner()
+			hsm := newHostStateMachine(tc.Host, &BareMetalHostReconciler{}, prov, true)
+			info := makeDefaultReconcileInfo(tc.Host)
+			result := hsm.ReconcileState(info)
+
+			assert.Equal(t, tc.ExpectedDetach, prov.calledNoError("Detach"), "ExpectedDetach mismatch")
+			assert.Equal(t, tc.ExpectedDirty, result.Dirty(), "ExpectedDirty mismatch")
+			assert.Equal(t, tc.ExpectedOperationalStatus, info.host.OperationalStatus())
+			assert.Equal(t, tc.ExpectedState, info.host.Status.Provisioning.State)
+		})
+	}
+}
+
+func TestDetachError(t *testing.T) {
+	testCases := []struct {
+		Scenario                  string
+		Host                      *metal3v1alpha1.BareMetalHost
+		ExpectedOperationalStatus metal3v1alpha1.OperationalStatus
+		ExpectedState             metal3v1alpha1.ProvisioningState
+		ClearError                bool
+		RemoveAnnotation          bool
+	}{
+		{
+			Scenario:                  "ProvisionerTemporaryError",
+			Host:                      host(metal3v1alpha1.StateProvisioned).build(),
+			ExpectedOperationalStatus: metal3v1alpha1.OperationalStatusDetached,
+			ExpectedState:             metal3v1alpha1.StateProvisioned,
+			ClearError:                true,
+		},
+		{
+			Scenario:                  "AnnotationRemovedAfterDetachError",
+			Host:                      host(metal3v1alpha1.StateProvisioned).build(),
+			ExpectedOperationalStatus: metal3v1alpha1.OperationalStatusOK,
+			ExpectedState:             metal3v1alpha1.StateProvisioned,
+			RemoveAnnotation:          true,
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.Scenario, func(t *testing.T) {
+			tc.Host.Annotations = map[string]string{
+				metal3v1alpha1.DetachedAnnotation: "true",
+			}
+			prov := newMockProvisioner()
+			hsm := newHostStateMachine(tc.Host, &BareMetalHostReconciler{}, prov, true)
+			info := makeDefaultReconcileInfo(tc.Host)
+
+			prov.setNextError("Detach", "some error")
+			result := hsm.ReconcileState(info)
+			assert.True(t, result.Dirty())
+			assert.Equal(t, 1, tc.Host.Status.ErrorCount)
+			assert.Equal(t, metal3v1alpha1.OperationalStatusError, info.host.OperationalStatus())
+			assert.Equal(t, v1alpha1.DetachError, info.host.Status.ErrorType)
+			assert.Equal(t, tc.ExpectedState, info.host.Status.Provisioning.State)
+
+			if tc.ClearError {
+				prov.clearNextError("Detach")
+			}
+			if tc.RemoveAnnotation {
+				tc.Host.Annotations = map[string]string{}
+			}
+			result = hsm.ReconcileState(info)
+			assert.Equal(t, 0, tc.Host.Status.ErrorCount)
+			assert.True(t, result.Dirty())
+			assert.Equal(t, tc.ExpectedOperationalStatus, info.host.OperationalStatus())
+			assert.Equal(t, tc.ExpectedState, info.host.Status.Provisioning.State)
+			assert.Empty(t, info.host.Status.ErrorType)
+		})
+	}
+}
+
 func TestProvisioningCancelled(t *testing.T) {
 	testCases := []struct {
 		Scenario string
@@ -587,17 +819,21 @@ func newMockProvisioner() *mockProvisioner {
 	return &mockProvisioner{
 		hasProvisioningCapacity: true,
 		nextResults:             make(map[string]provisioner.Result),
+		callsNoError:            make(map[string]bool),
 	}
 }
 
 type mockProvisioner struct {
 	hasProvisioningCapacity bool
 	nextResults             map[string]provisioner.Result
+	callsNoError            map[string]bool
 }
 
 func (m *mockProvisioner) getNextResultByMethod(name string) (result provisioner.Result) {
 	if value, ok := m.nextResults[name]; ok {
 		result = value
+	} else {
+		m.callsNoError[name] = true
 	}
 	return
 }
@@ -614,6 +850,14 @@ func (m *mockProvisioner) setNextError(methodName, msg string) {
 	m.nextResults[methodName] = provisioner.Result{
 		ErrorMessage: msg,
 	}
+}
+
+func (m *mockProvisioner) clearNextError(methodName string) {
+	m.nextResults[methodName] = provisioner.Result{}
+}
+
+func (m *mockProvisioner) calledNoError(methodName string) bool {
+	return m.callsNoError[methodName]
 }
 
 func (m *mockProvisioner) ValidateManagementAccess(data provisioner.ManagementAccessData, credentialsChanged, force bool) (result provisioner.Result, provID string, err error) {
@@ -650,7 +894,8 @@ func (m *mockProvisioner) Delete() (result provisioner.Result, err error) {
 }
 
 func (m *mockProvisioner) Detach() (result provisioner.Result, err error) {
-	return m.getNextResultByMethod("Detach"), err
+	res := m.getNextResultByMethod("Detach")
+	return res, err
 }
 
 func (m *mockProvisioner) PowerOn() (result provisioner.Result, err error) {
--- a/docs/api.md
+++ b/docs/api.md
@@ -459,3 +459,14 @@ put any value on this annotation **other than `metal3.io/capm3`**. Please make
 sure that you remove the annotation  **only if the value of the annotation is
 not `metal3.io/capm3`, but another value that you have provided**. Removing the
 annotation will enable the reconciliation again.
+
+## Detaching hosts
+
+It is possible to prevent management of a BareMetalHost object by adding
+an annotation `baremetalhost.metal3.io/detached`. This removes the host from
+the provisioner, which prevents any management of the physical host (e.g
+changing power state, or deprovisioning), but still allows the BMH status
+to be updated unlike the `paused` anotation. While in this state the
+OperationalStatus field will be `detached` but the provisioning state will
+be unmodified.  This API only has any effect for BareMetalHost resources
+that are in either `Provisioned` or `ExternallyProvisioned` state.
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -1665,6 +1665,16 @@ func (p *ironicProvisioner) Delete() (result provisioner.Result, err error) {
 	return operationContinuing(0)
 }
 
+// Detach removes the host from the provisioning system.
+// Similar to Delete, but ensures non-interruptive behavior
+// for the target system.  It may be called multiple times,
+// and should return true for its dirty  flag until the
+// deletion operation is completed.
+func (p *ironicProvisioner) Detach() (result provisioner.Result, err error) {
+	// Currently the same behavior as Delete()
+	return p.Delete()
+}
+
 func (p *ironicProvisioner) changePower(ironicNode *nodes.Node, target nodes.TargetPowerState) (result provisioner.Result, err error) {
 	p.log.Info("changing power state")
 
--- a/config/crd/bases/metal3.io_baremetalhosts.yaml
+++ b/config/crd/bases/metal3.io_baremetalhosts.yaml
@@ -596,6 +596,7 @@ spec:
                 - discovered
                 - error
                 - delayed
+                - detached
                 type: string
               poweredOn:
                 description: indicator for whether or not the host is powered on
--- a/config/render/capm3.yaml
+++ b/config/render/capm3.yaml
@@ -594,6 +594,7 @@ spec:
                 - discovered
                 - error
                 - delayed
+                - detached
                 type: string
               poweredOn:
                 description: indicator for whether or not the host is powered on
--- a/pkg/provisioner/demo/demo.go
+++ b/pkg/provisioner/demo/demo.go
@@ -288,6 +288,16 @@ func (p *demoProvisioner) Delete() (result provisioner.Result, err error) {
 	return result, nil
 }
 
+// Detach removes the host from the provisioning system.
+// Similar to Delete, but ensures non-interruptive behavior
+// for the target system.  It may be called multiple times,
+// and should return true for its dirty  flag until the
+// deletion operation is completed.
+func (p *demoProvisioner) Detach() (result provisioner.Result, err error) {
+	p.log.Info("detaching host")
+	return result, nil
+}
+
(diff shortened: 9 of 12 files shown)

### CODE REVIEW

**Comment by hardys** [comment]
/cc @maelk @fmuyassarov - this leaves the existing `paused` annotation behavior intact, and allows for a pivot where the BMH resources remain as inventory on the parent cluster.

In the community meeting we discussed the CAPI pivot process, can you confirm if this approach will be acceptable from that perspective since it's not impacting any existing APIs?

File: controllers/metal3.io/baremetalhost_controller.go [hunk 1]
```diff
@@ -227,10 +227,30 @@ func (r *BareMetalHostReconciler) Reconcile(ctx context.Context, request ctrl.Re
 		return ctrl.Result{Requeue: true, RequeueAfter: provisionerNotReadyRetryDelay}, nil
 	}
 
+	// If the unmanaged annotation is set we remove any host from the
+	// provisioner and take no further action
+	if annotations != nil {
+		if _, ok := annotations[metal3v1alpha1.UnmanagedAnnotation]; ok {
+			provResult, err := prov.Delete()
```
andfasano, line 234: One doubt I have about implementing this task in the main reconcile loop _outside_ the state machine it's that the annotations could bypass completely the (de)provisioning limit (currently introduced by @s3rj1k in #807)

hardys: @andfasano That is true, but note we're not doing any deprovisioning here, it only deletes the host from Ironic, which should be a much less expensive operation, and will just get retried if for any reason it fails?

hardys: Note that the reason I did it like this is the feedback I got on https://github.com/metal3-io/metal3-docs/pull/168 indicated (IIUC) that the preference is to avoid adding any new state for this, and also to avoid adding any `$state->Unmanaged` transition (since it'll be hard to reliably implement any subsequent `Unmanaged->$state` transition without adding more "last state" data to the status.)

Doing it in the reconcile loop means we avoid touching the state machine at all, so it's much easier to resume management via the state machine in future if/when the annotation is removed.

andfasano: > @andfasano That is true, but note we're not doing any deprovisioning here, it only deletes the host from Ironic, which should be a much less expensive operation, and will just get retried if for any reason it fails?

Agree that in this case the overall scenario sounds less expensive, but since the deleting was part of it I raised this discussion point. Such kind of [issues](https://github.com/metal3-io/baremetal-operator/pull/807#issuecomment-790802043) start to appear when dealing simultaneously with an high number of BMH, so probably the real question probably is about the chances of ending up in a similar difficult scenario when using the unmanaged annotation.

> the preference is to avoid adding any new state for this

Right, sorry if I wasn't clear but I was referring to the [delay mechanism](https://github.com/metal3-io/baremetal-operator/pull/807/files#diff-eeed3703d04e4c23a7d7af8cd0b7931b6b7990f23d826c49bdbc31c5f0a50291R165) currently being part of the host state machine as a very early check (potentially the presence of the annotation could be an additional triggering condition in the delay check, and the task post-poned after it)

fmuyassarov:
> /cc @maelk @fmuyassarov - this leaves the existing `paused` annotation behavior intact, and allows for a pivot where the BMH resources remain as inventory on the parent cluster.
> 
> In the community meeting we discussed the CAPI pivot process, can you confirm if this approach will be acceptable from that perspective since it's not impacting any existing APIs?

I believe it is okay, since this patch is not touching the pause annotation and logic in [`Reconcile()`](https://github.com/metal3-io/baremetal-operator/blob/0e6772bcef0ff4db9185b3ebaa6f3eeae58d6e60/controllers/metal3.io/baremetalhost_controller.go#L117).
/cc @kashifest

controllers/metal3.io/baremetalhost_controller.go
hardys: > Such kind of issues start to appear when dealing simultaneously with an high number of BMH

Understood - in this case there is no change in the state of the physical host, so I suspect these issues are not relevant, but perhaps @s3rj1k can confirm

The main risk AFAICS is that someone tries to annotate $many BMH resources simultaneously, then it's perhaps possible that the `prov.Delete()` will fail temporarily if the ironic-api is overloaded - in that case we just retry on the next reconcile.

> I was referring to the delay mechanism currently being part of the host state machine as a very early check

Right, but as proposed we'd have to either change the BMH state to `metal3v1alpha1.StateDeleting` (which I want to avoid so we can return easily to the existing state if/when the annotation is removed), or we'd have to make the state machine care about annotations as you mention.

If folks feel strongly this case really requires the limit/delay I can look at adding it, but I'm not currently convinced the additional complexity is justified?

kashifest:
> /cc @maelk @fmuyassarov - this leaves the existing `paused` annotation behavior intact, and allows for a pivot where the BMH resources remain as inventory on the parent cluster.
> 
> In the community meeting we discussed the CAPI pivot process, can you confirm if this approach will be acceptable from that perspective since it's not impacting any existing APIs?

As long as `paused` and `status Annotation` behavior is intact, CAPI pivot process should be good IMO.

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -119,6 +123,10 @@ const (
 	OperationalStatusError OperationalStatus = "error"
 
 	OperationalStatusDelayed = "delayed"
+
+	// OperationalStatusUnmanaged is the status value when the host is
+	// marked unmanaged via the unmanaged annotation
+	OperationalStatusUnmanaged OperationalStatus = "unmanaged"
```
zaneb, line 129: It is slightly confusing that there is also an "Unmanaged" provisioning state that has operational status "discovered". I wonder if there are any other possible names.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -227,10 +227,30 @@ func (r *BareMetalHostReconciler) Reconcile(ctx context.Context, request ctrl.Re
...
+			if err != nil {
+				return ctrl.Result{}, errors.Wrap(err, "failed to delete on unmanage")
+			}
+			if provResult.Dirty {
+				return ctrl.Result{Requeue: true, RequeueAfter: provResult.RequeueAfter}, nil
+			}
+			info.host.SetOperationalStatus(metal3v1alpha1.OperationalStatusUnmanaged)
```
zaneb, line 241: I think we also need some code to set the operational status back to ok (or error, or potentially even discovering) once the annotation is removed.

apis/metal3.io/v1alpha1/baremetalhost_types.go
hardys: Yeah I thought the same but wasn't able to come up with an alternative that conveyed the right meaning, suggestions welcome! :)

From your provisioner API suggestions:
* Ignored - This doesn't convey quite the same meaning to me, but it's fairly close if we document that an "ignored" BMH is frozen in its current state, without any management by the provisioner.
* Paused - overlaps with the existing paused annotation which has slightly different behavior   
* Deregistered - this could work as the operational status, but it reflects the internal state (deregistered from the provisioner) not really the state of the BMH resource? 

IMO unmanaged conveys the status most accurately, but if we think it's too confusing having both state and status with that name, I'm thinking "ignored" or perhaps "frozen"

controllers/metal3.io/baremetalhost_controller.go
s3rj1k: @hardys Yea, I also see a potential overload for ironic database backend on burst.

I would prefer that all ironic operations fall into delay mechanism and also keep state of BMH.

IMO: Changing state of BMH to de-provisioning makes unmanaged annotation useless, one could just de-provision a host and remove it from cluster.

andfasano: @zaneb potentially yes, since the 'prov.Delete()' is an action that will occupy a slot at the end. So I agree in moving the function later in the state machine, and moreover trying to grant a slot before applying it (as a side note, the delay mechanism will also take care of the backoff)

hardys: > @hardys Yea, I also see a potential overload for ironic database backend on burst.
> I would prefer that all ironic operations fall into delay mechanism and also keep state of BMH.

Ack, sounds like there are multiple reasons to move this into the state-machines so I'll take a look at that.

> IMO: Changing state of BMH to de-provisioning makes unmanaged annotation useless, one could just de-provision a host and remove it from cluster.

The motivation is to avoid deprovisioning (and removing the BMH) in a multi-tier deployment - there should never be a reason to apply this annotation to a host in the de-provisioning state, so we should probably make that impossible via the state-machine rework.

controllers/metal3.io/baremetalhost_controller.go
hardys: @zaneb Doesn't this have the same problem as tracking the previous state in the `Unmanaged` case?  E.g we don't know what the previous operational status was unless it's saved somewhere, so the only option is to set it back to OK, then expect subsequent reconciliation to modify the status afterwards?

hardys: I was thinking that OperationalStatus will be set by the state machine when we remove the unmanaged annotation, then revisit whatever state-handler aligns with the current state which will remove the unamanaged status and restore the state-appropriate value (or error)?

hardys: Ok I see now that not every state handler sets the operational status, so we either have to change that, or just set the status to OK on removal of the annotation AFAICS?

hardys: Ok I set it back to OK in the latest version

apis/metal3.io/v1alpha1/baremetalhost_types.go
hardys: I renamed the status and the annotation to "deregistered"

controllers/metal3.io/baremetalhost_controller.go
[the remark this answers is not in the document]
hardys: Ack - this is still TODO

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -227,10 +227,30 @@ func (r *BareMetalHostReconciler) Reconcile(ctx context.Context, request ctrl.Re
...
+			errStatus := r.Status().Update(ctx, host)
+			if errStatus != nil {
+				return ctrl.Result{}, errors.Wrap(errStatus, "Could not set OperationalStatusUnmanaged")
+			}
+			reqLogger.Info("host is unmanaged, removed from provisioner")
+			return ctrl.Result{Requeue: false}, nil
+		}
+	}
```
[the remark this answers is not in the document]
hardys: Ack, moved to `host_state_machine.go` in the latest version

controllers/metal3.io/baremetalhost_controller.go
hardys: Moved to `host_state_machine.go` in the latest version

hardys:
@zaneb @andfasano I reworked this to the state machine as suggested, the new provisioner API (and docs/tests) are still TODO.

PTAL, if you're happy with this revised approach I'll proceed with the remaining pieces - thanks!

controllers/metal3.io/host_state_machine.go
```diff
@@ -226,6 +230,34 @@ func (hsm *hostStateMachine) checkInitiateDelete() bool {
 	return true
 }
 
+func (hsm *hostStateMachine) checkDeregisteredHost(info *reconcileInfo) (result actionResult) {
+	// If the deregistered annotation is set we remove any host from the
+	// provisioner and take no further action
+	// Note this doesn't change the current state, only the OperationalStatus
+	annotations := hsm.Host.GetAnnotations()
+	if annotations != nil {
+		if _, ok := annotations[metal3v1alpha1.DeregisteredAnnotation]; ok {
```
zaneb, line 239: Here would be the place to switch on the state if we decide to do that.
We're thinking we don't want this to happen in the [De]Provisioning states.
We also don't want to change the operational status in the Unmanaged state imho, although we do still want to return early.

It occurs to me that if we leave IPA running (i.e. in any state except [Externally]Provisioned) then the hosts will get rediscovered by ironic and the nodes recreated. I guess if we don't move them to the manageable state that's OK though?

controllers/metal3.io/host_state_machine.go
```diff
@@ -226,6 +230,34 @@ func (hsm *hostStateMachine) checkInitiateDelete() bool {
...
+			if info.host.OperationalStatus() != metal3v1alpha1.OperationalStatusDeregistered {
+				_, err := hsm.Provisioner.Delete()
+				if err != nil {
+					return actionError{errors.Wrap(err, "failed to deregister deregistered host")}
+				}
+				info.host.SetOperationalStatus(metal3v1alpha1.OperationalStatusDeregistered)
+				info.log.Info("host is deregistered, removed from provisioner")
+				return actionComplete{}
+			}
+			return deregisterComplete{}
```
zaneb, line 249: Elsewhere we've adopted a policy that we should never just stop reconciling a Host altogether. e.g. in actionUnmanaged we return actionContinue with unmanagedRetryDelay (which is 10 minutes). It would seem appropriate to do the same here.

controllers/metal3.io/action_result.go
```diff
@@ -97,6 +97,13 @@ func (r deleteComplete) Dirty() bool {
 	return false
 }
 
+// deregisterComplete is a result indicating that the deregistering
+// action completed, similar to deleteComplete there is no need to
+// requeue or mark Dirty, but the CR is not deleted in this case.
```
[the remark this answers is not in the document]
hardys: Yeah I was surprised the only action with the behavior needed was deleteComplete - renaming it seems reasonable to me.

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -39,6 +39,10 @@ const (
 	// an immediate requeue)
 	PausedAnnotation = "baremetalhost.metal3.io/paused"
 
+	// Deregistered is the annotation which stops provisioner management of the host
+	// unlike in the paused case, the host status may be updated
+	DeregisteredAnnotation = "baremetalhost.metal3.io/deregistered"
```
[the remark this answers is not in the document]
hardys: Argh! If only `Unmanaged` and `paused` weren't already in use! :laughing: 

After thinking about the name I reached the conclusion that `deregistered` reflects the internal state quite well (and it was definitely confusing having a State and OperationalStatus both called `unmanaged` after I moved the code into the same file).

Then it seemed weird to dislocate that naming convention at the annotation level, so I just renamed it and personally I think deregistered is an OK name given that the two most obvious alternatives are already in use?

With the benefit of hindsight we should perhaps have called the `paused` annotation `frozen` or something, but ultimately we just have to pick a name then document what it does.

I don't want to burn a bunch more cycles on the naming, but does anyone else have strong opinions on this?

controllers/metal3.io/host_state_machine.go
hardys: @zaneb I was a little confused about that - what's the reason for the periodic retry even for resources in a static/done state?  Won't any change to the resource trigger reconciliation without the periodic retry?

controllers/metal3.io/host_state_machine.go
hardys: > It occurs to me that if we leave IPA running (i.e. in any state except [Externally]Provisioned) then the hosts will get rediscovered by ironic and the nodes recreated. I guess if we don't move them to the manageable state that's OK though?

That's assuming a discovery workflow exists, which it doesn't today, right?

I'd expect any discovery flow to be smart enough to avoid creating duplicate BMH resources when an existing one exists (it'd have to rely on matching `bootMACAddress` I guess) - that's a slightly different case than was solved by @honza in https://github.com/metal3-io/baremetal-operator/pull/780 though (which relies on matching the mac at the Ironic level AFAICS.

I guess one option is to only action this annotation when the state is `Provisioned` - then we know IPA is done, and the only valid transition from that point will be deleting (without deprovisioning), unless the annotation is removed?

apis/metal3.io/v1alpha1/baremetalhost_types.go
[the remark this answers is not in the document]
andfasano: Not a strong one, my inclination would be more on `detach` rather then `deregister` - just because we already have a `Registering` state, and to me sounds as something related. `detach` gives me more the idea of a BMH still alive but "detached" from an Ironic node.

controllers/metal3.io/host_state_machine.go
```diff
@@ -226,6 +230,34 @@ func (hsm *hostStateMachine) checkInitiateDelete() bool {
...
 
+func (hsm *hostStateMachine) checkDeregisteredHost(info *reconcileInfo) (result actionResult) {
+	// If the deregistered annotation is set we remove any host from the
+	// provisioner and take no further action
+	// Note this doesn't change the current state, only the OperationalStatus
+	annotations := hsm.Host.GetAnnotations()
+	if annotations != nil {
+		if _, ok := annotations[metal3v1alpha1.DeregisteredAnnotation]; ok {
+			if info.host.OperationalStatus() != metal3v1alpha1.OperationalStatusDeregistered {
+				_, err := hsm.Provisioner.Delete()
```
andfasano, line 241: Note: adding a 
```
if actionRes := hsm.ensureProvisioningCapacity(info); actionRes != nil {
			return actionRes
}
```
just before the `Provisioner.Delete()` will be enough to check for a free slot before addressing the removal.

zaneb: I don't see any reason to check for a free slot. We don't do that for any other calls that only touch ironic.

andfasano: Right, but in #807 the `nodes.Deleting` state is going to be included, so a node that is currently being deregistered will be considered occupying a slot.

andfasano: (note that I do not consider this a blocking point for the current PR, but I was noticing that the thanks to the latest @hardys refactoring it would have been simpler to add such check if required)

zaneb: Ah, I see that you are another victim of terrible naming in ironic.
A reasonable person might think that nodes.Deleting is the state that the node goes into when you delete it. It isn't.

nodes.Deleting is what happens when you deprovision the node, i.e. a Nova delete in OpenStack. It's triggered by a HTTP PATCH setting targetState=deleting. In metal³ this happens in the Deprovision() call.
When you actually delete a node it just disappears from the database and that's the end of it. It's triggered by a HTTP DELETE. In metal³ this happens in the Delete() call.
These two completely unrelated and yet destructive actions both have the same name. It confused me for the longest time.

apis/metal3.io/v1alpha1/baremetalhost_types.go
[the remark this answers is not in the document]
zaneb: BTW I intended "disengage" in the sense of a gearbox. We're basically putting it in neutral so you can attach another engine for a bit. It's quite a nice analogy, at least if you get a chance to explain it to the user ;D
My only concern with "detach", and why I listed it in 3rd place, is that it is more prone to conflicts with other things like networks or volumes (i.e. it risks repeating the same mistake we have made with "paused" and "unmanaged"). I still prefer it to "deregister" though.
"Deregister" makes sense if you understand the inside baseball (i.e. that there's a separate volatile data store, that existence of an entry in this datastore means that the power state will be continuously monitored, and that when this data store is cleared we silently re-register existing hosts in it.) However, to a user looking in from the outside and seeing only the API, I think it makes little sense at best, and at worst is actively misleading by implying that it resets the state back to the time before "registration".

controllers/metal3.io/host_state_machine.go
zaneb: Yes, in theory it's not required. I think we just decided as a general design principle that we didn't want to ever just start completely ignoring a resource. I don't have a memory of what specific risks we thought that might help reduce. Maybe @dhellmann does?

controllers/metal3.io/host_state_machine.go
hardys: This was discussed in an [earlier thread](https://github.com/metal3-io/baremetal-operator/pull/827#discussion_r599761953) and I'm not sure we achieved consensus re whether the Delete() is a sufficiently inexpensive operation to skip the ProvisioningCapacity check.

Personally I think it's probably fine - just deleting the node from the DB is a *way* less expensive operation than deprovisioning, and the main cause of failures is stuff like PXE timeouts not DB contention.

I think @s3rj1k and @andfasano were more in favor of adding such a check, sounds like @zaneb and I aren't so :man_shrugging: - perhaps we leave it out for now, and it can be easily added if we ever see evidence of DB contention etc?

apis/metal3.io/v1alpha1/baremetalhost_types.go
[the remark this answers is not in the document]
hardys: Ugh.  Naming is hard :disappointed: 

@dhellmann @maelk @fmuyassarov - any preference before I rename this again?

FWIW my vote is for `detached` - I take the point about potential overlap with networking/storage things @zaneb but in the context of the BMH resource that's probably not very likely, right?

controllers/metal3.io/host_state_machine.go
andfasano: > Ah, I see that you are another victim of terrible naming in ironic.
> A reasonable person might think that nodes.Deleting is the state that the node goes into when you delete it. It isn't.
> 
> nodes.Deleting is what happens when you deprovision the node, i.e. a Nova delete in OpenStack. It's triggered by a HTTP PATCH setting targetState=deleting. In metal³ this happens in the Deprovision() call.
> When you actually delete a node it just disappears from the database and that's the end of it. It's triggered by a HTTP DELETE. In metal³ this happens in the Delete() call.
> These two completely unrelated and yet destructive actions both have the same name. It confused me for the longest time.

Doh! Never mind then :). @hardys no need to add the check then

s3rj1k: @hardys That depends on amount of hosts, connection and memory limits in mysql db, on big clusters this will be an issue.

controllers/metal3.io/host_state_machine.go
hardys: Switched to unmanagedRetryDelay in the latest version

controllers/metal3.io/host_state_machine.go
```diff
@@ -226,6 +230,34 @@ func (hsm *hostStateMachine) checkInitiateDelete() bool {
...
+				if err != nil {
+					return actionError{errors.Wrap(err, "failed to deregister deregistered host")}
+				}
```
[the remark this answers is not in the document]
hardys: Added check for ErrorMessage and Dirty flag in the latest version

controllers/metal3.io/host_state_machine.go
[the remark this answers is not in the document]
hardys: > @hardys That depends on amount of hosts, connection and memory limits in mysql db, on big clusters this will be an issue. One would have to tweak mysql config.

Right, but that's not what the provisioningCapacity slots are for AFAIK - those exist to prevent "thundering herd" issues where you try to simultaneously provision $many hosts, which results multiple ironic API calls and related conductor processing, DoSing the PXE infrastructure, saturating the physical network fabric, etc

In this case the annotation results in exactly one API call to Ironic, and it sounds like the consensus is to just make the call - if we see DB/ironic-api bottlenecks at scale we could revisit the discussion and/or look at ways to improve the Ironic/DB deployment configuration

apis/metal3.io/v1alpha1/baremetalhost_types.go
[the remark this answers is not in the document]
hardys: Changed to `detached` in the latest revision

controllers/metal3.io/action_result.go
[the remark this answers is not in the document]
hardys: switched to unmanagedRetryDelay instead

controllers/metal3.io/host_state_machine.go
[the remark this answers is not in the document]
hardys: Moved this logic into `BareMetalHostReconciler`

hardys: Ok sounds like consensus is to skip the check, we can revisit in future if needed

apis/metal3.io/v1alpha1/baremetalhost_types.go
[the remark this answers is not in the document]
zaneb: > that's probably not very likely, right?

:rofl:

metal3-io/metal3-docs#132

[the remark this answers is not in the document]
hardys: > > that's probably not very likely, right?
> 
> rofl
> 
> [metal3-io/metal3-docs#132](https://github.com/metal3-io/metal3-docs/pull/132)

Thanks I missed that - and it adds a ports API to the BMH spec - I'm not sure that's a good idea, I was thinking any such additions would have new CRDs, and thus not overlap with BMH things, but it seems I've been wildly optimistic there.

controllers/metal3.io/baremetalhost_controller.go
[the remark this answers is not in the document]
hardys: Done in the latest version

controllers/metal3.io/host_state_machine.go
hardys: I think in practice this won't be a problem, the only use-case I'm aware of is to detach Provisioned BMH resources on the parent cluster, and as you say provided we don't move the hosts to a state where Ironic asserts power state things will work fine even if in a corner-case scenario IPA does run and discover a host.

If we need to enforce this, it may be better to only allow detaching in [Externally]Provisioned - another option now that the implementation uses unmanagedRetryDelay is we could call `prov.Detach()` unconditionally, so any hanging Ironic resources would be eventually removed?

hardys:
@andfasano @zaneb this is ready for another review pass - PTAL thanks!

If we're OK with the latest name I'll update metal3-io/metal3-docs#168 to match?

docs/api.md
```diff
@@ -453,3 +453,11 @@ put any value on this annotation **other than `metal3.io/capm3`**. Please make
...
 not `metal3.io/capm3`, but another value that you have provided**. Removing the
 annotation will enable the reconciliation again.
+
+## Detaching hosts
+
+It is possible to pause the management of a BareMetalHost object by adding
+an annotation `baremetalhost.metal3.io/detached`. This removes the host from
+the provisioner, and prevents any further state changes, but it does not
+fully pause reconciliation (so the status field can be updated, and the
+OperationalStatus field will reflect the `detached` status).
```
sadasu, line 463: "possible to pause" and "does not fully pause" are a little confusing.
Looking at the code, maybe we can reword to say that we can "watch the state of the BMH without being able to directly manage it."

pkg/provisioner/ironic/ironic.go
```diff
@@ -1682,6 +1682,16 @@ func (p *ironicProvisioner) Delete() (result provisioner.Result, err error) {
...
 }
 
+// Detach removes the host from the provisioning system.
+// Similar to Delete, but ensures non-interruptive behavior
+// for the target system.  It may be called multiple times,
+// and should return true for its dirty  flag until the
+// deletion operation is completed.
+func (p *ironicProvisioner) Detach() (result provisioner.Result, err error) {
+	// Currently the same behavior as Delete()
+	return p.Delete()
```
sadasu, line 1692: https://github.com/metal3-io/baremetal-operator/issues/410 wants us to power off the node as part of Delete(). Fix is being worked on in https://github.com/metal3-io/baremetal-operator/pull/816, and a hardPowerOff() performed during a Delete(). I don't think we want that behavior for Detach().

zaneb: The PowerOff() function now has an argument to force a hard shutdown, so we could just put the power off logic in the controller rather than the provisioner now. That's probably the right thing to do anyway, since wanting to power off is not provisioner-specific.

Failing that we can flip the implementations (so that Delete() powers off and then calls Detach()) but let's deal with that when we come to it. This looks fine for now.

controllers/metal3.io/host_state_machine.go
```diff
@@ -226,6 +230,24 @@ func (hsm *hostStateMachine) checkInitiateDelete() bool {
 	return true
 }
 
+func (hsm *hostStateMachine) checkDetachedHost(info *reconcileInfo) (result actionResult) {
+	// If the detached annotation is set we remove any host from the
+	// provisioner and take no further action
+	// Note this doesn't change the current state, only the OperationalStatus
```
zaneb, line 236: We should ignore the annotation if the provisioning state is Unmanaged, to avoid changing the operational state from "discovered" to "detached". (Or, alternatively, change it back when the annotation is removed below.)

Where did we get to with the idea of only applying this in Provisioned/ExternallyProvisioned states?

controllers/metal3.io/host_state_machine.go
```diff
@@ -226,6 +230,24 @@ func (hsm *hostStateMachine) checkInitiateDelete() bool {
...
+	annotations := hsm.Host.GetAnnotations()
+	if annotations != nil {
+		if _, ok := annotations[metal3v1alpha1.DetachedAnnotation]; ok {
+			return hsm.Reconciler.actionDetached(hsm.Provisioner, info)
+		}
+	}
+	if info.host.OperationalStatus() == metal3v1alpha1.OperationalStatusDetached {
+		info.host.SetOperationalStatus(metal3v1alpha1.OperationalStatusOK)
```
zaneb, line 244:
```
newStatus := metal3v1alpha1.OperationalStatusOK
if info.host.Status.ErrorType != "" {
    newStatus = metal3v1alpha1.OperationalStatusError
}
info.host.SetOperationalStatus(newStatus)
```

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -510,6 +510,26 @@ func (r *BareMetalHostReconciler) actionUnmanaged(prov provisioner.Provisioner,
 	return actionContinue{unmanagedRetryDelay}
 }
 
+func (r *BareMetalHostReconciler) actionDetached(prov provisioner.Provisioner, info *reconcileInfo) actionResult {
+	if info.host.OperationalStatus() != metal3v1alpha1.OperationalStatusDetached {
```
zaneb, line 514: I don't actually see any harm in dropping this condition. We should always be able to call Detach() idempotently.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -510,6 +510,26 @@ func (r *BareMetalHostReconciler) actionUnmanaged(prov provisioner.Provisioner,
...
+		provResult, err := prov.Detach()
+		if err != nil {
+			return actionError{errors.Wrap(err, "failed to detach")}
+		}
+		if provResult.ErrorMessage != "" {
+			return recordActionFailure(info, metal3v1alpha1.DetachError, provResult.ErrorMessage)
```
zaneb, line 520: We need to clear this error again somewhere - both the ErrorType/ErrorMessage the next time we call Detach without getting an ErrorMessage back, and the ErrorCount once we have successfully detached.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -510,6 +510,26 @@ func (r *BareMetalHostReconciler) actionUnmanaged(prov provisioner.Provisioner,
...
+		}
+		if provResult.Dirty {
+			return actionContinue{provResult.RequeueAfter}
+		}
+		info.log.Info("host is detached, removed from provisioner")
+	}
+	if info.host.SetOperationalStatus(metal3v1alpha1.OperationalStatusDetached) {
+		return actionUpdate{}
```
zaneb, line 528: This will requeue immediately, which is fine but we don't actually need to do that. We could return `actionUpdate{actionContinue{unmanagedRetryDelay}}`.

controllers/metal3.io/host_state_machine.go
hardys: > Where did we get to with the idea of only applying this in Provisioned/ExternallyProvisioned states?

I think that would be fine for the use-case driving this change, we'd need to ensure it's also possible to delete the BMH from those states without triggering deprovisioning though

docs/api.md
hardys: "possible to pause" is a copy/paste typo, and I like your suggestion @sadasu thanks!

pkg/provisioner/ironic/ironic.go
hardys: Right, since we've got this new provisioner API it enables the behavior to diverge in future, but right now the behavior is the same so I just call Delete()

controllers/metal3.io/host_state_machine.go
hardys: The latest version restricts the API to  [Externally]Provisioned)

controllers/metal3.io/host_state_machine.go
hardys: The latest version restricts the API to only  [Externally]Provisioned)

controllers/metal3.io/host_state_machine.go
hardys: Thanks!

controllers/metal3.io/baremetalhost_controller.go
hardys: Done!

controllers/metal3.io/baremetalhost_controller.go
hardys: Done!

controllers/metal3.io/baremetalhost_controller.go
hardys: Fixed in the latest version

hardys:
Thanks for the review @zaneb I think I resolved all your comments - I also in discussion with @andfasano noticed that the "delete without deprovision" aspect wasn't working, so that's now resolved too and I added more tests to cover.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -522,6 +528,24 @@ func (r *BareMetalHostReconciler) actionUnmanaged(prov provisioner.Provisioner,
...
+		return actionError{errors.Wrap(err, "failed to detach")}
+	}
+	if provResult.ErrorMessage != "" {
+		return recordActionFailure(info, metal3v1alpha1.DetachError, provResult.ErrorMessage)
+	}
+	if provResult.Dirty {
+		return actionContinue{provResult.RequeueAfter}
+	}
+	if clearErrorSetStatus(info.host, metal3v1alpha1.OperationalStatusDetached) {
+		info.log.Info("host is detached, removed from provisioner")
```
andfasano, line 543: I guess this could be a good point to reset the ErrorCount to 0

hardys: Thanks @andfasano - the only place I see `info.host.Status.ErrorCount = 0` is in `manageHostPower` - do we expect that to reset the counter for all non-detached cases?  I incorrectly assumed it was handled by the existing `clearError` code

hardys: Resolved in the latest version, thanks!

andfasano: Originally it was defined there, but then the `ErrorCount` resetting has been mostly moved into the `host_state_machine` (usually on state transitions as [here](https://github.com/metal3-io/baremetal-operator/blob/b7cc13f55e3e8db9fa0318b71b4c64b6d4d14fe0/controllers/metal3.io/host_state_machine.go#L300) or [here](https://github.com/metal3-io/baremetal-operator/blob/b7cc13f55e3e8db9fa0318b71b4c64b6d4d14fe0/controllers/metal3.io/host_state_machine.go#L308) for example).
In this specific case we are exercising an action - `Detach` - that could fail and increase the error count, so as long as the resource remains detached (with the annotation set and state `Provisioned` or `ExternallyProvisioned`),_ a previously failed - but now recovered - deletion can be cleared here when the `operationalStatus` becomes effectively `OperationalStatusDetached`.

I just realized that there could be another scenario to be handled regarding the error management IIUC the annotation handling (@hardys can you please confirm if it sound reasonable?): when the `actionDetached` is triggered, the `Detach` fails (so the `ErrorCount` is increased) but then the annotation is removed (by a human or any other external operator) before recovering. The resource is still marked to be in error (`ErrorCount` increased, `ErrorType` set to `DetachError` and `ErrorMessage` no empty), but the flow at this point will go back as usual in the `actionManageSteadyState`.

hardys: @andfasano right, so in that case we'll call `manage[Externally]Provisioned->actionManageSteadyState->manageHostPower` which should reset the `ErrorCount` right?

I think reproducing such a scenario will be difficult, but I can try to add a test if you think it's needed?

andfasano: Yes, but the main issue I see is that potentially other problems could pop out before reaching that point (ie maybe the `Adopt` fails), and while the `ErrorType` and `ErrorMessage` will be overwritten the `ErrorCount` will get increased from its current non-zero value (non-zero due another unrelated error).

I'd say that for the sake of clarity it could be cleaner to detect the "interruption" of a in-progress detach operation (annotation not set but `ErrorType` set to `DetachError`) [here](https://github.com/metal3-io/baremetal-operator/pull/827/files#diff-eeed3703d04e4c23a7d7af8cd0b7931b6b7990f23d826c49bdbc31c5f0a50291R251) and perform the cleanup, since it looks like the detaching has been abandoned.

Having a test to capture it would be really great.

hardys: Ok I added a commit which adds a test for the two error paths, e.g:

  * Detach() fails, then succeeds
  * Detach() fails, then the annotation is removed

In the latter case I reset the ErrorCount, although I guess it's pretty likely whatever caused the Detach() to fail will cause some subsequent error.

controllers/metal3.io/host_state_machine.go
```diff
@@ -227,6 +235,37 @@ func (hsm *hostStateMachine) checkInitiateDelete() bool {
...
+	}
+	if info.host.Status.ErrorType == metal3v1alpha1.DetachError {
+		clearError(info.host)
+		hsm.Host.Status.ErrorCount = 0
+		info.log.Info("removed detach error")
+		return actionUpdate{}
+	}
+	if info.host.OperationalStatus() == metal3v1alpha1.OperationalStatusDetached {
+		newStatus := metal3v1alpha1.OperationalStatusOK
+		if info.host.Status.ErrorType != "" {
```
hardys, line 259: @zaneb with the change to explicitly handle DetachError as suggested by @andfasano I'm not sure this code is reachable?

In the case where we detach a host that's e.g `ProvisionedRegistrationError` either the Detach will succeed (so the ErrorType/Message gets removed), or fail (in which case this ends up DetachError)

I couldn't think of a way to reach this in the test without e.g forcing OperationalStatusDetached without any annotation, but if you can think of a scenario where we'll hit this in practice I can add another test scenario.

controllers/metal3.io/host_state_machine.go
```diff
@@ -227,6 +235,37 @@ func (hsm *hostStateMachine) checkInitiateDelete() bool {
...
+	annotations := hsm.Host.GetAnnotations()
+	if annotations != nil {
+		if _, ok := annotations[metal3v1alpha1.DetachedAnnotation]; ok {
+			// Only allow detaching hosts in Provisioned/ExternallyProvisioned states
+			if info.host.Status.Provisioning.State == metal3v1alpha1.StateProvisioned || info.host.Status.Provisioning.State == metal3v1alpha1.StateExternallyProvisioned {
+				return hsm.Reconciler.actionDetached(hsm.Provisioner, info)
+			}
+		}
+	}
+	if info.host.Status.ErrorType == metal3v1alpha1.DetachError {
```
sadasu, line 251: So, this is the case where the "detached" annotation was present, the detach operation resulted in a DetachError but then the annotation was removed? If that is the case,  is there a need to also set the OperationalStatus to some thing like metal3v1alpha1.OperationalStatusOK?

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -539,6 +545,26 @@ func getCurrentImage(host *metal3v1alpha1.BareMetalHost) *metal3v1alpha1.Image {
...
+// actionDetached() detaches the host from the Provisioner
+func (r *BareMetalHostReconciler) actionDetached(prov provisioner.Provisioner, info *reconcileInfo) actionResult {
+	provResult, err := prov.Detach()
+	if err != nil {
+		return actionError{errors.Wrap(err, "failed to detach")}
+	}
+	if provResult.ErrorMessage != "" {
+		return recordActionFailure(info, metal3v1alpha1.DetachError, provResult.ErrorMessage)
+	}
+	if provResult.Dirty {
```
zaneb, line 557: If Dirty is true then we want to clear any DetachError (but not set the operational status or reset the ErrorCount).

controllers/metal3.io/host_state_machine.go
zaneb: It seems to me that if a detach succeeds (or is in the process of succeeding) then we should only clear the error if it is a DetachError.
For other operations we clear any error if any subsequent thing succeeds, on the principal that if something failed it was probably because we couldn't contact the hardware, so if it's back then time to start with a clean slate. But that doesn't really apply to Detach, which doesn't talk to the hardware (and also currently can't fail, although it's good that we're building all the integration so that it would work in theory).

controllers/metal3.io/host_state_machine.go
```diff
@@ -227,6 +235,37 @@ func (hsm *hostStateMachine) checkInitiateDelete() bool {
...
 
+func (hsm *hostStateMachine) checkDetachedHost(info *reconcileInfo) (result actionResult) {
+	// If the detached annotation is set we remove any host from the
+	// provisioner and take no further action
+	// Note this doesn't change the current state, only the OperationalStatus
+	annotations := hsm.Host.GetAnnotations()
+	if annotations != nil {
+		if _, ok := annotations[metal3v1alpha1.DetachedAnnotation]; ok {
+			// Only allow detaching hosts in Provisioned/ExternallyProvisioned states
+			if info.host.Status.Provisioning.State == metal3v1alpha1.StateProvisioned || info.host.Status.Provisioning.State == metal3v1alpha1.StateExternallyProvisioned {
```
zaneb, line 246: nit: I think a switch statement would be much clearer here

controllers/metal3.io/host_state_machine.go
zaneb: I just happened to be reading the source for `clearError()` and it does do that.

controllers/metal3.io/host_state_machine.go
```diff
@@ -227,6 +235,37 @@ func (hsm *hostStateMachine) checkInitiateDelete() bool {
 	return true
 }
 
+func (hsm *hostStateMachine) checkDetachedHost(info *reconcileInfo) (result actionResult) {
+	// If the detached annotation is set we remove any host from the
+	// provisioner and take no further action
+	// Note this doesn't change the current state, only the OperationalStatus
+	annotations := hsm.Host.GetAnnotations()
+	if annotations != nil {
+		if _, ok := annotations[metal3v1alpha1.DetachedAnnotation]; ok {
```
zaneb, line 244: nit: a `hasDetachedAnnotation()` convenience function wouldn't hurt.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -539,6 +545,26 @@ func getCurrentImage(host *metal3v1alpha1.BareMetalHost) *metal3v1alpha1.Image {
 	return nil
 }
 
+// actionDetached() detaches the host from the Provisioner
+func (r *BareMetalHostReconciler) actionDetached(prov provisioner.Provisioner, info *reconcileInfo) actionResult {
```
zaneb, line 549: nit: I think of this as the counterpart to `registerHost()` more than an equivalent to the other methods starting with `action`, which are all called from the state machines `handle` methods. So maybe `detachHost()` would be a good name.

controllers/metal3.io/host_state_machine.go
hardys: Yeah that's already handled by `clearError()`

controllers/metal3.io/host_state_machine.go
hardys: My point was that if we reach the point where `info.host.OperationalStatus() == metal3v1alpha1.OperationalStatusDetached` I'm not sure there will ever be any ErrorType except empty, because we clear errors via `clearErrorSetStatus` at the end of `actionDetached()`

Leaving this test for ErrorType is harmless, but I couldn't get test coverage without doing things which I don't think will ever actually happen, so I wanted to query if this logic is actually correct.

controllers/metal3.io/host_state_machine_test.go
```diff
@@ -146,6 +146,278 @@ func TestProvisioningCapacity(t *testing.T) {
 	}
 }
 
+func TestDetach(t *testing.T) {
+	testCases := []struct {
+		Scenario                  string
+		Host                      *metal3v1alpha1.BareMetalHost
+		Annotations               map[string]string
+		ExpectedDetach            bool
```
andfasano, line 154: I think this is a little bit misleading for how it has been defined (see also my other comment in the mock). Moreover, I'd suggest to remove it, as I think it should be sufficient to verify the host operationalStatus/state and action result

controllers/metal3.io/host_state_machine_test.go
```diff
@@ -146,6 +146,278 @@ func TestProvisioningCapacity(t *testing.T) {
 	}
 }
 
+func TestDetach(t *testing.T) {
+	testCases := []struct {
+		Scenario                  string
+		Host                      *metal3v1alpha1.BareMetalHost
+		Annotations               map[string]string
```
andfasano, line 153: nit: to remove some noise from the case definition, it could be simply something like a `hasDetachAnnotation`

controllers/metal3.io/host_state_machine_test.go
```diff
@@ -146,6 +146,278 @@ func TestProvisioningCapacity(t *testing.T) {
...
+			assert.Equal(t, tc.ExpectedState, info.host.Status.Provisioning.State)
+		})
+	}
+}
+
+func TestDetachError(t *testing.T) {
+	testCases := []struct {
+		Scenario                  string
+		Host                      *metal3v1alpha1.BareMetalHost
+		Annotations               map[string]string
```
andfasano, line 367: Looks unused

controllers/metal3.io/host_state_machine_test.go
```diff
@@ -649,6 +926,14 @@ func (m *mockProvisioner) Delete() (result provisioner.Result, err error) {
 	return m.getNextResultByMethod("Delete"), err
 }
 
+func (m *mockProvisioner) Detach() (result provisioner.Result, err error) {
+	res := m.getNextResultByMethod("Detach")
+	if res.ErrorMessage == "" {
+		m.isDetached = true
```
andfasano, line 932: That looks rather unusual, and as previously suggested I think it could be removed.
In case it would be necessary to verify that a provisioner method has been invoked, the mock could have a generic mechanism to record the calls (personally not a big fan of such approach, but sometimes it could be required)

controllers/metal3.io/host_state_machine_test.go
hardys: Proving the OperationalStatus is set isn't exactly the same as proving the provisioner.Detach() was actually called, but I can remove this if we don't think that coverage is important.

controllers/metal3.io/host_state_machine_test.go
hardys: Currently we just assume the mockProvisioner methods have been called without checking, my goal here was to prove the Detach() had actually been called, but given the scope of host_state_machine_test.go is the HSM primarily I can remove it.

controllers/metal3.io/host_state_machine_test.go
hardys: Done

controllers/metal3.io/host_state_machine_test.go
hardys: Thanks, yes copy/paste error - removed in the latest version

controllers/metal3.io/host_state_machine_test.go
hardys: I replaced this by a generic `calledNoError` method on the mockProvisioner - personally I think there's value in proving the Provisioner method got called (and also that it didn't in all the states where we don't expect `Detach()` to happen)

controllers/metal3.io/host_state_machine_test.go
hardys: I converted this to a generic method that proves if the provisioner method got called or not

controllers/metal3.io/host_state_machine_test.go
```diff
@@ -146,6 +146,278 @@ func TestProvisioningCapacity(t *testing.T) {
...
 }
 
+func TestDetach(t *testing.T) {
+	testCases := []struct {
+		Scenario                  string
+		Host                      *metal3v1alpha1.BareMetalHost
+		Annotations               map[string]string
+		ExpectedDetach            bool
+		ExpectedDirty             bool
+		ExpectedError             bool
```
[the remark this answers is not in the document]
hardys: Now removed, this case is handled via `TestDetachError` now (previously the results were conditional on `ExpectedError` not only the provisioner configuration)

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -539,6 +545,26 @@ func getCurrentImage(host *metal3v1alpha1.BareMetalHost) *metal3v1alpha1.Image {
...
+	provResult, err := prov.Detach()
+	if err != nil {
+		return actionError{errors.Wrap(err, "failed to detach")}
+	}
+	if provResult.ErrorMessage != "" {
+		return recordActionFailure(info, metal3v1alpha1.DetachError, provResult.ErrorMessage)
```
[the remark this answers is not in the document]
hardys: Thanks, done!

controllers/metal3.io/baremetalhost_controller.go
hardys: Thanks, renamed.

controllers/metal3.io/host_state_machine.go
hardys: Done, I added a local function for now, but this seems to be a recurring pattern, so perhaps we can refactor to add a shared host.HasAnnotation() in future (I didn't attempt it here because it means moving some annotation constants which feels outside the scope of this series).

controllers/metal3.io/host_state_machine.go
hardys: Yeah, done!

controllers/metal3.io/baremetalhost_controller.go
hardys: Ok adjusted to call clearError like in similar provResult.Dirty cases.

andfasano:
/lgtm
/hold
to allow @zaneb a final pass

controllers/metal3.io/host_state_machine.go
zaneb: Right, you're correct that as it stands I don't think we'll ever get here. What I'm suggesting is that we probably *shouldn't* clear any non-`DetachError` (which I think can only be a power management error, since we're only doing this in the steady states) at the end of (or during) the successful detach. If we only clear `DetachError`s in `actionDetached()` then this code will become reachable again.

zaneb (commented):
I have a couple of comments inline about the error handling. It would be nice if we could address those (particularly since it means we can undo the refactoring of `clearError()`, but given that both the Detach errors and Power Management errors are an entirely theoretical concept today, I also don't see any obstacle to merging this and fixing it in a follow-up.
/approve

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -539,6 +546,29 @@ func getCurrentImage(host *metal3v1alpha1.BareMetalHost) *metal3v1alpha1.Image {
...
+	}
+	if provResult.Dirty {
+		if clearError(info.host) {
+			return actionUpdate{actionContinue{provResult.RequeueAfter}}
+		}
+		return actionContinue{provResult.RequeueAfter}
+	}
+	if clearErrorSetStatus(info.host, metal3v1alpha1.OperationalStatusDetached) {
```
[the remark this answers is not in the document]
hardys: Thanks, added in the latest revision

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -539,6 +546,29 @@ func getCurrentImage(host *metal3v1alpha1.BareMetalHost) *metal3v1alpha1.Image {
...
+func (r *BareMetalHostReconciler) detachHost(prov provisioner.Provisioner, info *reconcileInfo) actionResult {
+	provResult, err := prov.Detach()
+	if err != nil {
+		return actionError{errors.Wrap(err, "failed to detach")}
+	}
+	if provResult.ErrorMessage != "" {
+		return recordActionFailure(info, metal3v1alpha1.DetachError, provResult.ErrorMessage)
+	}
+	if provResult.Dirty {
+		if clearError(info.host) {
```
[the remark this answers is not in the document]
hardys: Done

PR #829, opened 2021-03-23
 Change default of RAIDInterface

iurygregory:
Instead of using `""` we will be using `no-raid`.
The support for RAID was considering  RAIDInterface defaul value as `""`
to generate the clean steps, this is causing errors in vmedia
deployments with idrac because it uses `no-raid`:

"Node failed to start the first cleaning step. Error: node does not support
this clean step: {'interface': 'raid', 'step': 'delete_configuration'}"

### DIFF
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -1234,7 +1234,7 @@ func (p *ironicProvisioner) ironicHasSameImage(ironicNode *nodes.Node) (sameImag
 
 func (p *ironicProvisioner) buildManualCleaningSteps() (cleanSteps []nodes.CleanStep, err error) {
 	// Build raid clean steps
-	if p.bmcAccess.RAIDInterface() != "" {
+	if p.bmcAccess.RAIDInterface() != "no-raid" {
 		cleanSteps = append(cleanSteps, BuildRAIDCleanSteps(p.host.Status.Provisioning.RAID)...)
 	} else if p.host.Status.Provisioning.RAID != nil {
 		return nil, fmt.Errorf("RAID settings are defined, but the node's driver %s does not support RAID", p.bmcAccess.Driver())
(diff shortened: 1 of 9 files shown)


pkg/provisioner/ironic/ironic.go
```diff
@@ -1234,7 +1234,7 @@ func (p *ironicProvisioner) ironicHasSameImage(ironicNode *nodes.Node) (sameImag
 
 func (p *ironicProvisioner) buildManualCleaningSteps() (cleanSteps []nodes.CleanStep, err error) {
 	// Build raid clean steps
-	if p.bmcAccess.RAIDInterface() != "" {
+	if p.bmcAccess.RAIDInterface() != "" && p.bmcAccess.RAIDInterface() != "no-raid" {
```
zaneb, line 1237: TBH I don't think it's correct to ever assume that the default (empty string) means `no-raid`, since that's only true in the case of one driver (that doesn't return "" - see #319). I think a real fix would be to explicitly set the RAID interface in every BMC driver and only check for `no-raid` here.

iurygregory: Sure I can go and change the bmcs with "" to `no-raid` and update this here =)

dhellmann: Do we want to look for the string `no-raid` or do we want a bmcAccess method like `RAIDSupported()` that returns a bool?

iurygregory: @dhellmann I would say it's fair to look at `no-raid` since this would say raid it's not supported, so having `no-raid` as default instead of `""` would be better

iurygregory: Done

pkg/provisioner/ironic/testbmc/testbmc.go
```diff
@@ -75,7 +75,7 @@ func (a *testAccessDetails) PowerInterface() string {
 }
 
 func (a *testAccessDetails) RAIDInterface() string {
-	return ""
+	return "no-raid"
```
andfasano, line 78: What about using a constant for such value?

dtantsur:
/approve

I agree re adding a constant, but otherwise the change is correct.

pkg/provisioner/ironic/testbmc/testbmc.go
iurygregory: we can probably think about creating possible values for raid as a constant (In a different PR I would say, since this is blocking deployments with virtual media)

zaneb:
+1 for a constant but since the world is broken let's fix that first.
/lgtm

PR #830, opened 2021-03-24
Fix the bug that the inspector reverse proxy still runs even when Ironic TLS is disabled

namnx228:
Currently, the httpd reverse proxy used by ironic-inspector is launched even when we don't enable ironic TLS. This adds unnecessary resource consumption. Also, it creates a conflict as both Ironic-inspector and the httpd-reverse-proxy try to open port 5050, leading to a crash of the httpd-reverse-proxy. 
```yaml
httpd-reverse-proxy:
  Container ID: docker://72b71a7e6cbce69a4c7c7543e8431a39e58d3ef6f865260baa59923678416ad4
  Image:     192.168.111.1:5000/localimages/ironic-inspector
  Image ID:   docker-pullable://192.168.111.1:5000/localimages/ironic-inspector@sha256:df05e2eec8885b68d05179aadf3d9ef7bbc5dda423e2b0b31f1feaf32907b76e
  Port:     <none>
  Host Port:   <none>
  Command:
   /bin/runhttpd
  State:     Waiting
  Reason:    CrashLoopBackOff
  Last State:   Terminated
   Reason:    Completed
   Exit Code:  0
   Started:   Mon, 22 Mar 2021 07:01:50 +0000
   Finished:   Mon, 22 Mar 2021 07:01:50 +0000
  Ready:     False
  Restart Count: 779
  Environment Variables from:
   ironic-bmo-configmap-bkkmd9cb78 ConfigMap Optional: false
  Environment:            <none>
  Mounts:
   /var/run/secrets/kubernetes.io/serviceaccount from default-token-cpt2d (ro)
```
The fix only allows httpd-reverse-proxy to be launched when Ironic TLS is enabled.


ironic-deployment/default/ironic_bmo_configmap.env
```diff
@@ -8,4 +8,4 @@ IRONIC_INSPECTOR_ENDPOINT=http://172.22.0.2:5050/v1/
 CACHEURL=http://172.22.0.1/images
 IRONIC_FAST_TRACK=true
 IRONIC_KERNEL_PARAMS=console=ttyS0
-INSPECTOR_REVERSE_PROXY_SETUP=false
+INSPECTOR_REVERSE_PROXY_SETUP=true
```
maelk, line 11: why would this be true by default ? I would rather expect it to be false, and set to true via kustomization / replacement if deploying TLS. I know that the reverse proxy would not be deployed anyways, but having it set to true might be confusing in a non-reverse proxy scenario

tools/deploy.sh
```diff
@@ -210,6 +210,18 @@ fi
...
+    then
+      IRONIC_BMO_CONFIGMAP="${SCRIPTDIR}/ironic-deployment/keepalived/ironic_bmo_configmap.env"
+    else
+      IRONIC_BMO_CONFIGMAP="${SCRIPTDIR}/ironic-deployment/default/ironic_bmo_configmap.env"
+    fi
+    if grep "INSPECTOR_REVERSE_PROXY_SETUP" "${IRONIC_BMO_CONFIGMAP}" > /dev/null
+    then
+      sed "s/\(INSPECTOR_REVERSE_PROXY_SETUP\).*/\1=${DEPLOY_TLS}/" -i "${IRONIC_BMO_CONFIGMAP}"
+    else
+      echo "INSPECTOR_REVERSE_PROXY_SETUP=${DEPLOY_TLS}" >> "${IRONIC_BMO_CONFIGMAP}"
```
maelk, line 223: you should add something to revert those changes. at the end of this script we try to restore the state of the repo. Another solution would be to copy the whole thing to a temporary folder and delete it afterwards.

ironic-deployment/default/ironic_bmo_configmap.env
namnx228: Fixed! Thanks

tools/deploy.sh
namnx228: Fixed! Thanks

tools/deploy.sh
```diff
@@ -210,8 +210,22 @@ fi
 
 if [ "${DEPLOY_IRONIC}" == "true" ]; then
     pushd "${SCRIPTDIR}"
+    if [[ "${DEPLOY_KEEPALIVED}" == "true" ]]
```
elfosardo, line 213: nit: preferred form would be placing then on the same line as the if, see for example L211

tools/deploy.sh
```diff
@@ -210,8 +210,22 @@ fi
...
+    then
+      IRONIC_BMO_CONFIGMAP="${SCRIPTDIR}/ironic-deployment/keepalived/ironic_bmo_configmap.env"
+    else
+      IRONIC_BMO_CONFIGMAP="${SCRIPTDIR}/ironic-deployment/default/ironic_bmo_configmap.env"
+    fi
+    cp "${IRONIC_BMO_CONFIGMAP}" /tmp/ironic_bmo_configmap.env
+    if grep "INSPECTOR_REVERSE_PROXY_SETUP" "${IRONIC_BMO_CONFIGMAP}" > /dev/null
```
elfosardo, line 220: ditto

namnx228: Fixed! Thanks for review

elfosardo (approved):

maelk (approved):

fmuyassarov (commented):
/lgtm

PR #833, opened 2021-03-24
Pass go tests flags via env.

s3rj1k:
To make local development more convenient this PR add ability to pass go test flags via envars directly to unit-tests that are started in docker/podman instance.

So effectively one could use something like this:
```
CONTAINER_RUNTIME=docker TEST_FLAGS="-failfast" ./hack/unit.sh
```

Setting `failfast` flag and removing verbose output is super-helpful on local dev environment when screen is small (less to scroll) and CPU/MEM resources are limited, so running all tests takes some time.

(mainly this is scroll issue actually, but could also be helpful in other cases) 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 RUN_NAMESPACE = metal3
-GO_TEST_FLAGS = $(VERBOSE)
+GO_TEST_FLAGS = $(TEST_FLAGS)
 DEBUG = --debug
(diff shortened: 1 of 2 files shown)


s3rj1k:
@dhellmann Can you please also look at this, looks very trivial, but helps enormously, thanks.

elfosardo (approved):

PR #837, opened 2021-03-29
Stop passing entire BareMetalHost to Provisioner

zaneb:
Until now we have passed (a copy of) the entire BareMetalHost data structure to the Provisioner that is created for each reconcile. This enabled the Provisioner to use any information contained in the CR at any time. On the downside, this enabled the Provisioner to use any information contained in the CR at any time. This makes it difficult to reason about what the provisioner is doing and what information it requires (an architectural pattern known as a "big ball of mud". This will prove untenable if more provisioner implementations are to be added.

This PR alters the provisioning methods so that only the data reasonably required by each operation is passed to the corresponding function as an explicit argument. Only information that is used to identify the host is passed to the provisioner (and is thus implicitly available on every provisioner call):

* A copy of the CR metadata (including the resource name and namespace, and also labels and annotations in case those are needed to trigger custom behaviour in some provisioner)
* The BMC address
* The BMC credentials
* Whether to disable SSL host validation in communicating with the BMC
* The PXE MAC address
* The Provisioner ID (as returned from ValidateManagementAccess())

Method-specific data in arguments are passed in structures so that the interface can continue to evolve without mass changes of function signatures. While some data is passed that is not strictly required by the ironic provisioner because it is foreseeable that other provisioners may need to make use of it, it is also foreseeable that future provisioners will have unforeseeable requirements for more information (certainly the ironic provisioner makes use of data in some places that might be considered unexpected). Therefore future evolution is expected, but by making it explicit we can allow the controller and the BMH CRD to evolve separately from the provisioner.


hardys:
wow big PR! lgtm other than the rebase though.

hardys:
lgtm - I guess the only issue with the merge commit is if we needed to revert a specific commit from the series but I guess we can handle that with manual resolution if necessary.

Do we want some additional eyes on this, particularly given the size of the series - @andfasano @fmuyassarov or @dhellmann perhaps?

zaneb:
It's a large patch but also a pretty straightforward refactor. I'd prefer to just merge it before more conflicts come up.

fmuyassarov (commented):
LGTM! Thanks for dividing it into smaller chunks!

PR #838, opened 2021-03-29
Fix error reporting when creating an host with invalid BMC url

andfasano:
This PR addresses the problem reported in https://github.com/metal3-io/baremetal-operator/issues/843.

Note that it completes the changes introduced by @zaneb in [beea4d0](https://github.com/metal3-io/baremetal-operator/commit/beea4d0ead807a8f19b38d538db3502ee3504b97) which already fixed the problem of deleting a BMH resource being created with a malformed BMC address (such issue originally also addressed by @kirankt in #713)

### DIFF
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -1183,15 +1183,6 @@ func (r *BareMetalHostReconciler) buildAndValidateBMCCredentials(request ctrl.Re
 		return nil, nil, &EmptyBMCAddressError{message: "Missing BMC connection detail 'Address'"}
 	}
 
-	// pass the bmc address to bmc.NewAccessDetails which will do
-	// more in-depth checking on the url to ensure it is
-	// a valid bmc address, returning a bmc.UnknownBMCTypeError
-	// if it is not conformant
-	_, err = bmc.NewAccessDetails(host.Spec.BMC.Address, host.Spec.BMC.DisableCertificateVerification)
-	if err != nil {
-		return nil, nil, err
-	}
-
 	bmcCreds = &bmc.Credentials{
 		Username: string(bmcCredsSecret.Data["username"]),
 		Password: string(bmcCredsSecret.Data["password"]),
--- a/controllers/metal3.io/baremetalhost_controller_test.go
+++ b/controllers/metal3.io/baremetalhost_controller_test.go
@@ -828,18 +828,6 @@ func TestMissingBMCParameters(t *testing.T) {
 		},
 
 		{
-			Scenario: "malformed address",
-			Secret:   newBMCCredsSecret("bmc-creds-ok", "User", "Pass"),
-			Host: newHost("invalid-bmc-address",
-				&metal3v1alpha1.BareMetalHostSpec{
-					BMC: metal3v1alpha1.BMCDetails{
-						Address:         "unknown://notAvalidIPMIURL",
-						CredentialsName: "bmc-creds-ok",
-					},
-				}),
-		},
-
-		{
 			Scenario: "missing address",
 			Secret:   newBMCCredsSecret("bmc-creds-ok", "User", "Pass"),
 			Host: newHost("missing-bmc-address",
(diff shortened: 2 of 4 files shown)


controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -216,6 +216,16 @@ func (r *BareMetalHostReconciler) Reconcile(ctx context.Context, request ctrl.Re
 	}
 	prov, err := r.ProvisionerFactory(*host, *bmcCreds, info.publishEvent)
 	if err != nil {
+
+		// Allow anyhow to proceed with the resource deletion if required
+		if !host.DeletionTimestamp.IsZero() && hostHasFinalizer(host) {
+			info.log.Info("marked to be deleted", "timestamp", host.DeletionTimestamp)
```
hardys, line 222: Do we want to do this unconditionally?  What if the user wants the opportunity to fix e.g BMC credentials so that deprovisioning can be completed without errors?

andfasano: Isn't enough the check on DeletionTimeStamp? AFAIK, if set, it means that the client requested to delete the resource, but the process cannot be completed since the finalizer hasn't been removed yet.

zaneb: @andfasano you're correct
The danger here though is that we create a way for people to short-circuit the proper delete flow (i.e. skip deprovisioning). I commented on the previous PR for this:

> It would be safer if we checked the the Provisioning state was empty (i.e. we have never been able to create a provisioner for this Host, vs. we created a Host and provisioned it and now somebody has changed the URL).

zaneb (commented):
I think this bug is actually fixed already by https://github.com/metal3-io/baremetal-operator/pull/820/commits/beea4d0ead807a8f19b38d538db3502ee3504b97 (which merged today).
If we get a dodgy BMC address we will now just record a registration error for the Host instead of failing to create a provisioner. That should mean that we are able to do deletes as normal.

andfasano:
> I think this bug is actually fixed already by [beea4d0](https://github.com/metal3-io/baremetal-operator/commit/beea4d0ead807a8f19b38d538db3502ee3504b97) (which merged today).
> If we get a dodgy BMC address we will now just record a registration error for the Host instead of failing to create a provisioner. That should mean that we are able to do deletes as normal.

Thanks @zaneb, I've modified accordingly the PR description and code to just keep the new test

controllers/metal3.io/baremetalhost_controller_test.go
```diff
@@ -1854,3 +1861,66 @@ func TestUpdateRAID(t *testing.T) {
...
+func (r *ReconcileHelper) verifyNextReconcileError(expectedState metal3v1alpha1.ProvisioningState, expectedErrorMessage string) {
+	_, err, host := tryReconcileSingleStep(r.reconciler, r.host, r.request)
+	assert.EqualError(r.t, err, expectedErrorMessage)
+	assert.Equal(r.t, expectedState, host.Status.Provisioning.State)
+}
+
+func (r *ReconcileHelper) doDeleteHost() {
+	now := metav1.Now()
+	r.host.DeletionTimestamp = &now
+	r.host.Finalizers = []string{}
```
zaneb, line 1909: Shouldn't we be testing that the controller can remove the finalizer by itself? The actual bug is that it wasn't doing that so it would be good to test it, although in practice we know that if we can get the the Deleting state it should be able to.

controllers/metal3.io/baremetalhost_controller_test.go
```diff
@@ -1854,3 +1861,66 @@ func TestUpdateRAID(t *testing.T) {
...
+	r.reconciler.Client.Update(context.Background(), r.host)
+}
+
+func TestInvalidBMHCanBeDeleted(t *testing.T) {
+	host := newDefaultHost(t)
+	host.Spec.BMC.Address = fmt.Sprintf("%s%s%s", "<", host.Spec.BMC.Address, ">")
+	r := newReconcileHelper(t, host)
+
+	r.verifyNextReconcileOk(metal3v1alpha1.StateNone)
```
zaneb, line 1918: Can't this be done by

```
tryReconcile(t, r, h, func(host *metal3v1alpha1.BareMetalHost, result reconcile.Result) bool {
    return host.Status.Provisioning.State == metal3v1alpha1.StateNone
})
```

without factoring out `tryReconcileSingleStep()`?

controllers/metal3.io/baremetalhost_controller_test.go
andfasano: Thanks that's a good point, let me recheck it, as the DeletionTimestamp should be sufficient

controllers/metal3.io/baremetalhost_controller_test.go
andfasano: I'm looking at them as two different testing styles focusing on different elements. Here the sequence is completely unrolled to highlight each step, as I felt was important to follow the whole path (at least from the provisioning state point of view). Also this gives the opportunity to introduce in a simpler way "asynchronous" operations in the flow, like the Delete one. That's something that definitely could be achieved also with the `tryReconcile`, but still a couple of them would be needed.
I took also the opportunity to promote an (opinionated) scenario helper as an approach to push for test readability, given that it hides the complexity of the setup details and offers some verification/action methods where all the technical details have been stripped out.

zaneb: I'm not against it if it's generally useful. However, while probably fine in this instance, in general this is a brittle way to develop tests - the exact number of requeues gets hard-coded in the test, with the result that any change requires a rewrite of the tests. In most cases the thing we want to test is that the state converges to the expected one within a bounded number of requeues. So I'm not sure we want to encourage this style.
FWIW I believe there are other tests that do multiple tryReconcile()s with other operations intervening in the flow, so there is precedent for doing it that way.

zaneb:
Excellent, thanks Andrea!
/lgtm

andfasano:
(PR required a rebase, apparently not detected on github...)

PR #840, opened 2021-03-30
Don't immediately retry on Inspect fail

zaneb:
Ensure that if the ironic node enters the InspectFail state, we not only report the error but also wait until after the exponential backoff before retrying, instead of retrying immediately.

The previous attempt to fix this in 2d02462fb45dbb5dd9dd0189ce91e0c417bbab2e didn't take due to a missing return.

### DIFF
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -673,8 +673,8 @@ func (p *ironicProvisioner) InspectHardware(force bool) (result provisioner.Resu
 				p.log.Info("inspection already started")
 				result, err = operationContinuing(introspectionRequeueDelay)
 				return
-			default:
-				if nodes.ProvisionState(ironicNode.ProvisionState) == nodes.InspectFail && !force {
+			case nodes.InspectFail:
+				if !force {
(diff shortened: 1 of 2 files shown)


pkg/provisioner/ironic/ironic.go
```diff
@@ -673,16 +673,18 @@ func (p *ironicProvisioner) InspectHardware(force bool) (result provisioner.Resu
 				p.log.Info("inspection already started")
 				result, err = operationContinuing(introspectionRequeueDelay)
 				return
-			default:
-				if nodes.ProvisionState(ironicNode.ProvisionState) == nodes.InspectFail && !force {
+			case nodes.InspectFail:
+				if !force {
```
s3rj1k, line 677: How about this instead?

```
if force {
  fallthrough
}

to avoid nested if

```

zaneb: That's not allowed (`fallthrough` has to be the last statement).

s3rj1k: hmm, yea you are right, `fallthrough statement out of place`

Issue #828, opened 2021-03-23
Power management errors are never reported


zaneb:
The `PowerOn()` and `PowerOff()` method of the ironic Provisioner never return an ErrorMessage, so despite a considerable amount of accounting that is in place for power management errors, none will ever be reported. If changing the power state fails (i.e. ironic accepts our request but is unable to carry it out) we just keep silently trying forever.

To resolve this we will need to implement checking of the LastError field from ironic, along with a force flag as used by other methods such as Adopt() (see #749).

A complication is that after a soft power off failure, we want to attempt a hard power off without flagging an error. However, there's no way to tell if a failed attempt we detect was a soft or hard power off attempt. (Note that the current code flow is unnecessarily confusing: if we start a hard power off then we end up checking on its progress by calling softPowerOff(), even though that is off the table. This code should probably be refactored so that checking the state is common.) A solution would be to say that we will only attempt a soft power off on the first attempt; if it fails we will flag a power management error and once ErrorCount > 0 we will only ever force a hard power off.

dtantsur:
> after a soft power off failure, we want to attempt a hard power off without flagging an error

This sounds like something we could implement in ironic.

zaneb:
> > after a soft power off failure, we want to attempt a hard power off without flagging an error
> 
> This sounds like something we could implement in ironic.

+1, I doubt metal³ is the only potential consumer of that.

sadasu:
> > > after a soft power off failure, we want to attempt a hard power off without flagging an error
> > 
> > 
> > This sounds like something we could implement in ironic.
> 
> +1, I doubt metal³ is the only potential consumer of that.

I would think that each controller needs to make the decision for itself to proceed to the remedial action (in this case a hard power off) when soft power off isn't succeeding.

PR #841, opened 2021-03-30
Better handling of power management errors

sadasu:
Attempts to fix https://github.com/metal3-io/baremetal-operator/issues/828

This fix should have the following components based on the above bug:
1. If Ironic accepts the power state change but is unable to carry it out (expressed via LastError from ironic), that should result in PowerOn() and PowerOff() returning an ErrorMessage. 
2. Refactor PowerOff() in such a way that if soft power off fails, it proceeds to hard power off without reporting an error. It is possible that this logic is moved out of here and into Ironic itself.
3. Add a force flag to the PowerOff() so that it skips soft power off and directly proceeds to hard power off after the previous call results in an error.

### DIFF
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -1005,7 +1005,7 @@ func (r *BareMetalHostReconciler) manageHostPower(prov provisioner.Provisioner, 
 		"reboot process", desiredPowerOnState != info.host.Spec.Online)
 
 	if desiredPowerOnState {
-		provResult, err = prov.PowerOn()
+		provResult, err = prov.PowerOn(info.host.Status.ErrorType == metal3v1alpha1.PowerManagementError)
 	} else {
 		if info.host.Status.ErrorCount > 0 {
 			desiredRebootMode = metal3v1alpha1.RebootModeHard
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -1007,7 +1007,10 @@ func (r *BareMetalHostReconciler) manageHostPower(prov provisioner.Provisioner, 
 	if desiredPowerOnState {
 		provResult, err = prov.PowerOn()
 	} else {
-		provResult, err = prov.PowerOff(desiredRebootMode)
+		if info.host.Status.ErrorCount > 0 {
+			desiredRebootMode = metal3v1alpha1.RebootModeHard
+		}
+		provResult, err = prov.PowerOff(desiredRebootMode, info.host.Status.ErrorType == metal3v1alpha1.PowerManagementError)
 	}
 	if err != nil {
 		return actionError{errors.Wrap(err, "failed to manage power state of host")}
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -1535,16 +1535,22 @@ func (p *ironicProvisioner) PowerOn() (result provisioner.Result, err error) {
 			p.log.Info("waiting for power status to change")
 			return operationContinuing(powerRequeueDelay)
 		}
-		result, err = p.changePower(ironicNode, nodes.PowerOn)
-		switch err.(type) {
-		case nil:
-		case HostLockedError:
-		default:
-			return transientError(errors.Wrap(err, "failed to power on host"))
+		if ironicNode.LastError != "" {
+			p.log.Info("PowerOn operation failed", "msg", ironicNode.LastError)
+			return operationFailed(fmt.Sprintf("PowerOn operation failed: %s",
+				ironicNode.LastError))
 		}
-		p.publisher("PowerOn", "Host powered on")
+		if result, err = p.changePower(ironicNode, nodes.PowerOn); err == nil {
+			p.publisher("PowerOn", "Host powered on")
+			return result, nil
+		}
+		switch err.(type) {
+		case HostLockedError:
+			return retryAfterDelay(powerRequeueDelay)
+		default:
+			return transientError(errors.Wrap(err, "failed to PowerOn node"))
+		}
 	}
-
 	return result, nil
 }
 
@@ -1553,21 +1559,35 @@ func (p *ironicProvisioner) PowerOn() (result provisioner.Result, err error) {
 func (p *ironicProvisioner) PowerOff(rebootMode metal3v1alpha1.RebootMode) (result provisioner.Result, err error) {
 	p.log.Info(fmt.Sprintf("ensuring host is powered off (mode: %s)", rebootMode))
 
-	if rebootMode == metal3v1alpha1.RebootModeHard {
-		result, err = p.hardPowerOff()
-	} else {
-		result, err = p.softPowerOff()
+	if rebootMode != metal3v1alpha1.RebootModeHard && !forceHardPowerOff {
+		if result, err = p.softPowerOff(); err == nil {
+			return result, nil
+		} else {
+			switch err.(type) {
+			// In case of soft power off is unsupported or has failed,
+			// we activate hard power off.
+			case SoftPowerOffUnsupportedError, SoftPowerOffFailed:
+				p.log.Info("soft power off is unsupported or failed. Proceeding to hard power off.")
+				return operationFailed(result.ErrorMessage)
+			case HostLockedError:
+				return retryAfterDelay(powerRequeueDelay)
+			default:
+				return transientError(err)
+			}
+		}
 	}
-	if err != nil {
-		switch err.(type) {
-		// In case of soft power off is unsupported or has failed,
-		// we activate hard power off.
-		case SoftPowerOffUnsupportedError, SoftPowerOffFailed:
-			return p.hardPowerOff()
-		case HostLockedError:
-			return retryAfterDelay(powerRequeueDelay)
-		default:
-			return transientError(err)
+	if rebootMode == metal3v1alpha1.RebootModeHard || forceHardPowerOff {
+		if result, err = p.hardPowerOff(); err == nil {
+			return result, nil
+		} else {
+			switch err.(type) {
+			case HardPowerOffFailed:
+				return operationFailed(result.ErrorMessage)
+			case HostLockedError:
+				return retryAfterDelay(powerRequeueDelay)
+			default:
+				return transientError(err)
+			}
 		}
 	}
 	return result, nil
@@ -1583,6 +1603,11 @@ func (p *ironicProvisioner) hardPowerOff() (result provisioner.Result, err error
 	}
 
 	if ironicNode.PowerState != powerOff {
+		if ironicNode.LastError != "" {
+			p.log.Info("hard power off error", "msg", ironicNode.LastError)
+			result, _ = operationFailed(ironicNode.LastError)
+			return result, HardPowerOffFailed{}
+		}
 		if ironicNode.TargetPowerState == powerOff {
 			p.log.Info("waiting for power status to change")
 			return operationContinuing(powerRequeueDelay)
@@ -1621,6 +1646,8 @@ func (p *ironicProvisioner) softPowerOff() (result provisioner.Result, err error
 		// If the target state is unset while the last error is set,
 		// then the last execution of soft power off has failed.
 		if targetState == "" && ironicNode.LastError != "" {
+			p.log.Info("soft power off error", "msg", ironicNode.LastError)
+			result, _ = operationFailed(ironicNode.LastError)
 			return result, SoftPowerOffFailed{}
 		}
 		result, err = p.changePower(ironicNode, nodes.SoftPowerOff)
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -1556,10 +1556,10 @@ func (p *ironicProvisioner) PowerOn() (result provisioner.Result, err error) {
 
 // PowerOff ensures the server is powered off independently of any image
 // provisioning operation.
-func (p *ironicProvisioner) PowerOff(rebootMode metal3v1alpha1.RebootMode) (result provisioner.Result, err error) {
+func (p *ironicProvisioner) PowerOff(rebootMode metal3v1alpha1.RebootMode, force bool) (result provisioner.Result, err error) {
 	p.log.Info(fmt.Sprintf("ensuring host is powered off (mode: %s)", rebootMode))
 
-	if rebootMode != metal3v1alpha1.RebootModeHard && !forceHardPowerOff {
+	if rebootMode != metal3v1alpha1.RebootModeHard && !force {
 		if result, err = p.softPowerOff(); err == nil {
 			return result, nil
 		} else {
@@ -1575,9 +1575,9 @@ func (p *ironicProvisioner) PowerOff(rebootMode metal3v1alpha1.RebootMode) (resu
 				return transientError(err)
 			}
 		}
-	}
-	if rebootMode == metal3v1alpha1.RebootModeHard || forceHardPowerOff {
-		if result, err = p.hardPowerOff(); err == nil {
+	} else {
+		// Reboot mode is hard or force flag is set
+		if result, err = p.hardPowerOff(force); err == nil {
 			return result, nil
 		} else {
 			switch err.(type) {
@@ -1590,11 +1590,10 @@ func (p *ironicProvisioner) PowerOff(rebootMode metal3v1alpha1.RebootMode) (resu
 			}
 		}
 	}
-	return result, nil
 }
 
 // hardPowerOff sends 'power off' request to BM node and waits for the result
-func (p *ironicProvisioner) hardPowerOff() (result provisioner.Result, err error) {
+func (p *ironicProvisioner) hardPowerOff(force bool) (result provisioner.Result, err error) {
 	p.log.Info("ensuring host is powered off by \"hard power off\" command")
 
 	ironicNode, err := p.getNode()
@@ -1603,9 +1602,10 @@ func (p *ironicProvisioner) hardPowerOff() (result provisioner.Result, err error
 	}
 
 	if ironicNode.PowerState != powerOff {
-		if ironicNode.LastError != "" {
+		if ironicNode.LastError != "" && !force {
 			p.log.Info("hard power off error", "msg", ironicNode.LastError)
-			result, _ = operationFailed(ironicNode.LastError)
+			result, _ = operationFailed(fmt.Sprintf("hardPowerOff operation failed: %s",
+				ironicNode.LastError))
 			return result, HardPowerOffFailed{}
 		}
 		if ironicNode.TargetPowerState == powerOff {
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -1559,37 +1559,11 @@ func (p *ironicProvisioner) PowerOn(force bool) (result provisioner.Result, err 
 func (p *ironicProvisioner) PowerOff(rebootMode metal3v1alpha1.RebootMode, force bool) (result provisioner.Result, err error) {
 	p.log.Info(fmt.Sprintf("ensuring host is powered off (mode: %s)", rebootMode))
 
-	if rebootMode != metal3v1alpha1.RebootModeHard && !force {
-		if result, err = p.softPowerOff(); err == nil {
-			return result, nil
-		} else {
-			switch err.(type) {
-			// In case of soft power off is unsupported or has failed,
-			// we activate hard power off.
-			case SoftPowerOffUnsupportedError, SoftPowerOffFailed:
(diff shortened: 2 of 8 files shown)


pkg/provisioner/ironic/ironic.go
```diff
@@ -1743,42 +1743,67 @@ func (p *ironicProvisioner) PowerOn() (result provisioner.Result, err error) {
 
 	if ironicNode.PowerState != powerOn {
 		if ironicNode.TargetPowerState == powerOn {
-			p.log.Info("waiting for power status to change")
-			return operationContinuing(powerRequeueDelay)
+			if ironicNode.LastError != "" {
```
dtantsur, line 1746: Is it "" or nil? From API perspective it's JSON null by default.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1743,42 +1743,67 @@ func (p *ironicProvisioner) PowerOn() (result provisioner.Result, err error) {
 
 	if ironicNode.PowerState != powerOn {
 		if ironicNode.TargetPowerState == powerOn {
```
dtantsur, line 1745: Target states means something is in progress, you won't get a target power state set with last error.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1743,42 +1743,67 @@ func (p *ironicProvisioner) PowerOn() (result provisioner.Result, err error) {
...
-		result, err = p.softPowerOff()
+	softPowerOffSuccessful := true
+
+	if rebootMode != metal3v1alpha1.RebootModeHard && !forceHardPowerOff {
+		if result, err = p.softPowerOff(); err == nil {
+			return result, nil
+		} else {
+			switch err.(type) {
+			// In case of soft power off is unsupported or has failed,
+			// we activate hard power off.
```
dtantsur, line 1783: Honestly, I think this logic should be a new API flag on ironic. It's not BMO specific, and implementing it here needs tracking what action is in progress.

dtantsur (changes requested):

pkg/provisioner/ironic/ironic.go
sadasu: While that API flag is being added in Ironic, how should we handle soft power off failures in BMO?

pkg/provisioner/ironic/ironic.go
sadasu: Thanks for that. Fixed it.

pkg/provisioner/ironic/ironic.go
sadasu: I believe it is "". Also, used here https://github.com/metal3-io/baremetal-operator/blob/master/pkg/provisioner/ironic/ironic.go#L585

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -932,7 +932,7 @@ func (r *BareMetalHostReconciler) manageHostPower(prov provisioner.Provisioner,
 	if desiredPowerOnState {
 		provResult, err = prov.PowerOn()
 	} else {
-		provResult, err = prov.PowerOff(desiredRebootMode)
+		provResult, err = prov.PowerOff(desiredRebootMode, info.host.Status.ErrorType == metal3v1alpha1.PowerManagementError)
```
sadasu, line 935: Under what circumstances can we expect host.Status.ErrorType to be set in general and specifically to  metal3v1alpha1.PowerManagementError? This fix depends on Ironic setting host.Status.ErrorType to metal3v1alpha1.PowerManagementError when there is an issue while soft or hard powering off a host and that the ironicNode.LastError is also set during these failure scenarios.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1724,39 +1724,60 @@ func (p *ironicProvisioner) PowerOn() (result provisioner.Result, err error) {
 			p.log.Info("waiting for power status to change")
 			return operationContinuing(powerRequeueDelay)
 		}
-		result, err = p.changePower(ironicNode, nodes.PowerOn)
-		switch err.(type) {
-		case nil:
-		case HostLockedError:
-		default:
-			return transientError(errors.Wrap(err, "failed to power on host"))
+		if ironicNode.LastError != "" {
```
zaneb, line 1727: The problem here is that once a LastError shows up we'll always report it as the error message. and never try again. That's why we need a `force` flag to both `PowerOn` and `PowerOff`.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1724,39 +1724,60 @@ func (p *ironicProvisioner) PowerOn() (result provisioner.Result, err error) {
...
-		switch err.(type) {
-		// In case of soft power off is unsupported or has failed,
-		// we activate hard power off.
-		case SoftPowerOffUnsupportedError, SoftPowerOffFailed:
-			return p.hardPowerOff()
-		case HostLockedError:
-			return retryAfterDelay(powerRequeueDelay)
-		default:
-			return transientError(err)
+	if rebootMode == metal3v1alpha1.RebootModeHard || forceHardPowerOff {
```
zaneb, line 1769: This is just the `else` to the `if` above, isn't it?

pkg/provisioner/ironic/ironic.go
```diff
@@ -1772,6 +1793,11 @@ func (p *ironicProvisioner) hardPowerOff() (result provisioner.Result, err error
 	}
 
 	if ironicNode.PowerState != powerOff {
+		if ironicNode.LastError != "" {
```
zaneb, line 1796: This is where we really need the force flag.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1724,39 +1724,60 @@ func (p *ironicProvisioner) PowerOn() (result provisioner.Result, err error) {
...
 // provisioning operation.
-func (p *ironicProvisioner) PowerOff(rebootMode metal3v1alpha1.RebootMode) (result provisioner.Result, err error) {
+func (p *ironicProvisioner) PowerOff(rebootMode metal3v1alpha1.RebootMode, forceHardPowerOff bool) (result provisioner.Result, err error) {
 	p.log.Info(fmt.Sprintf("ensuring host is powered off (mode: %s)", rebootMode))
 
-	if rebootMode == metal3v1alpha1.RebootModeHard {
-		result, err = p.hardPowerOff()
-	} else {
-		result, err = p.softPowerOff()
+	if rebootMode != metal3v1alpha1.RebootModeHard && !forceHardPowerOff {
```
zaneb, line 1752: Given that these two parameters completely overlap in terms of their functionality, I think we should just have the `rebootMode` and push the logic into the caller (it's only called in one place anyway).

controllers/metal3.io/baremetalhost_controller.go
zaneb: It's set on line 942 and cleared on line 956. (There's actually a bug, because it should also be cleared around line 966.) So it will be set whenever an error is reported and cleared whenever we manage to do something (anything) without immediately failing.
This is the right logic for a force flag to ignore the LastError in ironic and retry. Each time a retry is started, the flag will be cleared so if it fails we'll report the error again.
However, I think it's probably the wrong logic for forcing us into the hard power off mode (which is how this param is currently being used). IIRC we discussed implementing it so that we only try a soft power off once and any retries go straight to pulling the plug. For that we want to check for a non-zero ErrorCount, because the ErrorCount doesn't get cleared until the power management operation actually succeeds.

sadasu: Thanks! That was my concern. Updated to use ErrorCount to determine the force flag. Do you want me to take care of the other bug regarding clearing the ErrorType here?

pkg/provisioner/ironic/ironic.go
sadasu: The flag gets passed on to hardPowerOff() so I found it useful to have it as a different parameter. Let me know if you still want this logic to move to the BMH controller after looking at the latest updates.

pkg/provisioner/ironic/ironic.go
sadasu: Fixed that.

pkg/provisioner/ironic/ironic.go
sadasu: Yes, it is. Added a comment to provide clarity.

pkg/provisioner/ironic/ironic.go
sadasu: Fixed it.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1820,6 +1846,8 @@ func (p *ironicProvisioner) softPowerOff() (result provisioner.Result, err error
 		// If the target state is unset while the last error is set,
 		// then the last execution of soft power off has failed.
 		if targetState == "" && ironicNode.LastError != "" {
+			p.log.Info("soft power off error", "msg", ironicNode.LastError)
+			result, _ = operationFailed(ironicNode.LastError)
 			return result, SoftPowerOffFailed{}
```
zaneb, line 1851: Returning the failure as an `error` here instead of putting it directly in `result.ErrorMessage` in inconsistent with every other method that returns (`provisioner.Result, error)`, which we then have to reverse in the caller.
Let's take this opportunity to `return operationFailed(ironicNode.LastError)`.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1782,6 +1802,12 @@ func (p *ironicProvisioner) hardPowerOff() (result provisioner.Result, err error
 	}
 
 	if ironicNode.PowerState != powerOff {
+		if ironicNode.LastError != "" && !force {
+			p.log.Info("hard power off error", "msg", ironicNode.LastError)
+			result, _ = operationFailed(fmt.Sprintf("hardPowerOff operation failed: %s",
```
zaneb, line 1807: This ends up being show to the user, so let's not include method names.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1734,46 +1734,66 @@ func (p *ironicProvisioner) PowerOn() (result provisioner.Result, err error) {
...
+		if result, err = p.softPowerOff(); err == nil {
+			return result, nil
+		} else {
+			switch err.(type) {
+			// In case of soft power off is unsupported or has failed,
+			// we activate hard power off.
+			case SoftPowerOffUnsupportedError, SoftPowerOffFailed:
+				p.log.Info("soft power off is unsupported or failed. Proceeding to hard power off.")
+				return operationFailed(result.ErrorMessage)
+			case HostLockedError:
```
zaneb, line 1772: Maybe this should be a separate commit, but I actually think now we could get rid of all these error types and just returned the `result`/`error` combination we want at the source and ditch this whole switch statement to just `return p.softPowerOff()` (and likewise for `hardPowerOff()`.

pkg/provisioner/ironic/ironic.go
rdoxenham: Agreed on the above. `metal3v1alpha1.RebootMode` has two current states, soft and hard (https://github.com/metal3-io/baremetal-operator/blob/master/apis/metal3.io/v1alpha1/baremetalhost_types.go#L576)

We check the mode requested and act accordingly: https://github.com/metal3-io/baremetal-operator/blob/master/pkg/provisioner/ironic/ironic.go#L1755

pkg/provisioner/provisioner.go
```diff
@@ -139,12 +139,12 @@ type Provisioner interface {
...
 	// PowerOn ensures the server is powered on independently of any image
 	// provisioning operation.
-	PowerOn() (result Result, err error)
+	PowerOn(force bool) (result Result, err error)
 
 	// PowerOff ensures the server is powered off independently of any image
 	// provisioning operation. The boolean argument may be used to specify
 	// if a hard reboot (force power off) is required - true if so.
-	PowerOff(rebootMode metal3v1alpha1.RebootMode) (result Result, err error)
+	PowerOff(rebootMode metal3v1alpha1.RebootMode, forceHardPowerOff bool) (result Result, err error)
```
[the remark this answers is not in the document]
rdoxenham: "Hard" reboot mode as added in https://github.com/metal3-io/baremetal-operator/pull/795 will implement a hard power off. Maybe I'm misunderstanding why we're doing this twice here?

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -966,9 +966,9 @@ func (r *BareMetalHostReconciler) manageHostPower(prov provisioner.Provisioner,
 		"reboot process", desiredPowerOnState != info.host.Spec.Online)
 
 	if desiredPowerOnState {
-		provResult, err = prov.PowerOn()
+		provResult, err = prov.PowerOn(info.host.Status.ErrorCount != 0)
 	} else {
-		provResult, err = prov.PowerOff(desiredRebootMode)
+		provResult, err = prov.PowerOff(desiredRebootMode, info.host.Status.ErrorCount != 0)
```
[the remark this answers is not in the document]
sadasu: My original reservation with this was that the "desiredRebootMode" really did not change. Only the action we are taking on a PowerOff() failure with soft power off is changing. Hence my interpretation of the force flag.
I see that this is the preferred way to proceed for the community so, I am making this change to the desiredRebootMode.

pkg/provisioner/ironic/ironic.go
sadasu: Done.

pkg/provisioner/ironic/ironic.go
sadasu: Fixed.

pkg/provisioner/ironic/ironic.go
sadasu: Fixed as part of https://github.com/metal3-io/baremetal-operator/pull/841/commits/3d834e33e9b764f21298cd6a3661164133529f8d.

pkg/provisioner/provisioner.go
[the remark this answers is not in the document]
sadasu: @rdoxenham let take a look at the changes you mentioned. I believe this is just augmenting it by performing better error handling starting with how we handle LastError.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1734,46 +1734,41 @@ func (p *ironicProvisioner) PowerOn() (result provisioner.Result, err error) {
...
 // PowerOff ensures the server is powered off independently of any image
 // provisioning operation.
-func (p *ironicProvisioner) PowerOff(rebootMode metal3v1alpha1.RebootMode) (result provisioner.Result, err error) {
+func (p *ironicProvisioner) PowerOff(rebootMode metal3v1alpha1.RebootMode, force bool) (result provisioner.Result, err error) {
 	p.log.Info(fmt.Sprintf("ensuring host is powered off (mode: %s)", rebootMode))
 
-	if rebootMode == metal3v1alpha1.RebootModeHard {
-		result, err = p.hardPowerOff()
+	if rebootMode == metal3v1alpha1.RebootModeSoft {
+		return p.softPowerOff()
```
zaneb, line 1763: It's a little unfortunate that we have to record a user-visible failure here when soft power off isn't supported. This is still an improvement to the readability though, because we're no longer calling `softPowerOff()` to check the state of a hard power off we fell back to.
I think it should be possible to get the best of both in future with some refactoring...

pkg/provisioner/ironic/ironic.go
```diff
@@ -1820,11 +1824,17 @@ func (p *ironicProvisioner) softPowerOff() (result provisioner.Result, err error
 		// If the target state is unset while the last error is set,
 		// then the last execution of soft power off has failed.
 		if targetState == "" && ironicNode.LastError != "" {
-			return result, SoftPowerOffFailed{}
+			p.log.Info("soft power off error", "msg", ironicNode.LastError)
+			return operationFailed(ironicNode.LastError)
```
zaneb, line 1828: The code in this function up to this point is now essentially common to both `softPowerOff()` and `hardPowerOff()`, so we could extract it back up into `PowerOff()`.
With that done we could allow an immediate fall back from SoftPowerOffUnsupported to trying a hard power off without reporting an error, since we'd then only call `softPowerOff()` (or `hardPowerOff()`) at the *start* of a power off attempt.

zaneb (commented):
Thanks Sandhya, this is great.
Seeing this has given me some other ideas for refactorings, but this is a significant improvement by itself so I also wouldn't mind merging it as-is.
/approve

pkg/provisioner/ironic/ironic.go
```diff
@@ -1820,11 +1824,17 @@ func (p *ironicProvisioner) softPowerOff() (result provisioner.Result, err error
 		// If the target state is unset while the last error is set,
 		// then the last execution of soft power off has failed.
 		if targetState == "" && ironicNode.LastError != "" {
-			return result, SoftPowerOffFailed{}
```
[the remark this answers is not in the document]
sadasu: Agree. Done.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1847,12 +1825,16 @@ func (p *ironicProvisioner) softPowerOff() (result provisioner.Result, err error
...
-			return transientError(err)
+			switch err.(type) {
+			case HostLockedError:
+				return retryAfterDelay(powerRequeueDelay)
+			default:
+				return transientError(errors.Wrap(err, "failed to power off host"))
+			}
 		}
 		p.publisher("PowerOff", "Host soft powered off")
 	}
```
sadasu, line 1840: Should we be returning operationComplete() here like hardPowerOff() does?

zaneb: Ideally, yep, good spotting. The default value of result is equivalent to what operationComplete() returns so it doesn't technically make any difference, but from a documentation perspective we should return the result of changePower inside the if statement and operationComplete() outside. I must have missed this when I originally did the result functions.

sadasu:
@rdoxenham @andfasano any chance I can get a 2nd approval from either of you ?

rdoxenham:
> @rdoxenham @andfasano any chance I can get a 2nd approval from either of you ?

sorry, I don't have approve/lgtm rights on metal3-io, only OpenShift :(

pkg/provisioner/ironic/ironic.go
```diff
@@ -1734,46 +1734,41 @@ func (p *ironicProvisioner) PowerOn() (result provisioner.Result, err error) {
 			p.log.Info("waiting for power status to change")
 			return operationContinuing(powerRequeueDelay)
 		}
-		result, err = p.changePower(ironicNode, nodes.PowerOn)
-		switch err.(type) {
-		case nil:
-		case HostLockedError:
-		default:
-			return transientError(errors.Wrap(err, "failed to power on host"))
+		if ironicNode.LastError != "" && !force {
```
andfasano, line 1737: @sadasu (maybe I missed some bits but) can you please clarify the reason to check `ironicNode.LastError` before invoking `changePower`? Couldn't be something to be managed just after checking `changePower`?

sadasu: @andfasano When the 1st call to changePower() resulted in operationContinuing() and when PowerOn() is called again and this time the Last Error is not empty. We do not want to continue to changePower() without reporting the contents of LastError.

andfasano: My point was that, if LastError is set just after a call to changePower(), maybe an immediate `p.getNode()` after it would be sufficient to inspect the node _without_ waiting another loop (with the possibility eventually to return immediately an operationFailed).

In any case, it seems that in the current code the `operationFailed` is invoked just once (the second time). If the problem persists, the subsequent invocations will skip it given that the force flag will result to be set (PowerManagementError)

zaneb: I don't know that there's any reason to think a LastError will show up immediately after calling changePower(). In fact, quite the opposite - I'd expect it to almost always fail because of a timeout.

When the force flag is passed we return success, which will result in `clearError()` being called so that next time around the force flag will be unset and we _will_ report an error if another one occurs. This is consistent with all of the other methods that take a force flag.

If you call changePower first then either you will never see an error that appears asynchronously (presumably all of them, since the purpose of LastError is recording asynchronous errors) or you would have to only call it when the force flag is set, which means you couldn't change the power without first flagging a power management error.

elfosardo (approved):

sadasu:
/cc @maelk @furkatgofurov7 

This has been fixed for a while. I am looking for a /lgtm.

zaneb:
@andfasano can we merge this now?

pkg/provisioner/ironic/ironic.go
andfasano: Ok

andfasano (approved):
Reviewed the latest change and it looks ok to me. AFAIU, part of the duplication and errors check fix will be addressed in #816 

/lgtm

PR #842, opened 2021-03-31
Set bmh 'name' in metadata as hostname

kashifest:
This will set 'name' field of metaData as bmh name, otherwise by default ironic name gets set as hostname by default and ~ in hostname seems to be problematic for kubeadm during cloudinit nodeRegistration.

### DIFF
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -1456,6 +1456,7 @@ func (p *ironicProvisioner) Provision(hostConf provisioner.HostConfigData) (resu
 			"metal3-name":      p.host.ObjectMeta.Name,
 			"local-hostname":   p.host.ObjectMeta.Name,
 			"local_hostname":   p.host.ObjectMeta.Name,
+			"name":             p.host.ObjectMeta.Name,
 		}
 		metaDataRaw, err := hostConf.MetaData()
 		if err != nil {


kashifest:
To be exact, this is the issue we are facing in v1a3 CI . We dont see it in normal integration (v1a4) since in that case the node name is taken from data template:

`nodeRegistration.name: Invalid value: "metal3~node-1": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')`

pkg/provisioner/ironic/ironic.go
```diff
@@ -1455,7 +1455,6 @@ func (p *ironicProvisioner) Provision(hostConf provisioner.HostConfigData) (resu
 			"metal3-namespace": p.host.ObjectMeta.Namespace,
 			"metal3-name":      p.host.ObjectMeta.Name,
 			"local-hostname":   p.host.ObjectMeta.Name,
-			"local_hostname":   p.host.ObjectMeta.Name,
```
maelk, line 1458: I would maybe not remove that. There was an issue quite some time ago, where one of the two was used by cloud-init instead of the other. I don't remember the specifics

kashifest: Although it passed all the 4 CI runs... I take you suggestion and keep it as it is.

kashifest: When this line was not here, tests were passing, now I am seeing this after pivot back : 

```
{"level":"info","ts":1617200333.4666467,"logger":"controllers.BareMetalHost","msg":"registering and validating access to management controller","baremetalhost":"metal3/node-1","provisioningState":"provisioned","credentials":{"credentials":{"name":"node-1-bmc-secret","namespace":"metal3"},"credentialsVersion":"8800"}}
{"level":"info","ts":1617200333.5034752,"logger":"provisioner.ironic","msg":"node with name metal3-node-1 doesn't exist","host":"metal3-node-1"}
{"level":"info","ts":1617200333.5035117,"logger":"provisioner.ironic","msg":"looking for existing node by MAC","host":"metal3-node-1","MAC":"00:b0:5e:58:de:6b"} 

baremetal node list
| c222e16a-0211-4806-8820-a4a635f47270 | node-1        | b5eea54f-078a-4cf9-8040-34f7638cfbad | power on    | manageable         | False       |
| 0acc9a34-feff-4286-af3d-60fa9dab14ad | metal3-node-0 | bc7538b7-6b86-4c0a-8c4d-903ac73adfc3 | power on    | active             | False 
```

pkg/provisioner/ironic/ironic.go
```diff
@@ -55,7 +55,7 @@ const (
 	powerOff      = "power off"
 	softPowerOff  = "soft power off"
 	powerNone     = "None"
-	nameSeparator = "~"
+	nameSeparator = "-"
```
zaneb, line 58: This is [unsafe](https://github.com/metal3-io/baremetal-operator/pull/797#pullrequestreview-592348276); if we are going to change it, it must be to `.` not `-`.

zaneb (commented):
How is the ironic node name getting used as the hostname anyway? If we can find a way to stop that it would be preferable.

hardys:
> How is the ironic node name getting used as the hostname anyway? If we can find a way to stop that it would be preferable.

I thought the same thing - in the metadata we're setting e.g ` p.host.ObjectMeta.Name` which is the BMH name, not the name in Ironic, right? @dtantsur does Ironic inject the Ironic name in addition to the meta_data.json?

zaneb:
> The answer appears to be yes, so this name change isn't as "internal" as we originally envisaged - it seems Ironic injects `name` in addition to the keys we create in BMO, and presumably cloud-init uses that to set the hostname...

It's [done using `setdefault()`](https://opendev.org/openstack/ironic/src/branch/master/ironic/conductor/utils.py#L994), so if we set the `name` field explicitly in the metadata, that will resolve the problem.

kashifest:
Thanks @zaneb that seems to work. Tests have passed now.

kashifest:
Thanks @hardys for your insight.

asalkeld (approved):
/lgtm
/approve

PR #836, opened 2021-03-29
Extend RAID configuration for iDRAC BMC type

nasirkamal:
Extend RAID configuration for iDRAC BMC type

This PR also fix mis-configured function calls
buildManualCleaningSteps() and setTargetRAIDCfg().
They were using Status.Provisioning.RAID instead of
Spec.RAID. This now correctly performs RAID configuration.

PR #844, opened 2021-04-02
Fix bug: never set target_raid_config

Hellcatlk:
Signed-off-by: zouyu <zouy.fnst@cn.fujitsu.com>

Due to the change of the code in [baremetalhost_controller.go](https://github.com/metal3-io/baremetal-operator/blob/master/controllers/metal3.io/baremetalhost_controller.go#L217), the `saveHostProvisioningSettings()` called [here](https://github.com/metal3-io/baremetal-operator/blob/master/controllers/metal3.io/baremetalhost_controller.go#L687) didn't change `p.host`, so the value of `p.host.Status.Provisioning.RAID` is always nil.


andfasano:
@Hellcatlk looks like there's some overlapping with #836?

Hellcatlk:
> @Hellcatlk looks like there's some overlapping with #836?

Sorry, I missed this PR.

PR #847, opened 2021-04-05
Extend RAID configuration for iDRAC BMC type

nasirkamal:
This PR also fix mis-configured function calls
buildManualCleaningSteps() and setTargetRAIDCfg().
They were using Status.Provisioning.RAID instead of
Spec.RAID. This now correctly performs RAID configuration.

### DIFF
--- a/pkg/bmc/idrac.go
+++ b/pkg/bmc/idrac.go
@@ -87,7 +87,7 @@ func (a *iDracAccessDetails) PowerInterface() string {
 }
 
 func (a *iDracAccessDetails) RAIDInterface() string {
-	return "no-raid"
+	return "idrac-wsman"
 }
 
 func (a *iDracAccessDetails) VendorInterface() string {


demonCoder95:
@zaneb @dtantsur would be great if you could take a quick look at this small PR

dtantsur:
This PR has merge conflicts, could you rebase?

demonCoder95:
@dtantsur done

demonCoder95:
@dtantsur all done.

zaneb:
Two questions:

1. Does this work? (I know it *should*, but has anybody tried it)
2. With a RAID driver enabled we end up doing extra manual cleaning steps even for Hosts with no raid configured, I think on every deprovison/reprovision. Have we checked that the performance cost of this is acceptable?

demonCoder95:
> Two questions:
> 
> 1. Does this work? (I know it _should_, but has anybody tried it)

Yes, we've done thorough testing on DellEMC PowerEdge R740XD with the latest iDRAC firmware version. And the code is fully functional.

> 2. With a RAID driver enabled we end up doing extra manual cleaning steps even for Hosts with no raid configured, I think on every deprovison/reprovision. Have we checked that the performance cost of this is acceptable?

The only cleaning step performed in that case would be deleting existing configurations. I suppose when it comes to cleaning existing RAID configuration, the cost should be acceptable. We would like to have that disk erased properly before the reprovisioning takes place.

demonCoder95:
Great. Thanks! Let's get this merged then!

/approve

PR #848, opened 2021-04-06
modify power_interface for irmc

zhouhao3:
When `irmc` `power_interface` is used, Soft Reboot (Graceful Reset) and Soft Power Off (Graceful Power Off)
are only available if `ServerView` agents are installed.
Modify the value of `power_interface` to `ipmitool`, so that soft power off can be used directly.

Signed-off-by: Zhou Hao <zhouhao@fujitsu.com>

### DIFF
--- a/pkg/bmc/irmc.go
+++ b/pkg/bmc/irmc.go
@@ -74,7 +74,7 @@ func (a *iRMCAccessDetails) ManagementInterface() string {
 }
 
 func (a *iRMCAccessDetails) PowerInterface() string {
-	return ""
+	return "ipmitool"
 }
 
 func (a *iRMCAccessDetails) RAIDInterface() string {


pkg/bmc/irmc.go
```diff
@@ -74,7 +74,7 @@ func (a *iRMCAccessDetails) ManagementInterface() string {
 }
 
 func (a *iRMCAccessDetails) PowerInterface() string {
-	return ""
+	return "ipmitool"
```
dtantsur, line 77: We don't populate ipmi credentials above, will it work?

dtantsur (commented):
/ok-to-test

pkg/bmc/irmc.go
zhouhao3: In my test environment, ipmi related information can be successfully obtained.
```
| driver_info            | {'deploy_kernel': 'http://localhost:6181/images/ironic-python-agent.kernel',                |
|                        | 'deploy_ramdisk': 'http://localhost:6181/images/ironic-python-agent.initramfs',             |
|                        | 'irmc_address': '192.168.30.25', 'irmc_password': '******', 'irmc_username': 'admin',       |
|                        | 'ipmi_address': '192.168.30.25', 'ipmi_username': 'admin', 'ipmi_password': '******'}   
```

zhouhao3: After I manually changed the value of node's `power_interface` to `ipmitool`, `sofftshutdown` can be executed correctly.
@dtantsur So do you have any questions about this modification?

zhouhao3:
@dhellmann @zaneb PTAL

pkg/bmc/irmc.go
zaneb: Looks like the iRMC driver takes care of it: https://opendev.org/openstack/ironic/src/branch/master/ironic/drivers/modules/irmc/common.py#L160-L174

zaneb (commented):
/lgtm

dtantsur (approved):
/approve

PR #850, opened 2021-04-09
Add disk type field to BMH HardwareDetails

s3rj1k:
Motivation to add additional disk type is to allow custom storage provider assemble different types of storage based on needed storage profiles.

Something like:
 - `fast and expensive` storage on NVME
 - `general` storage on SSD
 - `backup`, `cold data` storage on HDD
 - `hybrid cache` storage with fast NVMEs backed by HDD
 - ...

### DIFF
--- a/apis/metal3.io/v1alpha1/baremetalhost_types.go
+++ b/apis/metal3.io/v1alpha1/baremetalhost_types.go
@@ -442,6 +442,16 @@ const (
 	TeraByte          = GigaByte * 1000
 )
 
+// DiskType is a disk type, i.e. HDD, SSD, NVME.
+type DiskType string
+
+// DiskType constants.
+const (
+	HDD  DiskType = "HDD"
+	SSD  DiskType = "SSD"
+	NVME DiskType = "NVME"
+)
+
 // CPU describes one processor on the host.
 type CPU struct {
 	Arch           string     `json:"arch,omitempty"`
@@ -457,8 +467,16 @@ type Storage struct {
 	// may not be stable across reboots.
 	Name string `json:"name,omitempty"`
 
-	// Whether this disk represents rotational storage
+	// Whether this disk represents rotational storage.
+	// This field is not recommended for usage, please
+	// prefer using 'Type' field instead, this field
+	// will be deprecated eventually.
 	Rotational bool `json:"rotational,omitempty"`
+
+	// Device type, one of: HDD, SSD, NVME.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Enum=HDD;SSD;NVME;
+	Type DiskType `json:"type,omitempty"`
 
 	// The size of the disk in Bytes
 	SizeBytes Capacity `json:"sizeBytes,omitempty"`
--- a/pkg/provisioner/ironic/hardwaredetails/hardwaredetails.go
+++ b/pkg/provisioner/ironic/hardwaredetails/hardwaredetails.go
@@ -95,12 +95,25 @@ func getNICDetails(ifdata []introspection.InterfaceType,
 	return nics
 }
 
+func getDiskType(diskdata introspection.RootDiskType) metal3v1alpha1.DiskType {
+	if diskdata.Rotational {
+		return metal3v1alpha1.HDD
+	}
+
+	if strings.HasPrefix(diskdata.Name, "/dev/nvme") {
+		return metal3v1alpha1.NVME
+	}
+
+	return metal3v1alpha1.SSD
+}
+
 func getStorageDetails(diskdata []introspection.RootDiskType) []metal3v1alpha1.Storage {
 	storage := make([]metal3v1alpha1.Storage, len(diskdata))
 	for i, disk := range diskdata {
 		storage[i] = metal3v1alpha1.Storage{
 			Name:               disk.Name,
 			Rotational:         disk.Rotational,
+			Type:               getDiskType(disk),
 			SizeBytes:          metal3v1alpha1.Capacity(disk.Size),
 			Vendor:             disk.Vendor,
 			Model:              disk.Model,
--- a/config/crd/bases/metal3.io_baremetalhosts.yaml
+++ b/config/crd/bases/metal3.io_baremetalhosts.yaml
@@ -495,7 +495,7 @@ spec:
                           description: The Linux device name of the disk, e.g. "/dev/sda". Note that this may not be stable across reboots.
                           type: string
                         rotational:
-                          description: Whether this disk represents rotational storage
+                          description: Whether this disk represents rotational storage. This field is not recommended for usage, please prefer using 'Type' field instead, this field will be deprecated eventually.
                           type: boolean
                         serialNumber:
                           description: The serial number of the device
@@ -504,6 +504,13 @@ spec:
                           description: The size of the disk in Bytes
                           format: int64
                           type: integer
+                        type:
+                          description: 'Device type, one of: HDD, SSD, NVME.'
+                          enum:
+                          - HDD
+                          - SSD
+                          - NVME
+                          type: string
                         vendor:
                           description: The name of the vendor of the device
                           type: string
--- a/config/render/capm3.yaml
+++ b/config/render/capm3.yaml
@@ -493,7 +493,7 @@ spec:
                           description: The Linux device name of the disk, e.g. "/dev/sda". Note that this may not be stable across reboots.
                           type: string
                         rotational:
-                          description: Whether this disk represents rotational storage
+                          description: Whether this disk represents rotational storage. This field is not recommended for usage, please prefer using 'Type' field instead, this field will be deprecated eventually.
                           type: boolean
                         serialNumber:
                           description: The serial number of the device
@@ -502,6 +502,13 @@ spec:
                           description: The size of the disk in Bytes
                           format: int64
                           type: integer
+                        type:
+                          description: 'Device type, one of: HDD, SSD, NVME.'
+                          enum:
+                          - HDD
+                          - SSD
+                          - NVME
+                          type: string
                         vendor:
                           description: The name of the vendor of the device
                           type: string


apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -460,6 +460,10 @@ type Storage struct {
 	// Whether this disk represents rotational storage
 	Rotational bool `json:"rotational,omitempty"`
 
+	// Device type, one of: HDD, SSD, NVME
```
dhellmann, line 463: Since we're interpreting discovered data to report the type, we should include the rules here in this comment since it's the main documentation a user will see via the CRD.

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -460,6 +460,10 @@ type Storage struct {
...
+	// +kubebuilder:validation:Enum=HDD;SSD;NVME
+	Type string `json:"type,omitempty"`
```
dhellmann, line 465: Should we use a go type for the values so we can define constants?

dhellmann (commented):
Normally we would want a design doc for an API change like this.

s3rj1k:
@dhellmann Will do PR for design docs, if generally this is acceptable change.

apis/metal3.io/v1alpha1/baremetalhost_types.go
s3rj1k: Yea we could, but other fields in here are strings, so I was thinking to be as similar as possible.

We could do type alias for string and use constants for it.

Main concern is, keep resulting JSON similar to one that current PR generates

dhellmann:
> @dhellmann Will do PR for design docs, if generally this is acceptable change.

I think I like the idea. It's a little bit like the HWCC, but I think it improves the usability of the API so I think that's OK. And it may make HWCC rules easier to write. @digambar15 what do you think?

apis/metal3.io/v1alpha1/baremetalhost_types.go
s3rj1k: Yea, seems reasonable, will do.

apis/metal3.io/v1alpha1/baremetalhost_types.go
dhellmann: Ah, yeah, good point. Those other values come directly out of Ironic and aren't modified. These are values we're defining ourselves. So, I could go either way.

s3rj1k:
@dhellmann @digambar15 PR is rebased, please take a look.

Ashughorla:
@s3rj1k Will you please look at [Proposal for Disk and NIC selector](https://github.com/metal3-io/metal3-docs/blob/master/design/hardware-classification-controller/support-for-new-parameters-hwcc-DiskAndNIC.md)
This Proposal is related to diskSelector parameter support in HWCC. This also recognize Raid disk.
Rule book for disk selection is present here [Rule Book](https://github.com/metal3-io/hardware-classification-controller/pull/69/files)
Implementation PR of diskSelector is here [Disk Selector](https://github.com/metal3-io/hardware-classification-controller/pull/64)

s3rj1k:
@Ashughorla Strange that this proposal covers both disks and NICs, I would have expected that two separate proposals.

I have some doubts about RAID disk type. Covering all the existing possibilities seems like impossible tasks, one would have to use some kind of DB, like smartmontools for regular disks.

Proposal covers additional HWCC CRD, this patch operates on BMH API data that is known to exist in every case.
(I can't have disk without kernel name).

dtantsur:
Should we deprecate `Rotational` then? What's our position on deprecating API things?

dhellmann:
> Should we deprecate `Rotational` then? What's our position on deprecating API things?

We've never updated the version of the API, because we've always been able to just add fields. (For what it's worth, this is a good example of why we should avoid `bool` fields.)

We need to formalize a process for deprecating and removing a field. I know in OpenShift it will take us several release cycles to move to a new API version and drop the old one. Other members of the community may have similar requirements, so we should take all of them into account.

s3rj1k:
@dhellmann @dtantsur We do need `Rotational` field data in some form (unexported ?) to set correct type of `HDD` for rotational disks.

s3rj1k:
https://github.com/metal3-io/baremetal-operator/issues/874#issuecomment-844137392

So assuming that we do not update API version, we do not need (allowed) to remove any fields in this PR.

@dtantsur @dhellmann 
cc: @zaneb

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -460,6 +470,15 @@ type Storage struct {
 	// Whether this disk represents rotational storage
 	Rotational bool `json:"rotational,omitempty"`
 
+	// Device type, one of: HDD, SSD, NVME.
+	// Field value is computed from "Name" and "Rotational" field values.
```
zaneb, line 474: Nobody cares how it's computed :)

apis/metal3.io/v1alpha1/baremetalhost_types.go
zaneb: You went to the trouble of defining a DiskType, so you might as well use it. Then you wouldn't have to cast back to string all over the place.

apis/metal3.io/v1alpha1/baremetalhost_types.go
s3rj1k: This is for auto-generated docs and CRD comments, should I remove this?

apis/metal3.io/v1alpha1/baremetalhost_types.go
s3rj1k: Well, this is so obvious, I don't have an idea why I missed that, thanks @zaneb 
Fixing in next rebase.

elfosardo (approved):

apis/metal3.io/v1alpha1/baremetalhost_types.go
zaneb: Someone looking at the CRD wants to know that it says "NVME" when it's an NVME, "SSD" when it's an SSD, and "HDD" when it's an HDD.
Explaining how it's calculated only introduces a suspicion that it might be wrong. We don't state how any of the other values are calculated.

s3rj1k: @zaneb Thanks for explaining, recent rebase removed that comment

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -460,6 +472,14 @@ type Storage struct {
 	// Whether this disk represents rotational storage
 	Rotational bool `json:"rotational,omitempty"`
 
+	// Device type, one of: HDD, SSD, NVME.
+	// When "Rotational" is "true", field value is set to "HDD".
+	// When "Rotational" is "false" and "Name" starts with "/dev/nvme",
+	// field value is set to "NVME".
+	// When previous cases are not covered, field value is set to "SSD".
+	// +kubebuilder:validation:Enum=HDD;SSD;NVME
```
dtantsur, line 480: Add UNKNOWN aka empty value?

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -460,6 +472,14 @@ type Storage struct {
 	// Whether this disk represents rotational storage
```
dtantsur, line 472: Even if we cannot remove it, let's mark it as deprecated in the documentation.

dtantsur (changes requested):

apis/metal3.io/v1alpha1/baremetalhost_types.go
s3rj1k: Hmm, missed that due to broken controller-tools, will fix, thanks

apis/metal3.io/v1alpha1/baremetalhost_types.go
s3rj1k: I don't think that it is deprecated, rather not recommended for usage.
Something like, use `that` field instead.

apis/metal3.io/v1alpha1/baremetalhost_types.go
s3rj1k: fixed

apis/metal3.io/v1alpha1/baremetalhost_types.go
dtantsur: Deprecated == not recommended for usage :) I'm fine with the proposed wording.

s3rj1k: So we need to update comments inside baremetal-operator or rather docs ?

dtantsur: This comment since it's rendered in a few places.

s3rj1k: @dtantsur added comment about field deprecation.

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -457,9 +469,21 @@ type Storage struct {
...
+	// prefer using 'Type' field instead, this field
+	// will be deprecated eventually.
 	Rotational bool `json:"rotational,omitempty"`
 
+	// Device type, one of: HDD, SSD, NVME, UNKNOWN.
+	// When "Rotational" is "true", field value is set to "HDD".
+	// When "Rotational" is "false" and "Name" starts with "/dev/nvme",
+	// field value is set to "NVME".
+	// When previous cases are not covered, field value is set to "SSD".
+	// When "Name" undefined field value is set to "UNKNOWN".
```
zaneb, line 483: TBH I still don't think that all of these implementation details are a helpful thing to put in the CRD help. We just said the Rotational field was deprecated and to prefer this one, but then we define this one in terms of the Rotational field?

pkg/provisioner/ironic/hardwaredetails/hardwaredetails.go
```diff
@@ -95,12 +95,29 @@ func getNICDetails(ifdata []introspection.InterfaceType,
 	return nics
 }
 
+func getDiskType(diskdata introspection.RootDiskType) metal3v1alpha1.DiskType {
+	if diskdata.Name == "" {
+		return metal3v1alpha1.UNKNOWN
```
zaneb, line 100: I don't know that we'll ever be able to remove the Rotational field when this discards some of that data in some circumstances. Wouldn't it be better to just fall back to returning "SSD" or "HDD" in this case?

s3rj1k: There is no possible way that name field could be empty, name filed here is kernel device name, "/dev/sda" for example.
If this name is empty device shouldn't exists in first place.

We be better to remove omitempty from struct field, to make this filed mandatory.

Also there is no way of knowing what type of device is this "SSD" or "HDD", how can we chose what default value is for this (for none existent device).

Defaulting to either one of them will look strange, none existent device of some type?

But when using unknown, this effectively sets disk type to empty string and this is exactly default value for uninitialized string in golang, this makes more sense when combined with empty kernel name

apis/metal3.io/v1alpha1/baremetalhost_types.go
s3rj1k: I do not see any issues with removing all of comments that describe implementation details.
Will do on next rebase.

pkg/provisioner/ironic/hardwaredetails/hardwaredetails.go
zaneb: > There is no possible way that name field could be empty

Then why do we have an enum value for something that can't happen?

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -442,6 +442,18 @@ const (
 	TeraByte          = GigaByte * 1000
 )
 
+// DiskType is a disk type, i.e. HDD, SSD, NVME.
+type DiskType string
+
+// DiskType constants.
+const (
+	UNKNOWN DiskType = ""
```
zaneb, line 450: This value doesn't match the enum values in the CRD.

pkg/provisioner/ironic/hardwaredetails/hardwaredetails.go
s3rj1k: > Then why do we have an enum value for something that can't happen?
This could happen if IPA has some bug and allowed by struct field to happen.

getDiskType() is expected to return something in this case it is empty string but with type aliasing, hence enum for empty string

apis/metal3.io/v1alpha1/baremetalhost_types.go
s3rj1k: Not following, why?

image

zaneb: Because the value here is `""`, not `"UNKNOWN"`.

pkg/provisioner/ironic/hardwaredetails/hardwaredetails.go
zaneb: You said it can't happen so my question was why, of all the infinite things that can't happen, do we choose to pretend to handle this one as if it could?

s3rj1k: Can't happen on normal working linux environment, linux device needs to have name.

Can happen when IPA sends malformed data, current CRD validation rules allow that.

s3rj1k: So device in terms of CRD validation can have any data set, can have Vendor, Model, SizeBytes, ... set but have empty Name.

(IPA uses lsblk output to get this data parsed)

Knowing all of this, as end user I can't grasp what to do with disk data that has all of fields set except name, how it was set, from where this data comes, how to match this data to actual disk inside linux.

apis/metal3.io/v1alpha1/baremetalhost_types.go
s3rj1k: ohh, yea, obviously, thanks!

apis/metal3.io/v1alpha1/baremetalhost_types.go
s3rj1k: done

apis/metal3.io/v1alpha1/baremetalhost_types.go
s3rj1k: fixed

pkg/provisioner/ironic/hardwaredetails/hardwaredetails.go
zaneb: I think you're focusing too much on the CRD validation. The HardwareDetails can only be set by the BMO, and they're set by this code based on the the output of IPA. The CRD validation is irrelevant except insofar as it must allow all of the values that IPA can send. We don't need to worry about values that IPA can't send, regardless of whether they would be allowed by the OpenAPI specification. We should treat the OpenAPI schema as documentation for the user on what values to expect.

s3rj1k: Still I do not know how to choose what type to set for a case when name is empty.

```
if diskdata.Name == "" {...}
```

How about just removing this case altogether?

zaneb: Yes, I think we should just remove this check (and the UNKNOWN enum value) since it can't happen.

s3rj1k: done, thanks

s3rj1k:
@zaneb Thanks!

PR #845, opened 2021-04-02
Always update image details

fmuyassarov:
Always update the image data within `ValidateManagementAccess()` regardless of Ironic node existence or not. Currently, image data update is happening only if Ironic node doesn't exist. That means, if we failed to write image data, we won't be able to update it again in the next retries since node already exists.
Fixes https://github.com/metal3-io/baremetal-operator/issues/839.

Issue #839, opened 2021-03-30
Image update in ironic on re-registration is fragile


zaneb:
Updating the image data during registration using `getImageUpdateOptsForNode()` happens only if the node doesn't exist. This is a bug, because if writing the image data it fails then we will never try it again so long as the node exist.

Without data we cannot adopt, so if this occurred we would be stuck in a permanent error loop if the host is in the Provisioned/Deprovisioning states.

_Originally identified in https://github.com/metal3-io/baremetal-operator/pull/784#discussion_r604217324_

fmuyassarov:
Do we need to add a retry ?
/kind bug

zaneb:
It retries fine, but on the retry the Node will already exist in Ironic. The image setting are currently only written in the path where the Node didn't exist and we had to recreate it. They should be written unconditionally, whether the node already existed or not.

PR #852, opened 2021-04-12
Always lazily update Node options in Ironic

zaneb:
It became clear from #839 and #849 that in order to ensure consistent behaviour, we must ensure that the Ironic Node's image data is up to date on every reconcile, and this was implemented in #845 by unconditionally updating it. This will result in a write to Ironic on every reconcile, which is excessive.

We have the ability to check whether the data we want to set is already present, but we largely don't because of the huge amount of messy boilerplate currently required to update each option. This PR refactors the code so that we can just pass a map of the options we want to set to a function and lazy updates come for free, along with standardised logging.

### DIFF
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -689,24 +689,24 @@ func (p *ironicProvisioner) InspectHardware(data provisioner.InspectData, force,
 				fallthrough
 			default:
 				p.log.Info("updating boot mode before hardware inspection")
-				op, value := buildCapabilitiesValue(ironicNode, data.BootMode)
-				updates := nodes.UpdateOpts{
-					nodes.UpdateOperation{
-						Op:    op,
-						Path:  "/properties/capabilities",
-						Value: value,
-					},
-				}
-				_, err = nodes.Update(p.client, ironicNode.UUID, updates).Extract()
-				switch err.(type) {
-				case nil:
-				case gophercloud.ErrDefault409:
-					p.log.Info("could not update host settings in ironic, busy")
-					result, err = retryAfterDelay(provisionRequeueDelay)
-					return
-				default:
-					result, err = transientError(errors.Wrap(err, "failed to update host boot mode settings in ironic"))
-					return
+
+				updates := updateOptsBuilder(p.debugLog).
+					SetPropertiesOpts(optionsData{
+						"capabilities": buildCapabilitiesValue(ironicNode, data.BootMode),
+					}, ironicNode).
+					Updates
+				if len(updates) > 0 {
+					_, err = nodes.Update(p.client, ironicNode.UUID, updates).Extract()
+					switch err.(type) {
+					case nil:
+					case gophercloud.ErrDefault409:
+						p.log.Info("could not update host settings in ironic, busy")
+						result, err = retryAfterDelay(provisionRequeueDelay)
+						return
+					default:
+						result, err = transientError(errors.Wrap(err, "failed to update host boot mode settings in ironic"))
+						return
+					}
 				}
 
 				p.log.Info("starting new hardware inspection")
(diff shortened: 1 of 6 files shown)


pkg/provisioner/ironic/update_opts.go
```diff
@@ -0,0 +1,145 @@
...
+		if !(present && optionValueEqual(current, desiredValue)) {
+			if log != nil {
+				if present {
+					log.Info("updating option data", "option", name, "value", desiredValue)
+				} else {
+					log.Info("adding option data", "option", name, "value", desiredValue)
+				}
+			}
+			return &nodes.UpdateOperation{
+				Op:    nodes.AddOp, // Add also does replace
```
zaneb, line 89: @dtantsur can you confirm that there's no reason to ever use 'replace' instead of 'add'? It would be easy enough to do.

pkg/provisioner/ironic/update_opts.go
```diff
@@ -0,0 +1,145 @@
...
+				Value: desiredValue,
+			}
+		}
+	} else {
+		if present {
+			if log != nil {
+				log.Info("removing option data", "option", name)
+			}
+			return &nodes.UpdateOperation{
+				Op:   nodes.RemoveOp,
```
andfasano, line 100: Using `nil` to indicate a removal operation can make more difficult to understand the intentions (in the points where the method is used). A dedicated function will be easier to read and less error-prone

pkg/provisioner/ironic/update_opts.go
```diff
@@ -0,0 +1,145 @@
...
+				Path: path,
+			}
+		}
+	}
+	return nil
+}
+
+func updateTopLevelOption(name string, currentValue, desiredValue interface{}, log logr.Logger) nodes.UpdateOpts {
```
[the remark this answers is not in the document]
zaneb: > Some of the naming for these functions are not clear to me. I'm not sure about the meaning of `TopLevel` in this case,

It's for top-level options like `/instance_uuid`, as opposed to nested stuff like `/instance_info/...`.

> as it could be used to prepare an update operation for any node properties,

No it can't, because you don't want to replace all of e.g. `/instance_info` wholesale.

pkg/provisioner/ironic/update_opts.go
```diff
@@ -0,0 +1,145 @@
...
+	path := fmt.Sprintf("/%s", name)
+
+	updateOp := getUpdateOperation(name, currentData, desiredValue, path, log)
+	if updateOp != nil {
+		updates = append(updates, *updateOp)
+	}
+	return updates
+}
+
+func updateOptionsSection(node *nodes.Node, section optionsSection, settings optionsData, log logr.Logger) nodes.UpdateOpts {
```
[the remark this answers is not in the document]
zaneb: Fair enough, it seems to be just these two so I'll implement that.

pkg/provisioner/ironic/update_opts.go
```diff
@@ -0,0 +1,145 @@
...
+		return true
+	}
+	return false
+}
+
+func getUpdateOperation(name string, currentData map[string]interface{}, desiredValue interface{}, path string, log logr.Logger) *nodes.UpdateOperation {
+	current, present := currentData[name]
+	if desiredValue != nil {
+		if !(present && optionValueEqual(current, desiredValue)) {
+			if log != nil {
```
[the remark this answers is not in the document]
zaneb: Done.

pkg/provisioner/ironic/update_opts.go
zaneb: That makes sense in the cases where we are unconditionally clearing a field, but I [tried this out](https://github.com/zaneb/baremetal-operator/blob/ironic-update-options-fluent/pkg/provisioner/ironic/update_opts.go) and it turns out that there are some cases where we are e.g. getting the value to pass from a pointer that may be nil. In those cases we don't want to have to look at the value before deciding what to call, we just want to pass the value we've been given and have ironic make it right.
Also the code was way too complicated for the amount of lifting it is doing, although that might just have been me going overboard :laughing:

pkg/provisioner/ironic/update_opts.go
```diff
@@ -0,0 +1,145 @@
...
+		}
+		return true
+	}
+	return false
+}
+
+func getUpdateOperation(name string, currentData map[string]interface{}, desiredValue interface{}, path string, log logr.Logger) *nodes.UpdateOperation {
+	current, present := currentData[name]
+	if desiredValue != nil {
+		if !(present && optionValueEqual(current, desiredValue)) {
```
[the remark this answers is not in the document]
zaneb: I simplified `optionValueEqual()` to avoid comparing scalar types by trying `reflect.DeepEqual()` first, and only falling back to reflection for maps/lists where the type is not `interface{}`.

pkg/provisioner/ironic/ironic.go
```diff
@@ -499,62 +499,27 @@ func (p *ironicProvisioner) ValidateManagementAccess(data provisioner.Management
...
-		)
+		p.getImageUpdateOptsForNode(ironicNode, data.CurrentImage, data.BootMode, updater)
 	}
-	if len(updates) != 0 {
-		_, err = nodes.Update(p.client, ironicNode.UUID, updates).Extract()
+	updater.SetTopLevelOpt("automated_clean",
+		data.AutomatedCleaningMode != metal3v1alpha1.CleaningModeDisabled,
+		ironicNode.AutomatedClean)
+
+	if len(updater.Updates) != 0 {
```
andfasano, line 521: Minor: an improvement point here to remove some code duplication (with the block [here](https://github.com/metal3-io/baremetal-operator/pull/852/files#diff-f05fc4a0c9679815bc34bcef100c2a8d525688f2804a1128c5b9e1e575b76c29R663)) could be to let the `updater` manage the update, ie by having a method like `updater.ApplyUpdatesTo(ironicNode)`. It could help also the readability while browsing the code for the points where the updates are being applied

pkg/provisioner/ironic/update_opts.go
andfasano: Thanks, I understood your point but still remain skeptical in flattening a well defined semantic (add/replace/remove) as the opportunities of unwanted mistakes could increase. The added value of skipping updates when not required still anyhow remains high, so we can eventually re-evaluate this specific point later into another PR  if required.

pkg/provisioner/ironic/update_opts.go
```diff
@@ -0,0 +1,162 @@
...
+			return &nodes.UpdateOperation{
+				Op:   nodes.RemoveOp,
+				Path: path,
+			}
+		}
+	}
+	return nil
+}
+
+type nodeUpdater struct {
```
andfasano, line 110: Nice!

andfasano (approved):

pkg/provisioner/ironic/update_opts.go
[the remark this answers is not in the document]
andfasano: That was really a good idea, as it helps in reducing the complexity.

fmuyassarov (commented):
LGTM!

pkg/provisioner/ironic/ironic.go
zaneb: Great idea. Upon implementation it turns out that it shares more data with the provisioner (we need the client, and also the logger that is not necessarily the one in the nodeUpdater), so I made it an ironicProvisioner method.

andfasano:
Changes looks fine to me. @zaneb do you want to squash the commits?

zaneb:
> @zaneb do you want to squash the commits?

No, definitely not :)

andfasano:
Ok :)
/lgtm

PR #854, opened 2021-04-13
Restart Ironic containers when certificates are changed

namnx228:
This PR is related to these PRs, and needs to wait for them to go in:
https://github.com/metal3-io/ironic-image/pull/247
It enables the feature of Ironic containers that automatically restart when the certificates are updated.

### DIFF
--- a/tools/deploy.sh
+++ b/tools/deploy.sh
@@ -22,6 +22,7 @@ MARIADB_HOST="${MARIADB_HOST:-"mariaDB"}"
 MARIADB_HOST_IP="${MARIADB_HOST_IP:-"127.0.0.1"}"
 KUBECTL_ARGS="${KUBECTL_ARGS:-""}"
 KUSTOMIZE="go run sigs.k8s.io/kustomize/kustomize/v3"
+RESTART_CONTAINER_CERTIFICATE_UPDATED=${RESTART_CONTAINER_CERTIFICATE_UPDATED:-"false"}
 
 SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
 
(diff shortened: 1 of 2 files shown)


dhellmann:
Why would we restart ironic inside of the pod instead of just restarting the entire pod with the new settings?

zaneb:
It seems to me that this is something that ought to happen at the Pod level, rather than being detected inside each individual container. (ideally at the Deployment level, but for Ironic we're currently constrained to running only one copy at a time, so we don't really have to worry about rolling updates or disruption budgets in this case.)

Something like https://github.com/stakater/Reloader might be a better solution.

namnx228:
@dhellmann @zane It is true that the optimal way is to have something outside of the pod that watches for the change of the certificate secret and restart the pod. We can adopt an external opensource component like Zane mentioned. However, it is required to assess the validity of a new opensource component before it can be a part of baremetal-operator. Otherwise, we need to write our own controller, but I think it is also too much for a simple task like restarting the Ironic pod.
For that reason, in this PR, I implement a straightforward approach to solve the problem. I also provide an option to turn on or off the feature as the suggestion of @hardys. Also, for a long run, I agree that we need to plan to implement an optimal approach for this kind of problem in the future. 
So, how do you think?

dhellmann:
> @dhellmann @zane It is true that the optimal way is to have something outside of the pod that watches for the change of the certificate secret and restart the pod. We can adopt an external opensource component like Zane mentioned. However, it is required to assess the validity of a new opensource component before it can be a part of baremetal-operator. Otherwise, we need to write our own controller, but I think it is also too much for a simple task like restarting the Ironic pod.

What is changing the certificate, and how is it doing that?

zaneb:
> What is changing the certificate, and how is it doing that?

This might answer your question: https://github.com/metal3-io/ironic-image/pull/247#issuecomment-809947277

dhellmann:
> > What is changing the certificate, and how is it doing that?
> 
> This might answer your question: [metal3-io/ironic-image#247 (comment)](https://github.com/metal3-io/ironic-image/pull/247#issuecomment-809947277)

So, forcibly and without notice? Fun. OK, I guess I don't object to dealing with it explicitly then. Let's carry on and figure out the best way to do that.

namnx228:
The PR:https://github.com/metal3-io/ironic-image/pull/247 has been merged, so this PR can go in now. Please take a look again @dhellmann @zaneb @dtantsur. Thank you

tools/deploy.sh
```diff
@@ -219,6 +220,11 @@ if [ "${DEPLOY_IRONIC}" == "true" ]; then
     else
       echo "INSPECTOR_REVERSE_PROXY_SETUP=${DEPLOY_TLS}" >> "${IRONIC_BMO_CONFIGMAP}"
     fi
+    if grep "RESTART_CONTAINER_CERTIFICATE_UPDATED" "${IRONIC_BMO_CONFIGMAP}" > /dev/null; then
```
dtantsur, line 223: nit: use `grep -q` instead of `grep .. > /dev/null`

dtantsur (approved):

tools/deploy.sh
namnx228: Fixed! Thank you

elfosardo (approved):

PR #856, opened 2021-04-14
Remove govet script and test

elfosardo:
It's already included in the golint job


elfosardo:
not sure why tide says that this is not mergeable :/

furkatgofurov7:
@elfosardo tide is broken, unfortunately, we have seen the same behavior in another patch in the repo.

PR #857, opened 2021-04-15
Add go.mod for BMH api

kashifest:
This PR adds go.mod for BMH api.

The reason to add the go module in api is to uplift BMO for CAPM3 without requiring controller runtime to be uplifted to > 0.5.14 which is the version currently supported by CAPI.

**Important:**
Once this PR goes in, those who need to uplift BMO, and they are only using BMO apis , you need to change you go modules to use 

`github.com/metal3-io/baremetal-operator/apis` instead of 
`github.com/metal3-io/baremetal-operator`

If you are using/referring other parts of code in BMO not residing in `apis/`, you need to use both  the modules
```
github.com/metal3-io/baremetal-operator
github.com/metal3-io/baremetal-operator/apis
```


dhellmann:
Can we put the module in the `/apis` directory so that if we add more APIs later and when we inevitably update the version (we've been v1a1 for way too long) we can avoid adding more modules?

kashifest:
> Can we put the module in the `/apis` directory so that if we add more APIs later and when we inevitably update the version (we've been v1a1 for way too long) we can avoid adding more modules?

Agreed. Good point. I have changed it.  Please take a look.

zaneb:
No idea how this multiple modules stuff works, but I support the goal and this looks as plausible as anything :)
/approve

PR #859, opened 2021-04-16
Add cert-manager setup for TLS in ironic deployment.

smoshiur1237:
Ironic deployment doesn't have cert-manager setup for TLS and this PR will add the cert-manager for  TLS.

### DIFF
--- a/ironic-deployment/certmanager/certificate.yaml
+++ b/ironic-deployment/certmanager/certificate.yaml
@@ -1,0 +1,69 @@
+apiVersion: cert-manager.io/v1
+kind: Issuer
+metadata:
+  name: selfsigned-issuer
+  namespace: capm3-system
+spec:
+  selfSigned: {}
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: ironic-cacert
+  namespace: capm3-system
+spec:
+  isCA: true
+  ipAddresses:
+  - IRONIC_HOST_IP
+  issuerRef:
+    kind: Issuer
+    name: selfsigned-issuer
+  secretName: ironic-cacert
+---
+apiVersion: cert-manager.io/v1
+kind: Issuer
+metadata:
+  name: ca-issuer
+  namespace: capm3-system
+spec:
+  ca:
+    secretName: ironic-cacert
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: ironic-cert
+  namespace: capm3-system
+spec:
+  ipAddresses:
+  - IRONIC_HOST_IP
+  issuerRef:
+    kind: Issuer
+    name: ca-issuer
+  secretName: ironic-cert
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: ironic-inspector-cert
+  namespace: capm3-system
+spec:
+  ipAddresses:
+  - IRONIC_HOST_IP
+  issuerRef:
+    kind: Issuer
+    name: ca-issuer
+  secretName: ironic-inspector-cert
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: mariadb-cert
+  namespace: capm3-system
+spec:
+  ipAddresses:
+  - MARIADB_HOST_IP
+  issuerRef:
+    kind: Issuer
+    name: ca-issuer
+  secretName: mariadb-cert
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -115,15 +115,15 @@ func init() {
 	}
 	ironicTrustedCAFile = os.Getenv("IRONIC_CACERT_FILE")
 	if ironicTrustedCAFile == "" {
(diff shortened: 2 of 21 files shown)


smoshiur1237:
//test-centos-integration

tools/deploy.sh
```diff
@@ -115,91 +115,91 @@ if [ "${DEPLOY_BASIC_AUTH}" == "true" ]; then
...
-        cp "${MARIADB_CACERT_FILE}" "${IRONIC_TLS_SCENARIO}/mariadb-ca.crt"
-        cp "${IRONIC_CERT_FILE}" "${IRONIC_TLS_SCENARIO}/ironic.crt"
-        cp "${IRONIC_INSPECTOR_CERT_FILE}" "${IRONIC_TLS_SCENARIO}/ironic-inspector.crt"
-        cp "${MARIADB_CERT_FILE}" "${IRONIC_TLS_SCENARIO}/mariadb.crt"
-        cp "${IRONIC_KEY_FILE}" "${IRONIC_TLS_SCENARIO}/ironic.key"
-        cp "${IRONIC_INSPECTOR_KEY_FILE}" "${IRONIC_TLS_SCENARIO}/ironic-inspector.key"
-        cp "${MARIADB_KEY_FILE}" "${IRONIC_TLS_SCENARIO}/mariadb.key"
-    fi
-fi
+#     if [ "${DEPLOY_IRONIC}" == "true" ]; then
```
macaptain, line 184: Would it be better to delete these commented out lines? By the looks of it, after your change they won't be needed again.

smoshiur1237: Yes I will do that once I do have a successful local run.

namnx228 (approved):
LGTM

ironic-deployment/certmanager/certificate.yaml
```diff
@@ -0,0 +1,69 @@
...
+    secretName: ironic-cacert
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: ironic-cert
+  namespace: capm3-system
+spec:
+  ipAddresses:
+  - 172.22.0.2
```
kashifest, line 39: I am a bit worried about hardcoding this IP? What might be the consequence if we use parameterization through configmap?

namnx228: Not sure if there is a way for cert-manager to get the IP from a configmap. It will be nice if there is. Otherwise, hardcoding the IP is the only way IMO

smoshiur1237: I was trying to parameterize that variable and seems that ipAddresses in cert-manager is accepting only string value and didn't have a way to use configmap. I do have the same opinion as Nam.

kashifest: ok lets tackle this separately.

kashifest: I am referring to something similar to https://github.com/kubernetes-sigs/cluster-api/blob/master/config/certmanager/certificate.yaml#L19

pkg/provisioner/ironic/ironic.go
```diff
@@ -115,7 +115,7 @@ func init() {
 	}
 	ironicTrustedCAFile = os.Getenv("IRONIC_CACERT_FILE")
 	if ironicTrustedCAFile == "" {
-		ironicTrustedCAFile = "/opt/metal3/certs/ca/crt"
+		ironicTrustedCAFile = "/opt/metal3/certs/ca/tls.crt"
```
s3rj1k, line 118: Why name change? This breaks a pattern for other `/opt/metal3/certs/*` files

smoshiur1237: The secret which is generated by cert-manager and when we mount the secret to the container the CA file is tls.crt and we cannot change that with cert-manager. So we need to change the `ironicTrustedCAFile` to `"/opt/metal3/certs/ca/tls.crt"`.   can you please explain a bit more, what will be broken, if the name is changed?

s3rj1k: You can set custom paths using ENV

s3rj1k: Not broken, please see other names of files related to certs, naming is different, this will look not pretty

s3rj1k: ironicClientCertFile = "/opt/metal3/certs/client/crt"

ironicClientPrivKeyFile = "/opt/metal3/certs/client/key"

s3rj1k: My point is, if we change naming scheme for one of this, lets make this uniform for all of them

smoshiur1237: I think I got your point. will update the PR and have the same naming change in other places.

namnx228 (approved):
LGTM

fmuyassarov (commented):
As this is no way to to pass the integration CI before other dependent PRs are also merged, we need to merge them all at the same time. Notice was also given on the [slack](https://kubernetes.slack.com/archives/CHD49TLE7/p1621881614043400) and in the community meeting. Force merging it.

PR #860, opened 2021-04-19
Propagate IPA_BASEURI env var to local ironic containers

macaptain:
This will let you set IPA_BASEURI in your environment in metal3-dev-env and have this variable propagate to the ironic-ipa-downloader container, so that the Ironic Python Agent can be downloaded from a chosen location.


fmuyassarov (commented):
/lgtm

tools/run_local_ironic.sh
```diff
@@ -8,6 +8,7 @@ IRONIC_IMAGE=${IRONIC_IMAGE:-"quay.io/metal3-io/ironic:master"}
 IRONIC_INSPECTOR_IMAGE=${IRONIC_INSPECTOR_IMAGE:-"quay.io/metal3-io/ironic-inspector"}
 IRONIC_KEEPALIVED_IMAGE=${IRONIC_KEEPALIVED_IMAGE:-"quay.io/metal3-io/keepalived"}
 IPA_DOWNLOADER_IMAGE=${IPA_DOWNLOADER_IMAGE:-"quay.io/metal3-io/ironic-ipa-downloader:master"}
+IPA_BASEURI=${IPA_BASEURI:-}
```
maelk, line 11: Should this have a default ? It was defaulted within the container, but now if the user does not provide it, will it still be defaulted properly

macaptain: My expectation is that if `IPA_BASEURI` is not set, the variable will be empty here, will be empty in the vars file, and therefore when the ironic-ipa-downloader image is run, we will fall back to the default in the ironic-ipa-downloader get-resource.sh script:

https://github.com/metal3-io/ironic-ipa-downloader/blob/master/get-resource.sh#L11

I don't want to duplicate this default here in baremetal-operator, because it will override the default in the image, which could be confusing if the default in get-resource.sh changes at some point in the future.

macaptain: The integration test passed which demonstrates that this change in itself with the current empty default is a no-op for the tests.

dtantsur (approved):

fmuyassarov:
@zaneb would you mind to merge it manually as it seems it doesn't get merged automatically.

PR #851, opened 2021-04-12
Define command for ironic-inspector run

elfosardo:
Make run command for ironic-inspector explicit and define the entrypoint
for local run.

PR #861, opened 2021-04-19
Point ironic-inspector to the ironic image in quay

elfosardo:
After ironic-inspector merges with ironic image, we need to change the
reference of the ironic-inspector container to point to the ironic image
in quay.

This should happen after https://github.com/metal3-io/ironic-image/pull/253


config/basic-auth/default/credentials_patch.yaml
```diff
@@ -13,12 +13,12 @@ spec:
           mountPath: "/opt/metal3/auth/ironic"
           readOnly: true
         - name: ironic-inspector-credentials
-          mountPath: "/opt/metal3/auth/ironic-inspector"
+          mountPath: "/opt/metal3/auth/ironic"
```
fmuyassarov, line 16: Sorry if I'm missing something, why do we need to change the `mountPath`?

fmuyassarov (commented):
LGTM apart from one question below.

ironic-deployment/ironic/ironic.yaml
```diff
@@ -85,13 +85,13 @@ spec:
             - mountPath: /shared
               name: ironic-data-volume
         - name: ironic-inspector
-          image: quay.io/metal3-io/ironic-inspector
+          image: quay.io/metal3-io/ironic
```
andfasano, line 88: I'll put https://github.com/metal3-io/baremetal-operator/pull/851/files#diff-acbaf87bacacde2d8867e2e1629c604256f8c9ebf6c900248f5d8c1620694fabR91 as well here in the same PR

config/basic-auth/default/credentials_patch.yaml
elfosardo: this is absolutely a mistake, was not done on purpose
I'll change it back

ironic-deployment/ironic/ironic.yaml
elfosardo: I see the point, but I'd rather keep the two changes separated just for greater clarity
The merge is pretty complex, involves various components, and I think having small changes is better verifiable so if something goes wrong  we have better chance to know what and where

dtantsur:
The dependency has merged, but the quay build has failed :(

fmuyassarov (commented):
/lgtm

PR #862, opened 2021-04-20
Add unit and lint test in travis

kashifest:
Enabling unit and lint tests in travis temporarily until praw is fixed.


furkatgofurov7 (commented):
Thanks, this will unblock other works going in until we have prow cluster back in place. LGTM.

PR #863, opened 2021-04-20
Fix travis lint test

zaneb:
Oops.

### DIFF
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -433,8 +433,6 @@ func (p *ironicProvisioner) ValidateManagementAccess(data provisioner.Management
 	driverInfo["deploy_kernel"] = deployKernelURL
 	driverInfo["deploy_ramdisk"] = deployRamdiskURL
 
-	result, err = operationComplete()
-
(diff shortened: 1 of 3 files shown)


s3rj1k:
@zaneb 
Maybe move this at the bottom of function, inside default case of switch?
```
result, err = operationComplete()
```

something like
```
	case nodes.Active:
		// The host is already running, maybe it's a master?
		p.debugLog.Info("have active host", "image_source", ironicNode.InstanceInfo["image_source"])
		return

	default:
		result, err = operationComplete()
		return
	}
}
```

zaneb:
@s3rj1k We'd have to repeat it 3 times or delete the other two cases that also return without overriding the return values, along with the associated comments.

One day I'll replace the return types from the provisioner with something better and we can put this code in its final form, but today I just want to stop Travis from telling everybody their PR is broken.

s3rj1k:
seems that bot does not work on me :)

furkatgofurov7 (commented):
Not sure, why tide did not work, let's try once again. 

/lgtm

zaneb:
I think we finally broke tide as well :)
I'll merge manually, thanks for the reviews.

PR #865, opened 2021-04-25
WIP: Enable bmh validation webhook

ardaguclu:
This PR enables BareMetalHost validation webhook to validate BareMetalHost CRD prior to create and update operations.


main.go
```diff
@@ -97,6 +105,8 @@ func main() {
 		"use the demo provisioner to set host states")
 	flag.StringVar(&healthAddr, "health-addr", ":9440",
 		"The address the health endpoint binds to.")
+	flag.IntVar(&webhookPort, "webhook-port", 0,
```
zaneb, line 108: What's the reason for defaulting it to disabled?

config/webhook/manifests.yaml
```diff
@@ -0,0 +1,25 @@
...
+  creationTimestamp: null
+  name: validating-webhook-configuration
+webhooks:
+  - clientConfig:
+      caBundle: Cg==
+      service:
+        name: webhook-service
+        namespace: system
+        path: /validate-metal3-io-v1alpha1-baremetalhost
+    failurePolicy: Ignore
```
zaneb, line 14: I guess we have got this far without validation at all, but to my mind this ought to be something we can rely on in future, not something that will _maybe_ enforce some invariants. Are there other opinions on this?

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -0,0 +1,53 @@
...
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/util/errors"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+// log is for logging in this package.
+var log = logf.Log.WithName("baremetalhost-validation")
+
+func (host *BareMetalHost) Validate(old *BareMetalHost) error {
```
zaneb, line 13: AIUI putting implementing code in the API package is frowned-upon, I think mostly because it generally grows to start pulling in dependencies that other packages don't want when they just need to vendor the API. (I'm aware that we already have code in the API package, but people frowned upon that which is how we found out.)

I'd suggest we put this code in a new pkg/validation module.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -0,0 +1,53 @@
...
+	if len(r.HardwareRAIDVolumes) > 0 && len(r.SoftwareRAIDVolumes) > 0 {
+		return fmt.Errorf("hardwareRAIDVolumes and softwareRAIDVolumes can not be set at the same time")
+	}
+
+	return nil
+}
+
+func validateUpdate(old, new *BareMetalHost) []error {
+	var errs []error
+	if new.Spec.BMC.Address != old.Spec.BMC.Address {
```
zaneb, line 44: Need to check that the old BMC address was actually set.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -0,0 +1,53 @@
...
+		errs = append(errs, fmt.Errorf("BMC address can not be changed once it is set"))
+	}
+
+	if new.Spec.BootMACAddress != old.Spec.BootMACAddress {
```
zaneb, line 48: ditto with the old Boot MAC address.

zaneb (commented):
This is great, thanks Arda!!
For review purposes, I think it would be helpful to put the implementation of specific validations into separate commits within this PR. Updating dependencies in a separate commit helps too.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
ardaguclu: Agree. This is a good idea.

main.go
ardaguclu: I did not want to affect make run command on local. Maybe I can enable webhook by default and deliberately disable on make run?

zaneb: That sounds plausible. I wonder what other projects do...

ardaguclu: I checked cluster-api-provider-metal3 as an example and it also defaults to disabled.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
ardaguclu: Unfortunately, apis/ package has no access to pkg/ package. After checking other repos, it seems that validation should stay under apis/.

ardaguclu:
Thanks @zaneb for review. I changed the commit history as you suggested.

main.go
furkatgofurov7: Hi! Just FYI, I am not sure if we want to leave it disabled in BMO, but if you are referring to [this](https://github.com/metal3-io/cluster-api-provider-metal3/blob/master/main.go#L78) in CAPM3 which is disabled by default, for now, it will be changed soon as part of preparing for CAPI v1alpha4 changes in CAPM3. You can check this doc for [ref](https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/book/src/developer/providers/v1alpha3-to-v1alpha4.md#required-kustomize-changes-to-have-a-single-manager-watching-all-namespaces-and-answer-to-webhook-calls) which includes `Changes in the main.go file`.

ardaguclu: thanks @furkatgofurov7, I'll change that way and handle `make run` works without webhook.

ardaguclu: There is a bug in kustomize to merge args parameters and if I change it to enabled by default, there will no longer need to upgrade kustomize to 3.10.0.

main.go
```diff
@@ -147,6 +157,10 @@ func main() {
 
 	setupChecks(mgr)
 
+	if webhookPort != 0 {
```
andfasano, line 160: Why do we need this?

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -0,0 +1,53 @@
...
+
+func (host *BareMetalHost) Validate(old *BareMetalHost) error {
+	log.Info("validate", "name", host.Name)
+	var errs []error
+
+	if err := validateRAID(host.Spec.RAID); err != nil {
+		errs = append(errs, err)
+	}
+
+	if old != nil {
```
andfasano, line 21: Wondering if it's really required to flatten create / update validations now. Looks like the code would be simpler if we'll keep them separated

main.go
ardaguclu: We need a way to disable webhook somehow(e.q. local runs) and cluster-api-provider-metal3 uses this way.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
ardaguclu: In conceptual aspect, update validations also cover create validations. In any cases, all create validations will be executed for also update. I mean even if we flatten this code block, eventually to prevent duplications we are going to conflate them again.

main.go
andfasano: Not sure if you have used kubebuilder or not for scaffolding the webhook, but wouldn't be clear to have something similar [approach](https://book.kubebuilder.io/cronjob-tutorial/main-revisited.html)?

ardaguclu: Good point @andfasano, that usage seems more clear. I'll change that way.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
andfasano: For a given resource, the create/delete handler is invoked only once, while the update many. So, once the creation step has been validated, it's not clear why such validation should be repeated also for the update _always_. Of course, there could be validation rules that are meaningful for both the create and update handlers, but there could be other that are valid only in just one of them. For example in the current code the BMC update rule makes sense only during an update.

zaneb: I agree that basically all of the validations in ValidateCreate() are also going to be required in ValidateUpdate() (the only exceptions are stuff that is immutable like the resource name... actually it's just that I think). But I agree with Andrea that there doesn't seem much point taking them from two separate functions, squashing them into one, and then splitting them again based on whether the argument is nil. Like, you could just call ValidateCreate() at the start of ValidateUpdate() and save a lot of boilerplate?

fmuyassarov:
Triggering prow jobs, as it is back on track.
/retest

fmuyassarov:
Please ignore shellcheck failure, this is totally unrelated to this PR. We have a fix for that in https://github.com/metal3-io/baremetal-operator/pull/869/commits/514cd5084ae1b1d077f7b6ecab67d384f41096c2

apis/metal3.io/v1alpha1/baremetalhost_validation.go
ardaguclu: I totally understand your concerns @zaneb and I spent some time about thinking and trying the problem. We have baremetalhost and baremetalhost_webhook under the apis directory and let's say we have a validation module under pkg.

apis/
* baremetalhost
* baremetalhost_webhook

pkg/
* validation

baremetalhost_webhook depends on validation
validation depends on baremetalhost

and this creates import cycle and not being compiled.

One option comes to my mind is move everything related to webhook under the pkg(I have not tried yet and not sure it is standard way or workaround).

I asked this problem on kubebuilder slack channel maybe we can get an idea about it.

andfasano: I've also have some concerns in putting more implementation code under the `/apis` - and in particular the webhooks.
I'm wondering if it's the case to review the current package organization to reduce the api surface, but it sounds something that could go out from the scope of the current PR.
Another approach could be for now to extract the validation code in a separate type, ie `BaremetalHostValidator`, so that it could be defined in a different package without the need of touching `apis/`

apis/metal3.io/v1alpha1/baremetalhost_validation.go
zaneb: Looking deeper, it might make sense to have a common function that returns `[]error` so that both types of error can be combined in a single list, and that's fine too.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
zaneb: I had a look through the controller-runtime code, and it seems that there are 3 requirements for the thing we pass to `NewWebhookManagedBy(mgr).For()`:

1. It must implement the [Validator](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/webhook/admission/validator.go#L30-L35) interface (i.e. have all of the `ValidateCreate()` &c. methods)
2. You must be able to unpack the BareMetalHost data into it
3. It must return the right group/version/kind when passed to [`GVKForObject()`](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/apiutil/apimachinery.go#L75)

So I _think_ it would work to define a new type in another package:

```
type BareMetalHostValidator struct {
    v1alpha1.BareMetalHost
}

func (b *BareMetalHostValidator) ValidateCreate() error {
...
```

This is the way to go if it works.

ardaguclu: Thanks for your valuable comments. I think we all agree that webhook specific code will not be under `apis/`. 
My proposal is;
* First let me move current code under `pkg/` without creating a new `BaremetalHostValidator`. If it works, we don't have to create a struct for that. Everything under `pkg/` depends on `api/`
* If it does not work, it seems that `BaremetalHostValidator` might solve our problem.

zaneb: You can't add a new method to a type defined in another package in golang, so you will definitely have to create a new struct in order to move the code.

ardaguclu: Sorry, I forgot that webhook.Validator is BareMetalHost not new struct;
`var _ webhook.Validator = &BareMetalHost{}`

ok. I'll implement BaremetaHostValidator, thanks.

ardaguclu: This method did not work. In the setup register stage, it returns `kind BareMetalHostValidator not found` error(Because there is none). If `BareMetalHostValidator` is casted to `BareMetalHost`, then it returns `BareMetalHost` is not implemented as admission webhook interface. It is correct because `BareMetalHostValidator` has been implemented with that interface not `BareMetalHost`.

Since there isn't really any example about this usage for kubebuilder(and kubebuilder slack channel there was a comment about because time constraints it is designed to do simple validations, hence there is no such use case example). I started thinking that, maybe we can put these validations under the `apis/` folder. And we can have a chance to move it under `pkg/` when changing the folder structure as @andfasano stated.

asalkeld: @zaneb I'd be concerned if you have 2 types returning the same gvk in the same process.
you are going to have really make sure the scheme is separate so the controller is getting the correct events. I'd suggest that controller-runtime *really* expects you to have the validation code in the api - and don't try to fight it...

andfasano: Btw, I also checked the controller-runtime, and IINW it appears that the dependency it's even [more strict](https://github.com/kubernetes-sigs/controller-runtime/blob/b704f447ea7c8f7059c6665143a4aa1f6da28328/pkg/webhook/admission/validator.go#L66), ie the decoded object from the admission request must implement the `Validator` interface.

andfasano: Yes, my point is if the underlying serialized object it's a BMH then there wouldn't be any way to workaround it (if not, there's still the registration point).
So to summarize it looks like there could be just two approaches:
* Put everything under `/apis`
* Refactor the packages structures

apis/metal3.io/v1alpha1/baremetalhost_webhook.go
```diff
@@ -30,14 +31,14 @@ func (r *BareMetalHost) SetupWebhookWithManager(mgr ctrl.Manager) error {
 		Complete()
 }
 
-//+kubebuilder:webhook:verbs=create;update,path=/validate-metal3-io-v1alpha1-baremetalhost,mutating=false,failurePolicy=fail,groups=metal3.io,resources=baremetalhosts,versions=v1alpha1,name=vbaremetalhost.kb.io
+//+kubebuilder:webhook:verbs=create;update,path=/validate-metal3-io-v1alpha1-baremetalhost,mutating=false,failurePolicy=fail,sideEffects=none,admissionReviewVersions=v1;v1beta1,groups=metal3.io,resources=baremetalhosts,versions=v1alpha1,name=vbaremetalhost.kb.io
```
zaneb, line 34: We don't have a v1 API.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -3,28 +3,43 @@ package v1alpha1
...
 
-	if old != nil {
-		if updateErrs := validateUpdate(old, host); updateErrs != nil {
-			errs = append(errs, updateErrs...)
-		}
+	return errs
+}
+
+// validateUpdate validates BareMetalHost resource for update
+// but also covers the validations of creation
```
zaneb, line 25: Maybe we should call them validateHost() and validateChanges() or something like that?

zaneb (commented):
Do you plan to reorganise the commits in this PR?

/test-integration

main.go
```diff
@@ -97,6 +106,9 @@ func main() {
 		"use the demo provisioner to set host states")
 	flag.StringVar(&healthAddr, "health-addr", ":9440",
 		"The address the health endpoint binds to.")
+	flag.IntVar(&webhookPort, "webhook-port", 9443,
+		"Webhook Server port (set to 0 to disable)")
+	flag.BoolVar(&enableWebhook, "enable-webhook", true, "Enable webhook (default is True)")
```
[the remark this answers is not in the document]
ardaguclu: AFAIR, @andfasano suggested it and it made sense to me(flag seems like showing the purpose more clear). But I could not find the conversation in anywhere. I'm fine with both options.

apis/metal3.io/v1alpha1/baremetalhost_webhook.go
ardaguclu: I think this is for `WebhookConfigurations`, currently we are using v1 and we support v1beta1 too with `admissionReviewVersions=v1;v1beta1`

zaneb: Oh, OK, it's versioning for the data that is posted to the webhook:
https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#request

Carry on then.

main.go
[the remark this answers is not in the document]
zaneb: It's a user interface, we shouldn't have two ways of doing the same thing that both have to be set in unison.

For readability we can do:

    enableWebhook := webhookPort != 0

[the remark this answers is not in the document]
andfasano: My main concern here it's just about the readability of the user interface and not the code: I think it's much more clear and easy to understand to have an explicit speaking flag (maybe with a default value bound to an env var, or something like [that](https://book.kubebuilder.io/multiversion-tutorial/webhooks.html)) instead of setting a port value to 0, which could have potentially other meanings to the eyes of the reader.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -0,0 +1,55 @@
...
+func (host *BareMetalHost) validateChanges(old *BareMetalHost) []error {
+	log.Info("validate update", "name", host.Name)
+	var errs []error
+
+	if err := host.validateHost(); err != nil {
+		errs = append(errs, err...)
+	}
+
+	if old.Spec.BMC.Address != "" && host.Spec.BMC.Address != old.Spec.BMC.Address {
+		errs = append(errs, fmt.Errorf("BMC address can not be changed once it is set"))
```
dtantsur, line 35: Why so? It's not hard to make a typo in the BMC address. Maybe only prevent changing the driver part?

dtantsur (commented):
This is great, I have one comment inline.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
zaneb: > Why so?

Because making a BareMetalHost to point to one server, getting it to particular state, and then changing it to point to a different server as if it were the same one is... suboptimal.

> It's not hard to make a typo in the BMC address.

You can delete and recreate.

apis/metal3.io/v1alpha1/baremetalhost_webhook.go
```diff
@@ -32,7 +32,7 @@ func (r *BareMetalHost) SetupWebhookWithManager(mgr ctrl.Manager) error {
 		Complete()
 }
 
-//+kubebuilder:webhook:verbs=create;update,path=/validate-metal3-io-v1alpha1-baremetalhost,mutating=false,failurePolicy=fail,sideEffects=none,admissionReviewVersions=v1;v1beta1,groups=metal3.io,resources=baremetalhosts,versions=v1alpha1,name=vbaremetalhost.kb.io
+//+kubebuilder:webhook:verbs=create;update,path=/validate-metal3-io-v1alpha1-baremetalhost,mutating=false,failurePolicy=ignore,sideEffects=none,admissionReviewVersions=v1;v1beta1,groups=metal3.io,resources=baremetalhosts,versions=v1alpha1,name=vbaremetalhost.kb.io
```
zaneb: As [discussed previously](https://github.com/metal3-io/baremetal-operator/pull/865#discussion_r622207670), I think that the failure policy should be fail, not ignore.

apis/metal3.io/v1alpha1/baremetalhost_webhook.go
```diff
@@ -0,0 +1,50 @@
...
+// log is for logging in this package.
+var baremetalhostlog = logf.Log.WithName("baremetalhost-resource")
+
+func (r *BareMetalHost) SetupWebhookWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewWebhookManagedBy(mgr).
+		For(r).
+		Complete()
+}
+
+//+kubebuilder:webhook:verbs=create;update,path=/validate-metal3-io-v1alpha1-baremetalhost,mutating=false,failurePolicy=fail,groups=metal3.io,resources=baremetalhosts,versions=v1alpha1,name=vbaremetalhost.kb.io
```
zaneb: Is the name here something we are supposed to be customising?

zaneb (commented):
This looks about ready to drop the WIP label to me.

apis/metal3.io/v1alpha1/baremetalhost_webhook.go
ardaguclu: Yes, it is customizable.

apis/metal3.io/v1alpha1/baremetalhost_webhook.go
ardaguclu: This is the reason of WIP label. Because there seem to be a couple of bugs in different places when we set FailurePolicy to fail. This PR https://github.com/metal3-io/metal3-dev-env/pull/683 is prerequisite to integration tests pass and switch back to FailurePolicy to fail.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	if old.Spec.BootMACAddress != "" && host.Spec.BootMACAddress != old.Spec.BootMACAddress {
```
andfasano, line 38: (not blocking) for the sake of maintenance I'd find useful to have a separate BMH method for each distinct validation rule, ie `validateBMCAddress(), validateBootMacAddress()` and so on

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -0,0 +1,55 @@
...
+	if err := validateRAID(host.Spec.RAID); err != nil {
+		errs = append(errs, err)
+	}
+
+	return errs
+}
+
+// validateChanges validates BareMetalHost resource on changes
+// but also covers the validations of creation
+func (host *BareMetalHost) validateChanges(old *BareMetalHost) []error {
```
andfasano, line 26: Still on the naming side, I'd find more clear a name like `validateUpdate` for this method

apis/metal3.io/v1alpha1/baremetalhost_webhook.go
```diff
@@ -0,0 +1,59 @@
...
+func (r *BareMetalHost) ValidateCreate() error {
+	baremetalhostlog.Info("validate create", "name", r.Name)
+	return errors.NewAggregate(r.validateHost())
+}
+
+// ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
+func (r *BareMetalHost) ValidateUpdate(old runtime.Object) error {
+	baremetalhostlog.Info("validate update", "name", r.Name)
+	bmh, casted := old.(*BareMetalHost)
+	if !casted {
```
andfasano, line 49: Wondering if such type conversion could ever fail...

main.go
```diff
@@ -97,12 +105,16 @@ func main() {
 		"use the demo provisioner to set host states")
 	flag.StringVar(&healthAddr, "health-addr", ":9440",
 		"The address the health endpoint binds to.")
+	flag.IntVar(&webhookPort, "webhook-port", 9443,
+		"Webhook Server port (set to 0 to disable)")
```
andfasano, line 109: I'm still continuing to find the decision to use the value `0` for a port to disable a whole new feature not a good interface design: it's not self-speaking (as the previously discussed `enable-webhook`), and could convey a misleading message (since it could be interpreted as _"choose the best port"_).
Configuration settings are effectively  part of the public API exposed by the component to the user, so having a dedicated setting to enable/disable a feature and another one to configure it (when enabled) could be clearer

ardaguclu:
@zaneb, although we fixed a couple of problems and prepared validating webhook for use, integration tests are still failing because of intermittent `connection refused` failures of webhook API. This seems like the same error we got https://github.com/kubernetes/kubernetes/issues/102013. This also can be due to certificate problems between namespaces(I don't have proof yet).

My suggestion is, setting `failPolicy` to `ignore`. In any case, webhook will work(like on my local env), but we'll guarantee that integration tests will work. On the other hand, even if we solve this problem, I worry about creating flakiness.

ardaguclu:
When setting `failurePolicy` to `fail`, integration tests are failing. Because validating webhook API throws `connection refused` error sometimes. For example, it works in first bmh apply but after that it starts failing.
There are possible reasons for that;

- All webhooks are running under `capi-webhook-system` namespace, that duplicates controller managers but it seems working without any problem. BMO is running under capi-system namespace. One possible solution is running second BMO controller like the others under capi-webhook-system namespace by duplicating BMO controller manager.
- Kustomize applies in an predefined unchangeable order. I detected ValidatingWebhook of BMO is firstly applied(before certificates, issuers ,etc.) unlike others. This might be the reason.
- It sometimes fails might show that service or dns problems(I saw coredns's readiness probe is failing a couple of times, but others webhooks should not be working in that case).

My suggestion is setting failurePolicy to ignore for now. I can create a new task to set it to fail and assign it to myself.
- Validating webhook of BMO, when it is run as standalone, works without any problem.
- Validating webhook of BMO, when it is run under CAPM3, might fail sometimes and if we set to ignore, it just does not validate and not affect users.

ardaguclu:
I filed an issue for that, maybe Kubernetes community has an idea about this https://github.com/kubernetes/kubernetes/issues/103108

main.go
asalkeld: i think this is quite standard when using controller-runtime

asalkeld (commented):
/lgtm

config/manager/manager.yaml
```diff
@@ -5,15 +5,18 @@ metadata:
   namespace: system
   labels:
     control-plane: controller-manager
+    webhook: controller-manager
 spec:
   selector:
     matchLabels:
       control-plane: controller-manager
+      webhook: controller-manager
```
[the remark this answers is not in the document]
ardaguclu: This is the last piece of puzzle, let me explain why. Before that, I'd like to say this is manually added and totally customizable and is not needed if BMO runs standalone.

Capm3 runs three controller-managers under `capm3-system` namespace;
- capm3-controller-manager
- capm3-ipam-controller-manager
- capm3-baremetal-operator-controller-manager

These managers are all have `control-plane: controller-manager` labels. If we enable webhook by applying this service patch https://github.com/metal3-io/baremetal-operator/blob/22484ec5ff91aeaac912cf7188e8af55fcee099a/config/webhook/service_patch.yaml#L12, webhook traffic is routed to each of these managers, because they all have matching labels. 2/3 of webhook requests are failing. Because only baremetal-operator-controller-manager serves webhook endpoint(and others throw connection refused error).

I added this to distinguish baremetal-operator-controller-manager and webhook requests only goes to this manager. The other option is changing baremetal-operator-controller-managers label(i.e. `control-plane: bmo-controller-manager`).

Or there is another way I don't know.

apis/metal3.io/v1alpha1/baremetalhost_webhook.go
```diff
@@ -0,0 +1,59 @@
...
+// log is for logging in this package.
+var baremetalhostlog = logf.Log.WithName("baremetalhost-resource")
+
+func (r *BareMetalHost) SetupWebhookWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewWebhookManagedBy(mgr).
+		For(r).
+		Complete()
+}
+
+//+kubebuilder:webhook:verbs=create;update,path=/validate-metal3-io-v1alpha1-baremetalhost,mutating=false,failurePolicy=fail,sideEffects=none,admissionReviewVersions=v1;v1beta,groups=metal3.io,resources=baremetalhosts,versions=v1alpha1,name=vbaremetalhost.kb.io
```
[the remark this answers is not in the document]
ardaguclu: It fails with `webhook-configuration: ValidatingWebhookConfiguration.admissionregistration.k8s.io \"capm3-baremetal-operator-validating-webhook-configuration\" is invalid: webhooks[0].name: Invalid value: \"metal3.io\": should be a domain with at least three segments separated by dots` 

We need to add something kb.metal3.io? :)

config/manager/manager.yaml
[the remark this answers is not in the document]
ardaguclu: According to the @asalkeld suggestion, I removed new label from selector part. Because @zaneb you are right, that might cause problems in upgrade.

main.go
andfasano: For enabling / disabling specifically webhooks? Kubebuilder [sample](https://book.kubebuilder.io/cronjob-tutorial/running.html?highlight=ENABLE_WEB#running-and-deploying-the-controller) uses an env var (which could be handy especially when testing without the webhook) whereas some operators could use a more explicit approach.

Anyhow, on a second thought, I'm wondering if this kind of configuration (the webhook port) it's really required or if it could be avoided at all.

apis/metal3.io/v1alpha1/baremetalhost_validation_test.go
```diff
@@ -0,0 +1,135 @@
...
+						SoftwareRAIDVolumes: []SoftwareRAIDVolume{
+							{
+								SizeGibibytes: nil,
+								Level:         "",
+								PhysicalDisks: nil,
+							},
+						},
+					}}},
+			oldBMH:    nil,
+			wantedErr: "hardwareRAIDVolumes and softwareRAIDVolumes can not be set at the same time",
```
apinnick, line 72: Suggest you change "can not" to "cannot".

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -0,0 +1,55 @@
...
+	return errs
+}
+
+func validateRAID(r *RAIDConfig) error {
+	if r == nil {
+		return nil
+	}
+
+	if len(r.HardwareRAIDVolumes) > 0 && len(r.SoftwareRAIDVolumes) > 0 {
+		return fmt.Errorf("hardwareRAIDVolumes and softwareRAIDVolumes can not be set at the same time")
```
apinnick, line 51: Suggest you change "can not" to "cannot"

config/default/kustomization.yaml
```diff
@@ -18,9 +18,9 @@ bases:
 - ../manager
 # [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix including the one in 
 # crd/kustomization.yaml
-#- ../webhook
+- ../webhook
 # [CERTMANAGER] To enable cert-manager, uncomment all sections with 'CERTMANAGER'. 'WEBHOOK' components are required.
-#- ../certmanager
+- ../certmanager
```
zaneb, line 23: Is there a reason we are also enabling certmanager in this PR? That seems to be what is failing the integration tests.

ardaguclu: This provides certificates and related resources for webhook api serves in https(which is mandatory). Integration test is now failing because previously cert-manager crd resources were applied(certificate, issuer, etc. crds) and I guess now it is not.

ardaguclu:
After taking BMO out of CAPM3, there are a couple of PRs opened to enable validating webhook for BMO;
https://github.com/metal3-io/metal3-dev-env/pull/742 is for BMO in management cluster can access cert-manager
https://github.com/metal3-io/metal3-dev-env/pull/749 is for BMO in target cluster can access cert-manager
https://github.com/metal3-io/metal3-dev-env/pull/750 is for BMO in target cluster uses PR's branch instead master

https://github.com/metal3-io/cluster-api-provider-metal3/pull/263 is the backport to 0.4. Because in 0.4, BMO is in CAPM3 and it needs to be changed the certificate and dns names according to the capm3-system namespace.

config/manager/manager.yaml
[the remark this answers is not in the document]
ardaguclu: @zaneb now this PR also covers CAPM3 v1alpha4 and v1alpha5. The above change(adding new label) is only needed for v1alpha4. Since BMO should support both versions, it is still relevant.

furkatgofurov7:
This seems to be in a good shape now and has taken quiet some time to pass the CI tests due to other needed changes in different repositories. @ardaguclu do you want to squash your commits or planning to do some more changes, otherwise, it LGTM.

ardaguclu:
Thank you very much @furkatgofurov7. Let me squash my commits, because I don't have any pending changes.

furkatgofurov7 (commented):
/lgtm

Issue #654, opened 2020-09-25
[Feature] Add support for ansible as Ironic deploy interface


s3rj1k:
Would be nice to have the ability to trigger [ansible inside ironic for deployment](https://docs.openstack.org/ironic/ussuri/admin/interfaces/deploy.html#ansible-deploy).
[Currently only `direct` deploy_interfaces is supported.](https://github.com/metal3-io/ironic-image/blob/master/ironic.conf#L10)

Issue is for baremetal-operator to support this ability as ironic image itself can be custom build.

PR #867, opened 2021-04-29
Add provisioner passthrough

redixin:
Add provisioner agnostic field of type`apiextensions.JSON` to be able to implement any specific features in any underlying provisiner like the one described in #654

### DIFF
--- a/apis/metal3.io/v1alpha1/baremetalhost_types.go
+++ b/apis/metal3.io/v1alpha1/baremetalhost_types.go
@@ -19,6 +19,7 @@ package v1alpha1
 import (
 	"time"
 
+    apiextensions "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


dhellmann (commented):
We have tried very hard to make the metal3 APIs describe their inputs fully, and avoid having "JSON blob" fields that require the user to understand the implementation details of the provisioner to use our API. This would be a significant departure from that approach.

Before going ahead with something like this, we are going to need a more detailed discussion of a design for the idea of alternative provisioning backends, using reviews in the metal3-docs repository. Please write up a design doc there so we can work out API changes, etc. before reviewing any code here.

You may also be interested in reviewing https://github.com/metal3-io/metal3-docs/pull/180

redixin:
@dhellmann https://github.com/metal3-io/metal3-docs/pull/181

redixin:
> avoid having "JSON blob" fields that require the user to understand the implementation details of the provisioner to use our API

If we expose `ansible_extra` in metal3 API, we will be a bit more tied to ironic provisioner. Personally I don't think this is a problem, and ready to go with this approach as well.

Any other provisioner may just either fail explicitly or silently ignore unknown parameters.

PR #870, opened 2021-05-05
Document parameter to set boot iso source

elfosardo:
Document the configuration parameter IRONIC_BOOT_ISO_SOURCE to set where
the boot iso will be served from.
Set its default for local ironic deployment.

Depends on https://github.com/metal3-io/ironic-image/pull/261


romfreiman:
How it will be set in ZTP flow?

fmuyassarov (commented):
/lgtm

PR #871, opened 2021-05-06
Run apis/ tests in make unit

ardaguclu:
Currently `make test` and `make unit-cover` commands do not run unit tests under `apis` folder. This PR enables running tests under `apis` folder as well to ensure that code's stability.


dhellmann:
Do we need to make similar changes for the linter, code generator, etc.?

ardaguclu:
Actually we need, but to achieve that some packages are needed to be added in `apis/go.mod` like 
`github.com/golangci/golangci-lint`
`sigs.k8s.io/controller-tools v0.4.1`

s3rj1k:
>  `github.com/golangci/golangci-lint`

Why do you need to add linter as a package import, why not use binary build?

Adding linter will grow build deps to an insane level, please don't do that.

Makefile
```diff
@@ -64,6 +64,7 @@ test: generate lint manifests unit ## Run common developer tests
 .PHONY: unit
 unit: ## Run unit tests
 	go test ./... $(GO_TEST_FLAGS) -coverprofile $(COVER_PROFILE)
+	cd apis/ && go test ./... $(GO_TEST_FLAGS) -coverprofile $(COVER_PROFILE)
```
maelk, line 67: what about unit-cover ? and is the fmt check needing the same trick ?

ardaguclu: As far as I can see, these are not running for apis/; 
`make unit-cover`
`make manifests`
`make test`
`make lint`

I could not run `make fmt-check` (probably I'm doing something wrong).

ardaguclu:
@s3rj1k, thanks for your comment. Since they are now different modules because of separation of go.mods, it needs to be imported to run linter. Maybe I mistunderstand you, could you please elaborate what do you mean by running as a binary?

Because this `cd apis/ && $(GOLANGCI_LINT) run` just does not work.

s3rj1k:
@ardaguclu You do not need to import linter to actually use it, just install like the docs say
https://golangci-lint.run/usage/install/

dhellmann:
> @s3rj1k, thanks for your comment. Since they are now different modules because of separation of go.mods, it needs to be imported to run linter. Maybe I mistunderstand you, could you please elaborate what do you mean by running as a binary?
> 
> Because this `cd apis/ && $(GOLANGCI_LINT) run` just does not work.

We do have the linter in `go.mod` in the root of the repo, but we don't want it in the `apis` module. The whole point of creating this separate sub-module was to ensure it had the shortest possible list of dependencies.

s3rj1k:
@dhellmann This was my point, to make less dependencies, not more.

Also we could remove linter import from main go.mod.

There is no need to import it like this to use it. From previous experience with that linter, source builds tend to break, upstream also mentions this in docs and recommend using binary builds or docker image.

s3rj1k:
@ardaguclu Probably a good idea to use v1.32.0 golangci and update it in separate CR.

To make sure that newer version of linter does not trigger new unfixed issues with code.

ardaguclu:
@s3rj1k I run latest version of linter and it works fine without any warning.

ardaguclu:
I'd propose, maybe we can handle each of them in separate PRs;
`make lint` I created an issue about golint https://github.com/metal3-io/baremetal-operator/issues/872
`make manifests`
`make fmt`
and this PR solves `make test` and `make unit-cover`

maelk (approved):

PR #873, opened 2021-05-10
Revert: Point ironic-inspector to the ironic image in quay

fmuyassarov:
Some of the CI jobs (especially Centos based integration tests) are failing with a common issue as below. This reverts https://github.com/metal3-io/baremetal-operator/pull/861.
to unblock CI and keep it running without failure. It seems there is a conflict in the httpd port and this probably needs to be fixed in Ironic image. See the [slack discussion](https://kubernetes.slack.com/archives/CHD49TLE7/p1620663788347600).
```
ERROR 1146 (42S02) at line 1: Table 'ironic.conductor_hardware_interfaces' doesn't exist
AH00558: httpd: Could not reliably determine the server's fully qualified domain name, using fc00:f853:ccd:e793::1. Set the 'ServerName' directive globally to suppress this message
(98)Address already in use: AH00072: make_sock: could not bind to address [::]:6180
no listening sockets available, shutting down
AH00015: Unable to open logs

```

### DIFF
--- a/ironic-deployment/ironic/ironic.yaml
+++ b/ironic-deployment/ironic/ironic.yaml
@@ -85,7 +85,7 @@ spec:
             - mountPath: /shared
               name: ironic-data-volume
         - name: ironic-inspector
-          image: quay.io/metal3-io/ironic
+          image: quay.io/metal3-io/ironic-inspector
           imagePullPolicy: Always
           command:
             - /bin/runironic-inspector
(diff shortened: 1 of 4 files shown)


kashifest:
/lgtm
Lets unblock the CI and solve the issues separately.

furkatgofurov7:
> /hold cancel
> /approve
> 
> Let's take another look at the whole problem. I think we should end up with only one httpd container, not 2-3 we're starting now.
> 
> @fmuyassarov can we ensure that more CI jobs run by default, even if non-blocking? Or is it too much for our CI system?

We have been discussing this a lot internally, fully agree to have at least v1alpha4 centos job as mandatory in the repo.

PR #877, opened 2021-05-12
Add api generation in make generate

### LINKED ISSUE #875: Gomod for public API broke controller-tools object generation
After https://github.com/metal3-io/baremetal-operator/commit/f4865dbd25f288543513b1afb3f0e614070ad226
There is no simple way on regenerating objects for public API.

Current `make generate` does **not** generate objects for public API (silently ignores).

To actually see that object generation is broken, one would need to run
```
go run sigs.k8s.io/controller-tools/cmd/controller-gen object paths="./apis/..."
```

Output:
```
-: pattern ./apis/...: main module (github.com/metal3-io/baremetal-operator) does not contain package github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1
Error: not all generators ran successfully
run `controller-gen object paths=./apis/... -w` to see all available markers, or `controller-gen object paths=./apis/... -h` for usage
exit status 1

```

To manually regenerate objects, as a temporary measure, one would need to remove gomod related files from `apis` folder and remove related `requires` and `replaces` from repository root go.mod.

kashifest:
This PR adds api generation in generate Make target

Fixes #875


andfasano:
I think that this could be a reasonable approach until we'll keep the two go.mod separated. Probably not something that will scale well, but for now seems enough.

/lgtm

s3rj1k:
@andfasano But this brings new dep for API `sigs.k8s.io/controller-tools v0.5.0`
And it is only needed for code-generating, effective this makes separate go.mod useless, external projects would need to import a lot of unneeded dependencies just like it was before separate go.mod.

 Also note that version of controller tools are different 
 https://github.com/metal3-io/baremetal-operator/blob/master/go.mod#L19
 
 v0.5.0 VS v0.4.1 in core go.mod
 
 I do not think that this is acceptable, controller-tools version should be same across whole repo

kashifest:
> @andfasano But this brings new dep for API `sigs.k8s.io/controller-tools v0.5.0`
> And it is only needed for code-generating, effectively this makes separate go.mod useless, external projects would need to import a lot of unneeded dependencies just like it was before separate go.mod.
> 
> Also note that version of controller tools are different
> https://github.com/metal3-io/baremetal-operator/blob/master/go.mod#L19
> 
> v0.5.0 VS v0.4.1 in core go.mod
> 
> I do not think that this is acceptable, controller-tools version should be same across whole repo

I can change that

kashifest:
/test-integration
@s3rj1k @andfasano PTAL again. I think we can take it in to fix apis generation until we have an alternative.

Makefile
```diff
@@ -131,11 +131,13 @@ deploy: manifests ## Deploy controller in the configured Kubernetes cluster in ~
 .PHONY: manifests
 manifests: ## Generate manifests e.g. CRD, RBAC etc.
 	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
```
zaneb, line 133: AFAIK everything we need to generate is always in the apis directory, isn't it? I think we could remove this line (and the equivalent below).
That would at least solve the problem of trying to keep the controller-gen version the same in two different go.mod files.

kashifest: makes sense. @s3rj1k WDYT?

s3rj1k: https://github.com/metal3-io/baremetal-operator/pull/888/files

zaneb:
I opened #888 to do this without having to add dependencies to the apis module, so closing in favour of that per Kashif's comment https://github.com/metal3-io/baremetal-operator/pull/888#issuecomment-845784611.
/close

Issue #872, opened 2021-05-07
Run golint as a binary to remove dependency


ardaguclu:
Currently golint has been used as a dependency in baremetal operator. This will increase the binary size and to run golint for `apis/` module, it forces to be added as a dependency for `apis/` as well.

As @s3rj1k suggested, golint can be used as a binary(https://golangci-lint.run/usage/install/). Thanks to that we will reduce the binary size and be able to run apis module using this binary without adding any new dependency.

The whole purpose of this issue is;
`make lint` and `./hack/golint.sh` will also cover `apis/`

ardaguclu:
Closing this issue. Because currently golint also covers apis.

PR #878, opened 2021-05-13
Run golint as binary and remove from go.mod

ardaguclu:
This PR;

- Removes golangci-lint dependency from go.mod as stated in https://github.com/metal3-io/baremetal-operator/issues/872
- Changes `make lint` and `./hack/golint.sh` enabling golangci-lint commands also cover `apis/`.


Makefile
```diff
@@ -86,7 +85,7 @@ linters: lint generate-check fmt-check
 
 .PHONY: lint
 lint:
-	$(GOLANGCI_LINT) run
+	IS_CONTAINER=true ./hack/golint.sh
```
dhellmann, line 88: This line should not set IS_CONTAINER. The reason for having IS_CONTAINER in the scripts is so a developer can run the tools in containers locally to avoid conflicts if we need a different version of a tool than another project needs.

hack/golint.sh
```diff
@@ -5,10 +5,18 @@ set -eux
 IS_CONTAINER=${IS_CONTAINER:-false}
 CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-podman}"
 
+GOLANGCI_LINT=bin/golangci-lint
+if [ ! -f "${GOLANGCI_LINT}" ]; then
+    curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s v1.39.0
```
dhellmann, line 10: Line 26 refers to a specific golint image that we maintain on quay. Could we put this command in the Dockerfile that builds that image?

s3rj1k: @dhellmann How about using official golangci-lint docker image as a stage in builder image?

dhellmann: Sure, that makes sense, too.

PR #880, opened 2021-05-17
Fix error: missing the following IPMI credentials in node's driver_info: ['ipmi_address']

### LINKED ISSUE #879: baremetal IPI power_interface for irmc miss IPMI credentials
When we deployed bare metal using IPI, we encountered the following errors:
```
| last_error             | Failed to validate power driver interface for node e88070b5-afe2-4132-b697-93ee2bf10c1b. Error: Missing the following IPMI credentials in node's driver_info: ['ipmi_address']
```
Check that there is no ipmi related information in driver_info
```
| driver_info            | {​​​​​​​'deploy_kernel': 'http://192.168.20.2:80/images/ironic-python-agent.kernel', 'deploy_ramdisk': 'http://192.168.20.2:80/images/ironic-python-agent.initramfs', 'irmc_address': '192.168.30.25', 'irmc_password': '******', 'irmc_username': 'admin'}​​​​​​​ |
```
But in our previous use, there will be ipmi information in driver_info.

I think it may be related to this PR. https://github.com/metal3-io/baremetal-operator/pull/848

Hellcatlk:

Fixes #879


pkg/bmc/irmc.go
```diff
@@ -52,6 +52,9 @@ func (a *iRMCAccessDetails) DriverInfo(bmcCreds Credentials) map[string]interfac
 		"irmc_username": bmcCreds.Username,
 		"irmc_password": bmcCreds.Password,
 		"irmc_address":  a.hostname,
+		"ipmi_username": bmcCreds.Username,
+		"ipmi_password": bmcCreds.Password,
+		"ipmi_address":  a.hostname,
```
hardys, line 57: In #848 there was a link to https://opendev.org/openstack/ironic/src/branch/master/ironic/drivers/modules/irmc/common.py#L160-L174 mentioned - do we know why Ironic isn't copying these fields, since it seems like that code should be copying the irmc credentials?

dtantsur:
/ok-to-test

But I agree with the Steve's concern.

pkg/bmc/irmc.go
zaneb: The function is called from two places: the [iRMC power driver's `get_power_state`](https://opendev.org/openstack/ironic/src/branch/master/ironic/drivers/modules/irmc/power.py#L234-L246) and the [iRMC management driver's `validate`](https://opendev.org/openstack/ironic/src/branch/master/ironic/drivers/modules/irmc/management.py#L220-L232) methods.

#848 short-circuits the former by going directly to the ipmitool driver.

After reading the power state conductor persists the node to the DB, but after validating the management driver it looks like nothing gets persisted. That would explain @zhouhao3's findings that it worked when changing the power driver to ipmitool after initial creation, but not when changing it from the start.

In any event, this can't hurt.

zhouhao3:
After merging this PR, we performed the IPI installation again. This modification can modify the #879  problem. @hardys @zaneb @dtantsur  What do you think of this modification?

dtantsur:
/test-integration
/approve

I think the ironic situation should be sorted out, but I agree that it won't hurt to add the credentials.

zhouhao3:
@hardys  Can you update this PR to `openshift/baremetal-operator`? This error still occurs during IPI deployment. Thanks!

PR #881, opened 2021-05-17
strip whitespace after decoding secret values

dhellmann:
We have had several users report problems with credential management
that turned out to be caused by extra whitespace in the encoded values
placed into the secrets they created. Most often this is caused by
using command line tools to encode the values. For example,

$ echo -n root | base64
cm9vdA==
$ echo root | base64
cm9vdAo=

This change assumes that extraneous whitespace in the encoded values
is not part of the real username or password, and strips it before
trying to use the credentials. The logic to extract the credentials
from the secret is moved into its own function so that new unit tests
can be added.

### DIFF
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -1204,6 +1204,20 @@ func (r *BareMetalHostReconciler) getBMCSecretAndSetOwner(request ctrl.Request, 
 	return bmcCredsSecret, nil
 }
 
+func credentialsFromSecret(bmcCredsSecret *corev1.Secret) *bmc.Credentials {
+	// We trim surrounding whitespace because those characters are
+	// unlikely to be part of the username or password and it is
+	// common for users to encode the values with a command like
+	//
+	//     echo "my-password" | base64
+	//
+	// which introduces a trailing newline.
+	return &bmc.Credentials{
+		Username: strings.TrimSpace(string(bmcCredsSecret.Data["username"])),
+		Password: strings.TrimSpace(string(bmcCredsSecret.Data["password"])),
+	}
+}
+
 // Make sure the credentials for the management controller look
 // right and manufacture bmc.Credentials.  This does not actually try
 // to use the credentials.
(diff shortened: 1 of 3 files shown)


dtantsur:
/test-integration

LGTM, but let's mention this behavior at least somewhere in the docs?

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1204,6 +1204,20 @@ func (r *BareMetalHostReconciler) getBMCSecretAndSetOwner(request ctrl.Request,
...
+func credentialsFromSecret(bmcCredsSecret *corev1.Secret) *bmc.Credentials {
+	// We trim surrounding whitespace because those characters are
+	// unlikely to be part of the username or password and it is
+	// common for users to encode the values with a command like
+	//
+	//     echo "my-password" | base64
+	//
+	// which introduces a trailing newline.
+	return &bmc.Credentials{
+		Username: strings.TrimSpace(string(bmcCredsSecret.Data["username"])),
```
andfasano, line 1216: Note: it looks like that `TrimSpace` consider the Unicode set of characters as a "space": https://golang.org/pkg/unicode/#IsSpace. For the sake of clarity I'd suggest to document it.

controllers/metal3.io/baremetalhost_controller_test.go
```diff
@@ -2000,3 +2001,50 @@ func TestInvalidBMHCanBeDeleted(t *testing.T) {
...
+					"password": []byte("password"),
+				},
+			},
+			expected: bmc.Credentials{
+				Username: "username",
+				Password: "password",
+			},
+		},
+		{
+			name: "newline",
```
andfasano, line 2030: Not sure if you want to cover all the Unicode cases, but I think it could be useful to have at least the case like

```
[]byte("   username"),
[]byte("password   "),
```

andfasano (approved):

controllers/metal3.io/baremetalhost_controller.go
dhellmann: That set looks the same as "whitespace" to me. Is it different?

controllers/metal3.io/baremetalhost_controller_test.go
dhellmann: I added a case for non-newline whitespace, but I don't think it's necessary to retest strings.TrimSpace() in our test suite.

dhellmann:
> /test-integration
> 
> LGTM, but let's mention this behavior at least somewhere in the docs?

Good point. I added a note to the API docs.

controllers/metal3.io/baremetalhost_controller.go
andfasano: No it's the same, so probably not required

Issue #180, opened 2019-05-03
Introspection gives every NIC a network name of "pod networking"


russellb:
Right now we hardcode a network name of "Pod networking" for every NIC.  This should be either removed, or made dynamic somehow.

https://github.com/metal3-io/baremetal-operator/blob/master/pkg/provisioner/ironic/ironic.go#L378

Example:

```
apiVersion: metal3.io/v1alpha1
kind: BareMetalHost
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"metal3.io/v1alpha1","kind":"BareMetalHost","metadata":{"annotations":{},"name":"kube-worker-0","namespace":"metal3"},"spec":{"bmc":{"address":"ipmi://192.168.111.1:6233","credentialsName":"kube-worker-0-bmc-secret"},"bootMACAddress":
"00:93:1e:b1:74:87","online":true}}
  creationTimestamp: "2019-05-03T18:24:45Z"
  finalizers:
  - baremetalhost.metal3.io
  generation: 2
  name: kube-worker-0
  namespace: metal3
  resourceVersion: "1382"
  selfLink: /apis/metal3.io/v1alpha1/namespaces/metal3/baremetalhosts/kube-worker-0
  uid: ba1d5285-6dd0-11e9-86cf-4c9a6490472b
spec:
  bmc:
    address: ipmi://192.168.111.1:6233
    credentialsName: kube-worker-0-bmc-secret
  bootMACAddress: 00:93:1e:b1:74:87
  hardwareProfile: ""
  online: true
status:
  errorMessage: ""
  goodCredentials:
    credentials:
      name: kube-worker-0-bmc-secret
      namespace: metal3
    credentialsVersion: "807"
  hardware:
    cpu:
      count: 2
      model: Intel(R) Core(TM) i7-7567U CPU @ 3.50GHz
      speedGHz: 3.50401
      type: x86_64
    nics:
    - ip: 172.22.0.54
      mac: 00:93:1e:b1:74:87
      model: 0x1af4 0x0001
      name: eth0
      network: Pod Networking
      speedGbps: 0
    - ip: 192.168.111.23
      mac: 00:93:1e:b1:74:89
      model: 0x1af4 0x0001
      name: eth1
      network: Pod Networking
      speedGbps: 0
    ramGiB: 4
    storage:
    - model: QEMU QEMU HARDDISK
      name: /dev/sda
      sizeGiB: 50
[...]

PR #884, opened 2021-05-19
Custom deploy procedure

dtantsur:
See https://github.com/metal3-io/metal3-docs/blob/master/design/deploy-steps.md


apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -413,6 +417,14 @@ type Image struct {
 	DiskFormat *string `json:"format,omitempty"`
 }
 
+// Custom deploy is a description of a customized deploy process.
+type CustomDeploy struct {
+	// Custom deploy method name.
+	// This name is specific to the deploy ramdisk used. If you don't have
+	// a custom deploy ramdisk, you shouldn't use CustomDeploy.
+	Method string `json:"method"`
```
redixin, line 425: What about args?

dtantsur: I was requested to omit args on the review of the design document. They can be added later if we have a good case (and an idea how to wire them in the API).

redixin: What do you mean `a good case`? Passing args to the method isn't a good case?

Also `Args` is required field in DeployStep https://github.com/gophercloud/gophercloud/blob/v0.17.0/openstack/baremetal/v1/nodes/requests.go#L427

dtantsur: This implementation passes empty args (btw I'm not sure why we require them in ironic).

dtantsur: By a good case I mean a use case for passing args. What I'm doing does not require non-empty args.

redixin: E.g when someone want to use ansible to deploy.

Your comment:

> #180 has been made much more generic, and you may be able to reuse its interface for your goals. E.g. you allow `CustomDeploy.Method = "ansible"` to designate the ansible deploy. Then you can use metaData for passing additional information.

https://github.com/metal3-io/metal3-docs/pull/181#discussion_r629368593

But regardless this comment, passing args to methods is pretty standard approach and methods without args are pretty much useless.

I understand that doc is merged and only `Method` is mentioned there, but am I only one who confused here?

"Ironic api allows to pass args but let's not use it."

I also can't see any discussion about removing args here: https://github.com/metal3-io/metal3-docs/pull/180

What is wrong with args?

dtantsur: The previous discussion is https://github.com/metal3-io/metal3-docs/pull/180#discussion_r626924686. The problem was building API that accepts arbitrary JSON data. We already have `MetaData` and `UserData` for that, although it may not work for the ansible case. In any case, I suggest you rebase your older design on top of https://github.com/metal3-io/metal3-docs/pull/180 and discuss it there. This patch implements the approved design.

elfosardo (approved):

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -869,6 +884,11 @@ func (host *BareMetalHost) NeedsProvisioning() bool {
 		// The host is not supposed to be powered on.
 		return false
 	}
+
+	return host.hasNewImage() || host.hasNewCustomDeploy()
```
hardys, line 888: I'm not sure we really want the new spec field to trigger reprovisioning?

Currently the only field which does that is the image, so as an existing user I might be surprised if I changed some value in the custom deploy field and all my hosts got rebuilt?

I'm thinking about consistency e.g with userData/networkData etc where those fields (or rather the secrets they reference) could change post-deploy, but we don't rebuild even though they influence the deployed OS - this is similar even though it won't reference a potentially-shared secret?

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -845,11 +852,22 @@ func (r *BareMetalHostReconciler) actionProvisioning(prov provisioner.Provisione
 	}
 
 	// If the provisioner had no work, ensure the image settings match.
-	if info.host.Status.Provisioning.Image != *(info.host.Spec.Image) {
+	if info.host.Spec.Image == nil && info.host.Status.Provisioning.Image.URL != "" {
```
hardys, line 855: How can this happen - shouldn't the Status.Provisioning.Image always be removed by deprovisioning?

If this is a corner case found writing the test, I'd break it out into a separate commit as I'm currently not seeing how it's related to this change?

dtantsur: > shouldn't the Status.Provisioning.Image always be removed by deprovisioning?

This is provisioning code. Previously we would never get here if Spec.Image is nil, now we can. The old code causes segfault if we use Image == nil and CustomDeploy != nil.

apis/metal3.io/v1alpha1/baremetalhost_types.go
dtantsur: If we don't do this, how will we unprovision? In case when Image == nil and CustomDeploy != nil, the only way to unprovision is to set CustomDeploy = nil.

hardys: Perhaps I'm confused, can you outline the expected provision/deprovision flow here?

I expect a BMH to deprovision only when I remove (or replace) the image URL, so from a users perspective I'd expect the same behavior even if I have chosen to use a custom deploy ramdisk?

controllers/metal3.io/baremetalhost_controller.go
hardys: Again perhaps I'm missing something, can you explain in more detail why we can now hit the provisioning path with an image set in the status?  IIUC that should never happen, because we should have deprovisioned first.  /cc @zaneb

dtantsur: I'm not sure why you're asking me, that's the code that existed before my change. I just made sure it doesn't fail with segfault now that Spec.Image can be niI.

I assume it's this code that synchronizes Status with Spec after deployment, but I may be wrong.

apis/metal3.io/v1alpha1/baremetalhost_types.go
dtantsur: Yes, but you don't have an image now. What would you do to deprovision?

My assumption was that a user removes CustomDeploy from Spec, similarly to how Image removal causes deprovisioning now.

controllers/metal3.io/baremetalhost_controller.go
dtantsur: Looking more at the code. I can remove this check if you think it's safe to, but I'm not 100% sure myself. Previously this code path would never be hit with Spec.Image == nil.

hardys: I'm asking you because the `info.host.Status.Provisioning.Image = metal3v1alpha1.Image{}` is new - and AFAIK we should never clear the image in the status during provisioning, only during deprovisioning

So, I'd like to understand under what circumstances the state machine can allow us to get into this state, whether the issue is existing code or not :)

apis/metal3.io/v1alpha1/baremetalhost_types.go
hardys: In the current flow, if the `spec.image` gets removed, that triggers deprovisioning - in the case where CustomDeploy is specified I'd expect the same behavior.

But I probably would *not* expect to trigger rebuild of the host, if I modify CustomDeploy on an already provisioned host - most other spec fields don't trigger rebuild, so I think I'd expect this to behave the same.

If someone removes CustomDeploy from the spec, I'd expect to revert to the existing behavior (use the default deploy steps and IPA kernel/ramdisk) - obviously there are likely cases where it doesn't make sense to remove CustomDeploy because that configuration is required for provisioning and deprovisioning, but that's up to the user?

controllers/metal3.io/baremetalhost_controller.go
hardys: To clarify, I'm saying we should only clear the image in [actionDeprovisioning](https://github.com/metal3-io/baremetal-operator/blob/master/controllers/metal3.io/baremetalhost_controller.go#L917) - this addition to `actionProvisioning` seems wrong, so I want to understand if/why it's needed.

apis/metal3.io/v1alpha1/baremetalhost_types.go
dtantsur: Sorry, I don't get it. How do you suggest we trigger reprovision if Image is nil already?

dtantsur: > use the default deploy steps and IPA kernel/ramdisk

We have untied this logic from custom image logic, so only the default steps. But then we end up with a provisioned node with Image==nil and CustomDeploy==nil?

hardys: Ok we discussed this, and the reason for my confusion is that it's now possible to have `CustomDeploy!=nil`, but with `Image==nil` e.g the Image field is now optional (which is mentioned in the [design](https://github.com/metal3-io/metal3-docs/blob/master/design/deploy-steps.md))

That is unfortunate since it changes the existing API behavior, and potentially means more code corner cases since we have two APIs now that trigger provisioning/deprovisioning.

I know @zaneb had some suggestions on using the image URL instead, but since we accepted the design with this API I'm not blocking on it unless others think we need to revisit that discussion.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -598,6 +598,7 @@ func (r *BareMetalHostReconciler) registerHost(prov provisioner.Provisioner, inf
 			AutomatedCleaningMode: info.host.Spec.AutomatedCleaningMode,
 			State:                 info.host.Status.Provisioning.State,
 			CurrentImage:          getCurrentImage(info.host),
+			CustomDeploy:          info.host.Spec.CustomDeploy,
```
andfasano, line 601: I guess a `.DeepCopy()` may be required here?

controllers/metal3.io/baremetalhost_controller_test.go
```diff
@@ -1035,6 +1035,45 @@ func TestNeedsProvisioning(t *testing.T) {
 	}
 }
 
+// TestNeedsProvisioning verifies the logic for deciding when a host
+// needs to be provisioned when custom deploy is used.
+func TestNeedsProvisioningCustomDeploy(t *testing.T) {
```
andfasano, line 1040: I'd find easier to understand this test if defined in the table driven mode, given the presence of multiple cases

controllers/metal3.io/baremetalhost_controller_test.go
```diff
@@ -1059,6 +1098,52 @@ func TestProvision(t *testing.T) {
...
+	}
+	host.Spec.Online = true
+	r := newTestReconciler(host)
+
+	tryReconcile(t, r, host,
+		func(host *metal3v1alpha1.BareMetalHost, result reconcile.Result) bool {
+			t.Logf("custom deploy: %v", host.Spec.CustomDeploy)
+			t.Logf("provisioning custom deploy: %v", host.Status.Provisioning.CustomDeploy)
+			t.Logf("provisioning state: %v", host.Status.Provisioning.State)
+			return host.Status.Provisioning.CustomDeploy != nil && host.Status.Provisioning.CustomDeploy.Method != ""
```
andfasano, line 1116: Just to be sure: `host.Status.Provisioning.CustomDeploy.Method == "install_everything"`

Also, to be on the safe side, it could be worth adding a check on the state, ie `.. && host.Status.Provisioning.State == metal3v1alpha1.StateProvisioned`

controllers/metal3.io/host_state_machine.go
```diff
@@ -443,6 +447,28 @@ func (hsm *hostStateMachine) provisioningCancelled() bool {
...
+	if hsm.Host.Status.Provisioning.CustomDeploy == nil {
+		return false
+	}
+	if hsm.Host.Status.Provisioning.CustomDeploy.Method == "" {
+		return false
+	}
+	if hsm.Host.Spec.CustomDeploy.Method != hsm.Host.Status.Provisioning.CustomDeploy.Method {
+		return true
+	}
+	if hsm.Host.Spec.Image != nil && hsm.Host.Spec.Image.URL != hsm.Host.Status.Provisioning.Image.URL {
```
andfasano, line 466: Looks like this one should be already covered by `imageProvisioningCancelled()`?

pkg/provisioner/ironic/ironic.go
```diff
@@ -220,8 +220,8 @@ func newProvisionerWithSettings(host metal3v1alpha1.BareMetalHost, bmcCreds bmc.
 
 func newProvisionerWithIronicClients(hostData provisioner.HostData, publisher provisioner.EventPublisher, clientIronic *gophercloud.ServiceClient, clientInspector *gophercloud.ServiceClient) (*ironicProvisioner, error) {
 	// Ensure we have a microversion high enough to get the features
-	// we need.
-	clientIronic.Microversion = "1.56"
+	// we need. 1.69 adds DeploySteps to provisioning.
+	clientIronic.Microversion = "1.69"
```
andfasano, line 224: Question: could be this one considered a potential breaking change worth to be documented?

pkg/provisioner/ironic/ironic.go
```diff
@@ -1305,6 +1338,19 @@ func (p *ironicProvisioner) getConfigDrive(data provisioner.ProvisionData) (conf
 	return
 }
 
+func (p *ironicProvisioner) getCustomDeploySteps(customDeploy *metal3v1alpha1.CustomDeploy) (deploySteps []nodes.DeployStep) {
+	if customDeploy != nil && customDeploy.Method != "" {
+		deploySteps = append(deploySteps, nodes.DeployStep{
+			Interface: nodes.InterfaceDeploy,
+			Step:      customDeploy.Method,
+			Args:      map[string]interface{}{},
+			Priority:  80,
```
andfasano, line 1347: Maybe a named const here could be clearer

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -817,12 +818,18 @@ func (r *BareMetalHostReconciler) actionProvisioning(prov provisioner.Provisione
...
+	}
+
 	provResult, err := prov.Provision(provisioner.ProvisionData{
-		Image:           *info.host.Spec.Image.DeepCopy(),
+		Image:           image,
 		HostConfig:      hostConf,
 		BootMode:        info.host.Status.Provisioning.BootMode,
 		HardwareProfile: hwProf,
 		RootDeviceHints: info.host.Status.Provisioning.RootDeviceHints.DeepCopy(),
+		CustomDeploy:    info.host.Spec.CustomDeploy.DeepCopy(),
```
zaneb, line 832: nit: I'd suggest putting this right after the Image to keep related data together.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -845,11 +852,16 @@ func (r *BareMetalHostReconciler) actionProvisioning(prov provisioner.Provisione
 	}
 
 	// If the provisioner had no work, ensure the image settings match.
-	if info.host.Status.Provisioning.Image != *(info.host.Spec.Image) {
+	if info.host.Spec.Image != nil && info.host.Status.Provisioning.Image != *(info.host.Spec.Image) {
 		info.log.Info("updating deployed image in status")
 		info.host.Status.Provisioning.Image = *(info.host.Spec.Image)
 	}
 
+	if info.host.Spec.CustomDeploy != nil && (info.host.Status.Provisioning.CustomDeploy == nil || *(info.host.Spec.CustomDeploy) != *(info.host.Status.Provisioning.CustomDeploy)) {
```
zaneb, line 860: That last comparison is I think valid because the struct only contains a string, but if it contained anything more complex (and the whole point of defining a struct is to give us an option of adding more complex stuff in future AIUI) then it wouldn't. If we use reflect.DeepEqual() here now we can save someone a headache in future. (Yes, the image comparison above is also dodgy.)

controllers/metal3.io/baremetalhost_controller_test.go
dtantsur: I'm not sure what you mean by "table driven mode", could you elaborate?

pkg/provisioner/ironic/ironic.go
dtantsur: Do we have any sort of release notes for that?

pkg/provisioner/ironic/ironic.go
```diff
@@ -907,7 +907,37 @@ func (p *ironicProvisioner) setDirectDeployUpdateOptsForNode(ironicNode *nodes.N
...
+	var optValues optionsData
+	if imageData != nil && imageData.URL != "" {
+		checksum, checksumType, _ := imageData.GetChecksum()
+		// NOTE(dtantsur): all fields are optional for custom deploy
+		optValues = optionsData{
+			"boot_iso":            nil,
+			"image_checksum":      nil,
+			"image_source":        imageData.URL,
+			"image_os_hash_algo":  checksumType,
+			"image_os_hash_value": checksum,
```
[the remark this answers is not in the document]
dtantsur: I'd avoid empty strings, they're not equivalent to `nil`.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -817,12 +818,18 @@ func (r *BareMetalHostReconciler) actionProvisioning(prov provisioner.Provisione
...
 	}
 
+	var image metal3v1alpha1.Image
+	if info.host.Spec.Image != nil {
+		image = *info.host.Spec.Image.DeepCopy()
+	}
+
 	provResult, err := prov.Provision(provisioner.ProvisionData{
-		Image:           *info.host.Spec.Image.DeepCopy(),
+		Image:           image,
```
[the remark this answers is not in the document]
dtantsur: I tried that, and I think you were the one to suggest we don't :) In any case, it's going to be very invasive.

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -740,6 +752,9 @@ type ProvisionStatus struct {
 
 	// The Bios set by the user
 	Firmware *FirmwareConfig `json:"firmware,omitempty"`
+
+	// Custom deploy procedure applied to the host.
+	CustomDeploy *CustomDeploy `json:"customDeploy,omitempty"`
```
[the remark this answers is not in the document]
dtantsur: Not too much? We'd still have to deal with Image.URL possibly be "", CustomDeploy possibly be ""/`nil` and so on. A potential user confusion because of conflating too barely related terms may not be worth it.

controllers/metal3.io/host_state_machine.go
dtantsur: It's not. The two functions are called with `&&` between them, so if `imageProvisioningCancelled` returns `true`, this will return `false` and the result will be `false`.

controllers/metal3.io/baremetalhost_controller_test.go
andfasano: Yes, sure, here's an example: https://github.com/metal3-io/baremetal-operator/blob/master/pkg/provisioner/ironic/updatehardwarestate_test.go.

Essentially in the test we're going to define a number of different relevant cases (each one with its own inputs and expectation), and such approach helps in reducing the duplications, reduce the risks of side effects, and definitely supports troubleshooting and maintenance.

(there's a great blog about it here https://dave.cheney.net/2019/05/07/prefer-table-driven-tests)

pkg/provisioner/ironic/ironic.go
andfasano: AFAIK I don't think we have rl, and probably this could be a good point. Currently, there's just the docs in the `\docs` folder, a possibility could be to add a `Requirements` section there

controllers/metal3.io/host_state_machine.go
andfasano: Yes, that's right, but it looks a little bit tricky, and since this method could trigger a disruptive  action like deprovisioning I think it's worth trying to simplify it as much as possible.
I guess that the goal of the `provisioningCancelled()` method is essentially to detect a change - during the `Provisioning` or `Provisioned` state - such that invalidates the current state, thus requiring a deprovisioning.

As per the proposal, the  `Image` field should be considered optional if `CustomDeploy` is set (but not if it isn't of course), and this may suggest to:
1)  Check first `CustomDeploy` and then `Image` 
    * Do not consider `Image` nullity as a trigger condition in such case if `CustomDeploy` is set
2) If `CustomDeploy` is nil, check `Image` field as before to ensure the previous behavior

Ie, something like that: 

```
func (hsm *hostStateMachine) provisioningCancelled() bool {

	if hsm.Host.Spec.CustomDeploy != nil {
		if hsm.Host.Spec.CustomDeploy.Method == "" {
			return true
		}
		if hsm.Host.Status.Provisioning.CustomDeploy != nil && hsm.Host.Spec.CustomDeploy.Method != hsm.Host.Status.Provisioning.CustomDeploy.Method {
			return true
		}
		// At this point spec.CustomDeploy value didn't change (and it's not empty). Only a discrepancy in the Image
		// could require a deprovisioning, but if it isn't set or empty then is fine
		if hsm.Host.Spec.Image != nil && hsm.Host.Status.Provisioning.Image.URL != "" && hsm.Host.Spec.Image.URL != hsm.Host.Status.Provisioning.Image.URL {
			return true
		}

		return false
	}

	// Fallback to the previous case
	return hsm.imageProvisioningCancelled()
}
```

zaneb: I also read this wrong, and I have created truly terrible bugs in the past (see #915) by getting this function wrong, so I agree that we should seek to simplify the logic.

pkg/provisioner/ironic/ironic.go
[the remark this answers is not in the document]
zaneb: To be clearer, there are theoretically 3 ways to clear a field in Ironic, only 2 of which are possible using SetInstanceInfoOpts:

* Update the field and set the value to `""` - passing `""` does this as you'd expect
* Update the field and set the value to `null` - there's no way to do this
* Delete the field - passing `nil` will do this

Sounds like we do want the latter though, when the return from GetChecksum() is false.

controllers/metal3.io/baremetalhost_controller.go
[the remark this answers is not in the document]
zaneb: Oh, yeah I guess it's probably even more messy on the provisioner side to flip it.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -551,6 +551,10 @@ func getCurrentImage(host *metal3v1alpha1.BareMetalHost) *metal3v1alpha1.Image {
 	return nil
 }
 
+func hasCustomDeploy(host *metal3v1alpha1.BareMetalHost) bool {
+	return host.Status.Provisioning.CustomDeploy != nil || host.Spec.CustomDeploy != nil
```
zaneb, line 555: I think the correct logic here is:

```
if host.Status.Provisioning.CustomDeploy != nil && host.Status.Provisioning.CustomDeploy.Method != "" {
    return true
}

switch host.Status.Provisioning.State {
case metal3v1alpha1.StateProvisioning, metal3v1alpha1.StateExternallyProvisioned:
    return host.Spec.CustomDeploy != nil && host.Spec.CustomDeploy.Method != ""
default:
    return false
}
```

dtantsur: Thank you!

controllers/metal3.io/host_state_machine.go
dtantsur: Will do, but the proposed snippet doesn't cover one case (removing custom deploy while keeping the image). I'll try to figure it out.

pkg/provisioner/ironic/ironic.go
dtantsur: I'll put something in configuration.md

pkg/provisioner/ironic/ironic.go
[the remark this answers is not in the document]
dtantsur: Aha, got it.

dtantsur:
/test-integration 

All comments addressed, please check again.

andfasano:
Thanks @dtantsur, current improvements seem fine to me.

/lgtm

s3rj1k:
@dtantsur What is the minimal Ironic release after this PR gets merged, also can this be stated somewhere in docs?
Thanks for your hard work.

dtantsur:
> What is the minimal Ironic release after this PR gets merged, also can this be stated somewhere in docs?

We're stating the minimum API version in configuration.md, which can be matched to the actual service version in https://docs.openstack.org/ironic/latest/contributor/webapi-version-history.html. This requires 1.62 (Wallaby).

controllers/metal3.io/host_state_machine.go
```diff
@@ -428,7 +428,26 @@ func (hsm *hostStateMachine) handleReady(info *reconcileInfo) actionResult {
 }
 
 func (hsm *hostStateMachine) provisioningCancelled() bool {
-	return hsm.imageProvisioningCancelled() && hsm.customDeployProvisioningCancelled()
+	if hsm.Host.Spec.CustomDeploy != nil {
+		if hsm.Host.Spec.CustomDeploy.Method == "" {
```
zaneb, line 432: Not sure why an empty custom deploy struct causes provisioning to be unconditionally cancelled? I think that this should go away and the line above should be changed to:

```
if hsm.Host.Spec.CustomDeploy != nil && hsm.Host.Spec.CustomDeploy.Method != "" {
```

In my view the only relevant thing is the Method, and the checks for nil are only to prevent crashes not something that should provide a subtle user interface to behaviour changes.

controllers/metal3.io/host_state_machine.go
```diff
@@ -428,7 +428,26 @@ func (hsm *hostStateMachine) handleReady(info *reconcileInfo) actionResult {
...
+			return true
+		}
+		if hsm.Host.Status.Provisioning.CustomDeploy != nil && hsm.Host.Spec.CustomDeploy.Method != hsm.Host.Status.Provisioning.CustomDeploy.Method {
+			return true
+		}
+		// At this point spec.CustomDeploy value didn't change (and it's not empty). Only a discrepancy in the Image
+		// could require a deprovisioning, but if it isn't set or empty then is fine
+		if hsm.Host.Spec.Image != nil && hsm.Host.Status.Provisioning.Image.URL != "" && hsm.Host.Spec.Image.URL != hsm.Host.Status.Provisioning.Image.URL {
```
zaneb, line 440: I think this should be:

```
if hsm.Host.Status.Provisioning.Image.URL != "" &&
    (hsm.Host.Spec.Image == nil ||
        hsm.Host.Spec.Image.URL != hsm.Host.Status.Provisioning.Image.URL) {
```

controllers/metal3.io/host_state_machine.go
```diff
@@ -428,7 +428,26 @@ func (hsm *hostStateMachine) handleReady(info *reconcileInfo) actionResult {
...
+			return true
+		}
+
+		return false
+	} else if hsm.Host.Status.Provisioning.CustomDeploy != nil && hsm.Host.Status.Provisioning.CustomDeploy.Method != "" {
+		return true
+	}
+
+	// Fallback to the previous case
```
zaneb, line 449: nit: 'previous' here refers to history not to control flow, so it will make less sense outside the context of the code review.

controllers/metal3.io/host_state_machine.go
```diff
@@ -428,6 +428,27 @@ func (hsm *hostStateMachine) handleReady(info *reconcileInfo) actionResult {
 }
 
 func (hsm *hostStateMachine) provisioningCancelled() bool {
+	if hsm.Host.Spec.CustomDeploy != nil && hsm.Host.Spec.CustomDeploy.Method != "" {
+		if hsm.Host.Status.Provisioning.CustomDeploy != nil && hsm.Host.Spec.CustomDeploy.Method != hsm.Host.Status.Provisioning.CustomDeploy.Method {
```
zaneb, line 432: The `Status.Provisioning.CustomDeploy.Method` is expected to be empty during provisioning. I think _in practice_ that `Status.Provisioning.CustomDeploy` will be nil also, but it seems like it would be more robust to check both:

```
if hsm.Host.Status.Provisioning.CustomDeploy != nil &&
    hsm.Host.Status.Provisioning.CustomDeploy.Method != "" &&
        hsm.Host.Spec.CustomDeploy.Method != hsm.Host.Status.Provisioning.CustomDeploy.Method {
```

dtantsur: May I follow-up with this? I'd prefer to avoid another round of looking for lgtm and fighting the CI because of something so minor.

dtantsur: https://github.com/metal3-io/baremetal-operator/pull/932

PR #887, opened 2021-05-20
Refactoring: make ProvisionData.Image a pointer

dtantsur:
This is preparation for having Image optional on deployment, as is the
case with custom deploy procedure. See this design for details:
https://github.com/metal3-io/metal3-docs/pull/180/

### DIFF
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -919,7 +919,7 @@ func (p *ironicProvisioner) getImageUpdateOptsForNode(ironicNode *nodes.Node, im
 func (p *ironicProvisioner) getUpdateOptsForNode(ironicNode *nodes.Node, data provisioner.ProvisionData) *nodeUpdater {
 	updater := updateOptsBuilder(p.debugLog)
 
-	p.getImageUpdateOptsForNode(ironicNode, &data.Image, data.BootMode, updater)
+	p.getImageUpdateOptsForNode(ironicNode, data.Image, data.BootMode, updater)
 
 	opts := optionsData{
 		"root_device": devicehints.MakeHintMap(data.RootDeviceHints),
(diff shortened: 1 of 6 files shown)


elfosardo (approved):

pkg/provisioner/ironic/ironic.go
```diff
@@ -1065,18 +1065,22 @@ func (p *ironicProvisioner) Adopt(data provisioner.AdoptData, force bool) (resul
 	return operationComplete()
 }
 
-func (p *ironicProvisioner) ironicHasSameImage(ironicNode *nodes.Node, image metal3v1alpha1.Image) (sameImage bool) {
+func (p *ironicProvisioner) ironicHasSameImage(ironicNode *nodes.Node, image *metal3v1alpha1.Image) (sameImage bool) {
 	// To make it easier to test if ironic is configured with
 	// the same image we are trying to provision to the host.
-	if image.DiskFormat != nil && *image.DiskFormat == "live-iso" {
+	if image != nil && image.DiskFormat != nil && *image.DiskFormat == "live-iso" {
```
zaneb, line 1071: This is correct insofar as it will prevent anything from crashing, but is this the logic we actually want?
If the image is nil, shouldn't we check that none of the image fields are set in ironic?

zaneb (commented):
There are a still a few places in ironic.go where we are using a *Image without checking for nil (specifically in functions that previously took a pointer but where the pointer was never nil):

* `setLiveIsoUpdateOptsForNode()` l. 846
* `setDirectDeployUpdateOptsForNode()`, ll. 881, 885
* `setUpForProvisioning()` l. 1005
* `Provision()` l. 1336

Is there a reason we wouldn't just continue to pass the Image info from the Spec (even if it is empty) to the Provisioner?

dtantsur:
> Is there a reason we wouldn't just continue to pass the Image info from the Spec (even if it is empty) to the Provisioner?

It's a pointer in the Spec, so in theory it can be nil (not right now, but with my next patch).

dtantsur:
Hmm, I see your point, I'll try it.

PR #888, opened 2021-05-20
Correctly run controller-gen for new apis module

### LINKED ISSUE #875: Gomod for public API broke controller-tools object generation

zaneb:
Fixes #875

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,6 @@ COVER_PROFILE = cover.out
 BIN_DIR := bin
 
 CRD_OPTIONS ?= "crd:trivialVersions=false,allowDangerousTypes=true,crdVersions=v1"
-CONTROLLER_GEN ?= go run sigs.k8s.io/controller-tools/cmd/controller-gen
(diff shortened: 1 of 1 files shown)


kashifest:
This looks good to me. I can close #877  in favor of this.

PR #889, opened 2021-05-21
Fix bug with re-inspection due to Node in "available" state

zaneb:
When an annotation appears to force re-inspection, the Host returns to
the Inspecting state from the Ready/Available state. Originally the
ironic node would always be in the "manageable" state for a ready Host,
but we have usually been leaving it in the "available" state after
deprovisioning since 18465130c8f42bda4aa8191b5093656dcc029d90, and after
the initial preparation since 9a55ce5bc63a2f24916f1f604c736570836fdaea.

Handle the "available" node state during inspection by moving the node
to the "manageable" state. Since the first call to InspectHardware may
not start a new inspection, we have to return a new value to indicate
when inspection has actually started and thus it is safe to clear the
annotation. This also fixes a bug where failing to set the boot mode due
to a conflict would cause the inspection annotation to be removed
without starting a new inspection.

### DIFF
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -667,7 +667,7 @@ func (r *BareMetalHostReconciler) actionInspecting(prov provisioner.Provisioner,
 	info.log.Info("inspecting hardware")
 
 	refresh := hasInspectAnnotation(info.host)
-	provResult, details, err := prov.InspectHardware(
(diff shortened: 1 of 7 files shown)


fmuyassarov (commented):
This looks good to me. Thanks for taking care of this quickly. I will wait until integration CI passes and meanwhile do a local test with inspection api.

fmuyassarov:
I've tested it locally and this indeed fixes the bug.
/lgtm

PR #892, opened 2021-05-24
Also run controller-gen on the controller directory to get rbac

asalkeld:
The RBAC manifests are built from the controller (not api) code, so we have to run the `make manifests` commands at least on both the API module and the main module.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -133,11 +133,13 @@ tools/bin/controller-gen: go.mod
 .PHONY: manifests
 manifests: tools/bin/controller-gen ## Generate manifests e.g. CRD, RBAC etc.
 	cd apis; ../$< $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=../config/crd/bases
+	$< rbac:roleName=manager-role paths="./..." output:rbac:artifacts:config=config/rbac
 	$(KUSTOMIZE) build config/default > config/render/capm3.yaml
 
 .PHONY: generate
 generate: tools/bin/controller-gen ## Generate code
 	cd apis; ../$< object:headerFile="../hack/boilerplate.go.txt" paths="./..."
+	$< object:headerFile="hack/boilerplate.go.txt" paths="./..."
 
 ## --------------------------------------
 ## Docker Targets


Makefile
```diff
@@ -131,13 +131,16 @@ tools/bin/controller-gen: go.mod
 	go build -o $@ sigs.k8s.io/controller-tools/cmd/controller-gen
 
 .PHONY: manifests
+
```
zaneb, line 134: This seems gratuitous ;)

Makefile
```diff
@@ -131,13 +131,16 @@ tools/bin/controller-gen: go.mod
...
 manifests: tools/bin/controller-gen ## Generate manifests e.g. CRD, RBAC etc.
 	cd apis; ../$< $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=../config/crd/bases
+	$< rbac:roleName=manager-role paths="./..." output:rbac:artifacts:config=config/rbac
 	$(KUSTOMIZE) build config/default > config/render/capm3.yaml
 
 .PHONY: generate
 generate: tools/bin/controller-gen ## Generate code
 	cd apis; ../$< object:headerFile="../hack/boilerplate.go.txt" paths="./..."
+	$< object:headerFile="hack/boilerplate.go.txt" paths="./..."
```
zaneb, line 143: Out of curiousity, do we need both generate and manifests, or might just manifests suffice (at least for now)?

zaneb (commented):
/lgtm

Makefile
asalkeld: it's an oops'y ;-)

Makefile
asalkeld: this is to just make sure that if we add something in another dir, it's not going confuse the hell out of someone for hours.

zaneb:
@kashifest this is a follow-up to fix a mistake I made in #888, could you take a look?

kashifest:
/lgtm
@asalkeld would you please add some text in the PR description
I will run a centos run just to be sure everything is in order.
/test-centos-integration

zaneb:
Thanks @kashifest. I updated the description so we should be good to go.
/approve

PR #900, opened 2021-05-31
Fix configdrive handling

dtantsur:
I've spotted a few errors with the configdrive handling:

* We forgot to add configdrive to 'deploy failed' handler, so a repeated
  deployment will always happen without a configdrive.
* ConfigDrive creation is conditional on UserData presence. This is
  incorrect, any field being populated should trigger configdrive.
* There was no check on empty networkData, so we always tried to
  deserialize empty string with YAML (which works, but is incorrect).

Create a new helper for configdrive and cover it with unit tests.

### DIFF
--- a/pkg/provisioner/ironic/configdrive_test.go
+++ b/pkg/provisioner/ironic/configdrive_test.go
@@ -1,0 +1,134 @@
+package ironic
+
+import (
+	"testing"
+
+	"github.com/gophercloud/gophercloud/openstack/baremetal/v1/nodes"
+	"github.com/gophercloud/gophercloud/openstack/baremetalintrospection/v1/introspection"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
+	"github.com/metal3-io/baremetal-operator/pkg/bmc"
+	"github.com/metal3-io/baremetal-operator/pkg/provisioner"
+	"github.com/metal3-io/baremetal-operator/pkg/provisioner/fixture"
+	"github.com/metal3-io/baremetal-operator/pkg/provisioner/ironic/clients"
+	"github.com/metal3-io/baremetal-operator/pkg/provisioner/ironic/testserver"
+)
+
+func TestEmpty(t *testing.T) {
+	nodeUUID := "33ce8659-7400-4c68-9535-d10766f07a58"
+
+	cases := []struct {
+		name     string
+		hostData provisioner.HostConfigData
+		expected nodes.ConfigDrive
+	}{
+		{
+			name:     "empty",
+			hostData: fixture.NewHostConfigData("", "", ""),
+			expected: nodes.ConfigDrive{},
+		},
+		{
+			name:     "everything",
+			hostData: fixture.NewHostConfigData("testUserData", "test: NetworkData", "test: Meta"),
+			expected: nodes.ConfigDrive{
+				MetaData: map[string]interface{}{
+					"local-hostname":   "myhost",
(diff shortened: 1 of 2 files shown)


pkg/provisioner/ironic/ironic.go
```diff
@@ -1221,6 +1221,59 @@ func (p *ironicProvisioner) Prepare(data provisioner.PrepareData, unprepared boo
 	return
 }
 
+func (p *ironicProvisioner) getConfigDrive(data provisioner.ProvisionData) (result nodes.ConfigDrive, err error) {
```
zaneb, line 1224: Elsewhere in this file a return parameter named `result` is usually a `provisioner.Result`... it might be less confusing to name this `configDrive`.

zaneb (commented):
This would be slightly easier to review if the various fixes were in separate patches (within this PR).

/lgtm

pkg/provisioner/ironic/ironic.go
dtantsur: I can follow-up if this PR is merged the way it is.

elfosardo (approved):

pkg/provisioner/ironic/ironic.go
```diff
@@ -1221,6 +1221,59 @@ func (p *ironicProvisioner) Prepare(data provisioner.PrepareData, unprepared boo
...
+			return result, errors.Wrap(err, "failed to unmarshal metadata from secret")
+		}
+	}
+
+	// Set metaData if any field is populated by a user.
+	if metaDataRaw != "" || networkDataRaw != "" || userData != "" {
+		result.MetaData = metaData
+		p.log.Info("triggering provisioning with config drive")
+	} else {
+		p.log.Info("triggering provisioning without config drive")
```
hardys, line 1271: I'm not sure this is correct, as it changes the existing behavior - above we default `metaData` then also consume any user-provided metaDataRaw, e.g we should always create a configDrive containing at least the default metaData?

pkg/provisioner/ironic/configdrive_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+	cases := []struct {
+		name     string
+		hostData provisioner.HostConfigData
+		expected nodes.ConfigDrive
+	}{
+		{
+			name:     "empty",
+			hostData: fixture.NewHostConfigData("", "", ""),
+			expected: nodes.ConfigDrive{},
```
hardys, line 29: As mentioned below IIUC this should always contain the default metaData, so this isn't expected to be empty, and in the "only meta data" case we should rename the case to "extra meta data" or something, to prove user-provided meta-data is merged in with the defaults.

hardys:
Overall lgtm but I think the revised logic still isn't quite correct, because we'll only pass the default meta-data when one of (meta|network|user)Data is specified?

Previously we only passed it when userData was specified, which I agree was incorrect, but I think the desired behavior is to always pass this metaData regardless of the user-provided data?

pkg/provisioner/ironic/configdrive_test.go
dtantsur: I'd be worried about changing this behavior tbh. Don't we want to leave an option of not having a configdrive at all?

dtantsur:
> I think the desired behavior is to always pass this metaData regardless of the user-provided data?

I don't really know. It's a serious user-visible change. Could I follow-up with it, so that it can be reviewed separately (and reverted if things go south)?

dtantsur:
I'll bring it to the meeting today. If there are no objections, I'll update the PR.

PR #901, opened 2021-05-31
Add new CRDs for BIOS configuration

bfournie:
Add a new CRD for the HostFirmwareSettings for BIOS configuration as described in the spec - https://github.com/metal3-io/metal3-docs/pull/173. A new CRD is necessary since its not practical to include the number of settings in BareMetalHost.
In addition a CRD is added for the FirmwareSchema used to provide a schema for setting the firmware values.


config/samples/metal3.io_v1alpha1_hostfirmwaresettings.yaml
```diff
@@ -0,0 +1,46 @@
+apiVersion: metal3.io/v1alpha1
+kind: HostFirmwareSettings
+metadata:
+  name: hostfirmwaresettings-sample
+spec:
+  # Add fields here
+  foo: bar
```
zaneb, line 7: Looks like this needs to be regenerated :)

apis/metal3.io/v1alpha1/hostfirmwaresettings_types.go
```diff
@@ -0,0 +1,169 @@
...
+	// Whether or not this setting's value is unique to this node, e.g.
+	// a serial number.
+	Unique *bool `json:"unique,omitempty"`
+}
+
+// HostFirmwareSettingsSpec defines the desired state of HostFirmwareSettings
+type HostFirmwareSettingsSpec struct {
+	// Important: Run "make" to regenerate code after modifying this file
+
+	WriteableFirmwareSettings []WriteableFirmwareSetting `json:"writeableFirmwareSettings,omitempty"`
```
zaneb, line 78: Is there a reason we are not just making this a `map[string]string`? I tend to prefer that as it eliminates the possibility of duplicate values and therefore the question of what to do about them. (On the other hand I have used it on occasion in APIs and it turned out later to be the Wrong Thing.)

If the reason is related to problems generating an open-ended OpenAPI schema then I believe there are workarounds available.

apis/metal3.io/v1alpha1/hostfirmwaresettings_types.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+// HostFirmwareSettingsStatus defines the observed state of HostFirmwareSettings
+type HostFirmwareSettingsStatus struct {
+	// Important: Run "make" to regenerate code after modifying this file
+
+	CurrentFirmwareSettings []FirmwareSetting `json:"currentFirmwareSettings,omitempty"`
```
zaneb, line 85: I started a [discussion about naming](https://github.com/metal3-io/metal3-docs/pull/173#discussion_r644959330) (sorry) in the metal3-docs review.

apis/metal3.io/v1alpha1/hostfirmwaresettings_types.go
```diff
@@ -0,0 +1,169 @@
...
+
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// WriteableFirmwareSetting describes a FirmwareSetting whose value can be changed via Ironic clean-steps.
+type WriteableFirmwareSetting struct {
+	// Identifier for the Firmware setting.
+	Name string `json:"name" required:"true"`
+
+	// Value of the Firmware setting.
+	Value string `json:"value" required:"true"`
```
zaneb, line 33: Is the value always a string? I think this means that you'll be unable to apply yaml like:

```
spec:
   writableFirmwareSettings:
     someSetting: 42
```

and instead you'll have to enclose all numbers/bools in quotes. (I could be wrong, you should check.) That would be undesirable I think.

zaneb (commented):
Note that you might need the patch from #892 to properly generate all of the files.

apis/metal3.io/v1alpha1/hostfirmwaresettings_types.go
zaneb: We should probably add the directives:

```
// + patchMergeKey=name
// +patchStrategy=merge
```

so that people can easily just patch in only the values they care about.

(Merge key is only needed if we keep the list-of-structs type.)

config/samples/metal3.io_v1alpha1_hostfirmwaresettings.yaml
bfournie: Yeah I made need to patch this after regeneration, will fix.

apis/metal3.io/v1alpha1/hostfirmwaresettings_types.go
bfournie: I like using the map as it will simplify things when the spec is updated and we need to validate the value. But yeah I ran into some issues when trying to generate the CRD. I will look into using this more.

apis/metal3.io/v1alpha1/hostfirmwaresettings_types.go
bfournie: Sure, CRD will be updated to map the metal3-doc.

apis/metal3.io/v1alpha1/hostfirmwaresettings_types.go
bfournie: The value is always a string, that's how Ironic it comes from Ironic - https://docs.openstack.org/api-ref/baremetal/?expanded=list-all-bios-settings-by-node-detail#node-bios-nodes, and how its gets set via clean-steps. All integers will need to be quoted.

elfosardo (approved):

apis/metal3.io/v1alpha1/hostfirmwaresettings_types.go
bfournie: Updated to use maps for status and spec and for the firmwareschema. The generated yaml seems fine using "additionalProperties".

apis/metal3.io/v1alpha1/hostfirmwaresettings_types.go
```diff
@@ -0,0 +1,81 @@
...
+	// FirmwareSchema is a reference to the Schema used to describe each
+	// FirmwareSetting. It can be set here to use a Schema in a different
+	// Namespace other than the one that the settings is in.
+	FirmwareSchema *SchemaReference `json:"schema,omitempty"`
+
+	// Settings are the desired firmware settings stored as name/value pairs.
+	// This will be populated with the actual firmware settings and only
+	// contain the settings that can be modified (i.e. not ReadOnly), to
+	// facilitate making changes.
+	Settings map[string]string `json:"settings" required:"true"`
```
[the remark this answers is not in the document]
zaneb: If I'm reading https://kubernetes.io/blog/2019/06/20/crd-structural-schema/#extensions correctly, we _don't_ need to set `x-kubernetes-preserve-unknown-fields: true` here because all of the fields are strings. If we had arbitrary JSON data we would need to set that.

It seems like we should consider setting `x-kubernetes-int-or-string: true` so that the user can pass an int in yaml and have it coverted automatically rather than just rejected. It appears the way to trigger that is to use the type k8s.io/apimachinery/pkg/util/intstr.IntOrString in place of string.

apis/metal3.io/v1alpha1/firmwareschema_types.go
```diff
@@ -0,0 +1,163 @@
...
+	ResetRequired *bool `json:"reset_required,omitempty"`
+
+	// Whether or not this setting's value is unique to this node, e.g.
+	// a serial number.
+	Unique *bool `json:"unique,omitempty"`
+}
+
+// FirmwareSchemaSpec defines the desired state of FirmwareSchema
+type FirmwareSchemaSpec struct {
+	// The schema will be saved in the status section, spec is empty
```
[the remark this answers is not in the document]
bfournie: Sure will change to use Spec.

apis/metal3.io/v1alpha1/firmwareschema_types.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+// FirmwareSchemaStatus defines the observed state of FirmwareSchema
+type FirmwareSchemaStatus struct {
+
+	// The count of the number of times this schema is used by
+	// a HostFirmwareSettings resource
+	ReferenceCount int `json:"referenceCount" required:"true"`
```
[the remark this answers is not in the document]
bfournie: Sure can move this to Spec.

apis/metal3.io/v1alpha1/hostfirmwaresettings_types.go
```diff
@@ -0,0 +1,81 @@
...
+	// `name` is the reference to the schema.
+	Name string `json:"name"`
+}
+
+// HostFirmwareSettingsSpec defines the desired state of HostFirmwareSettings
+type HostFirmwareSettingsSpec struct {
+	// FirmwareSchema is a reference to the Schema used to describe each
+	// FirmwareSetting. It can be set here to use a Schema in a different
+	// Namespace other than the one that the settings is in.
+	FirmwareSchema *SchemaReference `json:"schema,omitempty"`
```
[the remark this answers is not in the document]
bfournie: This came about from the discussion on allowing the HostFirmwareSettings to use schemas in another namespace than the one the host is in in order to conserve resources. I interpreted this as requiring the user to define the other namespace to use (and perhaps the name) as BMO by default would only use the host namespace. Alternatively, if possible the BMO could search all namespaces for a matching schema, if we do that then this SchemaReference wouldn't be needed here.

apis/metal3.io/v1alpha1/hostfirmwaresettings_types.go
```diff
@@ -0,0 +1,81 @@
...
+
+// HostFirmwareSettingsStatus defines the observed state of HostFirmwareSettings
+type HostFirmwareSettingsStatus struct {
+	// FirmwareSchema is a reference to the Schema used to describe each
+	// FirmwareSetting. By default, this will be a Schema in the same
+	// Namespace as the settings but it can be overwritten in the Spec
+	FirmwareSchema *SchemaReference `json:"schema,omitempty"`
+
+	// Settings are the actual firmware settings stored as name/value pairs
+	// +patchStrategy=merge
```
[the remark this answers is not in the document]
bfournie: As per the doc, the Settings in the Status would contain all the settings read from Ironic and, upon reading, the BMO would copy these settings to the Spec but only include those that do not have the "ReadOnly" or "Unique" set in the FirmwareSchema for this setting. The Values in the Spec could then be easily updated by the user without having to change/write the Names. The BMO can compare the spec values to the expected values in the Status and only send the differences to clean steps. So with that design we'd need Settings in both places.

apis/metal3.io/v1alpha1/hostfirmwaresettings_types.go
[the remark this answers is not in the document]
bfournie: > It wouldn't hurt to define a type for this like
> 
> ```
> type SettingsMap map[string]string
> ``

Good idea

apis/metal3.io/v1alpha1/hostfirmwaresettings_types.go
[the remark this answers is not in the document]
zaneb: I meant the `+patchStrategy=merge` :)

apis/metal3.io/v1alpha1/hostfirmwaresettings_types.go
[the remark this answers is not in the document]
zaneb: Both the FirmwareSettings and the FirmwareSchema are going to be created by the BMO at ~the same time, so I don't think there's an opportunity for the user to intervene. I am picturing that the BMO will have a global flag to put the Schema in a single namespace vs. the namespace of the Host. Whoever runs the BMO will set that flag.

apis/metal3.io/v1alpha1/hostfirmwaresettings_types.go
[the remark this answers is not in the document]
bfournie: > It seems like we should consider setting `x-kubernetes-int-or-string: true` so that the user can pass an int in yaml and have it coverted automatically rather than just rejected. It appears the way to trigger that is to use the type k8s.io/apimachinery/pkg/util/intstr.IntOrString in place of string.

Yeah using IntOrString seems to work well and would allow the user to set values for example:
"RetryCount": "10"
or
"RetryCount": 10

Thanks for the suggestion.

apis/metal3.io/v1alpha1/hostfirmwaresettings_types.go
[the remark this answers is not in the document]
bfournie: Got it. Removed the SchemaReference from Spec. Will follow up with addition of a global flag.

apis/metal3.io/v1alpha1/firmwareschema_types.go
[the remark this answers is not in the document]
zaneb: I've been reading up on [Garbage Collection](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/) in k8s, and I think we should be able to use the owner references to manage this automatically (saves a lot of work implementing this in the controller). You _can_ have multiple owner references, which I didn't know before. (Only one of these can be marked as a 'controller', which is what we do for Secrets so that's what confused me. However, we don't [need](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/api-machinery/controller-ref.md) to do that here.)

So it's safe to ditch this field.

apis/metal3.io/v1alpha1/hostfirmwaresettings_types.go
```diff
@@ -0,0 +1,80 @@
...
+
+// HostFirmwareSettingsSpec defines the desired state of HostFirmwareSettings
+type HostFirmwareSettingsSpec struct {
+
+	// Settings are the desired firmware settings stored as name/value pairs.
+	// This will be populated with the actual firmware settings and only
+	// contain the settings that can be modified (i.e. not ReadOnly), to
+	// facilitate making changes.
+	// +patchStrategy=merge
+	Settings map[string]intstr.IntOrString `json:"settings" required:"true"`
```
zaneb, line 43: nit: we could think about declaring a type for this and naming it and SettingsMap in some way that makes it possible to distinguish them. Or maybe just ditch the SettingsMap type again since it's now used in only one place.

zaneb (commented):
This looks basically good. Couple of very minor things inline.

apis/metal3.io/v1alpha1/firmwareschema_types.go
```diff
@@ -0,0 +1,148 @@
...
+	// The hardware model associated with this schema
+	// +optional
+	HardwareModel string `json:"hardwareModel,omitempty"`
+
+	// Map of firmware name to schema
+	Schema map[string]SettingSchema `json:"schema" required:"true"`
+}
+
+//+kubebuilder:object:root=true
+//+kubebuilder:subresource:status
```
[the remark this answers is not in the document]
bfournie: Done

apis/metal3.io/v1alpha1/hostfirmwaresettings_types.go
bfournie: I think that type will be useful and had added it in other patch. Updated.

zaneb (commented):
/lgtm

apis/metal3.io/v1alpha1/firmwareschema_types.go
```diff
@@ -0,0 +1,147 @@
...
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+// Additional data describing the firmware setting
+type SettingSchema struct {
+
+	// The type of setting.
+	// +kubebuilder:validation:Enum=Enumeration;String;Integer;Boolean;Password
```
dtantsur, line 28: Do we expect to support non-Redfish implementation? If yes, we cannot add this validation.

apis/metal3.io/v1alpha1/firmwareschema_types.go
```diff
@@ -0,0 +1,147 @@
...
+// Check whether the setting's name and value is valid using the schema
+func (host *FirmwareSchema) CheckSettingIsValid(name string, value intstr.IntOrString, schemas map[string]SettingSchema) bool {
+
+	schema, ok := schemas[name]
+	if !ok {
+		// The setting must exist in the status
+		return false
+	}
+
+	if schema.ReadOnly != nil && *schema.ReadOnly == true {
```
dtantsur, line 91: I wonder if we need any logging in this function. Otherwise debugging can be challenging.

Or can we maybe return an `error` instead and log it somewhere at the caller place?

dtantsur (commented):
/approve

Looks good to me, but please consider fixing my comments in a follow-up.

apis/metal3.io/v1alpha1/firmwareschema_types.go
bfournie: Yes, for non-redfish types there will be no BIOS registry data returned and therefore no schema, so no validations can be done on the settings. I will note this in the CRD.

apis/metal3.io/v1alpha1/firmwareschema_types.go
bfournie: Good point about the debugging, I will add more granular debugging to help determine error in settings.

apis/metal3.io/v1alpha1/firmwareschema_types.go
dtantsur: Well, other drivers may decide to implement schema as well.

apis/metal3.io/v1alpha1/firmwareschema_types.go
bfournie: Changed CheckSettingIsValid() in https://github.com/metal3-io/baremetal-operator/pull/938 to return an error which is now logged.

PR #906, opened 2021-06-01
Increase the default $BMO_CONCURRENCY for scale

### LINKED ISSUE #905: Optimise concurrency for scale
Currently the default value for the `$BMO_CONCURRENCY` variable is 3. We chose this value because (a) it wasn't 1 (so in testing we would not end up relying on only 1 BMH being reconciled at a time), and (b) it seemed like as good a number as any. With the benefit of more data, we can say a bit more about the optimal value.

We currently limit the number of nodes that can be provisioned at one time (`$PROVISIONING_LIMIT`) to 20 to limit the load on Ironic. Note that enforcement of the provisioning limit cannot be made thread-safe, so while it isn't a hard limit, we should endeavour to keep `$BMO_CONCURRENCY` << `$PROVISIONING_LIMIT` to avoid smashing right through it. Most of the ironic provisioning work happens asynchronously from the perspective of the baremetal-operator; we only schedule a reconcile to check on it every 10s.

However, before a new Host can start provisioning, it must pass through a number of states, including those that don't do ~anything (e.g. `match profile`, or `inspecting` when inspection is disabled). If a large number of Hosts are added simultaneously, they will have a tendency to go through this process in lockstep, since as each is advanced it will go to the back of the queue before it will get processed again. Ideally we want at least 20 Hosts to be available for provisioning as quickly as possible.

I wrote a little [simulation](https://gist.github.com/zaneb/d48bb5bb42fc40799ef846726392157a)  to experiment with different settings. If we assume 1000 new Hosts (why not) added at once and 20 reconcile steps (wild guess) each taking around 100ms (pretty accurate when there is no work to do), then it would take more than 10 minutes for the first 100 nodes to become available for provisioning. Real-world testing with 1000 nodes created simultaneously shows that indeed, very few hosts are provisioned in the first 10-20 minutes. This can be improved by increasing the number of concurrent reconciles allowed.

Another option is to add some random small delay when requeuing, instead of requeuing without delay. As each reconcile step is worked through, the Hosts should tend to spread out in time, so that some reach the provisioning state sooner than others. Using the same simulation, the optimal amount of jitter to add is:

```
1.9 * <Total number of hosts> * <Average reconcile time> / <Number of concurrent reconciles>
```

(I haven't determined the mathematical significance of the value 1.9, but it's accurate to 2s.f.)

Plotting the effect of both interventions:

thread_time

Adding jitter reduces the time for the first 100 hosts to go through 20 reconcile steps by about 10-15%, but increasing the concurrency has a much, much larger effect. A default value of 8 for `$BMO_CONCURRENCY` would seem to capture most of the returns at this scale.

zaneb:
Instead of defaulting the $BMO_CONCURRENCY value (the maximum number of
Hosts to reconcile concurrently) to a hard-coded value of 3, set it
instead to the number of CPU threads available, but constrained to a
range between 2 and 8.

We never default to 1 so that we don't inadvertantly rely on
single-threadedness in tests. 8 seems to be a reasonable value for a
large scale deployment, while still being substantially below the
default $PROVISIONING_LIMIT of 20.

There remains no restriction on the value that can be passed in the
environment variable.

Fixes #905

### DIFF
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"os"
 	"reflect"
+	"runtime"
 	"strconv"
 	"strings"
 	"time"
@@ -1312,7 +1313,13 @@ func (r *BareMetalHostReconciler) updateEventHandler(e event.UpdateEvent) bool {
 // SetupWithManager reigsters the reconciler to be run by the manager
 func (r *BareMetalHostReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
-	maxConcurrentReconciles := 3
+	maxConcurrentReconciles := runtime.NumCPU()
+	if maxConcurrentReconciles > 8 {
+		maxConcurrentReconciles = 8
+	}
+	if maxConcurrentReconciles < 2 {
+		maxConcurrentReconciles = 2
+	}
 	if mcrEnv, ok := os.LookupEnv("BMO_CONCURRENCY"); ok {
 		mcr, err := strconv.Atoi(mcrEnv)
 		if err != nil {
--- a/controllers/metal3.io/host_state_machine.go
+++ b/controllers/metal3.io/host_state_machine.go
@@ -291,6 +291,10 @@ func (hsm *hostStateMachine) ensureRegistered(info *reconcileInfo) (result actio
 		// We haven't yet reached the Registration state, so don't attempt
 		// to register the Host.
 		return
+	case metal3v1alpha1.StateMatchProfile:
+		// We don't call the provisioner in this state, so there is no point
+		// in checking the registration.
+		return
 	case metal3v1alpha1.StateDeleting:
 		// In the deleting state the whole idea is to de-register the host
 		return
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -33,7 +33,7 @@ client certificate authentication (mTLS) to be enabled.
 client certificate SAN validation.
 
 `BMO_CONCURRENCY` -- The number of concurrent reconciles performed by the
-Operator. Default is 3.
+Operator. Default is the number of CPUs, but no less than 2 and no more than 8.
 
 `PROVISIONING_LIMIT` -- The desired maximum number of hosts that could be (de)provisioned
 simultaneously by the Operator. The Operator will try to enforce this limit,


flaper87 (commented):
This looks good! Thanks for the PR

flaper87 (commented):
/lgtm

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1312,7 +1313,13 @@ func (r *BareMetalHostReconciler) updateEventHandler(e event.UpdateEvent) bool {
 // SetupWithManager reigsters the reconciler to be run by the manager
 func (r *BareMetalHostReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
-	maxConcurrentReconciles := 3
+	maxConcurrentReconciles := runtime.NumCPU()
+	if maxConcurrentReconciles > 8 {
```
andfasano, line 1317: Instead of using an hard-coded value, I was thinking that it could be useful to set as the upper limit something like `min(PROVISIONIG_LIMIT/2.5, 1)`, just to enforce the ratio between the two values _in case_ the user will set a particularly low value for `PROVISIONING_LIMIT`.

zaneb: > Is there a plan to keep this configurable as an option (via env)?

It is still configurable.

zaneb: > Instead of using an hard-coded value, I was thinking that it could be useful to set as the upper limit something like `min(PROVISIONIG_LIMIT/2.5, 1)`, just to enforce the ratio between the two values _in case_ the user will set a particularly low value for `PROVISIONING_LIMIT`.

I can't imagine anyone setting a smaller limit except on very small hardware, in which case the CPU count should scale the default appropriately for us. I can imagine people setting a bigger limit, but I don't think we want to increase the default concurrency in that case. I don't think we should tie these two settings together because ultimately they limit different things.

andfasano: It's not about tying them together, it's just about the currently proposed value for the _default_  upper limit of `maxConcurrentReconciles` - and I think there was an error in my previous comment, as I meant something like:

`maxConcurrentReconciles = min(PROVISIONING_LIMIT/2.5, runtime.NumCPU())`

which for a default PROVISIONING_LIMIT value of 20 will limit maxConcurrentReconciles to 8 on a system with more than 8 cores.

Given that the current proposal is to define the _default_ value based on the number of available CPU, the basic idea is to  try using all the available cpus as much possible - even though the benefits could be marginal (in general I'd expect that BMO will run within a container, which could have already resources and/or limits set), and at the same time trying to honor the fact that `maxConcurrentReconciles << PROVISIONING_LIMIT`

furkatgofurov7 (commented):
This looks good, the only thing is we have to update the [documentation](https://github.com/metal3-io/baremetal-operator/blob/fa1773f003d0bd8bfccdb5906d57eb204d2b1c3e/docs/configuration.md#L35) whether we hard-code or set the upper limit to $BMO_CONCURRENCY accordingly.

PR #907, opened 2021-06-02
Support deploy_iso in addition to deploy_kernel/ramdisk

dtantsur:
Redfish virtual media supports pre-generated ISO ramdisks in addition
to a kernel/initramfs pair. Other drivers will catch up as well once
https://storyboard.openstack.org/#!/story/2008880 is fixed.


elfosardo (approved):

pkg/provisioner/ironic/ironic.go
```diff
@@ -94,14 +95,12 @@ func init() {
...
-	if deployKernelURL == "" {
-		fmt.Fprintf(os.Stderr, "Cannot start: No DEPLOY_KERNEL_URL variable set\n")
-		os.Exit(1)
-	}
 	deployRamdiskURL = os.Getenv("DEPLOY_RAMDISK_URL")
-	if deployRamdiskURL == "" {
-		fmt.Fprintf(os.Stderr, "Cannot start: No DEPLOY_RAMDISK_URL variable set\n")
+	deployISOURL = os.Getenv("DEPLOY_ISO_URL")
+	if deployISOURL == "" && (deployKernelURL == "" || deployRamdiskURL == "") {
+		fmt.Fprintf(os.Stderr, "Cannot start: Either DEPLOY_ISO_URL or DEPLOY_KERNEL_URL and DEPLOY_RAMDISK_URL must be set\n")
```
andfasano, line 101: I tried to read this error message from the point of view of the user, and it looks like a little bit tricky. Wondering if it's worh splitting the checks, validating first `DEPLOY_RAMDISK_URL` and, if not set, validating `DEPLOY_KERNEL_URL` and `DEPLOY_RAMDISK_URL` as usual

pkg/provisioner/ironic/ironic.go
```diff
@@ -510,8 +510,13 @@ func (p *ironicProvisioner) ValidateManagementAccess(data provisioner.Management
...
 	// provisioning network from somewhere.
-	driverInfo["deploy_kernel"] = deployKernelURL
-	driverInfo["deploy_ramdisk"] = deployRamdiskURL
+	if deployKernelURL != "" && deployRamdiskURL != "" {
+		driverInfo["deploy_kernel"] = deployKernelURL
+		driverInfo["deploy_ramdisk"] = deployRamdiskURL
+	}
+	if deployISOURL != "" {
+		driverInfo["deploy_iso"] = deployISOURL
+	}
```
andfasano, line 519: It would be nice to cover this part with a small test

pkg/provisioner/ironic/ironic.go
dtantsur: Not sure I get what you mean. Could you write pseudo-code for your proposal?

pkg/provisioner/ironic/ironic.go
dtantsur: I think this code is not unit tested at all, but I'll double-check.

pkg/provisioner/ironic/ironic.go
andfasano: Sorry, just noticed a typo in my comment, anyhow by reviewing the changes I realized that some points are still unclear to me, in particular which are the allowed/valid combinations? From the code it seems that `ISO`, `ISO/RAMDISK`, `ISO/KERNEL`,`ISO/RAMDISK/KERNEL` and `KERNEL/RAMDISK` could pass the validation check (while ` `, `KERNEL` and `RAMDISK` will fail)?

pkg/provisioner/ironic/ironic.go
andfasano: Yeah it could be, we started to cover the `ValidateManagementAccess` some time ago but there were still some gap. A test for this part could also serve as a useful documentation for the allowed cases.

pkg/provisioner/ironic/ironic.go
zaneb: Classic operator precedence problem :laughing:

"Either DEPLOY_KERNEL_URL and DEPLOY_RAMDISK_URL or DEPLOY_ISO_URL must be set" would be unambiguous.

zaneb:
Q: do we expect the ISO-only option to work with _all_ drivers (in the foreseeable future), or just virtualmedia ones?

dtantsur:
> Q: do we expect the ISO-only option to work with all drivers (in the foreseeable future), or just virtualmedia ones?

Only virtual media for the foreseeable future. So we'll need to keep kernel/ramdisk support for machines that don't have virtual media.

hardys:
> > Q: do we expect the ISO-only option to work with all drivers (in the foreseeable future), or just virtualmedia ones?
> 
> Only virtual media for the foreseeable future. So we'll need to keep kernel/ramdisk support for machines that don't have virtual media.

But to clarify - Ironic does support iPXE boot of the ISO image, we just know that doesn't currently work with the iso images we're working with?

I got reports of someone trying to boot a RHEL ISO with `format: live-iso` and (non-virtualmedia) redfish BMC URL, that failed at the dracut stage (similar to FCOS IIRC), so I don't think there's anything at the metal3 or Ironic layer which prevents potentially PXE booting an iso in future, if/when we figure out how to make that work?

dtantsur:
> Ironic does support iPXE boot of the ISO image, we just know that doesn't currently work with the iso images we're working with?

Ironic supports iPXE live ISO boot. We're not planning on supporting booting **IPA** ISOs via iPXE at the moment.

pkg/provisioner/ironic/ironic.go
```diff
@@ -82,36 +83,23 @@ func NewMacAddressConflictError(address, node string) error {
...
 	deployRamdiskURL = os.Getenv("DEPLOY_RAMDISK_URL")
-	if deployRamdiskURL == "" {
-		fmt.Fprintf(os.Stderr, "Cannot start: No DEPLOY_RAMDISK_URL variable set\n")
-		os.Exit(1)
+	deployISOURL = os.Getenv("DEPLOY_ISO_URL")
+	if deployISOURL == "" && (deployKernelURL == "" || deployRamdiskURL == "") {
+		return errors.New("Either DEPLOY_KERNEL_URL and DEPLOY_RAMDISK_URL or DEPLOY_ISO_URL must be set")
+	}
+	if (deployKernelURL == "" && deployRamdiskURL != "") || (deployKernelURL != "" && deployRamdiskURL == "") {
+		return errors.New("DEPLOY_KERNEL_URL and DEPLOY_RAMDISK_URL can only be set together")
```
andfasano, line 94: Ok, my personal inclination would be to specify exactly which expected vars were not set (without forcing the user to look at the current container configuration, especially useful when you have just the logs), but it's anyhow an improvement and don't consider it blocking

pkg/provisioner/ironic/ironic.go
andfasano: I also checked and looks like there's not a direct way to check, probably the mock will need some additional retouch (in case `TestValidateManagementAccessCreateWithLiveIso` could be a good starting point to look at).

andfasano (approved):
/lgtm

pkg/provisioner/ironic/init_test.go
```diff
@@ -0,0 +1,152 @@
...
+			err := loadConfigFromEnv()
+			if tc.expectedError != "" {
+				assert.Regexp(t, tc.expectedError, err, "error message")
+			} else {
+				assert.Nil(t, err)
+				tc.env.Verify(t)
+			}
+		})
+	}
+}
```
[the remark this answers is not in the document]
zaneb: Moved this to the top of my list: #914

PR #908, opened 2021-06-08
Avoid always do delete_configuration clean step

Hellcatlk:

When I test on a server(irmc) without a RAID controller, this will cause an error.
I am very sorry for ignored this situation in the previous code.

### DIFF
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -775,15 +775,15 @@ func (r *BareMetalHostReconciler) actionPreparing(prov provisioner.Provisioner, 
 	}
 
 	prepareData := provisioner.PrepareData{
-		TargetRAIDConfig:  newStatus.Provisioning.RAID.DeepCopy(),
-		ExistedRAIDConfig: info.host.Status.Provisioning.RAID,
-		RootDeviceHints:   newStatus.Provisioning.RootDeviceHints.DeepCopy(),
-		FirmwareConfig:    newStatus.Provisioning.Firmware.DeepCopy(),
+		TargetRAIDConfig: newStatus.Provisioning.RAID.DeepCopy(),
+		ActualRAIDConfig: info.host.Status.Provisioning.RAID.DeepCopy(),
+		RootDeviceHints:  newStatus.Provisioning.RootDeviceHints.DeepCopy(),
+		FirmwareConfig:   newStatus.Provisioning.Firmware.DeepCopy(),
 	}
 	// When manual cleaning fails, we think that the existed RAID configuration
 	// is invalid and needs to be reconfigured.
 	if info.host.Status.ErrorType == metal3v1alpha1.PreparationError {
-		prepareData.ExistedRAIDConfig = nil
+		prepareData.ActualRAIDConfig = nil
 		dirty = true
 	}
 	provResult, started, err := prov.Prepare(prepareData, dirty)
@@ -896,7 +896,7 @@ func (r *BareMetalHostReconciler) actionProvisioning(prov provisioner.Provisione
 // fields of a host.
 func clearHostProvisioningSettings(host *metal3v1alpha1.BareMetalHost) {
 	host.Status.Provisioning.RootDeviceHints = nil
-	// Only clear software raid volumes
+	// Keep `HardwareRAIDVolumes` to avoid configuring the same hardware RAID repeatedly
 	if host.Status.Provisioning.RAID != nil {
 		host.Status.Provisioning.RAID.SoftwareRAIDVolumes = nil
 	}
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -775,12 +775,18 @@ func (r *BareMetalHostReconciler) actionPreparing(prov provisioner.Provisioner, 
 	}
 
 	prepareData := provisioner.PrepareData{
-		RAIDConfig:      newStatus.Provisioning.RAID.DeepCopy(),
-		RootDeviceHints: newStatus.Provisioning.RootDeviceHints.DeepCopy(),
-		FirmwareConfig:  newStatus.Provisioning.Firmware.DeepCopy(),
+		TargetRAIDConfig:  newStatus.Provisioning.RAID.DeepCopy(),
+		ExistedRAIDConfig: info.host.Status.Provisioning.RAID,
+		RootDeviceHints:   newStatus.Provisioning.RootDeviceHints.DeepCopy(),
+		FirmwareConfig:    newStatus.Provisioning.Firmware.DeepCopy(),
 	}
-	provResult, started, err := prov.Prepare(prepareData,
-		dirty || info.host.Status.ErrorType == metal3v1alpha1.PreparationError)
+	// When manual cleaning fails, we think that the existed RAID configuration
+	// is invalid and needs to be reconfigured.
+	if info.host.Status.ErrorType == metal3v1alpha1.PreparationError {
+		prepareData.ExistedRAIDConfig = nil
+		dirty = true
+	}
+	provResult, started, err := prov.Prepare(prepareData, dirty)
 	if err != nil {
 		return actionError{errors.Wrap(err, "error preparing host")}
 	}
@@ -890,7 +896,10 @@ func (r *BareMetalHostReconciler) actionProvisioning(prov provisioner.Provisione
 // fields of a host.
 func clearHostProvisioningSettings(host *metal3v1alpha1.BareMetalHost) {
 	host.Status.Provisioning.RootDeviceHints = nil
-	host.Status.Provisioning.RAID = nil
+	// Only clear software raid volumes
+	if host.Status.Provisioning.RAID != nil {
+		host.Status.Provisioning.RAID.SoftwareRAIDVolumes = nil
+	}
 	host.Status.Provisioning.Firmware = nil
 }
 
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -1121,7 +1121,7 @@ func saveHostProvisioningSettings(host *metal3v1alpha1.BareMetalHost) (dirty boo
 	}
 
 	// Copy RAID settings
-	specRAID := host.Spec.RAID.DeepCopy()
+	specRAID := host.Spec.RAID
 	// If RAID configure is nil or empty, means that we need to keep the current hardware RAID configuration
 	// or clear current software RAID configuration
 	if specRAID == nil || reflect.DeepEqual(specRAID, &metal3v1alpha1.RAIDConfig{}) {
@@ -1132,13 +1132,9 @@ func saveHostProvisioningSettings(host *metal3v1alpha1.BareMetalHost) (dirty boo
 		// }
 		specRAID = &metal3v1alpha1.RAIDConfig{}
 		if host.Status.Provisioning.RAID != nil && len(host.Status.Provisioning.RAID.HardwareRAIDVolumes) != 0 {
-			specRAID.HardwareRAIDVolumes = host.Status.Provisioning.RAID.DeepCopy().HardwareRAIDVolumes
+			specRAID.HardwareRAIDVolumes = host.Status.Provisioning.RAID.HardwareRAIDVolumes
 		}
 		specRAID.SoftwareRAIDVolumes = []metal3v1alpha1.SoftwareRAIDVolume{}
-	}
-	// If HardwareRAIDVolumes isn't nil, ignore SoftwareRAIDVolumes
-	if specRAID.HardwareRAIDVolumes != nil {
-		specRAID.SoftwareRAIDVolumes = nil
 	}
 	if !reflect.DeepEqual(host.Status.Provisioning.RAID, specRAID) {
 		host.Status.Provisioning.RAID = specRAID
@@ -1147,7 +1143,7 @@ func saveHostProvisioningSettings(host *metal3v1alpha1.BareMetalHost) (dirty boo
 
 	// Copy BIOS settings
 	if !reflect.DeepEqual(host.Status.Provisioning.Firmware, host.Spec.Firmware) {
-		host.Status.Provisioning.Firmware = host.Spec.Firmware.DeepCopy()
+		host.Status.Provisioning.Firmware = host.Spec.Firmware
 		dirty = true
 	}
 
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -1112,41 +1112,28 @@ func saveHostProvisioningSettings(host *metal3v1alpha1.BareMetalHost) (dirty boo
 	}
 
 	// Copy RAID settings
-	if host.Spec.RAID != host.Status.Provisioning.RAID {
-		// If RAID settings is nil, remove saved settings,
-		// else check hardware RAID and software RAID.
-		if host.Spec.RAID == nil {
-			host.Status.Provisioning.RAID = nil
-			dirty = true
-		} else {
-			if host.Status.Provisioning.RAID == nil {
-				host.Status.Provisioning.RAID = &metal3v1alpha1.RAIDConfig{}
-				dirty = true
-			}
-			// If HardwareRAIDVolumes isn't nil, we will ignore SoftwareRAIDVolumes.
-			if len(host.Spec.RAID.HardwareRAIDVolumes) != 0 {
-				// If software RAID has been saved, remove it.
-				if len(host.Status.Provisioning.RAID.SoftwareRAIDVolumes) != 0 {
-					host.Status.Provisioning.RAID.SoftwareRAIDVolumes = nil
-				}
-				// Compare hardware RAID settings
-				if !reflect.DeepEqual(host.Spec.RAID.HardwareRAIDVolumes, host.Status.Provisioning.RAID.HardwareRAIDVolumes) {
-					host.Status.Provisioning.RAID.HardwareRAIDVolumes = host.Spec.RAID.HardwareRAIDVolumes
-					dirty = true
-				}
-			} else {
-				// If hardware RAID has been saved, remove it.
-				if len(host.Status.Provisioning.RAID.HardwareRAIDVolumes) != 0 {
-					host.Status.Provisioning.RAID.HardwareRAIDVolumes = nil
-					dirty = true
-				}
-				// Compare software RAID settings
-				if !reflect.DeepEqual(host.Spec.RAID.SoftwareRAIDVolumes, host.Status.Provisioning.RAID.SoftwareRAIDVolumes) {
-					host.Status.Provisioning.RAID.SoftwareRAIDVolumes = host.Spec.RAID.SoftwareRAIDVolumes
-					dirty = true
-				}
-			}
+	specRAID := host.Spec.RAID.DeepCopy()
+	// If RAID configure is nil or empty, means that we need to keep the current hardware RAID configuration
+	// or clear current software RAID configuration
+	if specRAID == nil || reflect.DeepEqual(specRAID, &metal3v1alpha1.RAIDConfig{}) {
+		// Set the default value of RAID configure:
+		// {
+		//     HardwareRAIDVolumes: nil or Status.Provisioning.RAID.HardwareRAIDVolumes(not empty),
+		//     SoftwareRAIDVolume: [],
+		// }
+		specRAID = &metal3v1alpha1.RAIDConfig{}
+		if host.Status.Provisioning.RAID != nil && len(host.Status.Provisioning.RAID.HardwareRAIDVolumes) != 0 {
+			specRAID.HardwareRAIDVolumes = host.Status.Provisioning.RAID.DeepCopy().HardwareRAIDVolumes
 		}
+		specRAID.SoftwareRAIDVolumes = []metal3v1alpha1.SoftwareRAIDVolume{}
+	}
+	// If HardwareRAIDVolumes isn't nil, ignore SoftwareRAIDVolumes
+	if specRAID.HardwareRAIDVolumes != nil {
+		specRAID.SoftwareRAIDVolumes = nil
+	}
+	if !reflect.DeepEqual(host.Status.Provisioning.RAID, specRAID) {
+		host.Status.Provisioning.RAID = specRAID
+		dirty = true
 	}
 
 	// Copy BIOS settings
--- a/pkg/provisioner/ironic/raid.go
+++ b/pkg/provisioner/ironic/raid.go
@@ -19,18 +19,21 @@ const (
 )
 
 // setTargetRAIDCfg set the RAID settings to the ironic Node for RAID configuration steps
-func setTargetRAIDCfg(p *ironicProvisioner, raidInterface string, ironicNode *nodes.Node, data provisioner.PrepareData) (err error) {
-	err = checkRAIDConfigure(raidInterface, data.TargetRAIDConfig)
+func setTargetRAIDCfg(p *ironicProvisioner, raidInterface string, ironicNode *nodes.Node, data provisioner.PrepareData) (provisioner.Result, error) {
+	err := checkRAIDConfigure(raidInterface, data.TargetRAIDConfig)
 	if err != nil {
-		return err
+		return operationFailed(err.Error())
 	}
 
 	var logicalDisks []nodes.LogicalDisk
 
 	// Build target for RAID configuration steps
 	logicalDisks, err = BuildTargetRAIDCfg(data.TargetRAIDConfig)
-	if len(logicalDisks) == 0 || err != nil {
-		return
+	if err != nil {
+		return operationFailed(err.Error())
+	}
+	if len(logicalDisks) == 0 {
+		return provisioner.Result{}, nil
 	}
 
 	// set root volume
@@ -42,11 +45,15 @@ func setTargetRAIDCfg(p *ironicProvisioner, raidInterface string, ironicNode *no
 	}
 
 	// Set target for RAID configuration steps
-	return nodes.SetRAIDConfig(
+	err = nodes.SetRAIDConfig(
 		p.client,
 		ironicNode.UUID,
 		nodes.RAIDConfigOpts{LogicalDisks: logicalDisks},
 	).ExtractErr()
+	if err != nil {
+		return transientError(err)
+	}
+	return provisioner.Result{}, nil
 }
 
 // BuildTargetRAIDCfg build RAID logical disks, this method doesn't set the root volume
--- a/pkg/provisioner/ironic/raid.go
+++ b/pkg/provisioner/ironic/raid.go
@@ -2,6 +2,7 @@ package ironic
 
 import (
 	"fmt"
+	"reflect"
 
 	"github.com/gophercloud/gophercloud/openstack/baremetal/v1/nodes"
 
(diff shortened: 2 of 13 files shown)


dtantsur:
/approve

Could you update the design to reflect this change?

zaneb:
I fear this will prevent the user from disabling RAID once it has been set up (even deleting the BareMetalHost and starting again won't help; you'd have to do it manually in the BMC).

Specifically, I think it relies on an assumption that there is a distinction between a missing RAID config and a RAID config where both the Hardware and Software volume lists are empty. However, given that all 3 of these are marked `omitempty` I strongly suspect that will not be the case (i.e. if no volumes are specified then the RAID config will be nil). If this is not merely an assumption and you have verified that there is a difference, please confirm.

I myself have attempted to solve the problem of extra manual cleans occurring because of the delete_configuration step (before I was aware that it is breaking irmc hosts that do not support RAID), but I came to the conclusion that there was very little that could be done about it while remaining correct. For example, if a new Host is added with an empty RAID config, I think we do still want to clear it just in case RAID was previously enabled before we enrolled that Host. For reference, here is the patch and my notes: https://github.com/zaneb/baremetal-operator/commit/8b5b575e3beb9d23221262c1f755c35b34ee559a

Obviously the fact that this is breaking some hosts gives new urgency to the problem. But I'm not sure that we can do the right thing without some indication either from ironic or from the user that the host does or does not support RAID.

Hellcatlk:
> Specifically, I think it relies on an assumption that there is a distinction between a missing RAID config and a RAID config where both the Hardware and Software volume lists are empty. However, given that all 3 of these are marked omitempty I strongly suspect that will not be the case (i.e. if no volumes are specified then the RAID config will be nil). If this is not merely an assumption and you have verified that there is a difference, please confirm.

@zaneb If the raid field does not exist or is empty, the value of host.Spec.RAID is:
`2021-06-08T22:08:09.642-0400    INFO    controllers.BareMetalHost       DEBUG     {"baremetalhost": "metal3/example-baremetalhost", "raid": null}`

If the raid field isn't empty, eg:
```yaml
spec:
  raid:
    hardwareRAIDVolumes:
```
The value of host.Spec.RAID is:
`2021-06-08T22:08:34.936-0400    INFO    controllers.BareMetalHost       DEBUG     {"baremetalhost": "metal3/example-baremetalhost", "raid": {}}`

zaneb:
OK, thanks for checking that. So we're relying on a very subtle distinction that:

```
spec:
  raid:
```

means use whatever RAID settings happen to exist, but

```
spec:
  raid:
    hardwareRAIDVolumes:
```

means clear any existing hardware volumes.

As subtle as that is, I don't think we have any choice. On deprovisioning we set the RAID config in the status to nil, which is equivalent to 'keep as is', so that's the right thing in terms of deciding whether we retrigger Preparing again or not. (i.e. if the Spec does not tell us to leave it as is, we will trigger Preparing and delete the existing configuration.)


Second question: it appears that under this scheme both hardware and software RAID follow the same rules with respect to `delete_configuration`. It follows that users will be unable to use software RAID on hosts that don't have a hardware RAID controller - which naively I would have thought was the point of software RAID.

It seems to me that what we actually want is to do is to `delete_configuration` when we have:

```
spec:
  raid:
    hardwareRAIDVolumes:
```

but `erase_devices_metadata` the rest of the time, including when we have:

```
spec:
  raid:
    softwareRAIDVolumes:
```

but we don't currently have a way of telling the cases in those two examples apart. Perhaps getting rid of `omitempty` and adding a `// +optional` comment instead might give us the right behaviour (giving us either nil or an empty slice) without changing the CRD schema? Or possibly the user would still have to go even further and put `hardwareRaidVolumes: []`.

(Also, while it could make sense to keep the externally configured hardware RAID settings, IMHO it never makes sense to rely on existing software RAID configuration because you can't know how those might have been trashed by a previous installation. I'd have to think more about how this could be managed.)

Hellcatlk:
> but we don't currently have a way of telling the cases in those two examples apart. Perhaps getting rid of `omitempty` and adding a `// +optional` comment instead might give us the right behaviour (giving us either nil or an empty slice) without changing the CRD schema? Or possibly the user would still have to go even further and put `hardwareRaidVolumes: []`.

I used `// +optional` instead of `omitempty` 

For
```yaml
spec:
  raid:
```
It's value is `"raid": null`

For
```yaml
spec:
  raid:
    hardwareRAIDVolume:
```
It's value is `"raid": {"hardwareRAIDVolumes":null,"softwareRAIDVolumes":null}`

Hellcatlk:
@zaneb If you assign a value to hardwareRAIDVolumes in the following way: 
```yaml
spec:
  raid:
    hardwareRAIDVolumes: []
```
We can distinguish them.
2021-06-09T00:00:57.300-0400    INFO    controllers.BareMetalHost       DEBUG   {"baremetalhost": "metal3/example-baremetalhost", "raid": {}}
2021-06-09T00:00:57.300-0400    INFO    controllers.BareMetalHost       DEBUG   {"baremetalhost": "metal3/example-baremetalhost", "hardware": []}
2021-06-09T00:00:57.300-0400    INFO    controllers.BareMetalHost       DEBUG   {"baremetalhost": "metal3/example-baremetalhost", "software": null}

Hellcatlk:
The reason for the failure of test-integration seems to be `Invalid CAPM3 version : v1alpha5. Not in : v1alpha4`

furkatgofurov7:
> The reason for the failure of test-integration seems to be `Invalid CAPM3 version : v1alpha5. Not in : v1alpha4`

@Hellcatlk hi! We have been working on some changes in CI and this PR was affected, sorry. We also notified about it over the [slack](https://kubernetes.slack.com/archives/CHD49TLE7/p1623234051111900). I re-triggered it again, since fix for the problem has already landed, hopefully should be fine this time.

Hellcatlk:
@zaneb Although I do not think that PR is the best solution to this problem, but this is the simplest solution I could find currently. And besides that, we seem to have no choice, unless we can know whether a server has a RAID controller through `ironic`.

In addition, I will update the design document after this PR is merged to make it consistent with the implementation.

zaneb:
I started looking at the other things we'll likely need to do to make sure we don't end up in the wrong state: #910 (incomplete and I haven't bothered fixing the unit tests yet).

In the meantime I think this is probably a reasonable way forward to allow us to fix the breakage and keep working. It looks like we need some way for the user to distinguish between "clear all existing HW RAID config" and "use existing hardware RAID config". We should document it as recommending you pass an empty list to get the former, and nothing to get the latter.

Hellcatlk:
>In the meantime I think this is probably a reasonable way forward to allow us to fix the breakage and keep working. It looks like we need some way for the user to distinguish between "clear all existing HW RAID config" and "use existing hardware RAID config". We should document it as recommending you pass an empty list to get the former, and nothing to get the latter.

You mean, only the following three situations will clear the RAID configuration?
```yaml
spec:
  # Clear hardware RAID configuration
  raid:
    hardwareRAIDVolumes: []
```

```yaml
spec:
  # Clear hardware RAID configuration
  raid:
     hardwareRAIDVolumes: []
     softwareRAIDVolumes: []
``` 

```yaml
spec:
  # Clear softwareRAID configuration
  raid:
     softwareRAIDVolumes: []
```

Hellcatlk:
@zaneb I am sorry but I have to say that the most urgent thing right now is to catch up with OCP4.8, so can we just take this as shor-term fix for now? If disagree, I can open a new PR to just change the irmc raid interface to "no-raid". WDYH?

zaneb:
> Do you mean that only the following three situations will clear the RAID configuration?

I mean we should document it that way (even if it's not true that those are the _only_ ways to clear it) so that we have the flexibility to redefine what

```
spec:
  # ¯\_(ツ)_/¯
  raid:
    hardwareRAIDVolumes:
```

means.

Hellcatlk:
> 
> 
> > Do you mean that only the following three situations will clear the RAID configuration?
> 
> I mean we should document it that way (even if it's not true that those are the _only_ ways to clear it) so that we have the flexibility to redefine what
> 
> ```
> spec:
>   # ¯\_(ツ)_/¯
>   raid:
>     hardwareRAIDVolumes:
> ```
> 
> means.

Thanks, I add description to api.md.

Hellcatlk:
@zaneb I updated some logic, and now software RAID is always deleted.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -757,12 +757,18 @@ func (r *BareMetalHostReconciler) actionPreparing(prov provisioner.Provisioner,
...
-		RootDeviceHints: newStatus.Provisioning.RootDeviceHints.DeepCopy(),
-		FirmwareConfig:  newStatus.Provisioning.Firmware.DeepCopy(),
+		TargetRAIDConfig:  newStatus.Provisioning.RAID.DeepCopy(),
+		ExistedRAIDConfig: info.host.Status.Provisioning.RAID,
+		RootDeviceHints:   newStatus.Provisioning.RootDeviceHints.DeepCopy(),
+		FirmwareConfig:    newStatus.Provisioning.Firmware.DeepCopy(),
+	}
+	// When manual cleaning fails, we think that the existed RAID configuration
+	// is invalid and needs to be reconfigured.
+	if info.host.Status.ErrorType == metal3v1alpha1.PreparationError {
```
Hellcatlk, line 767: @zaneb 
> How can we avoid assuming that there is no RAID configured on the host after an error occurs in Preparing?

We seem to be able to avoid this problem in this way 。

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -890,7 +896,10 @@ func (r *BareMetalHostReconciler) actionProvisioning(prov provisioner.Provisione
 // fields of a host.
 func clearHostProvisioningSettings(host *metal3v1alpha1.BareMetalHost) {
 	host.Status.Provisioning.RootDeviceHints = nil
-	host.Status.Provisioning.RAID = nil
+	// Only clear software raid volumes
+	if host.Status.Provisioning.RAID != nil {
+		host.Status.Provisioning.RAID.SoftwareRAIDVolumes = nil
+	}
```
andfasano, line 902: What's the reason to cleanup only the software RAID here?

pkg/provisioner/ironic/raid.go
```diff
@@ -138,7 +144,57 @@ func buildTargetSoftwareRAIDCfg(volumes []metal3v1alpha1.SoftwareRAIDVolume) (lo
...
+				{
+					Interface: "raid",
+					Step:      "delete_configuration",
+				},
+				{
+					Interface: "deploy",
+					Step:      "erase_devices_metadata",
+				},
+			}...,
+		)
```
andfasano, line 172: Looks like the above steps are configured always, in case of the software RAID, regardless the current status vs the target one (whereas for the hardware RAID, later, a `DeepEqual` check is performed to detect an eventual discrepancy). Wouldn't be required a similar check also here before starting to assemble the cleansteps?

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1112,41 +1121,28 @@ func saveHostProvisioningSettings(host *metal3v1alpha1.BareMetalHost) (dirty boo
...
+		// }
+		specRAID = &metal3v1alpha1.RAIDConfig{}
+		if host.Status.Provisioning.RAID != nil && len(host.Status.Provisioning.RAID.HardwareRAIDVolumes) != 0 {
+			specRAID.HardwareRAIDVolumes = host.Status.Provisioning.RAID.DeepCopy().HardwareRAIDVolumes
 		}
+		specRAID.SoftwareRAIDVolumes = []metal3v1alpha1.SoftwareRAIDVolume{}
+	}
+	// If HardwareRAIDVolumes isn't nil, ignore SoftwareRAIDVolumes
+	if specRAID.HardwareRAIDVolumes != nil {
+		specRAID.SoftwareRAIDVolumes = nil
```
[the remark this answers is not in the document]
Hellcatlk: Our logic time ignores `softwaraRAIDVolumes` when `hardwareRAIDVolumes` isn't nil, so I think softwareRAIDVolume can not affect the value of dirty flag in this case.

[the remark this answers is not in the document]
Hellcatlk: Our logic time ignores `softwaraRAIDVolumes` when `hardwareRAIDVolumes` isn't nil, so I think softwareRAIDVolume can not affect the value of dirty flag in this case.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -775,12 +775,18 @@ func (r *BareMetalHostReconciler) actionPreparing(prov provisioner.Provisioner,
 	}
 
 	prepareData := provisioner.PrepareData{
-		RAIDConfig:      newStatus.Provisioning.RAID.DeepCopy(),
-		RootDeviceHints: newStatus.Provisioning.RootDeviceHints.DeepCopy(),
-		FirmwareConfig:  newStatus.Provisioning.Firmware.DeepCopy(),
+		TargetRAIDConfig:  newStatus.Provisioning.RAID.DeepCopy(),
+		ExistedRAIDConfig: info.host.Status.Provisioning.RAID,
```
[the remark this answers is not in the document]
Hellcatlk: Thanks, done.

pkg/provisioner/ironic/raid.go
```diff
@@ -138,7 +144,57 @@ func buildTargetSoftwareRAIDCfg(volumes []metal3v1alpha1.SoftwareRAIDVolume) (lo
...
+		return
+	}
+
+	// Hardware RAID
+	// Ignore SoftwareRAIDVolumes
+	if target != nil {
+		target.SoftwareRAIDVolumes = nil
+	}
+	if existed != nil {
+		existed.SoftwareRAIDVolumes = nil
```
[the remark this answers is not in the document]
Hellcatlk: Sorry, I forgot to call 'DeepCopy()'.

pkg/provisioner/ironic/raid.go
```diff
@@ -172,3 +221,19 @@ func BuildRAIDCleanSteps(raid *metal3v1alpha1.RAIDConfig) (cleanSteps []nodes.Cl
...
+func checkRAIDConfigure(raidInterface string, raid *metal3v1alpha1.RAIDConfig) error {
+	if raidInterface == "no-raid" && raid != nil && (len(raid.HardwareRAIDVolumes) != 0 || len(raid.SoftwareRAIDVolumes) != 0) {
+		return fmt.Errorf("raid settings are defined, but the node's driver %s does not support RAID", raidInterface)
+	}
+
+	if raidInterface == "agent" && raid != nil && len(raid.HardwareRAIDVolumes) != 0 {
+		return fmt.Errorf("node's driver %s does not support hardware RAID", raidInterface)
+	}
+
+	if raidInterface != "agent" && raid != nil && len(raid.HardwareRAIDVolumes) == 0 && len(raid.SoftwareRAIDVolumes) != 0 {
```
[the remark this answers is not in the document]
Hellcatlk: In this case, according to the [design documents](https://github.com/metal3-io/metal3-docs/pull/134/files#diff-3e3740820f58bf5a98b488649cfa00cee6375c7a7122e986b39313f416cad8a7R58-R60), we will ignore 'softwareRAIDVolumes'`.

pkg/provisioner/ironic/raid.go
```diff
@@ -172,3 +221,19 @@ func BuildRAIDCleanSteps(raid *metal3v1alpha1.RAIDConfig) (cleanSteps []nodes.Cl
 	)
 	return
 }
+
+func checkRAIDConfigure(raidInterface string, raid *metal3v1alpha1.RAIDConfig) error {
+	if raidInterface == "no-raid" && raid != nil && (len(raid.HardwareRAIDVolumes) != 0 || len(raid.SoftwareRAIDVolumes) != 0) {
```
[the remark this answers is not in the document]
Hellcatlk: Since we will [set the default value](https://github.com/metal3-io/baremetal-operator/pull/908/files/d3efbfbfce70825786f4b60fabd3490a974dfb4a#diff-a69d9029388ab766ed36b32180145f52785a9d4a153775510dbddfa928a72e1cR1127-R1138) for the empty raid field:
```go
{
		HardwareRAIDVolumes: nil or Status.Provisioning.RAID.HardwareRAIDVolumes(not empty),
		SoftwareRAIDVolume: [],
}
```
So if there is only ` raid= nil` will make mistakes.

controllers/metal3.io/baremetalhost_controller.go
Hellcatlk: Because the hardware raid cannot be modified by the operating system, I think it can be retained and reused.

pkg/provisioner/ironic/raid.go
Hellcatlk: Because I think software RAID may be modified in the process of use, and we can't guarantee that it is consistent with the content in the status.

controllers/metal3.io/baremetalhost_controller.go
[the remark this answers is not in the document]
Hellcatlk: > In addition I'd expect this method to modify just the Status

I just modified the copy of the spec, which will not affect the actual content of the spec.

pkg/provisioner/ironic/raid.go
```diff
@@ -138,7 +144,57 @@ func buildTargetSoftwareRAIDCfg(volumes []metal3v1alpha1.SoftwareRAIDVolume) (lo
...
 // BuildRAIDCleanSteps build the clean steps for RAID configuration from BaremetalHost spec
-func BuildRAIDCleanSteps(raid *metal3v1alpha1.RAIDConfig) (cleanSteps []nodes.CleanStep) {
+func BuildRAIDCleanSteps(raidInterface string, target *metal3v1alpha1.RAIDConfig, existed *metal3v1alpha1.RAIDConfig) (cleanSteps []nodes.CleanStep, err error) {
+	err = checkRAIDConfigure(raidInterface, target)
+	if err != nil {
+		return nil, err
+	}
+
+	// No RAID
+	if raidInterface == "no-raid" {
```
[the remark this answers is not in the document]
Hellcatlk: Done.

controllers/metal3.io/baremetalhost_controller.go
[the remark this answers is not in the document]
andfasano: Sorry @Hellcatlk I was just referring to change `ExistedRAIDConfig`. `TargetRAIDConfig` it's fine and I'd suggest to keep it.

pkg/provisioner/ironic/raid.go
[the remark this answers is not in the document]
andfasano: Fine, maybe a switch could be clearer here (minor point)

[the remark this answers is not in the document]
Hellcatlk: Done.

controllers/metal3.io/baremetalhost_controller.go
Hellcatlk: Done.

pkg/provisioner/ironic/raid.go
Hellcatlk: Software RAID may be changed by the OS. We cannot guarantee that the actual SoftwareRAID and status are the same, so I will always delete SoftwareRAID and clear `Status.Provisioning.RAID.SoftwareRAIDVolumes`.

controllers/metal3.io/baremetalhost_controller.go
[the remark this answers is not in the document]
Hellcatlk: Done.

controllers/metal3.io/baremetalhost_controller.go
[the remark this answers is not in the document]
Hellcatlk: Because `BuildRAIDCleanSteps` will also handle the situation where software RAID and hardware RAID exist at the same time, so I removed the logic to handle this situation from `saveHostProvisioningSettings` to avoid `DeepCopy`.

pkg/provisioner/ironic/raid.go
andfasano: This happens only during the `Preparing` state, so I guess it would be anyway fine for now - still it would be nice to have in future a more robust way to apply software RAID only when required.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1112,46 +1121,29 @@ func saveHostProvisioningSettings(host *metal3v1alpha1.BareMetalHost) (dirty boo
...
+		// Set the default value of RAID configure:
+		// {
+		//     HardwareRAIDVolumes: nil or Status.Provisioning.RAID.HardwareRAIDVolumes(not empty),
+		//     SoftwareRAIDVolume: [],
+		// }
+		specRAID = &metal3v1alpha1.RAIDConfig{}
+		if host.Status.Provisioning.RAID != nil && len(host.Status.Provisioning.RAID.HardwareRAIDVolumes) != 0 {
+			specRAID.HardwareRAIDVolumes = host.Status.Provisioning.RAID.DeepCopy().HardwareRAIDVolumes
 		}
+		specRAID.SoftwareRAIDVolumes = []metal3v1alpha1.SoftwareRAIDVolume{}
```
andfasano, line 1137: Can you please add a log trace here stating that the default RAID values are applied? Otherwise I guess it would be pretty difficult to troubleshoot that when looking at the logs

pkg/provisioner/ironic/raid.go
```diff
@@ -172,3 +226,21 @@ func BuildRAIDCleanSteps(raid *metal3v1alpha1.RAIDConfig) (cleanSteps []nodes.Cl
 	)
 	return
```
andfasano, line 227: Still on the troubleshooting side, do we have any way to easily understand which clean steps are going to be applied? If not, I'd suggest to add log trace here https://github.com/metal3-io/baremetal-operator/blob/62372c1daa01c631da46109078c9872128088f1e/pkg/provisioner/ironic/ironic.go#L1068 for that

Hellcatlk: Here will print the current clean step:
https://github.com/metal3-io/baremetal-operator/blob/62372c1daa01c631da46109078c9872128088f1e/pkg/provisioner/ironic/ironic.go#L1029-L1041

controllers/metal3.io/baremetalhost_controller.go
Hellcatlk: There seems to be a problem with printing logs here, unless I can add a new parameter(logger) to `saveHostProvisioningSettings` or make it a member of `BareMetalHostReconciler`.

pkg/provisioner/ironic/raid.go
Hellcatlk: Done.

pkg/provisioner/ironic/raid.go
andfasano: Fine, maybe for consistency we could keep the log in both points

controllers/metal3.io/baremetalhost_controller.go
andfasano: I see. This method it's also used in the get one, so probably not the best place for logging. I'd say that the logging in the prepare could be enough for now.

controllers/metal3.io/baremetalhost_controller.go
andfasano: @Hellcatlk in theory the `PreparationError` could be raised also for other reasons than an invalid RAID (ie potentially an invalid BIOS settings). Have you maybe verified functionally this case? If the original error doesn't not depend on the RAID settings I guess that it would be safe to re-apply the same cleaning steps configured for RAID by Ironic, but it'd be nice to confirm it to be on the safe side (instead, if the error was due the invalid RAID, the retry mechanism triggered should give enough room to the user to fix it as required).

pkg/provisioner/ironic/raid.go
Hellcatlk: Done.

fenggw-fnst:
Thanks a lot for your kind review @andfasano . It seems that all the comments have been resolved, may I ask if there are any other problems with this patch?

andfasano:
> Thanks a lot for your kind review @andfasano . It seems that all the comments have been resolved, may I ask if there are any other problems with this patch?

Thanks @fenggw-fnst, I think there was just still one question pending here: https://github.com/metal3-io/baremetal-operator/pull/908/files#r672274334

fenggw-fnst:
Thanks for pointing this out, we will deal with this question ASAP.

controllers/metal3.io/baremetalhost_controller.go
Hellcatlk: @andfasano Sorry I miss this.

> in theory the PreparationError could be raised also for other reasons than an invalid RAID (ie potentially an invalid BIOS settings). Have you maybe verified functionally this case?

Yes, no matter what causes the preparingError, it will reconfigure RAID. 

The following two places will configure the same RAID as `Status.Provisioning.RAID` when the user setting `Spec.RAID` is empty. 

https://github.com/metal3-io/baremetal-operator/blob/672ef32de7a1f719c3afdd38147277c88a3f78d0/controllers/metal3.io/baremetalhost_controller.go#L783-L788

https://github.com/metal3-io/baremetal-operator/blob/672ef32de7a1f719c3afdd38147277c88a3f78d0/controllers/metal3.io/baremetalhost_controller.go#L1135

hase1128:
@andfasano 
>I think there was just still one question pending here: https://github.com/metal3-io/baremetal-operator/pull/908/files#r672274334

We answered.
Please see https://github.com/metal3-io/baremetal-operator/pull/908/files#r673604198

Thank you.

pkg/provisioner/ironic/raid.go
```diff
@@ -12,12 +13,22 @@ import (
...
+const (
+	noRAIDInterface       string = "no-raid"
+	softwareRAIDInterface string = "agent"
+)
+
 // setTargetRAIDCfg set the RAID settings to the ironic Node for RAID configuration steps
-func setTargetRAIDCfg(p *ironicProvisioner, ironicNode *nodes.Node, data provisioner.PrepareData) (err error) {
+func setTargetRAIDCfg(p *ironicProvisioner, raidInterface string, ironicNode *nodes.Node, data provisioner.PrepareData) (err error) {
+	err = checkRAIDConfigure(raidInterface, data.TargetRAIDConfig)
+	if err != nil {
```
andfasano, line 24: @Hellcatlk I've just noticed here a latest moment issue that requires to be fixed (hopefully it's something pretty quick).
In case of failure of `checkRAIDConfigure` (or immediately later `BuildTargetRAIDCfg`) a simple error is returned and then managed as a `transientError` in the `startManualCleaning`.
This does not seem the recommended approach in case of misconfiguration: for example, providing a RAID configuration for an host that does not support RAID will result in the BMH stuck in the `Preparing` state without any error set, but just with a stack trace in the log, which is not optimal for the end user.
So I'd suggest to refactor the `setTargetRAIDCfg` method to return directly a `(res provisioner.Result, err error)`, so that any failure for the `checkRAIDConfigure` (or `BuildTargetRAIDCfg`) could be interpreted as `operationFailed`, whereas an eventual issue from `nodes.SetRAIDConfig` will remain as a `transientError`.
So, in case of wrong configuration settings, yhis will give the user directly enough information to fix the problem (without requiring to check the logs)

andfasano (changes requested):

pkg/provisioner/ironic/raid.go
Hellcatlk: Thanks, done.

andfasano:
Thanks @Hellcatlk for your contribution

/lgtm

fenggw-fnst:
Sorry, we found that there was a case that had been missed in this PR from the beginning, we have submitted #941 to fix that.

dtantsur:
@Hellcatlk someone on Slack reported an issue that may be a regression after this patch, could you take a look? https://kubernetes.slack.com/archives/CHD49TLE7/p1627834617015100

PR #448, opened 2020-03-03
Implement metadata interface for baremetal operator

dukov:
Metadata is required for cloud-init config drive which is used by
Ironic. User should be able to create custom metadata in a secret and
reference it in a BareMetalHost object.

PR #909, opened 2021-06-09
Always build a configdrive except for live ISO

dtantsur:
We have default meta data that we should pass in any case.

PR based on top of https://github.com/metal3-io/baremetal-operator/pull/900.

### DIFF
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -795,7 +795,7 @@ func (p *ironicProvisioner) getImageUpdateOptsForNode(ironicNode *nodes.Node, im
 	if hasCustomDeploy {
 		// Custom deploy process
 		p.setCustomDeployUpdateOptsForNode(ironicNode, imageData, updater)
-	} else if imageData.DiskFormat != nil && *imageData.DiskFormat == "live-iso" {
+	} else if imageData.IsLiveISO() {
 		// Set live-iso format options
 		p.setLiveIsoUpdateOptsForNode(ironicNode, imageData, updater)
 	} else {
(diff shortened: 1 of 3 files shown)


pkg/provisioner/ironic/ironic.go
```diff
@@ -1239,6 +1239,59 @@ func (p *ironicProvisioner) Prepare(data provisioner.PrepareData, unprepared boo
 	return
 }
 
+func (p *ironicProvisioner) getConfigDrive(data provisioner.ProvisionData) (configDrive nodes.ConfigDrive, err error) {
+	// In theory, Ironic can support configdrive with live ISO by attaching
+	// it to another virtual media slot. However, Dell does not support two
+	// virtual media devices at the same time, so we shouldn't try it.
```
hardys, line 1245: From recent discussions with @rhjanders I don't think Supermicro does either, so we should perhaps just say this is not reliably/commonly supported on all hardware types?

hardys:
This lgtm - one small nit re the comment but IMO it makes sense to always pass the metal3 meta-data regardless of the user input in the spec.

I suspect that the reason this ended up originally conditional on the userData is that was at first the only field included in the config-drive, then others got added later (@zaneb do you recall if that's accurate perhaps?)

zaneb:
> I suspect that the reason this ended up originally conditional on the userData is that was at first the only field included in the config-drive, then others got added later (@zaneb do you recall if that's accurate perhaps?)

Yes, I'm 99% sure it was unintentional for that reason. Metadata was added later in #448 and NetworkData in #348.

zaneb:
Refamiliarising myself with the context... we bumped this out of #900 because we realised that always building a config drive could break the live-iso case. With the live-iso case carved out here, it should be safe to always build a config drive, which is probably the way it should always have worked, even though it is a user-visible change of behaviour.
/lgtm

PR #910, opened 2021-06-10
Fixes to RAID configuration

zaneb:
Based on my (shaky) understanding of Ironic, we still have numerous inconsistencies in when we reconfigure RAID. This attempts to iron out some of them.


zaneb:
The next patch I didn't get to yet:
Saving the previous settings means that if a user provisions with RAID and then removes the RAID configuration, we would have enough info to know that the RAID config needs to be cleared even if the user didn't set it to an empty list (re discussion around https://github.com/metal3-io/baremetal-operator/pull/908#issuecomment-857353073).

pkg/provisioner/ironic/raid.go
```diff
@@ -29,6 +29,8 @@ func setTargetRAIDCfg(p *ironicProvisioner, ironicNode *nodes.Node, raid *metal3
 		p.log.Info("rootDeviceHints is used, the first volume of raid will not be set to root")
 	}
```
Hellcatlk, line 30: I think here can be changed to the following:
```go
	// set root volume
	if hasRootDeviceHints  {
		 p.log.Info("rootDeviceHints is used, the first volume of raid will not be set to root")
	}
        logicalDisks[0].IsRootVolume = &hasRootDeviceHints 
```

zaneb:
@Hellcatlk I don't have any more time to spend on this in the near future, but feel free to use these changes for reference or in your own PR. There's a bunch of TODOs in comments and commit messages. Open questions I have:

* Is the assumption true that HW RAID that we set up in the BMC is still valid regardless of what (provisioned) software running on the host has done to it?
* What does the erase_devices_metadata step actually do? When is it required?
* Do we actually need to run a manual cleaning to set up SW RAID, or would the automated cleaning involved in the move from available->manageable->available be sufficient?
* Ironic automatically tears down Software RAID during cleaning... what does that mean for us? Do we have to ensure the current RAID config is present in the Ironic Node after the DB has dropped for this to occur?
* How can we avoid assuming that there is no RAID configured on the host after an error occurs in Preparing? We need to do this without breaking the case where the error is due to trying to set HW RAID on a box that doesn't support it - the user has to be able to recover from that situation. I don't currently believe this can be solved without either adding a HasHardwareRAID flag to the BMH status (which we would set after successfully Preparing a host with HW RAID), or saving a copy of the previous successful settings.

pkg/provisioner/ironic/raid.go
zaneb: We'd have to invert the logic.

pkg/provisioner/ironic/raid.go
```diff
@@ -138,29 +140,56 @@ func buildTargetSoftwareRAIDCfg(volumes []metal3v1alpha1.SoftwareRAIDVolume) (lo
 	return
 }
 
+// hwRaidConfigured returns true if HardwareRAIDVolumes is non-nil.
+// Note that it still returns true if it is an empty list - in this case we
+// still want to clear the RAID settings. When it is nil, we just leave the
+// existing settings in place.
+func hwRaidConfigured(config *metal3v1alpha1.RAIDConfig) bool {
```
Hellcatlk, line 147: Maybe we can check the value of `ironicNode.RAIDConfig` to determine whether RAID has been configured.

zaneb: The ironic database could disappear at any time, so the k8s CR has to be the source of truth.

Hellcatlk:
> Do we actually need to run a manual cleaning to set up SW RAID, or would the automated cleaning involved in the move from available->manageable->available be sufficient?

I checked the clean_steps executed by the auto cleaning, and `create_configuration` does not be executed. 
So I think it is necessary to configure software RAID by manual cleaning.

> What does the erase_devices_metadata step actually do? When is it required?

I found [erase_devices_metadata](https://github.com/openstack/ironic-python-agent/blob/master/ironic_python_agent/hardware.py#L1325-L1364) in `ironic_python_agent`, then it called [destroy_disk_metadata](https://github.com/openstack/ironic-lib/blob/master/ironic_lib/disk_utils.py#L554-L617).
 
In addition, could we merge #908 first? At least it provides a way for users to avoid errors caused by the server not supporting RAID.

pkg/provisioner/ironic/raid.go
Hellcatlk: What do you think of adding a new member to `prepareData` to store the current RAID configuration, and blanking the the new member when clean failed to avoid deleting the saved RAID configuration? See here for [details](https://github.com/metal3-io/baremetal-operator/pull/908/files#diff-a69d9029388ab766ed36b32180145f52785a9d4a153775510dbddfa928a72e1cR759-R771).

hase1128:
@zaneb 
Can we close this PR?
All concerns in this PR are included in #908, and #908 has been merged.
So I think we can close this PR, right?

hase1128:
@zaneb 
PTAL
Can we close this PR?

kashifest:
/hold

kashifest:
/hold cancel

PR #911, opened 2021-06-10
Disable raid feature for iRMC server

Hellcatlk:

In view of the need for further discussion on #908, we can first disable RAID settings for iRMC to avoid clean fail on iRMC servers that do not support hardware RAID.


Hellcatlk:
@andfasano @dtantsur @zaneb

iurygregory (commented):
/lgtm - the reason to change the test to idrac is just because it has support for raid right?

Hellcatlk:
> 
> 
> /lgtm - the reason to change the test to idrac is just because it has support for raid right?

Yes.

PR #912, opened 2021-06-10
Make check for detached annotation more explicit

mkowalski:
Currently checking if the host is detached checks whether the
"baremetalhost.metal3.io/detached" annotation exists, but ignores its
value. This may cause some confusion in case we end up with a BMH
resource with the annotation value set to "false" - from the linguistic
perspective such a host is not detached, however for the current code
such a host would be treated as detached.

This PR makes it explicit so that only when the value of the annotation
is "true", we treat the host as detached.

### DIFF
--- a/controllers/metal3.io/host_state_machine.go
+++ b/controllers/metal3.io/host_state_machine.go
@@ -238,22 +238,22 @@ func (hsm *hostStateMachine) checkInitiateDelete() bool {
 	return true
 }
 
-// hasInspectAnnotation checks for existence of baremetalhost.metal3.io/detached
-func hasDetachedAnnotation(host *metal3v1alpha1.BareMetalHost) bool {
+// hasInspectAnnotation checks for existence and value of baremetalhost.metal3.io/detached
+func hasDetachedAnnotationTrue(host *metal3v1alpha1.BareMetalHost) bool {
 	annotations := host.GetAnnotations()
 	if annotations != nil {
-		if _, ok := annotations[metal3v1alpha1.DetachedAnnotation]; ok {
-			return true
+		if val, ok := annotations[metal3v1alpha1.DetachedAnnotation]; ok {
+			return val == "true"
 		}
 	}
 	return false
 }
 
 func (hsm *hostStateMachine) checkDetachedHost(info *reconcileInfo) (result actionResult) {
-	// If the detached annotation is set we remove any host from the
+	// If the detached annotation is set to "true" we remove any host from the
 	// provisioner and take no further action
 	// Note this doesn't change the current state, only the OperationalStatus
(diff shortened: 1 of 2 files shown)


hardys:
So I deliberately made this ignore the value, because that's what we already do for the [paused annotation](https://github.com/metal3-io/baremetal-operator/blob/master/docs/api.md#pausing-reconciliation) and also the [reboot annotation](https://github.com/metal3-io/baremetal-operator/blob/master/controllers/metal3.io/baremetalhost_controller.go#L397)

The idea IIUC is that these annotations can optionally provide context from whatever external system decided to detach/pause/reboot - so in that case we treat the presence of the annotation key as the trigger, and completely ignore the value.

@n1r1 @zaneb what are your thoughts?

mkowalski:
Do you have any insights on other projects using it like this? So far (but that's not a big sample) I've only seen it being set to `true` and this is what caused my confusion, because having `baremetalhost.metal3.io/detached: "false"` (what would be a logical option if there was `true` before) does the opposite of what an operator of the environment would expect.

mkowalski:
What is also an option is to have a check for value other than `false`, that could be a good trade-off between projects using it as string and projects using it as a boolean value. What do you think?

n1r1:
it's true that the reboot annotation value was ignored in the past, but now we use the value to determine the "reboot mode" which can be either soft or hard (see [this](https://github.com/metal3-io/baremetal-operator/blob/f5228fa38953b0e0e3f4763797013a3f099151dc/controllers/metal3.io/baremetalhost_controller.go#L416) )

as for this specific case, I don't have any strong preference.
I'm not a big fan of annotations for such cases, but I'm inclined not to parse the value unless really required.

hardys:
> What is also an option is to have a check for value other than `false`, that could be a good trade-off between projects using it as string and projects using it as a boolean value. What do you think?

IMO we should just document that this API (and other related ones I mentioned) ignore the value, because we don't have conventions of accepting true/false values from any other annotations AFAIK?  For example there is the [inspect annotation](https://github.com/metal3-io/baremetal-operator/blob/master/docs/inspectAnnotation.md#inspect-annotation) which accepts a value of `disabled` not `false` - this is clearly documented so it's probably fine, but it'd be weird to retrospectively enforce some convention on previously ignored values?

There are several other yaml inputs which could be considered "false", so my opinion is it'd be clearer and less confusing to just say the value is ignored, and if the annotation is found (with any value) then the host is detached.

mkowalski:
Ok, I agree with all the points above. I think it's much better if in the projects itself we _discourage_ setting the example value as `true` so that people do not get the natural feeling that it's the boolean value.

I'll close this one here and let's get only some doc improvements via https://github.com/metal3-io/baremetal-operator/pull/913.

PR #913, opened 2021-06-10
Extend docs for "baremetalhost.metal3.io/detached" annotation

mkowalski:
Currently it's not clear that the `baremetalhost.metal3.io/detached`
annotation ignores the value. This PR makes it clear in the API
documentation that the annotation is not a boolean value so that setting
it as `"false"` may not have an expected result.

Supersedes: #912


hardys:
/approve

Thanks @mkowalski!

PR #916, opened 2021-06-15
Don't deprovision provisioned host due to error

### LINKED ISSUE #915: Adoption errors cause deprovisioning
If a Host is in the `provisioned` state and an error occurs during Adopt() (i.e. a `provisioned registration error`), the Host will be deprovisioned. This is bad.

The same does _not_ apply to ordinary `registration error` errors, because registration is handled separately.

It _would_ apply to `power management error`, but we have never reported a power management error (c.f. #828), and won't until #841 is merged.

zaneb:
During provisioning, we want an error to cause the Host to be
deprovisioned again, but we very much do not want that to happen when
the Host is already provisioned. Due to an error in
0a7713bd3b4ff30aef2dac17c2d62a79bfe02872, provisioned Hosts were being
automatically deprovisioned whenever an error (other than a registration
error) occurred.

In practice this only manifested on Adopt() failures, since we don't yet
report power management failures.

Fixes #915

### DIFF
--- a/controllers/metal3.io/host_state_machine.go
+++ b/controllers/metal3.io/host_state_machine.go
@@ -428,9 +428,6 @@ func (hsm *hostStateMachine) handleReady(info *reconcileInfo) actionResult {
 }
 
 func (hsm *hostStateMachine) provisioningCancelled() bool {
-	if hsm.Host.Status.ErrorMessage != "" {
-		return true
-	}
 	if hsm.Host.Spec.Image == nil {
 		return true
 	}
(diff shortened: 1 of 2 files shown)


hardys:
Adding a test to cover this would be good, but otherwise lgtm, thanks!

zaneb:
Added a test.
/test-integration

controllers/metal3.io/host_state_machine.go
```diff
@@ -428,9 +428,6 @@ func (hsm *hostStateMachine) handleReady(info *reconcileInfo) actionResult {
 }
 
 func (hsm *hostStateMachine) provisioningCancelled() bool {
-	if hsm.Host.Status.ErrorMessage != "" {
```
sadasu, line 431: Do we need this condition to remain, for a different path? is there a way for us to check if we are in the Adopt() path?

PR #918, opened 2021-06-17
Use recreate strategy for ironic deployment

lentzi90:
Ironic cannot run with more than 1 replica (the second crashes immediately) but the default rolling update strategy will try to get a second replica up and running before it removes the old one. This means that the rollout always gets stuck unless we first manually scale down the deployment to 0, then update, and finally scale it up again.

The recreate strategy is a better fit since it does the scale down
automatically.

For reference, here is how we currently upgrade ironic in the upgrade tests:
https://github.com/Nordix/metal3-dev-env/blob/1ab2dda79c717083f8ced21cdaeb885c93559dc6/vm-setup/roles/v1aX_integration_test/tasks/upgrade.yml#L217-L257

### DIFF
--- a/ironic-deployment/ironic/ironic.yaml
+++ b/ironic-deployment/ironic/ironic.yaml
@@ -4,6 +4,10 @@ metadata:
   name: capm3-ironic
 spec:
   replicas: 1
+  strategy:
+    # We cannot run Ironic with more than one replica at a time. The recreate
+    # strategy makes sure that the old pod is gone before a new is started.
+    type: Recreate
   selector:
     matchLabels:
       name: capm3-ironic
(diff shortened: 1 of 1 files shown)


lentzi90:
Waiting for https://github.com/metal3-io/baremetal-operator/pull/918 to simplify the upgrade steps for ironic. Once that PR is merged, we can skip the scale down and scale up since it will be done automatically.

/hold

lentzi90:
/hold cancel
(Accidentally added it on the wrong PR...)

namnx228 (approved):
LGTM, this is a good improvement

smoshiur1237:
Looks good to me.

macaptain (approved):

ironic-deployment/ironic/ironic.yaml
```diff
@@ -4,6 +4,10 @@ metadata:
   name: capm3-ironic
 spec:
   replicas: 1
+  strategy:
+    # Ironic cannot run with more than one replica at a time, so prevent this
```
dtantsur, line 8: Let's please reword this, ironic can run with a lot of replicates, *we* cannot run it this way.

lentzi90: Right, will do!

lentzi90: Better?

smoshiur1237:
PR is ready to get approved and lgtm.

PR #923, opened 2021-06-21
Fix the bug when not using reverse proxy for Ironic-inspector

namnx228:
This PR adds an option to disable the ironic inspector reverse proxy when we enable TLS on Ironic communication. 
Edit: Instead of adding an option, this PR just fixes a bug in the non-proxy case.

### DIFF
--- a/ironic-deployment/tls/default/tls.yaml
+++ b/ironic-deployment/tls/default/tls.yaml
@@ -53,6 +53,9 @@ spec:
         - name: cert-mariadb-ca
           mountPath: "/certs/ca/mariadb"
           readOnly: true
+        - name: cert-ironic-inspector-ca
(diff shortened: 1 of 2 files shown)


namnx228:
/assign @dtantsur 
/cc @fmuyassarov 
Related to: https://github.com/metal3-io/ironic-image/pull/277

dtantsur:
I understand it's a valid case that somebody may have, but do we really need to support it explicitly? I'd rather limit the number of options we expose.

namnx228:
> I understand it's a valid case that somebody may have, but do we really need to support it explicitly? I'd rather limit the number of options we expose.

Fair enough. I will drop the change that enables the option. However, a part of this PR is needed to solve a bug in non-proxy case. I think we should fix the bug even when we don't explicitly support it.

PR #928, opened 2021-06-30
BIOS Configurations for iDRAC

ZainubW:
Extended BIOS Configuration for iDRAC.


ZainubW:
@zaneb Can you please approve it again?

dtantsur:
/approve

Is there any documentation where we list which drivers support BIOS/RAID?

ZainubW:
> /approve
> 
> Is there any documentation where we list which drivers support BIOS/RAID?

I don't think so there is any documentation yet. But we can add it in api.md?

PR #929, opened 2021-07-01
Fix auto clean failure in preparing state

Hellcatlk:
Auto cleaning step(erase_devices_metadata) in Prepare will fail with the following error:

```
Agent returned error for clean step {'step': 'erase_devices_metadata',
'priority': 10, 'interface': 'deploy', 'reboot_requested': False, 'abortable':
True, 'requires_ramdisk': True} on node 09174110-95b3-4bd0-99a9-839df8794139 :
Error performing clean_step erase_devices_metadata: Error erasing block device:
Failed to erase the metadata on the device(s): \"/dev/sda\": Unexpected error
while running command.\nCommand: dd bs=512 if=/dev/zero of=/dev/sda count=33\n
Exit code: 1\nStdout: ''\nStderr: \"dd: error writing '/dev/sda': Input/output
error\\n1+0 records in\\n0+0 records out\\n0 bytes copied, 0.00108204 s, 0.0 kB/s\\n\"
```

Since RAID cannot be used immediately after the configuration is completed, it takes a
certain amount of time to initialize. At this time, data cannot be written to the disk,
and the above error occurs.

This will result in entering the clean fail state and re-executing manual clean and
auto clean, then it causes the above error again.

### DIFF
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -1217,10 +1217,7 @@ func (p *ironicProvisioner) Prepare(data provisioner.PrepareData, unprepared boo
 			started = true
 		}
 		// Manual clean finished
-		result, err = p.changeNodeProvisionState(
-			ironicNode,
-			nodes.ProvisionStateOpts{Target: nodes.TargetProvide},
-		)
+		result, err = operationComplete()
 
 	case nodes.CleanFail:
 		// When clean failed, we need to clean host provisioning settings.


pkg/provisioner/ironic/ironic.go
```diff
@@ -1217,10 +1217,7 @@ func (p *ironicProvisioner) Prepare(data provisioner.PrepareData, unprepared boo
 			started = true
 		}
 		// Manual clean finished
-		result, err = p.changeNodeProvisionState(
-			ironicNode,
-			nodes.ProvisionStateOpts{Target: nodes.TargetProvide},
-		)
+		result, err = operationComplete()
```
hardys, line 1220: If we remove this change of the provision state, how will the ironic host make the manageable->cleaning->available transition?

Ref https://docs.openstack.org/ironic/latest/_images/states.svg

hardys:
Is this error specific to RAID and/or particular hardware?

We seem to be changing the underlying state transition behavior here, so it'd be good to clarify how (when the clean error eventually is resolved, e.g the RAID is ready etc) we'll make the manageble->cleaning->available transition since it's not immediately obvious from this patch?

/cc @dtantsur

Hellcatlk:
@hardys [Here](https://github.com/metal3-io/baremetal-operator/blob/master/pkg/provisioner/ironic/ironic.go#L1184-L1186) will make the manageble->cleaning->available transition. 
If there is a clean failed here, it will only execute the auto cleaning agauin and keep looping until RAID is available.

Hellcatlk:
Sorry for making you confused, let me clarify the problem and what this PR for.

> Is this error specific to RAID and/or particular hardware?

This error is specific to RAID, rather than to particular hardware.

The root cause is that doing auto clean after RAID configuration is just completed causes a write error, this error results in entering the `clean fail` state, which causes to redo of manual clean and RAID configuration then auto clean will fail again, leading to an Infinite loop.

The write error is inevitable no matter initMode is slow or fast unless setting initMode=no, which means "do not check disk", but this is obviously wrong, and changing initMode is not supported by Ironic either.

The solution of this PR is to drop the auto clean from RAID configuration phase. In the original design, there was no auto clean in RAID configuration, it was added as an improvement, and it is safe to be removed I think.

> If we remove this change of the provision state, how will the ironic host make the manageable->cleaning->available transition?

This PR removes auto cleaning from the `preparing` state not `provisioning` state, so auto clean will still be executed in [`provision`](https://github.com/metal3-io/baremetal-operator/blob/493c56fc9b120ce4c58ac25ef8eb63d276e54528/pkg/provisioner/ironic/ironic.go#L1356-L1358).

dtantsur:
Shouldn't we fix the Ironic RAID implementation instead?

Hellcatlk:
@dtantsur 

> Shouldn't we fix the Ironic RAID implementation instead?

Thanks for the suggestion, this is indeed more reasonable. Could you please add the implementation in Ironic? Or should I fix this?

hs0210:
@dtantsur 
> Shouldn't we fix the Ironic RAID implementation instead?

After consideration, we came up with two ideas to fix it on ironic side：
1. Set a time limit for RAID initialization.   
Constrains: The time required for RAID initialization depends on the performance and size of the disk. So it is difficult to calculate an appropriate time limit.
2. Get the progress of RAID initialization to determine whether it has ended.   
Constrains: This requires each bmc to have related APIs to obtain the RAID initialization progress.

Both the two ways have their constrains, but at present we only think of them to solve this problem on ironic side.
Do you have a better idea? If it's really difficult to fix this on ironic side, could we change the phase of automated cleaning like [#929](https://github.com/metal3-io/baremetal-operator/pull/929) to fix it?

dtantsur:
I'm struggling to understand 1) how this change fixes anything (see Steve's question), 2) why at least the same logic cannot be applied to Ironic.

Hellcatlk:
@dtantsur 
> 1) how this change fixes anything (see Steve's question)

The root cause of this error is the inability to distinguish the cause of clean failed in Preparing. Therefore, the failure of `auto clean` will also cause `manual clean` to be executed again. You can see [here](https://github.com/metal3-io/baremetal-operator/pull/929#issuecomment-882267676) for details.

This PR moves `auto clean` to the `Provisioning` stage. Since the `Provisioning` stage only has `auto clean`, even if `clean failed`, it will only re-execute `auto clean` .

fenggw-fnst:
@dtantsur 

After RAID configuration completed, it takes some time to get RAID really accessible, due to different initialization ways or the logical disk size generated, the time could be long or short, but this is a common issue with all RAID drivers rather than specific to some vendor's hardware.

Regarding solving this on Ironic side, the only way we can come up with is to add an abstract method (e.g. wait_raid_init) within class RAIDInterface, which is to wait until RAID really accessible, vendors need to implement this by their own. And we also need to modify BMO on Metal3 side to append wait_raid_init as a clean step to where we do RAID configuration.

This requires every vendor's implementation, which takes a lot of time. Besides, it still needs to modify BMO on Metal3 side. As a result, I wonder if we could solve this just by current PR?

dtantsur:
> After RAID configuration completed, it takes some time to get RAID really accessible, due to different initialization ways or the logical disk size generated, the time could be long or short, but this is a common issue with all RAID drivers rather than specific to some vendor's hardware.

This is a different problem from the one mentioned above. This has to be fixed or worked around on the Ironic side.

> Regarding solving this on Ironic side, the only way we can come up with is to add an abstract method (e.g. wait_raid_init) within class RAIDInterface, which is to wait until RAID really accessible, vendors need to implement this by their own. And we also need to modify BMO on Metal3 side to append wait_raid_init as a clean step to where we do RAID configuration.

Simpler, just don't return from create_configuration etc until RAID is actually ready.

fenggw-fnst:
> Simpler, just don't return from create_configuration etc until RAID is actually ready.

To achieve this, we can only think of 2 ways:
1. Continuously retry write operation to check if RAID is actually ready with setting a time limit.
2. Implement "wait" by either add a new method or implement [this synchronize case](https://github.com/openstack/ironic/blob/master/ironic/drivers/base.py#L1395-L1397), which both need vendors to implement their apis to provide the ways for getting RAID initialization progress.

But they both have problems as mentioned [here](https://github.com/metal3-io/baremetal-operator/pull/929#issuecomment-883056454).
Do you have any other good ideas?

> This is a different problem from the one mentioned above. This has to be fixed or worked around on the Ironic side.

Yes, I agree that this problem should be solved on the Ironic side as a fundamental solution, but this may take a long time.
So can we solve [this problem from Metal3 perspective](https://github.com/metal3-io/baremetal-operator/pull/929#issuecomment-883849354) by PR929 first?

dtantsur:
So, we have two problems that are being discussed by two different people:

> The root cause of this error is that manual clean and auto clean are executed in the Preparing stage, but their failure statuses are both clean failed.

This is fixed by this patch, and BMO is the right place.

> After RAID configuration completed, it takes some time to get RAID really accessible

I'm not sure how this patch fixes the 2nd problem. Does it just happen to work because of a small delay? It's equivalent of putting `sleep(5)` in Ironic pretty much, which is not a great fix. But if we have no other options, it's better to do it in Ironic since it's going to benefit all consumers, not only BMO.

> Do you have any other good ideas?

This is what Redfish does: https://opendev.org/openstack/ironic/src/branch/master/ironic/drivers/modules/redfish/raid.py#L849-L858. Essentially if reboot is requested, it waits for IPA to start after the reboot, effectively knowing when it's done.

fenggw-fnst:
>So, we have two problems that are being discussed by two different people

Yes, sorry for making you confused.

>I'm not sure how this patch fixes the 2nd problem. Does it just happen to work because of a small delay? It's equivalent of putting sleep(5) in Ironic pretty much, which is not a great fix. But if we have no other options, it's better to do it in Ironic since it's going to benefit all consumers, not only BMO.

This patch doesn't fix the 2nd problem, it just works around the impact of the 2nd problem in BMO (i.e. the 1st problem). By applying this patch, although the write errors may still occur due to the 2nd problem, as long as auto clean is not at the same stage as the RAID configuration, it will just retry until it succeeds. Therefore, the manual clean process (RAID configuration) and auto clean process can work just fine respectively.

We still need to fix the 2nd problem on the Ironic side, but at first we would like to make sure RAID feature available in BMO.

>This is what Redfish does

Thanks for sharing, we will refer to this to fix the 2nd problem.

pkg/provisioner/ironic/ironic.go
dtantsur: Could someone please answer this question?

hs0210: As @Hellcatlk has already replied [here](https://github.com/metal3-io/baremetal-operator/pull/929#issuecomment-882267676). If we remove the changes here, then the auto cleaning will be executed in [`provision`](https://github.com/metal3-io/baremetal-operator/blob/493c56fc9b120ce4c58ac25ef8eb63d276e54528/pkg/provisioner/ironic/ironic.go#L1356-L1358), so the ironic host can still make the manageable->cleaning->available transition.

dtantsur:
/lgtm

Looks good, but I'd prefer @hardys or @zaneb to take a deeper look and approve.

zaneb:
So this is effectively reverting #882 :(

It should work fine because we still have the code to handle the manageable->available transition in the Provisioning state, but we lose the benefits (faster provisioning, consistent ironic state when Host in the ready state) and the ability to build on it in future (I'd like to move directly from Deprovisioning->Preparing as soon as automated cleaning is started, to make fencing faster).

I agree that we need to distinguish between failures of manual and automated cleaning. Unfortunately I can't think of any obvious solutions other that to introduce yet another state, or sub-states of the kind proposed by @dtantsur in https://github.com/metal3-io/metal3-docs/blob/master/design/baremetal-operator/sub-states.md#baremetalhost-detailed-activity.

Hellcatlk:
> So this is effectively reverting #882 :(
> 
> It should work fine because we still have the code to handle the manageable->available transition in the Provisioning state, but we lose the benefits (faster provisioning, consistent ironic state when Host in the ready state) and the ability to build on it in future (I'd like to move directly from Deprovisioning->Preparing as soon as automated cleaning is started, to make fencing faster).
> 
> I agree that we need to distinguish between failures of manual and automated cleaning. Unfortunately I can't think of any obvious solutions other that to introduce yet another state, or sub-states of the kind proposed by @dtantsur in https://github.com/metal3-io/metal3-docs/blob/master/design/baremetal-operator/sub-states.md#baremetalhost-detailed-activity.

@zaneb #882 is a good improvement, and I believe that eventually we need it. But can we merge this PR (revert #882 temporarily) to make sure RAID feature available first? IMHO temporarily losing the benefits of #882 doesn't impact too much compared with before. We can proceed the solution to distinguish between failures of manual and automated cleaning later, and apply #882 in future.

zaneb:
@dtantsur mentioned [on Slack](https://kubernetes.slack.com/archives/CHD49TLE7/p1629708966125600?thread_ts=1629471651.115400&cid=CHD49TLE7) that we should still be able to the target state (`manageable` or `available`) after the current state reached `clean failed`.

hs0210:
> @dtantsur mentioned on Slack that we should still be able to see the target state (manageable or available) after the current state reached clean failed.

@zaneb @dtantsur Yes, it is true the target state can be used to determine which caused clean failed(manual clean or auto clean). And we have verified it with the following patch. When auto clean fails, it only re-executes auto clean.

```diff
diff --git a/pkg/provisioner/ironic/ironic.go b/pkg/provisioner/ironic/ironic.go
index 282363e3..8549e8ae 100644
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -1105,6 +1105,20 @@ func (p *ironicProvisioner) Prepare(data provisioner.PrepareData, unprepared boo
                        result, err = p.setMaintenanceFlag(ironicNode, false)
                        return
                }
+               if ironicNode.TargetProvisionState == string(nodes.Available) {
+                       result, err = p.changeNodeProvisionState(
+                               ironicNode,
+                               nodes.ProvisionStateOpts{Target: nodes.TargetManage},
+                       )
+                       if err != nil {
+                               return
+                       }
+                       result, err = p.changeNodeProvisionState(
+                               ironicNode,
+                               nodes.ProvisionStateOpts{Target: nodes.TargetProvide},
+                       )
+                       return
+               }
                result, err = p.changeNodeProvisionState(
                        ironicNode,
                        nodes.ProvisionStateOpts{Target: nodes.TargetManage},
```
But it still failed to prepare because when executing auto clean the following new error occurs and then it falls into an infinite loop. Very sorry that I couldn't figure out what cause this error, could you give me some advice?
```
ERROR ironic.conductor.utils [req-4644db12-5acc-4f2c-94b8-579b5331188e - - - - -] 
Unexpected error when processing next clean step. TypeError: 'NoneType' object is not subscriptable
```

dtantsur:
> Unexpected error when processing next clean step. TypeError: 'NoneType' object is not subscriptable

It's likely a bug in ironic. Could you fetch the whole traceback from the ironic-conductor logs?

dtantsur:
Also note that the `manage` action is fast, but not immediate. Your proposed snippet doesn't account for it.

PR #933, opened 2021-07-07
Do not disable iDRAC Redfish vendor interfaces

dtantsur:
There is no harm in having them enabled, they're not used implicitly.
In the future we may start using them, e.g. to eject virtual CD.

Requires https://github.com/metal3-io/ironic-image/pull/284.


elfosardo (approved):

dtantsur:
/test-integration

Docker rate-limit

PR #934, opened 2021-07-09
Modify namespace prefix in ironic deployment to support TLS

smoshiur1237:
The changes in this PR will help to support TLS with CAPM3 v1a4(BMO part of CAPM3) and v1a5(BMO outside of CAPM3). This changes is needed to cope with v1a4 and v1a5 separately.

### DIFF
--- a/ironic-deployment/basic-auth/default/auth.yaml
+++ b/ironic-deployment/basic-auth/default/auth.yaml
@@ -1,7 +1,7 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: capm3-ironic
+  name: ${NAMEPREFIX}-ironic
 spec:
   template:
     spec:
--- a/tools/deploy.sh
+++ b/tools/deploy.sh
@@ -25,6 +25,32 @@ RESTART_CONTAINER_CERTIFICATE_UPDATED=${RESTART_CONTAINER_CERTIFICATE_UPDATED:-"
 
 SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
 
+IRONIC_DEPLOY_FILES="${SCRIPTDIR}/ironic-deployment/basic-auth/default/auth.yaml \
+	${SCRIPTDIR}/ironic-deployment/basic-auth/default/kustomization.yaml \
+	${SCRIPTDIR}/ironic-deployment/basic-auth/keepalived/auth.yaml \
+	${SCRIPTDIR}/ironic-deployment/basic-auth/keepalived/kustomization.yaml \
+	${SCRIPTDIR}/ironic-deployment/basic-auth/tls/default/auth.yaml \
+	${SCRIPTDIR}/ironic-deployment/basic-auth/tls/default/kustomization.yaml \
+	${SCRIPTDIR}/ironic-deployment/basic-auth/tls/keepalived/auth.yaml \
+	${SCRIPTDIR}/ironic-deployment/basic-auth/tls/keepalived/kustomization.yaml \
+	${SCRIPTDIR}/ironic-deployment/certmanager/certificate.yaml \
+	${SCRIPTDIR}/ironic-deployment/default/kustomization.yaml \
+	${SCRIPTDIR}/ironic-deployment/ironic/ironic.yaml \
+	${SCRIPTDIR}/ironic-deployment/keepalived/keepalived_patch.yaml \
+	${SCRIPTDIR}/ironic-deployment/keepalived/kustomization.yaml \
+	${SCRIPTDIR}/ironic-deployment/tls/default/kustomization.yaml \
+	${SCRIPTDIR}/ironic-deployment/tls/default/tls.yaml \
+	${SCRIPTDIR}/ironic-deployment/tls/keepalived/kustomization.yaml \
+	${SCRIPTDIR}/ironic-deployment/tls/keepalived/tls.yaml"
+
+for DEPLOY_FILE in ${IRONIC_DEPLOY_FILES}; do
+  cp "$DEPLOY_FILE" "$DEPLOY_FILE".orig
+  # shellcheck disable=SC2094
+  envsubst <"$DEPLOY_FILE".orig> "$DEPLOY_FILE"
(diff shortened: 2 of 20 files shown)


ironic-deployment/basic-auth/default/auth.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: capm3-ironic
+  name: ${NAMEPREFIX}-ironic
```
lentzi90, line 4: Do we have to make this configurable? I would much prefer to just skip the prefix. We don't have the prefix on the ConfigMaps and Secrets either.

If we absolutely need a prefix, I would suggest using the built-in `namePrefix` in [kustomize](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/kustomization/#setting-cross-cutting-fields). That way we don't need the custom bash/sed logic and the manifests stay clean.

```suggestion
  name: ironic
```

Same for all the other `capm3` -> `${NAMEPREFIX}` changes.

namnx228: Yes, we need to have the name prefix configurable. The point is that Ironic needs to be in the same namespace as BMO. With CAPM3 v1a4, we have BMO deployed in CAPM3, so it is in `capm3-system` namespace. With v1a5, we plan to deploy BMO outside of CAPM3, so it is in `baremetal-operator-system` namespace. Therefore, we need to deploy Ironic in different namespace accordingly. However, the Ironic deployment is independent with BMO deployment. Therefore, `namePrefix` in kustomize doesn't help in this case. The only way is to use bash commands to check which CAPM3 we are using, and change the nameprefix accordingly.

furkatgofurov7:
/retitle Modify namespace prefix in ironic deployment to support TLS

tools/run_local_ironic.sh
```diff
@@ -93,7 +93,7 @@ data:
 kind: Secret
 metadata:
    name: ironic-cacert
-   namespace: capm3-system
+   namespace: ${IRONIC_NAMESPACE}
```
dtantsur, line 96: What a different approach here?

jan-est: If I am not mistaking ${IRONIC_NAMESPACE} was used during pivoting to define in which namespace we want to move Ironic.

furkatgofurov7: This PR is dependent on two other patches, one in m3-dev-env and other in CAPM3. 

1. First one, [here](https://github.com/metal3-io/metal3-dev-env/pull/721/files#diff-cfcb1fb44601d83d47282622bdbd46c08d3278b4a6f6a0da4b2903ba609885dfR176-R180) sets the namespace for the ironic depending on the API version of CAPM3, where in CAPM3 v1a4 we deploy BMO as part of it and namespace is `capm3-system`, whereas in CAPM3 v1a5 we are trying to decouple BMO from CAPM3, the namespace is `baremetal-operator-system`. So, depending on the which ever CAPM3 API version is used we set namespace accordingly. 
2. Second one, is an effort to decouple BMO from CAPM3 and you can check the PR [here](https://github.com/metal3-io/cluster-api-provider-metal3/pull/242), which takes BMO out from CAPM3 in CAPM3 v1a5.

m3-dev-env [PR721](https://github.com/metal3-io/metal3-dev-env/pull/721) is testing both changes from this PR and CAPM3 one in CI.

tools/deploy.sh
```diff
@@ -22,9 +22,36 @@ MARIADB_HOST_IP="${MARIADB_HOST_IP:-"127.0.0.1"}"
...
+	${SCRIPTDIR}/ironic-deployment/tls/default/tls.yaml \
+	${SCRIPTDIR}/ironic-deployment/tls/keepalived/kustomization.yaml \
+	${SCRIPTDIR}/ironic-deployment/tls/keepalived/tls.yaml"
+
+for DEPLOY_FILE in ${IRONIC_DEPLOY_FILES}; do
+  cp "$DEPLOY_FILE" "$DEPLOY_FILE".bak
+  # shellcheck disable=SC2094
+  envsubst <"$DEPLOY_FILE".bak> "$DEPLOY_FILE"
+  # To use sed need to change ${NAMEPREFIX} to NAMEPREFIX to every IRONIC_DEPLOY_FILES
+  #sed -i "s/NAMEPREFIX/${NAMEPREFIX}/g" "${DEPLOY_FILE}"
```
furkatgofurov7, line 52: are these redundant, if so we should remove them.

ironic-deployment/basic-auth/default/auth.yaml
lentzi90: Thanks for explaining! Ok so the namespace must be different, I do get that. But the Deployment name can always be the same, right? That would make this change smaller and simpler I think.

tools/run_local_ironic.sh
namnx228: @dtantsur using `${IRONIC_NAMESPACE}` and `${NAMEPREFIX}-system` are similar. However, I agree that we should make it synchronize with the others. Will fix it. Thanks

tools/deploy.sh
namnx228: Fixed! Thanks

ironic-deployment/basic-auth/default/auth.yaml
namnx228: Yes, technically the name doesn't need to change. However, in v1a4, if we want to deploy Ironic in `capm3-system` namespace,  we should add the `capm3-` prefix since the others have the same prefix (we use namePrefix in CAPM3 kustomize). The same thing will be applied in v1a5 when BMO and Ironic are deployed in `baremetal-operator-system` namespace. That is why we also need to change the name for the sake of synchronization.

tools/run_local_ironic.sh
dtantsur: Okay, understood. I assume this patch is not ready to merge yet, because the metal3-dev-env change that defined IRONIC_NAMESPACE hasn't merged?

furkatgofurov7: @dtantsur this can already go in, IRONIC_NAMESPACE is defined already and set to `capm3-system` as default in metal3-dev-env master, it will be set in 721 accordingly based on CAPM3 API version.

fmuyassarov (commented):
/lgtm

PR #935, opened 2021-07-09
Create a SecretManager type to manage Secrets

zaneb:
De-duplicate the labelling implementations for BMC creds and user data, into a single SecretManager type in a new secretutils package. Also adds the flexibility to add owner references that are not controller references. As much code as possible related to caching/watching of Secrets is moved into this new package.

### DIFF
--- a/controllers/metal3.io/baremetalhost_controller_test.go
+++ b/controllers/metal3.io/baremetalhost_controller_test.go
@@ -705,7 +705,7 @@ func TestSecretUpdateOwnerRefAndEnvironmentLabelOnStartup(t *testing.T) {
 	assert.True(t, *secret.OwnerReferences[0].Controller)
 	assert.True(t, *secret.OwnerReferences[0].BlockOwnerDeletion)
 
-	assert.Equal(t, LabelEnvironmentValue, secret.Labels[LabelEnvironmentName])
+	assert.Equal(t, "baremetal", secret.Labels["environment.metal3.io"])
 
 }
 
--- a/pkg/secretutils/label.go
+++ b/pkg/secretutils/label.go
@@ -1,0 +1,6 @@
+package secretutils
+
+const (
+	LabelEnvironmentName  = "environment.metal3.io"
+	LabelEnvironmentValue = "baremetal"
+)
--- a/pkg/secretutils/label.go
+++ b/pkg/secretutils/label.go
@@ -1,6 +1,32 @@
 package secretutils
 
+import (
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+)
+
 const (
 	LabelEnvironmentName  = "environment.metal3.io"
 	LabelEnvironmentValue = "baremetal"
 )
+
+// AddSecretSelector adds a selector to a cache.SelectorsByObject that filters
+// Secrets so that only those labelled as part of the baremetal environment get
+// cached. The input may be nil.
+func AddSecretSelector(selectors cache.SelectorsByObject) cache.SelectorsByObject {
+	secret := &corev1.Secret{}
+	newSelectors := cache.SelectorsByObject{
+		secret: {
+			Label: labels.SelectorFromSet(
+				labels.Set{
+					LabelEnvironmentName: LabelEnvironmentValue,
+				}),
+		},
+	}
+	if selectors == nil {
+		return newSelectors
+	}
+	selectors[secret] = newSelectors[secret]
+	return selectors
+}
--- a/pkg/secretutils/secret_manager.go
+++ b/pkg/secretutils/secret_manager.go
@@ -1,0 +1,140 @@
+package secretutils
+
+import (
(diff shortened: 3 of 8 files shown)


zaneb:
@andfasano I wasn't planning on doing this, but I realised that I needed a third implementation of applying the label and setting the owner reference (not controller reference), and also to re-use the same code in a separate controller, and possibly also in a separate project. So at that point I figured it made more sense to refactor into a new package.

pkg/secretutils/secret_manager.go
```diff
@@ -0,0 +1,118 @@
...
+	needsUpdate := false
+	if !metav1.HasLabel(secret.ObjectMeta, LabelEnvironmentName) {
+		log.Info("settting secret environment label")
+		metav1.SetMetaDataLabel(&secret.ObjectMeta, LabelEnvironmentName, LabelEnvironmentValue)
+		needsUpdate = true
+	}
+	if owner != nil {
+		if ownerIsController {
+			if !metav1.IsControlledBy(secret, owner) {
+				log.Info("setting secret controller reference")
```
andfasano, line 72: It could be useful to log also the `owner`

pkg/secretutils/secret_manager.go
```diff
@@ -0,0 +1,118 @@
...
+			alreadyOwned := false
+			ownerUID := owner.GetUID()
+			for _, ref := range secret.GetOwnerReferences() {
+				if ref.UID == ownerUID {
+					alreadyOwned = true
+					break
+				}
+			}
+			if !alreadyOwned {
+				log.Info("setting secret owner reference")
```
andfasano, line 88: Here as well

pkg/secretutils/secret_manager.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	return nil
+}
+
+// ObtainSecret retrieves a Secret and ensures that it has a label that will
+// ensure it is present in the cache (and that we can watch for changes), and
+// optionally that it has a particular owner reference. The owner reference may
+// optionally be a controller reference.
+func (sm *SecretManager) ObtainSecret(key types.NamespacedName, owner metav1.Object, ownerIsController bool) (*corev1.Secret, error) {
```
andfasano, line 110: name bikeshedding: the feeling is that `obtain` doesn't communicate properly the fact that not only the object is going to be retrieved, but it's also to be owned if it's required. Maybe something like `AcquireSecret` could sound a little bit better.

andfasano: On a second thought, I realized that even `acquire` may not work well when `owner` is nil, because in that case no ownership action will be issued. For that, I'd guess if it could be better to split this method in a `ObtainSecret(key)` (just returns the secret and apply the environment label if required) and a `AcquireSecret(key, owner, ownerIsController)` that also sets the owner

controllers/metal3.io/baremetalhost_controller_test.go
```diff
@@ -705,7 +705,7 @@ func TestSecretUpdateOwnerRefAndEnvironmentLabelOnStartup(t *testing.T) {
 	assert.True(t, *secret.OwnerReferences[0].Controller)
 	assert.True(t, *secret.OwnerReferences[0].BlockOwnerDeletion)
 
-	assert.Equal(t, LabelEnvironmentValue, secret.Labels[LabelEnvironmentName])
+	assert.Equal(t, "baremetal", secret.Labels["environment.metal3.io"])
```
andfasano, line 708: Any particular reason to not keep using the const here?

pkg/secretutils/label.go
```diff
@@ -0,0 +1,30 @@
...
+
+// CacheSecretSelectors returns a SelectorsByObject that filters Secrets so
+// that only those labelled as part of the baremetal environment get cached.
+func CacheSecretSelectors() cache.SelectorsByObject {
+	return cache.SelectorsByObject{
+		&corev1.Secret{}: {
+			Label: CacheLabelSelector(),
+		},
+	}
+}
```
andfasano, line 30: This approach doesn't look to scale well in case in future there could be the need to specify an additional cache selector for another kind of resource different from a secret (and maybe managed by another manager). One approach could be to transform this method so that it could receive a `cache.SelectorsByObject` and fill it for just the secret type, leaving the possibility to other contributors to set eventually other types.

pkg/secretutils/secret_manager.go
```diff
@@ -0,0 +1,140 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+// SecretManager is a type for fetching Secrets whether or not they are in the
+// client cache, labelling so that they will be included in the client cache,
+// and optionally setting an owner reference.
+type SecretManager struct {
+	log       logr.Logger
+	client    client.Client
+	apiReader client.Reader
```
ardaguclu, line 23: SecretManager seems pretty well written. It would be better to add some unit tests by mocking these fields?

ardaguclu: Maybe it is not easy. There are lots of k8s dependencies.

Issue #855, opened 2021-04-13
Validating Webhook for BareMetalHost


zaneb:
There's a number of values or changes in a BMH Spec that we know are invalid, but that we don't prevent the user from commiting. A validating webhook would allow us to reject these values synchronously before the CR is written, thus relieving both us and the user from asynchronous error reporting.

This bug will serve as a place to record known invalid configs that should be checked by a webhook, until such time as it is implemented.

- [ ] Hardware RAID + Software RAID requested simultaneously
- [ ] Other invalid RAID config???
- [ ] Resource name that is invalid in ironic (looks like a UUID or contains chars other than `[A-Za-z0-9~._-]`
- [ ] Resource name that contains a `~` (used in ironic Node name to separate name + namespace)
- [ ] Names that are invalid as part of a DNS hostname
- [ ] Changing BMC Address once set
- [ ] Changing Boot MAC once set
- [ ] Invalid BMC Config
- [ ] Duplicate BMC addresses
- [ ] Duplicate Boot MAC addresses

andfasano:
- Expected value on some annotations (ie `inspect.metal3.io`)?
- It would be useful to capture all these checks in one point only, so that they could be shared by both the webhook and controller

ardaguclu:
@s3rj1k are you working on it?

s3rj1k:
@ardaguclu No, just watching issue for updates

hardys:
> Changing Boot MAC once set

We had some discussion on this - it *is* valid to change this in some cases, for example if you deploy via virtualmedia it can potentially work with any MAC on the system, but if you later want to deploy via PXE (or as we're doing downstream, derive the provisioning-network NIC from this MAC) then it must be the correct nic.

There are reported cases e.g https://bugzilla.redhat.com/show_bug.cgi?id=2000081 of this being configured wrongly but deployment still succeeding, in which case it's likely we'd want to enable patching of the bootMACAddress to some other value (but we could still e.g validate that it's a valid mac for the system by comparing with the status.hardware data)

zaneb:
> > Changing Boot MAC once set
> 
> We had some discussion on this - it _is_ valid to change this in some cases, for example if you deploy via virtualmedia it can potentially work with any MAC on the system, but if you later want to deploy via PXE (or as we're doing downstream, derive the provisioning-network NIC from this MAC) then it must be the correct nic.

Changing from virtualmedia->PXE means changing the driver, so it also means a new BMC address (which is never safe).

[...]

PR #936, opened 2021-07-10
Add PreprovisioningImage API and integration

zaneb:
As proposed in https://github.com/metal3-io/metal3-docs/blob/master/design/baremetal-operator/image-builder-integration.md#custom-agent-image-controller


pkg/provisioner/provisioner.go
```diff
@@ -59,11 +59,19 @@ type HostConfigData interface {
 	MetaData() (string, error)
 }
 
+type PreprovisioningImage struct {
+	ImageURL     string
+	Format       metal3v1alpha1.ImageFormat
+	Checksum     string
```
dtantsur, line 65: Boot ISO/initramfs don't need checksum on the ironic side.

apis/metal3.io/v1alpha1/preprovisioningimage_types.go
```diff
@@ -0,0 +1,116 @@
...
+	ImageUrl string `json:"imageUrl,omitempty"`
+
+	// format is the type of image that is available at the download url:
+	// either iso or initrd.
+	// +optional
+	Format ImageFormat `json:"format,omitempty"`
+
+	// checksum is the image checksum.
+	// +optional
+	Checksum string `json:"checksum,omitempty"`
```
dtantsur, line 69: Checksums are only required for qcow2 instance image, at least for ironic.

apis/metal3.io/v1alpha1/preprovisioningimage_types.go
```diff
@@ -0,0 +1,116 @@
...
+	ConditionImageReady ImageStatusConditionType = "Ready"
+
+	// Error indicates that the operator was unable to build an image.
+	ConditionImageError ImageStatusConditionType = "Error"
+)
+
+// PreprovisioningImageStatus defines the observed state of PreprovisioningImage
+type PreprovisioningImageStatus struct {
+	// imageUrl is the URL from which the built image can be downloaded.
+	ImageUrl string `json:"imageUrl,omitempty"`
```
dtantsur, line 60: This may not work out for initramfs images of CoreOS specifically, since it has 3 components.

Actually, a minimum ISO image will also have 2 components.. dunno what to do about it.

controllers/metal3.io/preprovisioningimage_controller.go
```diff
@@ -0,0 +1,241 @@
...
+
+func (r *PreprovisioningImageReconciler) update(img *metal3.PreprovisioningImage, log logr.Logger) (bool, error) {
+	generation := img.GetGeneration()
+
+	url, format, errorMessage := getImageURL()
+	if errorMessage != "" {
+		log.Info("no suitable image URL available", "preferredFormat", format)
+		return setError(generation, &img.Status, reasonConfigurationError, errorMessage), nil
+	}
+
```
dtantsur, line 101: Maybe fail if `Architecture` is set? There is no support for it here.

zaneb:
I got this working and added unit tests, I believe it's ready for review now.
/test-integration

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -592,6 +596,128 @@ func (r *BareMetalHostReconciler) detachHost(prov provisioner.Provisioner, info
...
 
+type imageBuildError struct {
+	Message string
+}
+
+func (ibe imageBuildError) Error() string {
+	return ibe.Message
+}
+
+func (r *BareMetalHostReconciler) preprovImageCurrent(info *reconcileInfo, image *metal3v1alpha1.PreprovisioningImage) (bool, error) {
```
andfasano, line 607: naming: what about something like `preprovImageReady` to make it a little bit more clear?

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -592,6 +596,128 @@ func (r *BareMetalHostReconciler) detachHost(prov provisioner.Provisioner, info
...
+		}
+		controllerutil.SetControllerReference(info.host, &preprovImage, r.Scheme())
+		err = r.Create(context.TODO(), &preprovImage)
+		return nil, err
+	}
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to retrieve pre-provisioning image data")
+	}
+
+	if !apiequality.Semantic.DeepEqual(preprovImage.Spec, expectedSpec) {
```
andfasano, line 703: The reason for this block isn't very clear to me, could you please provide some details? Shouldn't be the BMH controller take care just of the resource creation (and wait for the resource to be ready)?

andfasano (commented):
/approve

pkg/provisioner/ironic/ironic.go
```diff
@@ -504,24 +504,122 @@ func (p *ironicProvisioner) ValidateManagementAccess(data provisioner.Management
...
+		}
+	}
+
+	if !config.havePreprovImgBuilder {
+		if allowISO && config.deployISOURL != "" {
+			deployImageInfo[deployISOKey] = config.deployISOURL
+			return deployImageInfo
+		}
+	}
+	if !config.havePreprovImgBuilder || (hostImage != nil && !allowISO) {
```
zaneb, line 612: Re: https://github.com/metal3-io/baremetal-operator/pull/936/files#r714032665 I think this has to change to:

    if !config.havePreprovImgBuilder || !allowISO {

because otherwise it is out of sync with `PreprovisioningImageFormats()`, which does not request a PreprovisioningImage CR to be created if this is a PXE-only host (at line 560).

controllers/metal3.io/baremetalhost_controller.go
zaneb: If any of the data changes (specifically, if a different network data Secret is specified), we need to update the PreprovisioningImage so that the next boot will use the latest data.

controllers/metal3.io/preprovisioningimage_controller.go
```diff
@@ -0,0 +1,237 @@
...
+}
+
+func (r *PreprovisioningImageReconciler) CanStart() bool {
+	deployISOURL := os.Getenv("DEPLOY_ISO_URL")
+	hasCfg := deployISOURL != ""
+	if hasCfg {
+		r.Log.Info("have deploy image data",
+			"iso_url", deployISOURL)
+	} else {
+		r.Log.Info("not starting preprovisioning image controller; no image data available")
```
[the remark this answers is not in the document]
andfasano: Got it thanks

controllers/metal3.io/baremetalhost_controller.go
zaneb: The Preprov controller is watching the Secret for changes, but if the user changes the BMH to point to a _different_ Secret then it is the BMH controller's responsibility to update the PreprovisioningImage. The PreprovisioningImage controller never updates its own Spec.

controllers/metal3.io/preprovisioningimage_controller.go
```diff
@@ -0,0 +1,237 @@
...
+	}, nil
+}
+
+func setCondition(generation int64, status *metal3.PreprovisioningImageStatus,
+	cond metal3.ImageStatusConditionType, newStatus metav1.ConditionStatus,
+	time metav1.Time, reason conditionReason, message string) {
+	newCondition := metav1.Condition{
+		Type:               string(cond),
+		Status:             newStatus,
+		LastTransitionTime: time,
```
bfournie, line 171: It looks like LastTransitionTime will be determined by SetStatusCondition and doesn't need to be explicitly set - https://github.com/kubernetes/apimachinery/blob/86c0c0f8c8e2ad6601321a412cddb1f99ff6fae5/pkg/api/meta/conditions.go#L37.

zaneb: This is true. By passing a timestamp we can make sure that all of the conditions that change simultaneously get the same timestamp... which is something that probably only I care about :laughing:

bfournie: OK, Yeah if you want them all to have the same timestamp passing it in will work.

apis/metal3.io/v1alpha1/preprovisioningimage_types.go
zaneb: Yeah, I understand this a bit better now.
We only need to supply a URL for the part that can be customised with network config, which AIUI is only the initrd. For the minimal ISO, the rootfs gets fetched over the network, so it is not involved in network setup and can have a fixed URL.

pkg/ironic/testbmc/testbmc.go
```diff
@@ -92,6 +92,10 @@ func (a *testAccessDetails) SupportsSecureBoot() bool {
 	return false
 }
 
+func (a *testAccessDetails) SupportsISOPreprovisioningImage() bool {
+	return false
+}
+
```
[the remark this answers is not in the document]
zaneb: I'm not sure that a _driver_ (as opposed to hardware) supporting one necessarily implies that it supports the other. If that is the case then we could change it in future if we find another need for it.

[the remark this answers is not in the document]
hardys: @zaneb Not blocking on this, but just to clarify - isn't it the exact same process, e.g in both cases we have to boot an ISO image, and only *-virtualmedia drivers support that?

I guess in future we could make it possible to iPXE boot an ISO image, but that's dependent on the target image, so I'm not sure I get the distinction between booting an ISO preprovisioning image, or a user-provided ISO - the underlying requirements (of the hardware, expressed via the driver) are the same are they not?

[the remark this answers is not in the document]
dtantsur: Strictly speaking, live ISO and deploy using ISO are different workflows in Ironic. They have a lot of technical similarities, and are normally implemented by the same drivers though. I guess I don't have a strong opinion on your proposal, but these are not 100% the same things.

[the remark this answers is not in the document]
zaneb: @hardys I added this validation to the list in #855 so we don't forget about it.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -598,6 +602,128 @@ func (r *BareMetalHostReconciler) detachHost(prov provisioner.Provisioner, info
...
+				"currentVersion", image.Status.NetworkData.Version)
+			return false, nil
+		}
+	}
+	if image.Status.NetworkData.Name != image.Spec.NetworkDataName {
+		info.log.Info("network data location in pre-provisioning image is out of date")
+		return false, nil
+	}
+
+	for _, cond := range image.Status.Conditions {
```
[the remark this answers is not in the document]
zaneb: Thanks, this is a good suggestion. I think I had not yet discovered those handy functions when I wrote this :smile:

PR #937, opened 2021-07-12
Move tools modules under hack/tools

ardaguclu:
This PR creates new go module under hack/tools to decouple dependencies of BMO and tools.

This prevents possible dependency conflicts and main go module includes lesser dependencies.


hack/Dockerfile.golint
```diff
@@ -1,6 +0,0 @@
-FROM registry.hub.docker.com/library/golang:1.16
-
-RUN go get -u golang.org/x/lint/golint
-
-LABEL io.k8s.display-name="Metal3 golint test image" \
-      io.k8s.description="This image is for running tests requiring golint"
```
zaneb, line 6: IIUC this is still used in CI here: https://github.com/metal3-io/project-infra/blob/master/prow/config/config.yaml#L312-L325

hack/golint.sh
```diff
@@ -15,6 +15,6 @@ else
     --volume "${PWD}:/go/src/github.com/metal3-io/baremetal-operator:ro,z" \
     --entrypoint sh \
     --workdir /go/src/github.com/metal3-io/baremetal-operator \
-    quay.io/metal3-io/golint:latest \
+    docker.io/golangci/golangci-lint:latest \
```
zaneb, line 18: quay.io/metal3-io/golint:latest is the one used in CI, so if we don't use that then potentially developers will get different results to what CI gets.

zaneb (commented):
Maybe we need to create a separate "tools" module or something? curl | sh just doesn't seem like a solution to me.
/cc @asalkeld

hack/golint.sh
ardaguclu: After your and @honza's comments, I realized that this is not simply changeable image. It is better to stay using `quay.io/metal3-io/golint:latest` and make changes on it, if it is necessary.

ardaguclu:
> Maybe we need to create a separate "tools" module or something? curl | sh just doesn't seem like a solution to me.
> /cc @asalkeld
@zaneb It is interesting but golangci-lint suggests curl | sh installation(not only in local but also in CI) https://golangci-lint.run/usage/install/#ci-installation

hack/Dockerfile.golint
ardaguclu: According to the https://github.com/golang/go/issues/38968 and https://github.com/golangci/golangci-lint/issues/1892, golint has been deprecated. In any case, at some point we will need to remove golint and replace it with revive. Why we are using go base image and install golint in here?, maybe we can use golangci-lint official image directly. 

Thanks to that change, we can just change golint to revive in here https://github.com/metal3-io/baremetal-operator/blob/62372c1daa01c631da46109078c9872128088f1e/.golangci.yml#L7. (I run revive in our repo and just trivial changes needed.)

cc: @honza

ardaguclu:
kubebuilder seems using this curl  | sh installation method for golangci-lint https://github.com/kubernetes-sigs/kubebuilder/blob/ed9853214b00408ca27ed26c02f24695ca92fcf7/Makefile#L82
As an example, istio operator also uses this way https://github.com/banzaicloud/istio-operator/blob/9c35d1c5dc420fa596e8ac032f43b0fcb42a986e/Makefile#L41

On the other hand, Prometheus operator is designed by adding extra tools.go under scripts in here https://github.com/prometheus-operator/prometheus-operator/blob/master/scripts/tools.go. To be honest, I don't fully grasp what we can get by taking golangci-lint dependency from one tools.go to another.

ardaguclu:
I agree with @s3rj1k about unneeded packages. But my main concern is about this comment in here https://golangci-lint.run/usage/install/#install-from-source

"Note: such go get installation aren't guaranteed to work. We recommend using binary installation."

zaneb:
> my main concern is about this comment in here https://golangci-lint.run/usage/install/#install-from-source
> 
> "Note: such go get installation aren't guaranteed to work. We recommend using binary installation."

Looking at their reasons:

* some users use `-u` flag for `go get`, which upgrades our dependencies. Resulting configuration wasn't tested and isn't guaranteed to work.

Don't do that then.

* `go.mod` replacement directive doesn't apply. It means a user will be using patched version of `golangci-lint` if we use such replacements.

I have no idea what this means TBH.

* it's stability depends on a user's Go version (e.g. on this compiler Go <= 1.12 bug).

The stability of *everything* depends on the Go version.

* we've encountered a lot of issues with Go modules hashes.

If Go modules don't work we have bigger problems.

* it allows installation from master branch which can't be considered stable.

Don't do that then.

* it's slower than binary installation

CI doesn't care.

I'm sure that's the recommendation that results in them getting the fewest support requests, but it doesn't mean it's the right thing for us.

ardaguclu:
> > my main concern is about this comment in here https://golangci-lint.run/usage/install/#install-from-source
> > "Note: such go get installation aren't guaranteed to work. We recommend using binary installation."
> 
> Looking at their reasons:
> 
> * some users use `-u` flag for `go get`, which upgrades our dependencies. Resulting configuration wasn't tested and isn't guaranteed to work.
> 
> Don't do that then.

Totally agree.
> 
> * `go.mod` replacement directive doesn't apply. It means a user will be using patched version of `golangci-lint` if we use such replacements.
> 
> I have no idea what this means TBH.

Me neither.
> 
> * it's stability depends on a user's Go version (e.g. on this compiler Go <= 1.12 bug).
> 
> The stability of _everything_ depends on the Go version.

Since we are using the latest version of Go, this does not apply to us.
> 
> * we've encountered a lot of issues with Go modules hashes.
> 
> If Go modules don't work we have bigger problems.

This reason statement might be misleading. AFAIU, golangci-lint comes with lots of additional dependencies and it might conflict with the needed ones. That can be problem in the future.
> 
> * it allows installation from master branch which can't be considered stable.
> 
> Don't do that then.

Agree.
> 
> * it's slower than binary installation
> 
> CI doesn't care.

If CI does not care this, I think we can conclude that CI also does not care the lots of dependencies come with golangci-lint. Thus, this PR's motivation should be shifted from reducing binary size to decouple golangci-lint dependency from main go module to better maintenance.
> 
> I'm sure that's the recommendation that results in them getting the fewest support requests, but it doesn't mean it's the right thing for us.

I have one last concern and if we also agree upon that I'll apply prometheus method by creating new tools under hack folder. We have go modules in main and go modules under apis as well. Now we will have another go modules under hacks. Is there any possibility that we are making things harder to manage?(maybe answer is no because golangci-lint will only be under hack.)

s3rj1k:
My main concern here is lots of packages related to golangci-lint that are tightly coupled with baremetal-operator.

```
go.mod replacement directive doesn't apply. It means a user will be using patched version of golangci-lint if we use such replacements.
```

@zaneb 
They say, if someone is importing or building golangci-lint out of their code tree, all go.mod replacements that they do are ignored, this is how go.mod works.

ardaguclu:
> My main concern here is lots of packages related to golangci-lint that are tightly coupled with baremetal-operator.
> 
> ```
> go.mod replacement directive doesn't apply. It means a user will be using patched version of golangci-lint if we use such replacements.
> ```
> 
> @zaneb
> They say, if someone is importing or building golangci-lint out of their code tree, all go.mod replacements that they do are ignored, this is how go.mod works.

@s3rj1k , if we don't care about reducing binary size, moving golangci-lint from main module to it's own separate module under hack/ also decouples these dependencies.

zaneb:
To be clear, I don't believe that this, or any, change could have any effect on the binary size - since none of the dependencies pulled in by golangci-lint are called, or even imported in the packages we're actually building, they shouldn't end up in the binary.

What we can do is move them out of the top-level Go module, so that when somebody imports packages from BMO they don't have to worry about potential conflict between the dependencies only included for our CI and whatever they are doing. Although I'd expect that in most cases people would only be importing the API module now that that's separate. So it's a nice, if fairly low priority, thing to improve.

FWIW I gather that the cool kids are doing a separate Go module per package(!) nowadays, so I think we are still well short of peak go module; I wouldn't worry about the overhead of having 3.

asalkeld:
> To be clear, I don't believe that this, or any, change could have any effect on the binary size - since none of the dependencies pulled in by golangci-lint are called, or even imported in the packages we're actually building, they shouldn't end up in the binary.
> 
> What we can do is move them out of the top-level Go module, so that when somebody imports packages from BMO they don't have to worry about potential conflict between the dependencies only included for our CI and whatever they are doing. Although I'd expect that in most cases people would only be importing the API module now that that's separate. So it's a nice, if fairly low priority, thing to improve.
> 
> FWIW I gather that the cool kids are doing a separate Go module per package(!) nowadays, so I think we are still well short of peak go module; I wouldn't worry about the overhead of having 3.

:+1: I would be more in favor of this approach. Move all tools to hack/tools/ and build binaries for them.
Also: a nice-to-have would be in the Makefile
```
${golintbin}: hack/tools/go.sum
```
(so when your tools are updated, then they are rebuilt..)

elfosardo (approved):

asalkeld (commented):
lgtm

zaneb (commented):
/lgtm

PR #938, opened 2021-07-13
Create and manage hostFirmwareSettings and firmwareSchema resources

bfournie:


apis/metal3.io/v1alpha1/firmwareschema_types.go
```diff
@@ -80,56 +82,65 @@ type FirmwareSchema struct {
...
 	switch schema.AttributeType {
 	case "Enumeration":
 		for _, av := range schema.AllowableValues {
 			if value.String() == av {
-				return true
+				return nil
 			}
 		}
-		return false
+		return fmt.Errorf("Setting %s uses invalid enum %s", name, value.String())
```
dtantsur, line 104: nit: "enumeration value"?

apis/metal3.io/v1alpha1/firmwareschema_types.go
```diff
@@ -80,56 +82,65 @@ type FirmwareSchema struct {
...
 
 	case "Integer":
 		if schema.LowerBound == nil || schema.UpperBound == nil {
```
dtantsur, line 107: In theory we may have only one bound, right? E.g. "more than 1"

apis/metal3.io/v1alpha1/firmwareschema_types.go
```diff
@@ -80,56 +82,65 @@ type FirmwareSchema struct {
...
+			return nil
+		}
+		if value.IntValue() >= *schema.LowerBound && value.IntValue() <= *schema.UpperBound {
+			return nil
 		}
-		return (value.IntValue() >= *schema.LowerBound && value.IntValue() <= *schema.UpperBound)
+		return fmt.Errorf("Setting %s integer %s is out of range, %d - %d", name, value.String(), *schema.LowerBound, *schema.UpperBound)
 
 	case "String":
 		if schema.MinLength == nil || schema.MaxLength == nil {
```
dtantsur, line 117: Same, in theory nothing prevents us from having only upper limit.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -717,6 +722,13 @@ func (r *BareMetalHostReconciler) actionInspecting(prov provisioner.Provisioner,
 		return result
 	}
 
+	// Get firmware settings and manage resources
+	err = r.storeHostFirmwareSettings(prov, info)
+	if err != nil {
+		// log but continue, failure to get settings is not an error
+		info.log.Info("could not get firmware settings for host")
```
dtantsur, line 729: Maybe log the error? Otherwise it's probably lost.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1161,6 +1188,243 @@ func saveHostProvisioningSettings(host *metal3v1alpha1.BareMetalHost) (dirty boo
...
+	if hfs.Status.Settings == nil {
+		// If settings cleared after pivot, recreate map and update settings
+		hfs.Status.Settings = make(metal3v1alpha1.SettingsMap)
+
+		currentSettings, _, err := prov.GetFirmwareSettings(false)
+		if err != nil {
+			info.log.Info("could not get firmware settings from provisioner")
+			return nil, nil
+		}
+		if err = r.updateHostFirmwareSettingsFromCurrent(currentSettings, hfs, nil); err != nil {
```
dtantsur, line 1210: nit: it's a bit weird that a method called `getSomething` also does an update.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1161,6 +1188,243 @@ func saveHostProvisioningSettings(host *metal3v1alpha1.BareMetalHost) (dirty boo
 	return
 }
 
+// Get the stored firmware settings and validate the Spec settings against the schema
+func (r *BareMetalHostReconciler) getHostFirmwareSettings(prov provisioner.Provisioner, info *reconcileInfo) (hfs *metal3v1alpha1.HostFirmwareSettings, err error) {
+
+	hfs = &metal3v1alpha1.HostFirmwareSettings{}
+	if r.Get(context.TODO(), info.request.NamespacedName, hfs) != nil {
+		// Could not get settings, don't return error as settings may not have been available at provisioner
```
dtantsur, line 1196: Maybe limit the number of ignored errors? E.g. ignore "not found", but not connection failures?

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1161,6 +1188,243 @@ func saveHostProvisioningSettings(host *metal3v1alpha1.BareMetalHost) (dirty boo
...
+		return nil, nil
+	}
+
+	if hfs.Status.Settings == nil {
+		// If settings cleared after pivot, recreate map and update settings
+		hfs.Status.Settings = make(metal3v1alpha1.SettingsMap)
+
+		currentSettings, _, err := prov.GetFirmwareSettings(false)
+		if err != nil {
+			info.log.Info("could not get firmware settings from provisioner")
```
dtantsur, line 1207: Same here?

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1161,6 +1188,243 @@ func saveHostProvisioningSettings(host *metal3v1alpha1.BareMetalHost) (dirty boo
...
+	if !k8serrors.IsNotFound(err) {
+		// Error reading the object
+		return errors.Wrap(err, "could not load host firmware settings")
+	}
+
+	// Get the current settings and schema
+	currentSettings, schema, err := prov.GetFirmwareSettings(true)
+	if err != nil {
+		info.log.Info("could not get firmware settings with schema, trying without")
+		// If host cannot return schema attempt to just get the settings
```
dtantsur, line 1269: How can it happen that we can get BIOS settings, but not a schema? In ironic schema is just a part of settings.

pkg/provisioner/ironic/ironic.go
```diff
@@ -825,6 +825,53 @@ func (p *ironicProvisioner) getUpdateOptsForNode(ironicNode *nodes.Node, data pr
...
+func (p *ironicProvisioner) GetFirmwareSettings(includeSchema bool) (settings metal3v1alpha1.SettingsMap, schema map[string]metal3v1alpha1.SettingSchema, err error) {
+
+	if p.nodeID == "" {
+		return nil, nil, provisioner.ErrNeedsRegistration
+	}
+
+	// Get the settings from Ironic via Gophercloud
+	var settingsList []nodes.BIOSSetting
+	var biosListErr error
+	if includeSchema {
```
dtantsur, line 838: It seems like we more or less always need a schema, why having this switch?

pkg/provisioner/ironic/ironic.go
```diff
@@ -825,6 +825,53 @@ func (p *ironicProvisioner) getUpdateOptsForNode(ironicNode *nodes.Node, data pr
...
+		opts := nodes.ListBIOSSettingsOpts{Detail: true}
```
dtantsur, line 839: For this to work you need to update the microversion, I think it's in ironic/clients.go

dtantsur (commented):
I haven't finished the review, but here are some comments.

apis/metal3.io/v1alpha1/firmwareschema_types.go
bfournie: Changed

apis/metal3.io/v1alpha1/firmwareschema_types.go
bfournie: Yes, good point, Changed this to check Lower and Upper separately.

apis/metal3.io/v1alpha1/firmwareschema_types.go
bfournie: Yep, changed this to evaluate them separately.

controllers/metal3.io/baremetalhost_controller.go
bfournie: Fixed.

controllers/metal3.io/baremetalhost_controller.go
bfournie: Yeah I agree, but we need to update the Status fields if they get removed after a pivot. i haven't found names that would express this better. Still thinking about it...

controllers/metal3.io/baremetalhost_controller.go
bfournie: Fixed to return an error if the Get fails but it isn't a NotFound returned.

controllers/metal3.io/baremetalhost_controller.go
bfournie: Fixed

controllers/metal3.io/baremetalhost_controller.go
bfournie: Now that the client is set to use the API with microversion 1.74 agreed this shouldn't happen, it was an artifact of my testing with an older microversion.

pkg/provisioner/ironic/ironic.go
bfournie: My thought here is once we have the schema it will no longer change so we just need to do a non-detailed query to get the settings without including the schema fields, reducing the API bandwidth.

pkg/provisioner/ironic/ironic.go
bfournie: Yes, very good point. Changed to use microversion 1.74 and updated docs/configuration.md.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1153,6 +1180,245 @@ func saveHostProvisioningSettings(host *metal3v1alpha1.BareMetalHost) (dirty boo
...
+		}
+	}
+
+	schema := &metal3v1alpha1.FirmwareSchema{}
+	if hfs.Status.FirmwareSchema != nil {
+		if r.Get(context.TODO(), client.ObjectKey{
+			Namespace: hfs.Status.FirmwareSchema.Namespace,
+			Name:      hfs.Status.FirmwareSchema.Name}, schema) != nil {
+			info.log.Info("failed to get schema for hostFirmwareSettings", "namespacename", hfs.Status.FirmwareSchema.Namespace, "name", hfs.Status.FirmwareSchema.Name)
+			return hfs, nil
```
dtantsur, line 1218: Is it a normal situation? I would assume that if we have a `FirmwareSchema` link it must resolve? Or do you expect a race here?

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1153,6 +1180,245 @@ func saveHostProvisioningSettings(host *metal3v1alpha1.BareMetalHost) (dirty boo
...
+		return nil, err
+	}
+
+	return hfs, nil
+}
+
+// Get the firmware settings from the provisioner and create hostFirmwareSettings and FirmwareSchema resources
+func (r *BareMetalHostReconciler) storeHostFirmwareSettings(prov provisioner.Provisioner, info *reconcileInfo) (err error) {
+
+	info.log.Info("hostFirmwareSettings")
```
dtantsur, line 1235: nit: storeHostFirmwareSettings

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1153,6 +1180,245 @@ func saveHostProvisioningSettings(host *metal3v1alpha1.BareMetalHost) (dirty boo
...
+
+	// First check if there is a hostFirmwareSettings resource for this host
+	hfs := &metal3v1alpha1.HostFirmwareSettings{}
+	if err = r.Get(context.TODO(), info.request.NamespacedName, hfs); err == nil {
+		info.log.Info("found existing hostFirmwareSettings resource")
```
dtantsur, line 1240: Should be Debug I think

pkg/provisioner/ironic/ironic.go
dtantsur: I wonder if the schema can change in run time, e.g. after a firmware update.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1153,6 +1180,245 @@ func saveHostProvisioningSettings(host *metal3v1alpha1.BareMetalHost) (dirty boo
...
+	return nil
+}
+
+// Update the HostFirmwareSettings resource using the settings and schema from provisioner
+func (r *BareMetalHostReconciler) updateHostFirmwareSettingsFromCurrent(settings metal3v1alpha1.SettingsMap, hfs *metal3v1alpha1.HostFirmwareSettings, schema *metal3v1alpha1.FirmwareSchema) (err error) {
+
+	// if no schema provided, get the one referenced by hostFirmwareSettings if available
+	if schema == nil && hfs.Status.FirmwareSchema != nil {
+		schema = &metal3v1alpha1.FirmwareSchema{}
+		if r.Get(context.TODO(), client.ObjectKey{
```
dtantsur, line 1322: I have a feeling that this is redundant since all places where updateHostFirmwareSettingsFromCurrent is called already try to fetch the current schema.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1153,6 +1180,245 @@ func saveHostProvisioningSettings(host *metal3v1alpha1.BareMetalHost) (dirty boo
...
+		if strings.Contains(k, "Password") {
+			continue
+		}
+		hfs.Status.Settings[k] = v
+
+		// Don't include setting in Spec if cannot be changed
+		if schema != nil {
+			if schemaSetting, ok := schema.Spec.Schema[k]; ok {
+				if (schemaSetting.ReadOnly != nil && *schemaSetting.ReadOnly == true) ||
+					(schemaSetting.Unique != nil && *schemaSetting.Unique == true) {
```
dtantsur, line 1347: Why do you skip unique fields? Worth a comment.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1153,6 +1180,245 @@ func saveHostProvisioningSettings(host *metal3v1alpha1.BareMetalHost) (dirty boo
...
+		}
+	}
+
+	if specDirty {
+		if err = r.Update(context.TODO(), hfs); err != nil {
+			return errors.Wrap(err, "could not update host firmware settings")
+		}
+	}
+
+	return r.Status().Update(context.TODO(), hfs)
```
dtantsur, line 1365: Should this go to an `else` block to avoid two subsequent calls to `Update`?

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -717,6 +722,13 @@ func (r *BareMetalHostReconciler) actionInspecting(prov provisioner.Provisioner,
 		return result
 	}
 
+	// Get firmware settings and manage resources
+	err = r.storeHostFirmwareSettings(prov, info)
+	if err != nil {
+		// log but continue, failure to get settings is not an error
+		info.log.Info("could not get firmware settings for host", "error", err)
```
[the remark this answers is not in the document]
bfournie: Changed this to info.log.Error()

controllers/metal3.io/baremetalhost_controller.go
bfournie: Changed this return an error.

controllers/metal3.io/baremetalhost_controller.go
bfournie: Maybe I'm missing something but it appears that Info is the lowest log level. Left it as info.log.Info() for now.

controllers/metal3.io/baremetalhost_controller.go
bfournie: Yes, good catch. Removed this.

controllers/metal3.io/baremetalhost_controller.go
bfournie: > I wonder if the schema can change in run time, e.g. after a firmware update.
Yeah I didn't consider this but its surely possible. Changed to get the schema each time an update is done instead of just the settings when the schema was already retrieved. I left GetFirmwareSettings() with the includeSchema param still in case we need to change this in the future.

>Why do you skip unique fields? Worth a comment.
I added a comment for this. The reasoning was that the Spec wouldn't be populated with settings unique to that host so that the resource could be copied to other resources without having to change or remove the unique settings manually. This field is set by some vendors for settings like "SysName" etc.

controllers/metal3.io/baremetalhost_controller.go
bfournie: So we only need to do the full Update() when the Spec changes, the second call is just for a Status().Update(). I think we still need the Status().Update if the other Update() is done, but I will retest this.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -815,6 +836,11 @@ func (r *BareMetalHostReconciler) actionPreparing(prov provisioner.Provisioner,
 		return result
 	}
 
+	// store settings updated after cleaning
+	if err = r.storeHostFirmwareSettings(prov, info); err != nil {
```
andfasano, line 840: The `storeHostFirmwareSettings` does a lot of things, is it fully required here?

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1153,6 +1179,227 @@ func saveHostProvisioningSettings(host *metal3v1alpha1.BareMetalHost) (dirty boo
...
+
+			info.log.Error(err, "failed to get schema for hostFirmwareSettings")
+			return nil, err
+		}
+	} else {
+		schema = nil
+	}
+
+	// Validate the Spec setting using schema, if available
+	if err = ValidateHostFirmwareSettings(hfs, schema); err != nil {
```
andfasano, line 1225: What's the reason to (re)validate the setting on get?

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1153,6 +1179,227 @@ func saveHostProvisioningSettings(host *metal3v1alpha1.BareMetalHost) (dirty boo
...
+	// First check if there is a hostFirmwareSettings resource for this host
+	hfs := &metal3v1alpha1.HostFirmwareSettings{}
+	if err = r.Get(context.TODO(), info.request.NamespacedName, hfs); err == nil {
+		info.log.Info("found existing hostFirmwareSettings resource")
+	} else if !k8serrors.IsNotFound(err) {
+		// Error reading the object
+		return errors.Wrap(err, "could not load host firmware settings")
+	} else {
+		// Create a new resource
+		if err = r.newHostFirmwareSettings(info, hfs); err != nil {
```
andfasano, line 1246: Wouldn't be safer to postpone the resource creation after a successful `prov.GetFirmwareSettings(true)`?

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -815,6 +836,11 @@ func (r *BareMetalHostReconciler) actionPreparing(prov provisioner.Provisioner,
 		return result
 	}
 
+	// store settings updated after cleaning
+	if err = r.storeHostFirmwareSettings(prov, info); err != nil {
+		info.log.Info("could not store firmware settings for host after cleaning", "error", err)
```
andfasano, line 841: In case of error it is expect to take some action, like an `actionError` or `actionFailed` (and to trigger the retry mechanism for transient error, there is the `recordActionFailure` helper method)

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1153,6 +1179,227 @@ func saveHostProvisioningSettings(host *metal3v1alpha1.BareMetalHost) (dirty boo
...
+		return nil, err
+	}
+
+	return hfs, nil
+}
+
+// Get the firmware settings from the provisioner and create hostFirmwareSettings and FirmwareSchema resources
+func (r *BareMetalHostReconciler) storeHostFirmwareSettings(prov provisioner.Provisioner, info *reconcileInfo) (err error) {
+
+	info.log.Info("storeHostFirmwareSettings")
```
andfasano, line 1235: That's not very informative

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1153,6 +1179,227 @@ func saveHostProvisioningSettings(host *metal3v1alpha1.BareMetalHost) (dirty boo
...
+		}
+		info.log.Info("updated hostFirmwareSettings settings without schema")
+	}
+
+	return nil
+}
+
+// Create a HostFirmwareSettings resource
+func (r *BareMetalHostReconciler) newHostFirmwareSettings(info *reconcileInfo, hfs *metal3v1alpha1.HostFirmwareSettings) (err error) {
+
```
andfasano, line 1289: I would have expected to find the `hfs` creation directly within the factory method (otherwise it's more an `update` rather than a `new` one)

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1153,6 +1179,227 @@ func saveHostProvisioningSettings(host *metal3v1alpha1.BareMetalHost) (dirty boo
...
+	// If settings cleared after then pivot recreate map, it will be repopulated below
+	if hfs.Status.Settings == nil {
+		hfs.Status.Settings = make(metal3v1alpha1.SettingsMap)
+	}
+
+	// Update Spec and Status
+	specDirty := false
+	for k, v := range settings {
+		// Some vendors include encrypted password fields, don't add these
+		if strings.Contains(k, "Password") {
```
andfasano, line 1317: Is the case relevant here?

controllers/metal3.io/hostfirmwaresettings_controller.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	// Check if settings has been cleared during pivot, and if so, restore firmwareSchema from annotation
+	if hfs.Status.FirmwareSchema == nil {
+		if _, ok := hfs.Annotations[metal3v1alpha1.FirmwareSchemaNameAnnotation]; ok {
+			hfs.Status.FirmwareSchema = &metal3v1alpha1.SchemaReference{
+				Name:      hfs.Annotations[metal3v1alpha1.FirmwareSchemaNameAnnotation],
+				Namespace: hfs.Annotations[metal3v1alpha1.FirmwareSchemaNamespaceAnnotation]}
+
+			if r.Status().Update(context.TODO(), hfs) != nil {
```
andfasano, line 78: So the purpose of this controller it's just to update the schema reference? Looks like a mixed responsibility situation between this controller and the baremetalhost one, which is slightly confusing for me.

controllers/metal3.io/hostfirmwaresettings_controller.go
```diff
@@ -0,0 +1,171 @@
...
+
+			// Error reading the object - requeue the request.
+			return ctrl.Result{}, errors.Wrap(err, "could not load firmware schema")
+		}
+	} else {
+		schema = nil
+	}
+
+	if err = ValidateHostFirmwareSettings(hfs, schema); err != nil {
+		reqLogger.Info("hostFirmwareSetting validation failed", "error", err.Error())
```
andfasano, line 110: Also this one is somehow confusing, especially considering that no action or status is reported to the user

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -717,6 +722,12 @@ func (r *BareMetalHostReconciler) actionInspecting(prov provisioner.Provisioner,
 		return result
 	}
 
+	// Get firmware settings and manage resources
+	if err = r.storeHostFirmwareSettings(prov, info); err != nil {
```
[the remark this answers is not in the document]
bfournie: Good question. The BIOS settings are independent of introspection, Ironic gets the settings directly from the BMC and can read them at any time (it currently gets them when host is moved to manageable or completes cleaning), it doesn't rely on IPA running to access the hardware on the host like introspection does. So BMO can also read them from Ironic at any time, even if introspection is disabled. They can also be read from externally provisioned hosts even if there is no introspection data.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -717,6 +721,11 @@ func (r *BareMetalHostReconciler) actionInspecting(prov provisioner.Provisioner,
 		return result
 	}
 
+	// Indicate that the hostFirmwareSettings resource should be updated by its controller
+	if err = r.markHostFirmwareSettingsForUpdate(info); err != nil {
```
andfasano, line 725: On a second thought I'm afraid that this approach won't work properly, as it doesn't seem enough to cope with the BareMetalHost lifecycle: as per the last discussions, every BMH should have its own HFS (and schema) - and moreover I guess that it should be deleted as long as the related BMH disappear.
That's also something to consider during an upgrade, where some pre-existing BMHs could exist withouth ofc any related HFS (and then should be created).
Given that, I think that a viable approach  could be to let the HFS controller watching also the BMH events, and move directly the logic for the creation (and update) directly in the HFS controller, trying to let the BMH controller just a simple reader of the HFS resources (just for building the cleaning steps).
In this way the relationship will be come a little bit more loosely coupled (even though I guess a sync point will still be required, at least to let understand the BMH that the its own HFS has been filled properly).

controllers/metal3.io/hostfirmwaresettings_controller.go
```diff
@@ -0,0 +1,295 @@
...
+	// Get settings from provisioner if update requested or settings are empty
+	if hfs.Status.ProvStatus.Update == true || hfs.Status.Settings == nil {
+
+		// Use the provisioner ID stored in annotation to access Ironic API
+		if _, ok := hfs.Annotations[metal3v1alpha1.ProvisionerIdAnnotation]; ok {
+			provID := hfs.Annotations[metal3v1alpha1.ProvisionerIdAnnotation]
+
+			info := &rInfo{log: reqLogger, hfs: hfs}
+
+			prov, err := r.ProvisionerFactory.NewProvisioner(provisioner.BuildEmptyHostData(provID), func(reason, message string) {})
```
andfasano, line 88: Good idea! For safey, it would be recommended to invoked the `IsReady()` method before using it

controllers/metal3.io/hostfirmwaresettings_controller.go
```diff
@@ -0,0 +1,295 @@
...
+
+	if err = r.updateResource(info, currentSettings, firmwareSchema); err != nil {
+		return errors.Wrap(err, "could not update hostFirmwareSettings")
+	}
+
+	return nil
+}
+
+// Update the HostFirmwareSettings resource using the settings and schema from provisioner
+func (r *HostFirmwareSettingsReconciler) updateResource(info *rInfo, settings metal3v1alpha1.SettingsMap, schema *metal3v1alpha1.FirmwareSchema) (err error) {
```
andfasano, line 136: Minor, but personally here I'd prefer a less generic name

controllers/metal3.io/baremetalhost_controller.go
bfournie: That's a really good point. I've removed the creation of the hfs and the setting of the update flag from the bmh and added event handlers to the hfs controller and Watches() for the bmh. This is cleaner and has reduced the coupling between the 2 controllers. The bmh just reads the associated hfs resource to get the clean steps.

controllers/metal3.io/hostfirmwaresettings_controller.go
bfournie: Added the check for IsReady() along with a reschedule if not ready.

controllers/metal3.io/hostfirmwaresettings_controller.go
bfournie: Changed name.

controllers/metal3.io/hostfirmwaresettings_controller.go
```diff
@@ -0,0 +1,430 @@
...
+			return true
+		} else if newState == metal3v1alpha1.StateProvisioned && oldState != metal3v1alpha1.StateProvisioned {
+			r.Log.Info("baremetalhost event state update", "oldstate", oldState, "newstate", newState)
+			return true
+		}
+
+		return false
+	}
+
+	// This is a hostfirmwaresettings resource, if the update increased the resource generation then process it
```
andfasano, line 340: Just as an extra safety, an explicit check on the hostfirmwaresettings type could prevent unwanted errors in case a new type will be added in the watch list

controllers/metal3.io/hostfirmwaresettings_controller.go
```diff
@@ -0,0 +1,430 @@
...
+	}
+	newHost, newHostExists := e.ObjectNew.(*metal3v1alpha1.BareMetalHost)
+	if newHostExists {
+		newState = newHost.Status.Provisioning.State
+	}
+	if oldHostExists || newHostExists {
+		// The data only needs to be retrieved from Ironic after the node has been moved to
+		// ready and when the node is fully provisionied, its not necessary on every state
+		// change as the data from Ironic will be the same.
+		if newState == metal3v1alpha1.StateReady && oldState == metal3v1alpha1.StatePreparing {
```
andfasano, line 329: That's not clear to me. What's the reason to reconcile when the BMH goes into `Ready` from `Preparing`?

controllers/metal3.io/hostfirmwaresettings_controller.go
```diff
@@ -0,0 +1,430 @@
...
+	reqLogger := r.Log.WithValues("hostfirmwaresettings", req.NamespacedName)
+	reqLogger.Info("start")
+
+	// Fetch the HostFirmwareSettings
+	hfs := &metal3v1alpha1.HostFirmwareSettings{}
+	if err = r.Get(ctx, req.NamespacedName, hfs); err != nil {
+		if k8serrors.IsNotFound(err) {
+			// A resource may not yet exist so create one in this namespace
+			info := &rInfo{log: reqLogger, namespaceName: req.NamespacedName}
+			if err = r.newHostFirmwareSettings(info); err != nil {
```
andfasano, line 79: Shouldn't be the decision to create a new HFS bound to the existence of the related BMH? It may sound strange to create the HFS without its parent object, otherwise

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -786,7 +790,20 @@ func (r *BareMetalHostReconciler) actionPreparing(prov provisioner.Provisioner,
 		prepareData.ActualRAIDConfig = nil
 		dirty = true
 	}
+
+	// Use settings in hostFirmwareSettings if available
+	hfs, err := r.getValidHostFirmwareSettings(info)
```
andfasano, line 795: There's a couple of points here that could deserve some attention/discussion.

The first one is related to the validation of the HFS settings: I can appreciate that re-applying it synchronously here may eliminate some issues, but in order to promote a better separation of the responsibilities between the controllers it seems more an operation that should belong to the HFS controller. In particular, its reconcile loop may take care of validating the current specs as soon as they are updated, and flag the HFS resource as valid/not-valid. In that way the BMH controller will have only to wait for a valid HFS resource here, before proceeding with the `Preparing` state, and it will become just a mere HFS reader (with the only responsibility to assemble the required data for the clean step). The HFS flag could be a simple boolean or, maybe better for future extensions, a proper state.

The second point it's still about the synchronization between the two controllers. Since they are independent, there could be a potential race (the effective probabilities are very low), where the HFS controller didn't yet fill correctly the HFS resource while the BMH controller arrives at this point and wants to go on. An explicit synchronization mechanism based on flag or status field could eliminate also this problem at all.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -786,7 +790,20 @@ func (r *BareMetalHostReconciler) actionPreparing(prov provisioner.Provisioner,
...
+	if err != nil {
+		// Return error if settings available but invalid
+		return actionError{errors.Wrap(err, "error with hostFirmwareSettings")}
```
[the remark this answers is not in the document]
bfournie: Good point. Added support for publishing an event when validation fails.

controllers/metal3.io/hostfirmwaresettings_controller.go
bfournie: Added check to confirm its a hostfirmwaresettings resource.

controllers/metal3.io/hostfirmwaresettings_controller.go
bfournie: I updated the comment to explain this a bit better. We don't need to retrieve data from ironic on every state transition but we do need to get it after cleaning as the settings could have changed (i.e. from Preparing to Ready).

controllers/metal3.io/hostfirmwaresettings_controller.go
bfournie: Yes, good point. I changed the logic to return an error if the BMH resource can't be retrieved instead of creating the HFS resource

controllers/metal3.io/baremetalhost_controller.go
bfournie: Sure, I moved the validation to the HFS controller and added a bool that can be read by the BMH controller. The flag is set after the Spec is compared to the Schema.

apis/metal3.io/v1alpha1/firmwareschema_types.go
```diff
@@ -54,85 +56,114 @@ type SettingSchema struct {
...
+				return nil
 			}
 		}
-		return false
+		return SchemaSettingError{name: name, message: fmt.Sprintf("unknown enumeration value - %s", value.String())}
 
 	case "Integer":
-		if schema.LowerBound == nil || schema.UpperBound == nil {
+		if schema.LowerBound == nil && schema.UpperBound == nil {
 			// return true if no settings to check validity
```
elfosardo, line 86: this comment probably needs to be changed

apis/metal3.io/v1alpha1/firmwareschema_types.go
```diff
@@ -54,85 +56,114 @@ type SettingSchema struct {
...
+		if schema.UpperBound != nil && value.IntValue() > *schema.UpperBound {
+			return SchemaSettingError{name: name, message: fmt.Sprintf("integer %s is above range %d", value.String(), *schema.UpperBound)}
 		}
-		return (value.IntValue() >= *schema.LowerBound && value.IntValue() <= *schema.UpperBound)
+		return nil
 
 	case "String":
-		if schema.MinLength == nil || schema.MaxLength == nil {
+		if schema.MinLength == nil && schema.MaxLength == nil {
 			// return true if no settings to check validity
```
elfosardo, line 99: ditto

pkg/provisioner/ironic/clients/client.go
```diff
@@ -95,8 +95,8 @@ func IronicClient(ironicEndpoint string, auth AuthConfig, tls TLSConfig) (client
 
 	// Ensure we have a microversion high enough to get the features
 	// we need. Update docs/configuration.md when updating the version.
-	// Version 1.69 introduces deploySteps argument to provisioning.
-	client.Microversion = "1.69"
+	// Version 1.74 allows retrival of the BIOS Registry
```
elfosardo, line 98: nit: retrieval

elfosardo (approved):

controllers/metal3.io/hostfirmwaresettings_controller.go
```diff
@@ -0,0 +1,448 @@
...
+	}
+
+	// only run again on changes
+	return ctrl.Result{}, nil
+}
+
+// Get the firmware settings from the provisioner and update hostFirmwareSettings
+func (r *HostFirmwareSettingsReconciler) updateHostFirmwareSettings(prov provisioner.Provisioner, info *rInfo) (err error) {
+
+	info.log.Info("retrieving firmware settings and saving to resource")
```
dtantsur, line 162: nit: Maybe add BMH and/or Node UUID?

controllers/metal3.io/hostfirmwaresettings_controller.go
```diff
@@ -0,0 +1,448 @@
...
+		}
+	}
+
+	// Create a provisioner that can access Ironic API
+	provID := bmh.Status.Provisioning.ID
+	prov, err := r.ProvisionerFactory.NewProvisioner(provisioner.BuildEmptyHostData(provID), info.publishEvent)
+	if err != nil {
+		return ctrl.Result{}, errors.Wrap(err, "failed to create provisioner")
+	}
+	info.log.Info("created a provisioner for host", "uuid", provID)
```
dtantsur, line 134: This may be confusing since Provisioner is an internal entity, not something user-observable. Do we really need to log this?

controllers/metal3.io/hostfirmwaresettings_controller.go
```diff
@@ -0,0 +1,448 @@
...
+			}
+		}
+		// Copy setting to Spec only if doesn't exist as setting may have been set by user
+		if _, ok := info.hfs.Spec.Settings[k]; !ok {
+			info.hfs.Spec.Settings[k] = intstr.FromString(v)
+			specDirty = true
+		}
+	}
+
+	if specDirty {
```
dtantsur, line 222: This is probably a good place to add some logging since we modify a resource

controllers/metal3.io/hostfirmwaresettings_controller.go
```diff
@@ -0,0 +1,448 @@
...
+			if schemaSetting, ok := schema.Spec.Schema[k]; ok {
+				if (schemaSetting.ReadOnly != nil && *schemaSetting.ReadOnly == true) ||
+					(schemaSetting.Unique != nil && *schemaSetting.Unique == true) {
+					continue
+				}
+			}
+		}
+		// Copy setting to Spec only if doesn't exist as setting may have been set by user
+		if _, ok := info.hfs.Spec.Settings[k]; !ok {
+			info.hfs.Spec.Settings[k] = intstr.FromString(v)
```
dtantsur, line 217: I must admit, I don't remember the design too well, but updating the spec seems confusing to me. Does it mean that if a setting gets changed by a 3rd party (from within the machine or even automatically by the BMC), we'll try to modify it because it no longer matches the cached spec?

I would expect the current settings to go to Status, while Spec only containing what the user wants us to assert.

controllers/metal3.io/hostfirmwaresettings_controller.go
```diff
@@ -0,0 +1,448 @@
...
+
+		// The setting must be in the Status
+		if _, ok := info.hfs.Status.Settings[name]; !ok {
+			return fmt.Errorf("Setting %s is not in the Status field", name)
+		}
+
+		// check validity of updated value
+		if schema != nil {
+			if err := schema.CheckSettingIsValid(name, val, schema.Spec.Schema); err != nil {
+				return err
```
dtantsur, line 413: An idea for a follow-up: collect all errors and report them at once.

pkg/provisioner/ironic/ironic.go
dtantsur: Reading the updated patch it seems like includeSchema is always `true` now?

pkg/provisioner/ironic/ironic.go
```diff
@@ -1001,19 +1049,50 @@ func (p *ironicProvisioner) buildManualCleaningSteps(bmcAccess bmc.AccessDetails
...
+	if data.ActualFirmwareSettings != nil {
+		// If we have the current settings from Ironic, update the settings to contain:
+		// 1. settings converted by BMC drivers that are different than current settings
+		for _, bmcsetting := range bmcsettings {
+			if val, exists := data.ActualFirmwareSettings[bmcsetting["name"]]; exists {
+				if bmcsetting["value"] != val {
+					newSettings = buildFirmwareSettings(newSettings, bmcsetting["name"], bmcsetting["value"])
+				}
+			} else {
+				p.log.Info("name converted from bmc driver not found in firmware settings", "name", bmcsetting["name"])
```
dtantsur, line 1068: Please log BMH and/or Node UUID

dtantsur (commented):
I'm leaving a bunch of comments here. Most are nits except for two:
1) A question/suggestion on API fields
2) A concern about updating Spec automatically.

apis/metal3.io/v1alpha1/firmwareschema_types.go
```diff
@@ -54,85 +56,114 @@ type SettingSchema struct {
...
 
-// Check whether the setting's name and value is valid using the schema
-func (host *FirmwareSchema) CheckSettingIsValid(name string, value intstr.IntOrString, schemas map[string]SettingSchema) bool {
-
-	schema, ok := schemas[name]
-	if !ok {
-		// The setting must exist in the status
-		return false
-	}
+func (schema *SettingSchema) IsValid(name string, value intstr.IntOrString) error {
```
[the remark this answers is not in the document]
bfournie: Fixed

apis/metal3.io/v1alpha1/firmwareschema_types.go
```diff
@@ -54,85 +56,114 @@ type SettingSchema struct {
...
-				return true
+				return nil
 			}
 		}
-		return false
+		return SchemaSettingError{name: name, message: fmt.Sprintf("unknown enumeration value - %s", value.String())}
 
 	case "Integer":
-		if schema.LowerBound == nil || schema.UpperBound == nil {
+		if schema.LowerBound == nil && schema.UpperBound == nil {
```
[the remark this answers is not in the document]
bfournie: Fixed

apis/metal3.io/v1alpha1/firmwareschema_types.go
```diff
@@ -54,85 +56,114 @@ type SettingSchema struct {
...
 			// return true if no settings to check validity
-			return true
+			return nil
+		}
+		if schema.LowerBound != nil && value.IntValue() < *schema.LowerBound {
+			return SchemaSettingError{name: name, message: fmt.Sprintf("integer %s is below range %d", value.String(), *schema.LowerBound)}
```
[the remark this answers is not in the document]
bfournie: Fixed to use "maximum/minimum"

apis/metal3.io/v1alpha1/firmwareschema_types.go
```diff
@@ -54,85 +56,114 @@ type SettingSchema struct {
...
+		}
+		if schema.UpperBound != nil && value.IntValue() > *schema.UpperBound {
+			return SchemaSettingError{name: name, message: fmt.Sprintf("integer %s is above range %d", value.String(), *schema.UpperBound)}
 		}
-		return (value.IntValue() >= *schema.LowerBound && value.IntValue() <= *schema.UpperBound)
+		return nil
 
 	case "String":
-		if schema.MinLength == nil || schema.MaxLength == nil {
+		if schema.MinLength == nil && schema.MaxLength == nil {
```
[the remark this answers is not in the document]
bfournie: Fixed

apis/metal3.io/v1alpha1/firmwareschema_types.go
```diff
@@ -54,85 +56,114 @@ type SettingSchema struct {
...
 			// return true if no settings to check validity
-			return true
+			return nil
 		}
-		return (len(value.String()) >= *schema.MinLength && len(value.String()) <= *schema.MaxLength)
+		strLen := len(value.String())
+		if schema.MinLength != nil && strLen < *schema.MinLength {
+			return SchemaSettingError{name: name, message: fmt.Sprintf("string %s length is below range %d", value.String(), *schema.MinLength)}
```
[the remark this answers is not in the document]
bfournie: Fixed

apis/metal3.io/v1alpha1/firmwareschema_types.go
```diff
@@ -54,85 +56,114 @@ type SettingSchema struct {
...
-		return (value.String() == "true" || value.String() == "false")
+		if value.String() == "true" || value.String() == "false" {
+			return nil
+		}
+		return SchemaSettingError{name: name, message: fmt.Sprintf("%s is not a boolean", value.String())}
 
 	case "Password":
 		// Prevent sets of password types
-		return false
+		return SchemaSettingError{name: name, message: "it is a Password type"}
```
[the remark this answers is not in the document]
bfournie: Changed

apis/metal3.io/v1alpha1/hostfirmwaresettings_types.go
```diff
@@ -33,6 +33,14 @@ type SchemaReference struct {
 	Name string `json:"name"`
 }
 
+type UpdateInfo struct {
```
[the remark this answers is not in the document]
bfournie: Yeah removing the nesting does simplify it a bit. Changed

apis/metal3.io/v1alpha1/hostfirmwaresettings_types.go
```diff
@@ -33,6 +33,14 @@ type SchemaReference struct {
...
+	// Stores the time that the settings were last updated
+	LastUpdated *metav1.Time `json:"lastUpdated,omitempty"`
+
+	// Track whether settings stored in the spec are valid based on the schema
+	SettingsValid bool `json:"settingsValid"`
```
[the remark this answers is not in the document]
bfournie: Changed to use IsValid

controllers/metal3.io/hostfirmwaresettings_controller.go
bfournie: Good point, added node uuid here so can remove log below

controllers/metal3.io/hostfirmwaresettings_controller.go
bfournie: Not really since now logging uuid. Removed

controllers/metal3.io/hostfirmwaresettings_controller.go
bfournie: Added log

controllers/metal3.io/hostfirmwaresettings_controller.go
bfournie: Note that the Status only gets copied to the Spec if that entry doesn't exist in the Spec. So the way it will work is first time we retrieve the data from Ironic both the Status and the Spec will be updated to the same values. This improves the user experience in that the actual value in the Spec can be just changed when desired without having to input the setting name. 

So your question is if the STATUS field gets updated by a 3rd party?  I would think it that case the reconciler would run and update the Status to the value from Ironic, in that case there would not be a mismatch with the Spec. The only time we'll try to modify the entry is if the Spec field actually gets updated

controllers/metal3.io/hostfirmwaresettings_controller.go
bfournie: Yeah that's a good idea to handle multiple invalid settings. I've updated it to handle multiple errors and publish events for each one

pkg/provisioner/ironic/ironic.go
bfournie: For includeSchema - yes its always true now as we always get the settings and full schema. i'd like to keep this logic though as we may find in the future there are situations where we just want to get the settings

bfournie: > Please log BMH and/or Node UUID
Added nodeID

controllers/metal3.io/hostfirmwaresettings_controller.go
dtantsur: No, I mean what if a setting is updated in the machine by something? E.g. some hardware have boot device as BIOS setting, ironic does update that. Will it cause any actions?

apis/metal3.io/v1alpha1/hostfirmwaresettings_types.go
```diff
@@ -53,6 +53,12 @@ type HostFirmwareSettingsStatus struct {
 
 	// Settings are the actual firmware settings stored as name/value pairs
 	Settings SettingsMap `json:"settings" required:"true"`
+
+	// Stores the time that the settings were last updated
+	LastUpdated *metav1.Time `json:"lastUpdated,omitempty"`
+
+	// Track whether settings stored in the spec are valid based on the schema
+	IsValid bool `json:"isValid"`
```
andfasano, line 61: As a further comment, and also to keep aligned the design adopted for sub-resources, this part looks like a good candidate to be implemented with `Conditions`(see the current work developed by @zaneb  on [PR 936](https://github.com/metal3-io/baremetal-operator/pull/936/files#diff-875ad84c1b2d03186f5fca344650cb756732a75844100709818981560a76e345R81)).
More recommendations about `Conditions` usage can be found [here](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties)

apis/metal3.io/v1alpha1/firmwareschema_types.go
bfournie: Fixed

controllers/metal3.io/hostfirmwaresettings_controller.go
```diff
@@ -0,0 +1,460 @@
...
+		Name: namespacedName.Name, Namespace: namespacedName.Namespace}
+	info.hfs.Status.Settings = make(metal3v1alpha1.SettingsMap)
+	info.hfs.Spec.Settings = make(metal3v1alpha1.DesiredSettingsMap)
+
+	if err = r.Create(context.TODO(), info.hfs); err != nil {
+		return errors.Wrap(err, "failure creating hostFirmwareSettings resource")
+	}
+	info.log.Info("created new hostFirmwareSettings resource")
+
+	return nil
```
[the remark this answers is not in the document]
bfournie: Yeah, good point, I can see that if the actual setting gets updated outside of Ironic it can cause a an inadvertent update to the old value. I've removed the syncing of the Spec on update. This will add an extra step for the user to create the entry in the Spec when making a change but it will prevent this spurious update.

apis/metal3.io/v1alpha1/hostfirmwaresettings_types.go
bfournie: Good idea and thanks for the links. That seems to be a standard way of providing info regarding the resource. I've added 3 conditions for changeRequested, Valid, and Error which can be used by the bmh to get the status of the resource.

controllers/metal3.io/hostfirmwaresettings_controller.go
```diff
@@ -0,0 +1,460 @@
...
+//
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.6.4/pkg/reconcile
+func (r *HostFirmwareSettingsReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, err error) {
+
+	reqLogger := r.Log.WithValues("hostfirmwaresettings", req.NamespacedName)
+	reqLogger.Info("start")
+
+	// Get the corresponding baremetalhost in this namespace, if one doesn't exist don't continue processing
+	bmh := &metal3v1alpha1.BareMetalHost{}
```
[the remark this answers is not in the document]
bfournie: I was able to duplicate some situations in which the provisioning id changed but the bmh and hfs resources stayed up when starting localbmo and the pods were recreated. It may be contrived using that script but I think its safer to get the bmh resource each time as the performance hit is quite small. In addition there may be cases where the bmh has been deleted but the hfs resource hasn't been removed yet.

controllers/metal3.io/hostfirmwaresettings_controller.go
bfournie: > No, I mean what if a setting is updated in the machine by something? E.g. some hardware have boot device as BIOS setting, ironic does update that. Will it cause any actions?

Yeah, good point, I can see that if the actual setting gets updated outside of Ironic it can cause a an inadvertent update to the old value. I've removed the syncing of the Spec on update. This will add an extra step for the user to create the entry in the Spec when making a change but it will prevent this spurious update.

apis/metal3.io/v1alpha1/hostfirmwaresettings_types.go
```diff
@@ -33,6 +33,16 @@ type SchemaReference struct {
 	Name string `json:"name"`
 }
 
+type SettingsConditionType string
+
+const (
+	// Indicates that the settings in the Spec are different than Status
+	ChangeRequested SettingsConditionType = "ChangeRequested"
```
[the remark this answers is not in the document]
bfournie: Well the comparison logic does happen in the provisioner during cleaning but we can avoid that whole step from the bmh if ChangeRequested is false. Its also a useful condition to present to the user, since updating the settings requires a reboot a user can determine if a reboot is required by looking at this condition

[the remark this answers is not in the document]
andfasano: Fine, maybe then a name like `UpdateRequested` could convey better the message?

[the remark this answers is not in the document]
bfournie: Sure, changed to use that.

dtantsur:
/approve

Thanks!

controllers/metal3.io/hostfirmwaresettings_controller.go
```diff
@@ -0,0 +1,466 @@
...
+	for k := range schema {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+
+	h := sha256.New()
+	h.Write([]byte(fmt.Sprintf("%v", keys)))
+	hash := fmt.Sprintf("%x", h.Sum(nil))[:8]
+
+	return "schema-" + hash
```
zaneb, line 465: IIRC we decided to put the vendor and model in the name as well as the shortened hash. Did that turn out not to be feasible?

controllers/metal3.io/hostfirmwaresettings_controller.go
```diff
@@ -0,0 +1,466 @@
...
+		Message:            message,
+	}
+	meta.SetStatusCondition(&status.Conditions, newCondition)
+}
+
+// Generate a name based on the schema keys which should be the same for similar hardware
+func GetSchemaName(schema map[string]metal3v1alpha1.SettingSchema) string {
+
+	// Schemas from the same vendor and model should be identical for both keys and values.
+	// Hash the keys of the map to get a identifier unique to this schema.
```
zaneb, line 454: Why not include the values as well? Just having the keys sounds like it would risk collisions between different firmware versions.

controllers/metal3.io/hostfirmwaresettings_controller.go
```diff
@@ -0,0 +1,466 @@
...
+	}
+
+	if newHostExists && !oldHostExists {
+		// The baremetalhost has been created
+		return true
+	}
+	if oldHostExists || newHostExists {
+		// Data needs to be retrieved from Ironic when node is moved to Ready  e.g. after cleaning and when
+		// provisioned, its not necessary on every state change as data from Ironic will be the same.
+		if (newState == metal3v1alpha1.StateReady || newState == metal3v1alpha1.StateProvisioned) && oldState != newState {
```
zaneb, line 355: I don't believe it's safe in general to try to build an edge-triggered controller like this. Kubernetes is designed for everything to be [level-triggered](https://hackernoon.com/level-triggering-and-reconciliation-in-kubernetes-1f17fe30333d). There is no guarantee that we will see the edge (e.g. if the controller is restarted).

If we have to trigger on an edge I think the only safe way of doing so is to run it inside the BaremetalHost reconciler before writing the new provisioning status.

zaneb: `StateReady` is being renamed to `StateAvailable` (#340), so wherever one is referenced we have to include the other.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1001,19 +1049,50 @@ func (p *ironicProvisioner) buildManualCleaningSteps(bmcAccess bmc.AccessDetails
...
+					newSettings = buildFirmwareSettings(newSettings, bmcsetting["name"], bmcsetting["value"])
+				}
+			} else {
+				p.log.Info("name converted from bmc driver not found in firmware settings", "name", bmcsetting["name"], "node", p.nodeID)
+			}
+		}
+
+		// 2. target settings that are different than current settings
+		if data.TargetFirmwareSettings != nil {
+			for k, v := range data.TargetFirmwareSettings {
```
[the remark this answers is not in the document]
bfournie: Good point. Adding a check to skip if already defined in the platform-agnostic settings

controllers/metal3.io/hostfirmwaresettings_controller.go
```diff
@@ -0,0 +1,466 @@
...
+	} else {
+		for _, error := range errors {
+			info.publishEvent("ValidationFailed", fmt.Sprintf("Invalid BIOS setting: %v", error))
+		}
+
+		reason = reasonConfigurationError
+		setCondition(generation, &info.hfs.Status, metal3v1alpha1.SettingsValid, metav1.ConditionFalse, reason, "Invalid BIOS setting")
+	}
+
+	return r.Status().Update(context.TODO(), info.hfs)
```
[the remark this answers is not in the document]
bfournie: Adding a check to see if any Status fields have changed before doing the Update

controllers/metal3.io/hostfirmwaresettings_controller.go
```diff
@@ -0,0 +1,466 @@
...
+	if !k8serrors.IsNotFound(err) {
+		// Error reading the object
+		return nil, err
+
+	}
+
+	firmwareSchema = &metal3v1alpha1.FirmwareSchema{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      schemaName,
+			Namespace: info.hfs.ObjectMeta.Namespace,
```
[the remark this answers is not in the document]
bfournie: Yes that is still in plan as a follow on PR.

controllers/metal3.io/hostfirmwaresettings_controller.go
bfournie: I'm not sure if we would practically see differences in values but I'll add the values into the hash to prevent that happening.

controllers/metal3.io/hostfirmwaresettings_controller.go
bfournie: Will change to remove the edge-triggering in a follow on PR.

controllers/metal3.io/hostfirmwaresettings_controller.go
```diff
@@ -0,0 +1,466 @@
...
+func (r *HostFirmwareSettingsReconciler) updateStatus(info *rInfo, settings metal3v1alpha1.SettingsMap, schema *metal3v1alpha1.FirmwareSchema) (err error) {
+
+	if info.hfs.Status.Settings == nil {
+		info.hfs.Status.Settings = make(metal3v1alpha1.SettingsMap)
+	}
+
+	// Update Status
+	for k, v := range settings {
+		// Some vendors include encrypted password fields, don't add these
+		if strings.Contains(k, "Password") {
```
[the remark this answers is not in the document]
zaneb: OK, I'm recalling some stuff now and there's a bit of a usability issue here. Someone trying to clone the settings from one host to another can't just copy the settings from a Status to a Spec because the Status includes read-only settings.
In the design doc we said we were going to solve this by copying all of the writeable values from the Status to the Spec, but we decided not to do that.
Maybe the Status needs to be split into "Settings" (writeable) and "Data" (read-only). WDYT?

controllers/metal3.io/hostfirmwaresettings_controller.go
zaneb: I thought we wouldn't have the schema data available until after inspection either?
There is something to be said for not mixing data from two different sources though, since they could change independently.

controllers/metal3.io/hostfirmwaresettings_controller.go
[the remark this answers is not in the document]
bfournie: > LocalSettings?
> 
I like that

> Btw, IIRC this part will also be addressed in future by another controller (to copy the right settings to other hosts to keep them in sync)
Yes that is the plan, although from the VRAN discussions the importance was more on configuring the settings on a host during initial deployment than on keeping them in sync post-deployment

PR #941, opened 2021-07-22
Fix missing case of BuildRAIDCleanSteps

Hellcatlk:

In #908, I missed a case in `BuildRAIDCleanSteps` from the beginning of that PR, so add the handling of the case in this PR.

In summary, it adds the case that when `hardwareRAIDVolumes` is nil, keep the actual RAID configuration.

### DIFF
--- a/pkg/provisioner/ironic/raid.go
+++ b/pkg/provisioner/ironic/raid.go
@@ -210,10 +210,14 @@ func BuildRAIDCleanSteps(raidInterface string, target *metal3v1alpha1.RAIDConfig
 	}
 
 	// Hardware RAID
-	// Ignore SoftwareRAIDVolumes
-	if target != nil {
-		target.SoftwareRAIDVolumes = nil
+	// If hardware RAID configuration is nil,
+	// keep old hardware RAID configuration
+	if target == nil || target.HardwareRAIDVolumes == nil {
+		return
 	}
+
+	// Ignore SoftwareRAIDVolumes
+	target.SoftwareRAIDVolumes = nil
 	if actual != nil {
 		actual.SoftwareRAIDVolumes = nil
 	}
@@ -232,7 +236,7 @@ func BuildRAIDCleanSteps(raidInterface string, target *metal3v1alpha1.RAIDConfig
 	)
 
 	// If hardware raid configuration is empty, only need to clear old configuration
-	if target == nil || len(target.HardwareRAIDVolumes) == 0 {
+	if len(target.HardwareRAIDVolumes) == 0 {
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


Hellcatlk:
@andfasano

pkg/provisioner/ironic/raid.go
```diff
@@ -210,6 +210,12 @@ func BuildRAIDCleanSteps(raidInterface string, target *metal3v1alpha1.RAIDConfig
 	}
 
 	// Hardware RAID
+	// If hardware RAID configuration is nil,
+	// keep old hardware RAID configuration
+	if target == nil || target.HardwareRAIDVolumes == nil {
+		return
+	}
+
```
andfasano, line 218: The motivation for that case is unclear to me, and it also seems not correct, being in contrast with this later [check](https://github.com/metal3-io/baremetal-operator/pull/941/files#diff-d3542d6731fdd4d0ef86ed2bc8f7bbad22a4a0af2b8c28d0107835a32f7afce1L235-L237) - which instead appeared to be fine: if the desired raid configuration (`target`) is nil (or `HardwareRAIDVolumes` is nil or empty) then it's just cleared from the server (cleaning up any potentially existing previous configuration) (an this could also cover the case of multiple loops on the `Preparing` state, where maybe something meanwhile has changed). Also the software raid part follows a similar logic.

So, assuming that the current master implementation looks fine, could you please elaborate with more details the use case that you think it's missing (unfortunately also the PR description isn't very informative for that) ?

Hellcatlk: In the current master implementation, if the target node does not have a RAID controller, ironic will return clean failed because the `delete_configuration` clean step failed.

Hellcatlk: Therefore, our goal is:

- If hardwareRAIDVolume = nil
Keep actual RAID configuration

- If hardwareRAIDVolume = []
Clear actual RAID configuration

- Other
Apply RAID configuration to node

Hellcatlk: I just found that there is a redundant `if` statement in the place that you mentioned, I dropped it.

Hellcatlk: Additional explaintion:

This is the missing case:
User apply a new bmh without `raid` field, and at the same time, the target node doesn't have RAID controller.

In above case, when entering BuildRAIDCleanSteps(), the values of target and actual are:
```go
target = {
    hardwareRAIDVolume = nil
    softwareRAIDVolume = []
}

actual = nil
```

Because of above values of target and actual, BMO will append `delete_configuration` into clean steps, then due to there is no RAID controller, an error will happen, which leads to reentry the preparing status (infinity loop).

fenggw-fnst:
We have opened a new PR #942 to replace this one in order to avoid affecting the openshift tracking PR[165](https://github.com/openshift/baremetal-operator/pull/165).
PR165 should only track #908, rather than includes this PR, since it is a bug fix.
Unfortunately, this PR and PR908 was on the same branch, so we have to close this one and open a new one.
Sorry for the inconvenience.

PR #942, opened 2021-07-22
Fix missing case of BuildRAIDCleanSteps

Hellcatlk:

This PR replaces #941. I closed PR941 to avoid affecting openshift [PR165](https://github.com/openshift/baremetal-operator/pull/165).
PR165 should only track #908, rather than includes PR941, since it is a bug fix.
Unfortunately, PR941 and PR908 was on the same branch, so I have to close PR941 and open this one, sorry about that.

This PR aims to address a case in BuildRAIDCleanSteps which I missed in PR908.

In summary, it add the handling of the case when hardwareRAIDVolumes is nil and the target node doesn't have a RAID controller.  We shouldn't delete the RAID configuration on that node in that case, otherwise the `ironic` node will enter the clean failed state because of  `delete_configuration` clean step failure.

### DIFF
--- a/pkg/provisioner/ironic/raid.go
+++ b/pkg/provisioner/ironic/raid.go
@@ -210,10 +210,14 @@ func BuildRAIDCleanSteps(raidInterface string, target *metal3v1alpha1.RAIDConfig
 	}
 
 	// Hardware RAID
-	// Ignore SoftwareRAIDVolumes
-	if target != nil {
-		target.SoftwareRAIDVolumes = nil
+	// If hardware RAID configuration is nil,
+	// keep old hardware RAID configuration
+	if target == nil || target.HardwareRAIDVolumes == nil {
+		return
 	}
+
+	// Ignore SoftwareRAIDVolumes
+	target.SoftwareRAIDVolumes = nil
 	if actual != nil {
 		actual.SoftwareRAIDVolumes = nil
 	}
@@ -232,7 +236,7 @@ func BuildRAIDCleanSteps(raidInterface string, target *metal3v1alpha1.RAIDConfig
 	)
 
 	// If hardware raid configuration is empty, only need to clear old configuration
-	if target == nil || len(target.HardwareRAIDVolumes) == 0 {
+	if len(target.HardwareRAIDVolumes) == 0 {
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/provisioner/ironic/raid.go
```diff
@@ -210,10 +210,14 @@ func BuildRAIDCleanSteps(raidInterface string, target *metal3v1alpha1.RAIDConfig
...
 	// Hardware RAID
-	// Ignore SoftwareRAIDVolumes
-	if target != nil {
-		target.SoftwareRAIDVolumes = nil
+	// If hardware RAID configuration is nil,
+	// keep old hardware RAID configuration
+	if target == nil || target.HardwareRAIDVolumes == nil {
+		return
 	}
+
```
Hellcatlk, line 218: @andfasano 
The missing that I think of is that:
In the current master implementation, if the target node does not have a RAID controller, the ironic node will enter the clean failed state because the delete_configuration clean step failed.

Therefore, our goal is:
- If hardwareRAIDVolume = nil
Keep actual RAID configuration
- If hardwareRAIDVolume = []
Clear actual RAID configuration
- Other
Apply RAID configuration to node

Hellcatlk: Additional explaintion:

This is the missing case:
User apply a new bmh without `raid` field, and at the same time, the target node doesn't have RAID controller.

In above case, when entering BuildRAIDCleanSteps(), the values of target and actual are:
```go
target = {
    hardwareRAIDVolume = nil
    softwareRAIDVolume = []
}

actual = nil
```

Because of above values of target and actual, BMO will append `delete_configuration` into clean steps, then due to there is no RAID controller, an error will happen, which leads to reentry the preparing status (infinity loop).

andfasano: (I guess it's `This is because when a node does not have a RAID controller`_`but supports RAID`_)

Thanks @Hellcatlk for the clarification and links (and also for having used a different branch), it's definitely clearer now. Could you please update the commit message to reflect this case (ie that hw RAID operations are skipped when a node does not have a RAID controller), since right now it still sound pretty generic?

Hellcatlk: Thanks, done.

Hellcatlk:
@andfasano Sorry, I just updated the commit message, causing `lgtm` to be removed.

Hellcatlk:
@dtantsur PTAL.

PR #945, opened 2021-07-28
Drop the duplicating logwatch

dtantsur:


zaneb:
Is this something to do with combining the ironic/inspector container images? There's no information here to go on.

dtantsur:
> Is this something to do with combining the ironic/inspector container images? There's no information here to go on.

Yep. If you check the container definitions, they're identical now.

fmuyassarov (commented):
I think you are missing one more file https://github.com/metal3-io/baremetal-operator/blob/399f5ef7ee3831014c1425250bc4fa49641a8709/tools/remove_local_ironic.sh#L9 
Also would you mind to clean up metal3-dev-env too? There are couple of references in [upgrade](https://github.com/metal3-io/metal3-dev-env/blob/598dce73b1da03e456ec5995078bb66d2b15cd56/vm-setup/roles/v1aX_integration_test/tasks/upgrade.yml) and [pivot](https://github.com/metal3-io/metal3-dev-env/blob/54abb30bb076e237de68b804d3bffe52235ccdb9/vm-setup/roles/v1aX_integration_test/tasks/move.yml) tests

dtantsur:
> I think you are missing one more file 

@fmuyassarov I'd avoid touching clean-up code for now just in case a clean-up from an older installation.

elfosardo (approved):

hardys:
/approve

As mentioned in the other comments, this is because the ironic-inspector-image merged with ironic-image since https://github.com/metal3-io/ironic-image/pull/253

PR #946, opened 2021-07-28
Explicitly set node bios_interface

bfournie:
Ironic sets the bios_interface by default for all drivers. For some drivers - particularly idrac-virtualmedia, the default should be overridden. If not explicitly set, Ironic will use the default.

This requires an upgrade to a Gophercloud release that supports BIOSInterface.

### DIFF
--- a/pkg/bmc/access_test.go
+++ b/pkg/bmc/access_test.go
@@ -439,6 +439,7 @@ func TestStaticDriverInfo(t *testing.T) {
 		input      string
 		needsMac   bool
 		driver     string
+		bios       string
 		boot       string
 		management string
 		power      string
@@ -450,6 +451,7 @@ func TestStaticDriverInfo(t *testing.T) {
 			input:      "ipmi://192.168.122.1:6233",
 			needsMac:   false,
 			driver:     "ipmi",
+			bios:       "",
 			boot:       "ipxe",
 			management: "",
 			power:      "",
@@ -462,6 +464,7 @@ func TestStaticDriverInfo(t *testing.T) {
 			input:      "libvirt://192.168.122.1",
 			needsMac:   true,
 			driver:     "ipmi",
+			bios:       "",
 			boot:       "ipxe",
 			management: "",
 			power:      "",
@@ -474,6 +477,7 @@ func TestStaticDriverInfo(t *testing.T) {
 			input:      "idrac://192.168.122.1",
 			needsMac:   false,
 			driver:     "idrac",
+			bios:       "",
 			boot:       "ipxe",
 			management: "",
 			power:      "",
@@ -486,6 +490,7 @@ func TestStaticDriverInfo(t *testing.T) {
 			input:      "irmc://192.168.122.1",
 			needsMac:   false,
 			driver:     "irmc",
+			bios:       "",
 			boot:       "pxe",
 			management: "",
 			power:      "",
@@ -498,6 +503,7 @@ func TestStaticDriverInfo(t *testing.T) {
 			input:      "redfish://192.168.122.1",
 			needsMac:   true,
 			driver:     "redfish",
+			bios:       "",
 			boot:       "ipxe",
 			management: "",
 			power:      "",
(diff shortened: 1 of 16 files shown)


pkg/bmc/access_test.go
```diff
@@ -594,6 +608,7 @@ func TestStaticDriverInfo(t *testing.T) {
 			input:      "idrac-virtualmedia+http://192.168.122.1",
 			needsMac:   true,
 			driver:     "idrac",
+			bios:       "redfish",
```
dtantsur, line 611: idrac-redfish

dtantsur (changes requested):

pkg/bmc/idrac_virtualmedia.go
```diff
@@ -70,6 +70,10 @@ func (a *redfishiDracVirtualMediaAccessDetails) Driver() string {
 	return "idrac"
 }
 
+func (a *redfishiDracVirtualMediaAccessDetails) BIOSInterface() string {
+	return "redfish"
```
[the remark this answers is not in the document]
bfournie: Thanks. Fixed.

pkg/provisioner/ironic/ironic.go
```diff
@@ -363,23 +363,28 @@ func (p *ironicProvisioner) ValidateManagementAccess(data provisioner.Management
...
+			VendorInterface:     bmcAccess.VendorInterface(),
+			Properties: map[string]interface{}{
+				"capabilities": bootModeCapabilities[data.BootMode],
+			},
+		}
+
+		if bmcAccess.BIOSInterface() != "" {
+			// Only override Ironic defaults if explicitly set
+			opts.BIOSInterface = bmcAccess.BIOSInterface()
+		}
```
andfasano, line 385: Q: Is this really required this check (ie, if `bmAccess.BIOSInterface()` is not set then `opts.BIOSInterface` will be an empty string, is there any issue for that in gophercloud/ironic api maybe)?

bfournie: Yes this is intentional and the way it currently works. If opts.BIOSInterface is set to "" then BIOSInterface will be omitted by Gophercloud - https://github.com/gophercloud/gophercloud/blob/master/openstack/baremetal/v1/nodes/requests.go#L189 and the Ironic default will be used. We only need to override the default when the default that Ironic picks is not desired.

zaneb: According to that link, `opts.BIOSInterface` is a `string` (not `*string`), so if it is not set then it will get the zero value for a string (which is `""`). So this check has no effect.

bfournie: Thanks. Fixed.

pkg/bmc/redfish.go
```diff
@@ -133,11 +137,14 @@ func (a *redfishAccessDetails) BuildBIOSSettings(firmwareConfig *metal3v1alpha1.
 }
 
 // iDrac Redfish Overrides
-
 func (a *redfishiDracAccessDetails) Driver() string {
 	return "idrac"
 }
 
+func (a *redfishiDracAccessDetails) BIOSInterface() string {
+	return ""
```
zaneb, line 145: It seems like users would probably expect us to be using `idrac-redfish` here also. IIUC it will default to `idrac-wsman` right now?

pkg/provisioner/ironic/ironic.go
```diff
@@ -363,23 +363,25 @@ func (p *ironicProvisioner) ValidateManagementAccess(data provisioner.Management
...
+			ManagementInterface: bmcAccess.ManagementInterface(),
+			PowerInterface:      bmcAccess.PowerInterface(),
+			RAIDInterface:       bmcAccess.RAIDInterface(),
+			VendorInterface:     bmcAccess.VendorInterface(),
+			Properties: map[string]interface{}{
+				"capabilities": bootModeCapabilities[data.BootMode],
+			},
+		}
+
+		opts.BIOSInterface = bmcAccess.BIOSInterface()
```
zaneb, line 382: This can just be set inline, can't it?

pkg/bmc/redfish.go
bfournie: Yes, good point. Fixed.

pkg/provisioner/ironic/ironic.go
bfournie: Right, since no check is needed. Fixed.

PR #948, opened 2021-08-02
Don't use soon removed APIs

fmuyassarov:
apiextensions.k8s.io/v1beta1 API of CustomResourceDefinition is going to be removed in 1.22 release of K8s. This patch replaces currently used beta APIs with already available v1.


zaneb:
I guess these must be files that are autogenerated by kubebuilder, and not regenerated from source. The generate job is checking them and still passing.
/lgtm

PR #949, opened 2021-08-06
Add LIVE_ISO_FORCE_PERSISTENT_BOOT_DEVICE variable

hardys:
When using format: live-iso, two distinct use-cases exist:
  - Boot an installer ISO (non-persistent one-time boot desired)
  - Boot an ISO appliance (persistent live-iso boot desired).

The original plan was to solve the installer-iso boot order via
uefibootmgr (running from the live-iso), but this doesn't work
reliably on all hardware.

So expose a variable to control this behavior - in future we may
want to consider a per-BMH variable, but for now expose a
per-BMO variable.


pkg/provisioner/ironic/ironic.go
```diff
@@ -352,7 +353,10 @@ func (p *ironicProvisioner) ValidateManagementAccess(data provisioner.Management
 		driverInfo["deploy_iso"] = p.config.deployISOURL
 	}
 
-	// If we have not found a node yet, we need to create one
+	if data.CurrentImage != nil && data.CurrentImage.IsLiveISO() && p.config.liveISOForcePersistentBootDevice != "" {
+		driverInfo["force_persistent_boot_device"] = p.config.liveISOForcePersistentBootDevice
+	}
```
zaneb, line 358: I'm not sure there's any need to set this here, given that we do getImageUpdateOptsForNode() later anyway.

We're not even always saving a potential write in the case where we're creating a new Node, because if liveISOForcePersistentBootDevice is set but the current image is not a live ISO, we're not doing anything here but we will set driver_info/force_persistent_boot_device to `"Default"` later.

hardys: Ack yeah I wasn't entirely sure if we needed it in both places, if we're sure this will work OK on the adopt case if we only set it in getImageUpdateOptsForNode() then removing it sounds fine - I'm assuming Ironic only asserts this boot-device during provisioning @derekhiggins?

derekhiggins: Ya, ironic will only verify BMC on adopt. I'll push up a new version without these lines.

flaper87 (commented):
This LGTM! I won't select approve or anything to avoid triggering any workflow :)

pkg/provisioner/ironic/ironic.go
```diff
@@ -731,6 +739,13 @@ func (p *ironicProvisioner) setDirectDeployUpdateOptsForNode(ironicNode *nodes.N
 	updater.
 		SetInstanceInfoOpts(optValues, ironicNode).
 		SetTopLevelOpt("deploy_interface", "direct", ironicNode.DeployInterface)
+
+	if p.config.liveISOForcePersistentBootDevice != "" {
+		driverOptValues := optionsData{
+			"force_persistent_boot_device": "Default",
```
dtantsur, line 745: Shouldn't this be p.config.liveISOForcePersistentBootDevice instead of "Default"?

derekhiggins: No because this is in setDirectDeployUpdateOptsForNode, it is used in the normal ipi flow. The idea here is to set it to default regardless of whats configured so that the current ipi behaviour isn't changed.

dtantsur: Then why condition it on `p.config.liveISOForcePersistentBootDevice != ""`? I think we should enforce the default value.

pkg/provisioner/ironic/ironic.go
```diff
@@ -698,6 +699,13 @@ func (p *ironicProvisioner) setLiveIsoUpdateOptsForNode(ironicNode *nodes.Node,
 	updater.
 		SetInstanceInfoOpts(optValues, ironicNode).
 		SetTopLevelOpt("deploy_interface", "ramdisk", ironicNode.DeployInterface)
+
+	if p.config.liveISOForcePersistentBootDevice != "" {
+		driverOptValues := optionsData{
+			"force_persistent_boot_device": p.config.liveISOForcePersistentBootDevice,
+		}
+		updater.SetDriverInfoOpts(driverOptValues, ironicNode)
+	}
```
dtantsur, line 708: You probably want an `else` branch with setting the default value or removing the existing value altogether.

pkg/provisioner/ironic/validatemanagementaccess_test.go
```diff
@@ -412,7 +416,7 @@ func TestValidateManagementAccessExistingSteadyStateNoUpdate(t *testing.T) {
 			}
 			assert.Equal(t, "", result.ErrorMessage)
 			assert.Equal(t, false, result.Dirty)
-			assert.Len(t, ironic.GetLastNodeUpdateRequestFor("uuid"), 0)
+			assert.Len(t, ironic.GetLastNodeUpdateRequestFor("uuid"), imageType.ExpectedUpdates)
```
derekhiggins, line 419: The existence of this test suggests that we explicitly don't want to update nodes in "ValidateManagementAccess",
yet here we are, is this ok?

hardys: I think probably to avoid this, we should add the `DriverInfo` (when creating the mock nodes in `testserver.NewIronic(t).Ready().CreateNode` above?

That way the node will already be in the expected state, e.g the steady-state-no-update case this test is trying to prove, and no PATCH update will be triggered?

derekhiggins: makes sense, fixed it, thanks

derekhiggins:
/test-integration


> toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit

derekhiggins:
> VolumeLimitExceeded: Maximum number of volumes allowed (20) exceeded for quota 'volumes'. (HTTP 413) (Request-ID: req-445bc040-7e05-45e3-8e3e-2b491028b6c0)

will try again later

PR #955, opened 2021-08-18
Update rbac api version to v1

kashifest:
k8s 1.22 doesn't support rbac.authorization.k8s.io/v1beta1, uplifting the version to v1 instead.


fmuyassarov (commented):
/lgtm

PR #963, opened 2021-09-02
Add kashifest and fmuyassarov as approvers

fmuyassarov:
Add @kashifest  and @fmuyassarov  as approvers in the following directories
```
/config
/ironic-deployment
/resources
/tools
```

### DIFF
--- a/config/OWNERS
+++ b/config/OWNERS
@@ -1,0 +1,5 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+approvers:
+ - fmuyassarov
+ - kashifest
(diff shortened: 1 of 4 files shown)


dtantsur:
/lgtm

Congrats!

dtantsur:
I assume we don't need to duplicate the main `OWNERS` in these files?

fmuyassarov:
> I assume we don't need to duplicate the main `OWNERS` in these files?

I think no, because root OWNERS takes privilege over all other OWNERS file I believe.

zaneb:
I'd have been inclined to put the  /resources/keepalived-docker up a level in /resources, but it's academic for now.
/approve

PR #964, opened 2021-09-02
convert CRD webhook to apiextensions.k8s.io/v1

kplimack:
The webhook for CRD conversion of the metal3 resources was still formatted for `apiextensions.k8s.io/v1beta1`, while we were expecting it in `
apiextensions.k8s.io/v1`.   When enabling the webhooks for BMO, this error arose:
```
error validating data: ValidationError(CustomResourceDefinition.spec.conversion): unknown field "webhookClientConfig" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceConversion
```

Signed-off-by: Jake Plimack <jplimack@tesla.com>

### DIFF
--- a/config/crd/kustomizeconfig.yaml
+++ b/config/crd/kustomizeconfig.yaml
@@ -5,12 +5,12 @@ nameReference:
   fieldSpecs:
   - kind: CustomResourceDefinition
     group: apiextensions.k8s.io
-    path: spec/conversion/webhookClientConfig/service/name
+    path: spec/conversion/webhook/clientConfig/service/name
 
 namespace:
 - kind: CustomResourceDefinition
   group: apiextensions.k8s.io
-  path: spec/conversion/webhookClientConfig/service/namespace
+  path: spec/conversion/webhook/clientConfig/service/namespace
   create: false
 
 varReference:
--- a/config/crd/patches/webhook_in_baremetalhosts.yaml
+++ b/config/crd/patches/webhook_in_baremetalhosts.yaml
@@ -7,11 +7,13 @@ metadata:
 spec:
   conversion:
     strategy: Webhook
-    webhookClientConfig:
-      # this is "\n" used as a placeholder, otherwise it will be rejected by the apiserver for being blank,
-      # but we're going to set it later using the cert-manager (or potentially a patch if not using cert-manager)
-      caBundle: Cg==
-      service:
-        namespace: system
-        name: webhook-service
-        path: /convert
+    webhook:
+      conversionReviewVersions: ["v1", "v1beta1"]
+      clientConfig:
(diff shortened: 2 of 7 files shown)


jplimack:
@kashifest I'm unclear why the tests are failing, can you help me out?

fmuyassarov:
> @kashifest / @fmuyassarov I'm unclear why the tests are failing, can you help me out?

I think you need to run `make generate`

kashifest:
> > @kashifest / @fmuyassarov I'm unclear why the tests are failing, can you help me out?
> 
> I think you need to run `make generate`

Yes precisely, you need to run `make generate` that would add the changes also in rendered `capm3.yaml` and that should make the CI happy

fmuyassarov:
@jplimack you can also run those unit tests and generate tests locally if you would like to verify them
```
export CONTAINER_RUNTIME=docker
./hack/unit.sh 
../hack/generate.sh
```

jplimack:
@fmuyassarov my unit tests pass when using `hack/unit.sh`, but when I run `make unit`, I arrive at the same `TestProvision` errors as prow
https://prow.apps.test.metal3.io/view/gs/metal3-prow/pr-logs/pull/metal3-io_baremetal-operator/964/unit/1435315738133925888

fmuyassarov:
> @fmuyassarov my unit tests pass when using `hack/unit.sh`, but when I run `make unit`, I arrive at the same `TestProvision` errors as prow
> https://prow.apps.test.metal3.io/view/gs/metal3-prow/pr-logs/pull/metal3-io_baremetal-operator/964/unit/1435315738133925888

I'm actually seeing failure even when running `hack/unit.sh' after checking out to your branch

zaneb (commented):
Does anybody know if there is a later version of kubebuilder we can update to so that we don't continue to create new problems like this?

zaneb:
> Does anybody know if there is a later version of kubebuilder we can update to so that we don't continue to create new problems like this?

Not until v3 [by the looks of it](https://github.com/kubernetes-sigs/kubebuilder/pull/1644). I don't know how feasible that update is right now.

fmuyassarov:
> > Does anybody know if there is a later version of kubebuilder we can update to so that we don't continue to create new problems like this?
> 
> Not until v3 [by the looks of it](https://github.com/kubernetes-sigs/kubebuilder/pull/1644). I don't know how feasible that update is right now.

Yep, v3 is what we need probably. Kubebuilder book has [migration instructions](migrate-the-webhooks) from v2 to v3. By quickly looking at the doc, it seems that migration should be straightforward. But I'm not sure if we want it now.

kashifest:
/hold

kashifest:
/hold cancel

PR #967, opened 2021-09-06
Fix comment for detached annotation and add StateAvailable 

furkatgofurov7:
Addresses small comment from [965](https://github.com/metal3-io/baremetal-operator/pull/965#discussion_r702077235). 

Upd.: Also, adds Hosts in StateAvialable to allow using a detached annotation.


kashifest:
lgtm

controllers/metal3.io/host_state_machine.go
```diff
@@ -254,7 +254,7 @@ func (hsm *hostStateMachine) checkDetachedHost(info *reconcileInfo) (result acti
 	// provisioner and take no further action
 	// Note this doesn't change the current state, only the OperationalStatus
 	if hasDetachedAnnotation(hsm.Host) {
-		// Only allow detaching hosts in Provisioned/ExternallyProvisioned states
+		// Only allow detaching hosts in Provisioned/ExternallyProvisioned/Ready states
 		switch info.host.Status.Provisioning.State {
 		case metal3v1alpha1.StateProvisioned, metal3v1alpha1.StateExternallyProvisioned, metal3v1alpha1.StateReady:
```
zaneb, line 259: Since we're here, we should probably include StateAvailable wherever we mention StateReady, although it feels like progress on #340 has stalled.

furkatgofurov7: Included `StateAvailable` everywhere as suggested by @zaneb, thank you. 
/cc @hardys

docs/api.md
```diff
@@ -574,7 +574,8 @@ changing power state, or deprovisioning), but still allows the BMH status
 to be updated unlike the `paused` anotation. While in this state the
 OperationalStatus field will be `detached` but the provisioning state will
 be unmodified.  This API only has any effect for BareMetalHost resources
-that are in either `Provisioned` or `ExternallyProvisioned` state.
+that are in either `Provisioned`, `ExternallyProvisioned`, `Ready` or `Available`
```
zaneb, line 577: I usually write this as Ready/Available, since they're the same state. In practice I don't think it's possible to get a Host into the Available state yet.

zaneb (commented):
/test-integration
/approve

docs/api.md
furkatgofurov7: Fixed. thank you

PR #968, opened 2021-09-07
Stop relying on -modfile to allow vendorization

honza:
The Makefile targets we use to build our tools rely on the `-modfile`
flag of the `go build` command.  Unfortunately, the golang build
toolchain doesn't respect the vendor directory relative to the `go.mod`
file specified by `-modfile`, and instead uses the vendor directory
relative to the current working directory.  Since each subdirectory in
this repository that contains a `go.mod` file needs to have its own
vendor directory in order to be built in vendorized scenarios, relying
on `-modfile` doesn't work.

The patch modifies the affected Makefile targets to instead change to
the directory of the tool being built instead of using the `-modfile`
flag.  This makes vendorization possible in downstream environments.

Finally, for consistency, we start building the kustomize tool in the
same way as controller-gen, and golangci-lint.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,8 @@ COVER_PROFILE = cover.out
 BIN_DIR := bin
 
 CRD_OPTIONS ?= "crd:trivialVersions=false,allowDangerousTypes=true,crdVersions=v1"
-KUSTOMIZE ?= go run -modfile=hack/tools/go.mod sigs.k8s.io/kustomize/kustomize/v3
+KUSTOMIZE = tools/bin/kustomize
+CONTROLLER_GEN = tools/bin/controller-gen
 
 # See pkg/version.go for details
 SOURCE_GIT_COMMIT ?= $(shell git rev-parse --short HEAD)
@@ -83,7 +84,7 @@ unit-verbose: ## Run unit tests with verbose output
 linters: lint generate-check fmt-check
 
 tools/bin/golangci-lint: hack/tools/go.mod
-	go build -o $@ -modfile=$< github.com/golangci/golangci-lint/cmd/golangci-lint
+	cd hack/tools; go build -o $(abspath $@) github.com/golangci/golangci-lint/cmd/golangci-lint
 
 .PHONY: lint
 lint: tools/bin/golangci-lint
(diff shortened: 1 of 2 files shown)


ardaguclu:
This is a good catch @honza, thanks. 
/lgtm

zaneb:
This is unfortunate :( the -modfile approach seemed quite elegant compared to switching directories.
I wonder if in future we should create a second Makefile so that we can do e.g. `make -C hack/tools controller-gen`

There is another use of `-modfile` in `tools/deploy.sh`. Does that also need changing? Otherwise this LGTM.

honza:
> There is another use of -modfile in tools/deploy.sh. Does that also need changing? 

Fixed

tools/deploy.sh
```diff
@@ -44,6 +44,10 @@ IRONIC_DEPLOY_FILES="${SCRIPTDIR}/ironic-deployment/basic-auth/default/auth.yaml
 	${SCRIPTDIR}/ironic-deployment/tls/keepalived/kustomization.yaml \
 	${SCRIPTDIR}/ironic-deployment/tls/keepalived/tls.yaml"
 
+if [ ! -f ${KUSTOMIZE} ]; then
```
zaneb, line 47: There's no need to make this conditional, checking whether the target is up to date is kinda what make does.

Makefile
```diff
@@ -117,23 +117,27 @@ run-test-mode: generate fmt-check lint manifests ## Run against the configured K
...
 
 .PHONY: deploy
-deploy: manifests ## Deploy controller in the configured Kubernetes cluster in ~/.kube/config
+deploy: $(KUSTOMIZE) manifests ## Deploy controller in the configured Kubernetes cluster in ~/.kube/config
 	cd config/manager && kustomize edit set image controller=${IMG}
 	$(KUSTOMIZE) build config/default | kubectl apply -f -
 
-tools/bin/controller-gen: hack/tools/go.mod
-	go build -o $@ -modfile=$< sigs.k8s.io/controller-tools/cmd/controller-gen
+tools/bin/controller-gen:
```
zaneb, line 132: I think it's wrong to remove the dependency here, because it means we will never recompile when the version changes.

Makefile
```diff
@@ -117,23 +117,27 @@ run-test-mode: generate fmt-check lint manifests ## Run against the configured K
...
+	cd hack/tools; go build -o ../../$@ sigs.k8s.io/controller-tools/cmd/controller-gen
+
+$(KUSTOMIZE):
```
zaneb, line 135: We should also depend on the go.mod file here for the same reason.

Makefile
```diff
@@ -117,23 +117,27 @@ run-test-mode: generate fmt-check lint manifests ## Run against the configured K
 	go run -ldflags $(LDFLAGS) ./main.go -namespace=$(RUN_NAMESPACE) -dev -test-mode -webhook-port=0
 
 .PHONY: install
-install: manifests ## Install CRDs into a cluster
+install: $(KUSTOMIZE) manifests ## Install CRDs into a cluster
 	$(KUSTOMIZE) build config/crd | kubectl apply -f -
```
zaneb, line 121: Could use `$<` in place of `$(KUSTOMIZE)` in all of these rules if we stop allowing people to replace it with their own binary.

Makefile
```diff
@@ -117,23 +117,27 @@ run-test-mode: generate fmt-check lint manifests ## Run against the configured K
...
 .PHONY: deploy
-deploy: manifests ## Deploy controller in the configured Kubernetes cluster in ~/.kube/config
+deploy: $(KUSTOMIZE) manifests ## Deploy controller in the configured Kubernetes cluster in ~/.kube/config
 	cd config/manager && kustomize edit set image controller=${IMG}
 	$(KUSTOMIZE) build config/default | kubectl apply -f -
 
-tools/bin/controller-gen: hack/tools/go.mod
-	go build -o $@ -modfile=$< sigs.k8s.io/controller-tools/cmd/controller-gen
+tools/bin/controller-gen:
+	cd hack/tools; go build -o ../../$@ sigs.k8s.io/controller-tools/cmd/controller-gen
```
zaneb, line 133: I think we could possibly do `$(abspath $@)` here to avoid all the relative path stuff, which would change if we ever moved the output.
There a similar opportunity with `../$<` in the manifests & generate rules.

Makefile
```diff
@@ -18,7 +18,7 @@ COVER_PROFILE = cover.out
 BIN_DIR := bin
 
 CRD_OPTIONS ?= "crd:trivialVersions=false,allowDangerousTypes=true,crdVersions=v1"
-KUSTOMIZE ?= go run -modfile=hack/tools/go.mod sigs.k8s.io/kustomize/kustomize/v3
+KUSTOMIZE ?= tools/bin/kustomize
```
zaneb, line 21: We're not being that consistent here by defining a variable for kustomize but not for controller-gen. I'm not sure how people were using this variable to override kustomize before, but I suspect that it will not work as expected with these changes. It's probably better to just not support customising it.
I think I'd be more tempted to only define `TOOLS_BIN_DIR = tools/bin`.

zaneb (commented):
/lgtm

Makefile
honza: Is there an elegant way of specifying the second dependency? All I found is `$(word 2,$^)`, and it's ugly.

zaneb: Yeah, you're correct, there's nothing better than that. The only elegant solution would be to split that rule into two (or three).

PR #981, opened 2021-09-15
Avoid logging BMC password when creds change

### LINKED ISSUE #980: BMC passwords can sometimes be logged
Since b8dba2c016befacf2ff65004a2ff73008d5ab63a we use the nodeUpdater with its built-in logging to update the ironic node when the credentials change. This inadvertantly meant the BMC credentials will be logged in this case.

This should be very rare in practice; BMC credentials don't often change. However, it should be fixed.

zaneb:
Since b8dba2c016befacf2ff65004a2ff73008d5ab63a we use the nodeUpdater
with its built-in logging to update the ironic node when the credentials
change. This inadvertantly meant the BMC credentials will be logged in
this case (which is very rare; credentials don't often change).

To prevent this, sanitise new values as they are being logged. Ironic
does not return existing values for passwords, so there is no danger of
them being logged through that route. Currently password fields are only
redacted if they are set as part of a map; in practice this will be safe
for the foreseeable future as the bmc credentials are always set through
a map returned from the driver and not individually.

Fixes #980

### DIFF
--- a/pkg/provisioner/ironic/update_opts.go
+++ b/pkg/provisioner/ironic/update_opts.go
@@ -3,6 +3,7 @@ package ironic
 import (
 	"fmt"
 	"reflect"
+	"strings"
 
 	"github.com/go-logr/logr"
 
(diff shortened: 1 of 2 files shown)


pkg/provisioner/ironic/update_opts.go
```diff
@@ -72,6 +73,27 @@ func deref(v interface{}) interface{} {
...
+		dataType.Key().Kind() != reflect.String {
+		return data
+	}
+
+	value := reflect.ValueOf(data)
+	safeValue := reflect.MakeMap(dataType)
+
+	for _, k := range value.MapKeys() {
+		safeDatumValue := value.MapIndex(k)
+		if strings.Contains(k.String(), "password") {
```
andfasano, line 88: Just to be on the safe side, the comparison could be case insensitive

zaneb: This covers all of the values that any of the bmc drivers use.

andfasano: I was referring to the keys, which IINW are defined in the code upper func calls (like `SetPropertiesOpts`). So maybe someone could define a `Password` prop by mistake (not sure if Ironic is also strict on the case) which will not get sanitised

andfasano: Probably a `strings.ToLower(k.String())` wouldn't hurt, but if Ironic will catch it (ie during the integration tests) then it'd be ok

/lgtm

PR #985, opened 2021-09-20
Fix fallback for ironic drivers that don't support soft power off

### LINKED ISSUE #984: Power off fails for drivers that don't support Soft power off
When an ironic driver doesn't support soft power off, we are supposed to fall back to doing a hard power off. This was inadvertently broken by #841. Now when the driver doesn't support soft power off we end up returning a 'transient' error and retrying in an infinite loop.

zaneb:
Fix a regression in #841 that caused ironic drivers that don't support soft power off (such as Fujitsu when the agent is not available) to fail in an infinite loop instead of falling back to a hard power off.

Also clean up the code so it is much less arcane.

Fixes #984


sadasu (commented):
/lgtm

I see the regression in the previous changes. Thanks for the fix.

pkg/provisioner/ironic/power_test.go
```diff
@@ -295,7 +298,52 @@ func TestPowerOff(t *testing.T) {
...
+	}
+	auth := clients.AuthConfig{Type: clients.NoAuth}
+	prov, err := newProvisionerWithSettings(host, bmc.Credentials{}, publisher,
+		ironic.Endpoint(), auth, inspector.Endpoint(), auth,
+	)
+	if err != nil {
+		t.Fatalf("could not create provisioner: %s", err)
+	}
+
+	_, err = prov.PowerOff(metal3v1alpha1.RebootModeSoft, false)
```
andfasano, line 338: Unfortunately the current mock server does not allow to specify multiple, different responses for the same API call, so I don't think there is an easy way currently to test such scenario without enhancing properly the mock (right now, when a response call is configured, like here in `.WithNodeStatesPowerUpdate(nodeUUID, http.StatusBadRequest)` for example, the mock server will _always_ provide the same answer for all the requests).
I think we could discuss this change in a separate PR as it may be useful also for other tests.

andfasano (commented):
/approve

pkg/provisioner/ironic/power_test.go
zaneb: Yeah, I agree that would be a helpful feature to have in the mocks. If that were possible then this could have been implemented in TestPowerOff (and the regression would not have happened) without having to write this separate TestSoftPowerOffFallback test with its own logic.

PR #987, opened 2021-09-23
Fix non-existing fmt target usage

fmuyassarov:
This fixed non-existing `fmt` target in the Makefile. Also, updates `fmt-check` target to actually do the formatting.
Currently, if there is formatting issue running `./hack/gofmt.sh` would print
```
fmuyassarov@est:~/go/src/github/baremetal-operator$ ./hack/gofmt.sh 
+ IS_CONTAINER=false
+ CONTAINER_RUNTIME=docker
+ [ false != false ]
+ docker run --rm --env IS_CONTAINER=TRUE --volume /home/fmuyassarov/go/src/github/baremetal-operator:/workdir:rw,z --entrypoint sh --workdir /workdir registry.hub.docker.com/library/golang:1.16 /workdir/hack/gofmt.sh
+ IS_CONTAINER=TRUE
+ CONTAINER_RUNTIME=podman
+ [ TRUE != false ]
+ TOP_DIR=.
+ export XDG_CACHE_HOME=/tmp/.cache
+ cd .
+ gofmt -l .
+ [ -n apis/metal3.io/v1alpha1/baremetalhost_types.go ]
+ make fmt
make: *** No rule to make target 'fmt'.  Stop.
```

and after this patch, 
```
fmuyassarov@est:~/go/src/github/baremetal-operator$ ./hack/gofmt.sh 
+ IS_CONTAINER=false
+ CONTAINER_RUNTIME=docker
+ [ false != false ]
+ docker run --rm --env IS_CONTAINER=TRUE --volume /home/fmuyassarov/go/src/github/baremetal-operator:/workdir:rw,z --entrypoint sh --workdir /workdir registry.hub.docker.com/library/golang:1.16 /workdir/hack/gofmt.sh
+ IS_CONTAINER=TRUE
+ CONTAINER_RUNTIME=podman
+ [ TRUE != false ]
+ TOP_DIR=.
+ export XDG_CACHE_HOME=/tmp/.cache
+ cd .
+ gofmt -l .
+ [ -n apis/metal3.io/v1alpha1/baremetalhost_types.go ]
+ echo Please run make fmt-check to fix formatting issues
+ exit 1
Please run make fmt-check to fix formatting issues
```
and I'm informed that I need to run make `fmt-check` to fix formatting.

### DIFF
--- a/hack/gofmt.sh
+++ b/hack/gofmt.sh
@@ -13,7 +13,7 @@ if [ "${IS_CONTAINER}" != "false" ]; then
 
   # shellcheck disable=SC2086
   if [ -n "$(gofmt -l .)" ]; then
-      make fmt
+      gofmt -s -d -e .
       exit 1
   fi
 else
--- a/Makefile
+++ b/Makefile
@@ -182,6 +182,10 @@ generate-check-local:
 fmt-check: ## Run gofmt and report an error if any changes are made
 	./hack/gofmt.sh
 
+.PHONY: fmt
+fmt: ## Run gofmt and fix files with formatting issues
+	gofmt -s -w .
+
 ## --------------------------------------
 ## Documentation
 ## --------------------------------------


fmuyassarov:
I accidentally spotted it after https://kubernetes.slack.com/archives/CHD49TLE7/p1632318260088000

namnx228 (commented):
LGTM

fmuyassarov:
> /test-integration

I think CI failure is not about this patch but the issue which is being fixed in https://github.com/metal3-io/metal3-dev-env/pull/806. So, I will rerun the test again once 806 is merged.

hack/gofmt.sh
```diff
@@ -13,7 +13,7 @@ if [ "${IS_CONTAINER}" != "false" ]; then
 
   # shellcheck disable=SC2086
   if [ -n "$(gofmt -l .)" ]; then
-      make fmt
+      echo "Please run make fmt-check to fix formatting issues"
```
zaneb, line 16: More useful output for CI might be something like e.g. `gofmt -s -d -e .`

zaneb (commented):
This is exactly backwards from how I'd expect it to behave - `make gofmt-check` reformats but does not check, while `hack/gofmt.sh` checks but does not reformat.

Why not just define the missing `fmt` target?

hack/gofmt.sh
```diff
@@ -13,7 +13,7 @@ if [ "${IS_CONTAINER}" != "false" ]; then
 
   # shellcheck disable=SC2086
   if [ -n "$(gofmt -l .)" ]; then
-      make fmt
+      gofmt -s -d -e .
```
fmuyassarov, line 16: I don't think we need `make fmt` here, because if there is an issue, and when linter Ci is triggered, it will fix the formatting issues too, but it will just happen on the temporary pod, not on the actually patch. So, it is responsibility of the patch author to run `make fmt`.

PR #988, opened 2021-09-29
Remove ironic-inspector-log-watch

kashifest:
This is a leftover work of #945 which has removed ironic-inspector-log-watch.


smoshiur1237:
LGTM

PR #994, opened 2021-10-13
Add RequiresProvisioningNetwork check into AccessDetails

ardaguclu:
This PR adds new function namely RequiresProvisioningNetwork into
AccessDetails. Thanks to that, when provisioning network is disabled
or not, it becomes possible to check which BMCs support this.

### DIFF
--- a/pkg/ironic/bmc/access.go
+++ b/pkg/ironic/bmc/access.go
@@ -70,6 +70,9 @@ type AccessDetails interface {
 	// Whether the driver supports booting a preprovisioning image in ISO format
 	SupportsISOPreprovisioningImage() bool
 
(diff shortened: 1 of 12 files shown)


dtantsur:
Why not add the check itself in the same change? Otherwise looks good.

ardaguclu:
> > Or there is another place additionally I should add?
> 
> I think it's reasonable to add the check to BMO itself as well (in openshift case for workers). See e.g. for boot mode: https://github.com/metal3-io/baremetal-operator/blob/master/pkg/provisioner/ironic/ironic.go#L363-L368

@dtantsur I'll do that check for `ProvisioningNetworkDisabled` cases. In upstream, is it possible to get info?

PR #995, opened 2021-10-14
Change hfs controller to be level-triggered when handling bmh updates

bfournie:
Change to use a poll cycle in the hfs reconciler to get data from Ironic instead of relying on state difference. In addition, the following changes were made to address follow-on comments in 
https://github.com/metal3-io/baremetal-operator/pull/938/.

- changed name of condition `UpdateRequested` to `FirmwareSettingsChangeDetected`
- add each hostFirmwareSettings as owner of its FirmwareSchema so that garbage collection will only be done when all removed
- Change the SetOwnerReference/SetControllerReference calls to before Create so the Update can be removed
- When building BIOS settings for cleaning make sure the hfs settings don't overwrite the vendor-agnostic settings
- When generating the hash for the firmwareSchema include the schema values in addition to the names

### DIFF
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -56,6 +56,7 @@ const (
 	unmanagedRetryDelay           = time.Minute * 10
 	preprovImageRetryDelay        = time.Minute * 5
 	provisionerNotReadyRetryDelay = time.Second * 30
+	subResourceNotReadyRetryDelay = time.Second * 60
 	rebootAnnotationPrefix        = "reboot.metal3.io"
 	inspectAnnotationPrefix       = "inspect.metal3.io"
 	hardwareDetailsAnnotation     = inspectAnnotationPrefix + "/hardwaredetails"
@@ -808,6 +809,14 @@ func (r *BareMetalHostReconciler) registerHost(prov provisioner.Provisioner, inf
 		return result
 	}
 
+	// Create the hostFirmwareSettings resource with same host name/namespace if it doesn't exist
+	if info.host.Name != "" {
+		if err = r.createHostFirmwareSettings(info); err != nil {
+			info.log.Info("failed creating hostfirmwaresettings")
+			return actionError{errors.Wrap(err, "failed creating hostFirmwareSettings")}
+		}
+	}
+
 	// Reaching this point means the credentials are valid and worked,
 	// so clear any previous error and record the success in the
 	// status block.
@@ -944,7 +953,8 @@ func (r *BareMetalHostReconciler) actionPreparing(prov provisioner.Provisioner, 
 	// Use settings in hostFirmwareSettings if available
 	hfs, err := r.getHostFirmwareSettings(info)
 	if err != nil {
-		info.log.Info("hostFirmwareSettings not available for cleaning")
+		// wait until hostFirmwareSettings are ready
+		return actionContinue{subResourceNotReadyRetryDelay}
 	}
 	if hfs != nil {
 		prepareData.ActualFirmwareSettings = hfs.Status.Settings.DeepCopy()
@@ -1318,6 +1328,38 @@ func saveHostProvisioningSettings(host *metal3v1alpha1.BareMetalHost) (dirty boo
 	return
 }
 
+func (r *BareMetalHostReconciler) createHostFirmwareSettings(info *reconcileInfo) error {
+
+	// Check if HostFirmwareSettings already exists
+	hfs := &metal3v1alpha1.HostFirmwareSettings{}
+	if err := r.Get(context.TODO(), info.request.NamespacedName, hfs); err != nil {
+		if k8serrors.IsNotFound(err) {
+			// A resource doesn't exist, create one
+			hfs.ObjectMeta = metav1.ObjectMeta{
+				Name:      info.host.Name,
+				Namespace: info.host.Namespace}
+			hfs.Status.Settings = make(metal3v1alpha1.SettingsMap)
+			hfs.Spec.Settings = make(metal3v1alpha1.DesiredSettingsMap)
+
+			// Set bmh as owner, this makes sure the resource is deleted when bmh is deleted
+			if err = controllerutil.SetControllerReference(info.host, hfs, r.Scheme()); err != nil {
+				return errors.Wrap(err, "could not set bmh as controller")
+			}
+			if err = r.Create(context.TODO(), hfs); err != nil {
+				return errors.Wrap(err, "failure creating hostFirmwareSettings resource")
+			}
+
+			info.log.Info("created new hostFirmwareSettings resource")
+
+		} else {
+			// Error reading the object
+			return errors.Wrap(err, "could not load hostFirmwareSettings resource")
+		}
+	}
+
+	return nil
+}
+
 // Get the stored firmware settings if there are valid changes
 func (r *BareMetalHostReconciler) getHostFirmwareSettings(info *reconcileInfo) (hfs *metal3v1alpha1.HostFirmwareSettings, err error) {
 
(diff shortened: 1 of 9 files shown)


andfasano:
After reviewing all the changes, on a second thought I'm getting convinced that probably the relationships between BMH and HFS could be relaxed a little bit, possibly leading to a simplified approach that will honor all the points previously discussed. Following the main general points:

1. BMH needs only to wait, in the Preparing state, for the related HFS resource to be ready (and here using the conditions looks fine). If the HFS is not ready, it will re-enqueue a reconciliation request before proceeding. If it is, the cleaning steps could be taken as usual.
2. HFS controller will stop listening BMH events, and it should take care only of filling (and validating) the HFS resource by fetching the hw settings at every reconcile loop, ensuring that its Conditions properly reflect the current situation. The reconciliation will be by default reenqueued on a reasonable time basis (ie 5 minutes, to be decided), to ensure that any external changes to the hw settings will be captured (for example, if any setting will be changed after the cleaning step in the Preparing state).
3. Given that, the responsibility to create an (empty) HFS resource if not present will come back under the BMH controller (something to be performed at every reconcile loop, if not under deletion)

I understand that this means a significant change to the current design, but the overall idea is to look for a simplified approach that maybe could be reused also in future for other similar situations.

Wdyt?

bfournie:
> After reviewing all the changes, on a second thought I'm getting convinced that probably the relationships between BMH and HFS could be relaxed a little bit, possibly leading to a simplified approach that will honor all the points previously discussed. Following the main general points:
> 
> 1. BMH needs only to wait, in the Preparing state, for the related HFS resource to be ready (and here using the conditions looks fine). If the HFS is not ready, it will re-enqueue a reconciliation request before proceeding. If it is, the cleaning steps could be taken as usual.
> 2. HFS controller will stop listening BMH events, and it should take care only of filling (and validating) the HFS resource by fetching the hw settings at every reconcile loop, ensuring that its Conditions properly reflect the current situation. The reconciliation will be by default reenqueued on a reasonable time basis (ie 5 minutes, to be decided), to ensure that any external changes to the hw settings will be captured (for example, if any setting will be changed after the cleaning step in the Preparing state).
> 3. Given that, the responsibility to create an (empty) HFS resource if not present will come back under the BMH controller (something to be performed at every reconcile loop, if not under deletion)
> 
> I understand that this means a significant change to the current design, but the overall idea is to look for a simplified approach that maybe could be reused also in future for other similar situations.
> 
> Wdyt?

I think its a good idea and will simplify the hfs controller. I've updated the PR to:
1. Remove the watch of bmh resource in hfs controller 
2. Remove the hfs EventUpdate handler filtering
3. Create the hfs resource in bmh controller when registering
4. Add a requeue to the hfs reconciler to poll the data from Ironic and pick up any changes as a result of cleaning (or other events)
5. In bmh controller, before entering the preparing state check that the hfs is ready

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -944,7 +952,8 @@ func (r *BareMetalHostReconciler) actionPreparing(prov provisioner.Provisioner,
 	// Use settings in hostFirmwareSettings if available
 	hfs, err := r.getHostFirmwareSettings(info)
 	if err != nil {
-		info.log.Info("hostFirmwareSettings not available for cleaning")
+		// wait until hostFirmwareSettings are ready
+		return actionDelayed{}
```
andfasano, line 956: Just realized that this `actionDelayed` was primarily meant to be used in the capacity check, so I'd say that an `actionContinue{subResourceNotAvailable}` could be a better match for this case (ie, with a const `subResourceNotAvailable = time.Second * 60`)

controllers/metal3.io/hostfirmwaresettings_controller.go
```diff
@@ -123,16 +124,12 @@ func (r *HostFirmwareSettingsReconciler) Reconcile(ctx context.Context, req ctrl
...
 		if k8serrors.IsNotFound(err) {
-			// A resource doesn't exist, create one
-			if err = r.newHostFirmwareSettings(info, req.NamespacedName); err != nil {
-
-				return ctrl.Result{}, errors.Wrap(err, "could not create firmware settings")
-			}
-		} else {
-			// Error reading the object - requeue the request.
-			return ctrl.Result{}, errors.Wrap(err, "could not load host firmware settings")
+			return ctrl.Result{}, errors.Wrap(err, "could not create firmware settings")
```
andfasano, line 129: Shouldn't be the error message something like `not found` (given that the creation has been moved within the BMH controller)? I'd expect to end up here if someone deletes the HFS resource before the next reconciliation loop is triggered. In such case it should be sufficient to wait enough so that the BMH controller will re-create it

controllers/metal3.io/baremetalhost_controller.go
bfournie: Good point, changed.

controllers/metal3.io/hostfirmwaresettings_controller.go
bfournie: Yeah that error message needs updating, fixed.

controllers/metal3.io/hostfirmwaresettings_controller.go
```diff
@@ -123,16 +124,12 @@ func (r *HostFirmwareSettingsReconciler) Reconcile(ctx context.Context, req ctrl
...
 		if k8serrors.IsNotFound(err) {
-			// A resource doesn't exist, create one
-			if err = r.newHostFirmwareSettings(info, req.NamespacedName); err != nil {
-
-				return ctrl.Result{}, errors.Wrap(err, "could not create firmware settings")
-			}
-		} else {
-			// Error reading the object - requeue the request.
-			return ctrl.Result{}, errors.Wrap(err, "could not load host firmware settings")
+			return ctrl.Result{}, errors.Wrap(err, "hostFirmwareSettings not found")
```
andfasano, line 129: Note that returning an error will cause an immediate reconciliation (rate limited). Considering that we'll have to wait the next BMH reconciliation loop for recreating the missing resource, maybe this case could be treated not as an error but as a waiting condition (with `RequeueAfter` set to a reasonable value, ie `provisionerRetryDelay`)

andfasano (approved):
Thanks @bfournie for the patch, the code looks simpler and the responsibilities across the controllers well assigned, addressing the points previously discussed.

/approve

controllers/metal3.io/hostfirmwaresettings_controller.go
bfournie: Changed to return requeue time

controllers/metal3.io/hostfirmwaresettings_controller.go
```diff
@@ -196,14 +193,35 @@ func (r *HostFirmwareSettingsReconciler) updateStatus(info *rInfo, settings meta
...
-	// Update Status
+	// Update Status on changes
 	for k, v := range settings {
 		// Some vendors include encrypted password fields, don't add these
 		if strings.Contains(k, "Password") {
 			continue
 		}
-		info.hfs.Status.Settings[k] = v
+
+		// If this setting has ReadOnly only or Unique schema flags then store it separately
```
dtantsur, line 207: Why treat Unique this way? Unique fields can be modified, no?

bfournie: This change came about from Zane's earlier comment that we should make it easier for a user to copy all of hosts's BIOS settings to another host to ensure they are configured the same way. By segregating the ReadOnly fields and fields that are unique to a host into the LocalData field then the Status.Settings can be easily copied to another host and no filtering would be needed.
BUT - you found a bug in that currently Unique fields are prevented from being set since only Status.Settings is checked for validity. Pushing an update with fix and test.

bfournie: I've changed it back to the original design to not have Unique and ReadOnly fields in a separate slice, it was just introducing too much complexity having these broken out separately.

elfosardo (approved):

Issue #996, opened 2021-10-14
Fail to import package: github.com/metal3-io/baremetal-operator/ironic


hs0210:
I failed to import package `github.com/metal3-io/baremetal-operator/ironic` in terraform-provider-ironic.
import:
```diff
diff --git a/ironic/resource_ironic_node_v1.go b/ironic/resource_ironic_node_v1.go
index 7823f5ec..b434eaec 100644
--- a/ironic/resource_ironic_node_v1.go
+++ b/ironic/resource_ironic_node_v1.go
@@ -9,6 +9,7 @@ import (
        "github.com/gophercloud/gophercloud/openstack/baremetal/v1/nodes"
        "github.com/gophercloud/gophercloud/openstack/baremetal/v1/ports"
        "github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+       _ "github.com/metal3-io/baremetal-operator/ironic"
 )
```
error:
```
$ go mod tidy
go: finding module for package github.com/metal3-io/baremetal-operator/ironic
github.com/openshift-metal3/terraform-provider-ironic/ironic imports
        github.com/metal3-io/baremetal-operator/ironic: module github.com/metal3-io/baremetal-operator@latest found (v0.0.0-20211013091910-8feb172f0068), but does not contain package github.com/metal3-io/baremetal-operator/ironic
```     

Then I tried use replace, but failed too.
import:
```diff
diff --git a/go.mod b/go.mod
index b6b33a21..6e9e7237 100644
--- a/go.mod
+++ b/go.mod
@@ -20,4 +20,7 @@ require (
        golang.org/x/lint v0.0.0-20210508222113-6edffad5e616 // indirect
        golang.org/x/net v0.0.0-20191126235420-ef20fe5d7933 // indirect
        gopkg.in/check.v1 v1.0.0-20190902080502-41f04d3bba15 // indirect
+       github.com/metal3-io/baremetal-operator/ironic v0.0.0
 )
+
+replace github.com/metal3-io/baremetal-operator/ironic => github.com/metal3-io/baremetal-operator/ironic v0.0.0-20211013091910-8feb172f0068
diff --git a/ironic/resource_ironic_node_v1.go b/ironic/resource_ironic_node_v1.go
index 7823f5ec..b434eaec 100644
--- a/ironic/resource_ironic_node_v1.go
+++ b/ironic/resource_ironic_node_v1.go
@@ -9,6 +9,7 @@ import (
[...]

PR #998, opened 2021-10-14
Rename ironic go module to correct path for go get can find

ardaguclu:
`go get github.com/metal3-io/baremetal-operator/ironic` fails because
it can  not find new ironic module which lies under the pkg directory.
This PR fixes this bug by renaming ironic module to be parallel with folder
it lies.

LICENSE file is also added for GOPROXY url
can cache these modules in pkg.go.dev.

Fixes:  https://github.com/metal3-io/baremetal-operator/issues/996

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/go-logr/logr v0.4.0
 	github.com/gophercloud/gophercloud v0.18.0
 	github.com/metal3-io/baremetal-operator/apis v0.0.0
-	github.com/metal3-io/baremetal-operator/ironic v0.0.0
+	github.com/metal3-io/baremetal-operator/pkg/ironic v0.0.0
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.11.0
 	github.com/stretchr/testify v1.7.0
@@ -20,4 +20,4 @@ require (
 
 replace github.com/metal3-io/baremetal-operator/apis => ./apis
 
-replace github.com/metal3-io/baremetal-operator/ironic => ./pkg/ironic
+replace github.com/metal3-io/baremetal-operator/pkg/ironic => ./pkg/ironic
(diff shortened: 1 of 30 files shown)


s3rj1k:
@ardaguclu Please see https://github.com/metal3-io/baremetal-operator/issues/996#issuecomment-943262303
before moving code again

ardaguclu:
> @ardaguclu Please see [#996 (comment)](https://github.com/metal3-io/baremetal-operator/issues/996#issuecomment-943262303) before moving code again.
> 
> I'd prefer to avoid moving all of the code around, if possible.

Do you suggest renaming module name(by adding `pkg/`) instead moving folder?

go.mod
```diff
@@ -20,4 +20,4 @@ require (
 
 replace github.com/metal3-io/baremetal-operator/apis => ./apis
 
-replace github.com/metal3-io/baremetal-operator/ironic => ./pkg/ironic
+replace github.com/metal3-io/baremetal-operator/pkg/ironic => ./pkg/ironic
```
dtantsur, line 23: Is it really necessary any more?

dtantsur (commented):
/approve

go.mod
s3rj1k: @dtantsur yep, if we change something in ./pkg/ironic inside local change, without rewrite rule gomod will try to fetch from remote location, and code would be without newer change

s3rj1k: we should keep that rewrite

ardaguclu: Yes, to be compatible with `replace github.com/metal3-io/baremetal-operator/apis => ./apis` usage. I'm not sure, but if we don't do that local changes in ironic package can not be seen in main module.

s3rj1k:
> > @dtantsur Merge maybe?
> 
> @s3rj1k it is approved. I think, if you lgtm, this PR will automatically be merged.

I do not have merge privileges sadly

PR #1000, opened 2021-10-18
Add BareMetalHost validation

zhouhao3:
This PR adds some BareMetalHost validations as issued in #855.

* Invalid BMC Config
* RAID requested when not supported by driver
* Missing Boot MAC when required by driver
* BIOS requested when not supported by driver
* Add validations for BootMode

### DIFF
--- a/apis/metal3.io/v1alpha1/baremetalhost_validation.go
+++ b/apis/metal3.io/v1alpha1/baremetalhost_validation.go
@@ -17,10 +17,11 @@ var log = logf.Log.WithName("baremetalhost-validation")
 func (host *BareMetalHost) validateHost() []error {
 	log.Info("validate create", "name", host.Name)
 	var errs []error
+	var bmcAccess bmc.AccessDetails
 
 	if host.Spec.BMC.Address != "" {
 		var err error
-		_, err = bmc.NewAccessDetails(host.Spec.BMC.Address, host.Spec.BMC.DisableCertificateVerification)
+		bmcAccess, err = bmc.NewAccessDetails(host.Spec.BMC.Address, host.Spec.BMC.DisableCertificateVerification)
 		if err != nil {
 			errs = append(errs, err)
 		}
@@ -29,6 +30,8 @@ func (host *BareMetalHost) validateHost() []error {
 	if err := validateRAID(host.Spec.RAID); err != nil {
 		errs = append(errs, err)
 	}
+
+	errs = append(errs, validateBMCAccess(host.Spec, bmcAccess)...)
 
 	if err := validateBMHName(host.Name); err != nil {
 		errs = append(errs, err)
@@ -53,6 +56,22 @@ func (host *BareMetalHost) validateChanges(old *BareMetalHost) []error {
 
 	if old.Spec.BootMACAddress != "" && host.Spec.BootMACAddress != old.Spec.BootMACAddress {
 		errs = append(errs, fmt.Errorf("bootMACAddress can not be changed once it is set"))
+	}
+
+	return errs
+}
+
+func validateBMCAccess(s BareMetalHostSpec, bmcAccess bmc.AccessDetails) []error {
+	var errs []error
+
+	if bmcAccess == nil {
+		return errs
+	}
+
+	if s.RAID != nil && len(s.RAID.HardwareRAIDVolumes) > 0 {
+		if bmcAccess.RAIDInterface() == "no-raid" {
+			errs = append(errs, fmt.Errorf("BMC driver %s does not support configuring RAID", bmcAccess.Type()))
+		}
 	}
 
 	return errs
--- a/apis/metal3.io/v1alpha1/baremetalhost_validation.go
+++ b/apis/metal3.io/v1alpha1/baremetalhost_validation.go
@@ -80,6 +80,10 @@ func validateBMCAccess(s BareMetalHostSpec, bmcAccess bmc.AccessDetails) []error
 		}
 	}
 
+	if bmcAccess.NeedsMAC() && s.BootMACAddress == "" {
+		errs = append(errs, fmt.Errorf("BMC driver %s requires a BootMACAddress value", bmcAccess.Type()))
+	}
+
 	return errs
 }
 
--- a/apis/metal3.io/v1alpha1/baremetalhost_validation.go
+++ b/apis/metal3.io/v1alpha1/baremetalhost_validation.go
@@ -74,6 +74,12 @@ func validateBMCAccess(s BareMetalHostSpec, bmcAccess bmc.AccessDetails) []error
 		}
 	}
 
+	if s.Firmware != nil {
+		if _, err := bmcAccess.BuildBIOSSettings((*bmc.FirmwareConfig)(s.Firmware)); err != nil {
+			errs = append(errs, err)
+		}
+	}
+
 	return errs
 }
 
--- a/apis/metal3.io/v1alpha1/baremetalhost_validation.go
+++ b/apis/metal3.io/v1alpha1/baremetalhost_validation.go
@@ -7,7 +7,7 @@ import (
 	"github.com/google/uuid"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 
-	_ "github.com/metal3-io/baremetal-operator/pkg/hardwareutils/bmc"
+	"github.com/metal3-io/baremetal-operator/pkg/hardwareutils/bmc"
 )
 
 // log is for logging in this package.
@@ -17,6 +17,14 @@ var log = logf.Log.WithName("baremetalhost-validation")
 func (host *BareMetalHost) validateHost() []error {
 	log.Info("validate create", "name", host.Name)
 	var errs []error
+
+	if host.Spec.BMC.Address != "" {
+		var err error
+		_, err = bmc.NewAccessDetails(host.Spec.BMC.Address, host.Spec.BMC.DisableCertificateVerification)
+		if err != nil {
+			errs = append(errs, err)
+		}
+	}
 
 	if err := validateRAID(host.Spec.RAID); err != nil {
 		errs = append(errs, err)
--- a/apis/metal3.io/v1alpha1/baremetalhost_validation.go
+++ b/apis/metal3.io/v1alpha1/baremetalhost_validation.go
@@ -84,6 +84,10 @@ func validateBMCAccess(s BareMetalHostSpec, bmcAccess bmc.AccessDetails) []error
 		errs = append(errs, fmt.Errorf("BMC driver %s requires a BootMACAddress value", bmcAccess.Type()))
 	}
 
+	if s.BootMode == UEFISecureBoot && !bmcAccess.SupportsSecureBoot() {
+		errs = append(errs, fmt.Errorf("BMC driver %s does not support secure boot", bmcAccess.Type()))
+	}
+
 	return errs
 }
 
--- a/apis/metal3.io/v1alpha1/baremetalhost_validation_test.go
+++ b/apis/metal3.io/v1alpha1/baremetalhost_validation_test.go
@@ -107,6 +107,31 @@ func TestValidateCreate(t *testing.T) {
 			oldBMH:    nil,
 			wantedErr: "Unknown BMC type 'test' for address test:127.0.1.1",
 		},
+		{
+			name: "RAIDWithUnsupportBMC",
+			newBMH: &BareMetalHost{
+				TypeMeta:   tm,
+				ObjectMeta: om,
+				Spec: BareMetalHostSpec{
+					RAID: &RAIDConfig{
+						HardwareRAIDVolumes: []HardwareRAIDVolume{
+							{
+								SizeGibibytes:         nil,
+								Level:                 "",
+								Name:                  "",
+								Rotational:            nil,
+								NumberOfPhysicalDisks: nil,
+							},
+						},
+					},
+					BMC: BMCDetails{
+						Address:         "ipmi://127.0.1.1",
+						CredentialsName: "test1",
+					},
+				}},
+			oldBMH:    nil,
+			wantedErr: "BMC driver ipmi does not support configuring RAID",
+		},
 	}
 
 	for _, tt := range tests {
--- a/apis/metal3.io/v1alpha1/baremetalhost_validation_test.go
+++ b/apis/metal3.io/v1alpha1/baremetalhost_validation_test.go
@@ -73,6 +73,11 @@ func TestValidateCreate(t *testing.T) {
 				TypeMeta:   tm,
 				ObjectMeta: om,
 				Spec: BareMetalHostSpec{
+					BootMACAddress: "01:02:03:04:05:06",
+					BMC: BMCDetails{
+						Address:         "irmc:127.0.1.1",
+						CredentialsName: "test1",
+					},
 					RAID: &RAIDConfig{
 						HardwareRAIDVolumes: []HardwareRAIDVolume{
 							{
@@ -95,6 +100,21 @@ func TestValidateCreate(t *testing.T) {
 			wantedErr: "hardwareRAIDVolumes and softwareRAIDVolumes can not be set at the same time",
 		},
 		{
+			name: "supportBMCType",
+			newBMH: &BareMetalHost{
+				TypeMeta:   tm,
+				ObjectMeta: om,
+				Spec: BareMetalHostSpec{
+					BootMACAddress: "01:02:03:04:05:06",
+					BMC: BMCDetails{
+						Address:         "irmc:127.0.1.1",
+						CredentialsName: "test1",
+					},
+				}},
+			oldBMH:    nil,
+			wantedErr: "",
+		},
+		{
 			name: "unsupportBMCType",
 			newBMH: &BareMetalHost{
 				TypeMeta:   tm,
@@ -107,6 +127,32 @@ func TestValidateCreate(t *testing.T) {
 				}},
 			oldBMH:    nil,
 			wantedErr: "Unknown BMC type 'test' for address test:127.0.1.1",
+		},
+		{
+			name: "RAIDWithSupportBMC",
+			newBMH: &BareMetalHost{
+				TypeMeta:   tm,
+				ObjectMeta: om,
+				Spec: BareMetalHostSpec{
+					RAID: &RAIDConfig{
+						HardwareRAIDVolumes: []HardwareRAIDVolume{
+							{
+								SizeGibibytes:         nil,
+								Level:                 "",
+								Name:                  "",
+								Rotational:            nil,
+								NumberOfPhysicalDisks: nil,
+							},
+						},
+					},
+					BootMACAddress: "01:02:03:04:05:06",
+					BMC: BMCDetails{
+						Address:         "irmc://127.0.1.1",
+						CredentialsName: "test1",
+					},
+				}},
+			oldBMH:    nil,
+			wantedErr: "",
 		},
 		{
 			name: "RAIDWithUnsupportBMC",
@@ -134,6 +180,24 @@ func TestValidateCreate(t *testing.T) {
 			wantedErr: "BMC driver ipmi does not support configuring RAID",
 		},
 		{
+			name: "FirmwareWithSupportBMC",
+			newBMH: &BareMetalHost{
+				TypeMeta:   tm,
+				ObjectMeta: om,
+				Spec: BareMetalHostSpec{
+					Firmware: &FirmwareConfig{
+						VirtualizationEnabled: &enable,
+					},
+					BootMACAddress: "01:02:03:04:05:06",
+					BMC: BMCDetails{
+						Address:         "irmc://127.0.1.1",
+						CredentialsName: "test1",
+					},
+				}},
+			oldBMH:    nil,
+			wantedErr: "",
+		},
+		{
 			name: "FirmwareWithUnsupportBMC",
 			newBMH: &BareMetalHost{
 				TypeMeta:   tm,
@@ -151,7 +215,7 @@ func TestValidateCreate(t *testing.T) {
 			wantedErr: "firmware settings for ipmi are not supported",
 		},
 		{
-			name: "BootMACAddressRequired",
+			name: "BootMACAddressRequiredWithoutBootMACAddress",
 			newBMH: &BareMetalHost{
 				TypeMeta:   tm,
 				ObjectMeta: om,
@@ -163,6 +227,21 @@ func TestValidateCreate(t *testing.T) {
 				}},
 			oldBMH:    nil,
 			wantedErr: "BMC driver libvirt requires a BootMACAddress value",
+		},
+		{
+			name: "BootMACAddressRequiredWithBootMACAddress",
+			newBMH: &BareMetalHost{
+				TypeMeta:   tm,
+				ObjectMeta: om,
+				Spec: BareMetalHostSpec{
+					BMC: BMCDetails{
+						Address:         "libvirt://127.0.1.1",
+						CredentialsName: "test1",
+					},
+					BootMACAddress: "00:00:00:00:00:00",
+				}},
+			oldBMH:    nil,
+			wantedErr: "",
 		},
 		{
 			name: "BootMACAddressRequired",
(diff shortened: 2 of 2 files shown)


apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -42,14 +55,93 @@ func (host *BareMetalHost) validateChanges(old *BareMetalHost) []error {
...
+	for _, v := range avaliableType {
+		m[v] = true
+	}
+
+	bmcType, err := getBMCType(b.Address)
+	if err != nil {
+		return err
+	}
+	if _, ok := m[bmcType]; !ok {
+		return fmt.Errorf("unsupport BMC Type %s", bmcType)
```
fmuyassarov, line 109:
```suggestion
		return fmt.Errorf("unsupported BMC Type %s", bmcType)
```

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -14,7 +15,19 @@ func (host *BareMetalHost) validateHost() []error {
 	log.Info("validate create", "name", host.Name)
 	var errs []error
 
-	if err := validateRAID(host.Spec.RAID); err != nil {
+	if err := validateRAID(host.Spec); err != nil {
```
ardaguclu, line 18: Instead passing whole `Spec`, we can pass just the required parameters. I think, only 1 or 2 fields are used.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -14,7 +15,19 @@ func (host *BareMetalHost) validateHost() []error {
...
+		errs = append(errs, err)
+	}
+
+	if err := validateBMC(host.Spec.BMC); err != nil {
+		errs = append(errs, err)
+	}
+
+	if err := validateBootMACAddress(host.Spec); err != nil {
```
ardaguclu, line 26: Only required fields can be passed.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -14,7 +15,19 @@ func (host *BareMetalHost) validateHost() []error {
...
+		errs = append(errs, err)
+	}
+
+	if err := validateFirmware(host.Spec); err != nil {
```
ardaguclu, line 30: Same

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -42,14 +55,93 @@ func (host *BareMetalHost) validateChanges(old *BareMetalHost) []error {
...
+
+	switch bmcType {
+	case "ilo5", "idrac", "irmc":
+		return nil
+	default:
+		return fmt.Errorf("currently the 'raid' field is only supported by ilo5/idrac/irmc")
+	}
+}
+
+func getBMCType(address string) (string, error) {
```
ardaguclu, line 80: It looks like we should use the same parse logic https://github.com/metal3-io/baremetal-operator/blob/master/pkg/ironic/bmc/access.go#L74 which does more stuff and less error prone.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -42,14 +55,93 @@ func (host *BareMetalHost) validateChanges(old *BareMetalHost) []error {
...
+	if b.Address == "" {
+		return nil
+	}
+
+	avaliableType := []string{
+		"ipmi", "idrac", "idrac-virtualmedia", "idrac-redfish",
+		"irmc", "ibmc", "ilo4", "ilo5", "ilo5-redfish",
+		"redfish", "redfish-virtualmedia", "libvirt",
+	}
+	m := make(map[string]bool)
```
ardaguclu, line 99: Suggested way is to use `map[string]struct`

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -42,14 +55,93 @@ func (host *BareMetalHost) validateChanges(old *BareMetalHost) []error {
...
+	for _, v := range avaliableType {
+		m[v] = true
+	}
+
+	bmcType, err := getBMCType(b.Address)
```
ardaguclu, line 104: Why we are calculating BMCType each type, we can extract in upper level and pass it required functions

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -42,14 +55,93 @@ func (host *BareMetalHost) validateChanges(old *BareMetalHost) []error {
...
+		return err
+	}
+	if _, ok := m[bmcType]; !ok {
+		return fmt.Errorf("unsupport BMC Type %s", bmcType)
+	}
 	return nil
 }
+
+func validateBootMACAddress(s BareMetalHostSpec) error {
+	if s.BMC.Address == "" {
```
ardaguclu, line 115: Are we sure checking `s.BMC.Address` instead `s.BootMACAddress`

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -42,14 +55,93 @@ func (host *BareMetalHost) validateChanges(old *BareMetalHost) []error {
...
+		return err
+	}
+
+	switch bmcType {
+	case "ilo4", "ilo5", "idrac", "irmc":
+		return nil
+	default:
+		return fmt.Errorf("currently the firmware field is only supported by ilo4/ilo5/irmc/idrac")
+	}
+}
```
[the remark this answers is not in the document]
ardaguclu: I agree with duplication is problematic. However, as we discussed previously, it is not easy currently. 
There are two ways to achieve this and prevent duplication;
- kubebuilder will support moving validations under `pkg` folder.
- or as @dtantsur stated in here https://github.com/openshift-metal3/terraform-provider-ironic/pull/53#issuecomment-925574411, these things are moved to metal3 agnostic module and safely used in here.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -42,14 +55,93 @@ func (host *BareMetalHost) validateChanges(old *BareMetalHost) []error {
...
+
+	return strings.Split(strs[0], "+http")[0], nil
+}
+
+func validateBMC(b BMCDetails) error {
+	if b.Address == "" {
+		return nil
+	}
+
+	avaliableType := []string{
```
ardaguclu, line 94: Can we define these as constants above

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -42,14 +55,93 @@ func (host *BareMetalHost) validateChanges(old *BareMetalHost) []error {
...
+	m := make(map[string]bool)
+	for _, v := range avaliableType {
+		m[v] = true
+	}
+
+	bmcType, err := getBMCType(b.Address)
+	if err != nil {
+		return err
+	}
+	if _, ok := m[bmcType]; !ok {
```
ardaguclu, line 108: I think we can add more validations like there is any incorrect character like spaces, etc.

ardaguclu (commented):
Thanks for PR @zhouhao3

apis/metal3.io/v1alpha1/baremetalhost_validation.go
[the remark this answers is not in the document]
s3rj1k: And seems that some shared code (validation helpers) are needed to be separated into isolated (by import deps) package to keep that isolated code with minimum dependencies, that will avoid import loops in future.

Something like `common` package or `shared` package, that does not import anything except golang core libs and possibly `api` package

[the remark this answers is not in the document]
andfasano: I think it's a good approach in general @s3rj1k , and moreover I would suggest a slightly different way to face it. The basic idea is to have a sort of validation framework reflection based, both in order to break the dependencies where needed and also to have a more structured way to contribute new validation rules.
In the past I've used the [go-playground validator](https://github.com/go-playground/validator) for that: it essentially allows to define and apply validations rules in different ways (mostly by annotating fields)' it's shipped with a number of prebuilt validations rules but it's also pretty easy to define new one.
For basic checks (ie `BootMACAddress` is not empty) its application it's really straitghforward. But I think that whenever we have an interface available it could be also adopted for more complex scenarios, reducing a lot the dependency issue. Another option on the plate.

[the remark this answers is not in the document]
s3rj1k: Something like https://github.com/asaskevich/govalidator with their standalone functions, no need to add struct tags to trigger struct validation, this validation specific tags can become an issue

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -42,14 +55,93 @@ func (host *BareMetalHost) validateChanges(old *BareMetalHost) []error {
...
+
+func validateBMC(b BMCDetails) error {
+	if b.Address == "" {
+		return nil
+	}
+
+	avaliableType := []string{
+		"ipmi", "idrac", "idrac-virtualmedia", "idrac-redfish",
+		"irmc", "ibmc", "ilo4", "ilo5", "ilo5-redfish",
+		"redfish", "redfish-virtualmedia", "libvirt",
```
dtantsur, line 97: We absolutely should not hardcode a list of drivers in the API, even if we don't actually believe that we'll ever have non-Ironic provisioners.

zhouhao3:
Because this PR needs to be called [NewAccessDetails](https://github.com/metal3-io/baremetal-operator/blob/master/pkg/ironic/bmc/access.go#L126)  after the change, it is necessary to wait for #1007 to merge before making changes.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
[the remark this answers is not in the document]
zhouhao3: I updated this PR to call the operator method for validation of RAID and BIOS.
Added validation method for BootMACAddress in webhook. PTAL, thanks!

apis/go.mod
```diff
@@ -3,8 +3,11 @@ module github.com/metal3-io/baremetal-operator/apis
 go 1.16
 
 require (
+	github.com/metal3-io/baremetal-operator/pkg/ironic v0.0.0-20211108163011-2e5b1ff824a2
```
ardaguclu, line 6: This module can not be imported in here because there is circular dependeny.

https://github.com/metal3-io/baremetal-operator/pull/1017 will fix it.

zhouhao3:
Hi, Since 1017 has merged, I updated this PR, PTAL.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -15,8 +15,29 @@ var log = logf.Log.WithName("baremetalhost-validation")
 func (host *BareMetalHost) validateHost() []error {
 	log.Info("validate create", "name", host.Name)
 	var errs []error
+	var bmcAccess bmc.AccessDetails
 
-	if err := validateRAID(host.Spec.RAID); err != nil {
+	if host.Spec.BMC.Address != "" {
+		var err error
+		bmcAccess, err = bmc.NewAccessDetails(host.Spec.BMC.Address, false)
```
ardaguclu, line 22: Why we are passing false in here?, because we don't know is it disabled or not?

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -53,5 +74,46 @@ func validateRAID(r *RAIDConfig) error {
 		return fmt.Errorf("hardwareRAIDVolumes and softwareRAIDVolumes can not be set at the same time")
 	}
 
+	if bmcAccess == nil || bmcAccess.RAIDInterface() == "no-raid" {
```
ardaguclu, line 77: According to the discussion in here https://github.com/metal3-io/baremetal-operator/pull/1040/files#r755873502, if bmc address is nil(thus bmcAccess is nil), we will allow pass through.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -53,5 +74,46 @@ func validateRAID(r *RAIDConfig) error {
...
+
+	return nil
+}
+
+func validateFirmware(firmware *FirmwareConfig, bmcAccess bmc.AccessDetails) error {
+	if firmware == nil {
+		return nil
+	}
+
+	if bmcAccess == nil {
```
ardaguclu, line 89: Same

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -53,5 +74,46 @@ func validateRAID(r *RAIDConfig) error {
...
+		return fmt.Errorf("firmware settings are not supported")
+	}
+
+	_, err := bmcAccess.BuildBIOSSettings((*bmc.FirmwareConfig)(firmware))
```
ardaguclu, line 93: It looks like firmware is already in `*bmc.FirmwareConfig` type, maybe we don't have to cast it?

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -53,5 +74,46 @@ func validateRAID(r *RAIDConfig) error {
...
+	}
+
+	if bmcAccess.NeedsMAC() && bootAddress == "" {
+		return fmt.Errorf("BMC driver %s requires a BootMACAddress value", bmcAccess.Type())
+	}
+
+	return nil
+}
+
+func validateBooMode(bootMode BootMode, bmcAccess bmc.AccessDetails) error {
```
ardaguclu, line 109: validateBootMode?

ardaguclu (commented):
Thanks @zhouhao3 for the PR. Generally it looks good to me, which check is for `Invalid BMC Config`?

apis/metal3.io/v1alpha1/baremetalhost_validation_test.go
```diff
@@ -71,6 +72,92 @@ func TestValidateCreate(t *testing.T) {
 			oldBMH:    nil,
 			wantedErr: "hardwareRAIDVolumes and softwareRAIDVolumes can not be set at the same time",
 		},
+		{
+			name: "unsupportBMCType",
```
ardaguclu, line 76: Could you please add at least 1 valid and 1 invalid test cases for all validations you are adding.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
zhouhao3: We still do this because parameter passed in is `Spec.Firmware`，it is in the v1alpha1 package.

zhouhao3:
@ardaguclu Thanks, all done, PTAL.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -53,5 +74,42 @@ func validateRAID(r *RAIDConfig) error {
 		return fmt.Errorf("hardwareRAIDVolumes and softwareRAIDVolumes can not be set at the same time")
 	}
 
+	if bmcAccess.RAIDInterface() == "no-raid" {
```
ardaguclu, line 77: If `bmcAccess` is nil,  this function panics. It would be better to handle it

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -53,5 +74,42 @@ func validateRAID(r *RAIDConfig) error {
...
+	if firmware == nil {
+		return nil
+	}
+
+	_, err := bmcAccess.BuildBIOSSettings((*bmc.FirmwareConfig)(firmware))
+	return err
+}
+
+func validateBootMACAddress(bootAddress string, bmcAccess bmc.AccessDetails) error {
+	if bmcAccess == nil {
```
ardaguclu, line 94: `bmc.AccessDetails` is in value form not reference, can it be nil?

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -53,5 +74,42 @@ func validateRAID(r *RAIDConfig) error {
...
+
+	if bmcAccess.NeedsMAC() && bootAddress == "" {
+		return fmt.Errorf("BMC driver %s requires a BootMACAddress value", bmcAccess.Type())
+	}
+
+	return nil
+}
+
+func validateBootMode(bootMode BootMode, bmcAccess bmc.AccessDetails) error {
+	if bmcAccess == nil {
```
ardaguclu, line 106: Same

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -15,8 +15,29 @@ var log = logf.Log.WithName("baremetalhost-validation")
 func (host *BareMetalHost) validateHost() []error {
 	log.Info("validate create", "name", host.Name)
 	var errs []error
+	var bmcAccess bmc.AccessDetails
 
-	if err := validateRAID(host.Spec.RAID); err != nil {
+	if host.Spec.BMC.Address != "" {
+		var err error
```
ardaguclu, line 21: We can remove this and use below line as;
`bmcAccess, err := bmc.NewAccessDetails(host.Spec.BMC.Address, host.Spec.BMC.DisableCertificateVerification)`

apis/metal3.io/v1alpha1/baremetalhost_validation.go
ardaguclu, line 77: if bmcAccess is nil, this will panic.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
ardaguclu: if bmcAccess is nil, this will panic.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
zhouhao3: We can't do this. Because the value of `bmcAccess` cannot be used outside of the `if`.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
zhouhao3: Thanks, done.

ardaguclu:
Thanks @zhouhao3. Code looks good to me, but validation rules need to be reviewed.
/ok-to-test
/test-v1b1-integration
/test-v1b1-centos-integration

cc: @dtantsur

ardaguclu:
Validations look sane too.
/lgtm

zhouhao3:
@dtantsur PTAL

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -53,5 +74,46 @@ func validateRAID(r *RAIDConfig) error {
 		return fmt.Errorf("hardwareRAIDVolumes and softwareRAIDVolumes can not be set at the same time")
 	}
 
+	if bmcAccess == nil {
+		return nil
+	}
+
+	if bmcAccess.RAIDInterface() == "no-raid" {
```
andfasano, line 81: Minor: wouldn't be better to do this check as first?

zhouhao3: yes, updated. PTAL.

zhouhao3:
@andfasano PTAL

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -17,15 +17,36 @@ var log = logf.Log.WithName("baremetalhost-validation")
...
+	}
+
+	if err := validateBootMACAddress(host.Spec.BootMACAddress, bmcAccess); err != nil {
+		errs = append(errs, err)
+	}
+
+	if err := validateBootMode(host.Spec.BootMode, bmcAccess); err != nil {
+		errs = append(errs, err)
+	}
+
```
andfasano, line 49: It looks like all these new validations are related to BMC, and every one repeats the nil check (so they will be conditionally active only when there's a valid bmc access). I guess it'd preferable to group all these validation funcs under a common one, in order to avoid duplicating the check for every distinct validation func.

ardaguclu: Good point and agree

zhouhao3: Done, thanks, PTAL.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -50,16 +58,38 @@ func (host *BareMetalHost) validateChanges(old *BareMetalHost) []error {
 	return errs
 }
 
-func validateRAID(r *RAIDConfig) error {
-	if r == nil {
-		return nil
+func validateBMCAccess(s BareMetalHostSpec, bmcAccess bmc.AccessDetails) []error {
```
zhouhao3, line 61: Maybe the name of this function is inappropriate, not sure what it should be called.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -17,11 +17,19 @@ var log = logf.Log.WithName("baremetalhost-validation")
...
+	if host.Spec.BMC.Address != "" {
+		var err error
+		bmcAccess, err = bmc.NewAccessDetails(host.Spec.BMC.Address, host.Spec.BMC.DisableCertificateVerification)
+		if err != nil {
+			errs = append(errs, err)
+		}
 	}
 
+	errs1 := validateBMCAccess(host.Spec, bmcAccess)
+	errs = append(errs, errs1...)
```
ardaguclu, line 31: Can't this be written in one line `errs = append(errs, validateBMCAccess(host.Spec, bmcAccess)...)`?

kashifest:
/hold

kashifest:
/hold cancel

zhouhao3:
@andfasano PTAL

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -50,16 +57,38 @@ func (host *BareMetalHost) validateChanges(old *BareMetalHost) []error {
...
+
+	if bmcAccess == nil {
+		return errs
 	}
 
-	if len(r.HardwareRAIDVolumes) > 0 && len(r.SoftwareRAIDVolumes) > 0 {
-		return fmt.Errorf("hardwareRAIDVolumes and softwareRAIDVolumes can not be set at the same time")
+	if s.RAID != nil {
+		if bmcAccess.RAIDInterface() == "no-raid" {
+			errs = append(errs, fmt.Errorf("the 'raid' field for %s is not supported", bmcAccess.Type()))
```
dtantsur, line 69: For consistency with other error messages: "BMC driver %s does not support configuring RAID".

By the way, all this logic is completely broken for software RAID :-/ I guess we need to solve it eventually.

dtantsur: Actually, no, only this validation is broken. You need to check for hardware RAID configuration, for software RAID we will set RAIDInterface to "agent" when handling it.

zhouhao3: Thanks, updated. PTAL.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -50,16 +57,38 @@ func (host *BareMetalHost) validateChanges(old *BareMetalHost) []error {
...
 	}
 
-	if len(r.HardwareRAIDVolumes) > 0 && len(r.SoftwareRAIDVolumes) > 0 {
-		return fmt.Errorf("hardwareRAIDVolumes and softwareRAIDVolumes can not be set at the same time")
+	if s.RAID != nil && len(s.RAID.HardwareRAIDVolumes) > 0 {
+		if bmcAccess.RAIDInterface() == "no-raid" {
+			errs = append(errs, fmt.Errorf("BMC driver %s does not support configuring RAID", bmcAccess.Type()))
+		}
+
+		if len(s.RAID.SoftwareRAIDVolumes) > 0 {
```
zaneb, line 72: This isn't really BMC-related, I think I'd prefer to have kept the validateRAID() function as well (this won't be the last RAID validation we need), even if it means repeating the if statement from line 67 in two places.

zhouhao3: Done, PTAL.

PR #1004, opened 2021-10-19
gitignore: Get rid of rules not associated with the project

zaneb:
Everybody has their own set of tools that generate different tempfiles
that should be ignored. The correct place to configure this is in the
global gitignore file "~/.config/git/ignore". Including ignore rules for
a select set of tools is an antipattern that encourages people to want
to add their own tools to the list in each of the hundreds of projects
they interact with instead of setting up their global rules once.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,55 +2,7 @@
 bin
 build/_output
 build/_test
-# Created by https://www.gitignore.io/api/go,vim,emacs,visualstudiocode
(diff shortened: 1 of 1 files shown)


s3rj1k:
https://github.com/metal3-io/baremetal-operator/issues/1005

fmuyassarov (commented):
/lgtm
/test-integration

PR #1007, opened 2021-10-21
Remove bmc package dependency on API

zaneb:
We'll want to use this package in validation (which must be defined in
the API), so don't have it depend on the API. Instead, define a local
structure for the FirmwareConfig data.

### DIFF
--- a/pkg/ironic/bmc/access.go
+++ b/pkg/ironic/bmc/access.go
@@ -7,8 +7,6 @@ import (
 	"strings"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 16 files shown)


s3rj1k:
shouldn't we move `type FirmwareConfig struct {` to separate package?
something like `pkg/ironic/bmc/api/`?

zaneb:
> shouldn't we move `type FirmwareConfig struct {` to separate package? something like `pkg/ironic/bmc/api/`?

No, why?

PR #1008, opened 2021-10-25
Use finalizer for BMC secret.

s3rj1k:
There is a race condition when deleting namespace using API/CLI that contains BMH and BMC Secrets.
Secrets can be deleting before actual de-provisioning of BMH has started and this leads to broken state with registration error.

By using finalizer that is set for secrets that are OwnerRef`ed this change ensures proper objects deletion order by un-setting secrets finalizer only after deprovisioning has ended, k8s will then be allowed to clean-up remaining BMC secrets.

### DIFF
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -526,6 +526,13 @@ func (r *BareMetalHostReconciler) actionDeleting(prov provisioner.Provisioner, i
 	}
 
 	// Remove finalizer to allow deletion
+	secretManager := secretutils.NewSecretManager(info.log, r.Client, r.APIReader)
+
+	err = secretManager.ReleaseSecret(info.bmcCredsSecret)
+	if err != nil {
+		return actionError{err}
+	}
+
 	info.host.Finalizers = utils.FilterStringFromList(
 		info.host.Finalizers, metal3v1alpha1.BareMetalHostFinalizer)
 	info.log.Info("cleanup is complete, removed finalizer",
@@ -624,7 +631,7 @@ func (r *BareMetalHostReconciler) preprovImageAvailable(info *reconcileInfo, ima
 			Namespace: image.ObjectMeta.Namespace,
 		}
 		secretManager := r.secretManager(info.log)
-		networkData, err := secretManager.AcquireSecret(secretKey, info.host, false)
+		networkData, err := secretManager.AcquireSecret(secretKey, info.host, false, false)
 		if err != nil {
 			return false, err
 		}
@@ -1424,7 +1431,7 @@ func (r *BareMetalHostReconciler) getBMCSecretAndSetOwner(request ctrl.Request, 
 	reqLogger := r.Log.WithValues("baremetalhost", request.NamespacedName)
 	secretManager := r.secretManager(reqLogger)
 
-	bmcCredsSecret, err := secretManager.AcquireSecret(host.CredentialsKey(), host, true)
+	bmcCredsSecret, err := secretManager.AcquireSecret(host.CredentialsKey(), host, true, true)
 	if err != nil {
 		if k8serrors.IsNotFound(err) {
 			return nil, &ResolveBMCSecretRefError{message: fmt.Sprintf("The BMC secret %s does not exist", host.CredentialsKey())}
--- a/pkg/secretutils/secret_manager.go
+++ b/pkg/secretutils/secret_manager.go
@@ -12,6 +12,13 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	metal3v1alpha1 "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
+	"github.com/metal3-io/baremetal-operator/pkg/utils"
+)
+
+const (
+	SecretsFinalizer = metal3v1alpha1.BareMetalHostFinalizer + "/secret"
 )
 
 // SecretManager is a type for fetching Secrets whether or not they are in the
@@ -58,7 +65,7 @@ func (sm *SecretManager) findSecret(key types.NamespacedName) (secret *corev1.Se
 // claimSecret ensures that the Secret has a label that will ensure it is
 // present in the cache (and that we can watch for changes), and optionally
 // that it has a particular owner reference.
-func (sm *SecretManager) claimSecret(secret *corev1.Secret, owner client.Object, ownerIsController bool) error {
+func (sm *SecretManager) claimSecret(secret *corev1.Secret, owner client.Object, ownerIsController, addFinalizer bool) error {
 	log := sm.log.WithValues("secret", secret.Name, "secretNamespace", secret.Namespace)
 	needsUpdate := false
 	if !metav1.HasLabel(secret.ObjectMeta, LabelEnvironmentName) {
@@ -98,6 +105,12 @@ func (sm *SecretManager) claimSecret(secret *corev1.Secret, owner client.Object,
 		}
 	}
 
+	if addFinalizer && !utils.StringInList(secret.Finalizers, SecretsFinalizer) {
+		log.Info("setting secret finalizer")
+		secret.Finalizers = append(secret.Finalizers, SecretsFinalizer)
+		needsUpdate = true
+	}
+
 	if needsUpdate {
 		if err := sm.client.Update(context.TODO(), secret); err != nil {
 			return errors.Wrap(err, fmt.Sprintf("failed to update secret %s in namespace %s", secret.ObjectMeta.Name, secret.ObjectMeta.Namespace))
@@ -111,12 +124,12 @@ func (sm *SecretManager) claimSecret(secret *corev1.Secret, owner client.Object,
 // will ensure it is present in the cache (and that we can watch for changes),
 // and optionally that it has a particular owner reference. The owner reference
 // may optionally be a controller reference.
-func (sm *SecretManager) obtainSecretForOwner(key types.NamespacedName, owner client.Object, ownerIsController bool) (*corev1.Secret, error) {
+func (sm *SecretManager) obtainSecretForOwner(key types.NamespacedName, owner client.Object, ownerIsController, addFinalizer bool) (*corev1.Secret, error) {
 	secret, err := sm.findSecret(key)
 	if err != nil {
 		return nil, errors.Wrap(err, fmt.Sprintf("failed to fetch secret %s in namespace %s", key.Name, key.Namespace))
 	}
-	err = sm.claimSecret(secret, owner, ownerIsController)
+	err = sm.claimSecret(secret, owner, ownerIsController, addFinalizer)
 
 	return secret, err
 }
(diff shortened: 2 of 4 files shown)


fmuyassarov:
@s3rj1k Can you please share the steps to reproduce it?
I've tried
1. BMO and BMH in the same namespace, then deleted the namespace, but since BMO pod is killed, BMH are stuck (which is expected)
2. BMO and BMH are in different namespaces, then I deleted the BMH namepace, and BMO could delete BMHs successfully and eventually namespace was deleted fine.

s3rj1k:
@fmuyassarov 
 - BMO, CAPM3 in namespace A
 - BMH, Machine in namespace B
 - Machines are linked(consumed) to BMH
 - BMHs are all in provisioned state
 - BMHs have at least two secrets, BMC access secret, UserData secret
 - There are at least 3 BMHs/Machines in namespace B
 - Manually deleting namespace B -> stuck

fmuyassarov:
> @fmuyassarov
> 
> * BMO, CAPM3 in namespace A
> * BMH, Machine in namespace B
> * Machines are linked(consumed) to BMH
> * BMHs are all in provisioned state
> * BMHs have at least two secrets, BMC access secret, UserData secret
> * There are at least 3 BMHs/Machines in namespace B
> * Manually deleting namespace B -> stuck

Unfortunately I can't reproduce it on my side with metal3-dev-env with this setup, maybe there is something different if you are running on physical servers.

s3rj1k:
@fmuyassarov so what do we do about this? 

Assuming that this change is mostly none-invasive, can we merge this without reproduce from your side?

I have 100% reproducibility and it is a kind of show stopper for me.

Also maybe there is a dependency on K8s version itself? (recent versions had some improvements with GC)

fmuyassarov:
> @fmuyassarov so what do we do about this?
> 
> Assuming that this change is mostly none-invasive, can we merge this without reproduce from your side?
> 
> I have 100% reproducibility and it is a kind of show stopper for me.
> 
> Also maybe there is a dependency on K8s version itself? (recent versions had some improvements with GC)

Yes sure, I'm not blocking it, just wanted to understand in which cases would this happen, because I've never faced it before, although I do delete namespace sometimes and assume everything to go smooth. And I'm not seeing any harm of using finilizers. 
/lgtm

fmuyassarov:
what's the your kubernetes version @s3rj1k ?

s3rj1k:
> what's the your kubernetes version @s3rj1k ?

1.20.8

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -534,6 +534,20 @@ func (r *BareMetalHostReconciler) actionDeleting(prov provisioner.Provisioner, i
...
+			info.bmcCredsSecret.Finalizers, metal3v1alpha1.BareMetalHostFinalizer+"/secret")
+
+		if err := r.Update(context.Background(), info.bmcCredsSecret); err != nil {
+			return actionError{errors.Wrap(err, fmt.Sprintf("failed to remove finalizer from secret %s in namespace %s",
+				info.bmcCredsSecret.ObjectMeta.Name, info.bmcCredsSecret.ObjectMeta.Namespace))}
+		}
+
+		info.log.Info("removed secret finalizer",
+			"remaining", info.bmcCredsSecret.Finalizers)
+	}
```
zaneb, line 549: Please move this implementation to the SecretManager as a `ReleaseSecret()` method or something like that.

zaneb (commented):
I'm not necessarily opposed to this, but as currently written it will leave a lot of Secrets permanently undeletable without manual intervention to remove the finalizer:

* The old BMC secret if the user changes the BMH Spec to point to the new one
* The Secrets used for userData/metaData/networkData

If we have to do this then we might as well not use owner references, because effectively it's the controller that has to decide when to delete these Secrets.

s3rj1k:
@zaneb currently only bmc secret has ownerref, thus only that secret would
have finalizer.

пн, 1 нояб. 2021 г., 18:00 Zane Bitter ***@***.***>:

zaneb:
> @zaneb currently only bmc secret has ownerref, thus only that secret would have finalizer.

That doesn't follow, because the code you've added is not conditional on a non-nil ownerRef. It's also untrue, because other Secrets _do_ have owner refs.

[`claimSecret()`](https://github.com/metal3-io/baremetal-operator/pull/1008/files#diff-a883846d77daf3b5a84b2aa01499e7497baed3e1e38ac61357c207c3e5886c48R64) is called (via [`obtainSecretForOwner()`](https://github.com/metal3-io/baremetal-operator/blob/master/pkg/secretutils/secret_manager.go#L114)) by both [`AcquireSecret()`](https://github.com/metal3-io/baremetal-operator/blob/master/pkg/secretutils/secret_manager.go#L128) (adds owner ref) and [`ObtainSecret()`](https://github.com/metal3-io/baremetal-operator/blob/master/pkg/secretutils/secret_manager.go#L138) (no owner ref). Both of these are used in numerous places, not just for the BMC creds:

* https://github.com/metal3-io/baremetal-operator/blob/master/controllers/metal3.io/baremetalhost_controller.go#L1427 (BMC creds)
* https://github.com/metal3-io/baremetal-operator/blob/master/controllers/metal3.io/host_config_data.go#L29
* https://github.com/metal3-io/baremetal-operator/blob/master/controllers/metal3.io/baremetalhost_controller.go#L627
* https://github.com/metal3-io/baremetal-operator/blob/master/controllers/metal3.io/preprovisioningimage_controller.go#L154

s3rj1k:
There is an if on line 72 that checks ownerref precence, way above patch
lines.
On code that adds finalizer.

Also in my env I didn't see ownerref on UserData, but If it was intended to
have that ref it is actually good.

Also I am not opposed to refactor, just pointibg out that in my case the
only ownerref on secret was on bmc secret.


пн, 1 нояб. 2021 г., 18:31 Zane Bitter ***@***.***>:

zaneb:
Ah yep, you're correct, I misread the indentation.
However, it's still the case that the networkData secret(s) also have an owner ref.
You could also make it conditional on the `ownerIsController` flag being true; that is only the case for the BMC Secret for now. However, this is an extremely non-obvious side-effect that probably warrants its own explicit API flag.

s3rj1k:
@zaneb @fmuyassarov PTAL

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -534,6 +534,28 @@ func (r *BareMetalHostReconciler) actionDeleting(prov provisioner.Provisioner, i
...
+	err = secretManager.ReleaseSecretByRef(info.host.Spec.NetworkData)
+	if err != nil {
+		return actionError{err}
+	}
+
+	err = secretManager.ReleaseSecretByRef(info.host.Spec.MetaData)
+	if err != nil {
+		return actionError{err}
+	}
+
```
andfasano, line 558: I think this part could deserve a test, to verify that the finalizer is removed on host deletions from the secrets

pkg/secretutils/secret_manager.go
```diff
@@ -96,6 +103,12 @@ func (sm *SecretManager) claimSecret(secret *corev1.Secret, owner client.Object,
 				needsUpdate = true
 			}
 		}
+
+		if !utils.StringInList(secret.Finalizers, SecretsFinalizer) {
+			log.Info("setting secret finalizer")
+			secret.Finalizers = append(secret.Finalizers, SecretsFinalizer)
+			needsUpdate = true
+		}
```
andfasano, line 111: And also this one, to verify that a newly claimed secret has the finalizer set

controllers/metal3.io/baremetalhost_controller.go
s3rj1k: I assume that the proper way of adding such a test is to add it into ReleaseSecretByRef and into ReleaseSecret

pkg/secretutils/secret_manager.go
s3rj1k: We do not test that secret has ownerRef set if it was intended to set it (assuming that I understand correctly and point is about fetching newly updated secret object and checking that it has finilazer set), so why we would test-verify finalizers and ignore other fields?

andfasano: Well, currently we have a finalizer dedicate test, checking that as part of the reconciliation loop the host has the finalizer set (see the [TestAddFinalizers](https://github.com/metal3-io/baremetal-operator/blob/2de940e5f2c0603495f039e90123ddaa0f549e60/controllers/metal3.io/baremetalhost_controller_test.go#L439) test). So probably it could be a good place to include also the checks on the secrets.

(we also have a test for the secret setup fields, including the OwnerReference [here](https://github.com/metal3-io/baremetal-operator/blob/2de940e5f2c0603495f039e90123ddaa0f549e60/controllers/metal3.io/baremetalhost_controller_test.go#L718) that could work as well, but the previous one has more focus on the finalization)

controllers/metal3.io/baremetalhost_controller.go
andfasano: For that I was thinking more on enhancing somehow the [TestDeleteHost](https://github.com/metal3-io/baremetal-operator/blob/2de940e5f2c0603495f039e90123ddaa0f549e60/controllers/metal3.io/baremetalhost_controller_test.go#L1315) with the new cases on secrets

pkg/secretutils/secret_manager.go
```diff
@@ -12,6 +12,13 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	metal3v1alpha1 "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
+	"github.com/metal3-io/baremetal-operator/pkg/utils"
+)
+
+const (
+	SecretsFinalizer = metal3v1alpha1.BareMetalHostFinalizer + "/secret"
```
zaneb, line 21: I don't think we should add the /secret suffix on here; better to use the finalizer that is already defined in the API module.

pkg/secretutils/secret_manager.go
```diff
@@ -96,6 +103,12 @@ func (sm *SecretManager) claimSecret(secret *corev1.Secret, owner client.Object,
 				needsUpdate = true
 			}
 		}
+
+		if !utils.StringInList(secret.Finalizers, SecretsFinalizer) {
```
zaneb, line 107: I think we should pass a specific `addFinalizer` flag to this function to trigger adding the finalizer, and do it only for the BMC creds secret.
Already in this patch we're missing removing the finalizer from `info.host.Spec.PreprovisioningNetworkData`, both here and in the PreprovisioningImage controller. Problems like this will only continue to multiply if adding the finalizer is implicit rather than explicit.

s3rj1k: So do we want to set finalizers for:
 - all secrets
 - all secrets that have ownerRef
 - only BMC secret that has ownerRef
 - only BMC secret regardless of ownerRef
 
 ?

s3rj1k: ^^^ @zaneb @fmuyassarov @andfasano

zaneb: Only BMC secret, which always has ownerRef.

s3rj1k:
Updated OR based on recent comments
@fmuyassarov @zaneb @andfasano PTAL

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -534,6 +534,13 @@ func (r *BareMetalHostReconciler) actionDeleting(prov provisioner.Provisioner, i
 		return actionError{errors.Wrap(err, "failed to remove finalizer")}
 	}
 
+	secretManager := secretutils.NewSecretManager(info.log, r.Client, r.APIReader)
+
+	err = secretManager.ReleaseSecret(info.bmcCredsSecret)
```
zaneb, line 539: This must be done before the finalizer is removed from the BMH, otherwise the finalizer on the Secret could be left behind forever.

pkg/secretutils/secret_manager.go
```diff
@@ -125,16 +138,54 @@ func (sm *SecretManager) obtainSecretForOwner(key types.NamespacedName, owner cl
...
+	}
+
+	sm.log.Info("removed secret finalizer",
+		"remaining", secret.Finalizers)
+
+	return nil
+}
+
+// ReleaseSecretByRef removes secrets manager finalizer from secret specified be reference when needed.
+func (sm *SecretManager) ReleaseSecretByRef(secref *corev1.SecretReference) error {
```
zaneb, line 177: This is not actually used any more, is it? Let's just remove it.

s3rj1k: not used, yes, was thinking to keep it for consistent public interface for SecretManager, just in case we start to think about other secrets with finalizers, will remove

controllers/metal3.io/baremetalhost_controller.go
s3rj1k: fixed

pkg/secretutils/secret_manager.go
s3rj1k: For core objects like secret (does not apply to objects created via CRD, like BMH):

> Finalizer must be fully qualified since there's a validation process, so it must be declared according to the pattern prometherion/whatever_you_want, taking care the first part must adhere to DNS-1123 specification.

s3rj1k:
@zaneb Tests are green, all comments accounted for, please take a look, thanks

zaneb (commented):
/lgtm

s3rj1k:
@fmuyassarov @andfasano Please also take a look and hopefully merge, thanks

fmuyassarov (approved):

PR #1011, opened 2021-10-27
Uplift to kubebuilder v3

### LINKED ISSUE #977: Migrate to kubebuilder v3
We need to migrate to kubebuilder v3 in order to get api version of `v1` of CRDs generated. It is only kubebuilder v3 which by default generates `v1` based CRDs, and for v2 kubebuilder we get normally get `v1beta1` (which is already removed in 1.22 k8s).

kashifest:
Fixes #977

### DIFF
--- a/PROJECT
+++ b/PROJECT
@@ -1,16 +1,47 @@
+domain: metal3.io
+layout:
+- go.kubebuilder.io/v3
 multigroup: true
+projectName: baremetal-operator
 repo: github.com/metal3-io/baremetal-operator
 resources:
-- group: metal3.io
+- api:
+    crdVersion: v1
+    namespaced: true
+  controller: true
+  domain: metal3.io
+  group: metal3.io
   kind: BareMetalHost
+  path: github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1
   version: v1alpha1
-- group: metal3.io
+  webhooks:
+    validation: true
+    webhookVersion: v1
+- api:
+    crdVersion: v1
+    namespaced: true
+  controller: true
+  domain: metal3.io
+  group: metal3.io
   kind: HostFirmwareSettings
+  path: github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1
   version: v1alpha1
-- group: metal3.io
+- api:
+    crdVersion: v1
+    namespaced: true
+  controller: true
+  domain: metal3.io
+  group: metal3.io
   kind: FirmwareSchema
+  path: github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1
   version: v1alpha1
-- group: metal3.io
+- api:
+    crdVersion: v1
+    namespaced: true
+  controller: true
+  domain: metal3.io
+  group: metal3.io
   kind: PreprovisioningImage
+  path: github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1
   version: v1alpha1
-version: "2"
+version: "3"
--- a/apis/metal3.io/v1alpha1/baremetalhost_types.go
+++ b/apis/metal3.io/v1alpha1/baremetalhost_types.go
@@ -1,6 +1,5 @@
 /*
 
-
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -37,16 +37,19 @@ spec:
           - configMapRef:
               name: ironic
         name: manager
-        readinessProbe:
-          httpGet:
-            path: /healthz
-            port: 9440
-          initialDelaySeconds: 3
-          periodSeconds: 3
+        securityContext:
+          allowPrivilegeEscalation: false
         livenessProbe:
           httpGet:
             path: /healthz
             port: 9440
-          initialDelaySeconds: 10
-          periodSeconds: 3
+          initialDelaySeconds: 15
+          periodSeconds: 20
+        readinessProbe:
+          httpGet:
+            path: /readyz
+            port: 9440
+          initialDelaySeconds: 5
+          periodSeconds: 10
+      serviceAccountName: controller-manager
       terminationGracePeriodSeconds: 10
--- a/hack/unit.sh
+++ b/hack/unit.sh
@@ -11,9 +11,7 @@ if [ "${IS_CONTAINER}" != "false" ]; then
   eval "$(go env)"
   cd "${GOPATH}"/src/github.com/metal3-io/baremetal-operator
   export XDG_CACHE_HOME="/tmp/.cache"
-
   export COVER_PROFILE="${ARTIFACTS}"/cover.out
-
   TEST_FLAGS=${TEST_FLAGS} make -e unit-cover
 
 else
--- a/main.go
+++ b/main.go
@@ -1,6 +1,5 @@
 /*
 
-
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
@@ -83,7 +82,7 @@ func setupWebhooks(mgr ctrl.Manager) {
 
 func main() {
 	var watchNamespace string
-	var metricsAddr string
+	var metricsBindAddr string
 	var enableLeaderElection bool
 	var preprovImgEnable bool
 	var devLogging bool
@@ -97,7 +96,7 @@ func main() {
 	// namespace.
 	flag.StringVar(&watchNamespace, "namespace", os.Getenv("WATCH_NAMESPACE"),
 		"Namespace that the controller watches to reconcile host resources.")
-	flag.StringVar(&metricsAddr, "metrics-addr", "127.0.0.1:8085",
+	flag.StringVar(&metricsBindAddr, "metrics-addr", "127.0.0.1:8085",
 		"The address the metric endpoint binds to.")
 	flag.BoolVar(&enableLeaderElection, "enable-leader-election", false,
 		"Enable leader election for controller manager. "+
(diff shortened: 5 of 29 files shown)


main.go
```diff
@@ -97,9 +97,9 @@ func main() {
 	// namespace.
 	flag.StringVar(&watchNamespace, "namespace", os.Getenv("WATCH_NAMESPACE"),
 		"Namespace that the controller watches to reconcile host resources.")
-	flag.StringVar(&metricsAddr, "metrics-addr", "127.0.0.1:8085",
+	flag.StringVar(&metricsBindAddr, "metrics-bind-address", "127.0.0.1:8085",
 		"The address the metric endpoint binds to.")
-	flag.BoolVar(&enableLeaderElection, "enable-leader-election", false,
+	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
```
zaneb, line 102: Changing these flags is going to break people. Notwithstanding that these are the new defaults, I don't think there's any actual problem with keeping the ones we had is there?

hack/unit.sh
```diff
@@ -11,9 +11,15 @@ if [ "${IS_CONTAINER}" != "false" ]; then
...
   export XDG_CACHE_HOME="/tmp/.cache"
-
   export COVER_PROFILE="${ARTIFACTS}"/cover.out
-
+  
+  # The following downloads a set of envtest binaries to run tests with
+  export K8S_VERSION=1.22.1
+  curl -sSLo envtest-bins.tar.gz "https://go.kubebuilder.io/test-tools/${K8S_VERSION}/$(go env GOOS)/$(go env GOARCH)"
+  mkdir /usr/local/kubebuilder
+  tar -C /usr/local/kubebuilder --strip-components=1 -zvxf envtest-bins.tar.gz
```
zaneb, line 20: What is this stuff actually for? Can we build from source in the `hack` module the way we do for other stuff?

Is this file even used by anything?

zaneb (commented):
Thanks for doing this! It would be helpful to know which parts of the patch are manual changes needed to get it to work with v3, which are the result of changing to the new version for autogeneration, and which (if any) are something else.

hack/unit.sh
fmuyassarov: unit.sh is used by [unit test job in Prow](https://github.com/metal3-io/project-infra/blob/b124cfa18a06e27e78cc7c1b86da4ce4d58433f0/prow/config/config.yaml#L281) on a pull request.

We need to download these tarball, because it contains all necessary binaries like etcd, kubelet, api-server and etc, which we need during the test - to set up and start etcd instance and the k8s API server.

main.go
kashifest: Hi, these are auto generated and we are using the same names in CAPI and CAPM3 as well. However, I am ok with keeping the old flags or change it when we have introduced releases in this repo so that it doesn't break any existing workflow.

hack/unit.sh
kashifest: The new suite test file which checks for example if apis are added correctly to schemas, require the binaries and I have followed the kubebuilder way of doing envtest from `https://book.kubebuilder.io/reference/envtest.html`. As long as the file's usage as @fmuyassarov indicated we do use them in prow. I am open to ideas if there are other alternatives to do this.

kashifest:
> Thanks for doing this! It would be helpful to know which parts of the patch are manual changes needed to get it to work with v3, which are the result of changing to the new version for autogeneration, and which (if any) are something else.

@zaneb most of it are auto-generated. I have manually changed the flags in main.go since those were also auto generated initially and the port numbers in the manifests. The other manual changes were to make the unit and manifest-lint to pass.

hack/unit.sh
zaneb: FWIW kubebuilder has been generating the suite_test file for a while now but we keep deleting it because it doesn't work (at least locally, not sure about CI). Perhaps that's because it was missing this stuff... but also this script isn't called when we run tests locally so that makes me suspect this won't work either.
TBH I'd be happy to just keep deleting the suite_test every time kubebuilder adds it. We can be reasonably confident that autogenerated code works :)

kashifest: Alright, I am not so strong about this either. I will push an update soon removing suite_test

kashifest:
I have addressed your comments @zaneb . PTAL again

config/manager/manager.yaml
```diff
@@ -37,10 +37,19 @@ spec:
...
         name: manager
+        securityContext:
+          allowPrivilegeEscalation: false
         livenessProbe:
           httpGet:
             path: /healthz
             port: 9440
-          initialDelaySeconds: 3
-          periodSeconds: 3
+          initialDelaySeconds: 15
```
andfasano, line 46: @kashifest did you also change these value or were part of the automatic update?

kashifest: These were also auto generated.

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -1,5 +1,5 @@
 /*
-
+Copyright 2021.
```
zaneb, line 2: Could we not, or is it doing to want to replace this every time we change something?

zaneb (commented):
/lgtm

apis/metal3.io/v1alpha1/baremetalhost_types.go
kashifest: Good question. Not sure if thats the case. Shall I remove it from all the files?

zaneb: I'd prefer that, it can only get out of date.

kashifest: Done

kashifest:
Shall we take this in then? cc @zaneb @furkatgofurov7 @fmuyassarov

fmuyassarov (commented):
Some comments inline, otherwise looks good

PROJECT
```diff
@@ -1,16 +1,44 @@
+domain: metal3.io
+layout:
+- go.kubebuilder.io/v3
 multigroup: true
```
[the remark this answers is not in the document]
kashifest: Its there check the [migration guide](https://book.kubebuilder.io/migration/multi-group.html)

PROJECT
```diff
@@ -1,16 +1,44 @@
...
 resources:
-- group: metal3.io
+- api:
+    crdVersion: v1
+    namespaced: true
+  controller: true
+  domain: metal3.io
+  group: metal3.io
   kind: BareMetalHost
+  path: github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1
```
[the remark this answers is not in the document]
kashifest: I have generated the webhook file initially using kubebuilder as instructed [here](https://book.kubebuilder.io/reference/webhook-overview.html?highlight=webhook#supporting-older-cluster-versions) which didn't add webhook in the Project file automatically. Not sure if we should add it manually here.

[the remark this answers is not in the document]
fmuyassarov: yes Kubebuilder by default doesn't do it and generates very minimal  file. So, since we have the webhook enabled, I would suggest that we add it
```
  webhooks:
    validation: true
    webhookVersion: v1
```

PROJECT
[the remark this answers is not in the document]
fmuyassarov: if you are talking about the below part, that's for v2
```
version: "2"
domain: tutorial.kubebuilder.io
repo: tutorial.kubebuilder.io/project
multigroup: true
```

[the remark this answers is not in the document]
fmuyassarov: all the available fields are mentioned in https://book.kubebuilder.io/reference/project-config.html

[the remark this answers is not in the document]
kashifest: I followed https://book.kubebuilder.io/migration/migration_guide_v2tov3.html#migrate-the-apis , and that suggests the same as it was for v2. The fields was added by kubebuilder v3. I am reluctant to remove it specially since it is added by kubebuilder v3 itself. 

`Please run kubebuilder edit --multigroup=true to enable multi-group support before migrating the APIs and controllers. Please see [this](https://book.kubebuilder.io/migration/multi-group.html) for more details.`

PROJECT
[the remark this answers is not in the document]
kashifest: Not sure if thats a necessity. Some examples:
1. https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/PROJECT -> V3 project
2. https://github.com/kubernetes-sigs/cluster-api/blob/main/PROJECT -> v2 project
3. https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/PROJECT - > v2 project

My point on not touching it manually being, if we are modifying this file manually, I see no point of this PR itself since we could have manually modified other files as well as we were doing.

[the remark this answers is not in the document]
fmuyassarov: ok sure

PROJECT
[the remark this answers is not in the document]
fmuyassarov: Ok, I got confused then. Because in the Kubebuilder book, they have not mentioned multigroup field, while it seems it is possible to add that field.

PROJECT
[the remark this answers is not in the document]
kashifest: @fmuyassarov I have tried adding the webhook generation through kubebuilder from scratch again and I found out I didnt use the right webhook flags while I generated it. When I added the right flags, I can see the changes added in the PROJECT file automatically. Thanks for pointing it out. I have added a separate commit for this. PTAL again.

config/webhook/service.yaml
```diff
@@ -0,0 +1,13 @@
...
+metadata:
+  name: webhook-service
+  namespace: system
+spec:
+  ports:
+    - port: 443
+      protocol: TCP
+      targetPort: 9443
+  selector:
+    control-plane: controller-manager
```
fmuyassarov, line 13: don't we have it already in https://github.com/metal3-io/baremetal-operator/blob/master/config/webhook/service_patch.yaml  ?

kashifest: Indeed. Removing it.

kashifest:
I would really appreciate an lgtm again before I need to do rebase again :)

andfasano:
Thanks @kashifest!

/lgtm

PR #1014, opened 2021-11-04
Add BMC subscription CRD and reconciler

honza:
This patch implements the fast eventing API:

https://github.com/metal3-io/metal3-docs/pull/167

We introduce a new CRD, and a new reconciler.  We use the Ironic API to
communicate with the BMC to manage node subscriptions.


main.go
```diff
@@ -189,6 +189,15 @@ func main() {
 		os.Exit(1)
 	}
 
+	if err = (&metal3iocontroller.BMCEventSubscriptionReconciler{
```
s3rj1k, line 192: Can this be made conditional, enabled by CLI flag value?

apis/metal3.io/v1alpha1/bmceventsubscription_types.go
```diff
@@ -0,0 +1,92 @@
...
+	Destination string `json:"destination,omitempty"`
+
+	// Arbitrary user-provided context for the event
+	Context string `json:"context,omitempty"`
+
+	// Messages of which type we should pass along; defaults to ["Alert"]
+	EventTypes []string `json:"eventTypes,omitempty"`
+
+	// The BMC protocol to use; defaults to "Redfish"
+	Protocol string `json:"protocol,omitempty"`
```
bfournie, line 48: Should this be set explicitly to "redfish" in the controller if not set? Right now it just relies on using what user is setting here.

apis/metal3.io/v1alpha1/bmceventsubscription_types.go
```diff
@@ -0,0 +1,92 @@
...
+	HostRef string `json:"hostRef,omitempty"`
+
+	// A webhook URL to send events to
+	Destination string `json:"destination,omitempty"`
+
+	// Arbitrary user-provided context for the event
+	Context string `json:"context,omitempty"`
+
+	// Messages of which type we should pass along; defaults to ["Alert"]
+	EventTypes []string `json:"eventTypes,omitempty"`
```
bfournie, line 45: Could validation be done to ensure valid types?

apis/metal3.io/v1alpha1/bmceventsubscription_types.go
```diff
@@ -0,0 +1,92 @@
...
+	// +kubebuilder:default:=0
+	ErrorCount int `json:"errorCount"`
+	// LastUpdated identifies when this status was last observed.
+	// +optional
+	LastUpdated *metav1.Time `json:"lastUpdated,omitempty"`
+}
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+//
+// BMCEventSubscription is the Schema for the ...
```
bfournie, line 68: Nit - missing comment

apis/metal3.io/v1alpha1/bmceventsubscription_types.go
bfournie: I think if there is a minimal validation that can be done it should be done here, similar to what we do for baremetalhosts and hostfirmwaresettings, in order to provide feedback to the user instead of relying on Ironic or the BMC to fail the configuration.

apis/metal3.io/v1alpha1/bmceventsubscription_types.go
bfournie: Again, it would be more useful to validate here and provide feedback instead of relying on Ironic to fail the config.

apis/metal3.io/v1alpha1/bmceventsubscription_types.go
```diff
@@ -0,0 +1,92 @@
...
+	// subscriptions to block delete operations until the subscription is removed
+	// from the BMC.
+	BMCEventSubscriptionFinalizer string = "bmceventsubscription.metal3.io"
+)
+
+type BMCEventSubscriptionSpec struct {
+	Taints []corev1.Taint `json:"taints,omitempty"`
+
+	// A reference to a BareMetalHost
+	HostRef string `json:"hostRef,omitempty"`
```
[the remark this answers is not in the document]
honza: I'm wondering if we should change this to be a `corev1.ObjectReference`.

apis/metal3.io/v1alpha1/bmceventsubscription_types.go
```diff
@@ -0,0 +1,92 @@
...
+
+	// A webhook URL to send events to
+	Destination string `json:"destination,omitempty"`
```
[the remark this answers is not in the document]
honza: We can check that it's non-empty, and we can run it through `net/url.Parse`, and maybe check that it has a trailing slash?

controllers/metal3.io/bmceventsubscription_controller.go
```diff
@@ -0,0 +1,282 @@
...
+		return headers, err
+	}
+
+	for headerName, headerValueBytes := range secret.Data {
+		header := map[string]string{}
+		header[headerName] = string(headerValueBytes)
+		headers = append(headers, header)
+	}
+
+	return headers, err
```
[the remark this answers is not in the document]
honza: I don't think we want to take over the user-provided secret. If we were creating a custom CR derived from a secret, I'd say sure, we want to set the owner ourselves, and clean up later. What if the same secret is reused? What if it's temporarily unused?

controllers/metal3.io/bmceventsubscription_controller.go
```diff
@@ -0,0 +1,278 @@
...
+			// garbage collected. For additional cleanup logic use
+			// finalizers.  Return and don't requeue
+			return ctrl.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		return ctrl.Result{}, errors.Wrap(err, "could not load subscription")
+	}
+
+	if subscription.Spec.HostRef == "" {
+		return ctrl.Result{}, errors.New("Missing HostRef")
```
andfasano, line 70: This looks like a critical problem (a webhook could be perfect for that), and it could be useful to show it to user - otherwise there wouldn't be no clue about it if not looking in the logs. Given that it could be something that user may fix, I'd suggest to use re-evaluate the request after a while. Note that returning an error will produce an immediate re-enqueue (that will get rate-limited).

pkg/provisioner/ironic/ironic.go
```diff
@@ -1867,3 +1868,40 @@ func (p *ironicProvisioner) loadBusyHosts() (hosts map[string]struct{}, err erro
...
+	method := nodes.CallVendorPassthruOpts{
+		Method: "delete_subscription",
+	}
+	opts := nodes.DeleteSubscriptionOpts{
+		Id: subscription.Status.SubscriptionID,
+	}
+	err = nodes.DeleteSubscription(p.client, p.nodeID, method, opts).ExtractErr()
+
+	if err != nil {
+		return retryAfterDelay(subscriptionRequeueDelay)
```
andfasano, line 1904: `err` is not sent back upwards here

pkg/provisioner/ironic/ironic.go
```diff
@@ -1867,3 +1868,40 @@ func (p *ironicProvisioner) loadBusyHosts() (hosts map[string]struct{}, err erro
...
+		},
+		nodes.CreateSubscriptionOpts{
+			Destination: subscription.Spec.Destination,
+			Context:     subscription.Spec.Context,
+			Protocol:    subscription.Spec.Protocol,
+			EventTypes:  subscription.Spec.EventTypes,
+			HttpHeaders: httpHeaders,
+		}).Extract()
+	if err != nil {
+		return retryAfterDelay(subscriptionRequeueDelay)
```
andfasano, line 1887: `err` is not sent back upwards here

controllers/metal3.io/bmceventsubscription_controller.go
```diff
@@ -0,0 +1,278 @@
...
+	"github.com/metal3-io/baremetal-operator/pkg/utils"
+)
+
+type BMCEventSubscriptionReconciler struct {
+	client.Client
+	Log                logr.Logger
+	ProvisionerFactory provisioner.Factory
+}
+
+func (r *BMCEventSubscriptionReconciler) Reconcile(ctx context.Context, request ctrl.Request) (result ctrl.Result, err error) {
```
andfasano, line 43: This reconcile loop is pretty big. I'd suggest to start refactoring logical blocks in (small) separate functions, to make the flow more readable

controllers/metal3.io/bmceventsubscription_controller.go
```diff
@@ -0,0 +1,278 @@
...
+	reconcileSubscriptionCounters.With(subscriptionMetricLabels(request)).Inc()
+	defer func() {
+		if err != nil {
+			reconcileSubscriptionErrorCounter.Inc()
```
andfasano, line 47: What's the intended usage for this counter?

controllers/metal3.io/bmceventsubscription_controller.go
```diff
@@ -0,0 +1,278 @@
...
+		}
+	}()
+
+	reqLogger := r.Log.WithValues("bmceventsubscription", request.NamespacedName)
+	reqLogger.Info("start")
```
andfasano, line 52: Personally, I'd appreciate also to have a "done" trace :)

controllers/metal3.io/bmceventsubscription_controller.go
```diff
@@ -0,0 +1,278 @@
...
+	var bmcCreds *bmc.Credentials
+	var bmcCredsSecret *corev1.Secret
+
+	bmcCreds, bmcCredsSecret, err = r.buildAndValidateBMCCredentials(request, host)
+	if err != nil || bmcCreds == nil {
+		return ctrl.Result{}, errors.Wrap(err, "failed to get bmc secret")
+	}
+
+	initialState := host.Status.Provisioning.State
+	info := &reconcileInfo{
```
andfasano, line 119: it seems that `initialState` and `info` could be avoided here?

controllers/metal3.io/bmceventsubscription_controller.go
```diff
@@ -0,0 +1,278 @@
...
+	}
+	if !ready {
+		reqLogger.Info("provisioner is not ready", "RequeueAfter:", provisionerNotReadyRetryDelay)
+		return ctrl.Result{Requeue: true, RequeueAfter: provisionerNotReadyRetryDelay}, nil
+	}
+
+	httpHeaders, err := r.getHTTPHeaders(*subscription)
+
+	if err != nil {
+		return ctrl.Result{}, errors.Wrap(err, "failed to get http headers")
```
andfasano, line 143: Same here, for the error management

controllers/metal3.io/bmceventsubscription_controller.go
```diff
@@ -0,0 +1,282 @@
...
+func (r *BMCEventSubscriptionReconciler) buildAndValidateBMCCredentials(request ctrl.Request, host *metal3v1alpha1.BareMetalHost) (bmcCreds *bmc.Credentials, bmcCredsSecret *corev1.Secret, err error) {
+	// Retrieve the BMC secret from Kubernetes for this host
+	bmcCredsSecret, err = r.getBMCSecret(request, host)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// Check for a "discovered" host vs. one that we have all the info for
+	// and find empty Address or CredentialsName fields
+	if host.Spec.BMC.Address == "" {
```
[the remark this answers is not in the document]
andfasano: That's a good point, but I'd leave all the burden to validate the secret just to the BMH controller (which also caches it), and simplify the code here if possible

controllers/metal3.io/bmceventsubscription_controller.go
```diff
@@ -0,0 +1,278 @@
...
+		Namespace: request.Namespace,
+	}
+	err = r.Get(ctx, namespacedHostName, host)
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			// Request object not found, could have been deleted after
+			// reconcile request.  Owned objects are automatically
+			// garbage collected. For additional cleanup logic use
+			// finalizers.  Return and don't requeue
+			return ctrl.Result{}, nil
```
andfasano, line 85: This doesn't sound right to me. There could be a genuine mistake in the `hostRef` field, so in this way the subscription will never start and the user will never receive a warning (or even in the log). I guess that an ErrorMessage could be useful to manage also this situation for informing the user, with a requeueing after a while to give the user the time to fix the issue.

controllers/metal3.io/bmceventsubscription_controller.go
```diff
@@ -0,0 +1,278 @@
...
+			if err != nil {
+				return ctrl.Result{}, errors.Wrap(err, "failed to update subscription status")
+			}
+
+		} else {
+			// Update an existing subscription by recreating it in the BMC
+
+			originalSubscription := subscription.DeepCopy()
+
+			if _, err := prov.AddBMCEventSubscriptionForNode(subscription, httpHeaders); err != nil {
```
andfasano, line 164: Q: Are all the update eligible to destroy/recreate the subscription?

controllers/metal3.io/bmceventsubscription_controller.go
honza: I added a webhook.

controllers/metal3.io/bmceventsubscription_controller.go
honza: Maybe @iurygregory knows?

iurygregory: @andfasano  We can only create/delete/list a subscription, update is not supported (only in very newer redfish versions that most of the HW vendors are not using in their BMCs)

apis/metal3.io/v1alpha1/bmceventsubscription_types.go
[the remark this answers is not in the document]
dtantsur: I think it's worth validating it before we hand it off to Ironic. On top of that, why `omitempty`, isn't it required?

apis/metal3.io/v1alpha1/bmceventsubscription_types.go
dtantsur: Do we even need this field now given that "Redfish" is the only supported value?

apis/metal3.io/v1alpha1/bmceventsubscription_validation.go
```diff
@@ -0,0 +1,66 @@
...
+		errs = append(errs, fmt.Errorf("HostName cannot be empty"))
+	}
+
+	if s.Spec.Destination == "" {
+		errs = append(errs, fmt.Errorf("Destination cannot be empty"))
+	} else {
+		destinationUrl, err := url.ParseRequestURI(s.Spec.Destination)
+
+		if err != nil {
+			errs = append(errs, fmt.Errorf("Destination is an invalid URL"))
```
dtantsur, line 44: Let's add the actual error to the message

apis/metal3.io/v1alpha1/bmceventsubscription_validation.go
```diff
@@ -0,0 +1,66 @@
...
+		} else {
+			if !strings.HasSuffix(destinationUrl.Path, "/") {
+				errs = append(errs, fmt.Errorf("Destination must have a trailing slash"))
```
dtantsur, line 47: This is a weird requirement, does it come from Redfish?

controllers/metal3.io/bmceventsubscription_controller.go
dtantsur: +1

controllers/metal3.io/bmceventsubscription_controller.go
```diff
@@ -0,0 +1,260 @@
...
+
+	host := &metal3v1alpha1.BareMetalHost{}
+	namespacedHostName := types.NamespacedName{
+		Name:      subscription.Spec.HostName,
+		Namespace: request.Namespace,
+	}
+	err = r.Get(ctx, namespacedHostName, host)
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			reqLogger.Error(err, "baremetalhost not found")
```
dtantsur, line 82: nit: log the host name

controllers/metal3.io/bmceventsubscription_controller.go
```diff
@@ -0,0 +1,260 @@
...
+		return ctrl.Result{Requeue: true}, nil
+	}
+
+	// Retrieve the BMC details from the host spec and validate host
+	// BMC details and build the credentials for talking to the
+	// management controller.
+	var bmcCreds *bmc.Credentials
+
+	bmcCreds, _, err = r.buildAndValidateBMCCredentials(request, host)
+	if err != nil || bmcCreds == nil {
```
dtantsur, line 111: We shouldn't return bmcCreds == nil without also returning an err, so the 2nd check seems redundant

controllers/metal3.io/bmceventsubscription_controller.go
```diff
@@ -0,0 +1,260 @@
...
+		} else {
+			// Update an existing subscription by recreating it in the BMC
+
+			originalSubscription := subscription.DeepCopy()
+
+			if _, err := prov.AddBMCEventSubscriptionForNode(subscription, httpHeaders); err != nil {
+				return ctrl.Result{}, errors.Wrap(err, "failed to update subscription")
+			}
+
+			if _, err := prov.RemoveBMCEventSubscriptionForNode(*originalSubscription); err != nil {
```
dtantsur, line 158: I don't know how many simultaneous subscriptions a BMC allows, I'd rather delete the old subscription *before* creating a new one

controllers/metal3.io/bmceventsubscription_controller.go
```diff
@@ -0,0 +1,260 @@
...
+		return ctrl.Result{}, nil
+	}
+
+	return
+
+}
+
+func (r *BMCEventSubscriptionReconciler) getBMCSecret(request ctrl.Request, host *metal3v1alpha1.BareMetalHost) (bmcCredsSecret *corev1.Secret, err error) {
+	secretKey := host.CredentialsKey()
+	bmcCredsSecret = &corev1.Secret{}
```
dtantsur, line 191: nit: why return a pointer? you dereference it eventually anyway (and `nil` should not be an option)

controllers/metal3.io/bmceventsubscription_controller.go
```diff
@@ -0,0 +1,260 @@
...
+			return nil, &ResolveBMCSecretRefError{message: fmt.Sprintf("The BMC secret %s does not exist", secretKey)}
+		}
+		return nil, err
+	}
+
+	return bmcCredsSecret, nil
+}
+
+func (r *BMCEventSubscriptionReconciler) getHTTPHeaders(subscription metal3v1alpha1.BMCEventSubscription) ([]map[string]string, error) {
+	headers := []map[string]string{}
```
dtantsur, line 204: Why not one map? Technically, it will be a map on wire anyway.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1867,3 +1868,40 @@ func (p *ironicProvisioner) loadBusyHosts() (hosts map[string]struct{}, err erro
...
+		},
+		nodes.CreateSubscriptionOpts{
+			Destination: subscription.Spec.Destination,
+			Context:     subscription.Spec.Context,
+			Protocol:    subscription.Spec.Protocol,
+			EventTypes:  subscription.Spec.EventTypes,
+			HttpHeaders: httpHeaders,
+		}).Extract()
+	if err != nil {
+		return provisioner.Result{RequeueAfter: subscriptionRequeueDelay}, err
```
dtantsur, line 1887: What's the point of requeueing on failure? At the very least we should check which error it does: it makes no sense to retry a subscription for a wrong driver.

apis/metal3.io/v1alpha1/bmceventsubscription_validation.go
honza: Yes

iurygregory: I think you only need to have the `/` in the end in some cases, otherwise this wouldn't work `https://<IP>/RedfishEvents/EventReceiver.php` right?  Only if you are trying the destination as `https://<IP>` you would need the extra `/`

https://developer.hpe.com/blog/the-redfish-event-service/

controllers/metal3.io/bmceventsubscription_controller.go
honza: This is the type that gophercloud expects.

iurygregory: Just a reminder, Redfish won't return any information about the headers you used to create a subscription when you retrieve information about the subscription you created 

i.e you are creating a subscription with "Content-Type: application/json", the subscription is created, if you try to retrieve information about this subscription you won't have information that it had headers

controllers/metal3.io/bmceventsubscription_controller.go
```diff
@@ -0,0 +1,260 @@
...
+		return headers, nil
+	}
+
+	secret := &corev1.Secret{}
+	secretKey := types.NamespacedName{
+		Name:      subscription.Spec.HTTPHeadersRef.Name,
+		Namespace: subscription.Spec.HTTPHeadersRef.Namespace,
+	}
+
+	err := r.Get(context.TODO(), secretKey, secret)
```
andfasano, line 216: Note: I'm not sure if it could have any impact here, but the main manager has a cache selector to avoid caching all the secrets, so I guess also such secret should have the environment label (in BMO there's a `SecretManager` [class](https://github.com/metal3-io/baremetal-operator/blob/c22cb7ba6df61e7a4b3bae751f3b36b0ae825f1c/pkg/secretutils/secret_manager.go#L34) to keep care of it, probably worth checking)

main.go
```diff
@@ -110,6 +116,8 @@ func main() {
 		"The address the health endpoint binds to.")
 	flag.IntVar(&webhookPort, "webhook-port", 9443,
 		"Webhook Server port (set to 0 to disable)")
+	flag.BoolVar(&enableSubscriptions, "enable-subscriptions", false,
```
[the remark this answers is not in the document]
honza: It was requested by @s3rj1k above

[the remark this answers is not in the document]
andfasano: If you refer to this [comment](https://github.com/metal3-io/baremetal-operator/pull/1014#discussion_r743112217) then the motivation it's still unclear to me (and it didn't seem to me there were any specific reasons to make it conditional in the [design doc](https://github.com/metal3-io/metal3-docs/pull/167)?).

controllers/metal3.io/bmceventsubscription_controller.go
honza: Following the lead from [our old bmh controller implementation](https://github.com/metal3-io/baremetal-operator/blob/388bcf6f7de45c1889b89918c29ed1dbb06a7108/controllers/metal3.io/baremetalhost_controller.go#L1218)

Should we refactor this to use the newish secret manager?

controllers/metal3.io/bmceventsubscription_controller.go
honza: We can introduce a new type here, and move the formatting into a map further down the stack.

pkg/provisioner/ironic/ironic.go
honza: Retrying gives the user the chance to fix the subscription CR. We should definitely be logging the error though.

controllers/metal3.io/bmceventsubscription_controller.go
dtantsur: If the list of maps is what Ironic expects, then I'm fine with keeping it.

pkg/provisioner/ironic/ironic.go
dtantsur: I guess I'm worried about meaningless retries in case of e.g. unsupported driver (which is not possible to fix without recreating the BMH).

controllers/metal3.io/bmceventsubscription_controller.go
honza: I just refactored this a bit, let me know if you think I should go at it harder :)

honza:
I just found a bug in the create or update method; I'll push a fix ASAP.
/hold

controllers/metal3.io/bmceventsubscription_controller.go
honza: I made them immutable. It simplifies the business logic in the reconciler while the webhook prevents updates.

apis/metal3.io/v1alpha1/bmceventsubscription_validation.go
```diff
@@ -0,0 +1,72 @@
...
+			}
+		}
+	}
+
+	return errs
+}
+
+// validateChanges validates BMCEventSubscription resource on changes
+// but also covers the validations of creation
+func (s *BMCEventSubscription) validateChanges(old *BMCEventSubscription) []error {
```
ardaguclu, line 63: AFAICS, update is not allowed for this CRD and validateChanges is not used. Maybe we can remove this?

config/webhook/manifests.yaml
```diff
@@ -27,3 +27,24 @@ webhooks:
     resources:
     - baremetalhosts
   sideEffects: None
+- admissionReviewVersions:
+  - v1
+  - v1beta
+  clientConfig:
+    service:
+      name: webhook-service
```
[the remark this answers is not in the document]
ardaguclu: I'm also curious about this. Same service and same port would redirect validation request according to the path?.

main.go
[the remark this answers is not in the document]
ardaguclu: If webhook will use same service and port, it'd be needed to check webhook-port is 0 or not for local runs.

ardaguclu (commented):
I have few comments mostly around webhook. Thanks @honza

config/webhook/manifests.yaml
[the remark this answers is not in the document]
honza: Yes, I think this is the name of webhook service. The name of the resource is specified below.

main.go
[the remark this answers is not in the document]
honza: > If webhook will use same service and port, it'd be needed to check webhook-port is 0 or not for local runs.

I think this is already taken care of by the `if enableWebhook` check further down?

> If you refer to this comment then the motivation it's still unclear to me (and it didn't seem to me there were any specific reasons to make it conditional in the design doc?).

Is there a more general heuristic to decide which controllers should be enabled/disable by default?

controllers/metal3.io/bmceventsubscription_controller.go
```diff
@@ -0,0 +1,283 @@
...
+}
+
+func (r *BMCEventSubscriptionReconciler) getProvisioner(request ctrl.Request, host *metal3v1alpha1.BareMetalHost) (prov provisioner.Provisioner, ready bool, err error) {
+	reqLogger := r.Log.WithValues("bmceventsubscription", request.NamespacedName)
+
+	bmcCreds, _, err := r.buildAndValidateBMCCredentials(request, host)
+	if err != nil {
+		return prov, ready, errors.Wrap(err, "failed to get bmc secret")
+	}
+
```
bfournie, line 189: Since this controller only communicates with the Ironic APi via gophercloud and doesn't need to access the BMC directly (for power cycle etc), I'd suggest we create the provisioner without BMC info, similar to what we do for the hostFirmwareSettings controller (https://github.com/metal3-io/baremetal-operator/blob/master/controllers/metal3.io/hostfirmwaresettings_controller.go#L139). This will simplify creation of the provisioner and will no longer need buildAndValidateBMCCredentials().

config/webhook/manifests.yaml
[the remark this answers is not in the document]
andfasano: Ok thanks

controllers/metal3.io/bmceventsubscription_controller.go
andfasano: Maybe also the finalizer block could deserve a little bit of refactoring, for the rest looks better now thanks

controllers/metal3.io/bmceventsubscription_controller.go
```diff
@@ -0,0 +1,283 @@
...
+	host := &metal3v1alpha1.BareMetalHost{}
+	namespacedHostName := types.NamespacedName{
+		Name:      subscription.Spec.HostName,
+		Namespace: request.Namespace,
+	}
+	err = r.Get(ctx, namespacedHostName, host)
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			reqLogger.Error(err, "baremetalhost not found", "host", host.Name)
+			return ctrl.Result{RequeueAfter: subscriptionRetryDelay}, nil
```
andfasano, line 85: I'm still wondering if it'd be better to report this "pending" status to the user, to let him/her know what's currently going on (it could be a temporary situation, or maybe a longer one). This because the absence of a BMH _could indicate_ (but not necessarily) that the underlying host has gone, and such case I assume the subscription will not be anymore active.

main.go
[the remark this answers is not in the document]
honza: Remove

controllers/metal3.io/bmceventsubscription_controller.go
```diff
@@ -0,0 +1,303 @@
...
+	for headerName, headerValueBytes := range secret.Data {
+		header := map[string]string{}
+		header[headerName] = string(headerValueBytes)
+		headers = append(headers, header)
+	}
+
+	return headers, err
+}
+
+func (r *BMCEventSubscriptionReconciler) buildAndValidateBMCCredentials(request ctrl.Request, host *metal3v1alpha1.BareMetalHost) (bmcCreds *bmc.Credentials, bmcCredsSecret *corev1.Secret, err error) {
```
bfournie, line 275: The fix to just use the provisioner for Ironic APi and not need the BMC info looks good, you should also be able remove this function.

controllers/metal3.io/bmceventsubscription_controller.go
```diff
@@ -0,0 +1,303 @@
...
+	// Retrieve the BMC secret from Kubernetes for this host
+	bmcCredsSecret, err = r.getBMCSecret(request, host)
```
bfournie, line 277: Can also remove getBMCSecret()

controllers/metal3.io/bmceventsubscription_controller.go
honza: :man_facepalming: thanks!

elfosardo (approved):

bfournie:
Seeing this error in the integration test:
`E1209 15:00:21.061725       1 reflector.go:138] pkg/mod/k8s.io/client-go@v0.21.4/tools/cache/reflector.go:167: Failed to watch *v1alpha1.BMCEventSubscription: failed to list *v1alpha1.BMCEventSubscription: bmceventsubscriptions.metal3.io is forbidden: User "system:serviceaccount:baremetal-operator-system:baremetal-operator-controller-manager" cannot list resource "bmceventsubscriptions" in API group "metal3.io" at the cluster scope`

I believe you need to add rbac settings for the bmceventsubscription to the new controller, for example, similar to this:
//+kubebuilder:rbac:groups=metal3.io,resources=hostfirmwaresettings,verbs=get;list;watch;create;update;patch;delete
//+kubebuilder:rbac:groups=metal3.io,resources=hostfirmwaresettings/status,verbs=get;update;patch

bfournie (commented):
Its looks like the same rbac related errors are still being logged, you may need to run 'make manifests' as I would expect to see an additional file and changes to role.yaml in config/rbac for the resource (e.g. https://github.com/metal3-io/baremetal-operator/tree/master/config/rbac).

honza:
Running `make manifests` yields no changes. I just rebased on top of master to see if that does anything.

honza:
It looks like a line had to be added to the baremetal controller file as well.

apis/metal3.io/v1alpha1/bmceventsubscription_validation_test.go
```diff
@@ -0,0 +1,81 @@
...
+			},
+			oldS:      nil,
+			wantedErr: "Hostname-only destination must have a trailing slash",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if err := tt.newS.validateSubscription(); !errorArrContains(err, tt.wantedErr) {
+				t.Errorf("BareMetalHost.ValidateBareMetalHost() error = %v, wantErr %v", err, tt.wantedErr)
```
furkatgofurov7, line 77: should this be ValidateSubscription() instead?

apis/metal3.io/v1alpha1/bmceventsubscription_webhook.go
```diff
@@ -0,0 +1,54 @@
...
+
+	"k8s.io/apimachinery/pkg/util/errors"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+)
+
+// log is for logging in this package.
```
furkatgofurov7, line 26: nit: `bmcsubscriptionlog`

controllers/metal3.io/bmceventsubscription_controller.go
```diff
@@ -0,0 +1,271 @@
...
+
+type BMCEventSubscriptionReconciler struct {
+	client.Client
+	Log                logr.Logger
+	ProvisionerFactory provisioner.Factory
+	APIReader          client.Reader
+}
+
+//+kubebuilder:rbac:groups=metal3.io,resources=bmceventsubscriptions,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=metal3.io,resources=bmceventsubscriptions/status,verbs=get;update;patch
```
furkatgofurov7, line 53: this seems has no effect, since I do not see the rbac rules for `bmceventsubscriptions/status`?

pkg/provisioner/provisioner.go
```diff
@@ -188,6 +190,10 @@ type Provisioner interface {
 
 	// GetFirmwareSettings gets the BIOS settings and optional schema from the host and returns maps
 	GetFirmwareSettings(includeSchema bool) (settings metal3v1alpha1.SettingsMap, schema map[string]metal3v1alpha1.SettingSchema, err error)
+
+	// AddBMCEventSubscriptionForNode creates the subscription, and updates Status.SubscriptionID
+	AddBMCEventSubscriptionForNode(subscription *metal3v1alpha1.BMCEventSubscription, httpHeaders HTTPHeaders) (result Result, err error)
+	RemoveBMCEventSubscriptionForNode(subscription metal3v1alpha1.BMCEventSubscription) (result Result, err error)
```
furkatgofurov7, line 196: nit: missing a comment

honza:
/test-v1b1-centos-integration
/test-v1b1-integration

Previous run didn't show any RBAC issues.

bfournie (approved):

elfosardo (approved):

apis/metal3.io/v1alpha1/bmceventsubscription_types.go
andfasano: @honza is still really needed?

apis/metal3.io/v1alpha1/bmceventsubscription_validation.go
```diff
@@ -0,0 +1,59 @@
...
+		destinationUrl, err := url.ParseRequestURI(s.Spec.Destination)
+
+		if err != nil {
+			errs = append(errs, errors.Wrap(err, "Destination is an invalid URL"))
+		} else {
+			// Require a trailing slash on a domain without a path:
+			//
+			// Good: https://localhost:1234/index.php
+			// Good: https://localhost:1234/
+			// Bad: https://localhost:1234
```
andfasano, line 51: nit: these comments seem already covered by the tests

andfasano (commented):
Thanks @honza for the latest improvements. There is just one point about handling the delete, but for the rest the changes now look definitely better and almost all the points have been addressed.

I have just one about the testing: I think that it would be still worth following an approach like the `tryReconcile` in the [baremetal_controller_test.go](https://github.com/metal3-io/baremetal-operator/blob/92ffa60c683d9fa2dcb1ae0e229449520eface61/controllers/metal3.io/baremetalhost_controller_test.go) for adding some coverage on the relevant tasks fo the reconcile loop (could be also on a followup)

controllers/metal3.io/bmceventsubscription_controller.go
```diff
@@ -0,0 +1,280 @@
...
+		return nil
+	}
+
+	headers, err := r.getHTTPHeaders(*subscription)
+
+	if err != nil {
+		reqLogger.Error(err, "failed to get http headers")
+		subscription.Status.Error = fmt.Sprintf("failed to retrieve HTTP headers secret")
+		updateErr := r.Status().Update(ctx, subscription)
+		if updateErr != nil {
```
andfasano, line 182: This will get shadowed in the main loop [call](https://github.com/metal3-io/baremetal-operator/pull/1014/files#diff-2e58d31f13465b9d9eae85cfdf77bfa0579b6d3cd553b1fc7c37162c4e1f6e22R121)

andfasano: Note: updating the status in case of a permanent error (ie wrong http header) will cause an hot loop. To avoid it, please add an update predicate function like the [one](https://github.com/metal3-io/baremetal-operator/blob/3f19cf4344fed69def3a167a86a1375405840cac/controllers/metal3.io/baremetalhost_controller.go#L1557-L1579) used in the BMH controller

andfasano (changes requested):
/hold

Looks like there's another problem with the update, @honza can you please address it?

controllers/metal3.io/bmceventsubscription_controller.go
```diff
@@ -0,0 +1,280 @@
...
+	if !ready {
+		reqLogger.Info("provisioner is not ready", "RequeueAfter:", provisionerNotReadyRetryDelay)
+		return ctrl.Result{RequeueAfter: provisionerNotReadyRetryDelay}, nil
+	}
+
+	if subscription.DeletionTimestamp.IsZero() {
+		// Not being deleted
+		if err := r.createSubscription(ctx, prov, subscription); err != nil {
+			subscription.Status.Error = fmt.Sprintf("failed to create a subscription")
+			err := r.Status().Update(ctx, subscription)
```
andfasano, line 122: This `err` will cover the previous one at line 120, so in case the create subscription will fail (but the update not) the reconcile loop will return without any error

controllers/metal3.io/bmceventsubscription_controller.go
```diff
@@ -0,0 +1,280 @@
...
+			if err != nil {
+				return ctrl.Result{}, errors.Wrap(err, "failed to update subscription status")
+			}
+			return ctrl.Result{}, errors.Wrap(err, "failed to create subscription")
+		}
+	} else {
+		// Being deleted
+		if err := r.deleteSubscription(ctx, prov, subscription); err != nil {
+			subscription.Status.Error = fmt.Sprintf("failed to delete subscription")
+			err := r.Status().Update(ctx, subscription)
```
andfasano, line 132: Same here

andfasano (changes requested):

andfasano:
@honza latest push fixed the mentioned problems. There is still one case that probably to be discussed/clarified further (deleting a subscription when the related host does not exist anymore) but it could be addressed in another PR.

/lgtm

PR #1017, opened 2021-11-09
Create new BMC module and revert ironic packages to main module

ardaguclu:
Previously created ironic module can not be imported by apis module,
thus bmc package can not be used for validations. This PR reverts
ironic packages back to main go module and creates new BMC module.

This BMC module will be the core module and must not import any BMO
related package to avoid circular dependencies. Main and apis go modules
will use this BMC module.

This PR is continuation of this https://github.com/metal3-io/baremetal-operator/pull/1007.


ardaguclu:
@zaneb @dtantsur could you please have a look at this change?

integration test failed due to docker rate limit issue, I'll retest later.

dtantsur (changes requested):
Soft -1 to moving ironic back, it's a step in the wrong direction

zhouhao3:
It seems that there is still no consensus on whether ironic packages should be reverted?
Is it possible to keep the ironic package and separate the bmc package? In this case, the bmc package can be used for validations.

ardaguclu:
> It seems that there is still no consensus on whether ironic packages should be reverted? Is it possible to keep the ironic package and separate the bmc package? In this case, the bmc package can be used for validations.

I'll ask this topic today's community meeting and let's see the other opinions about it.

dtantsur:
/lgtm

We've discussed this on the community meeting. While we recognize the desire to work on better structuring of the code, technical issues (mostly around Go modules) prevent the current model from being usable and solving the immediate problems.

Dockerfile
```diff
@@ -11,8 +11,8 @@ COPY apis/go.mod apis/go.mod
 COPY apis/go.sum apis/go.sum
 COPY hack/tools/go.mod hack/tools/go.mod
 COPY hack/tools/go.sum hack/tools/go.sum
-COPY pkg/ironic/go.mod pkg/ironic/go.mod
-COPY pkg/ironic/go.sum pkg/ironic/go.sum
+COPY pkg/bmc/go.mod pkg/bmc/go.mod
```
zaneb, line 14: Is it worth putting the bmc package under something like `pkg/<something>/bmc` so that we can easily add more packages to the module if that makes sense in the future?

I would suggest 'hardware', but there is already a package with that name :(

zaneb (commented):
/lgtm

Dockerfile
s3rj1k: `bmc` package is highly dependent on ironic implementation itself, it does not do real BMC communication itself, so putting it to `pkg/bmc` could be somewhat confusing

ardaguclu: What do you think about pkg/hardwareutils/bmc? 

Hardwareutils will be the module.

zaneb: That WFM.

ardaguclu: Done.

ardaguclu:
@zaneb @dtantsur @s3rj1k could you please take a look one last time?, thanks.

Dockerfile
dtantsur: -1 to hardwareutils. what's the point of it? we can add more packages if we need it, the initial argument doesn't seem correct to me, unless we want to publish these packages to some global registry (in which case hardwareutils is also a terrible name).

dtantsur:
> @zaneb @dtantsur @s3rj1k could you please take a look one last time?, thanks.

I'm -1 now because of the rename.

Dockerfile
ardaguclu: What if we use `ironic` as module name. But currently it only includes bmc package for now. 

Thanks to that, we will have ability to move other ironic packages in this module time after time (after decoupling apis and provisioner dependency). That looks like a good path to create encapsulated ironic module as we intended.

zaneb: > -1 to hardwareutils. what's the point of it?

Well, the point is that we need an extra layer in the directory hierarchy and the name `hardware` is already taken.

> we can add more packages if we need it,

Note that we can't if the bmc package remains under the top level `/pkg/` dir and not one level deeper. Or we could, but they'd all appear under `pkg/bmc/...`.

> the initial argument doesn't seem correct to me, unless we want to publish these packages to some global registry (in which case hardwareutils is also a terrible name).

To be clear, the 'name' would be `github.com/metal3-io/baremetal-operator/pkg/hardwareutils` (there is no global registry as such though, although there is a global module cache).

dtantsur: Wait, so a directory does not become a package, only a parent directory? So if we make `pkg/bmc` a module it owns everything in `pkg`? O__o Please tell me you're kidding.

zaneb: No, but if we put go.mod in `pkg/bmc` then that is the module and any new packages in that module will appear like `pkg/bmc/somethingusedbyapibutnotbmcrelated`. So putting go.mod in `pkg/hardwareutils` and the bmc package in `pkg/hardwareutils/bmc` is more flexible because we can add `pkg/hardwareutils/somethingusedbyapibutnotbmcrelated` later.

dtantsur: I don't think I see the problem in creating a new module for `somethingusedbyapibutnotbmcrelated`, but I guess we don't want to end up in an infinite bike-shedding either.

ardaguclu: Agree. I don't want neither. Because interestingly this PR becomes bottleneck for independent different PRs.

ardaguclu: Thanks @dtantsur

ardaguclu:
There might be conflict with this https://github.com/metal3-io/baremetal-operator/pull/1041. Maybe we can wait this https://github.com/metal3-io/baremetal-operator/pull/1041 is merged first and after that this PR.

Dockerfile
zaneb: > I don't think I see the problem in creating a new module

New modules are a pain. We have to handle them in every Makefile target, and maintain replacement directives for all of them (with matching versions) in every downstream repo. The fewer we have, the better.

zaneb (commented):
/lgtm

PR #1019, opened 2021-11-09
Use PreprovisioningImage to build PXE images

zaneb:
If the PreprovisioningImage has networking configuration requirements that must be met prior to provisioning and are not guaranteed to be satisfied by the fact that the provisioning network is able to PXE boot, then the image must be customised even if it is to be PXE booted.

An example is when booting from a CoreOS image, we need to download the IPA container, and this is not necessarily available over the provisioning network.

This PR adds the ability to request either ISO or initrd formats for the PreprovisioningImage. A consequence of this is that PreprovisioningImage resources will now be created for _every_ Host (instead of just those using virtualmedia drivers) when the `-build-preprov-image` option is passed.


dtantsur:
/approve

The idea looks valid, please fix the UT

pkg/provisioner/ironic/ironic.go
```diff
@@ -554,14 +554,6 @@ func (p *ironicProvisioner) PreprovisioningImageFormats() ([]metal3v1alpha1.Imag
...
 	if accessDetails.SupportsISOPreprovisioningImage() {
 		formats = append(formats, metal3v1alpha1.ImageFormatISO)
-	} else {
-		if p.config.deployKernelURL != "" && p.config.deployRamdiskURL != "" {
-			// This is a PXE driver (no ISO support) so it shouldn't require any
-			// network customisation to boot the image, and we have sufficient
-			// data available to configure. Therefore, do not request an image
-			// build.
-			return nil, nil
-		}
```
zhouhao3, line 564: Why delete this code?

controllers/metal3.io/preprovisioningimage_controller.go
```diff
@@ -116,12 +116,26 @@ func (r *PreprovisioningImageReconciler) update(img *metal3.PreprovisioningImage
...
+				errorMessage = "No DEPLOY_ISO_URL specified"
+			}
+		case metal3.ImageFormatInitRD:
+			if initrd := os.Getenv("DEPLOY_RAMDISK_URL"); initrd != "" {
+				return initrd, fmt, ""
+			}
+			if errorMessage == "" {
+				format = fmt
+				errorMessage = "No DEPLOY_RAMDISK_URL specified"
+			}
```
zhouhao3, line 137: If there are two kinds of `acceptformats`, and both of them fail to obtain the URL, Is it bettert to return more detailed error information (e.g Neither DEPLOY_ISO_URL and DEPLOY_RAMDISK_URL specified ) ?  This is convenient for users to better understand the current situation.

zaneb: This controller won't start at all unless one of these is provided, so the case you mention can't happen:
https://github.com/metal3-io/baremetal-operator/pull/1019/files#diff-1d7da700aa97796ea2261b0c8bedf79a7eeb24ee0dd2fff6782b9b2811c9455dR238-R247

zhouhao3: Fine, lgtm

PR #1020, opened 2021-11-11
📖 Update old branch name mention in docs

zhouhao3:
Signed-off-by: Zhou Hao <zhouhao@fujitsu.com>

### DIFF
--- a/docs/dev-setup.md
+++ b/docs/dev-setup.md
@@ -172,8 +172,8 @@ and clean it with
 
 It is also possible to develop Baremetal Operator using Tilt with CAPM3. Please
 refer to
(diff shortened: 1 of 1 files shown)


furkatgofurov7:
/retitle 📖 Update old branch name mention in docs

andfasano:
I guess we could skip the integration tests for the doc files

furkatgofurov7:
> I guess we could skip the integration tests for the doc files

@andfasano  good point, since integration test is configured to be required in each PR, the only way we could skip it is by force merging as of now. 
We have a [configuration in prow](https://github.com/metal3-io/project-infra/blob/master/prow/config/config.yaml#L251-L253) for such a case and we can configure any prow job to skip the test if it did not touch certain files, but `/test-v1b1-integration` is run in Jenkins not Prow, the only mention of it in Prow is making sure it is run in every PR in the repo. Also, this job is set as required in Github UI in the repository settings.

PR #1021, opened 2021-11-11
Add dev to run_local_ironic.sh to setup customized env

yanndegat:


tools/run_local_ironic.sh
```diff
@@ -103,9 +110,11 @@ type: Opaque
 EOF
 fi
 
-sudo "${CONTAINER_RUNTIME}" pull "$IRONIC_IMAGE"
-sudo "${CONTAINER_RUNTIME}" pull "$IRONIC_INSPECTOR_IMAGE"
-sudo "${CONTAINER_RUNTIME}" pull "$IRONIC_KEEPALIVED_IMAGE"
+if [ "$CONTAINER_PULL_IMAGES" == "true" ]; then
+      sudo "${CONTAINER_RUNTIME}" pull "$IRONIC_IMAGE"
+      sudo "${CONTAINER_RUNTIME}" pull "$IRONIC_INSPECTOR_IMAGE"
+      sudo "${CONTAINER_RUNTIME}" pull "$IRONIC_KEEPALIVED_IMAGE"
```
hardys, line 116: Can you explain how you're testing with the pull disabled - presumably you're pre-pulling from some local registry?

I'm wondering why you can't just override IRONIC_IMAGE to reference some local image? (see lines 7-10)

yanndegat: hi @hardys 

that's it, you have to either build it locally or pre pull images.

if ever you only override IRONIC_IMAGE with some local image (line 7)
```
 sudo "${CONTAINER_RUNTIME}" pull "$IRONIC_IMAGE"
```
fails, because there's no existing "$IRONIC_IMAGE" on a remote registry.

hardys: Ack thanks for clarifying @yanndegat - this approach seems OK but FWIW I think a more common solution is to run a local registry, then you can build/push to that (and mirror any other needed images) similar to how we test via https://github.com/metal3-io/metal3-dev-env

Assuming that isn't an option for you, this new variable seems reasonable.

kashifest:
/hold

kashifest:
/hold cancel

PR #1036, opened 2021-11-22
Make PreprovisioningImage controller more generic

zaneb:
Create a new ImageProvider interface that the controller uses to obtain the image. The default implementation just obtains the URLs from environment variables as before, but is much easier to swap out for a new implementation than writing a separate controller.

Since custom controllers may need to cache data for each image, the controller needs to notify the image provider when a possibly cached image is no longer required. To this end, add a finalizer so that a cached image is cleared before a resource is deleted, and also mark the resource not ready when the config changes so that the existing image cache can be dropped.

### DIFF
--- a/controllers/metal3.io/preprovisioningimage_controller.go
+++ b/controllers/metal3.io/preprovisioningimage_controller.go
@@ -175,7 +175,7 @@ func (r *PreprovisioningImageReconciler) update(img *metal3.PreprovisioningImage
 		Format:            format,
 		Architecture:      img.Spec.Architecture,
 		NetworkDataStatus: secretStatus,
-	}, networkDataContent)
+	}, networkDataContent, log)
 	if err != nil {
 		return false, err
 	}
--- a/controllers/metal3.io/preprovisioningimage_controller.go
+++ b/controllers/metal3.io/preprovisioningimage_controller.go
@@ -21,6 +21,7 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
+	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
@@ -33,6 +34,7 @@ import (
 	metal3 "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
 	"github.com/metal3-io/baremetal-operator/pkg/imageprovider"
 	"github.com/metal3-io/baremetal-operator/pkg/secretutils"
+	"github.com/metal3-io/baremetal-operator/pkg/utils"
 )
 
 const (
@@ -74,6 +76,30 @@ func (r *PreprovisioningImageReconciler) Reconcile(ctx context.Context, req ctrl
 			err = nil
 		}
 		return ctrl.Result{}, err
+	}
+
+	if !img.DeletionTimestamp.IsZero() {
+		log.Info("cleaning up deleted resource")
+		if err := r.discardExistingImage(&img, log); err != nil {
+			return ctrl.Result{}, err
+		}
+		img.Finalizers = utils.FilterStringFromList(
+			img.Finalizers, metal3.PreprovisioningImageFinalizer)
+		err := r.Update(ctx, &img)
+		if err != nil {
+			return ctrl.Result{}, errors.Wrap(err, "failed to remove finalizer")
+		}
+		return ctrl.Result{}, nil
+	}
+
+	if !utils.StringInList(img.Finalizers, metal3.PreprovisioningImageFinalizer) {
+		log.Info("adding finalizer")
+		img.Finalizers = append(img.Finalizers, metal3.PreprovisioningImageFinalizer)
+		err := r.Update(ctx, &img)
+		if err != nil {
+			return ctrl.Result{}, errors.Wrap(err, "failed to add finalizer")
+		}
+		return ctrl.Result{}, nil
 	}
 
 	changed, err := r.update(&img, log)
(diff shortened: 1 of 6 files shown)


controllers/metal3.io/preprovisioningimage_controller.go
```diff
@@ -90,56 +117,104 @@ func (r *PreprovisioningImageReconciler) Reconcile(ctx context.Context, req ctrl
 	return result, err
 }
 
+func configChanged(img *metal3.PreprovisioningImage, format metal3.ImageFormat, networkDataStatus metal3.SecretStatus) bool {
```
ardaguclu, line 120: This function might be a method of `PreprovisioningImage`?

controllers/metal3.io/preprovisioningimage_controller.go
```diff
@@ -90,56 +117,104 @@ func (r *PreprovisioningImageReconciler) Reconcile(ctx context.Context, req ctrl
...
-			}
+func (r *PreprovisioningImageReconciler) getImageFormat(spec metal3.PreprovisioningImageSpec, log logr.Logger) (format metal3.ImageFormat) {
+	for _, acceptableFormat := range spec.AcceptFormats {
+		if r.ImageProvider.SupportsFormat(acceptableFormat) {
+			return acceptableFormat
 		}
 	}
+
+	if len(spec.AcceptFormats) > 0 {
+		log = log.WithValues("preferredFormat", spec.AcceptFormats[0])
```
ardaguclu, line 199: Should return after this?. Otherwise, it will log `no acceptable image format supported` in any case.

zaneb: That's the idea. If there were any acceptable formats supported, we'd have returned on line 194.

Note that this line doesn't actually log anything, it's just adding more data to the log message below. I'd have done it inline but it has to be conditional because getting `spec.AcceptFormats[0]` would fail if the slice is empty.

controllers/metal3.io/preprovisioningimage_controller.go
zaneb: Adding methods in the API module is discouraged. This one would be harmless enough (requires no dependencies), but it isn't actually likely to be used anywhere else so there's no reason to bend the rules.

controllers/metal3.io/preprovisioningimage_controller.go
```diff
@@ -214,6 +293,20 @@ func setImage(generation int64, status *metal3.PreprovisioningImageStatus, url s
...
+func setUnready(generation int64, status *metal3.PreprovisioningImageStatus, message string) bool {
+	newStatus := status.DeepCopy()
+
+	time := metav1.Now()
+	reason := reasonImageSuccess
+	setImageCondition(generation, newStatus,
+		metal3.ConditionImageReady, metav1.ConditionFalse,
+		time, reason, message)
+
+	changed := !apiequality.Semantic.DeepEqual(status, &newStatus)
```
bfournie, line 305: Just wondering why this is necessary. The return value isn't currently being used.

controllers/metal3.io/preprovisioningimage_controller.go
```diff
@@ -90,56 +117,104 @@ func (r *PreprovisioningImageReconciler) Reconcile(ctx context.Context, req ctrl
...
-		log.Info("network data Secret does not exist")
-		return setError(generation, &img.Status, reasonImageMissingNetworkData, "NetworkData secret not found"), err
+	if configChanged(img, format, secretStatus) {
+		reason := "Config changed"
+		if meta.IsStatusConditionTrue(img.Status.Conditions, string(metal3.ConditionImageReady)) {
+			// Ensure we mark the status as not ready before we remove the build
+			// from the image cache.
+			setUnready(generation, &img.Status, reason)
+		} else {
+			if img.Status.Format != "" {
```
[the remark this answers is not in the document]
zaneb: Aha, yep, good spotting.

controllers/metal3.io/preprovisioningimage_controller.go
zaneb: Just for consistency with the others. Originally I was using the return value, but since we only call this if it's going to return true I figured it would be clearer not to.

bfournie: Makes sense.

controllers/metal3.io/preprovisioningimage_controller.go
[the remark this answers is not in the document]
zaneb: Fixed.

bfournie (approved):

controllers/metal3.io/preprovisioningimage_controller.go
```diff
@@ -90,56 +117,102 @@ func (r *PreprovisioningImageReconciler) Reconcile(ctx context.Context, req ctrl
...
-	if k8serrors.IsNotFound(err) {
-		log.Info("network data Secret does not exist")
-		return setError(generation, &img.Status, reasonImageMissingNetworkData, "NetworkData secret not found"), err
+	if configChanged(img, format, secretStatus) {
+		reason := "Config changed"
+		if meta.IsStatusConditionTrue(img.Status.Conditions, string(metal3.ConditionImageReady)) {
+			// Ensure we mark the status as not ready before we remove the build
+			// from the image cache.
+			setUnready(generation, &img.Status, reason)
+		} else {
```
dtantsur, line 155: nit: shouldn't this code branch be executed in both cases, not only when the annotation is not present?

controllers/metal3.io/preprovisioningimage_controller.go
```diff
@@ -214,6 +291,20 @@ func setImage(generation int64, status *metal3.PreprovisioningImageStatus, url s
 	return changed
 }
 
+func setUnready(generation int64, status *metal3.PreprovisioningImageStatus, message string) bool {
```
dtantsur, line 294: nit: is unready actually a word? :)

controllers/metal3.io/preprovisioningimage_controller.go
zaneb: No, see comment on ll. 152-153.
This ensures that we are able to write to the CR that the image is not ready before we remove it. If we removed it first and then writing to the CR fails, we could end up in a state where the CR claims the URL is valid but it is in fact not for a possibly extended period of time.

controllers/metal3.io/preprovisioningimage_controller.go
zaneb: Only [since 1066](https://en.wiktionary.org/wiki/unready), although there was also [this guy](https://en.wikipedia.org/wiki/%C3%86thelred_the_Unready).

PR #1040, opened 2021-11-24
Validate Host DNS name in BareMetalHost resource

hroyrh:
Validate host dns name part of BareMetalHost bmc address.

### DIFF
--- a/apis/metal3.io/v1alpha1/baremetalhost_validation.go
+++ b/apis/metal3.io/v1alpha1/baremetalhost_validation.go
@@ -3,9 +3,7 @@ package v1alpha1
 import (
 	"fmt"
 	"net"
-	"net/url"
 	"regexp"
-	"strings"
 
 	"github.com/google/uuid"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
@@ -156,27 +154,9 @@ func validateDNSName(hostaddress string) error {
 		return nil
 	}
 
-	if !strings.Contains(hostaddress, "://") {
-		hostaddress = "ipmi://" + hostaddress
-	}
-
-	hurl, err := url.Parse(hostaddress)
+	_, err := bmc.GetParsedURL(hostaddress)
 	if err != nil {
-		return fmt.Errorf("Url parsing error, Host DNS name is invalid")
-	}
-
-	host, port, err := net.SplitHostPort(hurl.Host)
-
-	if port == "" {
-		if !strings.Contains(err.Error(), "missing port in address") {
-			return fmt.Errorf("Error while splitting host and port, Host DNS name is invalid")
-		}
-		host = hurl.Host
-	}
-	valid, _ := regexp.MatchString(`^([a-zA-Z]+\:\/\/)?(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])(\:[0-9]+)?$`, hostname)
-
-	if !valid {
-		return fmt.Errorf("Host DNS name is invalid")
+		return fmt.Errorf("host DNS name is invalid")
 	}
 
 	return nil
--- a/apis/metal3.io/v1alpha1/baremetalhost_validation.go
+++ b/apis/metal3.io/v1alpha1/baremetalhost_validation.go
@@ -6,6 +6,7 @@ import (
 	"regexp"
 
 	"github.com/google/uuid"
+	"github.com/pkg/errors"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 
 	"github.com/metal3-io/baremetal-operator/pkg/hardwareutils/bmc"
@@ -156,7 +157,7 @@ func validateDNSName(hostaddress string) error {
 
 	_, err := bmc.GetParsedURL(hostaddress)
 	if err != nil {
-		return fmt.Errorf("host DNS name is invalid")
+		return errors.Wrap(err, "BMO validation")
 	}
 
 	return nil
--- a/apis/metal3.io/v1alpha1/baremetalhost_validation.go
+++ b/apis/metal3.io/v1alpha1/baremetalhost_validation.go
@@ -3,7 +3,9 @@ package v1alpha1
 import (
 	"fmt"
 	"net"
+	"net/url"
 	"regexp"
+	"strings"
 
 	"github.com/google/uuid"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
@@ -35,6 +37,10 @@ func (host *BareMetalHost) validateHost() []error {
 	errs = append(errs, validateBMCAccess(host.Spec, bmcAccess)...)
 
 	if err := validateBMHName(host.Name); err != nil {
+		errs = append(errs, err)
+	}
+
+	if err := validateDNSName(host.Spec.BMC.Address); err != nil {
 		errs = append(errs, err)
 	}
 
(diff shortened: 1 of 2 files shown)


apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -53,3 +60,24 @@ func validateRAID(r *RAIDConfig) error {
 
 	return nil
 }
+
+func validateDNSName(hostaddress string) error {
+
+	if hostaddress == "" {
```
ardaguclu, line 66: Is this valid?

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -53,3 +60,24 @@ func validateRAID(r *RAIDConfig) error {
...
+		return nil
+	}
+
+	hurl, _ := url.Parse(hostaddress)
+	host, _, err := net.SplitHostPort(hurl.Host)
+
+	if err != nil {
```
ardaguclu, line 73: Maybe error is not nil actually means that host is invalid?

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -53,3 +60,24 @@ func validateRAID(r *RAIDConfig) error {
...
+		host = hurl.Host
+	}
+	valid, _ := regexp.MatchString(`^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])$`, host)
```
ardaguclu, line 76: I think, we should also handle error case in here instead hiding it.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -53,3 +60,24 @@ func validateRAID(r *RAIDConfig) error {
...
+
+	if !(valid) {
```
ardaguclu, line 78: There is no need for parenthesis in here.

apis/metal3.io/v1alpha1/baremetalhost_validation_test.go
```diff
@@ -71,6 +71,17 @@ func TestValidateCreate(t *testing.T) {
 			oldBMH:    nil,
 			wantedErr: "hardwareRAIDVolumes and softwareRAIDVolumes can not be set at the same time",
 		},
+		{
```
ardaguclu, line 74: I think, we can add more test like happy path or more edge cases.

ardaguclu (commented):
Thanks @hroyrh for the PR

apis/metal3.io/v1alpha1/baremetalhost_validation.go
hroyrh: This is to validate empty BareMetalHostSpec, but there could be a more precise check - maybe check if address is specified when Spec is not empty ?

apis/metal3.io/v1alpha1/baremetalhost_validation.go
hroyrh: Yes, I should probably check if the port is empty rather than the error.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
ardaguclu: I mean; if hostaddress is empty, should we return error or return nil

apis/metal3.io/v1alpha1/baremetalhost_validation.go
ardaguclu: İf error is not nil, we are setting `host = hurl.host`, why we are doing that instead returning an error?.

apis/metal3.io/v1alpha1/baremetalhost_validation_test.go
hroyrh: Okay, I will add a few more test cases.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
zaneb: I can potentially see some value in creating a Host with no address and then specifying the address later. I can't think of any particular use cases though... the BMO can't do anything until there is an address.
tl;dr I don't know :laughing:

ardaguclu: It might be safe to return nil then :)

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -53,3 +60,24 @@ func validateRAID(r *RAIDConfig) error {
...
 }
+
+func validateDNSName(hostaddress string) error {
+
+	if hostaddress == "" {
+		return nil
+	}
+
+	hurl, _ := url.Parse(hostaddress)
+	host, port, _ := net.SplitHostPort(hurl.Host)
```
ardaguclu, line 71: @hroyrh, if port is empty `host = hurl.Host` seems ok to me. But why we are not checking error in here?

hroyrh: It raises error for port empty too and as for the validity of the url, the regex checks that.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -18,6 +21,10 @@ func (host *BareMetalHost) validateHost() []error {
 		errs = append(errs, err)
 	}
 
+	if err := validateDNSName(host.Spec.BMC.Address); err != nil {
```
ardaguclu, line 24: @hroyrh hardwareutils/bmc package is added in this module. Maybe it would be useful using bmc's functionality for that validation if it is possible.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -53,3 +60,24 @@ func validateRAID(r *RAIDConfig) error {
 
 	return nil
 }
+
+func validateDNSName(hostaddress string) error {
```
ardaguclu, line 64: @hroyrh what I was thinking is something like that;

```
if hostaddress == "" {
		return nil
	}

	hurl, err := url.Parse(hostaddress)
	if err != nil {
		if !strings.Contains(hostaddress, ":") {
			return fmt.Errorf("Invalid format")
		}
	}
	host, port, err := net.SplitHostPort(hurl.Host)
	if err != nil {
		return fmt.Errorf("another invalid format")
	}

	if port == "" {
		host = hurl.Host
	}
	valid, err := regexp.MatchString(`^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])$`, host)
	if err != nil {
		return fmt.Errorf("regex error")
	}

	if !valid {
		return fmt.Errorf("Host DNS name is invalid")
	}

	return nil
```

apis/metal3.io/v1alpha1/baremetalhost_validation.go
dtantsur: I agree, the validation itself should probably go to the bmc package, if it's not there already

kashifest:
/hold

kashifest:
/hold cancel

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -143,3 +147,17 @@ func validateBMHName(bmhname string) error {
...
+
+	if hostaddress == "" {
+		return nil
+	}
+
+	_, err := bmc.GetParsedURL(hostaddress)
+	if err != nil {
+		return fmt.Errorf("host DNS name is invalid")
```
dtantsur, line 159: Could you include the original error using `errors.Wrap`?

hroyrh: @dtantsur I've wrapped the errors, and added a few more test cases, can you please take another look.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -143,3 +148,17 @@ func validateBMHName(bmhname string) error {
...
+
+func validateDNSName(hostaddress string) error {
+
+	if hostaddress == "" {
+		return nil
+	}
+
+	_, err := bmc.GetParsedURL(hostaddress)
+	if err != nil {
+		return errors.Wrap(err, "BMO validation")
```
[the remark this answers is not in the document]
hroyrh: Actually, the wrapped error already contains something very similar plus some other details, example error message : `BMO validation: failed to parse BMC address information: BMC address hostname/IP : [-host.example.com.org] is invalid` ( see [line 394](https://github.com/metal3-io/baremetal-operator/pull/1040/files#diff-c63bbb835e932967f86894fa9e1ca2d32aa663f19c2ce0712f53e0b65ebd3bf9R394) in the test file )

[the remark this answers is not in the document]
dtantsur: Nice!

PR #1046, opened 2021-12-08
Fix kustomize binary path in Tiltfile

Xenwar:
Fix kustomize binary path in Tiltfile

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -162,5 +162,5 @@ validate_auth()
 include_user_tilt_files()
 
 load_provider_tiltfiles(["."])
-local("make bin/kustomize")
+local("make tools/bin/kustomize")
 enable_provider("metal3-bmo")


Tiltfile
```diff
@@ -162,5 +162,5 @@ validate_auth()
 include_user_tilt_files()
 
 load_provider_tiltfiles(["."])
-local("make bin/kustomize")
+local("make tools/bin/kustomize")
```
furkatgofurov7, line 165: This is indeed a wrong path https://github.com/metal3-io/baremetal-operator/blob/master/Makefile#L21

furkatgofurov7 (commented):
/lgtm

/cc @andfasano @fmuyassarov @dtantsur

furkatgofurov7:
cc @zaneb

fmuyassarov (approved):

PR #1047, opened 2021-12-09
Add an interface for the image builder to report a failure

zaneb:
Some failures to build images may be permanent, as opposed to transient
errors. Define a special error type that allows the imageprovider to
indicate a permanent failure, that is then reported in the CR.

Other errors are assumed to be transient and will continue to cause a
retry.

### DIFF
--- a/pkg/imageprovider/invalid.go
+++ b/pkg/imageprovider/invalid.go
@@ -1,0 +1,19 @@
+package imageprovider
+
+import "fmt"
+
+type ImageBuildInvalid struct {
+	err error
+}
+
+func (ibf ImageBuildInvalid) Error() string {
+	return fmt.Sprintf("Cannot generate image: %s", ibf.err.Error())
+}
+
+func (ibf ImageBuildInvalid) Unwrap() error {
+	return ibf.err
+}
+
+func BuildInvalidError(err error) ImageBuildInvalid {
+	return ImageBuildInvalid{err: err}
+}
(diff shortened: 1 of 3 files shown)


zaneb:
Tests are broken due to https://github.com/metal3-io/project-infra/pull/284

pkg/imageprovider/invalid.go
```diff
@@ -0,0 +1,19 @@
+package imageprovider
+
+import "fmt"
+
+type ImageBuildInvalid struct {
+	err error
```
andfasano, line 6: Minor: from the current usage it looks like that this custom error is essentially string based, so rather than composing the `error` interface it could be enough to implement directly the interface (and also the creation will be simplified) - unless there will be a need to really wrap an error

pkg/imageprovider/invalid.go
```diff
@@ -0,0 +1,19 @@
...
+
+func (ibf ImageBuildInvalid) Error() string {
+	return fmt.Sprintf("Cannot generate image: %s", ibf.err.Error())
+}
+
+func (ibf ImageBuildInvalid) Unwrap() error {
+	return ibf.err
+}
+
+func InvalidError(err error) ImageBuildInvalid {
```
andfasano, line 17: `InvalidError` sounds a little bit too generic

furkatgofurov7 (commented):
LGTM

PR #1049, opened 2021-12-10
Remove reset_required from FirmwareSchema

bfournie:
The reset_required field in the FirmwareSchema is both misleading and unnecessary. Changing HostFirmwareSettings is accomplished by resetting the host so this field doesn't provide any extra info; in addition not all vendors implement this field.


dtantsur:
/approve

I assume it should be handled by Ironic, not by the BMO user.

andfasano:
@bfournie is there any risk of breaking change?

PR #1051, opened 2021-12-11
Handle HostFirmwareSettings non-numeric string set for Integer type

bfournie:
For HostFirmwareSettings, handle the case where a non-numeric string is set for an integer type.


zhouhao3:
lgtm

Issue #1053, opened 2021-12-14
Host secret is not removed during host delete


andfasano:
**Description**
When deleting an host, sometimes the host secret is not deleted and remains live. By looking in the logs, the following error is reported: 
```
ERROR   controller.baremetalhost        Reconciler error        {"reconciler group": "metal3.io", "reconciler kind": "BareMetalHost", "name": "ostest-extraworker-0", "namespace": "openshift-machine-api", "error": "action \"deleting\" failed: failed to remove finalizer from secret ostest-extraworker-0-bmc-secret in namespace openshift-machine-api: Operation cannot be fulfilled on secrets \"ostest-extraworker-0-bmc-secret\": the object has been modified;
 please apply your changes to the latest version and try again", "errorVerbose": "Operation cannot be fulfilled on secrets \"ostest-extraworker-0-bmc-secret\": the object has been modified; please apply your changes to the latest version and try again
failed to remove finalizer from secret ostest-extraworker-0-bmc-secret in namespace openshift-machine-api
github.com/metal3-io/baremetal-operator/pkg/secretutils.(*SecretManager).ReleaseSecret
	/baremetal-operator/go/src/github.com/metal3-io/baremetal-operator/pkg/secretutils/secret_manager.go:166
github.com/metal3-io/baremetal-operator/controllers/metal3%2eio.(*BareMetalHostReconciler).actionDeleting
	/baremetal-operator/go/src/github.com/metal3-io/baremetal-operator/controllers/metal3.io/baremetalhost_controller.go:530
github.com/metal3-io/baremetal-operator/controllers/metal3%2eio.(*hostStateMachine).handleDeleting
<snip>   
```
**What happens**
During the deletion of an host, at a certain point the current flow takes care of [removing the secret finalizer](https://github.com/metal3-io/baremetal-operator/blob/3f19cf4344fed69def3a167a86a1375405840cac/controllers/metal3.io/baremetalhost_controller.go#L531).
In some cases, the update on the secret is processed very quickly and produces an immediate reconciliation loop (BMO listens to owned secret events) while the host is still alive.
[...]

PR #1054, opened 2021-12-14
Do not add secret finalizer while deleting a host

andfasano:
Fixes https://github.com/metal3-io/baremetal-operator/issues/1053

### DIFF
--- a/controllers/metal3.io/baremetalhost_controller_test.go
+++ b/controllers/metal3.io/baremetalhost_controller_test.go
@@ -455,6 +455,48 @@ func TestAddFinalizers(t *testing.T) {
 	)
 }
 
+// TestDoNotAddSecretFinalizersDuringDelete verifies that during a host deletion,
+// in case the removal of a secret finalizer triggers an immediate reconcile loop,
+// then the secret finalizer is not added again
+func TestDoNotAddSecretFinalizersDuringDelete(t *testing.T) {
+	host := newDefaultHost(t)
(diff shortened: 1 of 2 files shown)


s3rj1k:
@andfasano nice catch, thanks!

andfasano:
> @andfasano nice catch, thanks!

Thanks. Unfortunately it happens intermittently, so it took some time to nail it down. I also tested it functionally (manually) and it seems ok.

controllers/metal3.io/baremetalhost_controller_test.go
```diff
@@ -455,6 +455,48 @@ func TestAddFinalizers(t *testing.T) {
 	)
 }
 
+// TestDoNotAddSecretFinalizersDuringDelete verifies that during an host deletion,
```
bfournie, line 458: Nit - 'an' should be 'a'

controllers/metal3.io/baremetalhost_controller_test.go
```diff
@@ -455,6 +455,48 @@ func TestAddFinalizers(t *testing.T) {
...
+// in case the removal of a secret finalizer triggers an immediate reconcile loop,
+// then the secret finalizer is not again
```
bfournie, line 460: Nit - seems like some text is missing here ("finalizer is not again")

bfournie (approved):

elfosardo (approved):

PR #1055, opened 2021-12-17
Update error message when soft poweroff fails

namnx228:
This PR updates the error message in case of soft poweroff failure. 
Currently, the workflow when turning off a machine, BMO tries soft poweroff first. If failed, it performs a hard poweroff. The problem is that after soft poweroff fails, BMO would publish an error message got from ironic conductor, something like: `Failed to change power state to 'power off' by 'soft power off'. Error: Redfish exception occurred. Error: Setting power state to soft power off failed for node <node id>...`. This message can make user confused and think that this is a serious error and it can interrupt the poweroff workflow. 
This PR aims to clarify a bit more about the workflow by informing user about the followed hard poweroff after the soft one failure.

### DIFF
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -59,6 +59,7 @@ const (
 	rebootAnnotationPrefix        = "reboot.metal3.io"
 	inspectAnnotationPrefix       = "inspect.metal3.io"
 	hardwareDetailsAnnotation     = inspectAnnotationPrefix + "/hardwaredetails"
+	clarifySoftPoweroffFailure    = "Continuing with hard poweroff after soft poweroff fails. More details: "
 )
 
 // BareMetalHostReconciler reconciles a BareMetalHost object
@@ -1218,6 +1219,10 @@ func (r *BareMetalHostReconciler) manageHostPower(prov provisioner.Provisioner, 
 	}
 
 	if provResult.ErrorMessage != "" {
+		if !desiredPowerOnState && desiredRebootMode == metal3v1alpha1.RebootModeSoft &&
+			info.host.Status.ErrorType != metal3v1alpha1.PowerManagementError {
+			provResult.ErrorMessage = clarifySoftPoweroffFailure + provResult.ErrorMessage
+		}
 		return recordActionFailure(info, metal3v1alpha1.PowerManagementError, provResult.ErrorMessage)
 	}
 
--- a/controllers/metal3.io/host_state_machine_test.go
+++ b/controllers/metal3.io/host_state_machine_test.go
@@ -874,61 +874,84 @@ func TestProvisioningCancelled(t *testing.T) {
 }
 
 func TestErrorCountIncreasedOnActionFailure(t *testing.T) {
-
+	defaultError := "some error"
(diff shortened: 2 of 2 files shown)


kashifest:
/hold cancel

kashifest:
/hold cancel

controllers/metal3.io/host_state_machine_test.go
```diff
@@ -942,7 +942,11 @@ func TestErrorCountIncreasedOnActionFailure(t *testing.T) {
 			result := hsm.ReconcileState(info)
 
 			assert.Equal(t, 1, tt.Host.Status.ErrorCount)
-			assert.Equal(t, "some error", tt.Host.Status.ErrorMessage)
+			if tt.ProvisionerErrorOn == "PowerOff" {
+				assert.Equal(t, clarifySoftPoweroffFailure+"some error", tt.Host.Status.ErrorMessage)
```
dtantsur, line 946: I suggest you add the expected error to scenarios above rather than calculating it (you can still use "some error" as the default).

namnx228: I understand that adding expected error to scenarios would make the testcases more customizable, and also make the test assertion code section looks clean. However, if we follow this approach, there will be two variables added to the testcase struct: inputError and expectedError. The reason is that when poweroff fails, the original error message will be modified by inserting the clarification message to the head of it. Therefore, the input and expected error messages will be different in this case. However, I feel like adding two more variables can make the test looks a bit more complicated than it should be. HDYT? @dtantsur

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1218,6 +1219,9 @@ func (r *BareMetalHostReconciler) manageHostPower(prov provisioner.Provisioner,
 	}
 
 	if provResult.ErrorMessage != "" {
+		if !desiredPowerOnState && desiredRebootMode == metal3v1alpha1.RebootModeSoft {
```
zaneb, line 1222: We don't attempt a soft power off if `info.host.Status.ErrorType == metal3v1alpha1.PowerManagementError`:
https://github.com/metal3-io/baremetal-operator/blob/main/pkg/provisioner/ironic/ironic.go#L1803
so we should also check that here.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1218,6 +1219,10 @@ func (r *BareMetalHostReconciler) manageHostPower(prov provisioner.Provisioner,
 	}
 
 	if provResult.ErrorMessage != "" {
+		if !desiredPowerOnState && desiredRebootMode == metal3v1alpha1.RebootModeSoft &&
+			info.host.Status.ErrorType != metal3v1alpha1.PowerManagementError {
+			provResult.ErrorMessage = clarifySoftPoweroffFailure + provResult.ErrorMessage
```
andfasano, line 1224: nit: if it's used just here a const could be avoided for an hardcoded Sprintf

namnx228: It is not only used here, but also in the unit test

fmuyassarov (approved):

PR #1059, opened 2021-12-28
Create operator namespace with default kustomize config

fmuyassarov:
Install `baremetal-operator-system` namespace and remove
the requirement for manual installation from dev setup doc.

### DIFF
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -13,6 +13,7 @@ namePrefix: baremetal-operator-
 #  someName: someValue
 
 bases:
+- ../namespace
 - ../crd
 - ../rbac
 - ../manager
--- a/docs/dev-setup.md
+++ b/docs/dev-setup.md
@@ -6,12 +6,6 @@
 
    <https://kubernetes.io/docs/setup/minikube/>
 
-1. Create a namespace to host the operator
-
-    ```bash
-    kubectl create namespace baremetal-operator-system
-    ```
-
 1. Install cert-manager
 
     ```bash


lentzi90 (approved):
:+1:

config/default/kustomization.yaml
```diff
@@ -13,6 +13,7 @@ namePrefix: baremetal-operator-
 #  someName: someValue
 
 bases:
+- ../namespace
```
ardaguclu, line 16: I wonder https://github.com/metal3-io/baremetal-operator/blob/24103e39f5596873ee5820663d05487fadeee81d/config/default/kustomization.yaml#L9 prefix will also apply namespace. As far as I recall, this is the reason namespace is not used in here.

config/render/capm3.yaml
```diff
@@ -1,3 +1,10 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    control-plane: controller-manager
```
ardaguclu, line 5: I'm asking just to learn; do we really need this label for namespace creation?

ardaguclu (commented):
Thanks for the PR @fmuyassarov, I left some comments.

config/default/kustomization.yaml
fmuyassarov: Hi @ardaguclu . Sorry for my late reply, I was on vacation for some time.
Actually we don't need namespace manifest below to be added in config/render/capm3.yaml, because we already have it in https://github.com/metal3-io/baremetal-operator/blob/main/config/namespace/namespace.yaml and here we are just reusing it. This is the same as in case of [basic-auth](https://github.com/metal3-io/baremetal-operator/blob/23493e6d94bbcff1c4da4623197674022dd5ccae/config/basic-auth/default/kustomization.yaml#L6) or [tls](https://github.com/metal3-io/baremetal-operator/blob/23493e6d94bbcff1c4da4623197674022dd5ccae/config/tls/kustomization.yaml#L6). So, it was required to install the namespace only when using config/default (non tls or basic-auth). 
About the label, we already have the label for all the resource like [Deployment](https://github.com/metal3-io/baremetal-operator/blob/23493e6d94bbcff1c4da4623197674022dd5ccae/config/manager/manager.yaml#L7) and other objects related to the BMO deployment including the [namespace](https://github.com/metal3-io/baremetal-operator/blob/23493e6d94bbcff1c4da4623197674022dd5ccae/config/manager/manager.yaml#L7) that we are trying to reuse here. And that's perhaps for some sort of filtering, I'm not sure TBH.

fmuyassarov:
@ardaguclu @andfasano 
I think I now have the reason why namespace base is not added in the config/default.
Both [TLS](https://github.com/metal3-io/baremetal-operator/blob/23493e6d94bbcff1c4da4623197674022dd5ccae/config/tls/kustomization.yaml#L5) and [Basic-auth](https://github.com/metal3-io/baremetal-operator/blob/23493e6d94bbcff1c4da4623197674022dd5ccae/config/basic-auth/default/kustomization.yaml#L5) are based on `config/default ` + `config/namespace`.  If we are adding namespace installation within config/default, that measn both TLS and Basic-auth will fail because we are trying to install the same namespace twice (via `config/default ` and `config/namespace`).

Example:
```
$ kustomize build config/tls | kubectl apply -f - 
Error: accumulating resources: accumulation err='accumulating resources from '../namespace': '/tmp/baremetal-operator/config/namespace' must resolve to a file': recursed merging from path '/tmp/baremetal-operator/config/namespace': may not add resource with an already registered id: ~G_v1_Namespace|~X|baremetal-operator-system
error: no objects passed to apply
```
I think we can either close this PR assuming people will use TLS or Basic-auth which will install the namespace as well for them, or remove the `../../namesapce` from [TLS](https://github.com/metal3-io/baremetal-operator/blob/23493e6d94bbcff1c4da4623197674022dd5ccae/config/tls/kustomization.yaml#L5) and [Basic-auth](https://github.com/metal3-io/baremetal-operator/blob/23493e6d94bbcff1c4da4623197674022dd5ccae/config/basic-auth/default/kustomization.yaml#L5) and instead add it only within `config/default `.  Any thoughts ?

lentzi90:
> I think we can either close this PR assuming people will use TLS or Basic-auth which will install the namespace as well for them or they will just need to install it as mentioned in the doc, or remove the `../../namesapce` from [TLS](https://github.com/metal3-io/baremetal-operator/blob/23493e6d94bbcff1c4da4623197674022dd5ccae/config/tls/kustomization.yaml#L5) and [Basic-auth](https://github.com/metal3-io/baremetal-operator/blob/23493e6d94bbcff1c4da4623197674022dd5ccae/config/basic-auth/default/kustomization.yaml#L5) and instead add it only within `config/default `. Any thoughts ?

How about having a "base" without the namespace (I guess this is `config/default`?) and a proper "default" overlay that adds the namespace just like TLS and Basic-auth does? Then anyone that want to do their own config still has something minimal to start from (the "base"), while we still have default, TLS and Basic-auth left.
I would also be fine with completely dropping support for the "no-auth".

ardaguclu:
@fmuyassarov as far as I can see, this https://github.com/metal3-io/baremetal-operator/blob/main/config/kustomization.yaml might be used and nowhere is using it(it also renders namespace and default). 

I think, it would be better to remove namespace from tls and basic_auth. Because namespace creation is one of the backbone of these yamls and might be better to create it initially, instead some custom place.

lentzi90:
> @fmuyassarov as far as I can see, this https://github.com/metal3-io/baremetal-operator/blob/main/config/kustomization.yaml might be used and nowhere is using it(it also renders namespace and default).

Nice did not see that! This would be the "default overlay" I was talking about and can be used instead of config/default.

> I think, it would be better to remove namespace from tls and basic_auth. Because namespace creation is one of the backbone of these yamls and might be better to create it initially, instead some custom place.

Not sure if this is a good idea. If we are sure no user will ever want to run things in a different namespace it is fine to do this, but otherwise it will makes things complicated for them. Leaving the base without a namespace makes it easier to change the namespace since one can simply add it. If there is already a namespace in the base, it would first need to be removed.

ardaguclu:
> > @fmuyassarov as far as I can see, this https://github.com/metal3-io/baremetal-operator/blob/main/config/kustomization.yaml might be used and nowhere is using it(it also renders namespace and default).
> 
> Nice did not see that! This would be the "default overlay" I was talking about and can be used instead of config/default.
> 
> > I think, it would be better to remove namespace from tls and basic_auth. Because namespace creation is one of the backbone of these yamls and might be better to create it initially, instead some custom place.
> 
> Not sure if this is a good idea. If we are sure no user will ever want to run things in a different namespace it is fine to do this, but otherwise it will makes things complicated for them. Leaving the base without a namespace makes it easier to change the namespace since one can simply add it. If there is already a namespace in the base, it would first need to be removed.

Thanks @lentzi90. But current kustomization structure already does not work without namespaces. Even here https://github.com/metal3-io/baremetal-operator/blob/d3bcdd79e5114eef4901d82a73983090aa7679a6/config/default/kustomization.yaml#L2 assumes that namespace is created. If user wants to change namespace, they can just edit here https://github.com/metal3-io/baremetal-operator/blob/main/config/namespace/namespace.yaml.

Current structure already makes changing namespaces difficult because user has to know tls and basic_auth yamls should also be changed if they want to use different namespace.

PR #976, opened 2021-09-14
Added PhysicalDisks and Controller fields to Hardware RAID

MahnoorAsghar:

PR #1062, opened 2022-01-04
Add physicalDisks and controller parameters to HardwareRAID

demonCoder95:
This patch adds the physical disks listing to the HardwareRAID
configuration as well as the ability to specify the RAID
controller to be used for configuration of RAID during
provisioning.

This work is based on the specification available here:
https://github.com/metal3-io/metal3-docs/pull/196

This work reimplements the work in #976


demonCoder95:
@fmuyassarov @zaneb You can close PR #976 since that is being abandoned. This PR reimplements the same work that was completed over there and needed to accommodate a lot of changes to the upstream BMO repo.

Would be nice if this can be quickly reviewed and merged. #976 only needed some housekeeping work, it was already approved. Thanks a bunch!

pkg/provisioner/ironic/raid.go
```diff
@@ -99,11 +99,26 @@ func buildTargetHardwareRAIDCfg(volumes []metal3v1alpha1.HardwareRAIDVolume) (lo
 			}
 			nameCheckFlags[volume.Name] = index
 		}
+
+		// Check numberOfPhysicalDisks is same as len(physicalDisks)
+		if volume.NumberOfPhysicalDisks != nil {
+			if *volume.NumberOfPhysicalDisks != len(volume.PhysicalDisks) {
```
zaneb, line 105: This effectively makes volume.PhysicalDisks a required property doesn't it? I think we should only compare if they are both non-zero.

demonCoder95: ah! good catch! i'll fix it right away.

pkg/provisioner/ironic/raid.go
```diff
@@ -99,11 +99,26 @@ func buildTargetHardwareRAIDCfg(volumes []metal3v1alpha1.HardwareRAIDVolume) (lo
 			}
 			nameCheckFlags[volume.Name] = index
 		}
+
+		// Check numberOfPhysicalDisks is same as len(physicalDisks)
+		if volume.NumberOfPhysicalDisks != nil && volume.PhysicalDisks != nil {
```
zaneb, line 104: I assume you've verified that PhysicalDisks is nil (and not just an empty slice) when it is not specified?

zaneb (commented):
/approve

pkg/provisioner/ironic/raid.go
demonCoder95: Oh, you're right. It's not a pointer, so it will be an empty slice. Let me fix that.

docs/api.md
```diff
@@ -173,6 +173,10 @@ The sub-fields are:
   * *numberOfPhysicalDisks* -- Integer, number of physical disks to use for the
     logical disk. Defaults to minimum number of disks required for the
     particular RAID level.
+  * *physicalDisks* -- List of names of physical disks (Strings). This is an
+    optional field. If specified, the `controller` field must be specified too.
```
andfasano, line 177: I was looking for a validation rule, either in the webhook or controller code, but didn't find it. @demonCoder95 shouldn't be required, as per documentation?

demonCoder95: Could you elaborate on the scope of this validation? What do we need to validate?

andfasano: That the `controller` field is set when `physicalDisks` is specified

demonCoder95: Oh right, I've added it here. let me send the patch.

demonCoder95: All done. @andfasano

pkg/provisioner/ironic/raid.go
zaneb: Slices can be nil (even if they're not pointers), so it's never clear to me what the deserialization of any particular field will give us without trying it. Checking for zero length is a safe way to cover everything though.

zaneb:
I added the validations we'll eventually want to add to the webhook to the list in #855.

demonCoder95:
> I added the validations we'll eventually want to add to the webhook to the list in #855.

Oh great. I've also added the requested validation in the controller, though. Thanks for letting me know :)

demonCoder95:
@zaneb @andfasano Anything else that needs doing in this PR? Let's merge it if there's nothing else. I can move forward with #855 and add necessary validations then.

pkg/provisioner/ironic/raid.go
```diff
@@ -99,11 +99,33 @@ func buildTargetHardwareRAIDCfg(volumes []metal3v1alpha1.HardwareRAIDVolume) (lo
...
+
+		// Check that controller field is specified if PhysicalDisks are used
+		if len(volume.PhysicalDisks) != 0 {
+			if volume.Controller == "" {
+				return nil, errors.Errorf("'controller' must be specified if 'physicalDisks' are used!")
+			}
+		}
+
+		// Check numberOfPhysicalDisks is same as len(physicalDisks)
+		if volume.NumberOfPhysicalDisks != nil && len(volume.PhysicalDisks) != 0 {
```
andfasano, line 111: Q: what if `PhysicalDisk` is defined but not `NumberOfPhysicalDisks`?

demonCoder95: I'd incline towards what @zaneb has said. Specifying ``NumberOfPhysicalDisks`` just doesn't need to be conflicting with the number of ``PhysicalDisks``.

andfasano: Ok thanks for the clarification

pkg/provisioner/ironic/raid_test.go
```diff
@@ -12,6 +12,7 @@ import (
 func TestBuildTargetRAIDCfg(t *testing.T) {
 	var TRUE = true
 	var FALSE = false
+	physicalDisks := make([]interface{}, 0)
```
andfasano, line 15: I think it would be nice to have at least one case where the hardware physical volumes were set (and possibly some others to verify the error conditions)

demonCoder95: Oh alright. I'll add one right away.

andfasano:
Thanks @demonCoder95 

/lgtm

PR #1063, opened 2022-01-05
Fix software RAID

dtantsur:
The current code treats software RAID similarly to hardware RAID and
expects the driver's RAIDInterface value to be "agent" to use it.
This is never the case, as software RAID is not driver-specific.

This change adjusts the procedure to set the expected RAIDInterface
before setting the target configuration. This makes software RAID work
with any values of the driver's RAIDInterface.

### DIFF
--- a/pkg/hardwareutils/bmc/access_test.go
+++ b/pkg/hardwareutils/bmc/access_test.go
@@ -434,7 +434,6 @@ func TestStaticDriverInfo(t *testing.T) {
 		boot       string
 		management string
 		power      string
-		raid       string
 		vendor     string
 	}{
 		{
@@ -446,7 +445,6 @@ func TestStaticDriverInfo(t *testing.T) {
 			boot:       "ipxe",
 			management: "",
 			power:      "",
-			raid:       "no-raid",
 		},
 
 		{
@@ -458,7 +456,6 @@ func TestStaticDriverInfo(t *testing.T) {
 			boot:       "ipxe",
 			management: "",
 			power:      "",
-			raid:       "no-raid",
 		},
 
 		{
@@ -470,7 +467,6 @@ func TestStaticDriverInfo(t *testing.T) {
 			boot:       "ipxe",
 			management: "",
 			power:      "",
-			raid:       "no-raid",
 		},
 
 		{
@@ -482,7 +478,6 @@ func TestStaticDriverInfo(t *testing.T) {
 			boot:       "ipxe",
 			management: "",
 			power:      "",
-			raid:       "irmc",
 		},
 
 		{
@@ -494,7 +489,6 @@ func TestStaticDriverInfo(t *testing.T) {
 			boot:       "ipxe",
 			management: "",
 			power:      "",
-			raid:       "no-raid",
 		},
 
 		{
@@ -506,7 +500,6 @@ func TestStaticDriverInfo(t *testing.T) {
 			boot:       "redfish-virtual-media",
 			management: "",
 			power:      "",
-			raid:       "no-raid",
 		},
 
 		{
@@ -518,7 +511,6 @@ func TestStaticDriverInfo(t *testing.T) {
 			boot:       "redfish-virtual-media",
 			management: "",
 			power:      "",
-			raid:       "no-raid",
 		},
 
 		{
@@ -530,7 +522,6 @@ func TestStaticDriverInfo(t *testing.T) {
 			boot:       "redfish-virtual-media",
 			management: "",
 			power:      "",
-			raid:       "no-raid",
 		},
 
 		{
@@ -542,7 +533,6 @@ func TestStaticDriverInfo(t *testing.T) {
 			boot:       "ipxe",
 			management: "idrac-redfish",
 			power:      "idrac-redfish",
-			raid:       "no-raid",
 			vendor:     "idrac-redfish",
 		},
 
@@ -582,7 +572,6 @@ func TestStaticDriverInfo(t *testing.T) {
 			boot:       "idrac-redfish-virtual-media",
 			management: "idrac-redfish",
 			power:      "idrac-redfish",
-			raid:       "no-raid",
 			vendor:     "idrac-redfish",
 		},
 
@@ -595,7 +584,6 @@ func TestStaticDriverInfo(t *testing.T) {
 			boot:       "idrac-redfish-virtual-media",
 			management: "idrac-redfish",
 			power:      "idrac-redfish",
-			raid:       "no-raid",
 			vendor:     "idrac-redfish",
 		},
 
@@ -608,7 +596,6 @@ func TestStaticDriverInfo(t *testing.T) {
 			boot:       "idrac-redfish-virtual-media",
 			management: "idrac-redfish",
 			power:      "idrac-redfish",
-			raid:       "no-raid",
 			vendor:     "idrac-redfish",
 		},
 
@@ -621,7 +608,6 @@ func TestStaticDriverInfo(t *testing.T) {
 			boot:       "pxe",
 			management: "ibmc",
 			power:      "ibmc",
-			raid:       "no-raid",
 		},
 
 		{
@@ -633,7 +619,6 @@ func TestStaticDriverInfo(t *testing.T) {
 			boot:       "ilo-ipxe",
 			management: "",
 			power:      "",
-			raid:       "no-raid",
 		},
 
 		{
@@ -645,7 +630,6 @@ func TestStaticDriverInfo(t *testing.T) {
 			boot:       "ilo-virtual-media",
 			management: "",
 			power:      "",
-			raid:       "no-raid",
 		},
 
 		{
@@ -657,7 +641,6 @@ func TestStaticDriverInfo(t *testing.T) {
 			boot:       "ilo-ipxe",
 			management: "",
 			power:      "",
-			raid:       "ilo5",
 		},
 	} {
 		t.Run(tc.Scenario, func(t *testing.T) {
--- a/pkg/provisioner/ironic/raid.go
+++ b/pkg/provisioner/ironic/raid.go
@@ -19,7 +19,7 @@ const (
 
 // setTargetRAIDCfg set the RAID settings to the ironic Node for RAID configuration steps
 func setTargetRAIDCfg(p *ironicProvisioner, raidInterface string, ironicNode *nodes.Node, data provisioner.PrepareData) (provisioner.Result, error) {
-	err := checkRAIDConfigure(raidInterface, data.TargetRAIDConfig)
+	err := CheckRAIDInterface(raidInterface, data.TargetRAIDConfig)
 	if err != nil {
 		return operationFailed(err.Error())
 	}
@@ -178,7 +178,7 @@ func buildTargetSoftwareRAIDCfg(volumes []metal3v1alpha1.SoftwareRAIDVolume) (lo
 
 // BuildRAIDCleanSteps build the clean steps for RAID configuration from BaremetalHost spec
 func BuildRAIDCleanSteps(raidInterface string, target *metal3v1alpha1.RAIDConfig, actual *metal3v1alpha1.RAIDConfig) (cleanSteps []nodes.CleanStep, err error) {
-	err = checkRAIDConfigure(raidInterface, target)
+	err = CheckRAIDInterface(raidInterface, target)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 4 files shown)


dtantsur:
/test-v1b1-centos-integration 
/test-v1b1-integration

I must admit I haven't tested this change, but it looks reasonable, or at least more reasonable than the code we have :)

bfournie (approved):

Hellcatlk:
lgtm, but I want to do some testing before merging,
But I'm very sorry, 1.31~2.6 is the Chinese New Year holiday, I can't start this work right now.
I will start this work immediately after the holidays.

Hellcatlk:
After testing, this patch works fine.
But Ironic doesn't seem to have successfully configured the software raid,
In the preparing state, we can see the following raid configuration in bmh:
```
[core@worker-2 ~]$ sudo lsblk
NAME MAJ:MIN RM SIZE RO TYPE MOUNTPOINT
loop0 7:0 0 15.4G 0 loop /run/ephemeral
loop1 7:1 0 882.7M 0 loop /sysroot
sda 8:0 0 372.6G 0 disk
└─sda1 8:1 0 372.1G 0 part
  └─md0 9:0 0 372G 0 raid1
sdb 8:16 0 372.6G 0 disk
└─sdb1 8:17 0 372.1G 0 part
  └─md0 9:0 0 372G 0 raid1
sdc 8:32 0 1.8T 0 disk
└─sdc1 8:33 0 1.8T 0 part
  └─md0 9:0 0 372G 0 raid1
sdd 8:48 0 1.8T 0 disk
└─sdd1 8:49 0 1.8T 0 part
  └─md0 9:0 0 372G 0 raid1
sde 8:64 0 1.8T 0 disk
└─sde1 8:65 0 1.8T 0 part
  └─md0 9:0 0 372G 0 raid1
```

But in provisioned state, the software raid is gone:
```
[root@worker-2 core]# lsblk
NAME MAJ:MIN RM SIZE RO TYPE MOUNTPOINT
sda 8:0 0 372.6G 0 disk
├─sda1 8:1 0 1M 0 part
├─sda2 8:2 0 127M 0 part
├─sda3 8:3 0 384M 0 part /boot
└─sda4 8:4 0 372.1G 0 part /sysroot
sdb 8:16 0 372.6G 0 disk
sdc 8:32 0 1.8T 0 disk
sdd 8:48 0 1.8T 0 disk
sde 8:64 0 1.8T 0 disk
```

Following is the raid configuration:
```
spec:
  raid:
    softwareRAIDVolumes:
    - level: "1"
```

Is this reasonable?

zhouhao3:
Hi @zaneb @dtantsur, as @Hellcatlk replied above, can we merge this PR if that's normal?

dtantsur:
@Hellcatlk you may need to check if `/dev/md0` is set as a root device hint. I think Ironic sets it, but maybe BMO overwrites it then.

In any case, I think it's a great improvement, and we can iterate further.

dtantsur:
Otherwise, I'll need ironic and BMO logs to be sure.

pkg/provisioner/ironic/raid.go
```diff
@@ -251,20 +254,29 @@ func BuildRAIDCleanSteps(raidInterface string, target *metal3v1alpha1.RAIDConfig
...
-		}
-	case softwareRAIDInterface:
-		if raid != nil && len(raid.HardwareRAIDVolumes) != 0 {
-			return fmt.Errorf("node's driver %s does not support hardware RAID", raidInterface)
-		}
-	default:
-		if raid != nil && len(raid.HardwareRAIDVolumes) == 0 && len(raid.SoftwareRAIDVolumes) != 0 {
-			return fmt.Errorf("node's driver %s does not support software RAID", raidInterface)
-		}
+func checkRAIDInterface(raidInterface string, target, actual *metal3v1alpha1.RAIDConfig) (string, error) {
```
hs0210, line 257: > The logic here is wrong, see https://github.com/metal3-io/baremetal-operator/pull/1063. I wonder if we can avoid duplication of BMO code here. We can make that function public.

Hi, could you please help make this function public as you said above in https://github.com/openshift-metal3/terraform-provider-ironic/pull/53#discussion_r794712182

Hellcatlk:
> @Hellcatlk you may need to check if /dev/md0 is set as a root device hint. I think Ironic sets it, but maybe BMO overwrites it then.

@dtantsur We checked that the root device hint is `/dev/sda`.
```
Raid:
  Hardware RAID Volumes: <nil> 
  Software RAID Volumes:
    Level: 1 
Root Device Hints
  Device Name: /dev/sda
```

> In any case, I think it's a great improvement, and we can iterate further.

Agree.

pkg/hardwareutils/bmc/access_test.go
```diff
@@ -679,6 +682,13 @@ func TestStaticDriverInfo(t *testing.T) {
 				t.Fatalf("Unexpected bios interface %q, expected %q",
 					acc.BIOSInterface(), tc.bios)
 			}
+			if acc.RAIDInterface() != tc.raid {
```
zaneb, line 685: Does this really achieve anything other than making you change the code in two places whenever you change a driver?

pkg/provisioner/ironic/raid.go
```diff
@@ -251,20 +254,29 @@ func BuildRAIDCleanSteps(raidInterface string, target *metal3v1alpha1.RAIDConfig
...
 	}
-	return nil
+
+	// If software RAID is requested, change the RAID interface.
+	if len(target.SoftwareRAIDVolumes) != 0 {
+		return softwareRAIDInterface, nil
+	}
+
+	// If software RAID is being deleted, also change the interface.
+	if actual != nil && len(target.SoftwareRAIDVolumes) == 0 && len(actual.SoftwareRAIDVolumes) != 0 {
```
zaneb, line 277: Second condition is always true here. Should it be HardwareRAIDVolumes?

pkg/provisioner/ironic/raid.go
```diff
@@ -156,28 +163,24 @@ func buildTargetSoftwareRAIDCfg(volumes []metal3v1alpha1.SoftwareRAIDVolume) (lo
...
-			target.HardwareRAIDVolumes = nil
-		}
+		target.HardwareRAIDVolumes = nil
 		if actual != nil {
 			actual.HardwareRAIDVolumes = nil
 		}
 		if reflect.DeepEqual(target, actual) {
 			return
 		}
+		if len(target.SoftwareRAIDVolumes) == 0 && (actual == nil || actual.SoftwareRAIDVolumes == nil || len(actual.SoftwareRAIDVolumes) == 0) {
```
zaneb, line 181: `len(<slice>) == 0` is always true when `<slice> == nil` is true, so there's no need to check both.

pkg/provisioner/ironic/raid.go
dtantsur: Mmm, no, how are hardware volumes related? This handles the case when we have some volumes created, and the user wants to delete them, thus setting target to [].

pkg/hardwareutils/bmc/access_test.go
dtantsur: I don't know, it's an established pattern here. Better visibility at least?

dtantsur:
@Hellcatlk can you try removing `RootDeviceHints` from your BMH?

pkg/hardwareutils/bmc/access_test.go
zaneb: An established pattern of busywork ;)

(Checking stuff that is inferred from the URL, like the driver name and the boot interface, makes sense and is the purpose of this test. The rest is pure cargo-cult.)

pkg/provisioner/ironic/raid.go
zaneb: As it stands if you're changing from Software RAID (in actual) to Hardware RAID (in target) then it'll return softwareRAIDInterface, which is surprising. (I'd find it easier to believe that there is _no_ right answer here than that this is the right answer.)

dtantsur: > As it stands if you're changing from Software RAID (in actual) to Hardware RAID (in target) then it'll return softwareRAIDInterface, which is surprising.

There is no mention of hardware RAID here, I don't quite understand what you're referring to. This logic triggers when we have an existing software RAID (in actual), and the user requests no software RAID volumes (in target). To remove software RAID, we need to use the software RAID interface, otherwise it will stuck forever.

Hellcatlk:
@dtantsur Hi, did the [above test results](https://github.com/metal3-io/baremetal-operator/pull/1063#issuecomment-1044422540) meet your expectations? Is there anything else we need to do?

pkg/provisioner/ironic/raid.go
zaneb: Sorry, reading my comment again I think it was backwards. The issue is when you change from hardware RAID to software RAID.

Two configs are passed here, `target` which is the RAID setting we want to configure, and `actual`, which is the RAID setting that we already configured the host with.

If the host _was_ configured with hardware RAID and we're removing it, we use the hardware RAID driver to do so, as you'd expect. But if we're simultaneously replacing the hardware RAID config with a software RAID config, then we'll try to use the software RAID driver to remove the hardware RAID config, which is surprising.

zaneb: If this is intentional then you can delete the `len(target.SoftwareRAIDVolumes) == 0` part because it's already checked on line 273.

pkg/provisioner/ironic/raid.go
```diff
@@ -43,6 +43,13 @@ func setTargetRAIDCfg(p *ironicProvisioner, raidInterface string, ironicNode *no
 		p.log.Info("rootDeviceHints is used, the first volume of raid will not be set to root")
 	}
 
+	updater := updateOptsBuilder(p.debugLog)
+	updater.SetTopLevelOpt("raid_interface", targetRaidInterface, ironicNode.RAIDInterface)
```
zhouhao3, line 47: > This is never the case, as software RAID is not driver-specific.

In this case, why do we need to update the `raid_interface` here? If we want to configure software RAID, this will change the node's `raid_interface` to `agent`. If the node had hardware RAID before this would cause the problem that @zaneb  pointed out below.

> If the host was configured with hardware RAID and we're removing it, we use the hardware RAID driver to do so, as you'd expect. But if we're simultaneously replacing the hardware RAID config with a software RAID config, then we'll try to use the software RAID driver to remove the hardware RAID config, which is surprising

dtantsur: > In this case, why do we need to update the raid_interface here? If we want to configure software RAID, this will change the node's raid_interface to agent.

Yes. Otherwise you will not be able to build software RAID.

pkg/provisioner/ironic/raid.go
dtantsur: I'll leave a FIXME and update the docs.

pkg/provisioner/ironic/raid.go
```diff
@@ -273,20 +276,32 @@ func BuildRAIDCleanSteps(raidInterface string, target *metal3v1alpha1.RAIDConfig
...
-	case softwareRAIDInterface:
-		if raid != nil && len(raid.HardwareRAIDVolumes) != 0 {
-			return fmt.Errorf("node's driver %s does not support hardware RAID", raidInterface)
-		}
-	default:
-		if raid != nil && len(raid.HardwareRAIDVolumes) == 0 && len(raid.SoftwareRAIDVolumes) != 0 {
-			return fmt.Errorf("node's driver %s does not support software RAID", raidInterface)
-		}
+// CheckRAIDInterface checks the current RAID interface against the requested configuration
+func CheckRAIDInterface(raidInterface string, target, actual *metal3v1alpha1.RAIDConfig) (string, error) {
```
andfasano, line 280: Will this change impact the webhook validation as well somehow?

https://github.com/metal3-io/baremetal-operator/blob/360fad81fe1c50a36eb1a64bd86b70a4fc63bf3a/apis/metal3.io/v1alpha1/baremetalhost_validation.go#L102-L130

dtantsur: The logic there seems still correct to me.

hs0210:
It seems that the code is OK. But the soft raid does not seem to be successfully configured judging from @Hellcatlk's [test result](https://github.com/metal3-io/baremetal-operator/pull/1063#issuecomment-1044422540)...

elfosardo (approved):

elfosardo:
@hardys @fmuyassarov @andfasano @zaneb can you please have a look when you get the chance? thanks!

PR #1064, opened 2022-01-06
Use integer value in clean steps for HostFirmwareSettings Integer type

bfournie:
For these types the value in clean steps should be stored as an integer, otherwise vendors reject the Redfish request.


dtantsur:
/approve

Let me guess: the next issue will be with `bool` :)

bfournie:
> /approve
> 
> Let me guess: the next issue will be with `bool` :)

Good point! Will check that, although most vendors don't use the Boolean type (preferring an Enumeration type with "Enabled","Disabled" instead)

andfasano:
@bfournie I think that as a quick/temporary fix could make sense, but I'd suggest to review the entire approach later, maybe adopting an `interface{}` for managing the settings value.

/lgtm

PR #1066, opened 2022-01-11
Make the inspector use the same image as Ironic

lentzi90:


tools/run_local_ironic.sh
```diff
@@ -5,7 +5,7 @@ set -ex
 SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
 
 IRONIC_IMAGE=${IRONIC_IMAGE:-"quay.io/metal3-io/ironic:master"}
-IRONIC_INSPECTOR_IMAGE=${IRONIC_INSPECTOR_IMAGE:-"quay.io/metal3-io/ironic"}
+IRONIC_INSPECTOR_IMAGE=${IRONIC_IMAGE}
```
lentzi90, line 8: Not setting a default value here for `IRONIC_INSPECTOR_IMAGE` since `IRONIC_IMAGE` is guaranteed to have a value from the line above.

fmuyassarov (approved):

PR #1068, opened 2022-01-13
Check if bmc hostname follows DNS Standard

hroyrh:
Adds a function to check if the bmc.address follows DNS standard and exports the GetParsedURL function to make it available in other validations.

### DIFF
--- a/pkg/hardwareutils/bmc/access.go
+++ b/pkg/hardwareutils/bmc/access.go
@@ -173,7 +173,7 @@ func checkDNSValid(address string) error {
 	// Check if BMC address hostname follows DNS Standard
 	valid, _ := regexp.MatchString(`^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]{0,61}[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]{0,61}[A-Za-z0-9])$`, address)
 	if !valid {
-		return fmt.Errorf("BMC address hostname [%s] invalid, expected format : Type://DNS_Standard_Hostname:Port", address)
+		return fmt.Errorf("BMC address hostname/IP : [%s] is invalid", address)
 	}
 	return nil
 }
--- a/pkg/hardwareutils/bmc/access.go
+++ b/pkg/hardwareutils/bmc/access.go
@@ -88,10 +88,6 @@ type AccessDetails interface {
 }
 
 func GetParsedURL(address string) (parsedURL *url.URL, err error) {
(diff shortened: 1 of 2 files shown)


pkg/hardwareutils/bmc/access_test.go
```diff
@@ -368,7 +368,7 @@ func TestParse(t *testing.T) {
 		},
 	} {
 		t.Run(tc.Scenario, func(t *testing.T) {
-			url, err := getParsedURL(tc.Address)
+			url, err := GetParsedURL(tc.Address)
```
ardaguclu, line 371: I think, tests including invalid dns cases might be added?

pkg/hardwareutils/bmc/access.go
```diff
@@ -150,3 +157,23 @@ func NewAccessDetails(address string, disableCertificateVerification bool) (Acce
...
+
+func checkDNSValid(address string) error {
+
+	// Allowing empty BMC address
+	if address == "" {
+		return nil
+	}
+
+	// Check if its a IPv6/IPv4 address
+	if net.ParseIP(address) != nil {
```
ardaguclu, line 169: In this line, we have a DNS value, why we are parsing assuming that it is IP?

pkg/hardwareutils/bmc/access_test.go
ardaguclu: We really need lots of tests to assure that is working.

pkg/hardwareutils/bmc/access.go
hroyrh: In lot of tests we have IPv6 address ( and I think the BMC address hostname can be an IPv6 address) - this check is validate that. IPv4 addresses are anyways DNS standard compliant, so it doesn't need a check - I will remove IPv4 from the comment to avoid this confusion.

ardaguclu:
This PR looks good to me, but I'll not label to someone else have a look at;

/test-v1b1-centos-integration
/test-v1b1-integration

zhouhao3:
/test-v1b1-centos-integration

sems good to me, but like @ardaguclu said, we need to add corresponding test cases for this.

hroyrh:
Would it make sense to add test cases here or at the [baremetalhost_valitations level](https://github.com/metal3-io/baremetal-operator/pull/1040)
cc @ardaguclu @zhouhao3

zhouhao3:
>Would it make sense to add test cases here or at the https://github.com/metal3-io/baremetal-operator/pull/1040

I think both places need their own corresponding tests.

zhouhao3 (commented):
LGTM

PR #1070, opened 2022-01-14
Add physicalDisk and controller WebHook validation for HardwareRAID

demonCoder95:
This patch adds WebHook validation for Hardware RAID to validate the following:

1. ``physicalDisks`` specified without ``controller`` not allowed.
2. ``numberOfPhysicalDisks`` and length of ``physicalDisks`` must be consistent. 

This is in accordance with the WebHook validation outline provided in #855 . The functionality that enables these parameters in ``HardwareRAID`` is available in #1062 

The patch also provides unit test coverage for the validations added into the WebHook.


demonCoder95:
> /lgtm

Let's wait for someone to approve this as well, after reviewing? Perhaps @andfasano

PR #1074, opened 2022-01-20
Add make targets for setting images

lentzi90:
These make targets can be used to change the image used for BMO, Ironic,
Mariadb, keepalived and IPA downloader. This is similar to how things work in CAPM3.

The deploy target was removed. It has been broken for at least a year
since the name of the image was changed. I.e. it could apply the
manifests but the image substitution didn't work.

The idea is to unify how image replacement works in CAPM3 and BMO and solve https://github.com/metal3-io/metal3-dev-env/issues/908 by making use of these make targets in metal3-dev-env.

For reference, [this is how the image replacement works in CAPM3](https://github.com/metal3-io/cluster-api-provider-metal3/blob/2d686bf00a7d4babd8287dffc9f8bdabe07c26f2/Makefile#L370-L378).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,9 @@ help:  ## Display this help
 	@echo "  DEBUG            -- debug flag, if any ($(DEBUG))"
 
 # Image URL to use all building/pushing image targets
-IMG ?= baremetal-operator:latest
+IMG_NAME ?= baremetal-operator
+IMG_TAG ?= latest
+IMG ?= $(IMG_NAME):$(IMG_TAG)
 
 ## --------------------------------------
 ## Test Targets
@@ -129,8 +131,8 @@ uninstall: $(KUSTOMIZE) manifests ## Uninstall CRDs from a cluster
 	$< build config/crd | kubectl delete -f -
 
 .PHONY: deploy
-deploy: $(KUSTOMIZE) manifests ## Deploy controller in the configured Kubernetes cluster in ~/.kube/config
-	cd config/manager && kustomize edit set image controller=${IMG}
+deploy: $(KUSTOMIZE) manifests  ## Deploy controller in the configured Kubernetes cluster in ~/.kube/config
+	make set-manifest-image-bmo MANIFEST_IMG=$(IMG_NAME) MANIFEST_TAG=$(IMG_TAG)
 	$< build config/default | kubectl apply -f -
 
 $(CONTROLLER_GEN): hack/tools/go.mod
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -128,11 +129,6 @@ install: $(KUSTOMIZE) manifests ## Install CRDs into a cluster
 uninstall: $(KUSTOMIZE) manifests ## Uninstall CRDs from a cluster
 	$< build config/crd | kubectl delete -f -
 
-.PHONY: deploy
-deploy: $(KUSTOMIZE) manifests ## Deploy controller in the configured Kubernetes cluster in ~/.kube/config
-	cd config/manager && kustomize edit set image controller=${IMG}
```
lentzi90, line 133: This has not been working since the image name was changed from `controller` to `quay.io/metal3-io/baremetal-operator`. See [this commit](https://github.com/metal3-io/baremetal-operator/commit/209dbd095d8314c5e2fa106c69020fc06c34271e#diff-ce0d597a9e5f20a8fe61c3ba5e6185a6b90cfef2c00bb5b5ef3e97b15b79f480)

Makefile
```diff
@@ -151,6 +147,31 @@ manifests-generate: $(CONTROLLER_GEN)
...
 
+.PHONY: set-manifest-image-bmo
+set-manifest-image-bmo: manifests $(KUSTOMIZE)
+	$(info Updating container image for BMO to use ${MANIFEST_IMG}:${MANIFEST_TAG})
+	cd config/default && $(KUSTOMIZE) edit set image quay.io/metal3-io/baremetal-operator=${MANIFEST_IMG}:${MANIFEST_TAG}
+
+.PHONY: set-manifest-image-ironic
+set-manifest-image-ironic: manifests $(KUSTOMIZE)
+	$(info Updating container image for Ironic to use ${MANIFEST_IMG}:${MANIFEST_TAG})
+	cd ironic-deployment/default && $(KUSTOMIZE) edit set image quay.io/metal3-io/ironic=${MANIFEST_IMG}:${MANIFEST_TAG}
```
lentzi90, line 158: In CAPM3 we do a global substitution of all `image: *` in the manifest, but this does not work well with Ironic, since we have multiple different containers (e.g. mariadb). For this reason I decided to use `kustomize edit set image` instead so we replace only the mentioned images.

Makefile
```diff
@@ -151,6 +146,31 @@ manifests-generate: $(CONTROLLER_GEN)
...
 
+.PHONY: set-manifest-image-bmo
+set-manifest-image-bmo: $(KUSTOMIZE) manifests
+	$(info Updating container image for BMO to use ${MANIFEST_IMG}:${MANIFEST_TAG})
+	cd config/default && $(abspath $(KUSTOMIZE)) edit set image quay.io/metal3-io/baremetal-operator=${MANIFEST_IMG}:${MANIFEST_TAG}
+
+.PHONY: set-manifest-image-ironic
+set-manifest-image-ironic: $(KUSTOMIZE) manifests
+	$(info Updating container image for Ironic to use ${MANIFEST_IMG}:${MANIFEST_TAG})
+	cd ironic-deployment/default && $(abspath $(KUSTOMIZE)) edit set image quay.io/metal3-io/ironic=${MANIFEST_IMG}:${MANIFEST_TAG}
```
lentzi90, line 157: In CAPM3 we do a global substitution of all `image: *` in the manifest, but this does not work well with Ironic, since we have multiple different containers (e.g. mariadb). For this reason I decided to use `kustomize edit set image` instead so we replace only the mentioned images.

Please also note that I'm reusing the same variables in all targets (`MANIFEST_IMG` and `MANIFEST_TAG`). My thinking is that you cannot run all the targets at the same time so they may as well use the same. If you think it would be better with separate variables (e.g. `MANIFEST_IMG_BMO` and `MANIFEST_IMG_IRONIC`), please let me know.

dtantsur: You can run several targets, but I think it's an acceptable simplification.

Makefile
dtantsur: The kustomize part was not working properly, but I guess everything else was?

lentzi90: True. I was assuming that since the kustomize part was broken there was no point in the target as a whole. Should I rather keep the `kubectl apply` part and remove the rest?

zhouhao3 (approved):

Makefile
dtantsur: Can we maybe leave it in place, but make it depend on the new `set-manifest-image-bmo` target instead of running kustomize?

lentzi90: Sounds like a good solution. I'll update!

lentzi90: Done. Please take a look @dtantsur

Makefile
```diff
@@ -129,8 +131,8 @@ uninstall: $(KUSTOMIZE) manifests ## Uninstall CRDs from a cluster
 	$< build config/crd | kubectl delete -f -
 
 .PHONY: deploy
-deploy: $(KUSTOMIZE) manifests ## Deploy controller in the configured Kubernetes cluster in ~/.kube/config
-	cd config/manager && kustomize edit set image controller=${IMG}
+deploy: $(KUSTOMIZE) manifests  ## Deploy controller in the configured Kubernetes cluster in ~/.kube/config
+	make set-manifest-image-bmo MANIFEST_IMG=$(IMG_NAME) MANIFEST_TAG=$(IMG_TAG)
 	$< build config/default | kubectl apply -f -
```
fmuyassarov, line 136: `manifests` calls 
1. `manifests-generate`  (doesn't require kustomize)
2. `manifests-kustomize` (requires kustomize and it already [installs](https://github.com/metal3-io/baremetal-operator/blob/d3bcdd79e5114eef4901d82a73983090aa7679a6/Makefile#L151) it by itself )
I believe we can drop the `$(KUSTOMIZE)` here ?

Makefile
```diff
@@ -151,6 +153,31 @@ manifests-generate: $(CONTROLLER_GEN)
 manifests-kustomize: $(KUSTOMIZE)
 	$< build config/default > config/render/capm3.yaml
 
+.PHONY: set-manifest-image-bmo
+set-manifest-image-bmo: $(KUSTOMIZE) manifests
```
fmuyassarov, line 157: same here and other occurrences

Makefile
lentzi90: Actually `$<` points to kustomize here so we need it. I think `$<` is used instead of `$(KUSTOMIZE)` to make sure the path is correct, not sure if it is really needed though.

fmuyassarov: true, I missed that. Thanks

Makefile
```diff
@@ -129,8 +131,8 @@ uninstall: $(KUSTOMIZE) manifests ## Uninstall CRDs from a cluster
 	$< build config/crd | kubectl delete -f -
 
 .PHONY: deploy
-deploy: $(KUSTOMIZE) manifests ## Deploy controller in the configured Kubernetes cluster in ~/.kube/config
-	cd config/manager && kustomize edit set image controller=${IMG}
+deploy: $(KUSTOMIZE) manifests  ## Deploy controller in the configured Kubernetes cluster in ~/.kube/config
+	make set-manifest-image-bmo MANIFEST_IMG=$(IMG_NAME) MANIFEST_TAG=$(IMG_TAG)
```
dtantsur, line 135: I'm not sure I like the recursive make, but I understand why it's done.

PR #1075, opened 2022-01-20
Move from Available to Preparing if HostFirmwareSettings changed

bfournie:
Currently if the host is in the Available state and the HFS settings are changed it doesn't transition to the Preparing state as it would when the BareMetalHost settings are changed. This change fixes that.

### DIFF
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -951,7 +951,7 @@ func (r *BareMetalHostReconciler) actionMatchProfile(prov provisioner.Provisione
 func (r *BareMetalHostReconciler) actionPreparing(prov provisioner.Provisioner, info *reconcileInfo) actionResult {
 	info.log.Info("preparing")
 
-	dirty, newStatus, err := getHostProvisioningSettings(info.host)
+	bmhDirty, newStatus, err := getHostProvisioningSettings(info.host)
 	if err != nil {
 		return actionError{err}
 	}
@@ -962,49 +962,56 @@ func (r *BareMetalHostReconciler) actionPreparing(prov provisioner.Provisioner, 
 		RootDeviceHints:  newStatus.Provisioning.RootDeviceHints.DeepCopy(),
 		FirmwareConfig:   newStatus.Provisioning.Firmware.DeepCopy(),
 	}
-	// When manual cleaning fails, we think that the existed RAID configuration
+	// When manual cleaning fails, we think that the existing RAID configuration
 	// is invalid and needs to be reconfigured.
 	if info.host.Status.ErrorType == metal3v1alpha1.PreparationError {
 		prepareData.ActualRAIDConfig = nil
-		dirty = true
+		bmhDirty = true
 	}
 
-	// Use settings in hostFirmwareSettings if available
-	hfs, err := r.getHostFirmwareSettings(info)
+	// The hfsDirty flag is used to push the new settings to Ironic as part of the clean steps.
+	// The HFS Status field will be updated in the HostFirmwareSettingsReconciler when it reads the settings from Ironic.
+	// After manual cleaning is complete the HFS Spec should match the Status.
+	hfsDirty, hfs, err := r.getHostFirmwareSettings(info)
+
 	if err != nil {
 		// wait until hostFirmwareSettings are ready
 		return actionContinue{subResourceNotReadyRetryDelay}
 	}
-	if hfs != nil {
+	if hfsDirty {
 		prepareData.ActualFirmwareSettings = hfs.Status.Settings.DeepCopy()
 		prepareData.TargetFirmwareSettings = hfs.Spec.Settings.DeepCopy()
 	}
 
-	provResult, started, err := prov.Prepare(prepareData, dirty)
+	provResult, started, err := prov.Prepare(prepareData, bmhDirty || hfsDirty,
+		info.host.Status.ErrorType == metal3v1alpha1.PreparationError)
 
 	if err != nil {
 		return actionError{errors.Wrap(err, "error preparing host")}
 	}
 
 	if provResult.ErrorMessage != "" {
-		info.log.Info("handling cleaning error in controller")
-		clearHostProvisioningSettings(info.host)
+		if bmhDirty {
+			info.log.Info("handling cleaning error in controller")
+			clearHostProvisioningSettings(info.host)
+		}
 		return recordActionFailure(info, metal3v1alpha1.PreparationError, provResult.ErrorMessage)
 	}
 
-	if dirty && started {
+	if bmhDirty && started {
 		info.log.Info("saving host provisioning settings")
 		_, err := saveHostProvisioningSettings(info.host)
 		if err != nil {
 			return actionError{errors.Wrap(err, "could not save the host provisioning settings")}
 		}
 	}
+
 	if started && clearError(info.host) {
-		dirty = true
+		bmhDirty = true
 	}
 	if provResult.Dirty {
 		result := actionContinue{provResult.RequeueAfter}
-		if dirty {
+		if bmhDirty {
 			return actionUpdate{result}
 		}
 		return result
@@ -1384,34 +1391,35 @@ func (r *BareMetalHostReconciler) createHostFirmwareSettings(info *reconcileInfo
 }
 
 // Get the stored firmware settings if there are valid changes
-func (r *BareMetalHostReconciler) getHostFirmwareSettings(info *reconcileInfo) (hfs *metal3v1alpha1.HostFirmwareSettings, err error) {
+func (r *BareMetalHostReconciler) getHostFirmwareSettings(info *reconcileInfo) (dirty bool, hfs *metal3v1alpha1.HostFirmwareSettings, err error) {
 
 	hfs = &metal3v1alpha1.HostFirmwareSettings{}
 	if err = r.Get(context.TODO(), info.request.NamespacedName, hfs); err != nil {
 
 		if !k8serrors.IsNotFound(err) {
 			// Error reading the object
-			return nil, errors.Wrap(err, "could not load host firmware settings")
+			return false, nil, errors.Wrap(err, "could not load host firmware settings")
 		}
 
 		// Could not get settings, log it but don't return error as settings may not have been available at provisioner
 		info.log.Info("could not get hostFirmwareSettings", "namespacename", info.request.NamespacedName)
-		return nil, nil
+		return false, nil, nil
 	}
 
 	// Check if there are settings in the Spec that are different than the Status
 	if meta.IsStatusConditionTrue(hfs.Status.Conditions, string(metal3v1alpha1.FirmwareSettingsChangeDetected)) {
 
 		if meta.IsStatusConditionTrue(hfs.Status.Conditions, string(metal3v1alpha1.FirmwareSettingsValid)) {
-			return hfs, nil
+			info.log.Info("hostFirmwareSettings indicating ChangeDetected", "namespacename", info.request.NamespacedName)
+			return true, hfs, nil
 		}
 
 		info.log.Info("hostFirmwareSettings not valid", "namespacename", info.request.NamespacedName)
-		return nil, nil
+		return false, nil, nil
 	}
 
 	info.log.Info("hostFirmwareSettings no updates", "namespacename", info.request.NamespacedName)
-	return nil, nil
+	return false, nil, nil
 }
 
 func (r *BareMetalHostReconciler) saveHostStatus(host *metal3v1alpha1.BareMetalHost) error {
--- a/controllers/metal3.io/baremetalhost_controller_test.go
+++ b/controllers/metal3.io/baremetalhost_controller_test.go
@@ -17,6 +17,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
 
 	promutil "github.com/prometheus/client_golang/prometheus/testutil"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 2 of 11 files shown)


controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -966,45 +966,48 @@ func (r *BareMetalHostReconciler) actionPreparing(prov provisioner.Provisioner,
 	// is invalid and needs to be reconfigured.
 	if info.host.Status.ErrorType == metal3v1alpha1.PreparationError {
 		prepareData.ActualRAIDConfig = nil
-		dirty = true
+		bmhDirty = true
 	}
 
 	// Use settings in hostFirmwareSettings if available
-	hfs, err := r.getHostFirmwareSettings(info)
+	hfsDirty, hfs, err := r.getHostFirmwareSettings(info)
```
dtantsur, line 973: I guess you could just use the same `dirty` variable

zhouhao3 (approved):

controllers/metal3.io/host_state_machine.go
```diff
@@ -417,6 +417,13 @@ func (hsm *hostStateMachine) handleAvailable(info *reconcileInfo) actionResult {
 	} else if dirty {
 		hsm.NextState = metal3v1alpha1.StatePreparing
 		return actionComplete{}
+	} else {
+		// Check if hostFirmwareSettings have been changed
+		dirty, _, _ := hsm.Reconciler.getHostFirmwareSettings(info)
+		if dirty {
+			hsm.NextState = metal3v1alpha1.StatePreparing
```
andfasano, line 424: @bfournie it would be nice to have some coverage on this part (considering that a state change is a relevant action). As a good starting point for that you could have a look at the [baremetalhost controller test](https://github.com/metal3-io/baremetal-operator/blob/ea7afff67eefb9360cc278dc4f34e340461551f4/controllers/metal3.io/baremetalhost_controller_test.go)

controllers/metal3.io/host_state_machine.go
```diff
@@ -417,6 +417,13 @@ func (hsm *hostStateMachine) handleAvailable(info *reconcileInfo) actionResult {
 	} else if dirty {
 		hsm.NextState = metal3v1alpha1.StatePreparing
 		return actionComplete{}
+	} else {
+		// Check if hostFirmwareSettings have been changed
+		dirty, _, _ := hsm.Reconciler.getHostFirmwareSettings(info)
```
zaneb, line 422: I don't think we should ignore `err` here.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -966,45 +966,48 @@ func (r *BareMetalHostReconciler) actionPreparing(prov provisioner.Provisioner,
...
 		return actionContinue{subResourceNotReadyRetryDelay}
 	}
-	if hfs != nil {
+	if hfsDirty {
 		prepareData.ActualFirmwareSettings = hfs.Status.Settings.DeepCopy()
 		prepareData.TargetFirmwareSettings = hfs.Spec.Settings.DeepCopy()
 	}
 
-	provResult, started, err := prov.Prepare(prepareData, dirty)
+	provResult, started, err := prov.Prepare(prepareData, bmhDirty || hfsDirty)
```
zaneb, line 983: This is a change in behaviour... IIRC the flag causes us to start a new cleaning.

Previously we keep passing true until it returns started=true, at which point we save the provisioning settings and bmhDirty becomes false.
It looks to me like hfsDirty will continue to return true even after cleaning is started? I fear this will mess with the error handling.

bfournie: > This is a change in behaviour... IIRC the flag causes us to start a new cleaning.
> 
> Previously we keep passing true until it returns started=true, at which point we save the provisioning settings and bmhDirty becomes false. It looks to me like hfsDirty will continue to return true even after cleaning is started? I fear this will mess with the error handling.

Good point. I added another condition to the HFS to track whether cleaning has started and that will allow the error to be returned if Ironic fails the cleaning step. As we talked about using the condition prevents needed to add another field to the CRD.

controllers/metal3.io/host_state_machine.go
```diff
@@ -417,6 +417,13 @@ func (hsm *hostStateMachine) handleAvailable(info *reconcileInfo) actionResult {
 	} else if dirty {
 		hsm.NextState = metal3v1alpha1.StatePreparing
 		return actionComplete{}
+	} else {
```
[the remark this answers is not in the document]
bfournie: Fixed

controllers/metal3.io/host_state_machine.go
bfournie: Fixed

controllers/metal3.io/host_state_machine.go
bfournie: Good point, added a test for getHostFirmwareSettings()

pkg/provisioner/ironic/ironic.go
```diff
@@ -1310,7 +1310,8 @@ func (p *ironicProvisioner) Prepare(data provisioner.PrepareData, unprepared boo
 		// When clean failed, we need to clean host provisioning settings.
 		// If unprepared is false, means the settings aren't cleared.
 		// So we can't set the node's state to manageable, until the settings are cleared.
-		if !unprepared {
+		// Also check for late binding resources in which the settings are cleared after cleaning
+		if !unprepared || lateBinding {
```
zaneb, line 1314: I think this just has the opposite problem: if a previous attempt to change only the HFS (i.e. the BMH settings are not dirty) failed and we recorded that error, then when the controller retries we'll just report this error straight away and never move back to manageable. (The controller will clear the firmware settings from the BMH, meaning unprepared will be now true, but lateBinding will also remain true so we'll always take this path.)

I'm thinking if we name this new flag `force` (for consistency with other Provisioner APIs), check `if !force` here, and pass `bmhDirty` for force:

```
prov.Prepare(prepareData, bmhDirty || hfsDirty, bmhDirty)
```

(this patch is the equivalent of passing `bmhDirty && !hfsDirty` for `force`)

zaneb: I guess my suggestion relies on us being able to distinguish between the existing firmware/RAID settings having been cleared due to an error, and them having been set to nothing due to nothing being specified in the Spec. I don't think this is the case, because otherwise it would trigger manual cleaning on every host.

bfournie: Yes the fix works to use bmhDirty for force. I was able to simulate a cleaning failure when changing the HFS settings and get the error back. I was also able to recover after fixing the situation that caused the cleaning failure:
```
bmh/ostest-worker-1 -n openshift-machine-api
NAME              STATE       CONSUMER                      ONLINE   ERROR   AGE
ostest-worker-1   preparing   ostest-qdqsj-worker-0-cww5k   false            78m
 'apply_configuration', 'abortable': False, 'priority': 0}: Redfish exception occurred. Error: Redfish BIOS apply configuration failed for node 5c2bb81b-6736-43f2-a412-1c021aa76d5a. Error: HTTP PATCH http://192.168.111.1:8000/redfish/v1/Systems/193c3aa9-1836-4575-b52c-b6359c1cd38e/BIOS/Settings returned code 500. Base.1.0.GeneralError: cannot serialize 11 (type int) Extended information: [{'@odata.type': '/redfish/v1/$metadata#Message.1.0.0.Message', 'MessageId': 'Base.1.0.GeneralError'}]
```
Thanks

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -966,45 +966,50 @@ func (r *BareMetalHostReconciler) actionPreparing(prov provisioner.Provisioner,
...
 		return actionContinue{subResourceNotReadyRetryDelay}
 	}
-	if hfs != nil {
+	if hfsDirty {
 		prepareData.ActualFirmwareSettings = hfs.Status.Settings.DeepCopy()
 		prepareData.TargetFirmwareSettings = hfs.Spec.Settings.DeepCopy()
 	}
 
-	provResult, started, err := prov.Prepare(prepareData, dirty)
+	provResult, started, err := prov.Prepare(prepareData, bmhDirty || hfsDirty, bmhDirty)
```
zaneb, line 984: I just realised I've been massively overthinking this.

The `force` flag now works exactly the same as all the other Provisioner APIs that have one. So we can pass the equivalent thing, which would be `info.host.Status.ErrorType == metal3v1alpha1.PreparationError`.

I think passing `bmhDirty` has a bug in the case where we are retrying but there is no difference between the Firmware/RAID settings in the BMH Spec and the empty ones in the Status.

zaneb (commented):
I finally figured out what to do with the force flag (see inline).

My remaining worry is that the started flag is going to continue to be false until the HFS reports that the settings are updated. If there is any chance that this controller sees the ironic Node return to `manageable` state while the HFS is still saying the settings differ (note that we may be reading from an outdated cache), then we will embark on an unnecessary second round of manual cleaning. Is there any mechanism to prevent this?

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -966,45 +966,51 @@ func (r *BareMetalHostReconciler) actionPreparing(prov provisioner.Provisioner,
...
-	if dirty && started {
+	if bmhDirty && started {
 		info.log.Info("saving host provisioning settings")
 		_, err := saveHostProvisioningSettings(info.host)
 		if err != nil {
 			return actionError{errors.Wrap(err, "could not save the host provisioning settings")}
 		}
 	}
+
+	updateErrMsg := false
```
zaneb, line 1007: This is not entirely correct, because if there wasn't a previous error but we did save the host provisioning settings, then we also need to write. This should probably be something like:

```
needsUpdate := bmhDirty && started
```

controllers/metal3.io/hostfirmwaresettings_controller.go
```diff
@@ -43,7 +43,7 @@ import (
 const (
 	provisionerRetryDelay          = time.Second * 30
 	resourceNotAvailableRetryDelay = time.Second * 30
-	reconcilerRequeueDelay         = time.Minute * 5
+	reconcilerRequeueDelay         = time.Minute * 1
```
zaneb, line 46: Maybe we should consider reducing this only when FirmwareSettingsChangeDetected is true?

controllers/metal3.io/baremetalhost_controller.go
bfournie: Ah, I see what you mean, fixed to restore initial way

controllers/metal3.io/hostfirmwaresettings_controller.go
bfournie: That's a really good idea. Changed.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -966,45 +966,50 @@ func (r *BareMetalHostReconciler) actionPreparing(prov provisioner.Provisioner,
...
+		info.host.Status.ErrorType == metal3v1alpha1.PreparationError)
 
 	if err != nil {
 		return actionError{errors.Wrap(err, "error preparing host")}
 	}
 
 	if provResult.ErrorMessage != "" {
 		info.log.Info("handling cleaning error in controller")
-		clearHostProvisioningSettings(info.host)
+		if bmhDirty {
```
andfasano, line 993: Shouldn't be the log message within the `if bmhDirty {` block?

controllers/metal3.io/baremetalhost_controller_test.go
```diff
@@ -78,6 +79,35 @@ func newHost(name string, spec *metal3v1alpha1.BareMetalHostSpec) *metal3v1alpha
 	}
 }
 
+func newHostFirmwareSettings(host *metal3v1alpha1.BareMetalHost, conditions []metav1.Condition) *metal3v1alpha1.HostFirmwareSettings {
+	hfs := &metal3v1alpha1.HostFirmwareSettings{
+		// TypeMeta: metav1.TypeMeta{
```
andfasano, line 84: nit: these comments could be removed

controllers/metal3.io/baremetalhost_controller_test.go
```diff
@@ -2522,3 +2552,117 @@ func TestPreprovImageAvailable(t *testing.T) {
...
+				t.Fatal(err)
+			}
+			assert.Equal(t, tc.Dirty, dirty, "dirty flag did not match")
+		})
+	}
+
+}
+
+func TestTransitionToPreparing(t *testing.T) {
+	t.Run("bmh firmware change transition", func(t *testing.T) {
```
andfasano, line 2611: (not blocking) Would be nice to have the usual cases setup, given that there is some duplication in the test code

controllers/metal3.io/host_state_machine.go
```diff
@@ -419,6 +419,14 @@ func (hsm *hostStateMachine) handleAvailable(info *reconcileInfo) actionResult {
 		return actionComplete{}
 	}
 
+	// Check if hostFirmwareSettings have changed
+	if dirty, _, err := hsm.Reconciler.getHostFirmwareSettings(info); err != nil {
```
andfasano, line 423: Wouldn't be  better to check it together with `getHostProvisioningSettings`? 

nit: just noticed that `getHostProvisioningSettings` is not part of `BareMetalHostReconciler`

controllers/metal3.io/hostfirmwaresettings_controller.go
```diff
@@ -162,6 +163,10 @@ func (r *HostFirmwareSettingsReconciler) Reconcile(ctx context.Context, req ctrl
 	}
 
 	// requeue to run again after delay
+	if meta.IsStatusConditionTrue(info.hfs.Status.Conditions, string(metal3v1alpha1.FirmwareSettingsChangeDetected)) {
+		// If the Spec is different that Status shorten the time between queries to get Ironic updates after cleaning
```
andfasano, line 167: This comment is a little bit unclear to me

controllers/metal3.io/baremetalhost_controller.go
bfournie: Yes, changed.

controllers/metal3.io/baremetalhost_controller_test.go
bfournie: Yes, good point, Removed.

controllers/metal3.io/baremetalhost_controller_test.go
bfournie: Will discuss this further but would like to do this as a followup.

controllers/metal3.io/host_state_machine.go
bfournie: I think there needs to be a separate function for getHostFirmwareSettings() as its needed in actionPreparing() https://github.com/metal3-io/baremetal-operator/blob/main/controllers/metal3.io/baremetalhost_controller.go#L973>

Since in handleAvailable() its just a check to determine whether we move to Preparing I think its safe to have them separate, if we move to Preparing first because of getHostProvisioningSettings() that is OK.

andfasano: Ok, that was I noticed, the fact that the action could have been marked as completed prematurely

controllers/metal3.io/hostfirmwaresettings_controller.go
bfournie: Fixed

controllers/metal3.io/baremetalhost_controller_test.go
bfournie: Updated tests to split TestTransitionToPreparing into 2 separate functions.

dtantsur:
Not sure I like the `force` vs `unprepared`.. but let it be for now.

/lgtm

Issue #1079, opened 2022-01-31
Question: is "dirty" still used to avoid infinite reconcile loops?


janiskemper:
Since ```actionFailed``` is always returned with ```dirty: true``` (https://github.com/metal3-io/baremetal-operator/blob/main/controllers/metal3.io/baremetalhost_controller.go#L354), how is this parameter still used? I understand that it is meant to avoid "update" calls in some cases. However, updates altering the error counter are already filtered out. Therefore, I think it doesn't follow the original purpose anymore, still indicated by the comment in https://github.com/metal3-io/baremetal-operator/blob/07a1a2ef8564839557d97d26c055ad98e917f2b9/controllers/metal3.io/baremetalhost_controller.go#L257. Is that correct?
Thanks in advance for the clarification :)

zaneb:
You appear to be right that there's no need to have a `dirty` field in `actionFailed`, and its `Dirty()` method could just always return `true`. I don't recall the history of why I thought that was needed previously. I suspect it dated from before we had an error count to update. Feel free to remove it.

In general the `Dirty()` method of the `actionResult` interface exists to prevent unnecessary writes to the k8s API.

You're correct that unnecessary reconciles are now filtered out by https://github.com/metal3-io/baremetal-operator/blob/07a1a2ef8564839557d97d26c055ad98e917f2b9/controllers/metal3.io/baremetalhost_controller.go#L1580-L1583. We always explicity ask for a reconcile (in the return from the `Result()` method) when we need one. The filtering came relatively late, so the comment you mention was correct until that filter was added. Feel free to tidy that up also.

janiskemper:
Thanks for clarification @zaneb. I can have a look at it when I have the time, but I suppose it would be a bigger refactoring, as it is used in a lot of tests etc. Should this issue be the base for future refactoring, or should we close this and open a new one for that?

zaneb:
It shouldn't require more than 3 lines to be changed as far as I can see. I think we can leave this issue open to cover that.

PR #1080, opened 2022-02-04
Only update the HFS status on changes

bfournie:
Only update the HFS status if the settings from Ironic have changed or if the conditions were updated.
Currently the reconciler can run an additional time because the Status is updated on each pass.

### DIFF
--- a/controllers/metal3.io/hostfirmwaresettings_controller.go
+++ b/controllers/metal3.io/hostfirmwaresettings_controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"crypto/sha256"
 	"fmt"
+	"reflect"
 	"sort"
 	"strconv"
 	"strings"
@@ -35,6 +36,8 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	metal3v1alpha1 "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
 	"github.com/metal3-io/baremetal-operator/pkg/provisioner"
@@ -179,11 +182,6 @@ func (r *HostFirmwareSettingsReconciler) updateHostFirmwareSettings(currentSetti
 		return errors.Wrap(err, "could not get/create firmware schema")
 	}
 
-	// Set hostFirmwareSetting to use this schema
-	info.hfs.Status.FirmwareSchema = &metal3v1alpha1.SchemaReference{
-		Namespace: firmwareSchema.ObjectMeta.Namespace,
-		Name:      firmwareSchema.ObjectMeta.Name}
-
 	if err = r.updateStatus(info, currentSettings, firmwareSchema); err != nil {
 		return errors.Wrap(err, "could not update hostFirmwareSettings")
 	}
@@ -194,9 +192,14 @@ func (r *HostFirmwareSettingsReconciler) updateHostFirmwareSettings(currentSetti
 // Update the HostFirmwareSettings resource using the settings and schema from provisioner
 func (r *HostFirmwareSettingsReconciler) updateStatus(info *rInfo, settings metal3v1alpha1.SettingsMap, schema *metal3v1alpha1.FirmwareSchema) (err error) {
 
-	if info.hfs.Status.Settings == nil {
-		info.hfs.Status.Settings = make(metal3v1alpha1.SettingsMap)
-	}
+	dirty := false
+	var newStatus metal3v1alpha1.HostFirmwareSettingsStatus
+	newStatus.Settings = make(metal3v1alpha1.SettingsMap)
+
+	// Set hostFirmwareSetting to use this schema
+	newStatus.FirmwareSchema = &metal3v1alpha1.SchemaReference{
+		Namespace: schema.ObjectMeta.Namespace,
+		Name:      schema.ObjectMeta.Name}
 
 	// Update Status on changes
 	for k, v := range settings {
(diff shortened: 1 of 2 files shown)


zaneb:
I suspect that after this change, reconcile will still run again once after an update (to stop that, you have to filter out status changes, see #1079 for details of how the host controller does it). Prior to this it's probably running constantly, limited only by the rate limit in the controller-runtime.

controllers/metal3.io/hostfirmwaresettings_controller.go
```diff
@@ -228,34 +235,46 @@ func (r *HostFirmwareSettingsReconciler) updateStatus(info *rInfo, settings meta
 	generation := info.hfs.GetGeneration()
 
 	if specMismatch {
-		setCondition(generation, &info.hfs.Status, metal3v1alpha1.FirmwareSettingsChangeDetected, metav1.ConditionTrue, reason, "")
+		if setCondition(generation, &newStatus, info, metal3v1alpha1.FirmwareSettingsChangeDetected, metav1.ConditionTrue, reason, "") {
```
zaneb, line 238: Don't we need to copy over the existing condtions to the newStatus first? Otherwise the state will always appear to be transitioning.

bfournie:
> I suspect that after this change, reconcile will still run again once after an update (to stop that, you have to filter out status changes, see #1079 for details of how the host controller does it). Prior to this it's probably running constantly, limited only by the rate limit in the controller-runtime.

Yes, it looks like right now when there is a change to settings (in Spec or Status) I'm seeing a 2nd reconcile run if the Conditions are updated. I should be able to filter that using an eventHandler similar to (bmh https://github.com/metal3-io/baremetal-operator/blob/main/controllers/metal3.io/baremetalhost_controller.go#L1588).

Previously the Status would get updated on each RequeueDelay (every 5 minutes) followed by a 2nd reconcile run immediately after but no more than that. I assume that all the data in the resource was the same (even the condition LastTtansitionTime) so the reconciler would not be invoked.

controllers/metal3.io/hostfirmwaresettings_controller.go
bfournie: No I don't believe so, setCondition() now updates the condition in newStatus and checks whether its different than the current one to set the dirty flag.  Status().Update will only be called if the Conditions have changed.

bfournie:
> > I suspect that after this change, reconcile will still run again once after an update (to stop that, you have to filter out status changes, see #1079 for details of how the host controller does it). Prior to this it's probably running constantly, limited only by the rate limit in the controller-runtime.
> 
> Yes, it looks like right now when there is a change to settings (in Spec or Status) I'm seeing a 2nd reconcile run if the Conditions are updated. I should be able to filter that using an eventHandler similar to (bmh https://github.com/metal3-io/baremetal-operator/blob/main/controllers/metal3.io/baremetalhost_controller.go#L1588).
> 
> Previously the Status would get updated on each RequeueDelay (every 5 minutes) followed by a 2nd reconcile run immediately after but no more than that. I assume that all the data in the resource was the same (even the condition LastTtansitionTime) so the reconciler would not be invoked.

I've added an eventHandler to check Generation, this works well to filter out changes the status e.g. Condition updates, that should not cause the reconciler to run again. PTAL.

controllers/metal3.io/hostfirmwaresettings_controller.go
```diff
@@ -332,11 +353,29 @@ func (r *HostFirmwareSettingsReconciler) SetupWithManager(mgr ctrl.Manager) erro
...
+
+	r.Log.Info("hostfirmwaresettings in event handler")
+
+	//If the update increased the resource Generation then let's process it
+	if e.ObjectNew.GetGeneration() != e.ObjectOld.GetGeneration() {
+		r.Log.Info("returning true as generation changed from event handler")
+		return true
+	}
+
+	r.Log.Info("returning false from event handler")
```
dtantsur, line 373: nit: I wonder how useful this is, especially without much context

controllers/metal3.io/hostfirmwaresettings_controller.go
```diff
@@ -205,13 +211,16 @@ func (r *HostFirmwareSettingsReconciler) updateStatus(info *rInfo, settings meta
 			continue
 		}
 
-		info.hfs.Status.Settings[k] = v
+		newStatus.Settings[k] = v
+	}
+	if !reflect.DeepEqual(info.hfs.Status.Settings, newStatus.Settings) {
+		dirty = true
```
andfasano, line 217: nit: `dirty` could directly be set in one line here, by comparing the schema or the the settings

controllers/metal3.io/hostfirmwaresettings_controller.go
bfournie: Removed this log.

controllers/metal3.io/hostfirmwaresettings_controller.go
bfournie: Good point, combined to one line

controllers/metal3.io/hostfirmwaresettings_controller.go
```diff
@@ -205,13 +208,16 @@ func (r *HostFirmwareSettingsReconciler) updateStatus(info *rInfo, settings meta
...
 	}
 
+	dirty = !reflect.DeepEqual(info.hfs.Status.FirmwareSchema, newStatus.FirmwareSchema) ||
+		!reflect.DeepEqual(info.hfs.Status.Settings, newStatus.Settings)
+
 	// Check if any Spec settings are different than Status
 	specMismatch := false
 	for k, v := range info.hfs.Spec.Settings {
-		if statusVal, ok := info.hfs.Status.Settings[k]; ok {
+		if statusVal, ok := newStatus.Settings[k]; ok {
```
andfasano, line 220: Q: IIUC, this potentially may reconcile two distinct facts, either that the user updated meanwhile the HFS specs, either that someone else changed directly the BIOS settings (maybe a manual operation?), can you please confirm? In any case, I guess it could be worth logging the event for future troubleshooting

controllers/metal3.io/hostfirmwaresettings_controller.go
```diff
@@ -228,34 +234,46 @@ func (r *HostFirmwareSettingsReconciler) updateStatus(info *rInfo, settings meta
...
+			dirty = true
 		}
 	}
 
-	t := metav1.Now()
-	info.hfs.Status.LastUpdated = &t
-	return r.Status().Update(context.TODO(), info.hfs)
+	// Update Status if it has changed
+	if dirty {
+		info.log.Info("Status has changed")
```
andfasano, line 268: I think it could be useful to log also the current status value

bfournie: I'm not sure we want to log the entire status, with hundreds of BIOS settings this can be quite large.

andfasano: Yeah definitely :) Very likely adding a log trace when the change is detected will be good enough

PR #1083, opened 2022-02-11
Use deep copy/equal for host root device hints

bfournie:
The root device hints have a rotational field which is a pointer to a bool. Use DeepCopy and DeepEqual to ensure its handled properly.


controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1309,22 +1309,23 @@ func getHostProvisioningSettings(host *metal3v1alpha1.BareMetalHost) (dirty bool
 // saveHostProvisioningSettings copies the values related to
 // provisioning that do not trigger re-provisioning into the status
 // fields of the host.
-func saveHostProvisioningSettings(host *metal3v1alpha1.BareMetalHost) (dirty bool, err error) {
+func saveHostProvisioningSettings(host *metal3v1alpha1.BareMetalHost, info *reconcileInfo) (dirty bool, err error) {
```
zaneb, line 1312: I'd be inclined to only pass the logger if that's the only thing we need from the info struct.

zaneb (commented):
/approve

PR #1086, opened 2022-02-15
Never try to update an existing instance UUID (fixes #1085)

### LINKED ISSUE #1085: BMO may try to update the node's instance_uuid
I've seen it in the CI: https://jenkins.nordix.org/job/metal3_metal3_dev_env_v1a5_integration_test_centos/23/. Probably reproducible by re-creating a BMH for a deployed node.

The problem is, instance UUID is intentionally not updatable (it serves as an optional lock). We need to be able to generate a UUID that survives the BMH recreating, probably be generating it from the address or the name.

dtantsur:
This field cannot be updated, only set or unset. This is because it
serves as a voluntary lock for node consumers. To avoid issues when
a BMH is re-created for an existing and deployed node, use UUID derived
from its BMC address instead of the BMH UUID.

### DIFF
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -884,8 +884,15 @@ func (p *ironicProvisioner) setCustomDeployUpdateOptsForNode(ironicNode *nodes.N
 }
 
 func (p *ironicProvisioner) getImageUpdateOptsForNode(ironicNode *nodes.Node, imageData *metal3v1alpha1.Image, bootMode metal3v1alpha1.BootMode, hasCustomDeploy bool, updater *nodeUpdater) {
-	// instance_uuid
-	updater.SetTopLevelOpt("instance_uuid", string(p.objectMeta.UID), ironicNode.InstanceUUID)
+	// Instance UUID marks a node as belonging to this Host.
+	instanceUUID := string(p.objectMeta.UID)
+	// Instance UUID cannot be changed (it serves as a voluntary lock), only deleted and set.
+	if ironicNode.InstanceUUID == "" {
+		updater.SetTopLevelOpt("instance_uuid", instanceUUID, ironicNode.InstanceUUID)
+	} else if ironicNode.InstanceUUID != instanceUUID {
+		// TODO(dtantsur): make this an error? having a correct instance UUID may be important for shared Ironic deployments
+		p.log.Info("unexpected value of instance UUID, not updating", "node", ironicNode.UUID, "expectedUUID", instanceUUID, "actualUUID", ironicNode.InstanceUUID)
+	}
 
 	// Secure boot is a normal capability that goes into instance_info (we
 	// also put it to properties for consistency, although it's not


zaneb:
Why would we want the UUID to survive recreating the host?
We should always delete the Node from ironic before allowing the host to be deleted. The only way to prevent that that I know of is to remove the finalizer early. We should probably delete any pre-existing Node that predated the Host, not try to use it as if it were a fresh one.

I don't really buy this explanation either though, because afaik there's no way to delete a Host and then recreate it straight into Provisioned state either.

pkg/provisioner/ironic/ironic.go
```diff
@@ -883,9 +884,21 @@ func (p *ironicProvisioner) setCustomDeployUpdateOptsForNode(ironicNode *nodes.N
 		SetTopLevelOpt("deploy_interface", "custom-agent", ironicNode.DeployInterface)
 }
 
+func (p *ironicProvisioner) getInstanceUUID() uuid.UUID {
+	// This procedure generates a predictable UUID that survives re-creating the host
+	return uuid.NewSHA1(uuid.NameSpaceURL, []byte(p.bmcAddress))
```
zaneb, line 889: I don't want to do this, IMHO it's masking a real problem.

pkg/provisioner/ironic/ironic.go
```diff
@@ -883,9 +884,21 @@ func (p *ironicProvisioner) setCustomDeployUpdateOptsForNode(ironicNode *nodes.N
...
 func (p *ironicProvisioner) getImageUpdateOptsForNode(ironicNode *nodes.Node, imageData *metal3v1alpha1.Image, bootMode metal3v1alpha1.BootMode, hasCustomDeploy bool, updater *nodeUpdater) {
-	// instance_uuid
-	updater.SetTopLevelOpt("instance_uuid", string(p.objectMeta.UID), ironicNode.InstanceUUID)
+	// Instance UUID marks a node as belonging to this Host.
+	instanceUUID := p.getInstanceUUID().String()
+	// Instance UUID cannot be changed (it serves as a voluntary lock), only deleted and set.
+	if ironicNode.InstanceUUID == "" {
+		updater.SetTopLevelOpt("instance_uuid", instanceUUID, ironicNode.InstanceUUID)
+	} else if ironicNode.InstanceUUID != instanceUUID && ironicNode.InstanceUUID != string(p.objectMeta.UID) {
+		// TODO(dtantsur): make this an error? it may be important for shared Ironic deployments that are shared
```
zaneb, line 899: I agree this should be an error. Ironic only returns 409 even though it will never come right, so we do need this detection code.
If we have to recover automatically then we should do it by deleting the Node. I'd prefer we address the root cause in the tests instead, but I'm starting to think that deleting the Node might be safe. Maybe.

pkg/provisioner/ironic/ironic.go
```diff
@@ -884,8 +884,15 @@ func (p *ironicProvisioner) setCustomDeployUpdateOptsForNode(ironicNode *nodes.N
...
 func (p *ironicProvisioner) getImageUpdateOptsForNode(ironicNode *nodes.Node, imageData *metal3v1alpha1.Image, bootMode metal3v1alpha1.BootMode, hasCustomDeploy bool, updater *nodeUpdater) {
-	// instance_uuid
-	updater.SetTopLevelOpt("instance_uuid", string(p.objectMeta.UID), ironicNode.InstanceUUID)
+	// Instance UUID marks a node as belonging to this Host.
+	instanceUUID := string(p.objectMeta.UID)
+	// Instance UUID cannot be changed (it serves as a voluntary lock), only deleted and set.
+	if ironicNode.InstanceUUID == "" {
+		updater.SetTopLevelOpt("instance_uuid", instanceUUID, ironicNode.InstanceUUID)
+	} else if ironicNode.InstanceUUID != instanceUUID {
+		// TODO(dtantsur): make this an error? having a correct instance UUID may be important for shared Ironic deployments
```
zaneb, line 893: The actual scenario seen in #1085 is that a reconcile for a stale CR is still in progress. In that case, returning a transient error would be the right thing to do as it will end the reconciliation without changing anything (further) in ironic that doesn't actually belong to this resource any more. Previously the write error from ironic was somewhat protecting us from that.

dtantsur: I'm not sure what you mean by "protect" here, it ended up with the BMH in some weird state where it couldn't be used any more.

furkatgofurov7:
/remove-lifecycle stale

This seems to be visible in CI time to time so this would be still relevant

furkatgofurov7:
Please label the PR appropriately based on https://github.com/metal3-io/baremetal-operator/blob/main/CONTRIBUTING.md#contributing-a-patch to make it easier to sort it out while releasing in the changelog, thanks!

PR #1089, opened 2022-02-21
imageprovider: an explicit state when an image is not ready

dtantsur:


controllers/metal3.io/preprovisioningimage_controller.go
```diff
@@ -110,6 +110,13 @@ func (r *PreprovisioningImageReconciler) Reconcile(ctx context.Context, req ctrl
 		log.Info("requeuing to check for secret", "after", delay)
 		result.RequeueAfter = delay
 	}
+
+	notReady := imageprovider.ImageNotReady{}
+	if errors.As(err, &notReady) {
```
ardaguclu, line 115: This isn't wrong and probably will work. But in such cases;
I prefer such usage;

```
if inr, ok := err.(imageprovider.ImageNotReady{}); ok {
				result.RequeueAfter = inr.RequeueAfter
			}
```
This looks more simpler to me, but as said your solution also looks good to me.

dtantsur: Your idea looks better, but there is an interesting discussion in the overview of https://pkg.go.dev/errors. Do you think it applies?

dtantsur: > In this case, it's not clear to me where such kind of error should be generated 

In the implementation. There is no guarantee that an image will be available quickly enough to return it immediately.

controllers/metal3.io/preprovisioningimage_controller.go
```diff
@@ -110,6 +110,13 @@ func (r *PreprovisioningImageReconciler) Reconcile(ctx context.Context, req ctrl
...
+		log.Info("image is not ready yet, requeuing", "after", notReady.RequeueAfter)
+		result.RequeueAfter = notReady.RequeueAfter
```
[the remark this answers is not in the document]
dtantsur: A potential counter-argument is that it's the implementation, not the controller, that knows how long to wait.

controllers/metal3.io/preprovisioningimage_controller.go
```diff
@@ -110,6 +110,13 @@ func (r *PreprovisioningImageReconciler) Reconcile(ctx context.Context, req ctrl
...
+	}
```
zaneb, line 118: This is not updating the status to say that the image is not ready, just leaving the CR in whatever state it was. I think you also/instead need to check for this right after the check for ImageBuildInvalid.

controllers/metal3.io/preprovisioningimage_controller.go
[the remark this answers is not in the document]
zaneb: This is similar to the BMH controller, where the provisioner tells the controller what the requeue delay should be.

You probably want to set a minimum value here. Otherwise, you must always set `result.Requeue = true` otherwise a `RequeueAfter` of 0 (the default!) will not result in a requeue and the CR will ~never be updated again.

controllers/metal3.io/preprovisioningimage_controller.go
zaneb: It's unlikely that anyone would wrap this error (slightly more likely they'd use it to wrap something else), but even so I think it's preferable to use `errors.As()` just as a general policy.

It's kind of like the difference between `type(x) == Foo` vs. `isinstance(x, Foo)` in Python. (But not the same thing, obviously.)

pkg/imageprovider/invalid.go
```diff
@@ -17,3 +24,7 @@ func (ibf ImageBuildInvalid) Unwrap() error {
 func BuildInvalidError(err error) ImageBuildInvalid {
 	return ImageBuildInvalid{err: err}
 }
+
```
zaneb, line 27: Can we try to keep the definition and its methods together instead of interleaving them?

controllers/metal3.io/preprovisioningimage_controller.go
dtantsur: I'm not sure I'm doing the right thing, but I've tried. Please check.

controllers/metal3.io/preprovisioningimage_controller.go
andfasano: > > In this case, it's not clear to me where such kind of error should be generated
> 
> In the implementation. There is no guarantee that an image will be available quickly enough to return it immediately.

Just for the sake of clarity, do you mean by one of the methods of the `ImageProvider` interface? (In such case I'm not sure if we have some documentation for the expected errors, it could be worth adding it for making clearer the relationships between the reconciler and the image provider; ie `BuildImage` could return a `ImageBuildInvalid` error)

controllers/metal3.io/preprovisioningimage_controller.go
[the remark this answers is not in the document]
andfasano: Indeed, result and error are kept as a separate return arguments

controllers/metal3.io/preprovisioningimage_controller.go
dtantsur: I'm definitely not changing the whole `BuildImage` signature for one error :-/ This value is specific to the error and should be there, but if you insist, I'll remove it and just have a constant retry.

As to your first comment, we don't have an access to `Result` in `update`, thus I cannot move the check there.

andfasano: IMHO a custom error type should contain only the strictly necessary additional fields to report the error, help the troubleshooting and the debugging activity. 
The current retry approach seems ok to me (as the request will be automatically re-enqueued with a rate limiting), but in case I'd consider ok to review the `ImageProvider` api if required.

/lgtm

controllers/metal3.io/preprovisioningimage_controller.go
```diff
@@ -110,6 +110,16 @@ func (r *PreprovisioningImageReconciler) Reconcile(ctx context.Context, req ctrl
 		log.Info("requeuing to check for secret", "after", delay)
 		result.RequeueAfter = delay
 	}
+
+	notReady := imageprovider.ImageNotReady{}
+	if errors.As(err, &notReady) {
+		log.Info("image is not ready yet, requeuing")
+		if setUnready(img.GetGeneration(), &img.Status, "Image is not ready yet") {
```
zaneb, line 117: It would be better to pass `err.Error()` as the reason.

controllers/metal3.io/preprovisioningimage_controller.go
```diff
@@ -110,6 +110,16 @@ func (r *PreprovisioningImageReconciler) Reconcile(ctx context.Context, req ctrl
...
+			changed = true
+		}
+		result.Requeue = true
```
zaneb, line 120: You need to set a delay here, otherwise we will have a hot loop.

controllers/metal3.io/preprovisioningimage_controller.go
zaneb: This looks fine, but the only reason for having the code here was because it was setting the retry delay from data in the error. I wasn't opposed to doing that, but if we are not going to then this code definitely belongs at the point where the error is first returned, i.e. https://github.com/metal3-io/baremetal-operator/blob/05d12b6768a9989a9a4e61dad6cd1f9e84a6e078/controllers/metal3.io/preprovisioningimage_controller.go#L186

PR #1090, opened 2022-02-23
Remove the option to disable authentication

dtantsur:
Its use case is questionable and definitely does not justify the
complexity it adds to deployment files.


dtantsur:
/test generate

not reproducible locally, these files never existed

mboukhalfa:
Thanks @dtantsur for the simplicity added, indeed no-auth was complicating ironic manifests and deployment script without any required use case. When rebased I will take another look

PR #1092, opened 2022-03-01
Copy labels from BaremetalHost to PreprovisioningImage

zaneb:
This may be needed e.g. to find linked objects.


dtantsur:
What is needed to make this PR ready, only tests?

zaneb:
I think so, yes.

PR #1093, opened 2022-03-07
Clarify documentation on BMC formats

dtantsur:
* Specify which driver uses which technology.
* +http and +https are supported for most drivers


jtomasek (approved):

zhouhao3 (approved):

andfasano:
Thanks @dtantsur it looks definitely better now

PR #1094, opened 2022-03-08
Expose CheckRAIDInterface from the RAID module

dtantsur:
We need to use it in the terraform provider as well:
https://github.com/openshift-metal3/terraform-provider-ironic/pull/53.

Also remove no-op RAID testing bits from the access tests.

### DIFF
--- a/pkg/provisioner/ironic/raid.go
+++ b/pkg/provisioner/ironic/raid.go
@@ -19,7 +19,7 @@ const (
 
 // setTargetRAIDCfg set the RAID settings to the ironic Node for RAID configuration steps
 func setTargetRAIDCfg(p *ironicProvisioner, raidInterface string, ironicNode *nodes.Node, data provisioner.PrepareData) (provisioner.Result, error) {
-	err := checkRAIDConfigure(raidInterface, data.TargetRAIDConfig)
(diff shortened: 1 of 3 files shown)


zhouhao3 (approved):

pkg/provisioner/ironic/raid.go
```diff
@@ -273,18 +273,21 @@ func BuildRAIDCleanSteps(raidInterface string, target *metal3v1alpha1.RAIDConfig
 	return
 }
 
-func checkRAIDConfigure(raidInterface string, raid *metal3v1alpha1.RAIDConfig) error {
+// CheckRAIDInterface checks the current RAID interface against the requested configuration
+func CheckRAIDInterface(raidInterface string, target *metal3v1alpha1.RAIDConfig) error {
```
andfasano, line 277: I'm likely overthinking about it, but instead of making public directly this method, could be worth maybe exposing it through a more generic validation interface? I'm wondering if we could have similar needs in future (still in the validation area), and providing a single entry point for the external user may avoid exposing directly implementation details, and possibly help the maintenance.

dtantsur: Maybe, but let's save the pressing problem now and think about refactoring later? This change (previously a part of the software RAID fix) has been blocking the RAID work on the terraform side for months already.

PR #1096, opened 2022-03-16
Fix NAMEPREFIX default value

mboukhalfa:
Update NAMEPREFIX default value to `baremetal-operator` instead of `capm3`.
`capm3` was used as default before https://github.com/metal3-io/metal3-dev-env/pull/721. but since BMO now is deployed out side CAPM3, the prefix `capm3` seems not used anywhere instead `baremetal-operator` used as default in dev-env https://github.com/metal3-io/metal3-dev-env/blob/38b624ca44d7684ae3123e75a6655c125749f51b/lib/common.sh#L204


Rozzii (commented):
LGTM

andfasano:
What's the underlying requirement for this change? The commit message doesn't provide any additional information, can you please expand it?

mboukhalfa:
> What's the underlying requirement for this change? The commit message doesn't provide any additional information, can you please expand it?

The NAMEPREFIX is used to create the namespace, and I assume previously BMO was running in `capm3` namespace but now it is deployed outside capm3 and run in different namespace. The default namespace prefix used in DEV-ENV now is `baremetal-operator` and it is always overriding this default value exported here in BMO. 

In some tests for example e2e we decided to move all variable to the e2e config file instead of exporting them so BMO ends up using this default value capm3 for namespace while we were expecting `baremetal-operator`  so we are exporting `NAMEPREFIX` in CAPM3 only to override this environment variable in BMO. I thought it would be convenient for dev-env and metal3 tests to set the most used `NAMEPREFIX=baremetal-operator` as default instead of `capm3` that I do not see it being used for BMO anymore.

fmuyassarov (commented):
/approve

PR #1097, opened 2022-03-21
Remove amd64 and linux to ensure multiarch container build

jeffdyoung:
Hello, I am testing the baremetal-operator on arm, and noticed that this builds an arm container with an x86 image. I'd like to remove the hard-coded arch so the image builds based on the os arch.


zaneb:
Should this be done for https://github.com/metal3-io/baremetal-operator/blob/main/Tiltfile#L101 as well?

/approve

PR #973, opened 2021-09-09
Add AGE column to BMH CRDs

kashifest:
This adds the AGE column to BMH CRDs to make the printed output consistent with the built in K8s resources.

Keeping things consistent with CAPI# kubernetes-sigs/cluster-api#5180

PR #1099, opened 2022-03-22
HardwareData custom resource for host inspection data

fmuyassarov:
Implements: https://github.com/metal3-io/metal3-docs/blob/main/design/baremetal-operator/hostdata_crd.md
Fixes: https://github.com/metal3-io/cluster-api-provider-metal3/issues/266

Some of the functionalities this patch adds:
- [x] when pivoted, inspection data is restored from hardwareData
- [x] when BMH deleted, hardwareData is also deleted. Maintained via ownerReferences
- [x] when inspection disabled, hardwareData is not created
- [x] hostData can not be deleted until owner BMH is not deleted. This is maintained via finalizer
- [x] requesting inspection of `available` BMH, re-creates the hardwareData with the latest inspection data obtained from Ironic

This requires changes on the way we store host status on the statusAnnotation. To be exact, we don't store hardwareDetails into the statusAnnotation anymore. See https://github.com/metal3-io/cluster-api-provider-metal3/pull/573.

For the move to succeed, we need to use clusterctl move annotations on the hardwareData. Related adjustment for the pivoting
workflow in the metal3-dev-env based CI is in https://github.com/metal3-io/metal3-dev-env/pull/969.

The PR is missing unit testing and I'm working on that commit now.


controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -125,6 +127,27 @@ func (r *BareMetalHostReconciler) Reconcile(ctx context.Context, request ctrl.Re
...
+	// Fetch the HostData
+	hostData := &metal3v1alpha1.HostData{}
+	hostDataKey := client.ObjectKey{
+		Name:      host.Name,
+		Namespace: host.Namespace,
+	}
+	err = r.Get(ctx, hostDataKey, hostData)
+	if err != nil && !k8serrors.IsNotFound(err) {
+		fmt.Println("Handle it")
+	}
```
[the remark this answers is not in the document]
zaneb: It's a tricky one. The failure could be due to e.g. missing CRD or RBAC permissions. Do we want to prevent all updates to BMHs in those circumstances? I suspect not, we probably just want to log the error and continue (but not attempt to create a replacement or anything).

config/rbac/hostdata_editor_role.yaml
```diff
@@ -0,0 +1,24 @@
...
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - metal3.io.metal3.io
+  resources:
+  - hostdata/status
+  verbs:
+  - get
```
lentzi90, line 24: Is this needed when it has access to the `hostdata` resources without the `/status`?

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -125,6 +127,27 @@ func (r *BareMetalHostReconciler) Reconcile(ctx context.Context, request ctrl.Re
...
+
+	// Host is being deleted, so we delete the finalizer from the hostData to allow its deletion.
+	if !host.DeletionTimestamp.IsZero() {
+		if controllerutil.ContainsFinalizer(hostData, hostDataFinalizer) {
```
furkatgofurov7, line 143: Is there any use case where Finalizer would not exist?

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -125,6 +127,27 @@ func (r *BareMetalHostReconciler) Reconcile(ctx context.Context, request ctrl.Re
...
+			controllerutil.RemoveFinalizer(hostData, hostDataFinalizer)
```
furkatgofurov7, line 144: It would be helpful to have a log about successful finalizer removal

config/rbac/role.yaml
```diff
@@ -86,6 +86,24 @@ rules:
...
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - metal3.io
+  resources:
+  - hostdata/finalizers
```
furkatgofurov7, line 104: This is new to me, why do we need to add `finalizers` in rbac? also, do not you need to add `hostdata/status` in this list?

config/crd/bases/metal3.io_hostdata.yaml
```diff
@@ -0,0 +1,209 @@
...
+    shortNames:
+    - hd
+    singular: hostdata
+  scope: Namespaced
+  versions:
+  - additionalPrinterColumns:
+    - description: Time duration since creation of HostData
+      jsonPath: .metadata.creationTimestamp
+      name: Age
+      type: date
```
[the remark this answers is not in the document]
fmuyassarov: What you said is correct and felt weird to me too at the beginning too. But surprisingly it works :man-shrugging: and we already use it for [BMH](https://github.com/metal3-io/baremetal-operator/pull/973)

Another k8s example: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#additional-printer-columns

config/samples/metal3.io_v1alpha1_hostdata.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: metal3.io.metal3.io/v1alpha1
+kind: HostData
+metadata:
+  name: hostdata-sample
+spec:
+  # TODO(user): Add fields here
```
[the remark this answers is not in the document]
fmuyassarov: it is autogenerated indeed, should we keep it ? not sure, but I can similar auto code for other CRDs in https://github.com/metal3-io/baremetal-operator/tree/main/config/samples.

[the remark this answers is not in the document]
fmuyassarov: I don't see a point of having it, I will be +1 to remove it actually

config/rbac/role.yaml
fmuyassarov: I don't know how status ended up here, but it there is no status in the CRD, probably it is a leftoover from the clean up.
hostdata/finalizers is necessary to update finalizers on hostData upon deletion

config/crd/bases/metal3.io_hostdata.yaml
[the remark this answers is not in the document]
lentzi90: I see, thanks for explaining!

config/rbac/hostdata_editor_role.yaml
fmuyassarov: good catch, removed it

lentzi90 (approved):

controllers/metal3.io/baremetalhost_controller.go
furkatgofurov7: What would or expected to happen when finalizer is manually deleted?

fmuyassarov: it allows the hostData to be deleted even if BMH doesn't have deletion timestamp

furkatgofurov7: My concern was, if user deletes finalizer manually before host deletion, and when actual deletion happens it won't be able to remove manually deleted finalizer and next try to [update](https://github.com/metal3-io/baremetal-operator/pull/1099/files#diff-a69d9029388ab766ed36b32180145f52785a9d4a153775510dbddfa928a72e1cR146-R148) hostData, which should throw an error I believe. I think we need to do the update inside the finalizer contains condition

fmuyassarov: It won't throw an error during update, only if the client fails to update the resource on the API. But, it makes sense to add the update to within that if, because, there is no point of updating the resource if no change was made.

fmuyassarov: updated

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -78,7 +79,6 @@ type reconcileInfo struct {
 	request           ctrl.Request
 	bmcCredsSecret    *corev1.Secret
 	events            []corev1.Event
-	errorMessage      string
```
zhouhao3, line 81: Why does it need to be removed? It looks like this change has nothing to do with the current commit, I think it's better to make it a separate commit or a separate PR to avoid misunderstandings.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -138,35 +162,36 @@ func (r *BareMetalHostReconciler) Reconcile(ctx context.Context, request ctrl.Re
...
-				t := metav1.Now()
-				host.Status.LastUpdated = &t
-			}
-			errStatus := r.Status().Update(ctx, host)
-			if errStatus != nil {
-				return ctrl.Result{}, errors.Wrap(errStatus, "Could not restore status from annotation")
-			}
-			return ctrl.Result{Requeue: true}, nil
 		}
-		reqLogger.Info("No status cache found")
```
zhouhao3, line 158: It looks like you removed this line of log output, do we no longer need this log info? Or you forget to copy this line?

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -906,6 +928,54 @@ func (r *BareMetalHostReconciler) actionInspecting(prov provisioner.Provisioner,
...
+				*metav1.NewControllerRef(info.host, metal3v1alpha1.GroupVersion.WithKind("BareMetalHost")),
+			},
+		},
+		Spec: metal3v1alpha1.HardwareDataSpec{
+			HardwareDetails: details,
+		},
+	}
+
+	err = r.Client.Get(context.Background(), hardwareDataKey, hardwareData)
+	if err == nil {
```
zhouhao3, line 960: Do we also need to consider the failure to obtain hardwareDate(hardwareData exists, but getting failed for some reason)? This case should be requeued to fetch again.

controllers/metal3.io/baremetalhost_controller.go
fmuyassarov: I removed it because it was unused code. Sure, I can leave it as it is

controllers/metal3.io/baremetalhost_controller.go
fmuyassarov: perhaps we can assume that if `k8serrors.IsNotFound(err)` is false, then the resource exits but it got failed some other reason ? So, it would look like 
```
	err = r.Client.Get(context.Background(), hardwareDataKey, hardwareData)
	if err == nil || !k8serrors.IsNotFound(err) {
		// hardwareData found and we reached here due to request for another inspection.
		// Delete it before re-creating.
		if controllerutil.ContainsFinalizer(hardwareData, hardwareDataFinalizer) {
			controllerutil.RemoveFinalizer(hardwareData, hardwareDataFinalizer)
		}
		if err := r.Update(context.Background(), hardwareData); err != nil {
			return actionError{errors.Wrap(err, "failed to remove hardwareData finalizer")}
		}
		if err := r.Client.Delete(context.Background(), hd); err != nil {
			return actionError{errors.Wrap(err, "failed to delete hardwareData")}
		}
	}
	// either hardwareData was deleted above, or not found. We need to re-create it
	if err := r.Client.Create(context.Background(), hd); err != nil {
		return actionError{errors.Wrap(err, "failed to create hardwareData")}
	}
```

zhouhao3: seems good to me

zhouhao3 (commented):
LGTM
You also need to address unit test failures.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -138,35 +163,37 @@ func (r *BareMetalHostReconciler) Reconcile(ctx context.Context, request ctrl.Re
...
-				t := metav1.Now()
-				host.Status.LastUpdated = &t
-			}
-			errStatus := r.Status().Update(ctx, host)
-			if errStatus != nil {
-				return ctrl.Result{}, errors.Wrap(errStatus, "Could not restore status from annotation")
-			}
-			return ctrl.Result{Requeue: true}, nil
+		}
+		if host.Status.LastUpdated.IsZero() {
```
andfasano, line 174: @fmuyassarov is this status updated valid, even in case of error from `r.getHostStatusFromAnnotation(host)`?

controllers/metal3.io/baremetalhost_controller.go
andfasano: nit: maybe this log trace should be moved

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -138,35 +163,37 @@ func (r *BareMetalHostReconciler) Reconcile(ctx context.Context, request ctrl.Re
...
+		errStatus := r.Status().Update(ctx, host)
+		if errStatus != nil {
+			return ctrl.Result{}, errors.Wrap(errStatus, "Could not restore status from annotation")
 		}
+		return ctrl.Result{Requeue: true}, nil
+	}
+	// The status annotation is unneeded, as the status subresource is
+	// already present. The annotation data will get outdated, so remove it.
+	if _, present := annotations[metal3v1alpha1.StatusAnnotation]; present {
+		delete(annotations, metal3v1alpha1.StatusAnnotation)
```
andfasano, line 191: A log trace could be helpful here, for troubleshooting

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -465,10 +492,7 @@ func clearRebootAnnotations(host *metal3v1alpha1.BareMetalHost) (dirty bool) {
 // which means we don't inspect even in Inspecting state
 func inspectionDisabled(host *metal3v1alpha1.BareMetalHost) bool {
 	annotations := host.GetAnnotations()
-	if annotations[inspectAnnotationPrefix] == "disabled" {
-		return true
-	}
-	return false
+	return annotations[inspectAnnotationPrefix] == "disabled"
 }
```
andfasano, line 496: Not blocking: it could be a good occasion to move this entire block into its own function, to reduce a little bit the size of the main reconcile loop

controllers/metal3.io/baremetalhost_controller.go
fmuyassarov: good point, I'm not sure why we need it even after reading the comment, but I fixed to keep it as before

controllers/metal3.io/baremetalhost_controller.go
fmuyassarov: hmm, that's already an independent function and not part of the reconcile(), no?

controllers/metal3.io/baremetalhost_controller.go
fmuyassarov: @andfasano do you mean after ?

```
		if errStatus != nil {
			return ctrl.Result{}, errors.Wrap(errStatus, "Could not restore status from annotation")
		}
```

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -138,35 +163,38 @@ func (r *BareMetalHostReconciler) Reconcile(ctx context.Context, request ctrl.Re
 	// Check if Status is empty and status annotation is present
 	// Manually restore data.
 	if !r.hostHasStatus(host) {
-		reqLogger.Info("Fetching Status from Annotation")
+		reqLogger.Info("Reconstructing Status from hardwareData and annotation")
 		objStatus, err := r.getHostStatusFromAnnotation(host)
-		if err == nil && objStatus != nil {
+		if err == nil && objStatus != nil && hardwareData.Name != "" {
```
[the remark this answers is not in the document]
fmuyassarov: Added log trace

controllers/metal3.io/baremetalhost_controller.go
fmuyassarov: That's a good point @andfasano . Would you mind if I address this an a follow up patch not to block the initial implementation one?

controllers/metal3.io/baremetalhost_controller.go
fmuyassarov: Updated, it should be relatively close to the previous state. PTAL @andfasano

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -138,12 +163,18 @@ func (r *BareMetalHostReconciler) Reconcile(ctx context.Context, request ctrl.Re
...
 		if err == nil && objStatus != nil {
+			if hardwareData.Name != "" {
+				host.Status.HardwareDetails = hardwareData.Spec.HardwareDetails
+				host.Status = *objStatus
+				reqLogger.Info("updating baremetalHost status with hardwareData")
+			}
 			host.Status = *objStatus
-			if host.Status.LastUpdated.IsZero() {
-				// Ensure the LastUpdated timestamp in set to avoid
+			if err == nil && host.Status.LastUpdated.IsZero() {
```
andfasano, line 176: Isn't redundant this nil check, given the previous one?

controllers/metal3.io/baremetalhost_controller.go
andfasano: Sounds fine for me

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -138,12 +163,18 @@ func (r *BareMetalHostReconciler) Reconcile(ctx context.Context, request ctrl.Re
...
 	// Manually restore data.
 	if !r.hostHasStatus(host) {
-		reqLogger.Info("Fetching Status from Annotation")
+		reqLogger.Info("Reconstructing Status from hardwareData and annotation")
 		objStatus, err := r.getHostStatusFromAnnotation(host)
+
 		if err == nil && objStatus != nil {
+			if hardwareData.Name != "" {
+				host.Status.HardwareDetails = hardwareData.Spec.HardwareDetails
+				host.Status = *objStatus
```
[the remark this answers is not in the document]
fmuyassarov: Good catch @andfasano . Fixed

controllers/metal3.io/baremetalhost_controller.go
fmuyassarov: It is indeed, thanks for pointing out. Fixed.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -138,12 +163,17 @@ func (r *BareMetalHostReconciler) Reconcile(ctx context.Context, request ctrl.Re
...
+		reqLogger.Info("Reconstructing Status from hardwareData and annotation")
 		objStatus, err := r.getHostStatusFromAnnotation(host)
+
 		if err == nil && objStatus != nil {
+			if hardwareData.Name != "" {
+				host.Status.HardwareDetails = hardwareData.Spec.HardwareDetails
+				reqLogger.Info("updating baremetalHost status with hardwareData")
+			}
 			host.Status = *objStatus
 			if host.Status.LastUpdated.IsZero() {
```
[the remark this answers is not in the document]
andfasano: Thanks @fmuyassarov I think now the approach looks more clear, and also the test helped to understand better the behavior

fmuyassarov:
/test-v1b1-centos-integration
/test-v1b1-ubuntu-integration

Thanks for reviews @andfasano . I have squashed last commit now. Would you mind to give approve label ?

andfasano:
@fmuyassarov looks like there some problem in the integration tests?

fmuyassarov:
@andfasano indeed it was infra issue. Ubuntu job passed after re-triggering. I will re-run Centos job too now. But I believe otherwise it is good to go.
/test-v1b1-centos-integration

fmuyassarov:
ok same story with centos. It passed when I triggered it in https://github.com/metal3-io/baremetal-operator/pull/1099#issuecomment-1125760745. I will keep giving it a try until it get passed.

PR #1100, opened 2022-03-23
Check for empty HFS status settings

tliu2021:
If the HFS status settings have not been populated, does not generate the
Invalid BIOS Setting events.

If the changes are detected, but the HFS status settings are empty, it waits
until hostFirmwareSettings are read.

Signed-off-by: tliu2021 <tali@redhat.com>

### DIFF
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -1428,6 +1428,11 @@ func (r *BareMetalHostReconciler) getHostFirmwareSettings(info *reconcileInfo) (
 	// Check if there are settings in the Spec that are different than the Status
 	if meta.IsStatusConditionTrue(hfs.Status.Conditions, string(metal3v1alpha1.FirmwareSettingsChangeDetected)) {
 
+		// Check if the status settings have been populated
+		if len(hfs.Status.Settings) == 0 {
+			return false, nil, errors.New("host firmware status settings not available")
+		}
+
 		if meta.IsStatusConditionTrue(hfs.Status.Conditions, string(metal3v1alpha1.FirmwareSettingsValid)) {
 			info.log.Info("hostFirmwareSettings indicating ChangeDetected", "namespacename", info.request.NamespacedName)
 			return true, hfs, nil
--- a/controllers/metal3.io/baremetalhost_controller_test.go
+++ b/controllers/metal3.io/baremetalhost_controller_test.go
@@ -2736,9 +2736,51 @@ func TestHFSTransitionToPreparing(t *testing.T) {
 			{Type: "ChangeDetected", Status: "True", Reason: "Success"},
 			{Type: "Valid", Status: "True", Reason: "Success"},
 		},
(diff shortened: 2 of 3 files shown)


controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1428,6 +1428,11 @@ func (r *BareMetalHostReconciler) getHostFirmwareSettings(info *reconcileInfo) (
 	// Check if there are settings in the Spec that are different than the Status
 	if meta.IsStatusConditionTrue(hfs.Status.Conditions, string(metal3v1alpha1.FirmwareSettingsChangeDetected)) {
 
+		// Check if the status settings have been populated
+		if len(hfs.Status.Settings) == 0 {
+			return false, nil, errors.New("host firmware status settings not available")
```
bfournie, line 1433: OK, this will prevent a transition out of Preparing in the case of, for example, the license not being set on certain hardware but if the user clears the Spec setting it will no longer be blocked.

controllers/metal3.io/baremetalhost_controller_test.go
```diff
@@ -2736,9 +2736,51 @@ func TestHFSTransitionToPreparing(t *testing.T) {
...
 
 	r.Update(goctx.TODO(), hfs)
 
 	waitForProvisioningState(t, r, host, metal3v1alpha1.StatePreparing)
 }
+
+// TestHFSEmptyStatusSettings ensures that BMH does not move to the next state in the
+// following two senarios:
+// 1. when a user provides the BIOS settings on a hardware server that does not
+// have the required license to configure BIOS
```
bfournie, line 2753: This is really the same scenario, its just that when for example the license is not provided for certain hardware there will never be data coming back from Ironic as opposed to just a delay waiting for Ironic to retrieve the data from the BMC.

controllers/metal3.io/baremetalhost_controller_test.go
```diff
@@ -2736,9 +2736,51 @@ func TestHFSTransitionToPreparing(t *testing.T) {
...
+		},
 	}
 
 	r.Update(goctx.TODO(), hfs)
 
 	waitForProvisioningState(t, r, host, metal3v1alpha1.StatePreparing)
 }
+
+// TestHFSEmptyStatusSettings ensures that BMH does not move to the next state in the
+// following two senarios:
```
bfournie, line 2751: Nit - typo "senarios"

bfournie (commented):
Change looks good. Maybe clarify the comment in TestHFSEmptyStatusSettings

controllers/metal3.io/baremetalhost_controller.go
tliu2021: Yes. I extended TestHFSEmptyStatusSettings to include this scenario.

controllers/metal3.io/baremetalhost_controller_test.go
tliu2021: The comment in TestHFSEmptyStatusSettings has been updated.

PR #1102, opened 2022-03-31
Fix make target for setting Ironic image

lentzi90:
I mistakenly thought `default` was the base of all Ironic
kustomizations, but it turns out `ironic` is the base, so that is where
we should set the image.

Edit: The problem is that some kustomizations depends on `ironic` directly instead of `default`, meaning that any changes made in `default` will not be picked up. See for example keepalived: https://github.com/metal3-io/baremetal-operator/blob/main/ironic-deployment/keepalived/kustomization.yaml

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -161,7 +161,7 @@ set-manifest-image-bmo: $(KUSTOMIZE) manifests
 .PHONY: set-manifest-image-ironic
 set-manifest-image-ironic: $(KUSTOMIZE) manifests
 	$(info Updating container image for Ironic to use ${MANIFEST_IMG}:${MANIFEST_TAG})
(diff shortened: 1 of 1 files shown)


Sunnatillo (commented):
LGTM

namnx228 (commented):
LGTM

zhouhao3 (approved):

kashifest (commented):
/lgtm

elfosardo (approved):

furkatgofurov7 (approved):

fmuyassarov:
/approve
I see quite many people giving lgtm and approve. Since looks good to everyone, giving approval too

PR #1103, opened 2022-04-01
Add kubesec to BMO

namnx228:
Kubesec is a security risk analysis tool for Kubernetes manifest files. This PR adds a github action script to BMO so it is scanned by Kubesec weekly.
Github action test result can be seen here: https://github.com/Nordix/baremetal-operator/actions/runs/2155435134
And test security alert can be seen here: https://github.com/Nordix/baremetal-operator/security/code-scanning?query=is%3Aopen+branch%3AAdd-kubesec-nam

### DIFF
--- a/.github/workflows/kubesec.yml
+++ b/.github/workflows/kubesec.yml
@@ -1,0 +1,58 @@
+name: Kubesec
+on:
+  push:
+    branches: [ main ]
+  schedule:
+    - cron: '30 7 * * 4'
+
+jobs:
+  setup:
+    runs-on: ubuntu-20.04
+    permissions:
+      actions: read
+      contents: read
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Collect all yaml
+        id: list_yaml
+        run: |
+          LIST_YAML="$(find * -type f -name '*.yaml')"
+          echo "::set-output name=value::$(IFS=$','; echo $LIST_YAML | jq -cnR '[inputs | select(length>0)]'; IFS=$'\n')"
+    outputs:
+      matrix: ${{ steps.list_yaml.outputs.value }}
+
+  lint:
+    needs: [ setup ]
+    name: Kubesec
+    runs-on: ubuntu-20.04
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+    strategy:
+      matrix:
+        value: ${{ fromJson(needs.setup.outputs.matrix) }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Run kubesec scanner
+        uses: controlplaneio/kubesec-action@43d0ddff5ffee89a6bb9f29b64cd865411137b14
+        with:
+          input: ${{ matrix.value }}
+          format: template
+          template: template/sarif.tpl
+          output: ${{ matrix.value }}.sarif
+          exit-code: "0"
+
+      - name: Save result into a variable
+        id: save_result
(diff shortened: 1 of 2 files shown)


.github/workflows/kubesec.yml
```diff
@@ -0,0 +1,64 @@
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+
+name: Kubesec
+
+on:
+  push: 
+    branches: [ Add-kubesec-nam ]
```
namnx228, line 10: This is just for testing. It will be deleted if this PR is okay to go in

.github/workflows/kubesec.yml
```diff
@@ -0,0 +1,58 @@
...
+      actions: read
+      contents: read
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Collect all yaml
+        id: list_yaml
+        run: |
+          LIST_YAML="$(find * -type f -name '*.yaml')"
```
lentzi90, line 21: I wonder if it would be good to limit this a bit or exclude some paths. For example, this will now scan `examples` and `config/samples` that I'm not sure if is useful. On the other hand, we can at least be sure that it scans everything now.
I don't have a strong opinion, just wanted to mention this.

.github/workflows/kubesec.yml
```diff
@@ -0,0 +1,58 @@
...
+    needs: [ setup ]
+    name: Kubesec
+    runs-on: ubuntu-20.04
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+    strategy: 
+      matrix: 
+        value: ${{ fromJson(needs.setup.outputs.matrix) }}
```
lentzi90, line 36: Have you considered putting all the files together in one big file instead of running separate jobs for them? I'm thinking it would take less time and resources probably, but maybe it would also be harder to tell where exactly the issue is when the scanner detects something.
No strong opinion here either, I just thought that 91 jobs is quite a lot :smile:

.github/workflows/kubesec.yml
namnx228: I agree that scanning something is not a part of production doesn't improve security much. However, the example is the first thing a new consumer/consumer approaches when they start to work with this repo, and there is no better way to show the project security portrait by tightening the security level of these example code.

lentzi90: Good point, let's keep it like this!

.github/workflows/kubesec.yml
namnx228: I tried to make it in one single file. You can see it in the first commit. However, one technical issue is that one single file is generated while github action is running, not a part of the repo, so we cannot see file preview and determine where is the issue. 
In fact, we don't need to worry about 91 jobs can consume more time because they runs parallelly. Also, github allocates unlimited resource for open source repo to use github action, so we don't need to worry about result consumption.

lentzi90: Nice! Thanks for the explanation!

lentzi90 (approved):

Rozzii (approved):
lgtm

furkatgofurov7 (approved):

fmuyassarov:
@smoshiur1237 can you squash the commits? otherwise looks good.
/hold
/approve

smoshiur1237:
Thanks @fmuyassarov squashed and remove the hold

smoshiur1237:
/test-v1b1-ubuntu-integration
test-v1b1-centos-integration

furkatgofurov7 (commented):
Adding it again since it was removed after squashing the commits 
/lgtm

PR #1104, opened 2022-04-04
Add NET_RAW capability for dnsmasq

lentzi90:
We are seeing failed jobs in the CI due to dnsmasq missing NET_RAW capability. This should solve the issue. But it would also be good to check with upstream what caused this change.


lentzi90:
/hold
Let's confirm before we merge this why the new capability is needed.

kashifest:
We have tested this locally and found out that, this requirement is a by-product of uplifting ironic base image to centos-9 stream. Test results show that dnsmasq which has centos-8 stream base image do not require NET_RAW capabilities but in centos-9 stream it does.

PR #1106, opened 2022-04-06
Set minReadySeconds for BMO and Ironic

lentzi90:
This is to make it easier to tell when the deployments are "available".
The default is to consider them available as soon as the pods have
started. But if there is some problem it is likely that the pods will
immediately crash. Therefore it makes sense to give them a few seconds
before we consider them available for real.

This is related to https://github.com/metal3-io/metal3-dev-env/issues/981 and https://github.com/metal3-io/metal3-dev-env/pull/983
It will be easier to check the health of the controllers when the deployments have `minReadySeconds` set.

### DIFF
--- a/ironic-deployment/ironic/ironic.yaml
+++ b/ironic-deployment/ironic/ironic.yaml
@@ -4,6 +4,7 @@ metadata:
   name: ${NAMEPREFIX}-ironic
 spec:
   replicas: 1
+  minReadySeconds: 10
   strategy:
     # We cannot run Ironic with more than one replica at a time. The recreate
     # strategy makes sure that the old pod is gone before a new is started.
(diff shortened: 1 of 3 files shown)


kashifest (commented):
/lgtm

kashifest:
/lgtm cancel
@lentzi90 make sure you run  this https://github.com/metal3-io/baremetal-operator#important-notes

lentzi90:
Thanks, fixed!

ironic-deployment/ironic/ironic.yaml
```diff
@@ -17,6 +17,7 @@ spec:
         name: ${NAMEPREFIX}-ironic
     spec:
       hostNetwork: true
+      minReadySeconds: 10
```
zhouhao3, line 20: `minReadySeconds` should be under the topmost `spec` field.

lentzi90: Thanks, fixed!

zhouhao3 (commented):
LGTM

fmuyassarov (approved):

PR #1109, opened 2022-04-12
Allow image provider to override kernel and add kernel params

dtantsur:
Currently, we assume that the same kernel can be used for all images.
This is not the case in multi-arch scenarios. Add a new optional field
to pass a kernel from the provider.

Also add support for extra kernel parameters. These are required by
CoreOS to link to the correct rootfs.

Requires: https://review.opendev.org/c/openstack/ironic/+/837379

### DIFF
--- a/apis/metal3.io/v1alpha1/preprovisioningimage_types.go
+++ b/apis/metal3.io/v1alpha1/preprovisioningimage_types.go
@@ -65,6 +65,16 @@ type PreprovisioningImageStatus struct {
 	// imageUrl is the URL from which the built image can be downloaded.
 	ImageUrl string `json:"imageUrl,omitempty"`
 
+	// kernelUrl is the URL from which the kernel of the image can be downloaded.
+	// Only makes sense for initrd images.
+	// +optional
+	KernelUrl string `json:"kernelUrl,omitempty"`
+
+	// extraKernelParams is a string with extra parameters to pass to the
+	// kernel when booting the image over network. Only makes sense for initrd images.
+	// +optional
+	ExtraKernelParams string `json:"extraKernelParams,omitempty"`
+
 	// format is the type of image that is available at the download url:
 	// either iso or initrd.
 	// +optional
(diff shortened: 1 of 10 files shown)


pkg/provisioner/ironic/ironic.go
```diff
@@ -590,11 +591,19 @@ func setDeployImage(driverInfo map[string]interface{}, config ironicConfig, acce
 				return deployImageInfo
 			}
 		case metal3v1alpha1.ImageFormatInitRD:
-			if allowInitRD {
+			if hostImage.KernelURL != "" {
+				deployImageInfo[deployKernelKey] = hostImage.KernelURL
+			} else if config.deployKernelURL == "" {
+				return nil
```
zaneb, line 597: I don't think returning here is consistent with the previous behaviour when there is no kernel available. The code on ll. 610-614 should still have a chance to run.

zaneb (commented):
/approve

apis/metal3.io/v1alpha1/preprovisioningimage_types.go
```diff
@@ -65,6 +65,16 @@ type PreprovisioningImageStatus struct {
...
 
+	// kernelUrl is the URL from which the kernel of the image can be downloaded.
+	// Only makes sense for initrd images.
+	// +optional
+	KernelUrl string `json:"kernelUrl,omitempty"`
+
+	// extraKernelParams is a string with extra parameters to pass to the
+	// kernel when booting the image over network. Only makes sense for initrd images.
+	// +optional
+	ExtraKernelParams string `json:"extraKernelParams"`
```
zhouhao3, line 76: This should also be an omitempty field.

pkg/provisioner/ironic/ironic.go
dtantsur: I don't think there is a point in it. If we reach this line, we have a request for initrd, but the kernel is neither in the image, nor in the configuration. There is no way we can succeed.

Even worse, we can end up with providing a pre-configured ISO when the image requests `ImageFormatInitRD`.

zhouhao3: It would be better to output  log info before returning.

dtantsur:
/retitle Allow image provider to override kernel and add kernel params

dtantsur:
/hold cancel

Github won't let me to remove [WIP] for the title :(

PR #1120, opened 2022-05-05
[WIP] Bmh statusannotation validation

hroyrh:
Add validations referenced in issue #855 : 
- Status annotation added after Status subresource already exists
- Status annotation containing invalid data
- Hardware details annotation containing invalid data
- Reboot annotation with invalid mode value
- Inspect annotation with invalid value


lentzi90 (commented):
Does this handle the reboot annotation with key (`reboot.metal3.io/{key}`)? I didn't see any test for it, sorry if I missed it.
It would be nice to have a test or two with this also to ensure we don't break this functionality. For docs on how it works, please see [this](https://metal3io.netlify.app/bmo/reboot_annotation.html).

hroyrh:
@lentzi90 I've added rebootannotation with keys to the validations.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -184,3 +244,87 @@ func validateImageURL(imageURL string) error {
...
+
+	return nil
+}
+
+func validHwdDetailsAnnotation(hwdetannotaiton string, inspect string) error {
+	if hwdetannotaiton == "" {
+		return nil
+	}
+
+	if inspect != "disabled" {
```
hroyrh, line 269: @lentzi90 I've a question about the hardwaredetailsannotation, if you check here - https://metal3io.netlify.app/bmo/external_inspection.html , we can add this annotation in case the `HardwareData` is not present. But, since that object is not part of the BMH resource, it is difficult to make that check as part of baremetalhost_validations. 
For now, I've just added the inspection disabled check for now, but that might cause some issues down the line, so I was wondering what would be the reasonable course here - add dependents objects to baremetalhost_validations or move this validation to another file.

lentzi90: Good question! I'm not sure what makes sense for this. @furkatgofurov7 @kashifest any ideas how we should handle this?

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -71,6 +77,11 @@ func (host *BareMetalHost) validateChanges(old *BareMetalHost) []error {
 		errs = append(errs, fmt.Errorf("bootMACAddress can not be changed once it is set"))
 	}
 
+	if old.Status != (BareMetalHostStatus{}) && host.Annotations[StatusAnnotation] != "" {
+		//if old.Status.OperationalStatus != "" && host.Annotations.StatusAnnotation != "" {
```
zaneb, line 81: Can delete this commented-out line now, presumably.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -71,6 +77,11 @@ func (host *BareMetalHost) validateChanges(old *BareMetalHost) []error {
 		errs = append(errs, fmt.Errorf("bootMACAddress can not be changed once it is set"))
 	}
 
+	if old.Status != (BareMetalHostStatus{}) && host.Annotations[StatusAnnotation] != "" {
```
zaneb, line 80: question: should this be something like

```
if old.Status != BareMetalHostStatus{} && host.Annotations[StatusAnnotation] != "" && host.Annotations[StatusAnnotation] != old.Annotations[StatusAnnotation] {
```

i.e. could this break other writes to the CR in the period from when the operator copies the status to the Status until the status annotation is removed? (I don't actually know.)

apis/metal3.io/v1alpha1/baremetalhost_validation.go
kashifest: Separate validation check makes sense since it involves multiple objects. If we want to keep it here, we can derive which hwannotaion  belongs to which BMH through owner-reference chain.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
kashifest: @zaneb has a valid point, I am thinking if we can break this long if condition somehow.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
furkatgofurov7: > @lentzi90 I've a question about the hardwaredetailsannotation, if you check here - https://metal3io.netlify.app/bmo/external_inspection.html , we can add this annotation in case the `HardwareData` is not present. 

I think we are mixing things up here. The external inspection user guide is referring to  `HardwareDetails data in the Status` which do still exist in the BMH. 
HardwareData is completely separate CRD AFAIK which stores `status.hardware` information, essentially the same as in BMH status, but the latter is left in purpose for backwards compatibility and planned to be dropped when there is a new API for BMO

kashifest: Ah true 😀

apis/metal3.io/v1alpha1/baremetalhost_validation.go
hroyrh: @kashifest i have included another if inside to break it the condition into two.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
hroyrh: So, as I understand it, the current check if good enough then. Since according to - https://github.com/metal3-io/metal3-docs/blob/main/design/baremetal-operator/external-introspection.md , even if we have hardwaredetails inside the status-annotation, the hardwaredetails-annotation will take precedence.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -175,6 +198,55 @@ func validateDNSName(hostaddress string) error {
...
+
+		objStatus, err := unmarshalStatusAnnotation([]byte(statusannotation))
+		if err != nil {
+			return fmt.Errorf("invalid status annotation, failed to fetch Status from annotation")
+		}
+
+		deco := json.NewDecoder(strings.NewReader(statusannotation))
+		deco.DisallowUnknownFields()
+		if err = deco.Decode(&BareMetalHostStatus{}); err != nil {
+			return fmt.Errorf("invalid field in StatusAnnotation")
```
dtantsur, line 239: same

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -184,3 +256,87 @@ func validateImageURL(imageURL string) error {
 
 	return nil
 }
+
+func checkStatusAnnotation(bmhstatus *BareMetalHostStatus) error {
+
+	operatstatus_allowed := []string{"", "OK", "discovered", "error", "delayed", "detached"}
+	errortype_allowed := []string{"", "provisioned registration error", "registration error", "inspection error", "preparation error", "provisioning error", "power management error"}
```
dtantsur, line 263: Could we not hardcode the values here? It will be very trivial to forget to update this place.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -184,3 +256,87 @@ func validateImageURL(imageURL string) error {
...
+	}
+
+	if inspect != "disabled" {
+		return fmt.Errorf("inspection has to be disabled for HardwareDetailsAnnotation, check if {'inspect.metal3.io' : 'disabled'}")
+	}
+
+	objStatus := &HardwareDetails{}
+	err := json.Unmarshal([]byte(hwdetannotaiton), objStatus)
+	if err != nil {
+		return fmt.Errorf("failed to fetch Hardware Details from annotation")
```
dtantsur, line 288: same here and below re adding the actual error

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -184,3 +256,87 @@ func validateImageURL(imageURL string) error {
...
+		return fmt.Errorf("invalid field in Hardware Details Annotation")
+	}
+
+	return nil
+}
+
+func validInspectAnnotation(inspectannotation string) error {
+
+	if !stringInSlice([]string{"disabled", ""}, inspectannotation) {
+		return fmt.Errorf("invalid value for Inspect Annotation")
```
dtantsur, line 303: nit: list valid values?

Also I'm aware that some people got used to putting "enabled" here. I wonder if we should support that...

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -184,3 +256,87 @@ func validateImageURL(imageURL string) error {
...
+	return nil
+}
+
+func validRebootAnnotation(rebootannotation string) error {
+	if rebootannotation == "" {
+		return nil
+	}
+
+	objStatus := &RebootAnnotationArguments{}
+	err := json.Unmarshal([]byte(rebootannotation), objStatus)
```
dtantsur, line 315: Is it actually a JSON? I would expect it to be a normal string.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
hroyrh: I think we can do that, but then we should also update the documentation.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
hroyrh: You mean hardcode the values in the baremetalhost_types file ?

apis/metal3.io/v1alpha1/baremetalhost_validation.go
hroyrh: We have these 2 types - `reboot.metal3.io: ""` and `reboot.metal3.io/my-unique-key: "{'mode':'soft'}"`, and JSON takes care of both of these.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
hroyrh: @dtantsur should we use the go-validator to validate for values instead ?

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -71,9 +77,26 @@ func (host *BareMetalHost) validateChanges(old *BareMetalHost) []error {
...
 }
 
+func validateAddStatusAnnotation(old *BareMetalHost, new *BareMetalHost) error {
+
+	oldStatusNotEmpty := (old.Status != (BareMetalHostStatus{}))
+	newAnnotationNotEmpty := (new.Annotations[StatusAnnotation] != "")
+	newAnnotationIsDifferent := (new.Annotations[StatusAnnotation] != old.Annotations[StatusAnnotation])
+
+	if oldStatusNotEmpty && newAnnotationNotEmpty && newAnnotationIsDifferent {
+		return fmt.Errorf("cannot add new statusannotation when status-subresource is already present")
```
dtantsur, line 94: instead of literal "statusannotation" you probably meant to insert the value of `StatusAnnotation`.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -175,6 +198,55 @@ func validateDNSName(hostaddress string) error {
...
+
+	return errs
+}
+
+func validStatusAnnotation(statusannotation string) error {
+	if statusannotation != "" {
+
+		objStatus, err := unmarshalStatusAnnotation([]byte(statusannotation))
+		if err != nil {
+			return fmt.Errorf("invalid status annotation, failed to fetch Status from annotation")
```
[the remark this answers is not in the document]
dtantsur: Yes, you can do it by using %w specifier.

Btw, there is no point in using `Errorf` is you're not substituting any values

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -184,3 +256,87 @@ func validateImageURL(imageURL string) error {
...
 	return nil
 }
+
+func checkStatusAnnotation(bmhstatus *BareMetalHostStatus) error {
+
+	operatstatus_allowed := []string{"", "OK", "discovered", "error", "delayed", "detached"}
+	errortype_allowed := []string{"", "provisioned registration error", "registration error", "inspection error", "preparation error", "provisioning error", "power management error"}
+
+	if !stringInSlice(operatstatus_allowed, string(bmhstatus.OperationalStatus)) {
+		return fmt.Errorf("invalid OperationalStatus in StatusAnnotation")
```
dtantsur, line 266: Did you mean `errors.New` or did you plan to substitute any values?

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -184,3 +256,87 @@ func validateImageURL(imageURL string) error {
...
+	}
+
+	if !stringInSlice(errortype_allowed, string(bmhstatus.ErrorType)) {
+		return fmt.Errorf("invalid ErrorType in StatusAnnotation")
```
dtantsur, line 270: same

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -184,3 +256,87 @@ func validateImageURL(imageURL string) error {
...
+
+func validRebootAnnotation(rebootannotation string) error {
+	if rebootannotation == "" {
+		return nil
+	}
+
+	objStatus := &RebootAnnotationArguments{}
+	err := json.Unmarshal([]byte(rebootannotation), objStatus)
+	if err != nil {
+		return fmt.Errorf("failed to fetch Reboot from annotation")
```
dtantsur, line 317: Please add the actual error (e.g. with %w)

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -184,3 +256,87 @@ func validateImageURL(imageURL string) error {
...
+	}
+
+	if !stringInSlice([]string{"hard", "soft", ""}, string(objStatus.Mode)) {
+		return fmt.Errorf("invalid RebootMode in RebootAnnotation")
```
dtantsur, line 321: nit: add supported modes?

apis/metal3.io/v1alpha1/baremetalhost_validation.go
hroyrh: @dtantsur do you mean, we should return the values of `New.StatusAnnotation` and `Old.StatusAnnotation` and state that they are not equal ?

dtantsur:
@hroyrh at least the gofmt failure seems real

/test-centos-integration-e2e-main

apis/metal3.io/v1alpha1/baremetalhost_validation.go
zaneb: He means put `cannot add new "baremetalhost.metal3.io/status" annotation`

apis/metal3.io/v1alpha1/baremetalhost_validation.go
dtantsur: My problem is that if we add a new value, we'll probably forgot to update both this list and the list in the API types.

hroyrh: Right now, I see two ways to resolve this, either use go-validator (in which case we only need to update the baremetalhost_types.go file) or add comments at both places so that both sides are updated.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -184,3 +256,95 @@ func validateImageURL(imageURL string) error {
...
 	return nil
 }
+
+// When making changes to this function for operationalstatus and errortype,
+// also make the corresponding changes in the OperationalStatus and
+// ErrorType fields in the struct definition of BareMetalHostStatus in
+// the file baremetalhost_types.go
+func checkStatusAnnotation(bmhstatus *BareMetalHostStatus) error {
+
+	operatstatus_allowed := []string{"", "OK", "discovered", "error", "delayed", "detached"}
```
zaneb, line 266: Please use defined [constants](https://github.com/metal3-io/baremetal-operator/blob/main/apis/metal3.io/v1alpha1/baremetalhost_types.go#L108-L160).

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -184,3 +256,95 @@ func validateImageURL(imageURL string) error {
...
+
+func unmarshalStatusAnnotation(content []byte) (*BareMetalHostStatus, error) {
+	objStatus := &BareMetalHostStatus{}
+	if err := json.Unmarshal(content, objStatus); err != nil {
+		return nil, errors.Wrap(err, "Failed to fetch Status from annotation")
+	}
+	return objStatus, nil
+}
+
+func stringInSlice(strlist []string, str string) bool {
```
zaneb, line 343: We are using go 1.19 in go.mod, so in theory we could now use https://pkg.go.dev/golang.org/x/exp/slices#Contains for this.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
zaneb: Is there a way to validate the status annotation against the OpenAPI definition of the Status subresource? That would be better than having to duplicate that validation by hand (in a way that can potentially get out of sync). The tricky part will be where to get that OpenAPI definition from - presumably we'd have to retrieve the CRD.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
hroyrh: I have used the constants although had to typecast them to string, does it look okay now ?

zaneb: Yes, but there's no need to cast everything - the point of using slices.Contains is that it can take a slice of e.g. `OperationalStatus` and search for `bmhstatus.OperationalStatus` in it.

zaneb:
> @zaneb regarding the OpenAPI definition for Status subresource, do we have a standard CRD that we follow. I was checking in the baremetal operator under the Config directory, but couldn't find an explicit specification. Do we use the status part of this spec - [baremetalhosts.yaml](https://github.com/metal3-io/baremetal-operator/blob/main/config/crd/bases/metal3.io_baremetalhosts.yaml)

Yes, exactly.

hroyrh:
@zaneb I haven't found any library to validate OpenAPIv3 schema for golang yet, and implementing it for this case might need non-trivial amount of work. I can open a separate pull request for it, and we can move this one forward since it deals with couple of other validations too. What do you think ?

zaneb:
> I haven't found any library to validate OpenAPIv3 schema for golang yet

The k8s API server (and possibly kubectl also?) is validating it, so that code already exists in k8s.
The tricky part is getting the schema in that format.

I agree that we don't need to block on this and could look into it later.

dtantsur:
```
failed to add finalizer: Operation cannot be fulfilled on baremetalhosts.metal3.io \"node-0\": StorageError: invalid object, Code: 4, Key: /registry/metal3.io/baremetalhosts/metal3/node-0, ResourceVersion: 0, AdditionalErrorMsg: Precondition failed: UID in precondition: d5c2070f-1dd7-4ab4-a0fa-7f09723a447a, UID in object meta: 
```

Not sure if it's related.

dtantsur:
Maybe you need to rebase the change to get https://github.com/metal3-io/baremetal-operator/commit/822b9beeca4f4993b8e8e81602c79a0e814cf6b9?

hroyrh:
> ```
> failed to add finalizer: Operation cannot be fulfilled on baremetalhosts.metal3.io \"node-0\": StorageError: invalid object, Code: 4, Key: /registry/metal3.io/baremetalhosts/metal3/node-0, ResourceVersion: 0, AdditionalErrorMsg: Precondition failed: UID in precondition: d5c2070f-1dd7-4ab4-a0fa-7f09723a447a, UID in object meta: 
> ```
> 
> Not sure if it's related.

@dtantsur yeah this seems to be the issue blocking CI for this PR. I have already rebased it and included the above mentioned PR, but this issue it still there.
I think its not accepting the changes I made to baremetal_types.go file, but need to investigate more to be sure of what the root cause is.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -76,9 +82,26 @@ func (host *BareMetalHost) validateChanges(old *BareMetalHost) []error {
...
+	if err := validateAddStatusAnnotation(old, host); err != nil {
+		errs = append(errs, err)
+	}
+
 	return errs
 }
 
+func validateAddStatusAnnotation(old *BareMetalHost, new *BareMetalHost) error {
+
+	oldStatusNotEmpty := (old.Status != (BareMetalHostStatus{}))
```
zaneb, line 94: This is not the check we use in the controller: https://github.com/metal3-io/baremetal-operator/blob/main/controllers/metal3.io/baremetalhost_controller.go#L1676-L1678

hroyrh: Thanks Zane, I've pushed a commit using the check used in the controller, will now run the CI test and see how it goes.

hroyrh:
Due to the CI issues its been causing, I've removed the validation - `validateAddStatusAnnotation` and will create a separate PR for it.

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -47,6 +47,19 @@ const (
 	// annotation is present and status is empty, BMO will reconstruct BMH Status
 	// from the status annotation.
 	StatusAnnotation = "baremetalhost.metal3.io/status"
+
+	// RebootAnnotation is the annotation which tells the host which mode to use
+	// when rebooting - hard/soft
+	RebootAnnotationPrefix = "reboot.metal3.io"
```
dtantsur, line 53: Could you make sure that duplicate constants do not exist in the controller/provisioner code? E.g. rebootAnnotationPrefix exists in baremetalhost_controller.go. These new constants should be used instead.

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -722,12 +735,16 @@ type BareMetalHostStatus struct {
 	// Important: Run "make generate manifests" to regenerate code
 	// after modifying this file
 
-	// OperationalStatus holds the status of the host
+	// OperationalStatus holds the status of the host and its
+	// validated by the checkStatusAnnotation function in
+	// baremetalhost_validation.go
```
dtantsur, line 740: This will be a part of auto-generated docs. I don't think we should refer to files there.

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -722,12 +735,16 @@ type BareMetalHostStatus struct {
...
 	// +kubebuilder:validation:Enum="";OK;discovered;error;delayed;detached
 	OperationalStatus OperationalStatus `json:"operationalStatus"`
 
 	// ErrorType indicates the type of failure encountered when the
-	// OperationalStatus is OperationalStatusError
+	// OperationalStatus is OperationalStatusError and its validated
+	// by the checkStatusAnnotation function in
+	// baremetalhost_validation.go
```
dtantsur, line 747: Same

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -175,6 +181,55 @@ func validateDNSName(hostaddress string) error {
...
 
+func validateAnnotations(host *BareMetalHost) []error {
+	var errs []error
+	var err error
+
+	for annotation, value := range host.Annotations {
+
+		switch {
+		case annotation == StatusAnnotation:
+			err = validStatusAnnotation(value)
```
dtantsur, line 192: nit: all these new functions should probably start with "validate", not "valid"

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -175,6 +181,55 @@ func validateDNSName(hostaddress string) error {
...
+		}
+		if err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	return errs
+}
+
+func validStatusAnnotation(statusannotation string) error {
```
dtantsur, line 211: nit: statusAnnotation

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -201,3 +256,86 @@ func validateRootDeviceHints(rdh *RootDeviceHints) error {
...
 	return nil
 }
+
+// When making changes to this function for operationalstatus and errortype,
+// also make the corresponding changes in the OperationalStatus and
+// ErrorType fields in the struct definition of BareMetalHostStatus in
+// the file baremetalhost_types.go
+func checkStatusAnnotation(bmhstatus *BareMetalHostStatus) error {
+
+	operatstatus_allowed := []string{"", string(OperationalStatusOK), string(OperationalStatusDiscovered), string(OperationalStatusError), string(OperationalStatusDelayed), string(OperationalStatusDetached)}
```
dtantsur, line 266: nit: go uses camel case, so operationalStatusAllowed, errorTypeAllowed

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -201,3 +256,86 @@ func validateRootDeviceHints(rdh *RootDeviceHints) error {
...
+		return nil
+	}
+
+	objStatus := &RebootAnnotationArguments{}
+	err := json.Unmarshal([]byte(rebootannotation), objStatus)
+	if err != nil {
+		return fmt.Errorf("failed to fetch Reboot from annotation, %w", err)
+	}
+
+	supported_modes := []string{"hard", "soft", ""}
```
dtantsur, line 326: nit: supportedModes

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -175,6 +181,55 @@ func validateDNSName(hostaddress string) error {
...
+	if statusannotation != "" {
+
+		objStatus, err := unmarshalStatusAnnotation([]byte(statusannotation))
+		if err != nil {
+			return fmt.Errorf("invalid status annotation, error=%w", err)
+		}
+
+		deco := json.NewDecoder(strings.NewReader(statusannotation))
+		deco.DisallowUnknownFields()
+		if err = deco.Decode(&BareMetalHostStatus{}); err != nil {
```
dtantsur, line 221: I don't quite understand why you parse BareMetalHostStatus twice - in unmarshalStatusAnnotation first, then here.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -201,3 +256,86 @@ func validateRootDeviceHints(rdh *RootDeviceHints) error {
...
 }
+
+// When making changes to this function for operationalstatus and errortype,
+// also make the corresponding changes in the OperationalStatus and
+// ErrorType fields in the struct definition of BareMetalHostStatus in
+// the file baremetalhost_types.go
+func checkStatusAnnotation(bmhstatus *BareMetalHostStatus) error {
+
+	operatstatus_allowed := []string{"", string(OperationalStatusOK), string(OperationalStatusDiscovered), string(OperationalStatusError), string(OperationalStatusDelayed), string(OperationalStatusDetached)}
+	errortype_allowed := []string{"", string(ProvisionedRegistrationError), string(RegistrationError), string(InspectionError), string(PreparationError), string(ProvisioningError), string(PowerManagementError)}
```
dtantsur, line 267: I suggest you move these to where the constants are defined so that we don't forget to update them.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -201,3 +256,86 @@ func validateRootDeviceHints(rdh *RootDeviceHints) error {
...
+
+	objStatus := &HardwareDetails{}
+	err := json.Unmarshal([]byte(hwdetannotaiton), objStatus)
+	if err != nil {
+		return fmt.Errorf("failed to fetch Hardware Details from annotation, error=%w", err)
+	}
+
+	deco := json.NewDecoder(strings.NewReader(hwdetannotaiton))
+	deco.DisallowUnknownFields()
+	if err = deco.Decode(objStatus); err != nil {
```
dtantsur, line 297: Same here: you're decoding HardwareDetails twice.

dtantsur (changes requested):

apis/metal3.io/v1alpha1/baremetalhost_types.go
hroyrh: Removed the file reference.

apis/metal3.io/v1alpha1/baremetalhost_types.go
hroyrh: Done

apis/metal3.io/v1alpha1/baremetalhost_validation.go
hroyrh: fixed it

apis/metal3.io/v1alpha1/baremetalhost_validation.go
hroyrh: made changes where the came case was not followed

apis/metal3.io/v1alpha1/baremetalhost_validation.go
hroyrh: done

apis/metal3.io/v1alpha1/baremetalhost_validation.go
hroyrh: done

apis/metal3.io/v1alpha1/baremetalhost_validation.go
hroyrh: removed the duplicate decode part

apis/metal3.io/v1alpha1/baremetalhost_validation.go
hroyrh: moved these to the types file alongside the declaration of its element type

apis/metal3.io/v1alpha1/baremetalhost_validation.go
hroyrh: fixed

apis/metal3.io/v1alpha1/baremetalhost_types.go
hroyrh: @dtantsur can we do this part of another PR ?

dtantsur: As you wish, just don't forget please.

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -722,12 +741,14 @@ type BareMetalHostStatus struct {
 	// Important: Run "make generate manifests" to regenerate code
 	// after modifying this file
 
-	// OperationalStatus holds the status of the host
+	// OperationalStatus holds the status of the host and its
+	// validated by the checkStatusAnnotation function
```
dtantsur, line 745: Same problem: the consumers of the API have no idea about functions in the Go code.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -55,6 +57,10 @@ func (host *BareMetalHost) validateHost() []error {
 		}
 	}
 
+	if annot_errors := validateAnnotations(host); annot_errors != nil {
```
dtantsur, line 60: annotationErrors

apis/metal3.io/v1alpha1/baremetalhost_validation_test.go
```diff
@@ -42,6 +42,97 @@ func TestValidateCreate(t *testing.T) {
 		Name:      "07564256-96ae-4315-ab03-8d34ece60fbb",
 		Namespace: "test-namespace",
 	}
+
+	om_sa_valid := metav1.ObjectMeta{
```
dtantsur, line 46: Please convert everything to camelCase, Go does not use snake_case.

apis/metal3.io/v1alpha1/baremetalhost_validation_test.go
```diff
@@ -511,6 +602,94 @@ func TestValidateCreate(t *testing.T) {
 			oldBMH:    nil,
 			wantedErr: "Image URL test1 is an invalid URL",
 		},
+		{
+			name: "validStatusAnnotation",
+			newBMH: &BareMetalHost{
+				TypeMeta: tm, ObjectMeta: om_sa_valid,
```
dtantsur, line 608: I wonder if it's useful to have these variables. This way, the test cases are separated from the actually tested configured. I'd just inline the content of ObjectMeta everywhere.

dtantsur:
/approve

The overall shape looks good. There are still some minor issues, please make sure to fix them here or in a follow-up PR.

apis/metal3.io/v1alpha1/baremetalhost_types.go
hroyrh: Done

apis/metal3.io/v1alpha1/baremetalhost_validation.go
hroyrh: Done

apis/metal3.io/v1alpha1/baremetalhost_validation_test.go
hroyrh: Moved the content inline, so removed these variables.

apis/metal3.io/v1alpha1/baremetalhost_validation_test.go
hroyrh: Thanks, that makes sense, I have moved the content inline.

PR #1121, opened 2022-05-05
test e2e

mboukhalfa:
Test running capm3 e2e tests from bmo


mboukhalfa:
/close
test failed because of infra however BMO has been used from tested_repo path cloned by project infra as expected and that's confirm that e2e test is working correctly from BMO repo

PR #1127, opened 2022-05-31
Get rid of a separate state for profile matching

dtantsur:
The logic executed in this state is really trivial, there is no reason
to have it. Move the actual profile matching to registering, which has
a bonus of reacting to changes.


controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -878,6 +878,16 @@ func (r *BareMetalHostReconciler) registerHost(prov provisioner.Provisioner, inf
 		return result
 	}
 
+	dirty, err = r.matchProfile(info)
+	if err != nil {
+		// FIXME(zaneb): This error requires a Spec change to fix, so we
+		// shouldn't treat it as transient
+		return actionError{errors.Wrap(err, "failed to match profile")}
```
zaneb, line 885: Now that we are doing this in a state where setting ProvisioningError makes some sense, can we fix this?

zaneb (commented):
We should be able to just reference the HardwareProfile in the Status directly here now: https://github.com/metal3-io/baremetal-operator/blob/main/controllers/metal3.io/baremetalhost_controller.go#L714-L716

zaneb:
The only issue I can see is that if an ExternallyProvisioned host has an invalid hw profile, then previously it wouldn't be a problem but now it will block registration. That seems unlikely though, and in any case it's the user's fault.
/approve

PR #1128, opened 2022-05-31
Skip the inspecting state when inspection is disabled

dtantsur:
Currently, a host moves to the inspecting state and requests
a preprovisioning image even when inspection is disabled.


docs/BaremetalHost_ProvisioningState.dot
```diff
@@ -12,6 +12,7 @@ digraph BaremetalHost {
     ExternallyProvisioned [label="Externally\nProvisioned"]
 
     Registering -> Inspecting [label="!externallyProvisioned && NeedsHardwareInspection()"]
+    Registering -> Preparing [label="!externallyProvisioned && !NeedsHardwareInspection()"]
```
zaneb, line 15: `!NeedsHardwareInspection()` isn't actually what we're checking here though, it's `inspectionDisabled()`.

controllers/metal3.io/host_state_machine.go
```diff
@@ -348,6 +348,8 @@ func (hsm *hostStateMachine) handleRegistering(info *reconcileInfo) actionResult
 	// if the credentials change and the Host must be re-registered.
 	if hsm.Host.Spec.ExternallyProvisioned {
 		hsm.NextState = metal3v1alpha1.StateExternallyProvisioned
+	} else if inspectionDisabled(hsm.Host) {
+		hsm.NextState = metal3v1alpha1.StatePreparing
```
zaneb, line 352: Need to also handle the transition from the ExternallyProvisioned state.

zaneb (commented):
It occurs to me only just now that to solve the problem of registration requiring a PreprovisioningImage during inspection, all we really need to do is return early from [this switch statement](https://github.com/metal3-io/baremetal-operator/blob/main/controllers/metal3.io/host_state_machine.go#L289-L307) in the inspecting state if inspection is disabled.

Skipping the state altogether shouldn't hurt though.

controllers/metal3.io/host_state_machine.go
dtantsur: I'm a bit blurry after the summit, could you elaborate? This is a branch for `!ExternallyProvisioned`

zaneb: There is another transition to Preparing on line 386, but I can't leave a comment there because GitHub, so I left it here.

zaneb:
#1137 implements my suggestion [above](https://github.com/metal3-io/baremetal-operator/pull/1128#pullrequestreview-998782929), which solves the problem without the need to change the state machine.

dtantsur:
> without the need to change the state machine.

I cannot agree it's a good thing. Starting inspection when it's disabled is confusing.

zaneb:
I found the original discussion where it was decided not to change the state machine: https://github.com/metal3-io/metal3-docs/pull/155#discussion_r552811030
It doesn't appear to me that there was any compelling technical reason that we chose to do it that way.

controllers/metal3.io/host_state_machine.go
```diff
@@ -377,10 +379,10 @@ func (hsm *hostStateMachine) handleExternallyProvisioned(info *reconcileInfo) ac
 		return hsm.Reconciler.actionManageSteadyState(hsm.Provisioner, info)
 	}
 
-	switch {
-	case hsm.Host.NeedsHardwareInspection():
+	// TODO(dtantsur): move this logic inside NeedsHardwareInspection?
+	if hsm.Host.NeedsHardwareInspection() && !inspectionDisabled(hsm.Host) {
```
zaneb, line 383: Moving it inside NeedsHardwareInspection would save updating the state machine diagram :)

PR #1130, opened 2022-06-02
Split node enrollment from provisioning configuration

dtantsur:
These actions are logically independent and can easily happen
    at different times (e.g. image options can only be set on an available
    node, preprovisioning images can be ready later, etc).
    
    This is a first part of the long-overdue refactoring of the provisioner
    interface and its Ironic implementation.


eranco74 (commented):
/lgtm

PR #1133, opened 2022-06-07
Fix grammatical errors in comments

hs0210:
Signed-off-by: Hu Shuai <hus.fnst@fujitsu.com>

### DIFF
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -1167,7 +1167,7 @@ func (p *ironicProvisioner) buildManualCleaningSteps(bmcAccess bmc.AccessDetails
 	var newSettings []map[string]interface{}
 	if data.ActualFirmwareSettings != nil {
 		// If we have the current settings from Ironic, update the settings to contain:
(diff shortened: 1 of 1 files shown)


pkg/provisioner/ironic/ironic.go
```diff
@@ -1167,7 +1167,7 @@ func (p *ironicProvisioner) buildManualCleaningSteps(bmcAccess bmc.AccessDetails
 	var newSettings []map[string]interface{}
 	if data.ActualFirmwareSettings != nil {
 		// If we have the current settings from Ironic, update the settings to contain:
-		// 1. settings converted by BMC drivers that are different than current settings
+		// 1. settings converted by BMC drivers that are different from current settings
```
bfournie, line 1170: "different than" is correct and certainly not an error. Although there is some discussion about which is the preferred phrasing, see for example https://www.merriam-webster.com/words-at-play/different-from-or-different-than, it hardly seems worth changing.

hs0210: Thank you for clarifying that. I will close this PR.

PR #1134, opened 2022-06-14
Do not consider empty RAID and root device hints updated

dtantsur:
Currently, due to the way we compare these fields in Spec and Status,
we always trigger an update at least once. Avoid it if the fields are
completely empty.

Extract the root device hints function and call it when matching the
profile to avoid a new update later on.

Improve logging around the diff handling.


Hellcatlk:
Thanks, lgtm.

furkatgofurov7:
Please label the PR appropriately based on https://github.com/metal3-io/baremetal-operator/blob/main/CONTRIBUTING.md#contributing-a-patch to make it easier to sort it out while releasing in the changelog, thanks!

PR #1135, opened 2022-06-16
Stop treating missing network as fatal error

derekhiggins:
Report that NetworkData isn't available and return a
blank string id the sectret is present but doesn't contain
the expected key. This is the same as the behaviour when the
secret isn't present at all.


zhouhao3:
lgtm

I think it's best to cover this code in UT.

dtantsur:
/approve

What we're doing downstream is kinda weird.. but I guess this is the right thing to do.

PR #1137, opened 2022-06-21
Don't register in Inspecting if inspection disabled

zaneb:
When inspection is disabled, the host still passes through the
inspecting state. Previously, we would still try to register the host in
this state, which would mean that it would require a valid
PreprovisioningImage to continue. Since no image is needed if we are not
actually going to inspect, skip registration in this case.


dtantsur:
I think https://github.com/metal3-io/baremetal-operator/pull/1128 is actually the right thing to do and has less potential side effects.

zaneb:
/close
Closing in favour of #1128.

PR #1151, opened 2022-07-22
ironic: Always add ImageFormatInitRD to supported formats

chdxD1:
In short: There is no way to use the Preprovisioning API with format InitRD with an external controller.

Ironic should always support initrd format for PreProvisioning API.

Right now this depends on DEPLOY_KERNEL_URL being set, which requires DEPLOY_RAMDISK_URL to be set, which enables the internal PreprovisioningImageReconciler, which we do not want with an external reconciler.

### DIFF
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -564,12 +564,9 @@ func (p *ironicProvisioner) PreprovisioningImageFormats() ([]metal3v1alpha1.Imag
 		return nil, err
 	}
 
-	var formats []metal3v1alpha1.ImageFormat
+	formats := []metal3v1alpha1.ImageFormat{metal3v1alpha1.ImageFormatInitRD}
 	if accessDetails.SupportsISOPreprovisioningImage() {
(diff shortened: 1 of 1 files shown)


pkg/provisioner/ironic/validatemanagementaccess_test.go
```diff
@@ -836,7 +836,7 @@ func TestPreprovisioningImageFormats(t *testing.T) {
 			Name:              "enabled virtualmedia",
 			Address:           "redfish-virtualmedia://example.test",
 			PreprovImgEnabled: true,
-			Expected:          []metal3v1alpha1.ImageFormat{"iso", "initrd"},
+			Expected:          []metal3v1alpha1.ImageFormat{"initrd", "iso"},
```
dtantsur, line 839: This reordering means that initrd will have a priority over iso, which I think is not intended for virtual media.

chdxD1: Ah you're right, I haven't caught that one because the coupling is really loose between the controllers.

PR #1152, opened 2022-07-23
Fix ironic httpd readness & liveness Probe

shibaPuppy:
It will be restarted because it is a path not in liveness probe
The default ironic deploy without ssl requires an http call.

```
# https error log
[core:debug] [pid 26:tid 45] protocol.c(1447): [client 127.0.0.1:34436] AH00566: request failed: malformed request line
127.0.0.1 - - [23/Jul/2022:14:44:06 +0000] "\x16\x03\x01\x02" 400 226 "-" "-"
```


elfosardo:
/lgtm
not enough power

PR #1154, opened 2022-07-29
Handle maintenance more gracefully

dtantsur:
First, do not set the maintenance flag for nodes in manageable state,
it's not required.
    
Second, if the maintenance flag is on when we don't expect it and Ironic
does not report a fault, move the nodes off maintenance.

Finally, remove stale instanceUUID before deletion if it's present.


dtantsur:
/hold

I would like to do some additional testing before this gets merged

dtantsur:
/hold cancel

I suspect it may fix the CI problem we're seeing.

PR #1149, opened 2022-07-19
🌱 Uplift go modules: k8s.io/* to v0.24.2, CRT to v0.12.3, controller tools to v0.9.1 and golangci-lint to v1.47.1

furkatgofurov7:
- k8s.io/* to v0.24.2,
- controller runtime to v0.12.3
- controller tools to v0.9.1
- golangci-lint to v1.47.1

**`Note:`** As part of the uplifting Cluster API Provider Metla3 to use CAPI [v1.2.0](https://github.com/kubernetes-sigs/cluster-api/releases/tag/v1.2.0) release, changes are proposed in this PR to align BMO go modules with the CAPM3/CAPI.

/hold until similar CAPM3 go module uplift changes are in place

PR #1157, opened 2022-08-05
Bump golangci-lint to v1.46.2.

s3rj1k:
Linter bump is needed to support building BMO with golang 1.18, as older version of linter fails to lint when default version of golang compiler is 1.18

(v1.46.2 version is chosen as most stable version that works with 1.17 and 1.18)


fmuyassarov (commented):
I think this change is overlapping with https://github.com/metal3-io/baremetal-operator/pull/1149/files

s3rj1k:
> I think this change is overlapping with https://github.com/metal3-io/baremetal-operator/pull/1149/files

That change bumps linter to v1.47.1 and it's a bad idea, that version has some broken revive linters
Not sure if they are used in BMO.

Still current change can be merged right now and enables to build BMO with golang 1.18 and does not touch k8s libraries.

I would prefer to get this landed before that big change, and get a tag release before that big k8s library bump lands.

I have issues with vendoring newer k8s API that are present in https://github.com/metal3-io/baremetal-operator/pull/1149/files

furkatgofurov7:
> That change bumps linter to v1.47.1 and it's a bad idea, that version has some broken revive linters
> Not sure if they are used in BMO.

@s3rj1k I have not seen any broken linters so far in #1149, and also I am not sure why it can be a bad idea to use that version but rather stick to v1.46.X.

s3rj1k:
@furkatgofurov7 https://github.com/golangci/golangci-lint/pull/2999

I assume that this issue will be fixed only in [golangci-lint](https://github.com/golangci/golangci-lint) v1.48.1

furkatgofurov7:
> I assume that this issue will be fixed only in [golangci-lint](https://github.com/golangci/golangci-lint) v1.48.1

Ack, v1.48.0 was released some time ago, and we could wait for the v1.48.1 and bump it in this PR or 1149 also, please let me know which works for you best.

s3rj1k:
@furkatgofurov7 v1.48.1 would be preferable, thanks

furkatgofurov7:
> @furkatgofurov7 v1.48.1 would be preferable, thanks

sure, would you update this PR once it is out or want me to do it as part of #1149 ?

s3rj1k:
> sure, would you update this PR once it is out or want me to do it as part of #1149 ?

I'll update this one, I hope to have one tag release before K8s libs are bumped

s3rj1k:
@dtantsur Thanks

PR #1160, opened 2022-08-17
Remove spec matcher string operation in deviceHint

### LINKED ISSUE #1159: Open up the possibility of using all deviceHint operators in certain deviceHint types
The way device hint handling is implemented right now in https://github.com/metal3-io/baremetal-operator/blob/main/pkg/provisioner/ironic/devicehints/devicehints.go#L18-L44 means that there is no possibility to use any other Ironic device hint operator other than the ones that are hard coded in the aforementioned implementation.

IMO: the serialNumber and the 3 different WWN hints, the DeviceName and HCTL don't require the "s==" prefix at the BMO level because for string type device hints IPA checks the presence of valid operators and if there is no valid operator then IPA will prepend  the hint with "s==" as implemented here https://opendev.org/openstack/ironic-lib/src/commit/95ce746ad101b7af32684c29384237061565a4af/ironic_lib/utils.py#L279-L284 . 

I understand that the hard coded operators are more explicit but at the same time they artificially limit the functionality of Ironic/IPA when used in combination with BMO. I am currently facing a situation where I would need the "\<in\>" operator in the serialNumber device hint but I can't use it because BMO blocks me despite the fact that Ironic/IPA supports the feature.

Would it be okay to stop prepending  the hard coded operators fro the  serialNumber,WWN,  WWNWithExtension, WWNWithVendorExtension, HCTL and DeviceName hints?

smoshiur1237:
Fixes #1159 
This PR will remove the spec matcher string operation `s==`, which will open up the possibility of using all deviceHint operators in certain deviceHint types.


Rozzii (approved):

dtantsur:
I think it's confusing if we only support operators for some fields. Even if we keep the defaults for other fields, we need to allow non-default operators everywhere. A bit more code, but also less confusion.

dtantsur:
Oh, and do we have any docs to update?

Rozzii:
I agree with you @dtantsur Model, Vendor and MinSizeGbyte hits would also need default value handling.

As @smoshiur1237 and me  are in a bit of a hurry we would like to have this PR merged as soon as possible and then in a follow PR related to the same issue we would handle the default operator handling for the remaining 3 deviceHint.

Related to documentation question, I think in metal3-docs we can mention that the ironic device hint matcher operators have became available after both of the BMO changes went through.

zaneb:
I'm not a huge fan of this approach:
https://github.com/metal3-io/baremetal-operator/issues/1159#issuecomment-1219952710
I'd prefer to see a design proposal with a considered approach to the best API interface, and then explicitly passing instructions to ironic rather than leaking its implementation through the API.

kashifest:
BMO didnt have explicit operators prior to this https://github.com/metal3-io/baremetal-operator/pull/495/commits/c2513a64e1e78c6b9eacc1cad04d297032bb1cd7. Do you recall @zaneb what was the reason for using explicit operator? Was it what you described here ?

dtantsur:
I assume the argument was "explicit better than implicit", as well as actually changing some defaults (size >= instead of ==). @dhellmann may remember better.

zaneb:
> BMO didnt have explicit operators prior to this [c2513a6](https://github.com/metal3-io/baremetal-operator/commit/c2513a64e1e78c6b9eacc1cad04d297032bb1cd7).

That patch was part of the PR that introduced root device hints (#495), so BMO didn't have _anything_ prior to that.

> Do you recall @zaneb what was the reason for using explicit operator? Was it what you described here ?

Essentially yes. Here's the comment thread: https://github.com/metal3-io/baremetal-operator/pull/495#discussion_r433367167

dhellmann:
> I assume the argument was "explicit better than implicit", as well as actually changing some defaults (size >= instead of ==). @dhellmann may remember better.

Yes, the goal was to make it easy for the user to enter the data and for us to be explicit about the type of query that would be run without leaking Ironic's API through the metal3 API (as @zaneb says in his comment).

The change proposed here will be backwards-incompatible with existing data, so you'll need to bump the API version to v2something.

zaneb:
> The change proposed here will be backwards-incompatible with existing data, so you'll need to bump the API version to v2something.

IIUC it's (mostly) backwards-compatible because ironic infers the `==` operator if no operator is passed. The other points still stand though.

dtantsur:
It is backward compatible, yes, just less explicit.

I'd be curious how a better API could look like. Maybe we just need `serialIn`? But that would be inconsistent with the existing fields that default to `<in>`...

kashifest:
> That patch was part of the PR that introduced root device hints (#495), so BMO didn't have _anything_ prior to that.

True. Missed that. 

I agree with @zaneb then since that was the initial goal to not leak ironic implementation implicitly but rather we should have an API in BMO.


> I'd be curious how a better API could look like. Maybe we just need `serialIn`? But that would be inconsistent with the existing fields that default to `<in>`...

I am curious too.

kashifest:
Let's perhaps continue the discussion on the issue #1159 since this PR won't see the light.

Rozzii:
I think we can close this @smoshiur1237 as this PR won't be merged. I will open a PR if the proposal process will be a success.

smoshiur1237:
Closing the PR as we discuss in the thread, needs new proposal.

PR #1162, opened 2022-08-24
Set external_http_url on the node when BMC is IPv6

honza:
When the BMC address is IPv6, we need to set the `external_http_url`
value on the node so that the node can access the provisioning image.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -44,6 +44,10 @@ but overflows could happen in case of slow provisioners and / or higher number o
 concurrent reconciles. For such reasons, it is highly recommended to keep
 BMO_CONCURRENCY value lower than the requested PROVISIONING_LIMIT. Default is 20.
 
+`IRONIC_EXTERNAL_URL_V6` -- This is the URL where Ironic will find the image for
+nodes that use IPv6. In dual stack environments, this can be used to tell Ironic which IP
+version it should set on the BMC.
+
 Kustomization Configuration
 ---------------------------
 
--- a/pkg/provisioner/ironic/factory.go
+++ b/pkg/provisioner/ironic/factory.go
@@ -3,6 +3,7 @@ package ironic
 import (
 	"errors"
 	"fmt"
+	"net/url"
 	"os"
 	"strconv"
 	"strings"
@@ -151,6 +152,17 @@ func loadConfigFromEnv(havePreprovImgBuilder bool) (ironicConfig, error) {
 			return c, fmt.Errorf("Invalid value for variable LIVE_ISO_FORCE_PERSISTENT_BOOT_DEVICE, must be one of Default, Always or Never")
 		}
 		c.liveISOForcePersistentBootDevice = forcePersistentBootDevice
+	}
+
+	c.externalURL = os.Getenv("IRONIC_EXTERNAL_URL_V6")
+
+	// Let's see if externalURL looks like a URL
+	if c.externalURL != "" {
+		_, externalURLParseErr := url.Parse(c.externalURL)
+
+		if externalURLParseErr != nil {
+			return c, externalURLParseErr
+		}
 	}
 
 	return c, nil
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -2,6 +2,7 @@ package ironic
 
 import (
 	"fmt"
+	"net"
 	"strings"
 	"time"
 
@@ -75,6 +76,7 @@ type ironicConfig struct {
 	deployISOURL                     string
 	liveISOForcePersistentBootDevice string
 	maxBusyHosts                     int
+	externalURL                      string
 }
 
 // Provisioner implements the provisioning.Provisioner interface
@@ -352,6 +354,7 @@ func (p *ironicProvisioner) ValidateManagementAccess(data provisioner.Management
 	}
 
 	driverInfo := bmcAccess.DriverInfo(p.bmcCreds)
+	driverInfo = setExternalURL(p, driverInfo)
 	deployImageInfo := setDeployImage(driverInfo, p.config, bmcAccess, data.PreprovisioningImage)
 
 	// If we have not found a node yet, we need to create one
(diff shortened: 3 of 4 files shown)


docs/configuration.md
```diff
@@ -44,6 +44,8 @@ but overflows could happen in case of slow provisioners and / or higher number o
 concurrent reconciles. For such reasons, it is highly recommended to keep
 BMO_CONCURRENCY value lower than the requested PROVISIONING_LIMIT. Default is 20.
 
+`IRONIC_EXTERNAL_URL` -- This is the URL where Ironic will find the image for the node.
+
```
kashifest, line 48: Trying to understand this, is it similar to virtual media? Can you please some more information on the use case where this would be suitable?

docs/configuration.md
```diff
@@ -44,6 +44,8 @@ but overflows could happen in case of slow provisioners and / or higher number o
 concurrent reconciles. For such reasons, it is highly recommended to keep
 BMO_CONCURRENCY value lower than the requested PROVISIONING_LIMIT. Default is 20.
 
+`IRONIC_EXTERNAL_URL` -- This is the URL where Ironic will find the image for the node.
```
dtantsur, line 47: Let's maybe call it `IRONIC_EXTERNAL_URL_V6` to make it clear it's only used for IPv6?

docs/configuration.md
dtantsur: So, some background (I agree it should be here in the docs): on a dual stack setup, Ironic has no way to understand which IP version to use for telling a BMC the address of its HTTP server. So the idea here is: default to v4 and use v6 when BMC itself is v6.

pkg/provisioner/ironic/ironic.go
```diff
@@ -580,6 +583,36 @@ func (p *ironicProvisioner) PreprovisioningImageFormats() ([]metal3v1alpha1.Imag
 	return formats, nil
 }
 
+func (p *ironicProvisioner) setExternalURL(driverInfo map[string]interface{}) map[string]interface{} {
```
dtantsur, line 586: Let's start with making sure `driverInfo` does NOT already contain `external_http_url`? Which, I guess, implies we need to set it to `nil` explicitly?

pkg/provisioner/ironic/ironic.go
```diff
@@ -580,6 +583,36 @@ func (p *ironicProvisioner) PreprovisioningImageFormats() ([]metal3v1alpha1.Imag
...
+	if p.config.externalURL == "" {
+		return driverInfo
+	}
+
+	ip := net.ParseIP(p.bmcAddress)
+	if ip == nil {
+		// Maybe it's a hostname?
+		ips, err := net.LookupIP(p.bmcAddress)
+		if err != nil {
```
dtantsur, line 595: This case is not exactly normal and probably deserves a log record.

pkg/provisioner/ironic/factory.go
```diff
@@ -153,6 +153,8 @@ func loadConfigFromEnv(havePreprovImgBuilder bool) (ironicConfig, error) {
 		c.liveISOForcePersistentBootDevice = forcePersistentBootDevice
 	}
 
+	c.externalURL = os.Getenv("IRONIC_EXTERNAL_URL")
```
dtantsur, line 156: Do we need any validation?

pkg/provisioner/ironic/ironic.go
```diff
@@ -580,6 +583,36 @@ func (p *ironicProvisioner) PreprovisioningImageFormats() ([]metal3v1alpha1.Imag
...
+
+		ip = ips[0]
+	}
+
+	// In the case of IPv4, we don't have to do anything.
+	if ip.To4() != nil {
+		return driverInfo
+	}
+
+	driverInfo["external_http_url"] = p.config.externalURL
```
dtantsur, line 611: I wonder if we can unit-test this case.

kashifest: +1

docs/configuration.md
honza: I updated the docs but it feels cumbersome. Suggestions for improvement are appreciated.

docs/configuration.md
```diff
@@ -44,6 +44,10 @@ but overflows could happen in case of slow provisioners and / or higher number o
 concurrent reconciles. For such reasons, it is highly recommended to keep
 BMO_CONCURRENCY value lower than the requested PROVISIONING_LIMIT. Default is 20.
 
+`IRONIC_EXTERNAL_URL_V6` -- This is the URL where Ironic will find the image for
+the node. In dual stack environments, this can be used to tell Ironic which IP
```
dtantsur, line 48: "nodes that use IPv6."

pkg/provisioner/ironic/validatemanagementaccess_test.go
```diff
@@ -1148,3 +1148,35 @@ func TestSetDeployImage(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("could not create provisioner: %s", err)
+	}
+
+	prov.config.externalURL = "XXX"
+
+	driverInfo := make(map[string]interface{}, 0)
+	updatedDriverInfo := setExternalURL(prov, driverInfo)
+
+	assert.Equal(t, "XXX", updatedDriverInfo["external_http_url"])
```
dtantsur, line 1181: Could you add a test on removing the value in case of v4 or no externalURL in the config?

dtantsur:
Left a couple of comments, but I'm fine with a follow-up.

/approve

docs/configuration.md
zaneb: It's pretty awkward that the v4 URL and the v6 URL are set through different mechanisms (the former is set in the ironic config iirc?).

dtantsur:
/test-centos-integration-main 

Thank you for the update, looks good. Leaving other folks a chance to take a look.

iurygregory:
@kashifest can you take another look at this PR? Thanks!

docs/configuration.md
iurygregory: @zaneb ironic doesn't have the concept of v4/v6 for the BMC, in case people configure the
`[deploy]/external_http_url and [deploy]/http_url` with v6 information this would be the default it would use, now we added the option to override the url at node level so the operator can specify what they want for each node. In our use case for dual-stack we want to use this for v6 only (so BMO would identify the IPversion of the BMC and set if there is the value from the config)

zaneb: So if the URLs are settable on a per-host basis, it sounds like one option would be to pass both the v4 and v6 URLs to the BMO and have it always set the right one for each host. (Obviously another would be to modify ironic to accept both URLs in the config and either select one automatically or allow the BMO to specify through the API which one to use.)

Having two completely different data flows for the v4 and v6 URLs in a dual-stack cluster is a serious maintenance burden for whoever has to deploy metal³. Sometimes such things are necessary in the short term, but ideally they should come with at least a plan for unwinding them.

Issue #1045, opened 2021-12-05
Question: is it possible to use common secret all over the bmh(s)?


ileixe:
Now, secret for BMC address/password could not be shared to multiple bmhs emitting ownership error below

```
[...]

PR #1073, opened 2022-01-20
Delete controller flag for baremetalhost for secret

ileixe:
Secret may be shared across the several baremetal hosts. Delete the
controller flag which have been resricted the share.

Fix #1045

PR #1164, opened 2022-09-07
Delete controller flag for baremetalhost for secret

dtantsur:
Secret may be shared across the several baremetal hosts. Delete the
controller flag which have been resricted the share.

Replaces https://github.com/metal3-io/baremetal-operator/pull/1073
Fix https://github.com/metal3-io/baremetal-operator/issues/1045

### DIFF
--- a/pkg/secretutils/secret_manager.go
+++ b/pkg/secretutils/secret_manager.go
@@ -65,7 +65,7 @@ func (sm *SecretManager) findSecret(key types.NamespacedName) (secret *corev1.Se
 // claimSecret ensures that the Secret has a label that will ensure it is
 // present in the cache (and that we can watch for changes), and optionally
 // that it has a particular owner reference.
-func (sm *SecretManager) claimSecret(secret *corev1.Secret, owner client.Object, ownerIsController, addFinalizer bool) error {
+func (sm *SecretManager) claimSecret(secret *corev1.Secret, owner client.Object, addFinalizer bool) error {
 	log := sm.log.WithValues("secret", secret.Name, "secretNamespace", secret.Namespace)
 	needsUpdate := false
 	if !metav1.HasLabel(secret.ObjectMeta, LabelEnvironmentName) {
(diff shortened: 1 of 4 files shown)


dtantsur:
I've only fixed a unit test, so approving

/approve
/test-centos-integration-main
/test-ubuntu-integration-main

kashifest:
I will give some time to others to have a chance to review. Feel free to unhold.

zaneb:
Copying [my comment from the previous PR](https://github.com/metal3-io/baremetal-operator/pull/1073#pullrequestreview-857557963):

> We might need some additional code around https://github.com/metal3-io/baremetal-operator/blob/main/pkg/secretutils/secret_manager.go#L97 to make sure that if the secret is already owned by us but with the controller flag set, we clear the flag. That said, it doesn't look like there's anything that will fail if we don't do this for existing secrets.

Basically, this works for new BMHs, but behaviour will be inconsistent when there are existing ones in the system.

dtantsur:
/approve cancel
/lgtm cancel
/unhold

Okay, will see what I can do.

dtantsur:
@zaneb wdyt https://github.com/metal3-io/baremetal-operator/pull/1164/commits/bd07ede0d62b3b31e2a621640bdd2976c9e8a93b

pkg/secretutils/secret_manager.go
```diff
@@ -74,34 +74,39 @@ func (sm *SecretManager) claimSecret(secret *corev1.Secret, owner client.Object,
...
-				if err := controllerutil.SetOwnerReference(owner, secret, sm.client.Scheme()); err != nil {
-					return errors.Wrap(err, "failed to set secret owner reference")
-				}
-				needsUpdate = true
+			updated = append(updated, ref)
+		}
+		if needsCleanup {
+			ownerLog.Info("removing the controller flag from the owner reference")
+			// Make sure to do this before SetOwnerReference below
+			secret.SetOwnerReferences(updated)
```
zaneb, line 102: It looks to me like `SetOwnerReference()` already does what we want (replaces the existing owner reference with a new one that does not set the controller flag, so it seems to me that we could just call it whenever `!alreadyOwned || needsCleanup` instead of having two different paths.

pkg/secretutils/secret_manager.go
```diff
@@ -74,34 +74,39 @@ func (sm *SecretManager) claimSecret(secret *corev1.Secret, owner client.Object,
...
-			}
-		} else {
-			alreadyOwned := false
-			ownerUID := owner.GetUID()
-			for _, ref := range secret.GetOwnerReferences() {
-				if ref.UID == ownerUID {
+			"ownerUID", ownerUID)
+
+		alreadyOwned := false
+		needsCleanup := false
```
zaneb, line 84: nit: would prefer a name like `hasControllerRef`

pkg/secretutils/secret_manager.go
```diff
@@ -78,30 +78,24 @@ func (sm *SecretManager) claimSecret(secret *corev1.Secret, owner client.Object,
...
+		alreadyOwned := false
+		ownerUID := owner.GetUID()
+		for _, ref := range secret.GetOwnerReferences() {
+			// We used to add controller references to BMC
+			// secrets. This was wrong, update.
+			if ref.UID == ownerUID && ref.Controller == nil {
+				alreadyOwned = true
+				break
+			} else if ref.Controller != nil && *ref.Controller {
+				ownerLog.Info("updating secret to no longer have an owner of type controller")
```
zaneb, line 90: teeechnically this is only true if `ref.UID == owner.UID`, but in practice it seems unlikely that anyone else will add an owner reference to the same Secret.

zaneb (commented):
/lgtm

PR #1173, opened 2022-10-12
Exclude hosts with virtual media from PROVISIONING_LIMIT

dtantsur:
The limit was designed to avoid glitches when too many hosts DHCP or
boot from PXE. Virtual media requires neither, and with a pre-built ISO
is actually quite efficient.

### DIFF
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -1919,6 +1919,16 @@ func (p *ironicProvisioner) IsReady() (result bool, err error) {
 }
 
 func (p *ironicProvisioner) HasCapacity() (result bool, err error) {
+	bmcAccess, err := p.bmcAccess()
+	if err != nil {
+		return false, err // shouldn't actually happen so late in the process
+	}
+
+	// If the current host uses virtual media, do not limit it. Virtual
+	// media deployments may work without DHCP and can share the same image.
+	if bmcAccess.SupportsISOPreprovisioningImage() {
+		return true, nil
+	}
 
 	hosts, err := p.loadBusyHosts()
 	if err != nil {
@@ -1938,7 +1948,7 @@ func (p *ironicProvisioner) loadBusyHosts() (hosts map[string]struct{}, err erro
 
 	hosts = make(map[string]struct{})
 	pager := nodes.List(p.client, nodes.ListOpts{
-		Fields: []string{"uuid,name,provision_state,driver_internal_info,target_provision_state"},
+		Fields: []string{"uuid,name,provision_state,boot_interface"},
 	})
 
 	page, err := pager.AllPages()
@@ -1958,7 +1968,11 @@ func (p *ironicProvisioner) loadBusyHosts() (hosts map[string]struct{}, err erro
 			nodes.Inspecting, nodes.InspectWait,
 			nodes.Deploying, nodes.DeployWait,
 			nodes.Deleting:
-			hosts[node.Name] = struct{}{}
+			// FIXME(dtantsur): this is a bit silly, but we don't have an easy way
+			// to reconstruct AccessDetails from a DriverInfo.
+			if !strings.Contains(node.BootInterface, "virtual-media") {
+				hosts[node.Name] = struct{}{}
+			}
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/provisioner/ironic/ironic.go
```diff
@@ -1938,7 +1948,7 @@ func (p *ironicProvisioner) loadBusyHosts() (hosts map[string]struct{}, err erro
 
 	hosts = make(map[string]struct{})
 	pager := nodes.List(p.client, nodes.ListOpts{
-		Fields: []string{"uuid,name,provision_state,driver_internal_info,target_provision_state"},
+		Fields: []string{"uuid,name,provision_state,boot_interface"},
```
andfasano, line 1951: I don't remember the reason why these fields were requested initially, but if they're not required anymore then it seems fine to do some cleanup

pkg/provisioner/ironic/ironic.go
```diff
@@ -1919,6 +1919,16 @@ func (p *ironicProvisioner) IsReady() (result bool, err error) {
...
+	bmcAccess, err := p.bmcAccess()
+	if err != nil {
+		return false, err // shouldn't actually happen so late in the process
+	}
+
+	// If the current host uses virtual media, do not limit it. Virtual
+	// media deployments may work without DHCP and can share the same image.
+	if bmcAccess.SupportsISOPreprovisioningImage() {
+		return true, nil
+	}
```
andfasano, line 1931: I understand that this could be an handy centralized place to perform this check, and it may sound as a reasonable location froma technical point of view. Anyhow I'd evaluate also the possibility to "push it up" in the stack - where the `ensureCapacity` is invoked, essentially in the `updateHostStateFrom` and `checkDelayedHost`. Not to save any comp cycle, but just to move such logic, which is strictly host dependent, out from the provisioner - so that the opt out decision could be made earlier

pkg/provisioner/ironic/ironic.go
```diff
@@ -1919,6 +1919,16 @@ func (p *ironicProvisioner) IsReady() (result bool, err error) {
...
 
 func (p *ironicProvisioner) HasCapacity() (result bool, err error) {
+	bmcAccess, err := p.bmcAccess()
+	if err != nil {
+		return false, err // shouldn't actually happen so late in the process
+	}
+
+	// If the current host uses virtual media, do not limit it. Virtual
+	// media deployments may work without DHCP and can share the same image.
+	if bmcAccess.SupportsISOPreprovisioningImage() {
```
andfasano, line 1929: General question: the delay mechanism gets applied also during deprovisioning, did you maybe also evaluate the impact of disabling it for vmedia hosts also during this phase?

pkg/provisioner/ironic/ironic.go
dtantsur: I think we wanted to use agent_url from driver_internal_info, but then dropped this idea.

pkg/provisioner/ironic/ironic.go
dtantsur: I'm not sure I'm following, this logic is also provisioner-dependent, it relies on knowing how Ironic works. We'd also need to move the filtering of the busy hosts up the stack somehow, it's going to be a significant change.

pkg/provisioner/ironic/ironic.go
dtantsur: It matters even less there: cleaning works completely locally.

pkg/provisioner/ironic/ironic.go
andfasano: I was thinking on a more general mechanism to avoid applying the delay check mechanism at all based on some predicate (in this case it'd be based on `bmcAccess.SupportsISOPreprovisioningImage()`, but in future it could be something else, maybe an host label) (btw, using a label wouldn't need to recompile necessary the code in case a new scenario will pop up).
I won't block on this, but I'd consider a more generic mechanism in future if other scenarios will arise

pkg/provisioner/ironic/ironic.go
andfasano: Ok. Just wondering if it could be worth moving this refactoring into a separate commit for the sake of clarity with a proper comment (another PR seems overkilling in this case)

PR #1176, opened 2022-10-20
:bug: Power off nodes upon deletion

### LINKED ISSUE #410: Host should be powered down once deleted

honza:
This is a continuation of #816 which in turn tries to fix #410. 

Co-authored-by: Sandhya Dasu <sadasu@redhat.com> @sadasu


furkatgofurov7:
Heads up: centOS failure is unrelated to these changes, we are facing issues with CI. 

xref: https://kubernetes.slack.com/archives/CHD49TLE7/p1666273231196429

pkg/provisioner/ironic/delete_test.go
```diff
@@ -225,3 +225,102 @@ func deleteTest(t *testing.T, detach bool) {
...
+			ironic: testserver.NewIronic(t).WithDefaultResponses().Node(
+				nodes.Node{
+					UUID:           nodeUUID,
+					ProvisionState: "active",
+					Maintenance:    true,
+					PowerState:     powerOn,
+				},
+			).Delete(nodeUUID),
+			priorErrors:                 maxPowerOffRetryCount + 1,
+			expectedPowerStateInRequest: "",
```
zhouhao3, line 282: There should be a problem with the value assigned to expectedPowerStateInRequest.
 According to the above code, the machine should be poweroff when priorityErrors > maxPowerOffRetryCount.

Cellebyte:
what is the progress here?

pkg/provisioner/ironic/delete_test.go
honza: Thanks!

pkg/provisioner/ironic/ironic.go
```diff
@@ -1777,7 +1782,7 @@ func (p *ironicProvisioner) Delete() (result provisioner.Result, err error) {
 // deletion operation is completed.
 func (p *ironicProvisioner) Detach() (result provisioner.Result, err error) {
 	// Currently the same behavior as Delete()
-	return p.Delete()
+	return p.Delete(true)
```
dtantsur, line 1785: Detaching must not result in a power off or any other potentially destructive actions.

dtantsur:
/lgtm cancel

One issue inline, otherwise looking good.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -494,7 +495,7 @@ func (r *BareMetalHostReconciler) actionDeleting(prov provisioner.Provisioner, i
 		return deleteComplete{}
 	}
 
-	provResult, err := prov.Delete()
+	provResult, err := prov.Delete(info.host.Status.ErrorCount > maxPowerOffRetryCount)
```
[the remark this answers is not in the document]
honza: Thanks for the detailed explanation!

pkg/provisioner/ironic/ironic.go
```diff
@@ -1707,7 +1707,7 @@ func (p *ironicProvisioner) Deprovision(force bool) (result provisioner.Result,
 // Delete removes the host from the provisioning system. It may be
 // called multiple times, and should return true for its dirty flag
 // until the deprovisioning operation is completed.
-func (p *ironicProvisioner) Delete() (result provisioner.Result, err error) {
+func (p *ironicProvisioner) Delete(force bool) (result provisioner.Result, err error) {
```
zaneb, line 1710: Given that the `force` parameter is not used anymore, we probably don't need to add it.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -494,10 +496,35 @@ func (r *BareMetalHostReconciler) actionDeleting(prov provisioner.Provisioner, i
...
+	}
+
 	provResult, err := prov.Delete()
 	if err != nil {
 		return actionError{errors.Wrap(err, "failed to delete")}
 	}
+
+	if provResult.ErrorMessage != "" {
+		return recordActionFailure(info, metal3v1alpha1.DeleteError, provResult.ErrorMessage)
+	}
```
zaneb, line 526: I'm not against adding this (and the DeleteError constant in general) for completeness (a hypothetical non-ironic provisioner could potentially fail here, even though the ironic one never does). But if we do I'd prefer we did it in a separate PR (or at least a separate commit) so it is clear to future-us that this is independent of the power-off-on-deletion thing.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -494,10 +496,35 @@ func (r *BareMetalHostReconciler) actionDeleting(prov provisioner.Provisioner, i
...
+		if err != nil {
+			return actionError{errors.Wrap(err, "failed to power off")}
+		}
+
+		if provResult.ErrorMessage != "" {
+			return recordActionFailure(info, metal3v1alpha1.PowerManagementError, provResult.ErrorMessage)
+		}
+
+		if provResult.Dirty {
+			return actionContinue{provResult.RequeueAfter}
```
zaneb, line 515: We need to do

```
result := actionContinue{provResult.RequeueAfter}
if clearError(info.host) {
    return actionUpdate{result}
}
return result
```

So that the PowerManagementError gets cleared and we don't end up setting `force` every time.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -494,10 +496,35 @@ func (r *BareMetalHostReconciler) actionDeleting(prov provisioner.Provisioner, i
...
+		}
```
zaneb, line 516: If a PowerManagementError is set we should clear it here and reset host.Status.ErrorCount to 0.
We'll also need to set a flag to indicate that at the next return we want to return `actionUpdate`, not just `actionContinue` (otherwise the changes won't get written to the CR).

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -494,10 +496,35 @@ func (r *BareMetalHostReconciler) actionDeleting(prov provisioner.Provisioner, i
...
 
+	// Only try to power off `maxPowerOffRetryCount` times before giving up, and deleting the host anyway
+	if info.host.Status.ErrorType != metal3v1alpha1.PowerManagementError || (info.host.Status.ErrorType == metal3v1alpha1.PowerManagementError && info.host.Status.ErrorCount < maxPowerOffRetryCount) {
+		info.log.Info("host ready to be powered off")
+		provResult, err := prov.PowerOff(
+			metal3v1alpha1.RebootModeHard,
+			info.host.Status.ErrorType == metal3v1alpha1.PowerManagementError)
+
+		if err != nil {
+			return actionError{errors.Wrap(err, "failed to power off")}
```
zaneb, line 507: I just noticed that we have some code in `handleDeprovisioning()` that ensures we don't get stuck in an error loop in the case where the host is not registered in Ironic and we don't actually have credentials to register it: https://github.com/metal3-io/baremetal-operator/blob/main/controllers/metal3.io/host_state_machine.go#L523-L528

We'll need similar code here, because failures are subject to the `maxPowerOffRetryCount` but errors will retry forever unless we short-circuit them.

An alternative would be to not attempt to power off if the `ErrorType` going in is a Deprovisioning error. I'm not certain that this would be the right thing though, because there could conceivably be cases where deprovisioning is going to fail (e.g. host is borked so it won't boot the IPA image) but we can still turn the power off (BMC still works).

zaneb (commented):
Refamiliarising myself with how all of this works :sweat_smile:

honza:
This might be overkill but the suggested changes led me to create a new step in the state machine.  When a delete is requested, instead of `StateDeleting`, we now enter `StatePoweringOffBeforeDelete`.  This separates the two operations nicely, and allows us to handle errors in the usual way (i.e. as was said above regarding skipping to delete when too many errors happen, or we don't have host creds). 

When we clear `ErrorCount`, we do it as part of the `actionComplete` result, and thus don't need an explicit update (as is idiomatic in the existing code).

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -207,6 +207,10 @@ const (
 	// learn about the hardware components available there
 	StateInspecting ProvisioningState = "inspecting"
 
+	// StateDeleting means we are in the process of cleaning up the host
```
dtantsur, line 210: The comment needs an update

controllers/metal3.io/host_state_machine.go
```diff
@@ -223,15 +224,15 @@ func (hsm *hostStateMachine) checkInitiateDelete(log logr.Logger) bool {
...
+		hsm.NextState = metal3v1alpha1.StatePoweringOffBeforeDelete
 	case metal3v1alpha1.StateProvisioning, metal3v1alpha1.StateProvisioned:
 		if hsm.Host.OperationalStatus() == metal3v1alpha1.OperationalStatusDetached {
 			if delayDeleteForDetachedHost(hsm.Host) {
 				log.Info("Delaying detached host deletion")
 				deleteDelayedForDetached.Inc()
 				return false
 			}
-			hsm.NextState = metal3v1alpha1.StateDeleting
+			hsm.NextState = metal3v1alpha1.StatePoweringOffBeforeDelete
```
dtantsur, line 235: We cannot power off detached hosts, maybe move straight to deleting?

honza: Changed to skip to delete for detached hosts

controllers/metal3.io/host_state_machine.go
```diff
@@ -561,6 +566,37 @@ func (hsm *hostStateMachine) handleDeprovisioning(info *reconcileInfo) actionRes
 	return actResult
 }
 
+func (hsm *hostStateMachine) handlePoweringOffBeforeDelete(info *reconcileInfo) actionResult {
+	actResult := hsm.Reconciler.actionPowerOffBeforeDeleting(hsm.Provisioner, info)
+	skipToDelete := func() actionResult {
```
dtantsur, line 571: nit: most of this function is repeating handleDeprovisioning. It would be great to refactor them.

PR #1179, opened 2022-11-06
:seedling: Add SECURITY_CONTACTS file

Rozzii:


SECURITY_CONTACTS
```diff
@@ -0,0 +1,11 @@
...
+
+Please do not disclose any security issue publicly e.g. Pull Requests, Issues, Comments.
+
+In this repository security reports are handled according to the
+Metal3-io project's security policy. For more information about the security
+policy consult the User-Guide [here](https://metal3io.netlify.app/security_policy.html).
+
+As this repository has a rolling release cycle and there are no release branches the security
+fix will be merged to the main branch and depending on the severity of the vulnerability the issue
+will either got it's own patch release or it will be released together with the next upcoming regular release.
```
tuminoid, line 11: too complex wording + a typo or two, so maybe something like:

will either be released in a patch release or in the next upcoming regular release.

SECURITY_CONTACTS
```diff
@@ -0,0 +1,11 @@
+# Reporting a security vulnerability
+
+Please do not disclose any security issue publicly e.g. Pull Requests, Issues, Comments.
+
+In this repository security reports are handled according to the
+Metal3-io project's security policy. For more information about the security
+policy consult the User-Guide [here](https://metal3io.netlify.app/security_policy.html).
+
+As this repository has a rolling release cycle and there are no release branches the security
```
tuminoid, line 9: add a comma between "branches" and "the"

SECURITY_CONTACTS
```diff
@@ -0,0 +1,11 @@
+# Reporting a security vulnerability
+
+Please do not disclose any security issue publicly e.g. Pull Requests, Issues, Comments.
```
tuminoid, line 3: should mention also not disclosing to project members or OWNERS directly?

tuminoid (commented):
leaving this review only to this PR, as your dozen other PR are identical to tihs PR

SECURITY_CONTACTS
Rozzii: Done.

SECURITY_CONTACTS
Rozzii: Done.

SECURITY_CONTACTS
Rozzii: Done.

SECURITY_CONTACTS
```diff
@@ -0,0 +1,12 @@
+# Reporting a security vulnerability
+
+Please do:
+  - not disclose any security issue publicly e.g. Pull Requests, Issues, Comments.
+  - not disclose any security issue to any owner of the repository or to any other contributor.
```
tuminoid, line 5: I'd like this line to have word "directly" in it, as now it sounds like we don't want to know at all

SECURITY_CONTACTS
```diff
@@ -0,0 +1,12 @@
...
+
+In this repository security reports are handled according to the
+Metal3-io project's security policy. For more information about the security
+policy consult the User-Guide [here](https://metal3io.netlify.app/security_policy.html).
+
+As this repository has a rolling release cycle and there are no release branches, the security
+fix will be merged to the main branch. The fix will either be released in a patch release or in the next upcoming regular release based on the severity of the issue.
```
tuminoid, line 12: Line wrap properly please.

SECURITY_CONTACTS
Rozzii: Dibe

SECURITY_CONTACTS
Rozzii: Done

SECURITY_CONTACTS
```diff
@@ -0,0 +1,15 @@
+# Reporting a security vulnerability
+
+Please do:
+  - not disclose any security issue publicly e.g. Pull Requests, Comments.
```
furkatgofurov7, line 4: ...security issue(s) publicly e.g. through GitHub Pull Requests, Issues and/or Comments.

SECURITY_CONTACTS
```diff
@@ -0,0 +1,15 @@
...
+  - not disclose any security issue directly to any owner of the repository or
+    to any other contributor.
+
+In this repository security reports are handled according to the
+Metal3-io project's security policy. For more information about the security
+policy consult the User-Guide [here](https://metal3io.netlify.app/security_policy.html).
```
furkatgofurov7, line 10:
```suggestion
In this repository, security reports are handled according to the
Metal3-io project's security policy. For more information about the security
policy consult the User Guide [here](https://metal3io.netlify.app/security_policy.html).
```

furkatgofurov7 (commented):
Thanks for the PR, few nits:

kashifest:
/lgtm
Lets fix the nits in a followup

SECURITY_CONTACTS
Rozzii: We are referring to it as User-guide or user-guider in many places https://github.com/metal3-io/metal3-docs/tree/main/docs bot in directory names and in text

SECURITY_CONTACTS
Rozzii: I understand that the sentence could be more explicit but I would like to advocate for not changing it as it does not add that much benefit IMO for the following reasons.

-in case of "security issue"  IMO if the statement is prohibiting in relation to a single entity then it is prohibiting in relation to multiples of the same entity in this context. So I don't think we need the (s) at the end .

- related to the second adjustment, I think if we ask the contributor to not disclose the issue publicly then we also prohibit the contributor on all platform. I don't think it is important to highlight that it is a GitHub PR or a GitLab PR as both of them are undesired.
 
I trust the contributors will understand what we mean by this statement, if they don't we can always change it in the future.

zaneb:
> I can understand that it is a third-party service and we could use a GitHub link that will work even if something happens with Netlify.

I guess a better question might be why we don't have a metal3.io subdomain (e.g. docs.metal3.io) as a CNAME record for metal3io.netlify.app. There are TLS certs to figure out I'm sure, but based on no research at all I feel confident that this must be a service that netlify would offer?

Anyhow, as Kashif said, let's get this merged and we can tweak it (across all of the repos) later.
/approve

PR #1180, opened 2022-11-07
🌱 build: introduce support for FROM overrides

jwstein3400:
This change moves the base container images into ARGs which can be overridden at build time.

N/A


jwstein3400:
@dtantsur When will the centos/ubuntu integration status be reported?

PR #1183, opened 2022-11-23
Run make fmt to fix issues

dtantsur:
No idea why the CI does not complain, but the local build fails with:

 pkg/provisioner/ironic/testserver/ironic.go:135: File is not `gofmt`-ed with `-s` (gofmt)


dtantsur:
/approve

Whitespace-only changes, self-approving to unblock everyone using Go 1.19

PR #1184, opened 2022-11-23
:bug: Always retry failed cleaning on deprovisioning (fixes #1182)

### LINKED ISSUE #1182: Lack of retries for cleaning may cause undesired side effects
I was playing with manually adding/removing k8s nodes to my cluster. I deprovisioned a BMH and removed the corresponding node. The BMH become `available`, but the node came back. Why?

What I deduced from the logs is this. Cleaning for the BMH has failed, the Ironic node got into `clean failed` state. BMO has moved it to `manageable` and left there since `manageable` in Ironic translates to `available` in BMO. Then, following `online: true`, the host was powered on and the still present k8s installation rejoined the cluster.

This behaviour is even [documented in the source code](https://github.com/metal3-io/baremetal-operator/blob/cc171d0d8ee21107255db5a5239bb29f2afb0477/pkg/provisioner/ironic/ironic.go#L1653-L1656), but I think it's wrong since it may:
1. as in this example, cause side effects when the machine powers back on into the old operating system;
2. cause sensitive data leaks if the host is decommissioned as it is;
3. confuse users with previous Ironic background since in Ironic a node cannot enter `available` without going through cleaning (unless disabled).

If cleaning is enabled for the host, it is retried until success.

/kind bug

dtantsur:
Not running cleaning immediately after a failure may:

1. cause side effects when the machine powers back on into the old
   operating system (and e.g. rejoins the cluster as a worker);
2. confuse users with previous Ironic background since in Ironic a node
   cannot enter available without going through cleaning (unless disabled).

This change moves hosts manageable -> available immediately. Users who want
to opt-out have to set automatedCleanMode to disabled or detach the host.

As of this change, we still give up cleaning after 3 attempts, so it's
still possible to end up with an unclean host causing issues.

### DIFF
--- a/controllers/metal3.io/host_state_machine.go
+++ b/controllers/metal3.io/host_state_machine.go
@@ -517,6 +517,8 @@ func (hsm *hostStateMachine) handleDeprovisioning(info *reconcileInfo) actionRes
 			// If the provisioner gives up deprovisioning and
 			// deletion has been requested, continue to delete.
 			if hsm.Host.Status.ErrorCount > 3 {
+				info.log.Info("Giving up on host clean up after 3 attempts. The host may still be operational " +
+					"and cause issues in your clusters. You should clean it up manually now.")
 				return skipToDelete()
 			}
 		case actionError:
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -1660,27 +1660,30 @@ func (p *ironicProvisioner) Deprovision(force bool) (result provisioner.Result, 
 		)
 
 	case nodes.CleanFail:
-		p.log.Info("cleaning failed")
+		if !force {
+			p.log.Info("cleaning failed", "lastError", ironicNode.LastError)
+			return operationFailed(fmt.Sprintf("Cleaning failed: %s", ironicNode.LastError))
+		}
+		p.log.Info("retrying cleaning")
 		if ironicNode.Maintenance {
-			p.log.Info("clearing maintenance flag")
+			p.log.Info("clearing maintenance flag", "maintenanceReason", ironicNode.MaintenanceReason)
 			return p.setMaintenanceFlag(ironicNode, false, "")
 		}
-		// This will return us to the manageable state without completing
-		// cleaning. Because cleaning happens in the process of moving from
-		// manageable to available, the node will still get cleaned before
-		// we provision it again.
+		// Move to manageable for retrying.
 		return p.changeNodeProvisionState(
 			ironicNode,
 			nodes.ProvisionStateOpts{Target: nodes.TargetManage},
 		)
 
 	case nodes.Manageable:
-		// We end up here after CleanFail. Because cleaning happens in the
-		// process of moving from manageable to available, the node will still
-		// get cleaned before we provision it again. Therefore, just declare
-		// deprovisioning complete.
-		p.log.Info("deprovisioning node is in manageable state")
-		return operationComplete()
+		// We end up here after CleanFail, retry cleaning. If a user
+		// wants to delete a host without cleaning, they can always set
+		// automatedCleaningMode: disabled.
+		p.log.Info("deprovisioning node is in manageable state", "automatedClean", ironicNode.AutomatedClean)
+		return p.changeNodeProvisionState(
+			ironicNode,
+			nodes.ProvisionStateOpts{Target: nodes.TargetProvide},
+		)
 
 	case nodes.Available:
 		p.publisher("DeprovisioningComplete", "Image deprovisioning completed")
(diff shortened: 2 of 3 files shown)


kashifest (commented):
lgtm. I really like this change.
/cc @zaneb @honza

pkg/provisioner/ironic/ironic.go
```diff
@@ -1645,27 +1645,26 @@ func (p *ironicProvisioner) Deprovision(force bool) (result provisioner.Result,
 		)
 
 	case nodes.CleanFail:
```
zaneb, line 1647: We need to treat this like the Error state and do:

```
if !force {
    return operationFailed(fmt.Sprintf("Cleaning failed: %s", ironicNode.LastError))
}
```

(Speaking of which, it looks like l. 1638 is the last remaining place where we're not using the semantic `operation*()` functions for return values.)

zaneb (commented):
The issue with this (which I should have mentioned here instead of [just on the issue](https://github.com/metal3-io/baremetal-operator/issues/1182#issuecomment-1329316249)) is that if there is some systematic failure, it will just sit in deprovisioning forever; there's no point where we ever report the cause of the problem to the user, nor can we increase the delay between retries.

dtantsur:
/hold

Needs careful thinking, will get back to it after the holidays

zaneb:
One reason this matters is that when we are trying to delete a host, we give up after 3 attempts at deprovisioning so that if the host is just gone, you won't be stuck with the BMH forever: https://github.com/metal3-io/baremetal-operator/blob/main/controllers/metal3.io/host_state_machine.go#L517-L521

I guess in the case where it is just gone we'll fail at `Error` rather than `CleanFailed`, so that will still work. But if we retry cleaning forever then in a situation where it can never work (e.g. BMC works but network card in the host is fried, or something) you will be stuck with a BMH that cannot be deleted.

dtantsur:
> I guess in the case where it is just gone we'll fail at Error rather than CleanFailed, so that will still work. But if we retry cleaning forever then in a situation where it can never work (e.g. BMC works but network card in the host is fried, or something) you will be stuck with a BMH that cannot be deleted.

You can delete such node by disabling cleaning or detaching it. I'm very much against giving up on cleaning if it's requested by a user.

dtantsur:
/hold cancel
/test-ubuntu-integration-main

This seems to work well in my testing, including stopping retries if automated cleaning mode is changed to `disabled`.

zaneb:
> You can delete such node by disabling cleaning or detaching it.

I could maybe buy that, but we're still not reporting the error so the user has no idea what is going on. It will just sit in 'deprovisioning' forever.

zaneb:
> > I could maybe buy that, but we're still not reporting the error so the user has no idea what is going on. It will just sit in 'deprovisioning' forever.
> 
> This applies to everything, no?

Not everything.

> It's the same for deployment and inspecting: we briefly show the error on the BMH, then retry.

For deployment I think that's true, because the state changes to deprovisioning and we don't keep track of how many times we go through that cycle. We should fix that.

For inspecting I think from memory we do the exponential backoff correctly.

> Why should cleaning be different?

Everything should be different.

> How else can we report an error?

In principle: return a failure that will record the reason in the status and bump the errorCount, thus increasing the delay until the next retry.

dtantsur:
> In principle: return a failure that will record the reason in the status and bump the errorCount, thus increasing the delay until the next retry.

That's what my patch does, at least as far as I understand (and see in my testing). Could you point out what I'm missing? Any explicit step to make the delay exponential?

dtantsur:
I tested by starting cleaning, then aborting it after the node enters `clean wait` using the Ironic CLI. This is the outcome:
```yaml
status:
  errorCount: 1
  errorMessage: 'Cleaning failed: By request, the clean operation was aborted'
  errorType: provisioning error
```

On the 2nd attempt, the error message is buggy, although Ironic reports the same failure:
```yaml
status:
  errorCount: 2
  errorMessage: 'Cleaning failed: '
  errorType: provisioning error
```

I have a feeling that more time has passed after the 2nd attempt than after the 1st one, but I'm not 100% sure.

dtantsur:
The empty error message is probably this Ironic issue: https://storyboard.openstack.org/#!/story/2010603. We'll fix it separately from this PR.

zaneb:
> That's what my patch does

Sorry, my bad, you are correct. I thought we were still discussing the previous version of the patch and I missed that you already fixed this.

controllers/metal3.io/host_state_machine.go
```diff
@@ -503,28 +503,18 @@ func (hsm *hostStateMachine) handleDeprovisioning(info *reconcileInfo) actionRes
...
-
 		switch r := actResult.(type) {
 		case actionComplete:
 			hsm.NextState = metal3v1alpha1.StateDeleting
 			hsm.Host.Status.ErrorCount = 0
-		case actionFailed:
-			// If the provisioner gives up deprovisioning and
-			// deletion has been requested, continue to delete.
-			if hsm.Host.Status.ErrorCount > 3 {
-				return skipToDelete()
```
zaneb, line 520: This feels like it might belong in a separate PR, so we can discuss (and if necessary some day revert) it independently.
The case where the underlying host just goes away is a fairly common one, and people justifiably complain about this making the BMH undeletable. So while I understand the argument, this is also an intentional feature we're removing.

dtantsur: Deleting unmanaged hosts is impossible for other reasons, not because of cleaning IIRC. There is a bug report for that from people with cleaning disabled.

Splitting this change.

zaneb (commented):
/lgtm

PR #1191, opened 2022-12-02
🐛 Fix endless requeue of HostFirmwareSettings

### LINKED ISSUE #1190: HostFirmwareSettings requeued indefinitely after deletion?
I tried out the "[test mode](https://github.com/metal3-io/baremetal-operator/blob/main/docs/dev-setup.md#running-without-ironic)" (`make run-test-mode`) recently for some simulations. After creating a lot of BMHs and then deleting them again, I noticed that BMO never "settled down" afterward. The logs just kept rolling by. I got curious and investigated what was going on and this is what I found:

BMO indefinitely re-queues reconciliation of HostFirmwareSettings after the BMH is deleted.

I'm not that familiar with the code though so I would like to gather some input on this before attempting to "fix" it.
Would it be safe to just *not* re-queue if the BMH doesn't exist? Or do we need to re-queue and the issue is rather why reconsiliation is triggerd for the deleted object to begin with?

1. Setup a cluster
2. Install CRDs and cert-manager
3. run BMO in test mode (`make run-test-mode`)
4. Add lots of BMHs
5. Delete all BMHs
6. Observe logs continuing to spit out the following with references to long gone resources:
   ```
   could not get baremetalhost, not running reconciler
   ```

Reconciliation of long gone resources should stop and not be re-queued.

The relevant code:
https://github.com/metal3-io/baremetal-operator/blob/03f8a05111547afb8e81036540001849a88d6534/controllers/metal3.io/hostfirmwaresettings_controller.go#L121-L124

**Environment:**

- Baremetal Operator version: main
- Environment (metal3-dev-env or other): fixture provisioner (not sure if relevant)

/kind bug

lentzi90:

The reconciler was endlessly re-queuing HostFirmwareSettings after the BMH was deleted. For BMH we treat missing resource as a special case that we don't re-queue to avoid this endless loop. This PR makes it the same for HostFirmwareSettings.

Fixes #1190

### DIFF
--- a/controllers/metal3.io/hostfirmwaresettings_controller.go
+++ b/controllers/metal3.io/hostfirmwaresettings_controller.go
@@ -120,6 +120,9 @@ func (r *HostFirmwareSettingsReconciler) Reconcile(ctx context.Context, req ctrl
 	bmh := &metal3v1alpha1.BareMetalHost{}
 	if err = r.Get(context.TODO(), req.NamespacedName, bmh); err != nil {
 		reqLogger.Info("could not get baremetalhost, not running reconciler")
+		if k8serrors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
 		return ctrl.Result{Requeue: true, RequeueAfter: resourceNotAvailableRetryDelay}, nil
 	}
 


controllers/metal3.io/hostfirmwaresettings_controller.go
```diff
@@ -120,6 +120,9 @@ func (r *HostFirmwareSettingsReconciler) Reconcile(ctx context.Context, req ctrl
 	bmh := &metal3v1alpha1.BareMetalHost{}
 	if err = r.Get(context.TODO(), req.NamespacedName, bmh); err != nil {
 		reqLogger.Info("could not get baremetalhost, not running reconciler")
+		if k8serrors.IsNotFound(err) {
+			return ctrl.Result{}, nil
```
mboukhalfa, line 124: should not return error not found instead of nil ?

lentzi90: Not completely sure, but we don't set any error for the BMH controller either if it is missing: https://github.com/metal3-io/baremetal-operator/blob/main/controllers/metal3.io/baremetalhost_controller.go#L127

mboukhalfa: /lgtm

zaneb: returning an error would cause an immediate requeue, so this is correct.

PR #1196, opened 2022-12-16
🌱 Separate mariadb component

lentzi90:

Make it optional to include the mariadb container. Split it into its of component that can be selected if needed.

NOTE: Running with MariaDB requires TLS to be enabled.

Supporting all the different combinations is becoming very hard since they affect each other. For example, if you want TLS + mariadb, then maraidb need patches for TLS and other components also need to know about these certificates. I don't see the point in supporting all different combinations and would suggest that we only support secure and well tested ones, i.e. those that we test in CI.
Alternatively, if we want a more "dynamic" way to handle things, we could consider going with helm charts instead. IMO there is not much point in doing that unless we actually test the combinations in some way. Instead I think we should provide a good base kustomization that works and users can build upon if needed.

### DIFF
--- a/ironic-deployment/README.md
+++ b/ironic-deployment/README.md
@@ -4,6 +4,10 @@ This folder contains kustomizations for Ironic.
 They are mainly used through the [deploy.sh](../tools/deploy.sh) script, which takes care of generating the necessary config for basic-auth and TLS.
 
 - **base** - This is the kustomize base that we start from.
-- **components** - In here you will find re-usable kustomize components for running Ironic with TLS, basic-auth or keepalived. Note that the basic-auth component is missing the actual credentials. This is on purpose, to make sure that the user is setting the password. Similarly the TLS component needs to have the proper IP/SAN set for the certificates.
+- **components** - In here you will find re-usable kustomize components for running Ironic with TLS, basic-auth, keepalived or mariadb.
+   - **basic-auth** - Enable basic authentication. Note that the basic-auth component is missing the actual credentials. This is on purpose, to make sure that the user is setting the password.
+   - **tls** - Enable TLS. The TLS component needs to have the proper IP/SAN set for the certificates.
+   - **keepalived** - Add a keepalived container to the deployment. This is useful when using a VIP for exposing the Ironic endpoint, so that the IP can move with the pod.
+   - **mariadb** - Use MariaDB instead of SQLite. TLS required for this to work.
 - **default** - A minimal, fully working, Ironic kustomization including configmap and password. Use only for development! The DB password is hard coded in the repo and there is no TLS or basic-auth.
 - **overlays** - Here you will find ready made overlays that use the above mentioned components.
--- a/ironic-deployment/components/mariadb/mariadb_patch.yaml
+++ b/ironic-deployment/components/mariadb/mariadb_patch.yaml
@@ -1,0 +1,67 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: ironic
+spec:
+  template:
+    spec:
+      containers:
+        - name: mariadb
+          image: quay.io/metal3-io/mariadb
+          imagePullPolicy: Always
+          livenessProbe:
+           exec:
+             command: ["sh", "-c", "mysqladmin status -uironic -p$(printenv MARIADB_PASSWORD)"]
+           initialDelaySeconds: 30
+           periodSeconds: 30
+           timeoutSeconds: 10
+           successThreshold: 1
+           failureThreshold: 10
+          readinessProbe:
+           exec:
+             command: ["sh", "-c", "mysqladmin status -uironic -p$(printenv MARIADB_PASSWORD)"]
+           initialDelaySeconds: 30
+           periodSeconds: 30
+           timeoutSeconds: 10
+           successThreshold: 1
+           failureThreshold: 10
+          volumeMounts:
+            - mountPath: /shared
+              name: ironic-data-volume
+            - name: cert-mariadb
+              mountPath: "/certs/mariadb"
+              readOnly: true
+            - name: cert-mariadb-ca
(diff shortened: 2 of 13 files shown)


lentzi90:
/hold
Holding to make sure we have time to discuss this properly before merging.
There are 2 point I would like to discuss in particular:

1. This changes the default to deploy Ironic without MariaDB.
2. The MariaDB component requires TLS.

For 1, I think the intention was to change this at some point (see this [comment](https://github.com/metal3-io/baremetal-operator/blob/42be6bbf7af3991fc318626048c554e3510ca4c1/ironic-deployment/default/ironic_bmo_configmap.env#L12)), which is why I took the chance and did it now. I'm happy to change it back though if there are objections.

For 2, I'm not sure how important this is but just to make everyone aware. Before this, it was not possible to deploy Ironic without the MariaDB container at all (even when actually using SQLite as we do in metal3-dev-env). Now it is possible, but not without TLS. My point here is more that I'm not sure we should support deploying without TLS *at all*, rather than allowing to deploy MariaDB without TLS.

lentzi90:
Tests run from metal3-dev-env:
- [centos main integration without mariadb](https://jenkins.nordix.org/job/metal3_metal3_dev_env_main_integration_test_centos/40/)
- [ubuntu main integration without mariadb](https://jenkins.nordix.org/job/metal3_metal3_dev_env_main_integration_test_ubuntu/193/)

dtantsur:
> This changes the default to deploy Ironic without MariaDB.

That at least was my intention and that's what we're doing in OpenShift nowadays. I think defaulting to something small and lean, while keeping an option for scale-out, is the way to go.

> I'm not sure we should support deploying without TLS at all

If it's technically impossible to deploy without TLS now, I don't think we should sneak that in. We can wait until somebody complains, but honestly people should not deploy without TLS.

lentzi90:
So to reiterate, this PR makes it possible to deploy without the mariadb container and switches the default to run with sqlite.

This table show the possible combinations after this PR:

| | Sqlite | MariaDB |
|-|---------|-------------|
| TLS | :heavy_check_mark: | :heavy_check_mark: |
| no TLS | :heavy_check_mark: | :x: |

The reason why it is not possible to combine mariadb with no TLS, is that we would need two different mariadb components for it. TLS affects the mariadb manifests since we need to add certificates for mariadb when running with TLS. It also goes the other way, if we run with TLS and add mariadb we need to configure ironic to know about mariadb and its certificates. I didn't think it would be worth it to add a specific component for mariadb + no TLS, so I left it out.

Another thing worth mentioning is how it works currently (before this PR). The key thing to understand then is that the mariadb container is always present, even when ironic is configured to use sqlite. This means that it was possible to assume that we always have mariadb in the TLS component. I.e. TLS + sqlite really means TLS + mariadb + sqlite. The TLS component always creates certificates for mariadb and mounts them in all the necessary places. This is why we didn't have an issue with these combinations before. Now with this PR it surfaced since the TLS component would now try to add certificates for a non-existing container (mariadb).

ironic-deployment/README.md
```diff
@@ -4,6 +4,10 @@ This folder contains kustomizations for Ironic.
 They are mainly used through the [deploy.sh](../tools/deploy.sh) script, which takes care of generating the necessary config for basic-auth and TLS.
 
 - **base** - This is the kustomize base that we start from.
-- **components** - In here you will find re-usable kustomize components for running Ironic with TLS, basic-auth or keepalived. Note that the basic-auth component is missing the actual credentials. This is on purpose, to make sure that the user is setting the password. Similarly the TLS component needs to have the proper IP/SAN set for the certificates.
+- **components** - In here you will find re-usable kustomize components for running Ironic with TLS, basic-auth, keepalived or mariadb.
+   - **basic-auth** - Enable basic authentication. Note that the basic-auth component is missing the actual credentials. This is on purpose, to make sure that the user is setting the password.
+   - **tls** - Enable TLS. The TLS component needs to have the proper IP/SAN set for the certificates.
+   - **keepalived** - Add a keepalived container to the deployment. This is useful when using a VIP for exposing Ironic, so that the IP can move with the pod.
```
kashifest, line 10:
```suggestion
   - **keepalived** - Add a keepalived container to the deployment. This is useful when using a VIP for exposing Ironic endpoint, so that the IP can move with the pod.
```

kashifest (approved):
/hold
@dtantsur this looks like ready to go in (One small nit left)

ironic-deployment/components/mariadb/mariadb_patch.yaml
```diff
@@ -0,0 +1,67 @@
...
+          readinessProbe:
+           exec:
+             command: ["sh", "-c", "mysqladmin status -uironic -p$(printenv MARIADB_PASSWORD)"]
+           initialDelaySeconds: 30
+           periodSeconds: 30
+           timeoutSeconds: 10
+           successThreshold: 1
+           failureThreshold: 10
+          volumeMounts:
+            - mountPath: /shared
```
dtantsur, line 29: I wonder if this is actually needed...

dtantsur:
/test-centos-integration-main
/test-ubuntu-integration-main
/lgtm

Unhold when ready

lentzi90:
Other PRs merged, and CI looks happy :tada: 
/hold cancel

kashifest:
/lgtm
Since it was rebased and lgtm went away.

lentzi90:
Come on Jenkins, you can do it!
I'll just wait for a while and see if it sparks to life...
/shrug

kashifest:
> /test-centos-integration-main
> /test-ubuntu-integration-main

How ?

lentzi90:
> > /test-centos-integration-main
> > /test-ubuntu-integration-main
> 
> How ?

I opened a new PR with the exact same commit here: https://github.com/metal3-io/baremetal-operator/pull/1211
When I trigger the tests on that they start running here :rofl:

PR #1197, opened 2022-12-19
🐛 uplift x/net to 0.4.0

tuminoid:
x/net has GO-2022-1144 and GO-2022-0969 vulnerabilities, and x/text as dependency to x/net has GO-2022-1059 that are fixed by this uplift.

BMO does not use vulnerable symbols directly.


kashifest (approved):

lentzi90 (commented):
/lgtm

PR #1201, opened 2023-01-09
🐛 .github: secure github actions

tuminoid:
Set top-level permission to read, and allow write only where applicable.

Pin actions to a sha to prevent unwanted changes.


kashifest (approved):

lentzi90 (commented):
/lgtm

PR #1204, opened 2023-01-17
✨ add explicit securitycontexts to controller

tuminoid:
It is always good to not rely on the defaults, but be explicit. Set explicit, secure securityContext for the BMO controller manager deployment and containers.

CAPI has the same starting from upcoming v1.4.0 and cert-manager etc has them already.

Setting explicit securityContext has its downsides as well, for tilt. Tilt's live update cannot handle securityContext which sets the user as non-root, as it requires root to deploy the binaries on rebuild. To workaround this, strip function is added to Tiltfile to handle BMO securityContexts.

Also, reindent the manifest properly.

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -140,11 +140,27 @@ def enable_provider(name):
 
     # Apply the kustomized yaml for this provider
     yaml = str(kustomizesub(context + "/config"))
+    yaml = strip_sec_ctx(yaml)
     k8s_yaml(blob(yaml))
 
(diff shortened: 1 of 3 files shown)


lentzi90 (commented):
/lgtm
Just noting here that we don't collect the manifests of BMO, Ironic, etc. yet, so it is hard to verify that these changes were actually applied. Hopefully we will get around to fixing that soon...

kashifest (approved):

PR #1207, opened 2023-01-27
🌱 Separate updating images from ValidateManagementAccess

dtantsur:
Contrary to its name, ValidateManagementAccess also enrolls the node and
configures various deploy-related parameters. This change splits
cleaning, image, custom deploy and ramdisk settings to a new function
with a goal to eventually delineate them.

Fix tryUpdateNode to return the updated node.

PR #1209, opened 2023-01-31
[WIP] 🐛 restart agent if the PreprovisioningImage has changed

dtantsur:
Requires https://github.com/metal3-io/baremetal-operator/pull/1207


zaneb:
This will have the effect of rebooting all the hosts whenever the PreprovisioningImage controller gets restarted, which was the thing we were trying to avoid.
It would seem safer to rely on e.g. the Generation of the PreprovisioningImage.

dtantsur:
Okay, we actually randomize the URL. This feels wrong. Or to put it another way: if we think that on every restart the images change, why is it wrong to reboot the nodes?

zaneb:
> > This will have the effect of rebooting all the hosts whenever the PreprovisioningImage controller gets restarted
> 
> Why? Are image URLs randomized somehow?

Yes.
 
> > It would seem safer to rely on e.g. the Generation of the PreprovisioningImage.
> 
> Generations are changed when the spec is updated, not the status, right?

[Yes](https://github.com/openshift/image-customization-controller/blob/main/pkg/imagehandler/imagehandler.go#L87).

> I've considered this option, but then we need to increase $something when the status changes.

Why? Given the same Spec we should build an equivalent image, so it's when the Spec changes that we want to trigger a reboot. The conditions in the Status already indicate which Generation they apply to and we already [check](https://github.com/openshift/baremetal-operator/blob/master/controllers/metal3.io/baremetalhost_controller.go#L645) that the built image corresponds to the latest Generation of the Spec before we use it.

The tricky part would be if we had annotations that could affect the content of the image, since those don't update the Generation. We're not doing that in ICC but it's [allowed by the default controller implementation](https://github.com/metal3-io/baremetal-operator/blob/main/controllers/metal3.io/preprovisioningimage_controller.go#L185). Regrettably, the assisted folks wrote their own controller so it could be doing anything.

dtantsur:
> Given the same Spec we should build an equivalent image, so it's when the Spec changes that we want to trigger a reboot. 

That's very explicitly not the case for assisted installer integration, whose input also depends on InfraEnv objects that we have no visibility into. So yeah:

>  Regrettably, the assisted folks wrote their own controller so it could be doing anything.

PR #1211, opened 2023-02-01
🌱 Separate mariadb kustomize component

lentzi90:
This is the exact same as https://github.com/metal3-io/baremetal-operator/pull/1196 to try to make the tests run...

**What this PR does / why we need it:**

Make it optional to include the mariadb container. Split it into its of component that can be selected if needed.

NOTE: Running with MariaDB requires TLS to be enabled.

Supporting all the different combinations is becoming very hard since they affect each other. For example, if you want TLS + mariadb, then maraidb need patches for TLS and other components also need to know about these certificates. I don't see the point in supporting all different combinations and would suggest that we only support secure and well tested ones, i.e. those that we test in CI.
Alternatively, if we want a more "dynamic" way to handle things, we could consider going with helm charts instead. IMO there is not much point in doing that unless we actually test the combinations in some way. Instead I think we should provide a good base kustomization that works and users can build upon if needed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -164,7 +164,7 @@ set-manifest-image-ironic: $(KUSTOMIZE) manifests
 .PHONY: set-manifest-image-mariadb
 set-manifest-image-mariadb: $(KUSTOMIZE) manifests
 	$(info Updating container image for Mariadb to use ${MANIFEST_IMG}:${MANIFEST_TAG})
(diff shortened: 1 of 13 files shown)


furkatgofurov7:
Clarification: since this was using the same commit as in #1196 both are merged.

PR #1212, opened 2023-02-03
🌱  set maintenance reason when turning maintenance on

dtantsur:
Redo of https://github.com/metal3-io/baremetal-operator/pull/1088

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.19
 
 require (
 	github.com/go-logr/logr v1.2.3
-	github.com/gophercloud/gophercloud v0.22.0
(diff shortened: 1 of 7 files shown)


honza:
My only comment is that it's too bad that we're setting the reason to an empty string most of the time. Can we attach meaning to some of those?

Also, I wonder if it might be more legible if the empty reason was hidden behind a constant?

```go
p.setMaintenanceFlag(ironicNode, false, MaintenanceReason.None)
```

dtantsur:
> My only comment is that it's too bad that we're setting the reason to an empty string most of the time. Can we attach meaning to some of those?

We use empty reason when *unsetting* maintenance, when a reason have no sense.

PR #1213, opened 2023-02-08
:bug: Accept http_proxy/https_proxy from env when `make docker`

Hellcatlk:

Fix the error that occur when executing `make docker` on a host that accesses the Internet through a proxy.

None.


dtantsur:
/approve

But why not `no_proxy` as well?

PR #1226, opened 2023-02-21
 🐛 allow namespace to continue terminating when bmh is deprovisioning at the same time

pixelsoccupied:
When BMH is deleted and NS is also in terminating state around the same time (usually GitOps/ArgoCD is involved), BMO may not be able to creating additional CRs as part of the deprovisioning steps. This condition essentially hangs the cleanup of BMH, its Secret (used for BMC) while the NS stays at a terminating state.

As long as registerHost is handling errors during this time, the issue goes away.

/cc @dtantsur
/cc @rhjanders
/cc @zaneb 

Fixes #

### DIFF
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -757,10 +757,15 @@ func (r *BareMetalHostReconciler) registerHost(prov provisioner.Provisioner, inf
 		return actionError{err}
 	}
 	switch info.host.Status.Provisioning.State {
-	case metal3v1alpha1.StateRegistering, metal3v1alpha1.StateExternallyProvisioned:
+	case metal3v1alpha1.StateRegistering, metal3v1alpha1.StateExternallyProvisioned, metal3v1alpha1.StateDeleting:
 		// No need to create PreprovisioningImage if host is not yet registered
 		// or is externally provisioned
 		preprovImgFormats = nil
+	case metal3v1alpha1.StateDeprovisioning:
+		// PreprovisioningImage is not required for deprovisioning when cleaning is disabled
+		if info.host.Spec.AutomatedCleaningMode == metal3v1alpha1.CleaningModeDisabled {
+			preprovImgFormats = nil
+		}
 	}
 
 	preprovImg, err := r.getPreprovImage(info, preprovImgFormats)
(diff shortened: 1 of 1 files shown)


controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -832,9 +838,13 @@ func (r *BareMetalHostReconciler) registerHost(prov provisioner.Provisioner, inf
 
 	// Create the hostFirmwareSettings resource with same host name/namespace if it doesn't exist
 	if info.host.Name != "" {
-		if err = r.createHostFirmwareSettings(info); err != nil {
-			info.log.Info("failed creating hostfirmwaresettings")
-			return actionError{errors.Wrap(err, "failed creating hostFirmwareSettings")}
+		if info.host.Spec.AutomatedCleaningMode == metal3v1alpha1.CleaningModeDisabled && !info.host.DeletionTimestamp.IsZero() {
```
zaneb, line 841: I don't think we need to worry about the cleaning mode here. If the host is being deleted then there's no point creating a HostFirmwareSetting CR for it.

Ditto for HardwareData, so let's handle that at the same time.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -761,6 +761,12 @@ func (r *BareMetalHostReconciler) registerHost(prov provisioner.Provisioner, inf
 		// No need to create PreprovisioningImage if host is not yet registered
 		// or is externally provisioned
 		preprovImgFormats = nil
+	case metal3v1alpha1.StateDeprovisioning, metal3v1alpha1.StateDeleting:
```
zaneb, line 764: I suspect we should add Deleting to the list of states on l. 760 where we unconditionally avoid creating a PreprovisioningImage.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -761,6 +761,12 @@ func (r *BareMetalHostReconciler) registerHost(prov provisioner.Provisioner, inf
...
+		// This is an edge case when the deprovisioning happens because the namespace with BMHs is being deleted.
```
zaneb, line 765: This comment is not strictly accurate outside the context of this review. We will get here if the host is deprovisioning or deleting for _any_ reason.
I'd put something like "PreprovisioningImage is not required for deprovisioning when cleaning is disabled".

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -757,10 +757,15 @@ func (r *BareMetalHostReconciler) registerHost(prov provisioner.Provisioner, inf
...
 	}
 	switch info.host.Status.Provisioning.State {
-	case metal3v1alpha1.StateRegistering, metal3v1alpha1.StateExternallyProvisioned:
+	case metal3v1alpha1.StateRegistering, metal3v1alpha1.StateExternallyProvisioned, metal3v1alpha1.StateDeleting:
 		// No need to create PreprovisioningImage if host is not yet registered
 		// or is externally provisioned
 		preprovImgFormats = nil
+	case metal3v1alpha1.StateDeprovisioning:
+		// PreprovisioningImage is not required for deprovisioning when cleaning is disabled
+		if info.host.Spec.AutomatedCleaningMode == metal3v1alpha1.CleaningModeDisabled {
```
zaneb, line 766: I don't love that we're encoding an assumption about the implementation of Ironic (i.e. that it doesn't need a PreprovisioningImage when cleaning is disabled) into the controller (rather than the ironic provisioner), but the alternative of passing the state and cleaning mode to PreprovisioningImageFormats() doesn't make a lot of conceptual sense either, so I think this is probably the best hack for now.

zaneb (commented):
/lgtm

controllers/metal3.io/baremetalhost_controller.go
dtantsur: I'm afraid we already have way too many assumptions about Ironic baked everywhere, including the API.

PR #1228, opened 2023-02-23
🐛 Do not abandon cleaning after 3 failures

dtantsur:
This behavior is arbitrary, opaque and violates the assumption that
cleaning results in actual cleaning. It hides infrastructure problems
and may cause deleted Kubernetes nodes reappear in the cluster.
Users that don't care about nodes being cleaned should opt out of
cleaning instead of relying on implicit behavior.

The issue that hosts cannot be deleted in the unmanages state has to be
addressed separately since it also affects hosts with cleaning disabled.

Follow-up to #1184


mboukhalfa:
testing the new trigger format

PR #1230, opened 2023-03-07
:seedling: Delay deprovision/delete when multiple finalizers exist

carbonin:
Delaying moving the host to deprovisioning or deleting allows other components to do whatever logic they need to do before BMO decides if the host should be deprovisioned or deleted.

In one specific case a finalizer could be added to a detached host which will attach the host to allow it to be cleaned at the time it is deleted. This logic needs to happen before BMO proceeds with its own delete logic.

@dtantsur PTAL


carbonin:
@dtantsur is there anyone else I should ping to get the approved label?

honza:
This seems sane.

/approve

zaneb:
Mmmmm, this is not really the semantics of a finalizer. A finalizer means that the k8s API shouldn't remove a CR that already has a DeletionTimestamp. It's a bit weird an unexpected that it would block the controller from doing anything. Something like the pause annotation would seem more appropriate for that.

PR #1233, opened 2023-03-12
:bug: Revert "Delay deprovision/delete when multiple finalizers exist"

zaneb:
This reverts #1230.

Finalizers are a signal to the k8s API, not to controllers. Controllers should ignore finalizers other than their own. This change overloaded the semantics of the finalizer to communicate between controllers. Annotations are the way to do this.

### DIFF
--- a/controllers/metal3.io/host_state_machine.go
+++ b/controllers/metal3.io/host_state_machine.go
@@ -223,10 +223,6 @@ func (hsm *hostStateMachine) checkInitiateDelete() bool {
 	default:
 		hsm.NextState = metal3v1alpha1.StateDeleting
 	case metal3v1alpha1.StateProvisioning, metal3v1alpha1.StateProvisioned:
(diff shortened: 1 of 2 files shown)


dtantsur:
Frankly, I hate multiplying vague annotations just as much as I hate reusing finalizers for something slightly different.

zaneb:
Me too, and I'm not convinced that one is needed here either, but at least it would be using the k8s API in a semantically correct way and have no potential to break backwards compatibility.

filanov:
> Me too, and I'm not convinced that one is needed here either, but at least it would be using the k8s API in a semantically correct way and have no potential to break backwards compatibility.

What is the backwards compatibility that you are talking about?

zaneb:
> What is the backwards compatibility that you are talking about?

If something somewhere is adding a finalizer to BMH for some reason then this could break it, as it would mean deleted hosts would just stay in Provisioned unless the finalizer was removed.

PR #1235, opened 2023-03-13
✨  Enable concurrency in BMO controllers through cmdline flags

kashifest:


main.go
```diff
@@ -95,6 +96,10 @@ func main() {
 	var runInTestMode bool
 	var runInDemoMode bool
 	var webhookPort int
+	var baremetalhostConcurrency int
+	var preprovisioningimageConcurrency int
+	var hostfirmwaresettingsConcurrency int
+	var bmceventsubscriptionConcurrency int
```
dtantsur, line 102: These will default to 0, I believe?

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1682,9 +1682,15 @@ func (r *BareMetalHostReconciler) updateEventHandler(e event.UpdateEvent) bool {
...
 // SetupWithManager registers the reconciler to be run by the manager
-func (r *BareMetalHostReconciler) SetupWithManager(mgr ctrl.Manager, preprovImgEnable bool) error {
+func (r *BareMetalHostReconciler) SetupWithManager(mgr ctrl.Manager, preprovImgEnable bool, options controller.Options) error {
+
+	maxConcurrentReconciles := 0
+	if options.MaxConcurrentReconciles > 0 {
+		maxConcurrentReconciles = options.MaxConcurrentReconciles
+	} else {
+		maxConcurrentReconciles = runtime.NumCPU()
+	}
```
dtantsur, line 1692: Do we want to replicate this logic and the handling of BMO_CONCURRENCY below in other controllers?

kashifest: Not sure since the other controllers don’t have it currently, I didn’t add it. Wdyt ?

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1682,9 +1682,15 @@ func (r *BareMetalHostReconciler) updateEventHandler(e event.UpdateEvent) bool {
 }
 
 // SetupWithManager registers the reconciler to be run by the manager
-func (r *BareMetalHostReconciler) SetupWithManager(mgr ctrl.Manager, preprovImgEnable bool) error {
+func (r *BareMetalHostReconciler) SetupWithManager(mgr ctrl.Manager, preprovImgEnable bool, options controller.Options) error {
+
+	maxConcurrentReconciles := 0
+	if options.MaxConcurrentReconciles > 0 {
```
zaneb, line 1688: How can this ever be true?

zaneb (commented):
I'm very confused by this change, which doesn't appear to do anything? It sets an option explicitly to its default value?
Why not make the existing BMO_CONCURRENCY environment variable apply to all controllers?

kashifest:
> I'm very confused by this change, which doesn't appear to do anything? It sets an option explicitly to its default value? Why not make the existing BMO_CONCURRENCY environment variable apply to all controllers?

Yep sorry, I didnt notice my local branch was not pushed properly, so the flag vars were missing in the commit. The idea is to allow configuring concurrency through command line flags. I will also replicate BMO_CONCURRENCY env code to other controllers.

main.go
kashifest: I have changed the default value to 10

kashifest:
@zaneb @dtantsur I have chose to keep only  command line arguments and removed configuring concurrency through env var since it didn't make any sense to me and also value of 8 and 2 was also something which is not properly explained. Do you think it looks good this way or you would still want to keep the env var way of configuring the controller concurrency? Currently the PR is trying to follow similar approach as CAPI/ CAPM3 controllers. 

/hold 
Until a common understanding is reached

zaneb:
https://github.com/metal3-io/baremetal-operator/issues/905
https://github.com/metal3-io/baremetal-operator/pull/906

kashifest:
> #905 #906

I have seen #906 at leas previously. Are you of the opinion that you would still want to keep the code to make this configurable via env var? or the value 8 should be default instead of 10 as I am setting now ?

zaneb:
I'm not opposed to changing the numbers, but I'd expect the new ones to be at least as "properly explained" as the previous ones were in #905.

Scaling by the number of vcpus so that we don't use proportionally more resources on small boxes made sense to me, but maybe there's a reason that's mistaken? (Certainly this was somewhat informed by tuning a lot of Python deployments, where the GIL makes it impossible to effectively utilise more threads than there are vcpus.)

I also don't really care whether it is configured by an env var or CLI arg. If there's a standard way to do the latter, that would be ideal. However, I don't see any reason to break existing deployers who may be relying on the env var. (Also, having different concurrency values for each controller seem pointless to me, and can't ever be standardised.)

PR #1236, opened 2023-03-16
:sparkles: Delay delete of detached hosts based on annotation metadata

carbonin:

This adds optional metadata to the detached annotation which allows a user to delay the deletion of a host.

This is useful in a case where another controller wants the BMH detached to avoid extra operations on the host, but also wants to take advantage of deprovisioning. Such a controller would set the detached annotation deprovisioning action to "delay" and also set its own finalizer on the BMH. Then, when the BMH is marked for deletion BMO would wait until the detached annotation is removed giving the other controller time to take whatever actions are needed (setting cleaning, changing images, etc)

This replaces the previous logic which watched finalizers directly. Having a controller wait on other finalizers is not desirable as it is not backward compatible in that other controllers could be setting finalizers today and not expect this new logic. It could also lead to deadlock-like behavior if this pattern were repeated in multiple controllers watching the same resource.

cc @zaneb @dtantsur


carbonin:
Marking this as a draft for now while I test it.

Looking for some feedback on the "API" and the names.

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -660,6 +660,18 @@ type RebootAnnotationArguments struct {
...
+type DetachedDeprovisionAction string
+
+const (
+	DetachedDeprovisionActionDelay  = "delay"
+	DetachedDeprovisionActionDelete = "delete"
+)
+
+type DetachedAnnotationArguments struct {
+	// DeprovisionAction indicates the desired deprovition logic when the detached annotation is present
+	DeprovisionAction DetachedDeprovisionAction `json:"deprovisionAction,omitempty"`
```
zaneb, line 672: 'deprovisionAction' doesn't seem like the right term, because it doesn't control what happens when you deprovision, but rather what happens when you delete. In fact as long as this annotation is set, we never deprovision at all.

controllers/metal3.io/host_state_machine.go
```diff
@@ -223,11 +224,10 @@ func (hsm *hostStateMachine) checkInitiateDelete() bool {
 	default:
 		hsm.NextState = metal3v1alpha1.StateDeleting
 	case metal3v1alpha1.StateProvisioning, metal3v1alpha1.StateProvisioned:
-		if len(hsm.Host.Finalizers) > 1 {
-			// Allow other finalizers to run before changing state
-			return false
-		}
 		if hsm.Host.OperationalStatus() == metal3v1alpha1.OperationalStatusDetached {
+			if delayDeleteForDetachedHost(hsm.Host) {
+				return false
```
zaneb, line 229: We need a log here.
Prometheus counter wouldn't hurt either.

zaneb (commented):
Conceptually I like this because there are two different reasons you might want to detach: because you just don't want to manage the host any more, or because you want to suspend the power checks but still get control back later to deprovision.

I don't care if the revert of #1230 is in the same PR, but can we have it in a separate commit please.

controllers/metal3.io/host_state_machine.go
carbonin: Log is no problem. Are you thinking of a new counter or would an existing one apply to this situation?

zaneb: New counter

carbonin: Done

controllers/metal3.io/metrics.go
```diff
@@ -123,6 +123,11 @@ var provisionerNotReady = prometheus.NewCounter(prometheus.CounterOpts{
 	Help: "Number of times a host is not provision ready",
 })
 
+var deleteDelayedForDetached = prometheus.NewCounter(prometheus.CounterOpts{
+	Name: "metal3_delete_delated_for_detached_total",
```
zaneb, line 127: typo: s/delated/delayed/

zaneb (commented):
Looks good except for a couple of typos.

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -660,6 +660,18 @@ type RebootAnnotationArguments struct {
...
 
+type DetachedDeleteAction string
+
+const (
+	DetachedDeleteActionDelay  = "delay"
+	DetachedDeleteActionDelete = "delete"
+)
+
+type DetachedAnnotationArguments struct {
+	// DeprovisionAction indicates the desired deprovition logic when the detached annotation is present
```
dtantsur, line 671: And now it's my turn to spot a typo: s/Deprovision/Delete/

zaneb: I see your `s/Deprovision/Delete/` and raise you `s/deprovition/delete/`

carbonin: :man_facepalming:

PR #1238, opened 2023-03-23
⚠️ 🐛 Update mariadb_patch.yaml

CJCShadowsan:
Add securityContext to ensure compliance with other recent securityContext improvements

It fixes an issue with mariadb being deployed in a pod where securityContext is set elsewhere.

Without it, it'll throw errors for config - And if you try set the same ironic ID as the other deployments, it will fail to write mariadb logs and refuse to start.

This sets the id to the mysql user (27), allowing it to start up accordingly.
Fixes #

### DIFF
--- a/ironic-deployment/components/mariadb/mariadb_patch.yaml
+++ b/ironic-deployment/components/mariadb/mariadb_patch.yaml
@@ -25,6 +25,9 @@ spec:
            timeoutSeconds: 10
            successThreshold: 1
            failureThreshold: 10
+          securityContext:
+            runAsGroup: 27
+            runAsUser: 27
           volumeMounts:
             - mountPath: /shared
               name: ironic-data-volume


tuminoid:
Thanks for bringing this up @CJCShadowsan. There is a hole in our testing apparently, allowing mariadb to be forgotten as its not a default database option.

/ok-to-test

I'd prefer adding more full-fledged securityContext for mariadb as well so we get all containers covered properly. I'm also thinking we may have issue with the `/shared` mount as gid 27 from mariadb is not the same as `fsGroup` we use for the deployment.

tuminoid:
We are reverting BMO non-root in #1239. I will reintroduce it next week and include proper `mariadb` support as well. We have additional issues related to unversioned `keepalived` images, which broke the release branches. We need to restore CI for those branches as priority.

I will close this PR as its not needed anymore, but I thank you @CJCShadowsan for raising this issue.

/close

PR #1242, opened 2023-03-31
⚠️ run BMO deployment as non-root (v2)

tuminoid:
BMO ironic has no reason to run as root. Make it run as "ironic" user.

dnsmasq requires elevated capabiities. k8s is missing the feature of ambient capabilities, so it requires us to setcap the binaries with expected capabilities and container must be running with "allowPrivilegeEscalation: true" in the manifest to allow elevation.

Read the ambient capabilities KEP for more details: https://github.com/kubernetes/enhancements/blob/master/keps/sig-security/2763-ambient-capabilities/README.md

Add securityContext to BMO deployment manifest and keepalived component, with correct UIDs and GIDs. This is important to be able to share files via /shared.

Modify keepalived image to run as ironic user, which we use the same UID and GID as the ironic-image.

This commit requires ironic-image with PR https://github.com/metal3-io/ironic-image/pull/410
as well as mariadb-image PR https://github.com/metal3-io/mariadb-image/pull/8
to be merged for this to work.

This v2 of the PR fixes issues identified after merging 1st PR:
- mariadb was missing securityContext and failed to run
- keepalived changes were not backwards compatible, and due using only single tag for all versions, new image broke all release branches. `NET_ADMIN` capability is not in the default capability set, and having keepalived binary have `net_admin` in its file capabilities prevents it from running if the container does not have this capability permitted.

### DIFF
--- a/ironic-deployment/base/ironic.yaml
+++ b/ironic-deployment/base/ironic.yaml
@@ -19,134 +19,190 @@ spec:
     spec:
       hostNetwork: true
       containers:
-        - name: ironic-dnsmasq
-          image: quay.io/metal3-io/ironic
-          imagePullPolicy: Always
-          securityContext:
-             capabilities:
-               add: ["NET_ADMIN", "NET_RAW"]
-          command:
-            - /bin/rundnsmasq
-          livenessProbe:
-           exec:
-             command: ["sh", "-c", "ss -lun | grep :67 && ss -lun | grep :69"]
-           initialDelaySeconds: 30
-           periodSeconds: 30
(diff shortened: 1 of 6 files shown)


tuminoid:
/hold

Until we have mariadb-image PR in, and testing via dev-env completed.

tuminoid:
Mariadb PR should be good to merge, so it solves that. We still need to add support for old tags for mariadb/ironic in CI/dev-env before merging this, and then sync the release branches as necessary, and maybe then we can re-merge this.

PR #1245, opened 2023-04-05
:seedling: redact the password info of image URL in logs

PlotCondor:
When ironic downloads images from an external server that requires basic authentication, the user and password would be added to URL. The password info shall be hidden in the logs.

**What this PR does / why we need it**: baremetal operator log should not show any password info.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*: N/A
Fixes #

### DIFF
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -3,6 +3,8 @@ package ironic
 import (
 	"fmt"
 	"net"
+	"net/url"
+	"reflect"
 	"strings"
 	"time"
 
@@ -686,6 +688,14 @@ func setDeployImage(config ironicConfig, accessDetails bmc.AccessDetails, hostIm
 	return nil
 }
 
+func redactedProbableURL(probableURL string) string {
+	u, err := url.ParseRequestURI(probableURL)
+	if err != nil {
+		return probableURL
+	}
+	return u.Redacted()
+}
+
 func (p *ironicProvisioner) tryUpdateNode(ironicNode *nodes.Node, updater *nodeUpdater) (updatedNode *nodes.Node, success bool, result provisioner.Result, err error) {
 	if len(updater.Updates) == 0 {
 		updatedNode = ironicNode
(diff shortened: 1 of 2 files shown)


pkg/provisioner/ironic/ironic.go
```diff
@@ -686,6 +688,17 @@ func setDeployImage(config ironicConfig, accessDetails bmc.AccessDetails, hostIm
 	return nil
 }
 
+func redactedProbableURL(probableURL interface{}) interface{} {
+	if probableURL == nil || reflect.TypeOf(probableURL).Kind() != reflect.String {
```
dtantsur, line 692: Why do you need reflection and interface{} here? a URL can only be a string.

pkg/provisioner/ironic/update_opts.go
```diff
@@ -88,6 +88,7 @@ func sanitisedValue(data interface{}) interface{} {
 		if strings.Contains(k.String(), "password") {
 			safeDatumValue = reflect.ValueOf("<redacted>")
 		}
+		safeDatumValue = reflect.ValueOf(redactedProbableURL(safeDatumValue.Interface()))
```
dtantsur, line 91: ... I'd rather have the reflection concentrated in this function.

pkg/provisioner/ironic/ironic.go
PlotCondor: https://github.com/metal3-io/baremetal-operator/blob/58b41390fce5ab99896f5968f740dc442951b26c/pkg/provisioner/ironic/update_opts.go#L110 Here the input is an `interface{}`. I will change the input and return type to string and move the reflection logics to the function `sanitisedValue()`

pkg/provisioner/ironic/update_opts.go
PlotCondor: Will do.

PR #1246, opened 2023-04-05
🌱 Disable scheduled workflows from running in forks

kashifest:
Workflows use resources that are only available when run in the [metal3-io/baremetal-operator](https://github.com/metal3-io/baremetal-operator) repository, so workflow run would always fail when they run in a forked repository. They are of no value to fork repositories. They are triggered by the schedule event, so they cause regular annoying and confusing workflow failure notifications for every fork owner. Rather checking these workflows only in upstream repo is enough.

### DIFF
--- a/.github/workflows/kubesec.yml
+++ b/.github/workflows/kubesec.yml
@@ -10,6 +10,9 @@ permissions:
 
 jobs:
   setup:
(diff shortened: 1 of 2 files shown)


tuminoid (approved):
/lgtm

PR #1247, opened 2023-04-05
🌱  Enable concurrency in BMO controllers through cmdLine flags

kashifest:
This PR adds concurrency in controllers through command line flags

### DIFF
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -19,10 +19,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"os"
 	"reflect"
-	"runtime"
-	"strconv"
 	"strings"
 	"time"
 
@@ -1732,33 +1729,7 @@ func (r *BareMetalHostReconciler) updateEventHandler(e event.UpdateEvent) bool {
 }
 
 // SetupWithManager registers the reconciler to be run by the manager
-func (r *BareMetalHostReconciler) SetupWithManager(mgr ctrl.Manager, preprovImgEnable bool) error {
-
-	maxConcurrentReconciles := runtime.NumCPU()
-	if maxConcurrentReconciles > 8 {
-		maxConcurrentReconciles = 8
-	}
-	if maxConcurrentReconciles < 2 {
-		maxConcurrentReconciles = 2
-	}
-	if mcrEnv, ok := os.LookupEnv("BMO_CONCURRENCY"); ok {
-		mcr, err := strconv.Atoi(mcrEnv)
-		if err != nil {
-			return errors.Wrap(err, fmt.Sprintf("BMO_CONCURRENCY value: %s is invalid", mcrEnv))
-		}
-		if mcr > 0 {
-			ctrl.Log.Info(fmt.Sprintf("BMO_CONCURRENCY of %d is set via an environment variable", mcr))
-			maxConcurrentReconciles = mcr
-		} else {
-			ctrl.Log.Info(fmt.Sprintf("Invalid BMO_CONCURRENCY value. Operator Concurrency will be set to a default value of %d", maxConcurrentReconciles))
-		}
-	} else {
-		ctrl.Log.Info(fmt.Sprintf("Operator Concurrency will be set to a default value of %d", maxConcurrentReconciles))
-	}
-
-	opts := controller.Options{
-		MaxConcurrentReconciles: maxConcurrentReconciles,
-	}
+func (r *BareMetalHostReconciler) SetupWithManager(mgr ctrl.Manager, preprovImgEnable bool, maxConcurrentReconcile int) error {
 
 	controller := ctrl.NewControllerManagedBy(mgr).
 		For(&metal3api.BareMetalHost{}).
@@ -1766,7 +1737,7 @@ func (r *BareMetalHostReconciler) SetupWithManager(mgr ctrl.Manager, preprovImgE
 			predicate.Funcs{
 				UpdateFunc: r.updateEventHandler,
 			}).
-		WithOptions(opts).
+		WithOptions(controller.Options{MaxConcurrentReconciles: maxConcurrentReconcile}).
 		Owns(&corev1.Secret{}, builder.MatchEveryOwner)
 
 	if preprovImgEnable {
--- a/controllers/metal3.io/preprovisioningimage_controller.go
+++ b/controllers/metal3.io/preprovisioningimage_controller.go
@@ -31,6 +31,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
 
 	metal3 "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
 	"github.com/metal3-io/baremetal-operator/pkg/imageprovider"
@@ -351,9 +352,10 @@ func (r *PreprovisioningImageReconciler) CanStart() bool {
 	return false
 }
 
-func (r *PreprovisioningImageReconciler) SetupWithManager(mgr ctrl.Manager) error {
+func (r *PreprovisioningImageReconciler) SetupWithManager(mgr ctrl.Manager, maxConcurrentReconcile int) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&metal3.PreprovisioningImage{}).
+		WithOptions(controller.Options{MaxConcurrentReconciles: maxConcurrentReconcile}).
 		Owns(&corev1.Secret{}, builder.MatchEveryOwner).
 		Complete(r)
 }
--- a/main.go
+++ b/main.go
@@ -21,8 +21,10 @@ import (
 	"fmt"
 	"os"
 	"runtime"
+	"strconv"
 	"strings"
 
+	"github.com/pkg/errors"
 	"go.uber.org/zap/zapcore"
 	k8sruntime "k8s.io/apimachinery/pkg/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
@@ -123,6 +125,7 @@ func main() {
 	var webhookPort int
 	var restConfigQPS float64
 	var restConfigBurst int
+	var controllerConcurrency int
 
 	// From CAPI point of view, BMO should be able to watch all namespaces
 	// in case of a deployment that is not multi-tenant. If the deployment
@@ -164,6 +167,8 @@ func main() {
 			"If omitted, the default Go cipher suites will be used. \n"+
 			"Preferred values: "+strings.Join(tlsCipherPreferredValues, ", ")+". \n"+
 			"Insecure values: "+strings.Join(tlsCipherInsecureValues, ", ")+".")
+	flag.IntVar(&controllerConcurrency, "controller-concurrency", 0,
+		"Number of CRs of each type to process simultaneously")
 	flag.Parse()
 
 	logOpts := zap.Options{}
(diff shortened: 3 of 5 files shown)


kashifest:
Opening  a new PR since while fixing https://github.com/metal3-io/baremetal-operator/pull/1235 the branch got corrupted while fixing rebasing.

kashifest:
Comments on previous PR
https://github.com/metal3-io/baremetal-operator/pull/1235#issuecomment-1471122373
https://github.com/metal3-io/baremetal-operator/pull/1235#issuecomment-1471410139
https://github.com/metal3-io/baremetal-operator/pull/1235#issuecomment-1475487599

kashifest:
I'll put @zaneb's latest comment here in writing as well to keep the conversation going: 

"I'm not opposed to changing the numbers, but I'd expect the new ones to be at least as "properly explained" as the previous ones were in https://github.com/metal3-io/baremetal-operator/issues/905.

Scaling by the number of vcpus so that we don't use proportionally more resources on small boxes made sense to me, but maybe there's a reason that's mistaken? (Certainly this was somewhat informed by tuning a lot of Python deployments, where the GIL makes it impossible to effectively utilise more threads than there are vcpus.)

I also don't really care whether it is configured by an env var or CLI arg. If there's a standard way to do the latter, that would be ideal. However, I don't see any reason to break existing deployers who may be relying on the env var. (Also, having different concurrency values for each controller seem pointless to me, and can't ever be standardised.)"

kashifest:
@zaneb thanks for your review. IMO these numbers highly depend on the infrastructure the deployment would be running as well as resource consumption. Anyways, I dont want to do any simulation at this point since my goal is to only standardize this configuration through command line flags.  I can still keep the number to 8 as you have done a simulation to support it. I think configuring it through controller flags is more standard but since you think this would break existing deploys, I have kept it as it is but added a check if the value is provided in the command line flag or not. PTAL again if it looks ok to you.

kashifest:
Any thoughts on this @zaneb @honza ?

main.go
```diff
@@ -116,6 +121,14 @@ func main() {
...
 	flag.IntVar(&webhookPort, "webhook-port", 9443,
 		"Webhook Server port (set to 0 to disable)")
+	flag.IntVar(&baremetalhostConcurrency, "baremetalhost-concurrency", 10,
+		"Number of baremetalhosts to process simultaneously")
+	flag.IntVar(&preprovisioningimageConcurrency, "preprovisioningimage-concurrency", 10,
+		"Number of preprovisioningimages to process simultaneously")
+	flag.IntVar(&hostfirmwaresettingsConcurrency, "hostfirmwaresettings-concurrency", 10,
+		"Number of hostfirmwaresettings to process simultaneously")
+	flag.IntVar(&bmceventsubscriptionConcurrency, "bmceventsubscription-concurrency", 10,
+		"Number of bmceventsubscriptions to process simultaneously")
```
zaneb, line 131: Why would anybody want to control these independently?

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1707,30 +1707,35 @@ func (r *BareMetalHostReconciler) updateEventHandler(e event.UpdateEvent) bool {
 }
 
 // SetupWithManager registers the reconciler to be run by the manager
-func (r *BareMetalHostReconciler) SetupWithManager(mgr ctrl.Manager, preprovImgEnable bool) error {
+func (r *BareMetalHostReconciler) SetupWithManager(mgr ctrl.Manager, preprovImgEnable bool, options controller.Options) error {
```
zaneb, line 1710: It's weird that an Options structure is passed in, but all of the options except MaxConcurrentReconciles are ignored.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1707,30 +1707,35 @@ func (r *BareMetalHostReconciler) updateEventHandler(e event.UpdateEvent) bool {
...
+			maxConcurrentReconciles = 8
+		}
+		if maxConcurrentReconciles < 2 {
+			maxConcurrentReconciles = 2
 		}
-		if mcr > 0 {
-			ctrl.Log.Info(fmt.Sprintf("BMO_CONCURRENCY of %d is set via an environment variable", mcr))
-			maxConcurrentReconciles = mcr
+		if mcrEnv, ok := os.LookupEnv("BMO_CONCURRENCY"); ok {
+			mcr, err := strconv.Atoi(mcrEnv)
```
zaneb, line 1725: Why not move all of this out into a common function, so that each controller can use the same logic?

main.go
kashifest: Just adding a flag, not imposing that anybody should be controlling this independently.

controllers/metal3.io/baremetalhost_controller.go
kashifest: Good point, fixed.

controllers/metal3.io/baremetalhost_controller.go
kashifest: Fixed

kashifest:
/test-ubuntu-integration-main
@zaneb @honza can I have some review on this before it needs rebase again?

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1707,8 +1707,37 @@ func (r *BareMetalHostReconciler) updateEventHandler(e event.UpdateEvent) bool {
...
+	if options.MaxConcurrentReconciles > 0 {
+		ctrl.Log.Info(fmt.Sprintf("Controller Concurrency will be set to %d according to command line flag", options.MaxConcurrentReconciles))
+	} else {
+		mCR, err := getMaxConcurrentReconciles()
+		if err != nil {
+			return err
+		}
+		maxConcurrentReconciles = mCR
+	}
+	options.MaxConcurrentReconciles = maxConcurrentReconciles
```
zaneb, line 1722: This always sets it to 0 AFAICT.

Could we have single implementation instead of having to fix this in 5 different places? Like move everything inside getMaxConcurrentReconciles maybe.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1707,8 +1707,37 @@ func (r *BareMetalHostReconciler) updateEventHandler(e event.UpdateEvent) bool {
...
 
 // SetupWithManager registers the reconciler to be run by the manager
-func (r *BareMetalHostReconciler) SetupWithManager(mgr ctrl.Manager, preprovImgEnable bool) error {
+func (r *BareMetalHostReconciler) SetupWithManager(mgr ctrl.Manager, preprovImgEnable bool, options controller.Options) error {
 
+	maxConcurrentReconciles := 0
+	if options.MaxConcurrentReconciles > 0 {
+		ctrl.Log.Info(fmt.Sprintf("Controller Concurrency will be set to %d according to command line flag", options.MaxConcurrentReconciles))
+	} else {
+		mCR, err := getMaxConcurrentReconciles()
```
zaneb, line 1716: This is ~never called since the default value is 10, so we're not preserving the previous behaviour.

kashifest: Default is set to 0 now.

controllers/metal3.io/baremetalhost_controller.go
kashifest: not sure if I got your point

zaneb: Say the user uses the CLI flag to set `options.MaxConcurrentReconciles` to, say, 15. Then:

```
	maxConcurrentReconciles := 0
```
This will be important later.
```
	if options.MaxConcurrentReconciles > 0 {
		ctrl.Log.Info(fmt.Sprintf("Controller Concurrency will be set to %d according to command line flag", options.MaxConcurrentReconciles))
```
Prints a log message saying we will use 15.
```
	} else {
		mCR, err := getMaxConcurrentReconciles()
		if err != nil {
			return err
		}
		maxConcurrentReconciles = mCR
	}
```
The road not taken.
```
	options.MaxConcurrentReconciles = maxConcurrentReconciles
```
Unconditionally sets `options.MaxConcurrentReconciles` to 0 regardless of the value passed in, then passes the options struct to the manager.

kashifest: ow snap, yes I had the assignment line which I misplaced during refactoring, good catch , fixed now.

kashifest:
@zaneb lets decide the fate of this PR, I have rebased it multiple times now.

controllers/metal3.io/preprovisioningimage_controller.go
```diff
@@ -351,9 +353,22 @@ func (r *PreprovisioningImageReconciler) CanStart() bool {
...
+	if options.MaxConcurrentReconciles > 0 {
+		ctrl.Log.Info(fmt.Sprintf("Controller Concurrency will be set to %d according to command line flag", options.MaxConcurrentReconciles))
+	} else {
+		mCR, err := getMaxConcurrentReconciles()
+		if err != nil {
+			return err
+		}
+		maxConcurrentReconciles = mCR
+	}
+	options.MaxConcurrentReconciles = maxConcurrentReconciles
```
dtantsur, line 367: I think you need to place this line instead of line 365. Otherwise you're setting options.MaxConcurrentReconciles to 0 if options.MaxConcurrentReconciles > 0.

controllers/metal3.io/preprovisioningimage_controller.go
```diff
@@ -351,9 +353,22 @@ func (r *PreprovisioningImageReconciler) CanStart() bool {
 	return false
 }
 
-func (r *PreprovisioningImageReconciler) SetupWithManager(mgr ctrl.Manager) error {
+func (r *PreprovisioningImageReconciler) SetupWithManager(mgr ctrl.Manager, options controller.Options) error {
```
dtantsur, line 356: Since the only supported option is concurrency, you could just accept it as an integer.

controllers/metal3.io/preprovisioningimage_controller.go
dtantsur, line 367:
```suggestion
		options.MaxConcurrentReconciles = maxConcurrentReconciles
	}
```

controllers/metal3.io/baremetalhost_controller.go
zaneb: It's fixed here (albeit in a slightly convoluted way), but not in the other 3 copies.
This is why I suggested moving this to a single implementation.

main.go
zaneb: We _are_ imposing that they have to be controlled independently because there is no option to allow them all to be set together (other than the existing env var).
If there's no reason to control them independently (and I can't think of one) then we're just making it needlessly hard for the user (and ourselves) and moving even further away from any kind of standard way to do this in kubebuilder-generated controllers.

kashifest: I will remove this since its not what we want and keep one flag for BMH controller only but IMO, controlling the concurrency of different controllers with the flag value of one controller, isn't standard way either.

kashifest:
@dtantsur @zaneb PTAL again

kashifest:
Still waiting for the reviews @zaneb @dtantsur

kashifest:
ping @zaneb @dtantsur

main.go
```diff
@@ -164,6 +167,8 @@ func main() {
 			"If omitted, the default Go cipher suites will be used. \n"+
 			"Preferred values: "+strings.Join(tlsCipherPreferredValues, ", ")+". \n"+
 			"Insecure values: "+strings.Join(tlsCipherInsecureValues, ", ")+".")
+	flag.IntVar(&bmoConcurrency, "bmo-concurrency", 0,
```
zaneb, line 170: Since this is used for everything now, perhaps we could rename it to just `concurrency` (or `controller-concurrency`?) and say it is the "Number of CRs of each type to process simultaneously".

zaneb (commented):
Happy with this. Thanks Kashif!
/approve

main.go
kashifest: done

lentzi90 (commented):
/lgtm

PR #1175, opened 2022-10-19
🌱 Refine ironic provisioner log data format

sailorvii:
Refine the log data format from "1666158340.7038972" to "2022-10-19T05:45:40.674Z"

The original format is badly readable.

**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

PR #1249, opened 2023-04-12
Use same settings for provisioner and controller logs 

zaneb:
Previously the ironic provisioner log was created in a separate module,
and didn't have access to the CLI arguments. Originally the provisioner
was always in structured mode even when the -dev flag was passed to put
it into development mode, and then after #1175 vice-versa.

Create the provisioning logger in main.go using the same settings as the
controller logger, and pass it in to the provisioner factory. Only change the
timestamp formatting in development log mode.


lentzi90 (commented):
/lgtm

PR #1107, opened 2022-04-11
Improve webhook verification

zhouhao3:
Part of #855 

- Add validations for ImageUrl
- Add validations for Image.DiskFormat

PR #1250, opened 2023-04-14
:bug: Revert live-iso validation as it is blocking pxe boot when non-virtualmedia drivers are used

hroyrh:
This reverts commit 40986755ad389c5edc51e4feef045d3a6fb19e65 part of this PR - [https://github.com/metal3-io/baremetal-operator/pull/1107/commits](https://github.com/metal3-io/baremetal-operator/pull/1107/commits)

This validation is blocking pxe booting when live-iso image format is used along with non-virtualmedia drivers.

Fixes #


dtantsur:
/approve

I'm impressed that IPMI+ISO actually works in practice (we do test this in the Ironic CI, but I know there are a lot of limitations to it), but we've got evidence that it does.

PR #1252, opened 2023-04-17
🐛BMH add architecture before inspection 

yanfeng1992:

In a multi-architecture scenario, can an annotation be added to bmh to represent the architecture? Through the new annotation field, it is used to generate the kernel and add kernel params in the preprovisioning image for bmh of different architectures

### DIFF
--- a/apis/metal3.io/v1alpha1/baremetalhost_types.go
+++ b/apis/metal3.io/v1alpha1/baremetalhost_types.go
@@ -408,6 +408,12 @@ type BareMetalHostSpec struct {
 	// A custom deploy procedure.
 	// +optional
 	CustomDeploy *CustomDeploy `json:"customDeploy,omitempty"`
+
+	// Architecture means CPU architecture.
+	// Allowed values are x86_64, aarch64, ppc64le, &c.
+	// It's passed to the preprovisioningimage.
+	// +optional
+	Architecture string `json:"architecture,omitempty"`
 }
 
 // AutomatedCleaningMode is the interface to enable/disable automated cleaning
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -659,6 +659,9 @@ func getHostArchitecture(host *metal3v1alpha1.BareMetalHost) string {
 		host.Status.HardwareDetails.CPU.Arch != "" {
 		return host.Status.HardwareDetails.CPU.Arch
 	}
+	if host.Spec.Architecture != "" {
+		return host.Spec.Architecture
+	}
 	if hwprof, err := profile.GetProfile(host.Status.HardwareProfile); err == nil {
 		return hwprof.CPUArch
 	}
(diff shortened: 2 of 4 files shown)


yanfeng1992:
cc @zhouhao3

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -650,11 +651,24 @@ func (r *BareMetalHostReconciler) preprovImageAvailable(info *reconcileInfo, ima
...
+}
+
 func getHostArchitecture(host *metal3v1alpha1.BareMetalHost) string {
 	if host.Status.HardwareDetails != nil &&
 		host.Status.HardwareDetails.CPU.Arch != "" {
 		return host.Status.HardwareDetails.CPU.Arch
 	}
+	if has, architecture := hasArchAnnotation(host); has {
+		return architecture
+	}
```
zhouhao3, line 671: Even if the architecture of the nodes is different, this function can also get the architecture of the current host, so I think it can meet your needs without the code you add.

yanfeng1992: host.Status.HardwareDetails.CPU.Arch comes from the hardware information collected by inspect. My idea is to know the architecture of bmh before inspecting, so as to specify the image of the corresponding architecture for ironic @zhouhao3

zhouhao3: > My idea is to know the architecture of bmh before inspecting, so as to specify the image of the corresponding architecture for ironic.

I don't understand why you would do this.
Currently this function is only called when preparing the provision image. So even if you want to meet your ideas, you should not change it here.

yanfeng1992: Sorry, I have some misunderstandings here, can you provide a contact method to continue the discussion with you, thank you @zhouhao3

dtantsur:
Why does it have to be an annotation? I suggest we add a new normal field instead.

yanfeng1992:
> Why does it have to be an annotation? I suggest we add a new normal field instead.

There is no special reason. I thought that the annotation might change a little bit. How about adding the following fields in BareMetalHostSpec
// BMH architecture.
// +optional
Architecture string `json:"architecture,omitempty"`

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -407,6 +407,10 @@ type BareMetalHostSpec struct {
 	// A custom deploy procedure.
 	// +optional
 	CustomDeploy *CustomDeploy `json:"customDeploy,omitempty"`
+
+	// BMH architecture.
```
zaneb, line 411: I'd like to see a little more documentation here.

* 'architecture' = CPU architecture
* allowed values (historically we have used `x86_64`, so I think the allowed values are `x86_64`, `aarch64`, `ppc64le`, &c. (not `x86-64` or golang CPU arch values like `amd64, `arm64`, &c.)
* where this is used - it's passed to the preprovisioningimage

yanfeng1992: Thanks, I have modified @zaneb

yanfeng1992:
Will the newly added Architecture be put into the new API versions? Should this pr be closed? @dtantsur

dtantsur:
> Will the newly added Architecture be put into the new API versions? Should this pr be closed?

The current plan envision adding Architecture before adding new API versions, so please do keep it open.

PR #1257, opened 2023-04-19
🌱 Bump go module dependencies

kashifest:
This PR bumps the go modules to newer versions.


kashifest:
@zaneb @dtantsur can we take this in? We are planning a release of BMO after this lands

dtantsur:
/approve

Not sure how I feel about just bumping all dependencies... Feels arbitrary. Not a blocking concern though, but let's develop some sort of a policy around it.

PR #1258, opened 2023-04-21
:bug: Delete unmanaged host when deletion timestamp is non-zero

hroyrh:
When we try to delete an unmanaged BMH, it gets stuck since there is no deletion workflow right now. This Pr adds the workflow to delete an unmanaged BMH when the deletionTimestamp is non-zero.

Fixes #


zaneb:
I think #1259 fixes the underlying issue, which should mean this is not needed.

PR #1259, opened 2023-04-24
:bug: Don't check for ironic capacity in 'deleting' state

zaneb:
Ironic doesn't do any work to speak of in the host's 'deleting' state. It appears this was added (in c3c2c8841ca3ebe0341d4c6336346c397bdd99a2) due to confusion between ironic's 'deleting' (i.e. the 'deprovisioning' state in Metal³) and Metal³'s 'deleting' state.

A result of this was that hosts in the 'unmanaged' state could not be deleted, because by definition they do not have BMC credentials, but the check for Ironic capacity required them. Since unmanaged hosts do not go through 'deprovisioning', removing this unnecessary check from the 'deleting' state should be enough to allow them to be deleted again.

### DIFF
--- a/controllers/metal3.io/host_state_machine.go
+++ b/controllers/metal3.io/host_state_machine.go
@@ -104,7 +104,7 @@ func (hsm *hostStateMachine) updateHostStateFrom(initialState metal3v1alpha1.Pro
 		// avoid putting an excessive pressure on the provisioner
 		switch hsm.NextState {
 		case metal3v1alpha1.StateInspecting, metal3v1alpha1.StateProvisioning,
-			metal3v1alpha1.StateDeprovisioning, metal3v1alpha1.StateDeleting:
+			metal3v1alpha1.StateDeprovisioning:
 			if actionRes := hsm.ensureCapacity(info, hsm.NextState); actionRes != nil {
 				return actionRes
 			}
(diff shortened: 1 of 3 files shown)


controllers/metal3.io/host_state_machine.go
```diff
@@ -161,7 +161,7 @@ func (hsm *hostStateMachine) checkDelayedHost(info *reconcileInfo) actionResult
 	// host not yet tracked by the provisioner
 	switch info.host.Status.Provisioning.State {
 	case metal3v1alpha1.StateInspecting, metal3v1alpha1.StateProvisioning,
-		metal3v1alpha1.StateDeprovisioning, metal3v1alpha1.StateDeleting:
+		metal3v1alpha1.StateDeprovisioning:
```
andfasano, line 164: I guess then the `metal3v1alpha1.StateDeleting` should be removed also from the [recordActionDelayed](https://github.com/metal3-io/baremetal-operator/blob/c759a9d23f4eab33eb06ad5f5f74c0201551c71e/controllers/metal3.io/baremetalhost_controller.go#L339) for consistency

andfasano: It sounds also this [test case](https://github.com/metal3-io/baremetal-operator/blob/fa0829d7d7cd7d65a8f755f7a998167f85c8aeb8/controllers/metal3.io/host_state_machine_test.go#L190-L197) could be removed then

PR #1264, opened 2023-04-28
:sparkles: Support /dev/disk/by-path root device hints

zaneb:
Allow paths of the form `/dev/disk/by-path` in the root device hints field.

By-path support is needed because it is the only method of identifying a disk that is stable and doesn't require knowledge of the components of a particular server (e.g. the drive serial number). In particular, HCTL (which you might think would do that) is not stable because the H part is numbered in the order the SCSI hosts are discovered by the kernel.

Support for by-path was previously discussed in #317, but in that case it was proposed to add a separate field. This PR reuses the same field, so you can specify either device name. This avoids the weirdness of two separate fields, each containing a Linux device path, at the possible cost of confusing people into thinking that _any_ Linux device path is valid. To mitigate this, the validating webhook rejects other path types.

When available, the stable by-path name is now reported in the HardwareDetails. Existing consumers that rely on round-tripping this name back through root device hints should continue to work and in fact become more reliable, as the name will now refer to the same device on every boot.

### DIFF
--- a/apis/metal3.io/v1alpha1/baremetalhost_types.go
+++ b/apis/metal3.io/v1alpha1/baremetalhost_types.go
@@ -52,8 +52,9 @@ const (
 // RootDeviceHints holds the hints for specifying the storage location
 // for the root filesystem for the image.
 type RootDeviceHints struct {
-	// A Linux device name like "/dev/vda". The hint must match the
-	// actual value exactly.
+	// A Linux device name like "/dev/vda", or a by-path link to it like
+	// "/dev/disk/by-path/pci-0000:01:00.0-scsi-0:2:0:0". The hint must match
+	// the actual value exactly.
 	DeviceName string `json:"deviceName,omitempty"`
 
 	// A SCSI bus address like 0:0:0:0. The hint must match the actual
--- a/apis/metal3.io/v1alpha1/baremetalhost_validation.go
+++ b/apis/metal3.io/v1alpha1/baremetalhost_validation.go
@@ -5,6 +5,7 @@ import (
 	"net"
 	"net/url"
 	"regexp"
+	"strings"
 
 	"github.com/google/uuid"
 	"github.com/pkg/errors"
@@ -41,6 +42,10 @@ func (host *BareMetalHost) validateHost() []error {
 	}
 
 	if err := validateDNSName(host.Spec.BMC.Address); err != nil {
+		errs = append(errs, err)
+	}
+
+	if err := validateRootDeviceHints(host.Spec.RootDeviceHints); err != nil {
 		errs = append(errs, err)
 	}
 
@@ -177,5 +182,22 @@ func validateImageURL(imageURL string) error {
 		return fmt.Errorf("Image URL %s is an invalid URL", imageURL)
 	}
 
+	return nil
+}
+
+func validateRootDeviceHints(rdh *RootDeviceHints) error {
+	if rdh == nil || rdh.DeviceName == "" {
+		return nil
+	}
+
+	subpath := strings.TrimPrefix(rdh.DeviceName, "/dev/")
+	if rdh.DeviceName == subpath {
+		return fmt.Errorf("Device Name of root device hint must be a /dev/ path, not \"%s\"", rdh.DeviceName)
+	}
+
+	subpath = strings.TrimPrefix(subpath, "disk/by-path/")
+	if strings.Contains(subpath, "/") {
+		return fmt.Errorf("Device Name of root device hint must be path in /dev/ or /dev/disk/by-path/, not \"%s\"", rdh.DeviceName)
+	}
 	return nil
 }
--- a/apis/metal3.io/v1alpha1/baremetalhost_validation_test.go
+++ b/apis/metal3.io/v1alpha1/baremetalhost_validation_test.go
@@ -438,6 +438,62 @@ func TestValidateCreate(t *testing.T) {
 			wantedErr: "BMO validation: failed to parse BMC address information: BMC address hostname/IP : [fe80::fc33:62ff:fe33:8xff] is invalid",
 		},
 		{
+			name: "validRootDeviceHint",
+			newBMH: &BareMetalHost{
+				TypeMeta:   tm,
+				ObjectMeta: om,
+				Spec: BareMetalHostSpec{
+					RootDeviceHints: &RootDeviceHints{
+						DeviceName: "/dev/sda",
+					},
+				},
+			},
+			oldBMH:    nil,
+			wantedErr: "",
+		},
+		{
+			name: "validRootDeviceHintByPath",
+			newBMH: &BareMetalHost{
+				TypeMeta:   tm,
+				ObjectMeta: om,
+				Spec: BareMetalHostSpec{
+					RootDeviceHints: &RootDeviceHints{
+						DeviceName: "/dev/disk/by-path/pci-0000:01:00.0-scsi-0:2:0:0",
+					},
+				},
+			},
+			oldBMH:    nil,
+			wantedErr: "",
+		},
+		{
+			name: "invalidRootDeviceHintByUUID",
+			newBMH: &BareMetalHost{
+				TypeMeta:   tm,
+				ObjectMeta: om,
+				Spec: BareMetalHostSpec{
+					RootDeviceHints: &RootDeviceHints{
+						DeviceName: "/dev/disk/by-uuid/cdaacd50-3a4c-421c-91c0-fe9ba7b8b2f1",
+					},
+				},
+			},
+			oldBMH:    nil,
+			wantedErr: "Device Name of root device hint must be path in /dev/ or /dev/disk/by-path/, not \"/dev/disk/by-uuid/cdaacd50-3a4c-421c-91c0-fe9ba7b8b2f1\"",
+		},
+		{
+			name: "invalidRootDeviceHintNoPath",
+			newBMH: &BareMetalHost{
+				TypeMeta:   tm,
+				ObjectMeta: om,
+				Spec: BareMetalHostSpec{
+					RootDeviceHints: &RootDeviceHints{
+						DeviceName: "sda",
+					},
+				},
+			},
+			oldBMH:    nil,
+			wantedErr: "Device Name of root device hint must be a /dev/ path, not \"sda\"",
+		},
+		{
 			name: "invalidImageURL",
 			newBMH: &BareMetalHost{
 				TypeMeta:   tm,
--- a/config/crd/bases/metal3.io_baremetalhosts.yaml
+++ b/config/crd/bases/metal3.io_baremetalhosts.yaml
@@ -374,7 +374,8 @@ spec:
                               image.
                             properties:
                               deviceName:
-                                description: A Linux device name like "/dev/vda".
+                                description: A Linux device name like "/dev/vda",
+                                  or a by-path link to it like "/dev/disk/by-path/pci-0000:01:00.0-scsi-0:2:0:0".
                                   The hint must match the actual value exactly.
                                 type: string
                               hctl:
@@ -436,8 +437,9 @@ spec:
                   image being provisioned.
                 properties:
                   deviceName:
-                    description: A Linux device name like "/dev/vda". The hint must
-                      match the actual value exactly.
+                    description: A Linux device name like "/dev/vda", or a by-path
+                      link to it like "/dev/disk/by-path/pci-0000:01:00.0-scsi-0:2:0:0".
+                      The hint must match the actual value exactly.
                     type: string
                   hctl:
                     description: A SCSI bus address like 0:0:0:0. The hint must match
@@ -977,7 +979,8 @@ spec:
                                   the image.
                                 properties:
                                   deviceName:
-                                    description: A Linux device name like "/dev/vda".
+                                    description: A Linux device name like "/dev/vda",
+                                      or a by-path link to it like "/dev/disk/by-path/pci-0000:01:00.0-scsi-0:2:0:0".
                                       The hint must match the actual value exactly.
                                     type: string
                                   hctl:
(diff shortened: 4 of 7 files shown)


s3rj1k:
Well, I don't like the idea of reusing field of kernel device name.
Kernel device name is kernel device name.
`by-path` is userspace magic from udev that is generated stably only on single udev version.

(udev changes rules of generation from time to time)

Also what about `by-id`? Do you feel like this is weird to?

dtantsur:
> Also what about by-id? Do you feel like this is weird to?

I think by-id is more or less the same as WWN. Ironic does not support a separate by-id.

s3rj1k:
> > Also what about by-id? Do you feel like this is weird to?
> 
> I think by-id is more or less the same as WWN. Ironic does not support a separate by-id.

It is not the case for nvme disks

dtantsur:
> It is not the case for nvme disks

What do they use? Could you provide an example?

s3rj1k:
> > It is not the case for nvme disks
> 
> What do they use? Could you provide an example?

60-persistent-storage.rules

```
# NVMe
KERNEL=="nvme*[0-9]n*[0-9]", ATTR{wwid}=="?*", SYMLINK+="disk/by-id/nvme-$attr{wwid}"
KERNEL=="nvme*[0-9]n*[0-9]p*[0-9]", ENV{DEVTYPE}=="partition", ATTRS{wwid}=="?*", SYMLINK+="disk/by-id/nvme-$attr{wwid}-part%n"

KERNEL=="nvme*[0-9]n*[0-9]", ENV{DEVTYPE}=="disk", ATTRS{serial}=="?*", ENV{ID_SERIAL_SHORT}="$attr{serial}"
KERNEL=="nvme*[0-9]n*[0-9]", ENV{DEVTYPE}=="disk", ATTRS{wwid}=="?*", ENV{ID_WWN}="$attr{wwid}"
KERNEL=="nvme*[0-9]n*[0-9]", ENV{DEVTYPE}=="disk", ATTRS{model}=="?*", ENV{ID_MODEL}="$attr{model}"
KERNEL=="nvme*[0-9]n*[0-9]", ENV{DEVTYPE}=="disk", ATTRS{firmware_rev}=="?*", ENV{ID_REVISION}="$attr{firmware_rev}"
KERNEL=="nvme*[0-9]n*[0-9]", ENV{DEVTYPE}=="disk", ATTRS{nsid}=="?*", ENV{ID_NSID}="$attr{nsid}"
# obsolete symlink that might get overridden on adding a new nvme controller, kept for backward compatibility
KERNEL=="nvme*[0-9]n*[0-9]", ENV{DEVTYPE}=="disk", ENV{ID_MODEL}=="?*", ENV{ID_SERIAL_SHORT}=="?*", \
  OPTIONS="string_escape=replace", ENV{ID_SERIAL}="$env{ID_MODEL}_$env{ID_SERIAL_SHORT}", SYMLINK+="disk/by-id/nvme-$env{ID_SERIAL}"
KERNEL=="nvme*[0-9]n*[0-9]", ENV{DEVTYPE}=="disk", ENV{ID_MODEL}=="?*", ENV{ID_SERIAL_SHORT}=="?*", ENV{ID_NSID}=="?*",\
  OPTIONS="string_escape=replace", ENV{ID_SERIAL}="$env{ID_MODEL}_$env{ID_SERIAL_SHORT}_$env{ID_NSID}", SYMLINK+="disk/by-id/nvme-$env{ID_SERIAL}"

KERNEL=="nvme*[0-9]n*[0-9]p*[0-9]", ENV{DEVTYPE}=="partition", ATTRS{serial}=="?*", ENV{ID_SERIAL_SHORT}="$attr{serial}"
KERNEL=="nvme*[0-9]n*[0-9]p*[0-9]", ENV{DEVTYPE}=="partition", ATTRS{model}=="?*", ENV{ID_MODEL}="$attr{model}"
KERNEL=="nvme*[0-9]n*[0-9]p*[0-9]", ENV{DEVTYPE}=="partition", ATTRS{firmware_rev}=="?*", ENV{ID_REVISION}="$attr{firmware_rev}"
KERNEL=="nvme*[0-9]n*[0-9]p*[0-9]", ENV{DEVTYPE}=="partition", ATTRS{nsid}=="?*", ENV{ID_NSID}="$attr{nsid}"
# obsolete symlink that might get overridden on adding a new nvme controller, kept for backward compatibility
KERNEL=="nvme*[0-9]n*[0-9]p*[0-9]", ENV{DEVTYPE}=="partition", ENV{ID_MODEL}=="?*", ENV{ID_SERIAL_SHORT}=="?*", \
  OPTIONS="string_escape=replace", ENV{ID_SERIAL}="$env{ID_MODEL}_$env{ID_SERIAL_SHORT}", SYMLINK+="disk/by-id/nvme-$env{ID_SERIAL}-part%n"
KERNEL=="nvme*[0-9]n*[0-9]p*[0-9]", ENV{DEVTYPE}=="partition", ENV{ID_MODEL}=="?*", ENV{ID_SERIAL_SHORT}=="?*",  ENV{ID_NSID}=="?*",\
  OPTIONS="string_escape=replace", ENV{ID_SERIAL}="$env{ID_MODEL}_$env{ID_SERIAL_SHORT}_$env{ID_NSID}", SYMLINK+="disk/by-id/nvme-$env{ID_SERIAL}-part%n"
```

It will generate some ID and not allways based on WWN.
(https://forums.fogproject.org/topic/15739/identical-nvme-drives/17)

zaneb:
> Well, I don't like the idea of reusing field of kernel device name.

That's fair, it's definitely a trade-off. There are two potential alternatives that I see, but they all involve adding an extra field to the HardwareDetails - and the result of that will be that users will have to update their CRDs to use the latest version of the controller. Furthermore, consumers would have to make code changes to get improved behaviour. For completeness, the alternatives are either:

1. Accept either path in the root device hint, but report them separately in HardwareDetails; or
2. Add a separate `BusPath` hint consisting of the bit _after_ `/dev/disks/by-path/`, and also report this in HardwareDetails.

> Kernel device name is kernel device name.

Only until the host is rebooted. After that it might be or it might not.

> by-path is userspace magic from udev that is generated stably only on single udev version.

Perfect is the enemy of good. udev changes are definitely much rarer than reboots.

> It is not the case for nvme disks

Looking at my local NVME disk, there are two by-id symlinks (which is corroborated by the udev rules posted above): one of the form `nvme-eui.<WWN>` and the other of the form `nvme-<Model>_<Serial number>_<NVME Namespace ID>`.
I imagine it's possible that some NVME disks might not have a WWN.

Ironic-inspector reports, and ironic root device hints support matching on, WWN, Model, and/or Serial number. Inspector does not return a list of by-id symlinks for a disk. There is potentially a case to be made that ironic should report the NVME Namespace and allow matching on that, but otherwise I think the actual use cases of `by-id` are already covered.

avishayt:
Why not allow any symlink that points to a valid block device?  Or even limit it to anything under /dev/disk/by-*/?

s3rj1k:
> Why not allow any symlink that points to a valid block device? Or even limit it to anything under /dev/disk/by-*/?

Perfectly valid if we for some reason hesitate to extend CRD.

Also extending CRD shouldn't be a problem @zaneb, removing fields from CRD is a problem, but this is not the case here.

zaneb:
> Why not allow any symlink that points to a valid block device? Or even limit it to anything under /dev/disk/by-*/?

That would require it to be implemented in ironic. And the implementation in ironic would be messy, because currently all root device hints are resolved by matching the hints against the fields of the same name discovered by inspection, using the operator embedded in the hint. There's no provision for matching one member of a list or anything like that. That's why there is a separate `by_path` root device hint in ironic, despite that being a pretty obviously weird interface.

The specific thing you're suggesting, where we a block device by filesystem path and we don't care whether what we have is the original or a symlink, is not possible. What we get is a list of real block devices, for each of which we can discover _some_ of the symlinks to it (those that the kernel knows about), and we have to select one from that list. (It has to work this way because the other hints do _not_ give you a path to start with.)

FWIW there are links under `/dev/disk/by-*` that we definitely would _not_ want to support. e.g. if you used `by-uuid` you'd never be able to provision twice with the same root device hint (because the UUID identifies the _filesystem_ on the disk/partition).

> Also extending CRD shouldn't be a problem @zaneb, removing fields from CRD is a problem, but this is not the case here.

It's true we've done this many times, although mostly with the Spec, and clients don't need to be updated. In this case the bmo itself is the client, so if that gets updated and the CRD _doesn't_ then bmo will stop working. Most people will update them together, and it's an alpha API so we're allowed... I just would like us to start acting as if this were not an alpha API asap.

s3rj1k:
> I just would like us to start acting as if this were not an alpha API asap.

Let's be real here, code is good but we are in Hardware domain here that is ever-changing and ever-buggy and Ironic itself here is double edged sword.

I rather sacrifice "clean code" over usefulness.

Also patching up Ironic with `by-id` shouldn't be that difficult.

zaneb:
@s3rj1k I'm not clear on which of the possible alternatives I mentioned [above](https://github.com/metal3-io/baremetal-operator/pull/1264#issuecomment-1529154807) (or another one I haven't thought of?) you are actually proposing?

avishayt:
I agree that by-uuid and by-label are less useful because they are related to the filesystem and not the disk, but who knows, it may help someone for a one-time provisioning job.  Not important, though.

More importantly, by-id is very useful and has the most persistent IDs, but it is a list because by-id can have multiple symlinks pointing to the same device.  For example:
```
nvme-eui.5cd2e42a91419c24 -> ../../nvme0n1
nvme-INTEL_SSDPEKKF256G8L_PHHP942200RN256B -> ../../nvme0n1
```

Side note - correct me if I'm wrong, but I don't think by-path would ever have more than one entry per device. A disk with multiple paths would appear as multiple devices and then be grouped by multipathd?

dtantsur:
Hi @avishayt @s3rj1k, let's try to unpack what the objections here are, because we seem to be talking about several things.

1) We need `by-id`. This is a possible addition, but someone has to drive the change in Ironic/IPA. I don't think Zane is going to do it so it's on one of you then. I don't think the desire for by-id somehow invalidates the desire to expose the already existing by-path support.

2) Reusing path for `by-path`. If we make by-path a one-off addition, it may make sense to add a new field. However, if we envision supporting arbitrary paths in the future (by-id, by-diskseq, /dev/mapper/ and so on), I think we should not plan to add a new field for each of them. Then path is the natural place to use.

So, the argument that we need by-id eventually makes me feel stronger that the proposed path is right.

s3rj1k:
@dtantsur How one string field would handle an array of `by-ids`?

dtantsur:
> How one string field would handle an array of by-ids?

Why do you need an array though? If you want an "in/contains" operation, this is something that is not specific to by-id: you could use the same with models or serial numbers, for example.

s3rj1k:
> > How one string field would handle an array of by-ids?
> 
> Why do you need an array though? If you want an "in/contains" operation, this is something that is not specific to by-id: you could use the same with models or serial numbers, for example.

I would want an array inside a structure that discovers HW data from IronicInspector.

dtantsur:
> I would want an array inside a structure that discovers HW data from IronicInspector.

Sure, I agree, but this has little to do with the root device hints format. You're still going to use one of the values.

s3rj1k:
> > I would want an array inside a structure that discovers HW data from IronicInspector.
> 
> Sure, I agree, but this has little to do with the root device hints format. You're still going to use one of the values.

Yes, I am bringing this up as yearly as possible so that we design needed API changes with consideration for other possible `by-*` fields.

s3rj1k:
> > Yes, I am bringing this up as early as possible so that we design needed API changes with consideration for other possible by-* fields.
> 
> Right, thank you. Do you think the change Zane proposes actually contradicts it? Module maybe renaming `deviceName` to `devicePath` (can be done in a follow-up).

My only concern that we still do not exposes IronicInspector `by-path` field as reported by Ironic and try to hide it away in inspection results.

I would want Inspection results to be as wide as possible (without hiding fields) and `rootHints` can be magically mangled, this is fine I guess.

dtantsur:
> I would want Inspection results to be as wide as possible (without hiding fields)

This is a proposal we can and should discuss. As an Ironic developer, I have nothing against that :)

zaneb:
> I'd prefer deviceName renamed to devicePath, can be done in a follow-up to avoid triggering another long discussion here.

We can't rename fields, even in an alpha API. We could add another field and deprecate the old one I guess.

zaneb:
> More importantly, by-id is very useful and has the most persistent IDs, but it is a list because by-id can have multiple symlinks pointing to the same device. For example:
> 
> nvme-eui.5cd2e42a91419c24 -> ../../nvme0n1
> nvme-INTEL_SSDPEKKF256G8L_PHHP942200RN256B -> ../../nvme0n1

The most persistent IDs are already selectable using:

```
rootDeviceHints:
  wwn: eui.5cd2e42a91419c24
```

or

```
rootDeviceHints:
  vendor: INTEL
  serialNumber: SSDPEKKF256G8L
```

The only things that by-id adds is a layer of munging by udev and the ability (or, rather, necessity) to match on the NVME namespace. If the latter is a thing that is actually needed, it would be easier and better to add support for that directly than to add by-id.

FWIW I have never once seen or heard of anybody using any of the above fields. In practice, `deviceName` and `minSizeGigabytes` are the only ones that get used. I've seen `hctl` used as a workaround for not having by-path. Nothing else.

zaneb:
> My only concern that we still do not exposes IronicInspector by-path field as reported by Ironic and try to hide it away in inspection results.

@s3rj1k so if in addition to this change we added a field like `alternativeNames` or something to the HardwareDetails, and populated it with a list of the known aliases (just the `/dev/sdX` and by-path ones at the moment), you'd be happy?

dtantsur:
> We could add another field and deprecate the old one I guess.

That's what I meant, yes.

s3rj1k:
> @s3rj1k so if in addition to this change we added a field like `alternativeNames` or something to the HardwareDetails, and populated it with a list of the known aliases (just the `/dev/sdX` and by-path ones at the moment), you'd be happy?

I would be very much happy with that, thanks!

zaneb:
:+1: OK, I can propose that in a separate PR.

zaneb:
#1269

kashifest:
> FWIW I have never once seen or heard of anybody using any of the above fields. In practice, `deviceName` and `minSizeGigabytes` are the only ones that get used. I've seen `hctl` used as a workaround for not having by-path. Nothing else.

FYI we use WWN and serial number downstream.

zaneb:
Ha, I stand corrected :)
In our downstream use cases, people are usually providing the hints before inspection occurs.
There are definitely good reasons to use the WWN/serial if you get to wait until after inspection to choose a disk.

kashifest (approved):

PR #1265, opened 2023-04-28
Added baseline standard manifest file for bmo.

as20203:
Added a manifest file in config folder for enforcing the baseline configuration.


config/default/baseline.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+kind: Namespace
```
tuminoid, line 2: BMO namespace is surely created somewhere already?

as20203: This was done before the discussion last week, there is a namespace manifest already in config/namespace folder. I'll update this PR and add these changes there. Thanks for pointing out.

config/namespace/namespace.yaml
```diff
@@ -3,4 +3,7 @@ kind: Namespace
 metadata:
   labels:
     control-plane: controller-manager
-  name: baremetal-operator-system
\ No newline at end of file
+    kubernetes.io/metadata.name: baremetal-operator-system
+    pod-security.kubernetes.io/enforce: baseline
+    pod-security.kubernetes.io/enforce-version: latest
+  name: baremetal-operator-syste
```
tuminoid, line 9: typo

config/namespace/namespace.yaml
```diff
@@ -3,4 +3,7 @@ kind: Namespace
 metadata:
   labels:
     control-plane: controller-manager
-  name: baremetal-operator-system
\ No newline at end of file
+    kubernetes.io/metadata.name: baremetal-operator-system
+    pod-security.kubernetes.io/enforce: baseline
+    pod-security.kubernetes.io/enforce-version: latest
```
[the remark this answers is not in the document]
as20203: we decided to use the latest and I've also removed this line since latest is the default.

Tiltfile
```diff
@@ -158,6 +158,11 @@ def strip_sec_ctx(yaml):
             spec["securityContext"] = {}
             for container in spec.get("containers", []):
                 container["securityContext"] = {}
+        if data.get("kind") == "Namespace":
+            spec = data["metadata"]["labels"]
+            pod_security_standard_label =  "pod-security.kubernetes.io/enforce" 
+            spec.pop(pod_security_standard_label)
```
tuminoid, line 164: This will raise `KeyError` if the key is not found. Add default vaue of `None` as second argument, so it doesn't happen.

tuminoid (changes requested):

tuminoid:
/retitle :seedling: Added baseline standard manifest file for BMO
/test-ubuntu-integration-main

tuminoid:
I don't think we can enable baseline for BMO namespace. Ironic container violates any of the baseline requirements: https://kubernetes.io/docs/concepts/security/pod-security-standards/

1. `hostNetwork` is not allowed
2. `NET_BROADCAST` capability is not allowed

dtantsur:
> I don't think we can enable baseline for BMO namespace. Ironic container violates the baseline requirements

Correct. If we ever split away the dnsmasq container into a separate pod, we can concentrate these privileges in it. But we'll never be able to get rid of them completely.

PR #1267, opened 2023-05-03
🌱 pin distroless to static:nonroot with sha digest

tuminoid:
Change distroless from base:latest to static:nonroot like other controller images, and pin it with sha digest.


lentzi90 (commented):
/lgtm

PR #1269, opened 2023-05-03
:sparkles: Add a list of alternate disk names to HardwareDetails

zaneb:
This is a follow up (and depends on) #1264.

Give users a way of seeing all of the alternate names for the disk that
we have discovered, even if they are not stable across reboots.


apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -518,10 +519,16 @@ type CPU struct {
...
-	// may not be stable across reboots.
+	// A Linux device name of the disk, e.g.
+	// "/dev/disk/by-path/pci-0000:01:00.0-scsi-0:2:0:0". This will be a name
+	// that is stable across reboots if one is available.
 	Name string `json:"name,omitempty"`
 
+	// A list of alternate Linux device names of the disk, e.g. "/dev/sda".
+	// Note that this list is not exhaustive, and names may not be stable
+	// across reboots.
+	AlternateNames []string `json:"alternateNames,omitempty"`
```
s3rj1k, line 530: To clarify this will be a list of known device names from Inspection **excluding** the one in `Name` field?

Also maybe we should `sort | uniq` this list somewhere?

zaneb: In the proposed implementation it _includes_ the name in the `name` field, and entries are unique. Also, the true kernel device (e.g. `/dev/sda`) always appears first.

pkg/provisioner/ironic/hardwaredetails/hardwaredetails.go
```diff
@@ -110,8 +110,15 @@ func getDiskType(diskdata introspection.RootDiskType) metal3v1alpha1.DiskType {
 func getStorageDetails(diskdata []introspection.RootDiskType) []metal3v1alpha1.Storage {
 	storage := make([]metal3v1alpha1.Storage, len(diskdata))
 	for i, disk := range diskdata {
+		device := disk.Name
+		allNames := []string{device}
+		if disk.ByPath != "" {
+			device = disk.ByPath
```
dtantsur, line 116: This feels wrong to me, you're replacing `Name` with by-path.

zaneb: That was in the other PR that you already lgtmd :)

hardys:
@dtantsur or @s3rj1k perhaps you can take a look since you previously reviewed this and I'm no longer an approver :)

tuminoid:
Please add PR type (:seedling: etc) to the PR title.

hardys:
/retitle :sparkles: Add a list of alternate disk names to HardwareDetails

zaneb:
@hardys looks like you need to submit a review to move yourself out of OWNERS purgatory :smile: (welcome back!)

PR #1275, opened 2023-05-16
🌱 Reduce the preprovisioning image retry delay

dtantsur:
Five minutes is a lot. Neither the built-in image provider nor the one
we have in OpenShift takes so much to build images.

### DIFF
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -54,7 +54,7 @@ import (
 const (
 	hostErrorRetryDelay           = time.Second * 10
 	unmanagedRetryDelay           = time.Minute * 10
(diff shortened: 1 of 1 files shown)


MahnoorAsghar (approved):

zaneb:
Hold on, the controller is [watching](https://github.com/metal3-io/baremetal-operator/blob/main/controllers/metal3.io/baremetalhost_controller.go#L1747) the PreprovisioningImage so it will be notified _immediately_ when something changes.
This delay is for how long before we pointlessly requeue the BMH when _nothing_ has changed. If anything it should maybe be _longer_ than 5 minutes so that we don't pointlessly hammer the API when e.g. the preprovisioning image controller is down.
I think we should revert this.

PR #1277, opened 2023-05-25
:bug: Hard reboot shouldn't wait for soft reboot

zaneb:
If a hard reboot is requested, don't wait for an in-progress soft reboot.

### DIFF
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -1937,8 +1937,9 @@ func (p *ironicProvisioner) PowerOff(rebootMode metal3api.RebootMode, force bool
 
 	if ironicNode.PowerState != powerOff {
 		targetState := ironicNode.TargetPowerState
+		useSoftReboot := rebootMode == metal3api.RebootModeSoft
 		// If the target state is either powerOff or softPowerOff, then we should wait
-		if targetState == powerOff || targetState == softPowerOff {
+		if targetState == powerOff || (useSoftReboot && targetState == softPowerOff) {
 			p.log.Info("waiting for power status to change")
 			return operationContinuing(powerRequeueDelay)
 		}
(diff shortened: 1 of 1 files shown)


pkg/provisioner/ironic/ironic.go
```diff
@@ -1937,8 +1937,9 @@ func (p *ironicProvisioner) PowerOff(rebootMode metal3v1alpha1.RebootMode, force
 
 	if ironicNode.PowerState != powerOff {
 		targetState := ironicNode.TargetPowerState
+		useSoftReboot := rebootMode == metal3v1alpha1.RebootModeSoft
 		// If the target state is either powerOff or softPowerOff, then we should wait
-		if targetState == powerOff || targetState == softPowerOff {
+		if targetState == powerOff || (useSoftReboot && targetState == softPowerOff) {
```
dtantsur, line 1942: Not that it's wrong, but I don't quite understand the point of the change. If `useSoftReboot` is false, `targetState` will never be `softPowerOff`

zaneb: `targetState` is being read from the node in ironic, so it could have been set differently by a previous call to `PowerOff()`. e.g. if a soft power off was in progress ("please give consideration to possibly rebooting this host") but a hard power off request comes in ("fence this node, now") then previously we would have kept waiting to see if the first one worked.

The thing to be sure of is that ironic can handle this sequence of events.

kashifest (commented):
We are planning for a minor release of BMO soon, can we take this bug fix in before that @honza @zaneb

pkg/provisioner/ironic/ironic.go
dtantsur: I think Ironic will hold a lock that will prevent you from issuing another request immediately.

zaneb: :disappointed: 
Is there anything we can do from the Metal³ side that would be correct?

zaneb (commented):
/hold

PR #1279, opened 2023-05-30
Simplify dependencies of the apis module

zaneb:
Avoid most dependencies on controller-runtime from within the `apis` go module. This greatly reduces the number of external packages pulled in by the apis module, and hence the chance of conflicts for anyone who wants to vendor just the API.


apis/metal3.io/v1alpha1/baremetalhost_webhook.go
```diff
@@ -18,24 +18,14 @@ import (
 	"k8s.io/apimachinery/pkg/util/errors"
 
 	"k8s.io/apimachinery/pkg/runtime"
-	ctrl "sigs.k8s.io/controller-runtime"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
```
dtantsur, line 21: We can probably get rid of this part as well, if we make the webhook instance accept a logr.Logger.

zaneb: I made a patch to get rid of this already, then deleted it because it was a bit ugly and the only effect it had was to change logr from an implicit to an explicit dependency.
It _is_ possible to get rid of the controller-runtime module altogether if we did this and moved the [SchemeBuilder](https://github.com/metal3-io/baremetal-operator/blob/main/apis/metal3.io/v1alpha1/groupversion_info.go) stuff out of the API, but that seemed like a fairly disruptive change for not a huge benefit. We can't get rid of the apimachinery dependency in any event - if avoiding controller-runtime could change that then it would be more compelling.

PR #1281, opened 2023-06-06
🌱 Remove explicit version when importing v1alpha1

dtantsur:
We're going to have more API versions, hardcoding v1alpha1 will
no longer be reasonable. Use metal3api as a shorter name.


honza:
I'm not sure if this is the Go way but it
/lgtm

dtantsur:
For the context, it was raised as an action item while discussing https://github.com/metal3-io/metal3-docs/pull/332

/test-ubuntu-integration-main

kashifest (approved):

PR #1282, opened 2023-06-07
:sparkles: Add TLS configuration flags for BMO

as20203:
Added tls configuration flags for bmo.

This enables flags `--tls-min-version` and `--tls-max-version` flags, which can be used to limit TLS versions server accepts. Choices are `TLS12` and `TLS13`, so you can force TLS 1.2 only, TLS 1.3 only, or (default) TLS 1.2 and TLS 1.3 is negotiated between server and client.

This also disables TLS 1.0 and TLS 1.1 as they're insecure and obsolete and not supported by Go TLS library.

### DIFF
--- a/main.go
+++ b/main.go
@@ -21,11 +21,13 @@ import (
 	"fmt"
 	"os"
 	"runtime"
+	"strings"
 
 	"go.uber.org/zap/zapcore"
 	k8sruntime "k8s.io/apimachinery/pkg/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
+	cliflag "k8s.io/component-base/cli/flag"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
@@ -44,10 +46,26 @@ import (
 	// +kubebuilder:scaffold:imports
 )
 
+type TLSVersion string
+
+// Constants for TLS versions.
+const (
+	TLSVersion12 TLSVersion = "TLS12"
+	TLSVersion13 TLSVersion = "TLS13"
+)
+
+type TLSOptions struct {
+	TLSMaxVersion   string
+	TLSMinVersion   string
+	TLSCipherSuites string
+}
+
 var (
-	scheme     = k8sruntime.NewScheme()
-	setupLog   = ctrl.Log.WithName("setup")
-	healthAddr string
+	scheme               = k8sruntime.NewScheme()
+	setupLog             = ctrl.Log.WithName("setup")
+	healthAddr           string
+	tlsOptions           = TLSOptions{}
+	tlsSupportedVersions = []string{"TLS12", "TLS13"}
 )
 
 const leaderElectionID = "baremetal-operator"
@@ -132,6 +150,22 @@ func main() {
 		"Maximum queries per second from the controller client to the Kubernetes API server. Default 20")
 	flag.IntVar(&restConfigBurst, "kube-api-burst", 30,
 		"Maximum number of queries that should be allowed in one burst from the controller client to the Kubernetes API server. Default 30")
+	flag.StringVar(&tlsOptions.TLSMinVersion, "tls-min-version", "TLS12",
+		"The minimum TLS version in use by the webhook server.\n"+
+			fmt.Sprintf("Possible values are %s.", strings.Join(tlsSupportedVersions, ", ")),
+	)
+	flag.StringVar(&tlsOptions.TLSMaxVersion, "tls-max-version", "TLS13",
+		"The maximum TLS version in use by the webhook server.\n"+
+			fmt.Sprintf("Possible values are %s.", strings.Join(tlsSupportedVersions, ", ")),
+	)
+
+	tlsCipherPreferredValues := cliflag.PreferredTLSCipherNames()
+	tlsCipherInsecureValues := cliflag.InsecureTLSCipherNames()
+	flag.StringVar(&tlsOptions.TLSCipherSuites, "tls-cipher-suites", "",
+		"Comma-separated list of cipher suites for the webhook server. "+
+			"If omitted, the default Go cipher suites will be used. \n"+
+			"Preferred values: "+strings.Join(tlsCipherPreferredValues, ", ")+". \n"+
+			"Insecure values: "+strings.Join(tlsCipherInsecureValues, ", ")+".")
 	flag.Parse()
 
 	logOpts := zap.Options{}
@@ -149,7 +183,11 @@ func main() {
 	if leaderElectionNamespace == "" {
 		leaderElectionNamespace = watchNamespace
 	}
-
+	tlsOptionOverrides, err := GetTLSOptionOverrideFuncs(tlsOptions)
+	if err != nil {
+		setupLog.Error(err, "unable to add TLS settings to the webhook server")
+		os.Exit(1)
+	}
 	restConfig := ctrl.GetConfigOrDie()
 	restConfig.QPS = float32(restConfigQPS)
 	restConfig.Burst = restConfigBurst
@@ -158,14 +196,8 @@ func main() {
 		Scheme:             scheme,
 		MetricsBindAddress: metricsBindAddr,
 		WebhookServer: webhook.NewServer(webhook.Options{
-			Port: webhookPort,
-			TLSOpts: []func(*tls.Config){
-				func(c *tls.Config) {
-					if c.MinVersion < tls.VersionTLS12 {
-						c.MinVersion = tls.VersionTLS12
-					}
-				},
-			},
+			Port:    webhookPort,
+			TLSOpts: tlsOptionOverrides,
 		}),
 		LeaderElection:          enableLeaderElection,
 		LeaderElectionID:        leaderElectionID,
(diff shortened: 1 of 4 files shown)


main_test.go
```diff
@@ -0,0 +1,81 @@
+/*
+Copyright 2023 The Kubernetes Authors.
```
tuminoid, line 2: This is original work, so @kashifest we should switch to Metal3 Authors right?

kashifest: Yes

main.go
```diff
@@ -246,3 +284,77 @@ func main() {
...
+		return nil, err
+	}
+	tlsOptions = append(tlsOptions, func(cfg *tls.Config) {
+		cfg.MaxVersion = tlsMaxVersion
+	})
+	// Cipher suites should not be set if empty.
+	if options.TLSMinVersion == "VersionTLS13" && options.TLSMaxVersion == "VersionTLS13" {
+		setupLog.Info("warning: cipher suite flag set to null for TLS version 1.3")
+		options.TLSCipherSuites = ""
+	}
```
lentzi90, line 327: Is this correct? Should the comment be that "Cipher suites should not be set for TLS13"? Why log a warning when doing this? It would make more sense to only log the warning if the user tried setting cipher suites AND required TLS13.

main.go
```diff
@@ -246,3 +284,77 @@ func main() {
...
+					setupLog.Info(fmt.Sprintf("warning: use of insecure cipher '%s' detected.", cipher))
+				}
+			}
+		}
+		tlsOptions = append(tlsOptions, func(cfg *tls.Config) {
+			cfg.CipherSuites = suites
+		})
+	}
+
+	if options.TLSCertFile != "" && options.TLSPrivateKeyFile != "" {
```
lentzi90, line 349: What happens if the user specifies only the cert but no private key? It looks like it is silently ignored. I would expect error out in that case.

as20203: fixed this in accordance with the discussion on the IPAM TLS PR.

lentzi90 (commented):
/lgtm

tuminoid:
Building manager and running `./bin/baremetal-operator --help` says:

```
  -tls-max-version string
    	The maximum TLS version in use by the webhook server.
    	Possible values are VersionTLS10, VersionTLS11, VersionTLS12, VersionTLS13. (default "VersionTLS13")
  -tls-min-version string
    	The minimum TLS version in use by the webhook server.
    	Possible values are VersionTLS10, VersionTLS11, VersionTLS12, VersionTLS13. (default "VersionTLS12")
```

We are not supporting TLS 1.0 or TLS 1.1.

lentzi90 (commented):
/lgtm

tuminoid:
A bit nit-picky, but the options are still `VersionTLS12` format. Why not just `TLS12`?

as20203:
> A bit nit-picky, but the options are still `VersionTLS12` format. Why not just `TLS12`?

The reason to pass it like this is because this is how the function requires it. Below you can see we pass the version to the the funciton.

```
tlsMinVersion, err := cliflag.TLSVersion(options.TLSMinVersion)
	if err != nil {
		return nil, err
	}
```

Now inside this function there is a mapping of strings as below where you can see the keys 

```
var versions = map[string]uint16{
	"VersionTLS10": tls.VersionTLS10,
	"VersionTLS11": tls.VersionTLS11,
	"VersionTLS12": tls.VersionTLS12,
	"VersionTLS13": tls.VersionTLS13,
}
```
If we give only TLS12 it will error out in below function

```
func TLSVersion(versionName string) (uint16, error) {
	if len(versionName) == 0 {
		return DefaultTLSVersion(), nil
	}
	if version, ok := versions[versionName]; ok {
		return version, nil
	}
	return 0, fmt.Errorf("unknown tls version %q", versionName)
}
```

So in conclusion that's why we pass it like this.

tuminoid:
> > A bit nit-picky, but the options are still `VersionTLS12` format. Why not just `TLS12`?
> 
> The reason to pass it like this is because this is how the function requires it. Below you can see we pass the version to the the funciton.
> ```
> 
> So in conclusion that's why we pass it like this.

IMO that is not acceptable reason to use such strings. For example, in the etcd PR which was used as an example how to add this feature, there is a mapping function https://github.com/etcd-io/etcd/pull/15156/files#diff-b4c1a72913d9878ddd102a03fc27e7b78404fc89b80d705832e56b541819c55a that creates user friendly version strings.

lentzi90:
That makes sense, let's add a mapping function and simplify the strings.
/lgtm cancel

as20203:
> > > A bit nit-picky, but the options are still `VersionTLS12` format. Why not just `TLS12`?
> > 
> > 
> > The reason to pass it like this is because this is how the function requires it. Below you can see we pass the version to the the funciton.
> > ```
> > 
> > So in conclusion that's why we pass it like this.
> > ```
> 
> IMO that is not acceptable reason to use such strings. For example, in the etcd PR which was used as an example how to add this feature, there is a mapping function https://github.com/etcd-io/etcd/pull/15156/files#diff-b4c1a72913d9878ddd102a03fc27e7b78404fc89b80d705832e56b541819c55a that creates user friendly version strings.

@lentzi90 @tuminoid just wanted to confirm as per the pr the string are TLS1.2 and TLS1.3 should I keep them like this or simply TLS12 and TLS13?

lentzi90:
Either is fine for me as long as it is shown in the help text :slightly_smiling_face:

main.go
```diff
@@ -246,3 +286,77 @@ func main() {
...
+func GetTLSVersion(version string) (uint16, error) {
+	var v uint16
+
+	switch version {
+	case string(TLSVersion12):
+		v = tls.VersionTLS12
+	case string(TLSVersion13):
+		v = tls.VersionTLS13
+	default:
+		return 0, fmt.Errorf("unexpected TLS version %q (must be one of: TLS1.2, TLS1.3)", version)
```
tuminoid, line 358: This doesn't match the expected value: `TLS1.2` in help text vs `TLS12` in values.

main.go
```diff
@@ -246,3 +286,77 @@ func main() {
...
+		tlsOptions = append(tlsOptions, func(cfg *tls.Config) {
+			cfg.CipherSuites = suites
+		})
+	}
+
+	return tlsOptions, nil
+}
+
+// GetTLSVersion returns the corresponding tls.Version or error.
+func GetTLSVersion(version string) (uint16, error) {
```
tuminoid, line 349: This function also needs a test.

tuminoid (changes requested):

main.go
as20203: added a test for this function to detect incorrect value passed as flag.

main_test.go
```diff
@@ -0,0 +1,90 @@
...
+		klog.SetOutput(bufWriter)
+		klog.LogToStderr(false) // this is important, because klog by default logs to stderr only
+		_, err := GetTLSOptionOverrideFuncs(tlsMockOptions)
+		g.Expect(bufWriter.String()).Should(ContainSubstring("warning: Cipher suites should not be set for TLS version 1.3. Ignoring ciphers"))
+		g.Expect(err).Should(BeNil())
+	})
+}
+
+func TestGetTLSVersion(t *testing.T) {
+	t.Run("should error out when incorrect tls version passed", func(t *testing.T) {
```
lentzi90, line 84: This is only negative test. It would be good to add also a positive test (i.e. it returns the correct version and no error). The other tests could also benefit from positive tests

tuminoid:
As discussed offline with @as20203, I also want to see a test run via test dev-env PR with these options applied to see TLS 1.3 only mode actually works as expected.

main_test.go
as20203: Added positive tests for correct values as well.

tuminoid (approved):
OK, with the final fix of moving the `TLSOpts` to the right place in webhook configuration, this is now finally working as expected.

Final annoyance unrelated to the PR is adding TLS options to `config/default/manager.yaml` doesn't work, since the `args:` is being overwritten by kustomizations instead of merging the configs.

/lgtm

tuminoid:
Please rebase, linter should be fine now.

tuminoid (commented):
/lgtm

as20203:
@dtantsur @zaneb please review.

main.go
```diff
@@ -254,3 +286,76 @@ func main() {
...
+		return nil, err
+	}
+
+	if tlsMaxVersion != 0 && tlsMinVersion > tlsMaxVersion {
+		return nil, fmt.Errorf("TLS version flag min version (%s) is greater than max version (%s)",
+			options.TLSMinVersion, options.TLSMaxVersion)
+	}
+
+	tlsOptions = append(tlsOptions, func(cfg *tls.Config) {
+		cfg.MinVersion = tlsMinVersion
```
zaneb, line 311: In some ways this is a step backwards from what we had before, because even if the default minimum version increases in kubebuilder, we will now still be pinned to TLS 1.2.

tuminoid:
Thanks @zaneb, we'll address those in follow-up PR to increase code quality. I created https://github.com/metal3-io/baremetal-operator/issues/1320 for it.

main.go
```diff
@@ -44,10 +46,26 @@ import (
 	// +kubebuilder:scaffold:imports
 )
 
+type TLSVersion string
+
+// Constants for TLS versions.
+const (
+	TLSVersion12 TLSVersion = "TLS12"
```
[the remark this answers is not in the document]
tuminoid: Kubernetes uses strings like `VersionTLS12` for the same, which we indeed deemed not user-friendly, while `TLS12` is, in my opinion, short and explicit, and hence user-friendly, and also more consistent with the ecosystem.

main.go
tuminoid: New TLS versions are rarely introduced or even minimums stepped up, and changing them have far reaching consequences, that IMO we are better off setting the minimum explicitly instead of minimum coming from external dependency.

main.go
```diff
@@ -44,10 +46,26 @@ import (
...
+
 var (
-	scheme     = k8sruntime.NewScheme()
-	setupLog   = ctrl.Log.WithName("setup")
-	healthAddr string
+	scheme               = k8sruntime.NewScheme()
+	setupLog             = ctrl.Log.WithName("setup")
+	healthAddr           string
+	tlsOptions           = TLSOptions{}
+	tlsSupportedVersions = []string{"TLS12", "TLS13"}
```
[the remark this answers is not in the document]
tuminoid: To be fixed in #1320

main.go
```diff
@@ -254,3 +286,76 @@ func main() {
...
+func GetTLSVersion(version string) (uint16, error) {
+	var v uint16
+
+	switch version {
+	case string(TLSVersion12):
+		v = tls.VersionTLS12
+	case string(TLSVersion13):
+		v = tls.VersionTLS13
+	default:
+		return 0, fmt.Errorf("unexpected TLS version %q (must be one of: TLS12, TLS13)", version)
```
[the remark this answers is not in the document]
tuminoid: To be fixed in #1320

main.go
```diff
@@ -254,3 +286,76 @@ func main() {
...
+	tlsOptions = append(tlsOptions, func(cfg *tls.Config) {
+		cfg.MinVersion = tlsMinVersion
+	})
+
+	tlsOptions = append(tlsOptions, func(cfg *tls.Config) {
+		cfg.MaxVersion = tlsMaxVersion
+	})
+	// Cipher suites should not be set if empty.
+	if options.TLSMinVersion == string(TLSVersion13) &&
+		options.TLSMaxVersion == string(TLSVersion13) &&
```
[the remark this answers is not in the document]
tuminoid: To be fixed in #1320

main.go
```diff
@@ -132,6 +150,22 @@ func main() {
 		"Maximum queries per second from the controller client to the Kubernetes API server. Default 20")
 	flag.IntVar(&restConfigBurst, "kube-api-burst", 30,
 		"Maximum number of queries that should be allowed in one burst from the controller client to the Kubernetes API server. Default 30")
+	flag.StringVar(&tlsOptions.TLSMinVersion, "tls-min-version", "TLS12",
```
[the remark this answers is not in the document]
tuminoid: To be fixed in #1320

main.go
```diff
@@ -132,6 +150,22 @@ func main() {
...
+		"The minimum TLS version in use by the webhook server.\n"+
+			fmt.Sprintf("Possible values are %s.", strings.Join(tlsSupportedVersions, ", ")),
+	)
+	flag.StringVar(&tlsOptions.TLSMaxVersion, "tls-max-version", "TLS13",
```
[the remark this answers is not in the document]
tuminoid: To be fixed in #1320

PR #1287, opened 2023-06-19
 🌱 Add dependabot workflow and configuration

kashifest:
This PR adds dependabot configuration which enables dependabot to automatically open PRs for bumping go modules and tags while ignoring a few of the gomodule update for example controller-runtim or k8s.io/* for manual bumping. This will run as a github workflow.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@ RUN_NAMESPACE = metal3
 GO_TEST_FLAGS = $(TEST_FLAGS)
 DEBUG = --debug
 COVER_PROFILE = cover.out
+GO_VERSION ?= 1.19.10
 
 # CRD Generation Options
 #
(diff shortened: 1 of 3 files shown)


Makefile
tuminoid, line 1: This is not relevant for this PR. Mistakenly added to the commit?

tuminoid (changes requested):

Makefile
kashifest: Actually it is, we missed this in CAPM3 and IPAM, I am adding it for CAPM3 https://github.com/metal3-io/cluster-api-provider-metal3/pull/1028 and for IPAM we need to add it as well. It is used by dependabot [here](https://github.com/metal3-io/baremetal-operator/pull/1287/files#diff-844c68181213ce31638b949e99f9f4df97859cb53840020a31c603a236346ba6R24)

tuminoid: How do they work without it :) Anyways, resolved.

tuminoid (commented):
/lgtm

tuminoid (approved):

PR #1312, opened 2023-07-14
:bug: Proceed to power off after deprovisioning

honza:
Follow up to #1176


controllers/metal3.io/host_state_machine.go
```diff
@@ -536,30 +536,30 @@ func (hsm *hostStateMachine) handleDeprovisioning(info *reconcileInfo) actionRes
...
-				return skipToDelete()
+				return skipToPowerOff()
 			}
 		case actionError:
 			if r.NeedsRegistration() && !hsm.haveCreds {
 				// If the host is not registered as a node in Ironic and we
 				// lack the credentials to deprovision it, just continue to
 				// delete.
-				return skipToDelete()
+				return skipToPowerOff()
```
dtantsur, line 562: This is kinda useless, you cannot power off without credentials.

honza: Done

kashifest:
Can this change be reflected back also in the state diagram https://github.com/metal3-io/baremetal-operator/blob/main/docs/BaremetalHost_ProvisioningState.png ? We are planning for a minor release of BMO soon, can we take this bug fix in before that @honza @zaneb

honza:
Good catch!  I updated the diagram, please have a look.

controllers/metal3.io/host_state_machine.go
```diff
@@ -542,17 +542,23 @@ func (hsm *hostStateMachine) handleDeprovisioning(info *reconcileInfo) actionRes
...
+			hsm.NextState = metal3api.StatePoweringOffBeforeDelete
 			hsm.Host.Status.ErrorCount = 0
 		case actionFailed:
 			// If the provisioner gives up deprovisioning and
 			// deletion has been requested, continue to delete.
 			if hsm.Host.Status.ErrorCount > 3 {
 				info.log.Info("Giving up on host clean up after 3 attempts. The host may still be operational " +
 					"and cause issues in your clusters. You should clean it up manually now.")
-				return skipToDelete()
+				return skipToPowerOff()
```
zaneb, line 561: Given that `skipToPowerOff` and `skipToDelete` are only used one time each, I would suggest either inlining them or making the next state a parameter and having a single function.

docs/BaremetalHost_ProvisioningState.dot
```diff
@@ -51,11 +51,13 @@ digraph BaremetalHost {
     Provisioned -> Deprovisioning [label="!DeletionTimestamp.IsZero()"]
 
     ExternallyProvisioned [shape=doublecircle]
-    ExternallyProvisioned -> Deleting [label="!DeletionTimestamp.IsZero()"]
+    ExternallyProvisioned -> PoweringOffBeforeDelete [label="!DeletionTimestamp.IsZero()"]
```
zaneb, line 54: People have complained about all the `Deleting3 [shape=point]` stuff making it unclear that the host goes into the Deleting state no matter where it was when deleted.
I wonder if this is the point where it finally becomes completely untenable to hide this info for the sake of saving a few lines on the diagram. How bad does it look if we include all the edges?

controllers/metal3.io/host_state_machine.go
honza: Done; inlined the functions, and also changed the prometheus counter to be `deleteWithoutPowerOff` in the case of `skipToDelete`.

docs/BaremetalHost_ProvisioningState.dot
honza: Something like this:

BaremetalHost_ProvisioningState

zaneb: :thinking: that's... not outstanding, but maybe not so bad that it's worth the ongoing confusion to avoid it. WDYT?
We don't actually go to PoweringOffBeforeDelete from Unmanaged or Registering, do we?

zaneb: > We don't actually go to PoweringOffBeforeDelete from Unmanaged or Registering, do we?

We actually [do](https://github.com/metal3-io/baremetal-operator/blob/5a01d08058d63e1e98b4ebe5e6b54c6f50d1837e/controllers/metal3.io/host_state_machine.go#L226-L227), but the (inevitable) error is [handled](https://github.com/metal3-io/baremetal-operator/blob/5a01d08058d63e1e98b4ebe5e6b54c6f50d1837e/controllers/metal3.io/host_state_machine.go#L588-L593), so I guess not technically a bug even though it is a bit messy for the user.

honza: OK, I pushed the diagram changes.  I think it's reasonably clear.

docs/BaremetalHost_ProvisioningState.dot
```diff
@@ -5,40 +5,30 @@ digraph BaremetalHost {
...
 
     Registering -> Inspecting [label="!externallyProvisioned && NeedsHardwareInspection()"]
     Registering -> Preparing [label="!externallyProvisioned && inspectionDisabled()"]
     Registering -> ExternallyProvisioned [label="externallyProvisioned"]
-    Registering -> Deleting2 [label="!DeletionTimestamp.IsZero()"]
-
-    Deleting2 [shape=point]
+    Registering -> PoweringOffBeforeDelete [label="!DeletionTimestamp.IsZero()"]
 
     ExternallyProvisioned -> Inspecting [label="!externallyProvisioned && NeedsHardwareInspection()"]
```
zaneb, line 17: I just discovered that if we change all these ones with && or || to e.g.
```suggestion
    ExternallyProvisioned -> Inspecting [label="!externallyProvisioned &&\nNeedsHardwareInspection()"]
```
it makes everything a bit more compact, and this one in particular fixes that diabolical intersection of the two edges at the top right.

zaneb (commented):
/approve

docs/BaremetalHost_ProvisioningState.dot
honza: Like this?

BaremetalHost_ProvisioningState

docs/BaremetalHost_ProvisioningState.dot
```diff
@@ -5,57 +5,49 @@ digraph BaremetalHost {
 
     Unmanaged [shape=doublecircle]
     Unmanaged -> Registering [label="BMC.* != \"\""]
-    Unmanaged -> Deleting1 [label="!DeletionTimestamp.IsZero()"]
-
-    Deleting1 [shape=point]
+    Unmanaged -> PoweringOffBeforeDelete [label="!DeletionTimestamp.IsZero()"]
```
dtantsur, line 8: This make me worried. Unmanaged hosts cannot be powered off. Isn't it the cause of the issue that we've recently discussed?

docs/BaremetalHost_ProvisioningState.dot
```diff
@@ -5,57 +5,49 @@ digraph BaremetalHost {
...
 
-    Registering -> Inspecting [label="!externallyProvisioned && NeedsHardwareInspection()"]
-    Registering -> Preparing [label="!externallyProvisioned && inspectionDisabled()"]
+    Registering -> Inspecting [label="!externallyProvisioned &&\nNeedsHardwareInspection()"]
+    Registering -> Preparing [label="!externallyProvisioned &&\ninspectionDisabled()"]
     Registering -> ExternallyProvisioned [label="externallyProvisioned"]
-    Registering -> Deleting2 [label="!DeletionTimestamp.IsZero()"]
-
-    Deleting2 [shape=point]
+    Registering -> PoweringOffBeforeDelete [label="!DeletionTimestamp.IsZero()"]
```
dtantsur, line 15: Same

docs/BaremetalHost_ProvisioningState.dot
honza: We already have a follow up for this here #1356

docs/BaremetalHost_ProvisioningState.dot
honza: We already have a follow up for this here #1356

PR #1322, opened 2023-08-04
:bug: Trigger reconcile on Secret change

### LINKED ISSUE #1321: Resources not reconciled when Secrets updated
We use the `Owns()` method on the controller-runtime control to request a reconcile when a Secret we own is updated.

However, by default this only triggers a reconcile on Secrets that have the `controller` flag set in the owner reference. The networkData Secrets have never had this flag set, and since #1164 neither have the BMC Secrets.

zaneb:
The default meaning of the Owns() method in the controller-runtime controller is to only trigger reconciles on owner references where we have the 'controller' flag set.

We have never set the 'controller' flag on the networkData Secret used in both the BareMetalHost and PreprovisioningImage controllers.

Since #1164 we no longer set the 'controller' flag on our owner references for Secrets either, in order to allow the same BMC Secret to be shared amongst multiple hosts.

In each of these case we were either never or no longer triggering a reconcile when the Secret changed. Add the MatchEveryOwner option to ensure that we still trigger a reconcile even when the 'controller' flag is not set in the owner reference.

Fixes #1321

### DIFF
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -36,6 +36,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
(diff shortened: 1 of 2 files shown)


dtantsur:
/lgtm

I wonder if we do need owner references, similar to BMC secrets...

PR #1332, opened 2023-08-16
🌱 Kustomize: Remove use of deprecated functionality

lentzi90:

This PR removes use of deprecated functionality and replaces it with supported alternatives. It should have NO IMPACT on the generated manifests. However, it should be noted that there are differences between the deprecated features and the replacements, which could result in changes in some cases.

- Change `patchesStrategicMerge` to `patches`
- Change `bases` to `resources`
- Change `vars` to `replacements` (and remove `varReference`)
- Ensure all `kustomization.yaml` files has `apiVersion` and `kind` (this is just for consistency).

The biggest change is the `vars -> replacements` update. Replacements are not equivalent to variables and are sometimes not easy to convert. In our case, it was quite straightforward to make the change, but the result is a bit less ... "flexible", i.e. we have to specify exactly what to replace and where, instead of just defining the variables and have them automatically replaced wherever they may be.

Fixes #

Note for reviewers: You may find it easier to review each commit separately.

### DIFF
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -1,3 +1,6 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
 # Adds namespace to all resources.
 namespace: baremetal-operator-system
 
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -24,24 +24,24 @@ bases:
 # [PROMETHEUS] To enable prometheus monitor, uncomment all sections with 'PROMETHEUS'.
 #- ../prometheus
 
-patchesStrategicMerge:
+patches:
 # Protect the /metrics endpoint by putting it behind auth.
 # If you want your controller-manager to expose the /metrics
 # endpoint w/o any authn/z, please comment the following line.
-- manager_auth_proxy_patch.yaml
+- path: manager_auth_proxy_patch.yaml
 
 # Mount the controller config file for loading manager configurations
 # through a ComponentConfig type
-#- manager_config_patch.yaml
+#- path: manager_config_patch.yaml
 
 # [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix including the one in
 # crd/kustomization.yaml
-- manager_webhook_patch.yaml
+- path: manager_webhook_patch.yaml
 
 # [CERTMANAGER] To enable cert-manager, uncomment all sections with 'CERTMANAGER'.
 # Uncomment 'CERTMANAGER' sections in crd/kustomization.yaml to enable the CA injection in the admission webhooks.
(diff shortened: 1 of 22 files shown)


lentzi90:
/hold
I realized we have a lot of commented out patches that are affected by the vars -> replacement change. Let me think about how to handle it...

lentzi90:
/hold cancel
With some inspiration from [kubebuilder](https://github.com/kubernetes-sigs/kubebuilder/blob/68abac1dfb5550f7159dec4ce600cd9b6db925bb/docs/book/src/component-config-tutorial/testdata/project/config/default/kustomization.yaml#L51-L147) it should now be good to go

config/default/kustomization.yaml
```diff
@@ -24,61 +27,133 @@ bases:
...
- 
+
 configMapGenerator:
 - behavior: create
   envs:
   - ironic.env
   name: ironic
+
+# This is for CA injection for the webhook
+replacements:
```
lentzi90, line 60: Note to @kashifest and others about why these replacements are not in e.g. certmanager or crd kustomizations: It is not possible to add the replacements there since the resource they take the value from (the webhook service) does not exist in there. So we must actually add them here.
You can see that they do the same in kubebuilder nowadays here: https://github.com/kubernetes-sigs/kubebuilder/blob/68abac1dfb5550f7159dec4ce600cd9b6db925bb/docs/book/src/component-config-tutorial/testdata/project/config/default/kustomization.yaml#L51-L147

lentzi90:
Hmm... I see the inspiration I took from kubebuilder is actually now causing a difference in the generated manifest. I would say it is harmless, just injecting the CA in all CRDs instead of only BMH, but lets see what others think

config/render/capm3.yaml
lentzi90, line 1: The changes in this file is due to the `create: true` option in the `replacements`. It causes the annotation to be injected in all CRDs instead of just replacing the value where it exists. I would say this is not an issue, but please let me know if you think differently.
The reason why kubebuilder did it this way is (probably) the same as I, convenience. With `create: true` we don't have to list each individual CRD that should have the annotation, since it will be added for all of them. So we could easily do it the other way, remove `create: true` and enumerate where exactly it should be (for us just BMH). However, the not so nice thing then is that we have commented out patches to add this annotation to all CRDs, so then we should also have commented out replacements for all of them...

lentzi90:
/hold
Checking kubebuilder migration steps, if that makes sense instead of manual update of these.

kashifest:
@lentzi90 is this PR still relevant?

lentzi90:
Not in its current form. The idea is to sort out the `replacements` issue in kubebuilder first and then update this PR to do the same. I will mark it as WIP for now.

lentzi90:
Hmm now that I think about it it will probably be more work to rebase/update this than just creating a new PR. I will do that instead.

PR #1337, opened 2023-08-22
🌱 Bump golangci-lint to v1.54.2

kashifest:
v1.54.2 golangci-lint patch has introduced some new rules for gosec. This PR bump the golangci-lint and fixes the code since dependabot PR couldn't fix it automatically.


honza:
We just merged an upgrade of golangci-lint yesterday, let's keep an eye on the frequency, it's starting to get tedious.

/approve

kashifest:
> We just merged an upgrade of golangci-lint yesterday, let's keep an eye on the frequency, it's starting to get tedious.
> 
Yes, unfortunately the patch version came out after that and dependabot was trying to bump it but with issues.

PR #1338, opened 2023-08-22
[WIP] Allow selecting TLS version in ironic client transactions

elfosardo:
Add the possibility to specify Min and Max TLS versions used by the ironic client transactions

### DIFF
--- a/pkg/provisioner/ironic/clients/client.go
+++ b/pkg/provisioner/ironic/clients/client.go
@@ -27,6 +27,8 @@ type TLSConfig struct {
 	ClientPrivateKeyFile  string
 	InsecureSkipVerify    bool
 	SkipClientSANVerify   bool
+	TLSMinVersion         uint16
+	TLSMaxVersion         uint16
 }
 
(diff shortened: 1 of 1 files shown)


pkg/provisioner/ironic/clients/client.go
```diff
@@ -27,6 +27,8 @@ type TLSConfig struct {
 	ClientPrivateKeyFile  string
 	InsecureSkipVerify    bool
 	SkipClientSANVerify   bool
+	TLSMinVersion         uint16
+	TLSMaxVersion         uint16
```
zaneb, line 31: What's the point of this if it's not set in `loadTLSConfigFromEnv()`?

Also the values here are something like 772 for TLS 1.3 (`0x0304` in hex, but that's not much better). That's not a great interface for the user passing these values in the environment.

tuminoid:
Nice! Just to clarify, this config would affect which TLS versions BMO provisioner uses as TLS client when it talks to... whom?

PR #1349, opened 2023-09-01
🌱 fix issue with git directory ownership in golint

tuminoid:
Run docker as same uid as user, so Git doesn't complain about directory ownership. This makes linting work locally as well.


Sunnatillo (commented):
LGTM

kashifest (approved):

PR #1304, opened 2023-06-29
🌱 BMO kustomization components

lentzi90:
**What this PR does / why we need it**:

This reworks the BMO kustmozation to use components (a kustomize feature) instead of overlays. This simplifies the logic quite a lot because we do not need the nested overlays (e.g. TLS inside basic-auth instead of just adding them together).
The PR also updates some kustomize syntax to avoid deprecated features.

NOTE! There were quite a few webhook patches that were commented out. I have removed them all. If you think they are needed, please let me know!

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

PR #1353, opened 2023-09-07
⚠️ adopt manifests for kustomize v5

LittleChimera:

building kustomize layers with kustomize v5 outputs warnings of deprecation

```
$ kustomize build default/ > render/capm3.yaml 
# Warning: 'bases' is deprecated. Please use 'resources' instead. Run 'kustomize edit fix' to update your Kustomization automatically.
# Warning: 'patchesStrategicMerge' is deprecated. Please use 'patches' instead. Run 'kustomize edit fix' to update your Kustomization automatically.
# Warning: 'vars' is deprecated. Please use 'replacements' instead. [EXPERIMENTAL] Run 'kustomize edit fix' to update your Kustomization automatically.
# Warning: 'patchesStrategicMerge' is deprecated. Please use 'patches' instead. Run 'kustomize edit fix' to update your Kustomization automatically.
```

### DIFF
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -12,7 +12,7 @@ namePrefix: baremetal-operator-
 #commonLabels:
 #  someName: someValue
 
-bases:
+resources:
 - ../crd
 - ../rbac
 - ../manager
@@ -24,11 +24,11 @@ bases:
 # [PROMETHEUS] To enable prometheus monitor, uncomment all sections with 'PROMETHEUS'.
 #- ../prometheus
 
-patchesStrategicMerge:
+patches:
 # Protect the /metrics endpoint by putting it behind auth.
 # If you want your controller-manager to expose the /metrics
 # endpoint w/o any authn/z, please comment the following line.
-- manager_auth_proxy_patch.yaml
+- path: manager_auth_proxy_patch.yaml
 
 # Mount the controller config file for loading manager configurations
 # through a ComponentConfig type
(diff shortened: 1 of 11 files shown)


tuminoid:
Possibly related: https://github.com/metal3-io/baremetal-operator/pull/1304
/cc @lentzi90

config/default/kustomization.yaml
```diff
@@ -24,59 +24,101 @@ bases:
...
 # [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix including the one in
 # crd/kustomization.yaml
-- manager_webhook_patch.yaml
+- path: manager_webhook_patch.yaml
 
 # [CERTMANAGER] To enable cert-manager, uncomment all sections with 'CERTMANAGER'.
 # Uncomment 'CERTMANAGER' sections in crd/kustomization.yaml to enable the CA injection in the admission webhooks.
 # 'CERTMANAGER' needs to be enabled to use ca injection
-- webhookcainjection_patch.yaml
+- path: webhookcainjection_patch.yaml
```
lentzi90, line 44: A note for the future: This one will work as it is, but if we at some point want to enable to mutating webhook (which is now in the same file but commented out) then it will break as long as https://github.com/kubernetes-sigs/kustomize/issues/5049 is unresolved.
It is easy to fix though so no need to worry about it. Just split the patch into two files.

lentzi90 (commented):
Thanks for working on this!
I have been working upstream with kustomize and kubebuilder in https://github.com/kubernetes-sigs/kustomize/issues/5049 and https://github.com/kubernetes-sigs/kubebuilder/issues/3539 to get this fixed at the source. I believe we will end up removing the CA injection patches and just rely on `replacements` to set the annotations. However, I'm happy to take this in while still figuring out the details upstream. (We don't have the issue of multiple patches in one file, since we have the mutating webhook patch commented out.) When we reach a conclusion upstream I will also try to propagate it here to keep things consistent.

/lgtm

lentzi90:
No no I'm happy to take this. This will make it easier for me to progress with what I was actually trying to do, i.e. reorganize and add Kustomize Components :slightly_smiling_face:

lentzi90:
Yes let's get an approver to look at it!
/assign @kashifest

kashifest:
/retitle :seedling: adopt manifests for kustomize v5

PR #1355, opened 2023-09-13
⚠️ Use ironic inventory API instead of calling ironic-inspector

dtantsur:
Removes the direct dependency on ironic-inspector, leaving it an implementation detail of Ironic.

Note for reviewers: it's much easier to review each of 3 commits separately.


hardys:
@dtantsur it might perhaps be helpful to clarify the Ironic dependencies related to this change (since not all users are consuming the upstream/master images), e.g I see the [API docs](https://docs.openstack.org/api-ref/baremetal/#node-inventory) say `New in version 1.81` - 

Looking at the [OpenStack stable branches](https://releases.openstack.org) and the release notes it looks like this requires Ironic 21.3.0 so this should work with `2023.1` onwards I think?

dtantsur:
@hardys (hey Steve, been a while!) this is an interesting point, would you like to bring it to the upstream meeting? The thing is, we've never expressed *any* explicit requirement on Ironic, essentially assuming that everyone is using the latest Ironic.

Meanwhile, https://docs.openstack.org/ironic/latest/contributor/webapi-version-history.html is a helpful link to match API versions to Ironic releases, in this case - from the Antelope series (we forgot to put exact numbers, I'll fix that).

hardys:
> @hardys (hey Steve, been a while!) this is an interesting point, would you like to bring it to the upstream meeting? The thing is, we've never expressed _any_ explicit requirement on Ironic, essentially assuming that everyone is using the latest Ironic.

Sure I'll raise it in the meeting so we can hopefully ensure we're all aligned - IMO assuming everyone is using latest/main Ironic probably isn't ideal - but in this case it seems the change is compatible with the current stable `2023.1` Ironic which is probably fine, thanks for clarifying!

cmd/get-hardware-details/main.go
```diff
@@ -58,7 +59,7 @@ func main() {
 
 func getOptions() (o options) {
 	if len(os.Args) != 3 {
-		fmt.Println("Usage: get-hardware-details <inspector URI> <node UUID>")
+		fmt.Println("Usage: get-hardware-details <ironic URI> <node UUID>")
```
zaneb, line 62: This looks like a backwards-compat nightmare :(

Could we do some sort of temporary hack where we the substitute the well-known ironic port number for the well-known inspector port number when it appears maybe?

dtantsur: I wonder if this tool is even used (I'd rather have people use the `baremetal` CLI from Ironic, sigh). I'll see if I can invent something.

hardys: I think it's used in the [OpenShift installer](https://github.com/openshift/installer/blob/master/data/data/bootstrap/baremetal/files/usr/local/bin/master-bmh-update.sh#L45) ;)

zaneb: Yep, it's definitely used by us, but _probably_ nobody else.

cmd/get-hardware-details/main.go
```diff
@@ -34,22 +36,34 @@ func main() {
...
+	endpoint := opts.Endpoint
+	parsedEndpoint, err := url.Parse(endpoint)
 	if err != nil {
-		fmt.Printf("could not get inspector client: %s", err)
+		fmt.Printf("invalid ironic endpoint: %s", err)
 		os.Exit(1)
 	}
 
-	introData := introspection.GetIntrospectionData(inspector, opts.NodeID)
+	if parsedEndpoint.Port() == "5050" {
```
zaneb, line 46: Big comment explaining what this is doing and when it can be removed wouldn't hurt.

cmd/get-hardware-details/main.go
```diff
@@ -34,22 +36,34 @@ func main() {
...
+		parsedEndpoint.Host = strings.Replace(parsedEndpoint.Host, ":5050", ":6385", 1)
```
zaneb, line 47: nit (well, more of a pro tip) - you could also do:

```
parsedEndpoint.Host = net.JoinHostPort(parsedEndpoint.Hostname(), "6385")
```

zaneb (commented):
/approve

cmd/get-hardware-details/main.go
dtantsur: Okay, I'll follow-up with it.

go.mod
```diff
@@ -5,7 +5,7 @@ go 1.20
 require (
 	github.com/go-logr/logr v1.2.4
 	github.com/google/safetext v0.0.0-20230106111101-7156a760e523
-	github.com/gophercloud/gophercloud v1.6.0
+	github.com/gophercloud/gophercloud v1.5.1-0.20230912084133-c79ed6d3b371
```
lentzi90, line 8: I'm a bit uncomfortable with this. Is the goal to merge it with the development commit or wait for gophercloud v2 to be released?

lentzi90 (commented):
This looks like a really good refactoring/clean up!
I'm a bit uncomfortable with the gophercloud development branch pulled in instead of an actual release though... Putting hold so we don't merge it accidentally before we get more eyes on it.

/hold

kashifest:
Since this specific feature is using a feature from newest ironic release, Shall we have a BMO release before this patch lands to make sure whoever is using an older version ironic with BMO, shall atleast have a newer BMO release to work with. Also this looks like a ✨ to me.

go.mod
dtantsur: V2 may take more time to get releases. What's the cause of your discomfort? For better or worse, using git commits is a normal thing in the Go world. The last time I tried to hurry gophercloud folks about a release they went "wtf, just use a git hash".

dtantsur:
>  Shall we have a BMO release before this patch lands to make sure whoever is using an older version ironic with BMO, shall atleast have a newer BMO release to work with.

+1, wanna do that?

> Also this looks like a ✨ to me.

It's not user-visible though, hence my choice.

go.mod
kashifest: Would this be landing in v2 only? Do you have any guess when would that be? AFAIK when we need to test commits that has landed in dev branches but not ready for release yet, we use pseudo-versions with git hashes.  Use of pseudo version signals module is still under development and not stable. But since go has now stricter rules now to generate pseudo-numbers, it can be used incase we cannot wait for proper release to exist.

dtantsur: Yep, this is going to v2 only because this work involved breaking changes. I don't know when that happens, we can ask on slack.

I think I'm using what you're describing here. Since the last release before splitting the 1.* branch was 1.5.0, it assumes the new version will 1.5.1-something (which is wrong, but go tell go!)

cmd/get-hardware-details/main.go
s3rj1k: @zaneb being modest is a virtue

lentzi90:
> > Also this looks like a ✨ to me.
> 
> It's not user-visible though, hence my choice.

It is definitely breaking to users that depend on a specific (older) version of Ironic though since this bumps the required version. I would say :warning: or :sparkles: is needed to highlight this impact.

go.mod
lentzi90: > What's the cause of your discomfort?

I'm mainly worried about the upgrade path and the potential for API changes that would impact us. It is *probably* not an issue but I would not want to just skip over it without thinking it through. So how do we expect the upgrade will look like?

We start with a scenario where a user is running an older version of BMO/Ironic relying on the older API and gophercloud. What would be required to upgrade? My understanding is something like this:

1. Upgrade Ironic to a version that supports microversion 1.81. (Is something else needed here to "refresh" the inventory, I guess not?)
2. Upgrade BMO.

This seems easy enough, but definitely worth a mention in the release notes for BMO.

Now my worry is that there would be some migration steps needed to go from gophercloud v1 to v2 and we end up somewhere in the middle. Do you see any situation where some kind of patch would be needed to support the bump to v2? (I know of projects where bugs have made it impossible to upgrade from e.g. 0.3.1 to 0.4.0. Instead users were required to first upgrade to the latest patch release, e.g. 0.3.4 and from there on to 0.4.0.) For example, could there be a required upgrade path for gophercloud to do v1.7.1 -> v2.0.0 and we miss it because we do not have the v1.7.1 patch?

dtantsur: Hmm, you seem to be conflating the Ironic upgrade and the Gophercloud upgrade.

Gophercloud is just an HTTP library. 2.0.0 will see some Go API changes, but these are easy to address on any bump. For better or worse, Go does not update anything automatically (we will need to keep an eye on dependabot, of course). This is why I'm not sure what you mean by "we don't have the v1.7.1 patch". Gophercloud has no local state.

Ironic upgrade is a bit different. Historically, we have taken bumping microversions very lightly. Now Steven has raised a valid point that some of our consumers may not always use the latest-and-greatest versions. Fair.  We need to start having some sort of upgrade docs per release (which I assume don't exist now).

In any case, what are the suggested action items here?

lentzi90: Thanks for the explanation, that makes sense! 
It seems like there is not much reason to worry. We just need to document the Ironic version bump so users notice it.
Perhaps we can bring this up at the community meeting today also to make sure developers are aware of it, in case there are any concerns?

kashifest:
> > Shall we have a BMO release before this patch lands to make sure whoever is using an older version ironic with BMO, shall atleast have a newer BMO release to work with.
> 
> +1, wanna do that?

@metal3-io/metal3_release_team can you folks do a BMO release before this PR merges.

tuminoid:
> > > Shall we have a BMO release before this patch lands to make sure whoever is using an older version ironic with BMO, shall atleast have a newer BMO release to work with.
> > 
> > 
> > +1, wanna do that?
> 
> @metal3-io/metal3_release_team can you folks do a BMO release before this PR merges.

We checked the main branch content, and there is not a lot to release before we have first breaking change/new feature. 0.4.1 is hence out of the question already and if we're bumping to 0.5.0, this should fit right in with the other changes for later release in regular schedule.

kashifest:
> We checked the main branch content, and there is not a lot to release before we have first breaking change/new feature merged after 0.4.0. 0.4.1 is hence out of the question already and if we're bumping to 0.5.0, this should fit right in with the other changes for later release in regular schedule.

Good point, two ways, we go to v0.5.0 if we cut a release now, or we dont cut a release and do a minor release in due time. Its a pity again that we won't be able to patch releases 0.4.x anymore.

tuminoid:
> > We checked the main branch content, and there is not a lot to release before we have first breaking change/new feature merged after 0.4.0. 0.4.1 is hence out of the question already and if we're bumping to 0.5.0, this should fit right in with the other changes for later release in regular schedule.
> 
> Good point, two ways, we go to v0.5.0 if we cut a release now, or we dont cut a release and do a minor release in due time. Its a pity again that we won't be able to patch releases 0.4.x anymore.

OK, as discussed: we are already working to add CI for release branches, so we keep doing that and will release 0.4.1 from the branch when able. 0.5.0 will be release from main later on. 

The key here is that Ironic image is not pinning Ironic source, so releasing BMO from where ever does not magically build old Ironic anymore. We need Ironic source pinning for that.

Rozzii:
> > > We checked the main branch content, and there is not a lot to release before we have first breaking change/new feature merged after 0.4.0. 0.4.1 is hence out of the question already and if we're bumping to 0.5.0, this should fit right in with the other changes for later release in regular schedule.
> > 
> > 
> > Good point, two ways, we go to v0.5.0 if we cut a release now, or we dont cut a release and do a minor release in due time. Its a pity again that we won't be able to patch releases 0.4.x anymore.
> 
> OK, as discussed: we are already working to add CI for release branches, so we keep doing that and will release 0.4.1 from the branch when able. 0.5.0 will be release from main later on.
> 
> The key here is that Ironic image is not pinning Ironic source, so releasing BMO from where ever does not magically build old Ironic anymore. We need Ironic source pinning for that.

Yep, branching 0.4.0 out at an "earlier" and working on that branch will serve as a good separation line between what has come before this commit and what has come after. We also don't need to rush 0.5.0 release (or any release) with this approach .

Rozzii:
> Okay, so are we good to proceed with this PR?

Yes.

lentzi90 (commented):
/hold cancel
I'll let someone with better understanding of this put lgtm

lentzi90:
Forgot about this: https://github.com/metal3-io/baremetal-operator/blob/137006a488b28105200b3da320a57dd66561d99f/docs/configuration.md?plain=1#L73
Please update the required Ironic version there @dtantsur

PR #1365, opened 2023-09-26
📖 fix releasing.md on annotated tags

tuminoid:
-m also implies -a, so neither of them can be present.


tuminoid:
/override test-ubuntu-integration-main 
/override test-centos-e2e-integration-main 

Integration tests do not relate to documentation. Save trees.

lentzi90 (commented):
/lgtm

Issue #1347, opened 2023-08-30
Deleting the last BMH using a credentials secret, causes the secret to be also deleted.


raffaelespazzoli:
**What steps did you take and what happened:**
Deleting the last BMH using a credentials secret, causes the secret to be also deleted.
This is not desirable.

**What did you expect to happen:**
The secret to be left alone.

**Anything else you would like to add:**
I don't understand what was the intention with using the owner reference in this case. It does seem unnecessary.

**Environment:**

- Baremetal Operator version: OCP 4.13
- Environment (metal3-dev-env or other):

/kind bug

zaneb:
I think the logic was that we should not leave credentials lying around when they are not used by anything.
Some more information about why it's undesirable would be helpful.

zaneb:
We also user the owner reference as a reverse link from the Secret to the BMH that owns it, so we can easily tell which BMH(s) need to be reconciled when a Secret is modified without having to search all of them. controller-runtime makes this very simple to set up by just calling `Owns(&corev1.Secret{})` on the controller manager, so that may have been part of the reason it was used initially.

raffaelespazzoli:
the ownership concept in kubernetes generally refers to the fact a controller of resource A has created resource B in response to resource A configuration. So, in a way resource B exists only for the fact that resource A exists and therefore A owns B. The garbage collection logic aligns with this assumption and will delete B on behalf of controller A if A is deleted. This is to avoid writing trivial finalizers.
This is not the case for this BMH controller, you don't create the secret therefore you don't own it. I get that you get some convenient behavior by setting the owner, but what you are doing is conceptually wrong and it is hurting situations where the secret should not be deleted.

lentzi90:
I agree that the current behavior is surprising. Perhaps a global flag would be cleanest?

Rozzii:
So is there any conclusion as far as I can understand the closest to an acceptable compromise would be to turn this feature into a configurable one right ? @dtantsur @s3rj1k @lentzi90 @raffaelespazzoli @zaneb  ?

hardys:
This was also [recently discussed](https://github.com/metal3-io/baremetal-operator/pull/1380#discussion_r1347372523) in relation to networkData secrets - it seems we're not consistently setting ownership of those (and presumably meta/userData) so we might want to consider if there's a common (but configurable) way to solve this?

zaneb:
We do consistently set ownership of user/meta/networkData and BMC Secrets.
[...]

PR #1380, opened 2023-10-05
:sparkles: Pass preprovisioningNetworkData to Ironic

hardys:

Ironic supports [DHCP-less provisioning](https://docs.openstack.org/ironic/latest/admin/dhcp-less.html), which I expected to work via Metal<sup>3</sup>, particularly since we already provide the [IPAM controller](https://github.com/metal3-io/ip-address-manager) which can be used to statically assign control plane IPs.

[Recent discussions](https://kubernetes.slack.com/archives/CHD49TLE7/p1695719774588889) however highlighted that this flow is not currently possible via Metal<sup>3</sup>, unless you implement a custom `PreprovisioningImage` controller, or have other downstream customisations.

In the case where `--build-preprov-image` is passed to BMO, we enable creation of the `PreprovisioningImage` CR and an internal default controller (which does not embed the `preprovisioningNetworkData`, but downstream custom controllers are doing so, so we don't want to break existing users opting into this flow)

In the case where `--build-preprov-image` is *not* passed, Ironic can be configured to inject `network_data` into the Ramdisk, and IMO this seems like reasonable default behaviour when no image controller is enabled. 

In this mode if the user provides `preprovsioningNetworkDataName` we'll set the Node `network_data` on initial registration of the Ironic node.  

This means when IPA boots for inspection, the virtual media device can be mounted and the `openstack/latest/network_data.json` file retrieved to configure the network (e.g by `glean` in the upstream Ironic flow, but users can build a Ramdisk image with any tool/script that consumes this file)

Note that I don't think the current upstream IPA image currently includes simple-init/glean so this will still require a custom Ramdisk image, but it does mean we can support DHCP-less without mandating a custom `PreprovisioningImage` controller.

### DIFF
--- a/controllers/metal3.io/host_config_data.go
+++ b/controllers/metal3.io/host_config_data.go
@@ -94,6 +94,26 @@ func (hcd *hostConfigData) NetworkData() (string, error) {
 	return networkDataRaw, err
 }
 
+// PreprovisioningNetworkData get preprovisioning network configuration
+func (hcd *hostConfigData) PreprovisioningNetworkData() (string, error) {
+	if hcd.host.Spec.PreprovisioningNetworkDataName == "" {
+		return "", nil
+	}
+	networkDataRaw, err := hcd.getSecretData(
+		hcd.host.Spec.PreprovisioningNetworkDataName,
+		hcd.host.Namespace,
+		"networkData",
+	)
+	if err != nil {
+		_, isNoDataErr := err.(NoDataInSecretError)
+		if isNoDataErr {
+			hcd.log.Info("PreprovisioningNetworkData networkData key is not set, returning empty data")
+			return "", nil
+		}
+	}
+	return networkDataRaw, err
+}
+
 // MetaData get host metatdata
 func (hcd *hostConfigData) MetaData() (string, error) {
 	if hcd.host.Spec.MetaData == nil {
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -456,6 +456,25 @@ func (p *ironicProvisioner) ValidateManagementAccess(data provisioner.Management
 		// below.
 	}
 
+	// If no PreprovisioningImage builder is enabled we set the Node network_data
+	// this enables Ironic to inject the network_data into the ramdisk image
+	if !p.config.havePreprovImgBuilder {
+		networkDataRaw := data.PreprovisioningNetworkData
+		if networkDataRaw != "" {
+			var networkData map[string]interface{}
+			if yamlErr := yaml.Unmarshal([]byte(networkDataRaw), &networkData); yamlErr != nil {
+				p.log.Info("failed to unmarshal networkData from PreprovisioningNetworkData")
+				result, err = transientError(fmt.Errorf("invalid preprovisioningNetworkData: %w", yamlErr))
+				return
+			}
+			numUpdates := len(updater.Updates)
+			updater.SetTopLevelOpt("network_data", networkData, ironicNode.NetworkData)
+			if len(updater.Updates) != numUpdates {
+				p.log.Info("adding preprovisioning network_data for node", "node", ironicNode.UUID)
+			}
+		}
+	}
+
 	ironicNode, success, result, err := p.tryUpdateNode(ironicNode, updater)
 	if !success {
 		return
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -802,14 +802,25 @@ func (r *BareMetalHostReconciler) registerHost(prov provisioner.Provisioner, inf
 		return actionError{err}
 	}
 
+	hostConf := &hostConfigData{
+		host:          info.host,
+		log:           info.log.WithName("host_config_data"),
+		secretManager: r.secretManager(info.log),
+	}
+	preprovisioningNetworkData, err := hostConf.PreprovisioningNetworkData()
+	if err != nil {
+		return recordActionFailure(info, metal3api.RegistrationError, "failed to read preprovisioningNetworkData")
+	}
+
 	provResult, provID, err := prov.ValidateManagementAccess(
 		provisioner.ManagementAccessData{
-			BootMode:              info.host.Status.Provisioning.BootMode,
-			AutomatedCleaningMode: info.host.Spec.AutomatedCleaningMode,
-			State:                 info.host.Status.Provisioning.State,
-			CurrentImage:          getCurrentImage(info.host),
-			PreprovisioningImage:  preprovImg,
-			HasCustomDeploy:       hasCustomDeploy(info.host),
+			BootMode:                   info.host.Status.Provisioning.BootMode,
+			AutomatedCleaningMode:      info.host.Spec.AutomatedCleaningMode,
+			State:                      info.host.Status.Provisioning.State,
+			CurrentImage:               getCurrentImage(info.host),
+			PreprovisioningImage:       preprovImg,
+			PreprovisioningNetworkData: preprovisioningNetworkData,
+			HasCustomDeploy:            hasCustomDeploy(info.host),
 		},
 		credsChanged,
 		info.host.Status.ErrorType == metal3api.RegistrationError)
--- a/controllers/metal3.io/host_config_data_test.go
+++ b/controllers/metal3.io/host_config_data_test.go
@@ -92,17 +92,19 @@ func TestProvisionWithHostConfig(t *testing.T) {
 	testBMCSecret := newBMCCredsSecret(defaultSecretName, "User", "Pass")
 
 	testCases := []struct {
-		Scenario                 string
-		Host                     *metal3api.BareMetalHost
-		UserDataSecret           *corev1.Secret
-		PreprovNetworkDataSecret *corev1.Secret
-		NetworkDataSecret        *corev1.Secret
-		ExpectedUserData         string
-		ErrUserData              bool
-		ExpectedNetworkData      string
-		ErrNetworkData           bool
-		ExpectedMetaData         string
-		ErrMetaData              bool
+		Scenario                           string
+		Host                               *metal3api.BareMetalHost
+		UserDataSecret                     *corev1.Secret
+		PreprovNetworkDataSecret           *corev1.Secret
+		NetworkDataSecret                  *corev1.Secret
+		ExpectedUserData                   string
+		ErrUserData                        bool
+		ExpectedNetworkData                string
+		ErrNetworkData                     bool
+		ExpectedPreprovisioningNetworkData string
+		ErrPreprovisioningNetworkData      bool
+		ExpectedMetaData                   string
+		ErrMetaData                        bool
 	}{
 		{
 			Scenario: "host with user data only",
@@ -151,11 +153,12 @@ func TestProvisionWithHostConfig(t *testing.T) {
 					},
 					PreprovisioningNetworkDataName: "net-data",
 				}),
-			NetworkDataSecret:   newSecret("net-data", map[string]string{"networkData": "key: value"}),
-			ExpectedUserData:    "",
-			ErrUserData:         false,
-			ExpectedNetworkData: base64.StdEncoding.EncodeToString([]byte("key: value")),
-			ErrNetworkData:      false,
+			NetworkDataSecret:                  newSecret("net-data", map[string]string{"networkData": "key: value"}),
+			ExpectedUserData:                   "",
+			ErrUserData:                        false,
+			ExpectedNetworkData:                base64.StdEncoding.EncodeToString([]byte("key: value")),
+			ExpectedPreprovisioningNetworkData: base64.StdEncoding.EncodeToString([]byte("key: value")),
+			ErrNetworkData:                     false,
 		},
 		{
 			Scenario: "host with preprov and regular network data",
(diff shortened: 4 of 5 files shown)


hardys:
@dtantsur @Rozzii - as discussed on slack - in my environment this allows me to specify a secret with network_data.json contents via `preprovisioningNetworkDataName`, then retrieve the `network_data.json` inside the ramdisk from the attached virtual media device.

Also @zaneb if you could take a look that would be much appreciated, thanks!

pkg/provisioner/ironic/ironic.go
```diff
@@ -456,6 +456,20 @@ func (p *ironicProvisioner) ValidateManagementAccess(data provisioner.Management
...
 
+	// If no PreprovisioningImage builder is enabled we set the Node network_data
+	// this enables Ironic to inject the network_data into the ramdisk image
+	if !p.config.havePreprovImgBuilder {
+		p.log.Info("adding preprovisioning network_data for node", "node", ironicNode.UUID)
+		networkDataRaw := data.PreprovisioningNetworkData
+		if networkDataRaw != "" {
+			var networkData map[string]interface{}
+			if err := yaml.Unmarshal([]byte(networkDataRaw), &networkData); err != nil {
+				p.log.Error(err, "failed to unmarshal network_data.json from PreprovisioningNetworkData")
```
hardys, line 467: Should we return `operationFailed` here or `transientError` since the secret contents may get updated?

controllers/metal3.io/host_config_data.go
```diff
@@ -94,6 +94,31 @@ func (hcd *hostConfigData) NetworkData() (string, error) {
 	return networkDataRaw, err
 }
 
+// PreprovisioningNetworkData get preprovisioning network configuration
+func (hcd *hostConfigData) PreprovisioningNetworkData() (string, error) {
+	if hcd.host.Spec.PreprovisioningNetworkDataName == "" {
+		hcd.log.Info("PreprovisioningNetworkDataName is not set, returning empty data")
```
dtantsur, line 100: This may be useful for your debugging now, but it will be recorded on every reconcile. BMO logs are already very chatty, I suggest removing this.

controllers/metal3.io/host_config_data.go
```diff
@@ -94,6 +94,31 @@ func (hcd *hostConfigData) NetworkData() (string, error) {
...
+		return "", nil
+	}
+	networkData := &corev1.SecretReference{
```
dtantsur, line 103: nit: it looks like you don't use the SecretReference itself, only its fields?

pkg/provisioner/ironic/ironic.go
```diff
@@ -456,6 +456,20 @@ func (p *ironicProvisioner) ValidateManagementAccess(data provisioner.Management
 		// below.
 	}
 
+	// If no PreprovisioningImage builder is enabled we set the Node network_data
+	// this enables Ironic to inject the network_data into the ramdisk image
+	if !p.config.havePreprovImgBuilder {
+		p.log.Info("adding preprovisioning network_data for node", "node", ironicNode.UUID)
```
dtantsur, line 462: Let's only log if we're actually changing things.

pkg/provisioner/ironic/ironic.go
dtantsur: That's not the topic I'm strong at... I guess we'll get another reconcile if it gets updated (assuming we listen to these events).

pkg/provisioner/ironic/ironic.go
```diff
@@ -456,6 +456,20 @@ func (p *ironicProvisioner) ValidateManagementAccess(data provisioner.Management
...
+		networkDataRaw := data.PreprovisioningNetworkData
+		if networkDataRaw != "" {
+			var networkData map[string]interface{}
```
dtantsur, line 465: nit: you could probably use `json.RawMessage` since we don't care about the exact format

pkg/provisioner/ironic/ironic.go
hardys: Yeah that's what I wasn't sure of, I know we'll get another reconcile if the BMH changes but I wasn't sure if it's only the content of a referenced secret - I'll test to confirm

controllers/metal3.io/host_config_data.go
hardys: Ah good point thanks - it's a copy/paste from NetworkData above which does have to handle SecretReference, will remove!

controllers/metal3.io/host_config_data.go
hardys: It's actually something we already log for all the other *Data functions, but I agree - will remove and propose a follow-up to remove the others 👍

pkg/provisioner/ironic/ironic.go
zaneb: We'll get another reconcile if the Secret changes because of https://github.com/metal3-io/baremetal-operator/blob/main/controllers/metal3.io/baremetalhost_controller.go#L1741, provided that we own the Secret, which is set up by https://github.com/metal3-io/baremetal-operator/blob/main/controllers/metal3.io/baremetalhost_controller.go#L648 - and note that this is currently only called for the preprovisioningNetworkData after we see it show up in the PreprovisioningImage Status. So we if want to use it all the time then we should call AcquireSecret somewhere unconditional.

That said, we don't have great ways of handling errors in general when the error is user configuration. The validating webhook is the best place to catch those. Also, this should be logged at info, not error level.
What would we be hoping to achieve here by returning early? Will ironic reject the update if the data is not yaml? If not I'd advocate that we continue to do what the patch currently does - pass whatever garbage the user gave us on to ironic and carry on.

controllers/metal3.io/host_config_data.go
zaneb: The other ones are not called in every state is the difference I think.

zaneb (commented):
In principle, I'm in favour of this. The idea behind preprovisioningNetworkData is that it is the default for the networkData during provisioning, and the networkData itself prior to provisioning, so this would be consistent with that.

However, there is an issue with this implementation - during provisioning at least, we will bounce ironic back and forth between the preprovisioningNetworkData and the actual networkData (assuming they are different) on every reconcile. It'll set the former during ValidateManagementAccess() and switch back to the latter immediately afterwards during Provision(). I'd assume this carries a risk of races.

We could restrict the setting of the preprovisioningNetworkData to certain BMO states (presumably Inspecting and Deprovisioning), but I think we'd still have an issue where during provisioning if we have to boot IPA again then it might get different network data than if we don't. I think to get consistent behaviour we have to do this only when moving the ironic node to manageable (from either enrolled, clean failed, inspect failed, adopt failed, or available) or deleting.

If the ironic provisioner is making the decision of whether it needs the preprovisiningNetworkData, it may also pay to let it call the function to retrieve it when required (similar to provisioner.HostConfigData) rather than calling it unconditionally from the controller.

pkg/provisioner/ironic/ironic.go
hardys: > Will ironic reject the update if the data is not yaml? 

It [validates it against the network_data.json schema](https://opendev.org/openstack/ironic/src/branch/master/ironic/api/controllers/v1/node.py#L394) so the format has to be exactly right or we'll get an error on node registration from Ironic

I discovered this while testing so decided to return early here in an attempt to catch obviously incorrect secret contents, it also matches what we do on [deployment for the config-drive network_data](https://github.com/metal3-io/baremetal-operator/blob/main/pkg/provisioner/ironic/ironic.go#L1477)

hardys:
> However, there is an issue with this implementation - during provisioning at least, we will bounce ironic back and forth between the preprovisioningNetworkData and the actual networkData (assuming they are different) on every reconcile. It'll set the former during ValidateManagementAccess() and switch back to the latter immediately afterwards during Provision(). I'd assume this carries a risk of races.

I'm not sure this is correct - the code I'm adding here sets the [Node network_data ](https://docs.openstack.org/api-ref/baremetal/#create-node)(which not currently set at all by BMO) - @dtantsur can confirm but IIUC this API is only used to provide network configuration to the IPA Ramdisk pre-provisioning (so it makes sense to only consume `preprovisioningNetworkData` here)

Then on Provision we'll use the existing code, which creates a config drive with the `networkData` which is passed to the [Ironic API](https://docs.openstack.org/api-ref/baremetal/#change-node-provision-state) when we change the provision state - this data can potentially be different to the preprovisioningNetworkData AFAICS (although it's not a scenario I've tested)

Am I missing something? By setting currently-unset the Node network_data, I don't see how we can introduce a race with the config-drive data?

controllers/metal3.io/host_config_data.go
```diff
@@ -94,6 +94,26 @@ func (hcd *hostConfigData) NetworkData() (string, error) {
...
+	}
+	networkDataRaw, err := hcd.getSecretData(
+		hcd.host.Spec.PreprovisioningNetworkDataName,
+		hcd.host.Namespace,
+		"networkData",
+	)
+	if err != nil {
+		_, isNoDataErr := err.(NoDataInSecretError)
+		if isNoDataErr {
+			hcd.log.Info("PreprovisioningNetworkData secret key is not set, returning empty data")
```
zaneb, line 110: nit: it would be good to specify what the key is (`networkData`)

pkg/provisioner/ironic/ironic.go
zaneb: OK, we're returning transientError in the config-drive case, so there's precedent for that even if it is not really the right thing.

zaneb:
> I'm not sure this is correct - the code I'm adding here sets the [Node network_data ](https://docs.openstack.org/api-ref/baremetal/#create-node)(which is not currently set at all by BMO) - @dtantsur can confirm but IIUC this API is only used to provide network configuration to the IPA Ramdisk pre-provisioning (so it makes sense to only consume preprovisioningNetworkData here)

Oh, sorry, you're right. I assumed that this was how we were passing the networkData today, but evidently that's set in the ConfigDrive instead.

pkg/provisioner/ironic/ironic.go
hardys: Ok I went with `operationFailed` in the latest revision but will adjust to `transientError` since it sounds like that will work without adjusting the secret ownership.

Related to that, I thought of a disadvantage of adjusting the network data ownership - that will result in behaviour similar to https://github.com/metal3-io/baremetal-operator/issues/1347 right?  Which might be desirable in the static-ip case (one secret per BMH) but probably not in the case where there's a config applied per cluster or group of hosts (non-default vlan, bonding etc)

zaneb: > Ok I went with `operationFailed` in the latest revision but will adjust to `transientError` since it sounds like that will work without adjusting the secret ownership.

You'll want to adjust it anyway for the cases where there is _not_ an error.

> Related to that, I thought of a disadvantage of adjusting the network data ownership - that will result in behaviour similar to #1347 right?

Yes.

> Which might be desirable in the static-ip case (one secret per BMH) but probably not in the case where there's a config applied per cluster or group of hosts (non-default vlan, bonding etc)

If there's multiple owners, the Secret doesn't get deleted until the last owner is. I think it's fine. This is consistent with how e.g. networkData works, and with how preprovisioningNetworkData works when the PreprovisioningImage controller is enabled.

Rozzii (approved):
As we have discussed it prior outside this PR, I am in favor  of this approach, via custom patches this is already way of using "preprovisioningNetworkData" in some downstream organizations as far as I have heard. Thank you @hardys for implementing this for the upstream repo! I can't really give a confident opinion about the go code, as far as I can tell it is fine I don't see really any inconsistencies but I am not that familiar with the BMO code. 

In any case from my POV it LGTM!

pkg/provisioner/ironic/ironic.go
```diff
@@ -456,6 +456,25 @@ func (p *ironicProvisioner) ValidateManagementAccess(data provisioner.Management
...
 
+	// If no PreprovisioningImage builder is enabled we set the Node network_data
+	// this enables Ironic to inject the network_data into the ramdisk image
+	if !p.config.havePreprovImgBuilder {
+		networkDataRaw := data.PreprovisioningNetworkData
+		if networkDataRaw != "" {
+			var networkData map[string]interface{}
+			if yamlErr := yaml.Unmarshal([]byte(networkDataRaw), &networkData); yamlErr != nil {
+				p.log.Info("failed to unmarshal networkData from PreprovisioningNetworkData")
+				result, err = transientError(errors.Wrap(err, "invalid preprovisioningNetworkData"))
```
dtantsur, line 467: nit: use the standard `fmt.Errorf` instead of `errors.Wrap`

pkg/provisioner/ironic/ironic.go
```diff
@@ -456,6 +456,25 @@ func (p *ironicProvisioner) ValidateManagementAccess(data provisioner.Management
...
+				return
+			}
+			numUpdates := len(updater.Updates)
+			updater.SetTopLevelOpt("network_data", networkData, ironicNode.NetworkData)
+			if len(updater.Updates) != numUpdates {
```
dtantsur, line 472: nit: maybe we should change SetTopLevelOpt to return a boolean (updated or not)

hardys: Yes I thought the same - I can perhaps look at a follow-up which refactors things to do that, but it seemed outside of the scope of this PR.

pkg/provisioner/ironic/ironic.go
hardys: Thanks, copy-paste mistake, I also forgot to change `err` to `yamlErr` - now fixed!

Rozzii (approved):
/lgtm

controllers/metal3.io/host_config_data.go
```diff
@@ -94,6 +94,26 @@ func (hcd *hostConfigData) NetworkData() (string, error) {
 	return networkDataRaw, err
 }
 
+// PreprovisioningNetworkData get preprovisioning network configuration
+func (hcd *hostConfigData) PreprovisioningNetworkData() (string, error) {
+	if hcd.host.Spec.PreprovisioningNetworkDataName == "" {
+		return "", nil
+	}
+	networkDataRaw, err := hcd.getSecretData(
```
zaneb, line 102: I did some more digging into the difference between `ObtainSecret()` (used here) and `AcquireSecret()`, after having got it wrong several times :smile: 

Both add a label to the Secret so that it gets watched and cached.
Only `AcquireSecret()` adds an owner reference so that the BMH gets reconciled if the Secret changes.

In the case of network/user/metaData, what matters is the value at a point in time, when you start provisioning. It's the start of provisioning that kicks off the reconcile, so there's no need to re-reconcile when the Secret changes.

There's a strong argument that preprovisioningNetworkData is the same, and what matters is the value at the point in time where you start inspecting/preparing/deprovisioning.
There's probably another argument that since ironic may reboot the node in response to its own internal events (not the BMO), then we should always provide it with the most up-to-date data we know of. In practice we will get there eventually because we always requeue every few minutes at most, but there may be situations where something unintuitive happens.

This would be interesting to discuss further but not worth blocking on.

zaneb (commented):
/lgtm

PR #1386, opened 2023-10-10
🐛 Skip power off BMH if registration error OR no credentials

### LINKED ISSUE #1385: BMH with good BMC credentials and registration error stuck deleting
Seen in https://github.com/metal3-io/baremetal-operator/pull/1382
Still need to verify the reproduction steps, but this is what I believe happened:

1. A BMH (`first`) in `available` state was deleted without waiting for it to go away. Tthe important part is that it is still there, not that it was deleting. Just mentioning for completeness sake.)
2. Another BMH (`second`) was created with identical BMC details (it is the same machine, just "re-creating" the BMH).
3. `second` hits registration error since there is a node with identical MAC in Ironic already.
4. `second` is marked for deletion (cleaning up after the failed test).
5. `first` disappears eventually when the deletion is completed.
6. **`second` remains indefinitely** stuck trying to power off before deletion.

I believe this is due to a simple logic fail here:
https://github.com/metal3-io/baremetal-operator/blob/c4936230d2e3bb5f136166852534760b56f523c8/controllers/metal3.io/host_state_machine.go#L588-L592

The situation I described above have a BMH with `NeedsRegistration=true` (because registration failed due to conflict) and `haveCreds=true`. We get `true && false` so we do not `skipToDelete`. It is impossible to power off the BMH since it was not registered but this is what is attempted. I think the proper check would be `NeedsRegistration || !haveCreds`. In either of these cases we cannot power off the BMH and should just skip to deletion.

The BMH with registration error should eventually be deleted.

I'm planning to open a PR with the suggested fix.

**Environment:**

- Baremetal Operator version: main
- Environment (metal3-dev-env or other): other (bmo e2e)

/kind bug

lentzi90:
If a BMH is not registered OR we do not have proper credentials for it, we cannot power off before deleting. Therefore we should skip to deletion in either of these cases, not just when both hold.

If a BMH is not registered OR we do not have proper credentials for it, we cannot power off before deleting. Therefore we should skip to deletion in either of these cases, not just when both hold.

Fixes #1385

### DIFF
--- a/controllers/metal3.io/host_state_machine.go
+++ b/controllers/metal3.io/host_state_machine.go
@@ -585,8 +585,8 @@ func (hsm *hostStateMachine) handlePoweringOffBeforeDelete(info *reconcileInfo) 
 			return skipToDelete()
 		}
 	case actionError:
-		if r.NeedsRegistration() && !hsm.haveCreds {
-			// If the host is not registered as a node in Ironic and we
+		if r.NeedsRegistration() || !hsm.haveCreds {
+			// If the host is not registered as a node in Ironic or we
 			// lack the credentials to power it off, just continue to
 			// delete.
 			return skipToDelete()


honza:
There is already a PR for this, and a good discussion: would you mind having a look? #1356

lentzi90:
/hold
Looks like I'm just duplicating #1356.

lentzi90:
/hold cancel
#1356 is merged, but it did not touch this particular line.
I *think* it may be impossible to get into the specific case that this would fix with the changes in #1356. However, I also think the logic is wrong and we should therefore change it.

kashifest:
@honza @zaneb wdyt about this patch. This looks like a bug to me as well.

zaneb:
It was correct before. If we are getting an error that we are not registered _and_ we don't have credentials, then we should ignore the error because it is never getting better.
If we are getting an error and it's not about needing registration, then we report the error.
(And if we are getting an error that we are not registered and we _do_ have credentials, then we need to handle the error so we will requeue and do the registration again.)

/close

lentzi90:
@zaneb can you please explain to me why we should try to register again if we are already in `handlePoweringOffBeforeDelete`? It makes no sense to me.

I imagine this chain of events:

1. BMH and BMC credentials are created
2. The BMC credentials are bad (but they do exist) or there is something fundamentally wrong with the server
3. Registration error.
4. User decides to just start over, so deletes the BMH
5. The BMH cannot be powered off, cannot be registered and is stuck retrying this until the user provides working credentials or fixes the server so it can be registered.

zaneb:
If you look at the code for `ReconcileState()`, registration happens in [`ensureRegistered()`](https://github.com/metal3-io/baremetal-operator/blob/main/controllers/metal3.io/host_state_machine.go#L196), _before_ the [state handler is called](https://github.com/metal3-io/baremetal-operator/blob/main/controllers/metal3.io/host_state_machine.go#L201). So the only ways to get to the state handler are if registration succeeds, or we [hit the exception for not having any credentials](https://github.com/metal3-io/baremetal-operator/blob/main/controllers/metal3.io/host_state_machine.go#L326-L331).

If we get into the state handler and the node [is not registered in ironic](https://github.com/metal3-io/baremetal-operator/blob/main/pkg/provisioner/ironic/ironic.go#L177), it means that the ironic DB has been lost between the time we did the registration and the time we tried to handle the current state. The solution to this is to requeue so that next time around `ensureRegistered()` will re-register the node. We do this by propagating the error, which will cause controller-runtime to requeue.

The only exception to that when we find the node not registered in ironic is if `haveCreds` is false, in which case we haven't attempted registration and trying again won't help because we won't attempt it next time either. In _that_ case we proceed to deletion. This is the case that the condition on line 588 is checking for.

I think this is confusing because the 'registered' in `ensureRegistered()` means 'ironic is successfully talking to the BMC', but the 'registered' in `ErrNeedsRegistration`/`NeedsRegistration()` means 'a Node object exists in Ironic'. They're linked because the same code does both, but maybe this could have been more understandable. In my defense, everything was much worse before the state machine existed :smile: 

The case where credentials exist but are not sufficient to get us talking to the BMC is handled by #1356 - we try 3 times (note: each attempt may take a number of reconciles) and if we keep getting `actionFailed` (note: not `actionError`, which is a transient error) then we proceed to deletion.

Issue #1371, opened 2023-09-28
Add e2e test for basic operations


lentzi90:
Add an e2e test for power off/on and [reboot](https://book.metal3.io/bmo/reboot_annotation).

PR #1389, opened 2023-10-11
🌱 Add e2e basic operations test

mboukhalfa:
Adds basic operations e2e test: reboot and power off/on.
Import capm3 e2e test to reuse public functions.

fixes: #1371


test/e2e/basic_ops_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+		namespace, cancelWatches = framework.CreateNamespaceAndWatchEvents(ctx, framework.CreateNamespaceAndWatchEventsInput{
+			Creator:   clusterProxy.GetClient(),
+			ClientSet: clusterProxy.GetClientSet(),
+			Name:      fmt.Sprintf("%s-%s", specName, util.RandomString(6)),
+			LogFolder: artifactFolder,
+		})
+	})
+
+	It("should power cycle a newly created BMH", func() {
```
lentzi90, line 50: Sounds a bit like every new BMH should be rebooted automatically. Maybe better like this?
```suggestion
	It("should control power cycle of BMH though annotations", func() {
```

test/e2e/basic_ops_test.go
```diff
@@ -0,0 +1,123 @@
...
+		}, e2eConfig.GetIntervals("inspection", "wait-registering")...)
+
+		By("waiting for the BMH to become available")
+		WaitForBmhInState(ctx, WaitForBmhInStateInput{
+			Client: clusterProxy.GetClient(),
+			Bmh:    bmh,
+			State:  metal3api.StateAvailable,
+		}, e2eConfig.GetIntervals("inspection", "wait-available")...)
+
+		By("testing the reboot BMH")
```
lentzi90, line 101:
```suggestion
		By("setting the reboot annotation and checking that the BMH was rebooted")
```

test/e2e/basic_ops_test.go
```diff
@@ -0,0 +1,123 @@
...
+		capm3_e2e.AnnotateBmh(ctx, clusterProxy.GetClient(), bmh, rebootAnnotation, pointer.String("{\"force\": true}"))
+		waitForVmsState([]string{vmName}, shutoff, e2eConfig.GetIntervals(specName, "wait-vm-state")...)
+		waitForVmsState([]string{vmName}, running, e2eConfig.GetIntervals(specName, "wait-vm-state")...)
+
+		By("checking the power off BMH")
```
lentzi90, line 106:
```suggestion
		By("setting the power off annotation on the BMH and checking that it worked")
```

test/e2e/basic_ops_test.go
```diff
@@ -0,0 +1,123 @@
...
+		capm3_e2e.AnnotateBmh(ctx, clusterProxy.GetClient(), bmh, poweroffAnnotation, pointer.String("{\"force\": true}"))
+		waitForVmsState([]string{vmName}, shutoff, e2eConfig.GetIntervals(specName, "wait-vm-state")...)
+
+		// power on
+		By("checking the power on BMH")
```
lentzi90, line 111:
```suggestion
		By("removing the power off annotation and checking that the BMH powers on")
```

test/e2e/config/ironic.yaml
```diff
@@ -26,3 +26,4 @@ intervals:
   inspection/wait-available: ["10m", "10s"]
   default/wait-deployment: ["5m", "1s"]
   ironic/wait-deployment: ["10m", "2s"]
+  default/wait-vm-state: ["20m", "100ms"]
```
lentzi90, line 29: This seems a bit long? The whole test barely takes 20 minutes :smile:

test/e2e/basic_ops_test.go
```diff
@@ -0,0 +1,123 @@
...
+		WaitForBmhInState(ctx, WaitForBmhInStateInput{
+			Client: clusterProxy.GetClient(),
+			Bmh:    bmh,
+			State:  metal3api.StateAvailable,
+		}, e2eConfig.GetIntervals("inspection", "wait-available")...)
+
+		By("testing the reboot BMH")
+		capm3_e2e.AnnotateBmh(ctx, clusterProxy.GetClient(), bmh, rebootAnnotation, pointer.String("{\"force\": true}"))
+		waitForVmsState([]string{vmName}, shutoff, e2eConfig.GetIntervals(specName, "wait-vm-state")...)
+		waitForVmsState([]string{vmName}, running, e2eConfig.GetIntervals(specName, "wait-vm-state")...)
```
lentzi90, line 104: This is a good way to ensure that it is really doing what we think. However, it is also impossible to do with real hardware. I was hoping we could do these e2e tests so that they can run also against real hardware in various configurations.
Do you think there is any chance that we could check these things in a different way? Maybe [PoweredOn](https://github.com/metal3-io/baremetal-operator/blob/7c007b6ae4efe36d2f313a0573d45804b03954b0/apis/metal3.io/v1alpha1/baremetalhost_types.go#L790) could work if the reboot is not too fast?

lentzi90 (commented):
Excellent work @mboukhalfa !
Added a few nits below, and then one concern: I would like to avoid relying on `virsh` or similar in the tests. This makes it impossible to reuse them in other contexts. Do you think there is some way we could avoid it?

test/e2e/config/ironic.yaml
```diff
@@ -24,14 +24,15 @@ variables:
   BMC_PASSWORD: password
   BMC_ADDRESS: ipmi://192.168.222.1:16230
   BOOT_MAC_ADDRESS: "00:60:2f:31:81:01"
-
+  VM_NAME: "bmo-e2e-0"
```
lentzi90, line 27: I think we can remove this now, right?

lentzi90 (commented):
Awesome!
/approve

lentzi90 (commented):
/lgtm

test/e2e/basic_ops_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+		By("waiting for the BMH to become available")
+		WaitForBmhInProvisioningState(ctx, WaitForBmhInProvisioningStateInput{
+			Client: clusterProxy.GetClient(),
+			Bmh:    bmh,
+			State:  metal3api.StateAvailable,
+		}, e2eConfig.GetIntervals(specName, "wait-available")...)
+
+		By("setting the reboot annotation and checking that the BMH was rebooted")
+		capm3_e2e.AnnotateBmh(ctx, clusterProxy.GetClient(), bmh, rebootAnnotation, pointer.String("{\"force\": true}"))
```
kashifest, line 100: Shouldn't this `AnnotateBMH()` function reside here and capm3 should import it rather.

kashifest (commented):
/approve
One question in line.
/hold

test/e2e/basic_ops_test.go
lentzi90: That is a good point. It would probably make sense to move some of the e2e code around BMHs from CAPM3 here.
For me it does not matter much when/how we do it though. I'm fine with doing a refactor PR after implementing the tests for example. This also makes it easier to review the new code we add :slightly_smiling_face:

kashifest: ok with refactor later.
/hold cancel

PR #1394, opened 2023-10-12
🌱 Add golangci-lint github workflow

kashifest:
This PR will add golangci-lint github workflow.  Currently only the following linters are enabled

    - asciicheck
    - bodyclose
    - dogsled
    - exportloopref

The plan is to enable the rest in batches to ease the code changes and review.


.golangci.yaml
```diff
@@ -0,0 +1,160 @@
...
+    # - predeclared
+    # - revive
+    # - rowserrcheck
+    # - staticcheck
+    # - stylecheck
+    # - thelper
+    # - typecheck
+    # - unconvert
+    # - unused
+    # - whitespace
```
kashifest, line 42: These linters are commented out since the changes required in the code are overwhelming. I plan to enable them in batches so that its easy to change and also easy to review.

tuminoid: Agreed. There are years of cruft to be fixed, easier to do and also to review in batches.

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,35 @@
...
+    name: lint
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        working-directory:
+          - ""
+          - test
+          - apis
+          - pkg/hardwareutils
```
tuminoid, line 21: Should we have the `if:` conditional here too like in other workflows?

tuminoid (commented):
One question, otherwise looking good to me.

.github/workflows/golangci-lint.yml
lentzi90: I think not. This workflow should work well also in forks, right?

kashifest: Good question, haven't thought about it, but I am leaning towards @lentzi90's comment that it should work with forks also. So we can keep it as it is now.

tuminoid: It doesn't have anything metal3 specific like uploads etc, so it should be fine.

tuminoid (commented):
/lgtm

lentzi90 (commented):
/lgtm
Great job!

tuminoid:
/assign @zaneb  @dtantsur 
Can we get an approval for this before it rots? There is years of dirt to fix as the old linter skipped most of the code.

PR #1382, opened 2023-10-09
🌱 Add e2e external inspection test

lentzi90:
**What this PR does / why we need it**:

Adds an e2e test for [external inspection](https://book.metal3.io/bmo/external_inspection).
The test simply creates a BMH with annotations for disabling inspection and for adding hardware details. Then it checks that the BMH becomes available without going through inspection.

This also adds cert-manager images to the e2e config so that if they happen to be available locally, then they can be loaded into the kind cluster (if used). It helps save some time when running the test repeatedly.

Minor details:
- Don't exit on e2e test failure, gather artifacts/logs first
- Destroy the network before trying to undefine it
- Wait for namespace deletion between e2e tests since we are using the
  same VM for the BMHs in multiple tests

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #1370

PR #1402, opened 2023-10-24
🌱 Re-inspection by annotation test added

### LINKED ISSUE #1369: Add inspection annotation e2e test
Add an e2e test that tests the [inspection annotation](https://book.metal3.io/bmo/inspect_annotation). Adding the annotation should trigger an inspection even if the BMH was inspected earlier.

adilGhaffarDev:

Fixes #1369

### DIFF
--- a/test/e2e/config/fixture.yaml
+++ b/test/e2e/config/fixture.yaml
@@ -15,6 +15,7 @@ variables:
   DEPLOY_BMO: "true"
   DEPLOY_CERT_MANAGER: "true"
   BMO_KUSTOMIZATION: "../../config/overlays/fixture"
+  EXPECTED_HOST_NAME: ""
   # Test credentials. The tests will create a BMH with these.
   # For the fixture provider it does not matter much what we put here.
   BMC_USER: admin
--- a/test/e2e/re_inspection_test.go
+++ b/test/e2e/re_inspection_test.go
@@ -1,0 +1,137 @@
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/pointer"
+	"sigs.k8s.io/cluster-api/test/framework"
+	"sigs.k8s.io/cluster-api/util"
+
+	capm3_e2e "github.com/metal3-io/cluster-api-provider-metal3/test/e2e"
+
+	metal3api "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
+)
+
+var _ = Describe("Re-Inspection", func() {
+	var (
+		specName         = "re-inspection"
+		namespace        *corev1.Namespace
+		cancelWatches    context.CancelFunc
+		bmcUser          string
+		bmcPassword      string
+		bmcAddress       string
+		bootMacAddress   string
+		expectedHostName string
+	)
+	const (
+		wrongHostName = "wrongHostName"
+	)
+	BeforeEach(func() {
+		bmcUser = e2eConfig.GetVariable("BMC_USER")
+		bmcPassword = e2eConfig.GetVariable("BMC_PASSWORD")
+		bmcAddress = e2eConfig.GetVariable("BMC_ADDRESS")
+		bootMacAddress = e2eConfig.GetVariable("BOOT_MAC_ADDRESS")
+		expectedHostName = e2eConfig.GetVariable("EXPECTED_HOST_NAME")
+
+		namespace, cancelWatches = framework.CreateNamespaceAndWatchEvents(ctx, framework.CreateNamespaceAndWatchEventsInput{
+			Creator:   clusterProxy.GetClient(),
+			ClientSet: clusterProxy.GetClientSet(),
+			Name:      fmt.Sprintf("%s-%s", specName, util.RandomString(6)),
+			LogFolder: artifactFolder,
+		})
+	})
+
+	It("should re-inspect the annotated BMH", func() {
+		By("creating a secret with BMH credentials")
+		bmcCredentials := corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "bmc-credentials",
+				Namespace: namespace.Name,
+			},
+			StringData: map[string]string{
+				"username": bmcUser,
+				"password": bmcPassword,
+			},
+		}
+		err := clusterProxy.GetClient().Create(ctx, &bmcCredentials)
+		Expect(err).NotTo(HaveOccurred())
+
+		By("creating a BMH with inspection disabled and hardware details added with wrong HostName")
+		newHardwareDetails := strings.Replace(hardwareDetails, "localhost.localdomain", wrongHostName, 1)
+		bmh := metal3api.BareMetalHost{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      specName + "-reinspect",
+				Namespace: namespace.Name,
+				Annotations: map[string]string{
+					metal3api.InspectAnnotationPrefix:   "disabled",
+					metal3api.HardwareDetailsAnnotation: newHardwareDetails,
+				},
+			},
+			Spec: metal3api.BareMetalHostSpec{
+				BMC: metal3api.BMCDetails{
+					Address:         bmcAddress,
+					CredentialsName: "bmc-credentials",
+				},
+				BootMode:       metal3api.Legacy,
+				BootMACAddress: bootMacAddress,
+			},
+		}
+		err = clusterProxy.GetClient().Create(ctx, &bmh)
+		Expect(err).NotTo(HaveOccurred())
+
+		By("waiting for the BMH to be in registering state")
+		WaitForBmhInProvisioningState(ctx, WaitForBmhInProvisioningStateInput{
+			Client: clusterProxy.GetClient(),
+			Bmh:    bmh,
+			State:  metal3api.StateRegistering,
+		}, e2eConfig.GetIntervals(specName, "wait-registering")...)
+
+		By("waiting for the BMH to become available")
+		WaitForBmhInProvisioningState(ctx, WaitForBmhInProvisioningStateInput{
+			Client: clusterProxy.GetClient(),
+			Bmh:    bmh,
+			State:  metal3api.StateAvailable,
+		}, e2eConfig.GetIntervals(specName, "wait-available")...)
+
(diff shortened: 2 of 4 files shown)


test/e2e/inspection_test.go
```diff
@@ -145,6 +146,84 @@ var _ = Describe("Inspection", func() {
...
+			Bmh:    bmh,
+			State:  metal3api.StateInspecting,
+		}, e2eConfig.GetIntervals(specName, "wait-inspecting")...)
+
+		By("waiting for the BMH to become available")
+		WaitForBmhInProvisioningState(ctx, WaitForBmhInProvisioningStateInput{
+			Client: clusterProxy.GetClient(),
+			Bmh:    bmh,
+			State:  metal3api.StateAvailable,
+		}, e2eConfig.GetIntervals(specName, "wait-available")...)
```
hardys, line 201: Similar to the [discussion](https://github.com/metal3-io/baremetal-operator/pull/1382#discussion_r1350129525) in #1382 this has the issue that we're only proving the state-machine path, not that inspection actually worked (e.g `bmh.Status.HardwareDetails` was updated).

As a first step I wonder if we could add a common assertion that `bmh.Status.HardwareDetails` is not empty after each inspection?

Ideally we'd change something about the environment e.g add some device, reinspect then prove it exists in the re-inspected HardwareDetails, but I appreciate that may be difficult and goes beyond the scope of adding a new test-case.

/cc @lentzi90 and @dtantsur for additional input on this suggestion.

lentzi90: Good point. What do you think about using external inspection to skip inspection initially? Then we can provide the input and make sure that it is a bit different from what an inspection would result it. That way we get a clear diff to check when we trigger the inspection using the annotation :slightly_smiling_face: 
Bonus: We save some time by not waiting for the first inspection :hourglass:

adilGhaffarDev: in e2e we have minkube and VMs, I am not sure how we can change VM from our e2e test, if we can do that we replicate the ideal scenario that @hardys has explained.

lentzi90: Sorry, I'm confusing you. I mean do an [external inspection](https://github.com/metal3-io/baremetal-operator/blob/main/test/e2e/external_inspection_test.go) where we can actually put any data we want. In the link you can see the hardware details we use for the external inspection test. We can use the same here but make some changes to it first (it is just a string).
Then when we trigger the inspection with the annotation, IPA will find out what the actual hardware details are and correct them. So we can check that there is a diff before and after the inspection.

hardys: @lentzi90 that could work (and such a scenario was actually [discussed](https://github.com/metal3-io/baremetal-operator/pull/796/files#r586241286) when I implemented the hardwaredetails annotation)

My only concern with that is it's not actually testing re-inspection, instead it's testing moving from inspection-disabled -> inspection (which is also a valid but probably less common scenario for users?) - this means we won't catch e.g any issues with Ironic that could arise from re-inspecting an already inspected host.

As an alternative we could assert HardwareDetails is correct after the first inspect, then keep track of the `operationHistory` timestamps, but unless we forcibly modify the Status.HardwareDetails between inspection runs it's going to be difficult to validate the re-inspection data (I'm assuming we can't modify the VM configuration from the test).

lentzi90: Hmm interesting perspective. I'm not familiar with how different the external inspection and normal inspection are treated internally. Are they very different? My impression was that the sequence is more like inspection-disabled -> external inspection -> BMH is now in available state (inspection is not disabled at this point) -> inspect "again" -> BMH is again in available state.
Is there really a difference in the externally inspected and the "normally" inspected states?

hardys: @lentzi90 the code-path through the BMO is completely different in the external/disabled case:

- We consume the hardwaredetails when inspection is disabled in the [reconcile loop](https://github.com/metal3-io/baremetal-operator/blob/main/controllers/metal3.io/baremetalhost_controller.go#L146)
- This also means we short-circuit the Inspecting state to [do nothing](https://github.com/metal3-io/baremetal-operator/blob/main/controllers/metal3.io/baremetalhost_controller.go#L936) (so we don't drive Ironic through any of the normal inspection process)

In this flow the BMH reaches available state and inspection is still disabled (only the hardwaredetails annotation is [removed on consumption I](https://github.com/metal3-io/baremetal-operator/blob/main/controllers/metal3.io/baremetalhost_controller.go#L283)IRC)

lentzi90: Ok that part I get, what I don't understand is how it in any way would affect the actual inspection that is then triggered.
We *do* check the normal inspection flow in a separate test, so I'm here only interested in checking the functionality of the inspection annotation.

Is the "second" inspection taking into account the existing state in any way? I would expect an externally inspected BMH to be indistinguishable from a normally inspected BMH and that they behave in exactly the same way in every situation.

The annotation that disables inspection is removed/overwritten with the annotation that triggers the inspection, this is a good point. Perhaps we should remove it first and ensure that no inspection is started just from that. Then we can add the annotation a bit later to trigger the inspection.

Perhaps the title of the PR is what is causing this confusion. My intention with https://github.com/metal3-io/baremetal-operator/issues/1369 was that we should test the functionality described here: https://book.metal3.io/bmo/inspect_annotation
I don't think the it is relevant in this case how the BMH ended up in available state with some (potentially) outdated hardware details. The only important part is that adding an annotation to it will trigger an inspection that causes the hardware details to be updated.

hardys: @lentzi90 the state in Ironic will be different if we use external/disabled vs "real" inspection initially, but perhaps that doesn't matter if the main aim is to test the annotation (and we have good confidence that Ironic behaves correctly when inspection is triggered twice since that's the likely scenario in the "hardware changed" use-case described for the annotation)

lentzi90: Got it, thanks for explaining this!
It sounds to me like we would be fine with an external inspection for now, and this should allow a pretty simple test where the hardware details actually change. In the CAPM3 e2e tests I know we test re-inspection properly, and perhaps we should add it here also at some point also. I would be happy to hear about use-cases or more reasons for testing this!

adilGhaffarDev: I have changed the test @hardys @lentzi90 please take a look and let me know if it looks reasonable. Now first we do an external inspection with the wrong hostname and after that, we remove hardware-utils annotation and change `inspect.metal3.io: disable` to `inspect.metal3.io: ""`, bmh is reinspected after that and we can see the right host-name after re-inspection.

test/e2e/re_inspection_test.go
lentzi90, line 1: Perfect!

test/e2e/inspection_test.go
lentzi90, line 1: I think we don't need the changes in this file. It is enough with the separate re-inspection test

test/e2e/config/fixture.yaml
```diff
@@ -31,6 +31,7 @@ intervals:
   inspection/wait-inspecting: ["5s", "10ms"]
   inspection/wait-available: ["5s", "1ms"]
   external-inspection/wait-available: ["5s", "1ms"]
+  re-inspection/wait-available: ["5m", "1s"]
```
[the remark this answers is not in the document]
adilGhaffarDev: I added a new one because the others were very small and re-inspection was taking some time. Should I change `default/wait-available:` in both `fixture` and `ironic` to `["10m", "10s"]` and use that instead?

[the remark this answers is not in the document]
lentzi90: Fixture is different it should not take much time. Let me check!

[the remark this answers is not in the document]
adilGhaffarDev: available after re-inspection almost takes 3 minutes I think:
```
  STEP: waiting for the BMH to become available after re-inspection @ 10/31/23 08:40:01.681
  STEP: checking that the hardware details are corrected after re-inspection @ 10/31/23 08:42:54.127
```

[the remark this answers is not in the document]
lentzi90: For fixture it takes less than a second most of the time.
```
  STEP: adding InspectAnnotation to re-inspect @ 10/31/23 10:54:30.718
  STEP: waiting for the BMH to be in inspecting state after inspection annotaion @ 10/31/23 10:54:30.739
  STEP: waiting for the BMH to become available after re-inspection @ 10/31/23 10:54:30.756
  STEP: checking that the hardware details are corrected after re-inspection @ 10/31/23 10:54:30.961
```

Fixture is not what runs of you use the `ci-e2e.sh` script. If you want to run it you need to check the bottom of the readme here: https://github.com/metal3-io/baremetal-operator/blob/main/test/e2e/README.md

test/e2e/re_inspection_test.go
```diff
@@ -0,0 +1,136 @@
...
+		WaitForBmhInProvisioningState(ctx, WaitForBmhInProvisioningStateInput{
+			Client: clusterProxy.GetClient(),
+			Bmh:    bmh,
+			State:  metal3api.StateAvailable,
+		}, e2eConfig.GetIntervals(specName, "wait-available")...)
+
+		By("checking that the hardware details are corrected after re-inspection")
+		key = types.NamespacedName{Namespace: bmh.Namespace, Name: bmh.Name}
+		Expect(clusterProxy.GetClient().Get(ctx, key, &bmh)).To(Succeed())
+		Expect(bmh.Status.HardwareDetails.Hostname).To(Equal(rightHostName))
```
lentzi90, line 130: We need to make this configurable so that the test can be used in other contexts. I think it will be fine to just add a variable to the e2e config with the expected hostname. To verify that it works, we can check with the fixture provider. The hardware details that it returns are hard coded here: https://github.com/metal3-io/baremetal-operator/blob/3874f1e2523335b50836712ff5597c297a48b1b6/pkg/provisioner/fixture/fixture.go#L136-L176

As you can see it does not set any hostname, so for fixture we would set `""`. If we can get the test passing on both fixture and ironic with separate e2e configs then it should be flexible enough for other environments also :slightly_smiling_face:

test/e2e/config/fixture.yaml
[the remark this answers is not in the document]
lentzi90: One more thing! If you run into any issues running the tests with the fixture provider on Mac, please bring them up! I want this to be easy for developers to run :slightly_smiling_face:

test/e2e/re_inspection_test.go
adilGhaffarDev: added in e2e config, please check

test/e2e/config/fixture.yaml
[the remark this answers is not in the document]
adilGhaffarDev: I was running tests in VM but I will try on my mac too.

lentzi90 (commented):
Awesome job!
/lgtm

kashifest (commented):
/approve

PR #1403, opened 2023-10-25
🌱 Implement BMH provisioning E2E test

### LINKED ISSUE #1367: Add provisioning e2e test
We should add a test that provisions and deprovisions a BareMetalHost.
This will require some kind of disk image and potentially an image server if we can not pull it directly from internet.
My suggestion would be to use a small image to avoid unnecessary download times. Nothing too archaic though. It is better that this is well known and tested than that we save seconds and MBs in the CI.

maxrantil:
Introduce an end-to-end test validating the provisioning and deprovisioning flow of a BareMetalHost.

Fixes #1367

### DIFF
--- a/hack/ci-e2e.sh
+++ b/hack/ci-e2e.sh
@@ -54,11 +54,36 @@ docker exec vbmc vbmc list
 # These variables are used by the tests. They override variables in the config file.
 # This IP is defined by the network we created above.
 # Together with the VBMC_PORT this becomes the BMC_ADDRESS used by the BMH in the test.
-export BMC_ADDRESS="ipmi://192.168.222.1:${VBMC_PORT}"
+IP_ADDRESS="192.168.222.1"
+export BMC_ADDRESS="ipmi://${IP_ADDRESS}:${VBMC_PORT}"
 export BOOT_MAC_ADDRESS
 # These are the VBMC defaults (used since we did not specify anything else for `vbmc add`).
 export BMC_USER=admin
 export BMC_PASSWORD=password
+CIRROS_VERSION="0.6.2"
+IMAGE_FILE="cirros-${CIRROS_VERSION}-x86_64-disk.img"
+export IMAGE_CHECKSUM="c8fc807773e5354afe61636071771906"
+export IMAGE_URL="http://${IP_ADDRESS}/${IMAGE_FILE}"
+IMAGE_FOLDER="${REPO_ROOT}/test/e2e/images"
+
+## Setup image server
+# Create a directory for images
+
+mkdir -p "${IMAGE_FOLDER}"
+pushd "${IMAGE_FOLDER}"
+
+## Setup image server
+# Check if IMAGE_FILE already exists
+if [[ ! -f "${IMAGE_FILE}" ]]; then
+    wget "https://download.cirros-cloud.net/${CIRROS_VERSION}/${IMAGE_FILE}"
+else
+    echo "${IMAGE_FILE} already exists. Skipping download."
+fi
+
+popd
+
+# Run image server
+docker run --rm --name image-server-e2e -d -p 80:8080 -v "${IMAGE_FOLDER}:/usr/share/nginx/html" nginxinc/nginx-unprivileged
 
 # We need to gather artifacts/logs before exiting also if there are errors
 set +e
--- a/hack/clean-e2e.sh
+++ b/hack/clean-e2e.sh
@@ -4,11 +4,13 @@ REPO_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
 cd "${REPO_ROOT}" || exit 1
 
 minikube delete
-docker stop vbmc
-docker rm vbmc
+docker rm -f vbmc
+docker rm -f image-server-e2e
 virsh -c qemu:///system destroy --domain bmo-e2e-0
 virsh -c qemu:///system undefine --domain bmo-e2e-0 --remove-all-storage
 virsh -c qemu:///system net-destroy baremetal-e2e
 virsh -c qemu:///system net-undefine baremetal-e2e
 
-rm -rf "${REPO_ROOT}/test/e2e/_artifacts"
+rm -rfv "${REPO_ROOT}/test/e2e/_artifacts"
+rm -rfv "${REPO_ROOT}/artifacts.tar.gz"
+rm -rfv "${REPO_ROOT}/test/e2e/images"
--- a/test/e2e/basic_provisioning_test.go
+++ b/test/e2e/basic_provisioning_test.go
@@ -1,0 +1,145 @@
+package e2e
+
+import (
+	"context"
+	"fmt"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/cluster-api/test/framework"
+	"sigs.k8s.io/cluster-api/util"
+	"sigs.k8s.io/cluster-api/util/patch"
+
+	metal3api "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
+)
+
+var _ = Describe("Provisioning", func() {
+	var (
+		specName       = "provisioning-ops"
+		namespace      *corev1.Namespace
+		cancelWatches  context.CancelFunc
+		bmcUser        string
+		bmcPassword    string
+		bmcAddress     string
+		bootMacAddress string
+	)
+
+	BeforeEach(func() {
+		bmcUser = e2eConfig.GetVariable("BMC_USER")
+		bmcPassword = e2eConfig.GetVariable("BMC_PASSWORD")
+		bmcAddress = e2eConfig.GetVariable("BMC_ADDRESS")
+		bootMacAddress = e2eConfig.GetVariable("BOOT_MAC_ADDRESS")
+
+		namespace, cancelWatches = framework.CreateNamespaceAndWatchEvents(ctx, framework.CreateNamespaceAndWatchEventsInput{
+			Creator:   clusterProxy.GetClient(),
+			ClientSet: clusterProxy.GetClientSet(),
+			Name:      fmt.Sprintf("%s-%s", specName, util.RandomString(6)),
+			LogFolder: artifactFolder,
+		})
+	})
+
+	It("should provision and then deprovision a BMH", func() {
+		By("Creating a secret with BMH credentials")
+		bmcCredentials := corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "bmc-credentials",
+				Namespace: namespace.Name,
+			},
+			StringData: map[string]string{
+				"username": bmcUser,
+				"password": bmcPassword,
+			},
+		}
+		err := clusterProxy.GetClient().Create(ctx, &bmcCredentials)
+		Expect(err).NotTo(HaveOccurred())
+
+		By("Creating a BMH with inspection disabled and hardware details added")
+		bmh := metal3api.BareMetalHost{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      specName,
+				Namespace: namespace.Name,
+				Annotations: map[string]string{
+					metal3api.InspectAnnotationPrefix:   "disabled",
+					metal3api.HardwareDetailsAnnotation: hardwareDetails,
+				},
+			},
+			Spec: metal3api.BareMetalHostSpec{
+				Online: true,
+				BMC: metal3api.BMCDetails{
+					Address:         bmcAddress,
+					CredentialsName: "bmc-credentials",
+				},
+				BootMode:       metal3api.Legacy,
+				BootMACAddress: bootMacAddress,
+			},
+		}
+		err = clusterProxy.GetClient().Create(ctx, &bmh)
+		Expect(err).NotTo(HaveOccurred())
+
+		By("Waiting for the BMH to be in registering state")
+		WaitForBmhInProvisioningState(ctx, WaitForBmhInProvisioningStateInput{
+			Client: clusterProxy.GetClient(),
+			Bmh:    bmh,
+			State:  metal3api.StateRegistering,
+		}, e2eConfig.GetIntervals(specName, "wait-registering")...)
+
+		By("Waiting for the BMH to become available")
+		WaitForBmhInProvisioningState(ctx, WaitForBmhInProvisioningStateInput{
+			Client: clusterProxy.GetClient(),
+			Bmh:    bmh,
+			State:  metal3api.StateAvailable,
+		}, e2eConfig.GetIntervals(specName, "wait-available")...)
+
+		By("Patching the BMH to test provisioning")
+		helper, err := patch.NewHelper(&bmh, clusterProxy.GetClient())
+		Expect(err).NotTo(HaveOccurred())
+		bmh.Spec.Image = &metal3api.Image{
+			URL:      e2eConfig.GetVariable("IMAGE_URL"),
+			Checksum: e2eConfig.GetVariable("IMAGE_CHECKSUM"),
+		}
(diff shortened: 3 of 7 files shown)


test/e2e/basic_provisioning_test.go
```diff
@@ -0,0 +1,138 @@
...
+		By("Patching the BMH to test provisioning")
+		bmhPatch := client.MergeFrom(bmh.DeepCopy())
+		bmh.Spec.Image = &metal3api.Image{
+			URL:      e2eConfig.GetVariable("IMAGE_URL"),
+			Checksum: e2eConfig.GetVariable("IMAGE_CHECKSUM"),
+		}
+		bmh.Spec.RootDeviceHints = &metal3api.RootDeviceHints{
+			DeviceName: "/dev/vda",
+		}
+		err = clusterProxy.GetClient().Patch(ctx, &bmh, bmhPatch)
```
lentzi90, line 104: This is probably working just fine, but we have more commonly been using a patch helper. I think that will be slightly easier to understand and more consistent with a patch helper, then we get something like this:
```suggestion
		helper, err := patch.NewHelper(&bmh, clusterProxy.GetClient())
		Expect(err).NotTo(HaveOccurred())
		bmh.Spec.Image = &metal3api.Image{
			URL:      e2eConfig.GetVariable("IMAGE_URL"),
			Checksum: e2eConfig.GetVariable("IMAGE_CHECKSUM"),
		}
		bmh.Spec.RootDeviceHints = &metal3api.RootDeviceHints{
			DeviceName: "/dev/vda",
		}
		Expect(helper.Patch(ctx, &bmh)).To(Succeed())
```

test/e2e/basic_provisioning_test.go
```diff
@@ -0,0 +1,138 @@
...
+
+		By("Waiting for the BMH to become provisioned")
+		WaitForBmhInProvisioningState(ctx, WaitForBmhInProvisioningStateInput{
+			Client: clusterProxy.GetClient(),
+			Bmh:    bmh,
+			State:  metal3api.StateProvisioned,
+		}, e2eConfig.GetIntervals(specName, "wait-provisioned")...)
+
+		By("Triggering the deprovisioning of the BMH")
+		bmhPatch = client.MergeFrom(bmh.DeepCopy())
```
lentzi90, line 122: I would suggest the same patch helper here

test/e2e/config/ironic.yaml
```diff
@@ -36,3 +40,6 @@ intervals:
   default/wait-namespace-deleted: ["10m", "1s"]
   ironic/wait-deployment: ["10m", "2s"]
   default/wait-power-state: ["10m", "100ms"]
+  default/wait-provisioning: ["20s", "1s"]
+  default/wait-provisioned: ["5m", "5s"]
```
lentzi90, line 44: Wow this is super fast compared to the provisioning in CAPM3! I guess that tiny image helps quite a lot :smile:

test/e2e/config/fixture.yaml
```diff
@@ -32,3 +36,6 @@ intervals:
   default/wait-deployment: ["5m", "1s"]
   default/wait-namespace-deleted: ["20s", "1s"]
   ironic/wait-deployment: ["10m", "2s"]
+  default/wait-provisioning: ["20s", "1s"]
+  default/wait-provisioned: ["5m", "5s"]
```
lentzi90, line 40: This seems excessive. The whole test took ~25 seconds with the fixture provider so I think we can lower both the `5m` deadline and the `5s` interval.
```suggestion
  default/wait-provisioned: ["1m", "10ms"]
```

test/e2e/config/fixture.yaml
```diff
@@ -32,3 +36,6 @@ intervals:
...
+  default/wait-deprovisioning: ["1m", "2s"]
```
lentzi90, line 41: We can probably have the same here
```suggestion
  default/wait-deprovisioning: ["1m", "10ms"]
```

Tiny nit: There is no newline at the end of the file (this is why github shows the :no_entry: sign here). It doesn't really matter by causes unnecessary diffs when adding more lines

lentzi90 (commented):
Great job!
I'm very excited to get this in, since it means we cover all the most important operations.

hack/ci-e2e.sh
```diff
@@ -59,6 +59,108 @@ export BOOT_MAC_ADDRESS
 # These are the VBMC defaults (used since we did not specify anything else for `vbmc add`).
 export BMC_USER=admin
 export BMC_PASSWORD=password
+# This is the IP address of the host where the image server is running.
+IMAGE_FILE="cirros-0.5.2-x86_64-disk.img"
+export IMAGE_MD5SUM="b874c39491a2377b8490f5f1e89761a4"
+IMAGE_IP_ADDRESS=$(echo $BMC_ADDRESS | sed -n 's|ipmi://\([^:]*\).*|\1|p')
```
lentzi90, line 65: I think we can make it even easier to understand. Let's create a new variable, e.g. `IP_ADDRESS` or `MINIKUBE_IP`. Then use this to set both the IMAGE_URL and the BMC_ADDRESS.

```
IP_ADDRESS="192.168.222.1"
BMC_ADDRESS="ipmi://${IP_ADDRESS}:${VBMC_PORT}"
IMAGE_URL="http://${IP_ADDRESS}/${IMAGE_FILE}"
```

hack/ci-e2e.sh
```diff
@@ -59,6 +59,108 @@ export BOOT_MAC_ADDRESS
...
+# Check if container "image-server" is running
+if [ "$(docker ps -q -f name=image-server)" ]; then
+    echo "Container 'image-server' is already running. Removing..."
+    docker rm -f image-server
+fi
+
+# Run image server
+docker run --rm --name image-server -d -p 80:8080 -v "$(pwd)/images:/usr/share/nginx/html" nginxinc/nginx-unprivileged
+
+## Setup image server
```
lentzi90, line 91: Everything from here down until line 163 looks like duplicate or older version of the same code. Probably some mistake when squashing?

hack/ci-e2e.sh
```diff
@@ -54,11 +54,33 @@ docker exec vbmc vbmc list
...
+# Create a directory for images
+mkdir -p images
+pushd images
+
+## Setup image server
+# Check if IMAGE_FILE already exists
+if [[ ! -f $IMAGE_FILE ]]; then
+    wget http://download.cirros-cloud.net/0.5.2/$IMAGE_FILE
+else
+    echo "$IMAGE_FILE already exists. Skipping download."
```
lentzi90, line 77: Nit: Use brackets and quoting everywhere for consistency and safety, in case there would be special characters in the string.
Any reason why you chose 0.5.2 btw? I have no problem with it, just curious since I see 0.6.2 is the latest.
```suggestion
if [[ ! -f "${IMAGE_FILE}" ]]; then
    wget "https://download.cirros-cloud.net/0.5.2/${IMAGE_FILE}"
else
    echo "${IMAGE_FILE} already exists. Skipping download."
```

hack/clean-e2e.sh
```diff
@@ -6,9 +6,11 @@ cd "${REPO_ROOT}" || exit 1
 minikube delete
 docker stop vbmc
 docker rm vbmc
+docker rm -f image-server-e2e
```
lentzi90, line 9: Nit: Let's keep these consistent
```suggestion
docker rm -f vbmc
docker rm -f image-server-e2e
```

lentzi90 (commented):
Just a couple more comments. This is progressing really nicely!
/metal3-bmo-e2e-test

lentzi90:
Great job! Please squash and I will put LGTM on it :slightly_smiling_face: (no need to add me as co-author for those small fixes :sweat_smile: )

maxrantil:
I did change the image to the latest CirrOS (I think I took 0.5.2 because of how the MD5SUM file looked and thought that was the latest at first glance). Now I experiment with the export of variables to the conf file from the ci-hack script and got some problems. I will squash the commits after I got this fixed.

maxrantil:
this part:
`# These variables are used by the tests. They override variables in the config file.
# This IP is defined by the network we created above.
# Together with the VBMC_PORT this becomes the BMC_ADDRESS used by the BMH in the test.
IP_ADDRESS="192.168.222.1"
export BMC_ADDRESS="ipmi://${IP_ADDRESS}:${VBMC_PORT}"
export BOOT_MAC_ADDRESS
# These are the VBMC defaults (used since we did not specify anything else for `vbmc add`).
export BMC_USER=admin
export BMC_PASSWORD=password
IMAGE_FILE="cirros-0.6.2-x86_64-disk.img"
export IMAGE_MD5SUM="c8fc807773e5354afe61636071771906"
export IMAGE_URL="http://${IP_ADDRESS}/${IMAGE_FILE}"`

After testing it, they do NOT override the variables in the config file. Any thoughts on that @lentzi90 ? Should I take back the function I had earlier for doing that?

lentzi90 (commented):
/lgtm

lentzi90:
New changes were fixing a mismatch in variable naming between the hack script and e2e config. (It did not fail any tests since we anyway had the same value in both places.)
/lgtm
/metal3-bmo-e2e-test

hack/ci-e2e.sh
```diff
@@ -54,11 +54,33 @@ docker exec vbmc vbmc list
...
 # This IP is defined by the network we created above.
 # Together with the VBMC_PORT this becomes the BMC_ADDRESS used by the BMH in the test.
-export BMC_ADDRESS="ipmi://192.168.222.1:${VBMC_PORT}"
+IP_ADDRESS="192.168.222.1"
+export BMC_ADDRESS="ipmi://${IP_ADDRESS}:${VBMC_PORT}"
 export BOOT_MAC_ADDRESS
 # These are the VBMC defaults (used since we did not specify anything else for `vbmc add`).
 export BMC_USER=admin
 export BMC_PASSWORD=password
+IMAGE_FILE="cirros-0.6.2-x86_64-disk.img"
```
mquhuy, line 63: This looks really great! Thank you a lot xD
I'm wondering though if we should put the cirros img version into a variable, for future upgrade? This `0.6.2` does appear again and again

mquhuy: Also, maybe we should change `IMAGE_FILE` to something more specific? Like `CIRROS_IMAGE_FILE` or `IMAGE_SERVER_E2E_FILE` or something?

test/e2e/config/ironic.yaml
```diff
@@ -24,6 +24,8 @@ variables:
   BMC_PASSWORD: password
   BMC_ADDRESS: ipmi://192.168.222.1:16230
   BOOT_MAC_ADDRESS: "00:60:2f:31:81:01"
+  IMAGE_URL: "http://192.168.222.1/cirros-0.6.2-x86_64-disk.img"
```
mquhuy, line 27: Same for this IMAGE_URL.

mquhuy (changes requested):
This looks awesome, and I think it's a great improvement.
Just some tiny bit comments, otherwise lgtm.

hack/ci-e2e.sh
lentzi90: Good point about the version. I don't have a strong opinion on it but it would sure simplify a bit when changing it later.

Regarding `IMAGE_FILE` I think we should not change it. We happen to use cirros now, but there is nothing preventing us from using other images in the future so I don't think it should include the OS. Perhaps there would be some benefit to `DISK_IMAGE_FILE` or similar, to make it obvious what kind of image it is? But honestly to me it seems good like it is already. The context in the script should explain it pretty good :slightly_smiling_face:

maxrantil: I could do a variable for the cirros version if you like but then we need to download the MD5 file and do a checksum test aswell. I'll dig into that and will also change the name to ``IMAGE_FILE_E2E``.  Thanks for the review!

lentzi90: Wait wait, please don't change it to `IMAGE_FILE_E2E`. The context should already be clear. The whole file is about e2e so no need to name the variables e2e.
Good point about the MD5 also. I don't think we should download the MD5 file. It is just unnecessarily complicating things at this point. The only issue (which is really small I would say) is that we have the version (0.6.2) in multiple places, which makes it slightly harder to update. If you want to improve this, you can create a variable for it (e.g. `CIRROS_VERSION`). It can then be used in the file name and URL. Updating the image would then only require changing this variable and the MD5 sum.

maxrantil: Got it. Good points.

test/e2e/basic_provisioning_test.go
```diff
@@ -0,0 +1,138 @@
...
+		Expect(err).NotTo(HaveOccurred())
+		bmh.Spec.Image = nil
+		Expect(helper.Patch(ctx, &bmh)).To(Succeed())
+
+		By("Waiting for the BMH to become deprovisioned")
+		WaitForBmhInProvisioningState(ctx, WaitForBmhInProvisioningStateInput{
+			Client: clusterProxy.GetClient(),
+			Bmh:    bmh,
+			State:  metal3api.StateDeprovisioning,
+		}, e2eConfig.GetIntervals(specName, "wait-deprovisioning")...)
```
kashifest, line 132: Dont we need to check if the `deprovisioning` succeeded and the BMH becomes `available` eventually? cc @lentzi90

lentzi90: Ah good point! I somehow mixed things up when the `By` mentioned `deprovisioned`. It should be changed to `deprovisioning` and then we should add a wait for `available` to make sure that deprovisioning succeeded!

maxrantil: Fixed.

maxrantil: Oh, I might have missed one thing you mentioned @lentzi90 , I just added the ``available`` after ``deprovisioned`` now. So the ``deprovisioning`` state is not checked right now. Is that necessary to have?

lentzi90 (commented):
/lgtm

test/e2e/config/fixture.yaml
```diff
@@ -32,3 +34,6 @@ intervals:
   default/wait-deployment: ["5m", "1s"]
   default/wait-namespace-deleted: ["20s", "1s"]
   ironic/wait-deployment: ["10m", "2s"]
+  default/wait-provisioning: ["20s", "1s"]
+  default/wait-provisioned: ["1m", "10ms"]
+  default/wait-deprovisioning: ["1m", "2s"]
```
kashifest, line 39: Although we are not testing this currently, shouldn't we also have `default/wait-available` declared here as we have in ironic.yaml . Whats the plan for running fixture by the way?

maxrantil: Good catch, I've missed that one, Also noticed that ``default/wait-registration`` was missing. I have both ready for a commit if needed. What do you think @lentzi90 ?

lentzi90: Good find! Yes this is needed or we get a timeout when running the fixture provider. We can use the same value as for `inspection/wait-available`.

> Whats the plan for running fixture by the way?

There is no plan currently. I try to make sure I run it manually (locally on my laptop) for every PR for now. The fixture tests are not important per se, but they give a good indication if the tests are flexible enough to be able to run on bare metal. I.e. if we can run fixture, then there is no dependency on libvirt or similar.
Maybe we should add a test that runs the fixture tests just for this reason, what do you think?

kashifest: Yeah thats a good point, I was suggesting mainly from the point since we are introducing changes in this file in every tests and it would be good to test them also, otherwise issues will just pile up, but if you are testing in your laptop already, there wouldn't be much issues. Anyways, its good to have it in CI incase if someone forgets to test it.

lentzi90: Added an issue for it: https://github.com/metal3-io/baremetal-operator/issues/1406

lentzi90 (commented):
/lgtm

lentzi90:
Previous test lost contact to jenkins worker...
/test-ubuntu-integration-main

PR #1407, opened 2023-10-30
🌱 add verify-release.sh

tuminoid:
Add `./hack/verify-release.sh` to check if the release process has been completed, and/or regular chores have been done before release. This script has been adapted from the one I wrote for CAPM3.

Another PR(s) will be created to fix the issues identified.

PR #1408, opened 2023-10-30
🌱 bump CAPI to v1.5.3

tuminoid:
Bump CAPI to v1.5.3, while fixing inconsistencies in the modules, and removing unnecessary replace that was even further changing the version.

Inconsistencies spotted via #1407


kashifest (commented):
/approve

PR #1410, opened 2023-10-30
🌱 Integrate sushy-tools and Redfish into e2e testing

### LINKED ISSUE #1374: Test Redfish in e2e tests
The initial e2e tests are just using VBMC and IPMI. We should add sushy-tools and Redfish. I think a good way to do this is to run some/all tests with both. So for example, in the inspection test, we would run it once with IPMI and once with Redfish (potentially in parallel).

/kind feature

maxrantil:
This PR augments the e2e testing framework by integrating sushy-tools and Redfish, enhancing test coverage across different management interfaces alongside the existing VBMC and IPMI tests.

Fixes #1374

### DIFF
--- a/hack/ci-e2e.sh
+++ b/hack/ci-e2e.sh
@@ -1,9 +1,20 @@
 #!/usr/bin/env bash
 
+# -----------------------------------------------------------------------------
+# Description: This script sets up the environment and runs E2E tests for the
+#              BMO project. It uses either vbmc or sushy-tools based on 
+#              the BMO_E2E_EMULATOR environment variable.
+# Usage:       export BMO_E2E_EMULATOR="vbmc"  # Or "sushy-tools"
+#              ./ci-e2e.sh
+# -----------------------------------------------------------------------------
+
 set -eux
 
 REPO_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
 cd "${REPO_ROOT}" || exit 1
+
+# BMO_E2E_EMULATOR can be set to either "vbmc" or "sushy-tools"
+BMO_E2E_EMULATOR=${BMO_E2E_EMULATOR:-"sushy-tools"}
 
 # Ensure requirements are installed
 "${REPO_ROOT}/hack/e2e/ensure_go.sh"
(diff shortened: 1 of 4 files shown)


hack/ci-e2e.sh
```diff
@@ -1,10 +1,21 @@
 #!/usr/bin/env bash
 
+# -----------------------------------------------------------------------------
+# Description: This script sets up the environment and runs E2E tests for the
+#              BMO project. It uses either vbmc or sushy-tools based on 
+#              the USE_TOOL environment variable.
+# Usage:       export USE_TOOL="vbmc"  # Or "sushy-tools"
```
tuminoid, line 7: Could we use more specific and descriptive name for the variable? `BMO_E2E_EMULATOR` or something like that.

hack/ci-e2e.sh
```diff
@@ -1,10 +1,21 @@
...
+#              ./ci-e2e.sh
+# -----------------------------------------------------------------------------
+
 set -eux
 
 REPO_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
 cd "${REPO_ROOT}" || exit 1
 
+# USE_TOOL can be set to either "vbmc" or "sushy-tools"
+USE_TOOL=${USE_TOOL:-"vbmc"}  # Defaults to vbmc if not set
```
tuminoid, line 17:
```suggestion
USE_TOOL=${USE_TOOL:-"vbmc"}
```
Comment is only stating the obvious, better not have it.

hack/ci-e2e.sh
```diff
@@ -32,58 +43,100 @@ minikube start
...
 export BMC_PASSWORD=password
-CIRROS_VERSION="0.6.2"
-IMAGE_FILE="cirros-${CIRROS_VERSION}-x86_64-disk.img"
-export IMAGE_CHECKSUM="c8fc807773e5354afe61636071771906"
-export IMAGE_URL="http://${IP_ADDRESS}/${IMAGE_FILE}"
-IMAGE_FOLDER="${REPO_ROOT}/test/e2e/images"
 
-## Setup image server
-# Create a directory for images
+if [[ "$USE_TOOL" == "vbmc" ]]; then
```
tuminoid, line 73:
```suggestion
if [[ "${USE_TOOL}" == "vbmc" ]]; then
```

hack/ci-e2e.sh
```diff
@@ -32,58 +43,100 @@ minikube start
...
+	# Start VBMC
+	docker run --name vbmc --network host -d \
+	-v /var/run/libvirt/libvirt-sock:/var/run/libvirt/libvirt-sock \
```
tuminoid, line 76:
```suggestion
        	-v /var/run/libvirt/libvirt-sock:/var/run/libvirt/libvirt-sock \
```
Easier to read if you indent multi-line commands.

hack/ci-e2e.sh
```diff
@@ -32,58 +43,100 @@ minikube start
...
+	# Together with the VBMC_PORT this becomes the BMC_ADDRESS used by the BMH in the test.
+	export BMC_ADDRESS="ipmi://${IP_ADDRESS}:${VBMC_PORT}"
+
+	# Add BMH VM to VBMC
+	docker exec vbmc vbmc add "${VM_NAME}" --port "${VBMC_PORT}"
+	docker exec vbmc vbmc start "${VM_NAME}"
+	docker exec vbmc vbmc list
+
+
+elif [[ "$USE_TOOL" == "sushy-tools" ]]; then
```
tuminoid, line 90:
```suggestion
elif [[ "${USE_TOOL}" == "sushy-tools" ]]; then
```

hack/ci-e2e.sh
```diff
@@ -32,58 +43,100 @@ minikube start
...
+	# Sushy-tools variables
+	SUSHY_EMULATOR_FILE="${REPO_ROOT}"/test/e2e/sushy-tools/sushy-emulator.conf
+	SUSHY_PORT="8000"
+
+	# Start sushy-tools
+	# Privileged mode is needed to access libvirt.
+	sudo docker run --name sushy-tools -d --privileged --network host \
+	-v "${SUSHY_EMULATOR_FILE}":/etc/sushy/sushy-emulator.conf:Z \
```
tuminoid, line 98: Same about indentation.

hack/ci-e2e.sh
```diff
@@ -32,58 +43,100 @@ minikube start
...
-CIRROS_VERSION="0.6.2"
-IMAGE_FILE="cirros-${CIRROS_VERSION}-x86_64-disk.img"
-export IMAGE_CHECKSUM="c8fc807773e5354afe61636071771906"
-export IMAGE_URL="http://${IP_ADDRESS}/${IMAGE_FILE}"
-IMAGE_FOLDER="${REPO_ROOT}/test/e2e/images"
 
-## Setup image server
-# Create a directory for images
+if [[ "$USE_TOOL" == "vbmc" ]]; then
+	# Start VBMC
```
tuminoid, line 74: Please use 2 or 4 spaces, follow the convention in the script. I think you're inserting tabs here.

hack/ci-e2e.sh
```diff
@@ -32,58 +43,100 @@ minikube start
...
+	-e SUSHY_EMULATOR_CONFIG=/etc/sushy/sushy-emulator.conf \
+	quay.io/metal3-io/sushy-tools:latest sushy-emulator
+
+	sleep 5
+
+	# Fetch the ID of the system with the name 'bmo-e2e-0'
+	ids=$(curl -s "http://${IP_ADDRESS}:${SUSHY_PORT}/redfish/v1/Systems" | jq -r '.Members[] | .["@odata.id"] | split("/")[-1] | rtrimstr("\"")')
+
+	# Iterate over each ID and fetch its details
+	for id in $ids; do
```
tuminoid, line 109:
```suggestion
	for id in ${ids}; do
```

hack/ci-e2e.sh
```diff
@@ -32,58 +43,100 @@ minikube start
...
+		data=$(curl -s "http://${IP_ADDRESS}:${SUSHY_PORT}/redfish/v1/Systems/${id}")
+		name=$(echo "$data" | jq -r '.Name')
```
tuminoid, line 111:
```suggestion
		name=$(echo "${data}" | jq -r '.Name')
```

hack/ci-e2e.sh
```diff
@@ -32,58 +43,100 @@ minikube start
...
+		if [[ "$name" == "bmo-e2e-0" ]]; then
```
tuminoid, line 112:
```suggestion
		if [[ "${name}" == "bmo-e2e-0" ]]; then
```

hack/ci-e2e.sh
```diff
@@ -32,58 +43,100 @@ minikube start
...
+			DESIRED_ID="$id"
```
tuminoid, line 113:
```suggestion
			DESIRED_ID="${id}"
```
Also, why are we saving this id into DESIRED_ID, if we're only using it on the next line, and then breaking?

hack/ci-e2e.sh
```diff
@@ -32,58 +43,100 @@ minikube start
...
-mkdir -p "${IMAGE_FOLDER}"
-pushd "${IMAGE_FOLDER}"
-
-## Setup image server
-# Check if IMAGE_FILE already exists
-if [[ ! -f "${IMAGE_FILE}" ]]; then
-    wget "https://download.cirros-cloud.net/${CIRROS_VERSION}/${IMAGE_FILE}"
 else
-    echo "${IMAGE_FILE} already exists. Skipping download."
+    echo "Invalid tool specified: $USE_TOOL"
```
tuminoid, line 120:
```suggestion
    echo "Invalid tool specified: ${USE_TOOL}"
```

tuminoid (commented):
Some comments and nits.

hack/ci-e2e.sh
maxrantil: Got it. Yes, it's a tab.

hack/ci-e2e.sh
maxrantil: Yes I see what you mean and you're right, I'll change it, it's a relic from an old fix.

hack/ci-e2e.sh
```diff
@@ -32,58 +43,97 @@ minikube start
...
+  # Privileged mode is needed to access libvirt.
+  sudo docker run --name sushy-tools -d --privileged --network host \
+    -v "${SUSHY_EMULATOR_FILE}":/etc/sushy/sushy-emulator.conf:Z \
+    -v /var/run/libvirt:/var/run/libvirt:Z \
+    -e SUSHY_EMULATOR_CONFIG=/etc/sushy/sushy-emulator.conf \
+    quay.io/metal3-io/sushy-tools:latest sushy-emulator
+
+  sleep 5
+
+  # Fetch the ID of the system with the name 'bmo-e2e-0'
```
dtantsur, line 104: nit: You can actually use the name directly in the URI.

test/e2e/sushy-tools/sushy-emulator.conf
```diff
@@ -0,0 +1,36 @@
+# Listen on all local IP interfaces
+SUSHY_EMULATOR_LISTEN_IP = u'0.0.0.0'
```
dtantsur, line 2: This is not great. You use neither authentication nor TLS. This will open VM management on your machine to anyone. Cannot we assume localhost? If not, I'd probably add TLS and auth.

lentzi90: Good point! I think we can use `192.168.222.1` here. Localhost will probably not work since Ironic would not be able to reach it then.

hack/ci-e2e.sh
```diff
@@ -32,58 +43,97 @@ minikube start
...
+  docker exec vbmc vbmc list
+
+elif [[ "${BMO_E2E_EMULATOR}" == "sushy-tools" ]]; then
+  # Sushy-tools variables
+  SUSHY_EMULATOR_FILE="${REPO_ROOT}"/test/e2e/sushy-tools/sushy-emulator.conf
+  SUSHY_PORT="8000"
+
+  # Start sushy-tools
+  # Privileged mode is needed to access libvirt.
+  sudo docker run --name sushy-tools -d --privileged --network host \
```
lentzi90, line 96: This about privileged mode seems to be no longer true or some misunderstanding. I got suspicious since VBMC should need the exact same kind of access to libvirt and did not require it. Now I have tested and confirmed that it is not necessary. 
```suggestion
  docker run --name sushy-tools -d --network host \
```

hack/ci-e2e.sh
```diff
@@ -32,58 +43,97 @@ minikube start
...
+    -v "${SUSHY_EMULATOR_FILE}":/etc/sushy/sushy-emulator.conf:Z \
+    -v /var/run/libvirt:/var/run/libvirt:Z \
+    -e SUSHY_EMULATOR_CONFIG=/etc/sushy/sushy-emulator.conf \
+    quay.io/metal3-io/sushy-tools:latest sushy-emulator
+
+  sleep 5
```
lentzi90, line 102: Do you think it would be possible to do some kind of retry instead of just sleep here? Maybe curl up to 5-10 times until we get success?

hack/ci-e2e.sh
```diff
@@ -32,58 +43,97 @@ minikube start
...
+CIRROS_VERSION="0.6.2"
+IMAGE_FILE="cirros-${CIRROS_VERSION}-x86_64-disk.img"
+export IMAGE_CHECKSUM="c8fc807773e5354afe61636071771906"
+export IMAGE_URL="http://${IP_ADDRESS}/${IMAGE_FILE}"
+IMAGE_DIR="${REPO_ROOT}/test/e2e/images"
+
+## Download and run image server
+mkdir -p "${IMAGE_DIR}"
+pushd "${IMAGE_DIR}"
+wget "https://download.cirros-cloud.net/${CIRROS_VERSION}/${IMAGE_FILE}"
```
lentzi90, line 131: nit: Wget's progress report is a bit spammy in the logs. I think it would be nice to make it quiet
```suggestion
wget --quiet "https://download.cirros-cloud.net/${CIRROS_VERSION}/${IMAGE_FILE}"
```

lentzi90 (commented):
Great job once again!
Some comments below :slightly_smiling_face:

hack/ci-e2e.sh
maxrantil: Whoaw that's great.

hack/ci-e2e.sh
maxrantil: When knowing that you could use the `VM_NAME` directly in the URI, there is no need for the `sleep` anymore.

lentzi90 (commented):
/lgtm

lentzi90:
Issues with the jenkins workers...
/metal3-bmo-e2e-test

hack/clean-e2e.sh
```diff
@@ -4,8 +4,12 @@ REPO_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
 cd "${REPO_ROOT}" || exit 1
 
 minikube delete
+docker stop vbmc
```
mquhuy, line 7: Minor issue: this step seems to be redundant. `docker rm -f` should stop and clean the container if it's running. Otoh `docker stop` will give out an error if the container is not running, which means we will always have some weird errors unless we run both the emulators.

test/e2e/config/ironic.yaml
```diff
@@ -33,7 +33,7 @@ intervals:
   inspection/wait-registration-error: ["1m", "5s"]
   external-inspection/wait-available: ["20s", "1s"]
   default/wait-inspecting: ["2m", "10s"]
-  default/wait-available: ["10m", "10s"]
+  default/wait-available: ["10m", "1s"]
```
mquhuy, line 36: I've not checked, but I wonder if this breaks anything? Is it enough for vbmc, and do we use this wait for anything else?

mquhuy (commented):
It looks very good! I just have some minor comments.

hack/clean-e2e.sh
maxrantil: Ok true, I can't remember how they got in here but I think you are right. Let's remove them.

test/e2e/config/ironic.yaml
maxrantil: Do you mean the change from `10s` to `1s`? I had some problems here while debugging and thought that the 10s interval check would have been able to have missed the state. That's why I changed it. But after finding the bug I don't think this was the problem.

Maybe I'm missing something here, what do you mean by breaks anything?

lentzi90 (commented):
/lgtm

test/e2e/config/ironic.yaml
lentzi90: The `"1s"` is just how often it checks. The deadline is `"10m"` still so this should not really change anything :slightly_smiling_face:

mquhuy: Yes. Basically I was wondering if this wait time is used somewhere else, and if it is, would 1s be sufficient in those cases.

mquhuy: Ah. Alright, I forgot about that xD Nevermind then.

maxrantil: It is used in the Interval-test and in the Provisioning-test at least. But like lentzi90 is saying its just how often it checks. The limit before ending the check is till 10 minutes.

kashifest (commented):
/approve

PR #1418, opened 2023-10-31
🌱 fix various issues with go modules

tuminoid:
Fix mismatched versions, incorrect replaces, etc. Findings are from verify-release.sh (PR #1407)


tuminoid:
WIP: waiting for dependabot to merge couple bumps, so I can rebase.

test/go.mod
```diff
@@ -5,21 +5,19 @@ go 1.20
 require (
 	github.com/metal3-io/baremetal-operator/apis v0.4.0
 	github.com/metal3-io/cluster-api-provider-metal3/test v1.5.1
-	github.com/onsi/ginkgo/v2 v2.12.0
-	github.com/onsi/gomega v1.27.10
+	github.com/onsi/ginkgo/v2 v2.13.0
```
lentzi90, line 8: The ginkgo version is unfortunately also defined in the Makefile: https://github.com/metal3-io/baremetal-operator/blob/dfc23753f8c37dc565e2e487a18c65845f8169e2/Makefile#L25
Can you update that as well?

tuminoid: Done, good catch. Is it there for the e2e?

lentzi90 (commented):
/lgtm

test/go.mod
lentzi90: Yeah I added it together with the e2e tests. Probably we could do it in a better way so that we don't need the separate version there

tuminoid:
/test-ubuntu-integration-main 
Maybe CI is back online now ...

PR #1420, opened 2023-10-31
🐛 Log node updates with a normal logger

dtantsur:


zaneb:
This is going to get pretty chatty. Every single bit of data we write to ironic will get logged. I'm not sure we want that.
I would support adding a separate `-debug` CLI flag so that the level can be set independently of the other side effects of the `-dev` flag (which also disables structured logging).

dtantsur:
> This is going to get pretty chatty. Every single bit of data we write to ironic will get logged. I'm not sure we want that.

Yes, we absolutely do want that. This is much less chatty that the stuff we already log. If you take a look at what we log, you'll see what I mean. The reason for proposing this patch is a bug where BMO gets stuck trying to update deploy_interface on an externally provisioned node - an operation that just should not happen, but it does, and I don't even know what the values are. So, while I support the idea of having the debug flag, this is one of the first things that must never be covered by it.

dtantsur:
Here are some numbers from the recent run:
```
$ grep "option data" ~/Work/stdout.log | wc -l
85
```
This may seem a lot, but it's only 4% of all message. For contrast:
```
$ grep "registering and validating access" ~/Work/stdout.log | wc -l
175
$ grep controllers.BMCEventSubscription ~/Work/stdout.log | wc -l
134
```
(The CI does not even use BMC subscription)

PR #1427, opened 2023-11-03
🌱 Clean up logging and error handling in the webhooks

dtantsur:
Logging:
- Rename the webhook loggers for consistency and easier grepping
- Remove redundant logging from the validation code

Errors:
- Provide wrapped errors in a consistent fashion
- Do not refer to internal constant names in user-visible strings
- Provide allowed values when saying that something is not allowed
- Use the stdlib error routines instead of the deprecated pkg/errors.
- Do not shadow the errors package name by local variables

### DIFF
--- a/apis/metal3.io/v1alpha1/baremetalhost_webhook.go
+++ b/apis/metal3.io/v1alpha1/baremetalhost_webhook.go
@@ -23,22 +23,22 @@ import (
 )
 
 // log is for logging in this package.
-var baremetalhostlog = logf.Log.WithName("baremetalhost-resource")
+var baremetalhostlog = logf.Log.WithName("webhooks").WithName("BareMetalHost")
 
 //+kubebuilder:webhook:verbs=create;update,path=/validate-metal3-io-v1alpha1-baremetalhost,mutating=false,failurePolicy=fail,sideEffects=none,admissionReviewVersions=v1;v1beta,groups=metal3.io,resources=baremetalhosts,versions=v1alpha1,name=baremetalhost.metal3.io
 
 // ValidateCreate implements webhook.Validator so a webhook will be registered for the type
 func (r *BareMetalHost) ValidateCreate() (admission.Warnings, error) {
-	baremetalhostlog.Info("validate create", "name", r.Name)
+	baremetalhostlog.Info("validate create", "namespace", r.Namespace, "name", r.Name)
 	return nil, errors.NewAggregate(r.validateHost())
 }
 
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
 func (r *BareMetalHost) ValidateUpdate(old runtime.Object) (admission.Warnings, error) {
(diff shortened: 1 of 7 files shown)


apis/metal3.io/v1alpha1/baremetalhost_webhook.go
```diff
@@ -23,7 +23,7 @@ import (
 )
 
 // log is for logging in this package.
-var baremetalhostlog = logf.Log.WithName("baremetalhost-resource")
+var baremetalhostlog = logf.Log.WithName("webhooks.BareMetalHost")
```
zaneb, line 26: Prefer `Log.WithName("webhooks").WithName("BareMetalHost")` for consistency with e.g. https://github.com/metal3-io/baremetal-operator/blob/main/main.go#L243

apis/metal3.io/v1alpha1/bmceventsubscription_webhook.go
```diff
@@ -23,7 +23,7 @@ import (
 )
 
 // bmcsubscriptionlog is for logging in this package.
-var bmcsubscriptionlog = logf.Log.WithName("bmceventsubscription-resource")
+var bmcsubscriptionlog = logf.Log.WithName("webhooks.BMCEventSubscription")
```
zaneb, line 26: ditto

apis/metal3.io/v1alpha1/bmceventsubscription_validation.go
```diff
@@ -29,23 +28,22 @@ var bmclog = logf.Log.WithName("bmceventsubscription-validation")
 
 // validateSubscription validates BMCEventSubscription resource for creation
 func (s *BMCEventSubscription) validateSubscription() []error {
-	bmclog.Info("validate create", "name", s.Name)
```
zaneb, line 32: `bmclog` above is now unused

zaneb (commented):
/approve

apis/metal3.io/v1alpha1/baremetalhost_webhook.go
zaneb, line 1: We should also log the namespace as well as the resource name.

apis/metal3.io/v1alpha1/bmceventsubscription_validation.go
```diff
@@ -29,23 +28,22 @@ var bmclog = logf.Log.WithName("bmceventsubscription-validation")
...
 	var errs []error
 
 	if s.Spec.HostName == "" {
-		errs = append(errs, fmt.Errorf("HostName cannot be empty"))
+		errs = append(errs, errors.New("hostName cannot be empty"))
```
zaneb, line 34: We should be validating this client-side, you don't need a webhook just to make sure a field is not empty.

dtantsur: I agree, but I'd rather keep this PR refactoring-only

PR #1435, opened 2023-11-07
✨ Implement API version negotiation

dtantsur:
Rather than bumping the microversion every time we need new features,
we will keep the baseline version and try to negotiate a higher version
in the provisioner. The new AvailableFeatures object provides
information about the supported features.

Includes:
- https://github.com/metal3-io/baremetal-operator/pull/1433

### DIFF
--- a/pkg/provisioner/ironic/clients/features.go
+++ b/pkg/provisioner/ironic/clients/features.go
@@ -1,0 +1,57 @@
+package clients
+
+import (
+	"fmt"
+
+	"github.com/go-logr/logr"
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/utils"
+)
+
+// AvailableFeatures represents features that Ironic API provides.
+// See https://docs.openstack.org/ironic/latest/contributor/webapi-version-history.html
+type AvailableFeatures struct {
+	MaxVersion int
+}
+
+// The minimum microversion that BMO can work with. We need to be really
+// conservative when updating this value since doing so unconditionally breaks
+// operators of older Ironic, even if the feature we need is optional.
+// Update docs/configuration.md when updating the version.
+// Version 1.81 allows retrival of Node inventory
+const baseline = "1.81"
+
+func GetAvailableFeatures(client *gophercloud.ServiceClient) (features AvailableFeatures, err error) {
+	mvs, err := utils.GetSupportedMicroversions(client)
+	if err != nil {
+		return
+	}
+
+	if mvs.MaxMajor != 1 || mvs.MaxMinor < 81 {
+		err = fmt.Errorf("Ironic API 1.81 or newer is required, got %d.%d", mvs.MaxMajor, mvs.MaxMinor)
+		return
+	}
+
+	features.MaxVersion = mvs.MaxMinor
+	return
+}
+
+func (af AvailableFeatures) Log(logger logr.Logger) {
+	// NOTE(dtantsur): update this every time we have more features of interest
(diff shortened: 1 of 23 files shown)


pkg/provisioner/ironic/dependencies.go
```diff
@@ -1,57 +1,31 @@
...
 	}
 	return ready, err
 }
 
-func (i *ironicDependenciesChecker) checkEndpoint(client *gophercloud.ServiceClient) (ready bool) {
-
-	// NOTE: Some versions of Ironic inspector returns 404 for /v1/ but 200 for /v1,
-	// which seems to be the default behavior for Flask. Remove the trailing slash
-	// from the client endpoint.
-	endpoint := strings.TrimSuffix(client.Endpoint, "/")
```
lentzi90, line 42: Is it certain that removing this will not cause any issues? Could there be users relying on this?

lentzi90 (commented):
Exciting! :star_struck:

pkg/provisioner/ironic/dependencies.go
dtantsur: From the Ironic's perspective, these URLs are identical (inspector did have this bug at some point). What scenario do you have in mind?

lentzi90: I'm just wondering why this was here to begin with and why it would be safe to remove it now. Sounds like the answer to the first question is that inspector had a bug. Then the question remains, can we be certain that no user is running a version of inspector that is affected?
I guess this is old enough that anyone still using an affected inspector would be out of support anyway :)

dtantsur: BMO no longer accesses Inspector directly, so it does not matter which version a user is running. It's between Ironic and Inspector now.

pkg/provisioner/ironic/clients/features.go
```diff
@@ -0,0 +1,57 @@
...
+	features.MaxVersion = mvs.MaxMinor
+	return
+}
+
+func (af AvailableFeatures) Log(logger logr.Logger) {
+	// NOTE(dtantsur): update this every time we have more features of interest
+	logger.Info("supported Ironic API features",
+		"maxVersion", fmt.Sprintf("1.%d", af.MaxVersion),
+		"chosenVersion", af.ChooseMicroversion(),
+		"firmwareUpgrades", af.HasFirmwareUpgrades())
```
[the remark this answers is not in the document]
dtantsur: I was thinking that the log messages would read like .. "supported Ironic API features ... firmwareUpgrades = true". WDYT?

pkg/provisioner/ironic/dependencies.go
lentzi90: Right! I forgot about that!
Nothing to see here, move along :smile:

lentzi90 (commented):
/lgtm

pkg/provisioner/ironic/clients/features.go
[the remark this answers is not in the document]
iurygregory: I think this could me miss interpreted (they would think they can only use to Upgrade firmware, if they had to downgrade maybe they would do manually before using BMO? Updates make more sense I think..)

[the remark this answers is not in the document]
dtantsur: Feel free to adjust the namings in your patch to make them consistent with what you propose.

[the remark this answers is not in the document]
iurygregory: Awesome! Tks @dtantsur

PR #1437, opened 2023-11-08
🌱 Add github workflow for e2e fixture tests

### LINKED ISSUE #1406: E2e: Add fixture tests to CI
**User Story**

As a developer I would like to ensure that the BMO e2e tests flexible enough to run in many different environments, including on actual hardware.

**Detailed Description**

We rely on libvirt, VBMC/sushy-tools and similar for CI. This introduces the risk that the e2e tests become dependent on the platform.
As a way to mitigate this, I initially made the e2e tests so that they can be run with the fixture provider (test-mode). This way of running the tests does not use libvirt, VBMC or any similar tools, which means it can be used as a way of ensuring that we do not depend on them. However, we do not run the fixture provider tests in CI currently. This issue is for adding the fixture tests to CI to ensure that we do not accidentally tie the tests to the CI platform.

The tests with fixture provider is quite light weight. We could consider running them in a github action. Otherwise Jenkins is always an option.

/kind feature

maxrantil:
Implement lightweight fixture-based tests to ensure CI platform flexibility.

Fixes #1406

### DIFF
--- a/.github/workflows/e2e-fixture-test.yml
+++ b/.github/workflows/e2e-fixture-test.yml
@@ -1,0 +1,35 @@
+name: E2E Fixture Test
+
+on:
+  pull_request:
+    branches: [ main ]
+    paths-ignore:
+      - '**/*.md'
+      - 'docs/**'
+      - '.gitignore'
+      - 'hack/*.sh'
+      - 'LICENSE'
+      - 'SECURITY_CONTACTS'
+      - 'DCO'
+      - 'OWNERS'
+
+permissions:
+  contents: read
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+
+    - name: Build BMO e2e Docker Image
+      env:
+        IMG: quay.io/metal3-io/baremetal-operator:e2e
+      run: make docker
+
+    - name: Set Up Environment and Run BMO e2e Tests
+      env:
+        E2E_CONF_FILE: ${GITHUB_WORKSPACE}/test/e2e/config/fixture.yaml
+        USE_EXISTING_CLUSTER: "false"
+      run: make test-e2e
--- a/test/e2e/basic_provisioning_test.go
+++ b/test/e2e/basic_provisioning_test.go
@@ -78,13 +78,6 @@ var _ = Describe("Provisioning", func() {
 		err = clusterProxy.GetClient().Create(ctx, &bmh)
 		Expect(err).NotTo(HaveOccurred())
 
-		By("Waiting for the BMH to be in registering state")
-		WaitForBmhInProvisioningState(ctx, WaitForBmhInProvisioningStateInput{
-			Client: clusterProxy.GetClient(),
-			Bmh:    bmh,
-			State:  metal3api.StateRegistering,
-		}, e2eConfig.GetIntervals(specName, "wait-registering")...)
-
 		By("Waiting for the BMH to become available")
 		WaitForBmhInProvisioningState(ctx, WaitForBmhInProvisioningStateInput{
 			Client: clusterProxy.GetClient(),
--- a/test/e2e/config/ironic.yaml
+++ b/test/e2e/config/ironic.yaml
@@ -33,7 +33,7 @@ intervals:
   default/wait-registering: ["1m", "5s"]
   inspection/wait-registration-error: ["1m", "5s"]
   external-inspection/wait-available: ["20s", "1s"]
-  default/wait-inspecting: ["2m", "10s"]
+  default/wait-inspecting: ["2m", "2s"]
   default/wait-available: ["10m", "1s"]
   default/wait-deployment: ["5m", "1s"]
   default/wait-namespace-deleted: ["10m", "1s"]
--- a/test/e2e/basic_ops_test.go
+++ b/test/e2e/basic_ops_test.go
@@ -82,13 +82,6 @@ var _ = Describe("basic", func() {
 		err = clusterProxy.GetClient().Create(ctx, &bmh)
 		Expect(err).NotTo(HaveOccurred())
 
-		By("waiting for the BMH to be in registering state")
-		WaitForBmhInProvisioningState(ctx, WaitForBmhInProvisioningStateInput{
-			Client: clusterProxy.GetClient(),
-			Bmh:    bmh,
-			State:  metal3api.StateRegistering,
-		}, e2eConfig.GetIntervals(specName, "wait-registering")...)
-
 		By("waiting for the BMH to become available")
 		WaitForBmhInProvisioningState(ctx, WaitForBmhInProvisioningStateInput{
 			Client: clusterProxy.GetClient(),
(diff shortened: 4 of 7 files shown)


.github/workflows/e2e-fixture-test.yml
```diff
@@ -0,0 +1,33 @@
...
+
+    - name: Install Dependencies
+      run: |
+        "${GITHUB_WORKSPACE}/hack/e2e/ensure_kubectl.sh"
+        "${GITHUB_WORKSPACE}/hack/e2e/ensure_cmctl.sh"
+
+    - name: Build BMO e2e Docker Image
+      run: |
+        export IMG=quay.io/metal3-io/baremetal-operator:e2e
+        make docker
```
lentzi90, line 27: I'm not sure how big difference this is really, but I think it is more common to set the environment this way in github actions.
```suggestion
    - name: Build BMO e2e Docker Image
      env:
        IMG: quay.io/metal3-io/baremetal-operator:e2e
      run: make docker
```

.github/workflows/e2e-fixture-test.yml
```diff
@@ -0,0 +1,33 @@
...
+
+    - name: Set Up Environment and Run BMO e2e Tests
+      run: |
+        export E2E_CONF_FILE=${GITHUB_WORKSPACE}/test/e2e/config/fixture.yaml
+        export USE_EXISTING_CLUSTER="false"
```
lentzi90, line 32: Same for these environment variables

test/e2e/basic_provisioning_test.go
```diff
@@ -78,13 +78,6 @@ var _ = Describe("Provisioning", func() {
 		err = clusterProxy.GetClient().Create(ctx, &bmh)
 		Expect(err).NotTo(HaveOccurred())
 
-		By("Waiting for the BMH to be in registering state")
-		WaitForBmhInProvisioningState(ctx, WaitForBmhInProvisioningStateInput{
-			Client: clusterProxy.GetClient(),
-			Bmh:    bmh,
-			State:  metal3api.StateRegistering,
-		}, e2eConfig.GetIntervals(specName, "wait-registering")...)
-
```
lentzi90, line 87: FYI (other reviewers): This was removed because the `registering` state passes so quickly that it is sometimes missed in the fixture tests. I think it is not important to check that we go through this state since it is the next state we really care about (`inspecting` or `available`).

.github/workflows/e2e-fixture-test.yml
```diff
@@ -0,0 +1,31 @@
...
+      - 'docs/**'
+      - '.gitignore'
+      - 'LICENSE'
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
```
lentzi90, line 17: We should pin this, similar to how we have done in other workflows: https://github.com/metal3-io/baremetal-operator/blob/74bfd27aeb2490263013bc39ca10a39223b9a11f/.github/workflows/dependabot.yml#L21

```suggestion
    - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
```

lentzi90 (commented):
/lgtm

lentzi90:
/retitle 🌱 Add github workflow for e2e fixture tests

lentzi90 (commented):
/lgtm

.github/workflows/e2e-fixture-test.yml
```diff
@@ -0,0 +1,31 @@
...
+
+on:
+  pull_request:
+    branches: [ main ]
+    paths-ignore:
+      - '**/*.md'
+      - 'docs/**'
+      - '.gitignore'
+      - 'LICENSE'
+
```
tuminoid, line 11: We need `permissions` set on the main level, and/or in the `jobs`.

tuminoid (commented):
/hold

.github/workflows/e2e-fixture-test.yml
maxrantil: Thanks for the feedback! It's now implemented.

.github/workflows/e2e-fixture-test.yml
```diff
@@ -0,0 +1,34 @@
+name: E2E Fixture Test
+
+on:
+  pull_request:
+    branches: [ main ]
+    paths-ignore:
+      - '**/*.md'
+      - 'docs/**'
+      - '.gitignore'
+      - 'LICENSE'
```
tuminoid, line 10: We should also add:
- SECURITY_CONTACTS
- DCO
- OWNERS

So it might be easier to add a matching rule for files that are all caps and have no suffix, unless changes to `PROJECT` has effect on e2e.

It would also be nice to ignore `hack/*.sh`, since they're for Prow/local.

tuminoid:
/unhold
Permissions issue is fixed.

lentzi90 (commented):
/lgtm

.github/workflows/e2e-fixture-test.yml
```diff
@@ -0,0 +1,38 @@
...
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+
+    - name: Install Dependencies
+      run: "${GITHUB_WORKSPACE}/hack/e2e/ensure_cmctl.sh"
```
mquhuy, line 27: With https://github.com/metal3-io/baremetal-operator/pull/1440 merged, we don't need this anymore.

maxrantil: this aswell?
    - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
    ??

tuminoid: Checkout action checks out repo, so it is needed.

mquhuy: No no. I meant only the `ensure_cmctl.sh` part (the script doesn't exist anymore anyway). Clicked on line 27 but don't know why it included the other 3 lines, sorry :sweat_smile:

lentzi90: Ah good point @mquhuy ! Yes we should remove the call to `ensure_cmctl.sh`

lentzi90 (commented):
/lgtm

test/e2e/config/ironic.yaml
```diff
@@ -33,7 +33,7 @@ intervals:
   default/wait-registering: ["1m", "5s"]
   inspection/wait-registration-error: ["1m", "5s"]
   external-inspection/wait-available: ["20s", "1s"]
-  default/wait-inspecting: ["2m", "10s"]
+  default/wait-inspecting: ["2m", "2s"]
```
kashifest, line 36: any reason to change this in this PR ?

maxrantil: Using 10-second intervals means that it can miss the state, so I changed it to 2 to one make the test more robust.

maxrantil: You are right to ask because I now see that this is in the ironic config, but after trying it out I noticed some flakes here so that why this fix is in this PR even that it has nothing to do with fixture.

test/e2e/basic_provisioning_test.go
dtantsur: Yep, I remember having this concern when looking at the e2e patch.

PR #1438, opened 2023-11-09
⚠️ Kustomizations usable off the shelf

### LINKED ISSUE #1361: Make Ironic and BMO kustomizations usable off the shelf
**User Story**

As a user/operator I would like to use the BMO/Ironic kustomizations as bases for my overlays to simplify deployment.

**Detailed Description**

It is not currently possible to use the kustomizations directly as is (ref https://github.com/metal3-io/metal3-docs/issues/298). For TLS and basic-auth variants it is required to add files that kustomize will use to generate secrets. This means that the Ironic/BMO kustomizations are only usable if you clone the repo and make changes to it first.

I would want to make these kustomizations usable directly as is. Then it would be possible to reference them in overlays as URLs even.
To make this possible I see two options:

1. (preferred) Remove secret generators, [such as this](https://github.com/metal3-io/baremetal-operator/blob/bdfa1bd1bd5e3d2a98534ab1fd22616677c1568a/config/basic-auth/default/kustomization.yaml#L8-L16), from the kustomizations. All the rest stays as is. This means that the manifests are created just like before but without the secrets. The user is then responsible for adding the secrets in any way they see fit. This can be for example, by creating an overlay that adds the secrets, or using the external secrets operator. Until the secrets are created, the pods will stay pending with explanation that it is missing a secret to mount. This is a common pattern and should be easy to understand for the user.
2. Keep the secret generators, but provide example credentials for them so that the kustomizations can be built without changes. It would be up to the user to overwrite the example credentials in their overlay or through other means. I would avoid this alternative since it is easy to make mistakes and end up with example credentials used in the actual deployment. It is much better (IMO) to get the pending pods as in alternative 1.

An easy way to check if this issue has been resolved is to do `kustomize build` for each kustomization under `config` and `ironic-deployment`. They should all succeed.

/kind feature

lentzi90:
The PR is marked as :warning: breaking because the structure of the kustomizations changed. In practice I do not think it should cause problems, because the kustomizations that we change here were not usable without modifications (adding credentials) anyway. The `deploy.sh` script should work just as before.

This restructures the BMO kustomizations to make use of components
(similar to what was already done for the Ironic kustomizations).

It also makes the kustomizations usable "off the shelf" by removing the
secret generation. Previously it was necessary to clone the repository
and add the relevant files (e.g. by running `tools/deploy.sh`) before it
was possible to build the kustomizations that included secrets generated
from the files. With this commit, the secrets will NOT be generated as
part of the kustomizations and they can thus be used as is. However, the
user is instead required to add the necessary secrets.

The secrets can be added in any way. One common way is that the users
creates their own overlays that add the secrets. These kustomizations
can be based on and use the kustomizations in the repo since they are
now buildable without changes. Another way to add the secrets is via the
external secrets operator.

Finally, basic-auth and TLS is enabled for the e2e tests. This also works as an example for how to make use of the kustomizations and for what secrets are needed.

Fixes #1361, #1371

Note for reviewers:

Most changes are just from moving files around to get the kustomize components structure. Some files needed slight modifications because of the new structure though and this can sometimes show as a large diff.


test/e2e/e2e_suite_test.go
lentzi90, line 1: The changes here is just to install Ironic before BMO, since BMO will try to mount the CA certificate that comes with Ironic.

lentzi90:
Tests looks good!
/hold cancel

ironic-deployment/overlays/e2e/ironic-auth-config
lentzi90, line 1: These credentials would be used in the e2e tests. I'm curious to hear opinions on this. My reasoning for including them in the repo is that

1. They act like a concrete example that users can look at to understand both how to create the secrets and what to put in them
2. The tests should (ideally) not expose Ironic to anything outside the CI system anyway and they are relatively short lived

I'm open for other solutions though. We could generate them automatically for the tests and static examples can be provided without being part of the CI.

/cc @tuminoid

config/default/kustomization.yaml
lentzi90, line 1: This was basically copied to `config/base/kustomization.yaml` and then simplified by using that as base

tools/deploy.sh
lentzi90, line 1: The changes to this file can be summarized like this:

1. Gereration of credentials was moved up so that the files will be in place when we start building the kustomizations.
2. A temporary overlay is used for BMO (just like Ironic was already doing).

config/base/kustomization.yaml
lentzi90, line 1: This was copied from `config/default` and then the paths were adapted. The only other change is that this does NOT create the ironic configmap. This is because the content of the configmap depends on the environment so the user will anyway need to change it in almost all cases.

ironic-deployment/overlays/e2e/ironic-auth-config
tuminoid: I agree that we need to have proper usable examples for the users, so they don't need to guess what they're supposed to look like etc, but I'd prefer generating them, and just documenting the proper format. 

1. While the risk of exposure is small, it still exists, better remove it
2. There is always someone who will deploy them in production anyways ...
3. If we generate them, we also test our codebase versus accidental hardcoding, ie. make it more realistic check

If in step 3 we use random username/password, it'll make debugging/developing a bit harder though, so let's document it well where/how to fetch the creds.

What do you think?

config/README.md
```diff
@@ -0,0 +1,31 @@
...
+- **components** - In here you will find re-usable kustomize components for TLS
+  and basic-auth.
+   - **basic-auth** - Enable basic authentication. Note that the basic-auth
+      component is missing the actual credentials. This is on purpose, to make
+      sure that the user is setting the password instead of using sample
+      credentials. The required secrets are:
+      - `ironic-credentials`
+      - `ironic-inspector-credentials`
+   - **tls** - Enable TLS. A CA certificate is needed here to verify the
+      connection to Ironic. If you deploy Ironic together with Ironic in a
```
kashifest, line 24: Do you mean 
```suggestion
      connection to Ironic. If you deploy Ironic-inspector together with Ironic in a
```

config/overlays/e2e/kustomization.yaml
```diff
@@ -24,3 +23,20 @@ patches:
...
 - name: quay.io/metal3-io/baremetal-operator
   newTag: e2e
+
+# We cannot use suffix hashes since the kustomizations we build on
+# cannot be aware of what suffixes we add.
+generatorOptions:
+  disableNameSuffixHash: true
+
+# NOTE: These credentials are only meant for testing!
+# DO NOT USE IN PRODUCTION
```
kashifest, line 33: hummm, wondering how many would read this 😄

ironic-deployment/overlays/e2e/ironic-auth-config
lentzi90: It does make sense to me, but makes things a bit more complicated. I need to think about how to do it.

kashifest (commented):
Great work. lgtm. I will leave it open for some time for review before I stamp approval.

config/README.md
lentzi90: Thanks good find. I mean Ironic together with BMO. Will fix!

config/overlays/e2e/kustomization.yaml
lentzi90: Well it should be already obvious that the overlay is for e2e :smile: 
But on a serious note, as discussed with @tuminoid above, I will most likely change this so that we generate the credentials on the fly instead of putting them here.

kashifest: Yeah saw that after I commented , I agree with @tuminoid

ironic-deployment/overlays/e2e/ironic-auth-config
lentzi90: Done, and looks like it is working!
I have not added any more documentation yet. Perhaps that can be in a separate PR? We don't have any docs for it at the moment anyway so no degradation...

The credentials are created in the `ci-e2e.sh` script. This means that outside of CI, users can still choose to set up a static overlay for use with a baremetal lab or similar.

kashifest:
@dtantsur @honza @zaneb PTAL. This patch will help us get rid of deploy script and allow us to use conventional kustomize overlays.

lentzi90:
> @dtantsur @honza @zaneb PTAL. This patch will help us get rid of deploy script and allow us to use conventional kustomize overlays.

To be clear, it will still be possible to use the deploy script for those who want to do so. I don't expect people in general want to use a script to deploy though, and this will make it possible to just use the kustomizations as is, without modifications or scripts.

lentzi90:
> I wonder how much easier this change becomes if we throw Inspector out in the window.

Not that much I think. I would still want to shuffle around the files like I do here. The Ironic and inspector auth configs would reduce to just one I guess, but that is just a few lines.

PR #1440, opened 2023-11-10
✨ Replace uses of kubectl and cmctl binaries in e2e

### LINKED ISSUE #1373: Remove use of `os.Exec` in e2e tests
We currently use `cmctl` and `kubectl` directly though `os.Exec` in the e2e tests. It would be nice to rewrite this to avoid `os.Exec`. If we can avoid using these binaries we will not have to download them and keep scripts to ensure they are installed in the e2e test.

Most of the logic should be pretty straight forward to translate into go. For building kustomizations, we can use the [krusty](https://pkg.go.dev/sigs.k8s.io/kustomize/api/krusty) library. We can also take inspiration from CAPI on [how to install cert-manager](https://github.com/kubernetes-sigs/cluster-api/blob/main/cmd/clusterctl/client/cluster/cert_manager.go#L168).

/kind feature

mquhuy:
Fixes #1373 

I chose different implementations from what was suggested in #1373:
- ~~Instead of using `krusty` package, I copied [the kustomize build workflow used by capi](https://github.com/Nordix/cluster-api/blob/f44c0ebd94aa7409af0db22bfce0953e2fd702ab/test/framework/clusterctl/repository.go#L226). This is only for simplicity, and to be aligned with CAPI. The `krusty` package seemed to be not too complicated, but imo the simplier, the better.~~
Changed (back) to `krusty` after discussion with @lentzi90.
- The `cert-manager` workflow from CAPI takes into control many of CAPI internal data structures, therefore is not easy to replicate. For our usecase, I think simply download the cert-manager manifests, apply and ~~wait for all deployments to be ready~~ perform a dry-run of self-signed issuer&certificate creation on cert-manager webhook should be enough.

I'm fully open to all suggestion/negotiation.


test/e2e/common.go
```diff
@@ -208,3 +209,18 @@ func WaitForBmhInPowerState(ctx context.Context, input WaitForBmhInPowerStateInp
...
+	}
+	kustomize := exec.NewCommand(
+		exec.WithCommand(kustomizePath),
+		exec.WithArgs("build", source))
+	stdout, stderr, err := kustomize.Run(ctx)
+	if err != nil {
+		return []byte{}, errors.Wrapf(err, "failed to execute kustomize: %s", stderr)
+	}
+	return stdout, nil
+}
```
lentzi90, line 226: This does not get rid of the `os.exec`, we just use it through a nicer wrapper. IMO this is not worth it. We still need the binary (kustomize instead of kubectl) and we still execute it from the go code.

test/e2e/cert_manager.go
```diff
@@ -0,0 +1,36 @@
...
+
+	"github.com/pkg/errors"
+)
+
+func checkCertManagerAPI(clusterProxy framework.ClusterProxy) error {
+	certManagerAPIVersion := "cert-manager.io/v1"
+	clientset := clusterProxy.GetClientSet()
+	_, err := clientset.Discovery().ServerResourcesForGroupVersion(certManagerAPIVersion)
+	return err
+}
```
lentzi90, line 19: This is nice!

test/e2e/cert_manager.go
```diff
@@ -0,0 +1,36 @@
...
+
+func installCertManager(ctx context.Context, clusterProxy framework.ClusterProxy, cmVersion string) error {
```
lentzi90, line 21: Nice and simple, I like it!

test/e2e/e2e_suite_test.go
```diff
@@ -103,26 +101,40 @@ var _ = SynchronizedBeforeSuite(func() []byte {
...
+				"cert-manager-cainjector",
+			}
+			for _, name := range cmDeploymentNames {
+				cmDeployment := &v1.Deployment{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      name,
+						Namespace: "cert-manager",
+					},
+				}
+				framework.WaitForDeploymentsAvailable(ctx, framework.WaitForDeploymentsAvailableInput{
```
lentzi90, line 122: This is unfortunately not enough. The webhook is not always available and registered as fast as the deployment becomes available.
Could you add a function that waits and retries the same way as clusterctl and cmctl does it? It should try to create a self signed issuer and certificate again and again until it succeeds.

hack/ci-e2e.sh
```diff
@@ -20,8 +21,9 @@ BMO_E2E_EMULATOR=${BMO_E2E_EMULATOR:-"sushy-tools"}
 "${REPO_ROOT}/hack/e2e/ensure_go.sh"
 export PATH="${PATH}:/usr/local/go/bin"
 "${REPO_ROOT}/hack/e2e/ensure_minikube.sh"
-"${REPO_ROOT}/hack/e2e/ensure_kubectl.sh"
-"${REPO_ROOT}/hack/e2e/ensure_cmctl.sh"
```
lentzi90, line 24: Please remove the script also :slightly_smiling_face:

test/e2e/common.go
mquhuy: Thank you. Now I realize I chose this method simply because CAPI is using it (the exec wrapper is from them too), but I agree that it's not worth it. I'll change to using one of the packages.

test/e2e/e2e_suite_test.go
mquhuy: Ah, okay. I misunderstood your idea before xD

test/go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/metal3-io/baremetal-operator/test
 go 1.20
 
 require (
+	github.com/cert-manager/cert-manager v1.10.0
```
lentzi90, line 6: The version seems strange considering that we have v1.13 in the e2e config.

test/e2e/e2e_suite_test.go
```diff
@@ -103,26 +101,30 @@ var _ = SynchronizedBeforeSuite(func() []byte {
...
 		// Install BMO
 		By("Installing BMO")
 		kustomization := e2eConfig.GetVariable("BMO_KUSTOMIZATION")
-		cmd := exec.Command("kubectl", "--kubeconfig", clusterProxy.GetKubeconfigPath(), "apply", "-k", kustomization) //nolint:gosec
-		output, err := cmd.CombinedOutput()
-		Expect(err).NotTo(HaveOccurred(), fmt.Sprintf("%s", output))
+		manifest, err := buildKustomizeManifest(kustomization)
+		Expect(err).NotTo(HaveOccurred())
+		err = clusterProxy.Apply(ctx, manifest)
+		Expect(err).NotTo(HaveOccurred())
```
lentzi90, line 127: Nice!

test/e2e/e2e_suite_test.go
```diff
@@ -103,26 +101,30 @@ var _ = SynchronizedBeforeSuite(func() []byte {
...
-			Expect(err).NotTo(HaveOccurred(), fmt.Sprintf("%s", output))
+			cmVersion := e2eConfig.GetVariable("CERT_MANAGER_VERSION")
+			err = installCertManager(ctx, clusterProxy, cmVersion)
+			Expect(err).NotTo(HaveOccurred())
+
+			By("Waiting for cert-manager webhook")
+			Eventually(func() bool {
+				err := checkCertManagerWebhook(ctx, clusterProxy)
+				return err == nil
+			}, e2eConfig.GetIntervals("default", "wait-available")...).Should(BeTrue())
```
lentzi90, line 114: I may be missing something, but is it not possible to just return the error directly and way for `Succeed` instead?
```suggestion
			Eventually(func() err {
				return checkCertManagerWebhook(ctx, clusterProxy)
			}, e2eConfig.GetIntervals("default", "wait-available")...).Should(Succeed())
```

lentzi90 (commented):
Great job! Just some minor comments

lentzi90:
This is the test to trigger for e2e. The other tests are not relevant :sweat_smile: 
/metal3-bmo-e2e-test

test/go.mod
mquhuy: Yes. Basically using newer versions would trigger upgrades in many of the dependencies, which'd make everything a mess :-s 1.10 is the newest one I could find that didn't trigger too many breaks.

test/e2e/e2e_suite_test.go
mquhuy: Thanks a lot! That's the way to go :sweat_smile: I'm still pretty not familiar yet with the CAPI test framework. Initially used `BeNil()` as the function returns an error, and the wait just never worked :sweat_smile:

test/go.mod
mquhuy: My apologies. Apparently I tried many other versions but not 1.13 :sweat_smile: It works.

lentzi90: Ok I see. We can take this for now probably, but at some point we will need to bump it

lentzi90:
> It seems that kubectl still needs to be there 😅

Any idea why?

mquhuy:
/metal3-bmo-e2e-test

> > It seems that kubectl still needs to be there 😅
> 
> Any idea why?

It's because [CAPI test framework `ClusterProxy.Apply()` uses `exec.KubectlApply`](https://github.com/kubernetes-sigs/cluster-api/blob/main/test/framework/cluster_proxy.go#L249), which [uses `kubectl` command](https://github.com/kubernetes-sigs/cluster-api/blob/main/test/framework/exec/kubectl.go#L29).

I can make a separate apply function to get rid of the `kubectl`, but I'm wondering if we should stick to the test framework? Or maybe we can make that contribution to CAPI instead?

lentzi90:
I see! It is fine like this I would say. No need to make it unnecessarily complicated :slightly_smiling_face:

test/e2e/config/ironic.yaml
```diff
@@ -27,6 +27,7 @@ variables:
   BOOT_MAC_ADDRESS: "00:60:2f:31:81:01"
   IMAGE_URL: "http://192.168.222.1/cirros-0.6.2-x86_64-disk.img"
   IMAGE_CHECKSUM: "c8fc807773e5354afe61636071771906"
+  CERT_MANAGER_VERSION: "v1.13.0"
```
lentzi90, line 30: This needs to be added in `fixture.yaml` also!
(We will have automated tests to check the fixture tests when #1437 is merged.)

mquhuy: Done.

lentzi90 (commented):
/lgtm

Rozzii (approved):

kashifest (commented):
/approve

PR #1448, opened 2023-11-17
🌱 Implement Helper Functions for E2E Tests

### LINKED ISSUE #1436: Implement Helper Functions to Streamline E2E Testing
**User Story**
As a developer, I would like to implement specific helper functions for the end-to-end (E2E) tests to streamline the testing process and shorten the code. This enhancement is aimed at improving code readability, maintainability, and potentially reducing the time for test execution.

**Detailed Description**
I propose the implementation of the following functionalities:

Function for Creating Secrets: A function that encapsulates the logic of creating secrets in the test setup. This function will help in reducing repetitive code across different test cases where secret creation is a common prerequisite.

Enhancement of `WaitForBmhInProvisioningState`: Revise this function to accept an optional parameter for disallowed states. This will add flexibility in tests by allowing us to specify states that the BMH should avoid while reaching a target state.

Development of a New Function`WaitForBmhDeleted` to Handle BMH Deletion: Similar to the above to handle more complex conditions or states during the deletion process. 

/kind feature

maxrantil:
**What this PR does**:
As the title says

Fixes #1436

### DIFF
--- a/test/e2e/basic_ops_test.go
+++ b/test/e2e/basic_ops_test.go
@@ -20,6 +20,7 @@ import (
 var _ = Describe("basic", func() {
 	var (
 		specName       = "basic-ops"
+		secretName     = "bmc-credentials"
 		namespace      *corev1.Namespace
 		cancelWatches  context.CancelFunc
 		bmcUser        string
@@ -47,18 +48,7 @@ var _ = Describe("basic", func() {
 
 	It("should control power cycle of BMH though annotations", func() {
 		By("creating a secret with BMH credentials")
-		bmcCredentials := corev1.Secret{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      "bmc-credentials",
-				Namespace: namespace.Name,
-			},
-			StringData: map[string]string{
-				"username": bmcUser,
-				"password": bmcPassword,
-			},
-		}
-		err := clusterProxy.GetClient().Create(ctx, &bmcCredentials)
-		Expect(err).NotTo(HaveOccurred())
+		CreateBMHCredentialsSecret(ctx, clusterProxy.GetClient(), namespace.Name, secretName, bmcUser, bmcPassword)
 
 		By("creating a BMH")
 		bmh := metal3api.BareMetalHost{
@@ -79,7 +69,7 @@ var _ = Describe("basic", func() {
 				BootMACAddress: bootMacAddress,
 			},
 		}
-		err = clusterProxy.GetClient().Create(ctx, &bmh)
+		err := clusterProxy.GetClient().Create(ctx, &bmh)
 		Expect(err).NotTo(HaveOccurred())
 
 		By("waiting for the BMH to become available")
--- a/test/e2e/basic_provisioning_test.go
+++ b/test/e2e/basic_provisioning_test.go
@@ -15,7 +15,6 @@ import (
 	"sigs.k8s.io/cluster-api/test/framework"
 	"sigs.k8s.io/cluster-api/util"
 	"sigs.k8s.io/cluster-api/util/patch"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	metal3api "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
 )
@@ -23,6 +22,7 @@ import (
 var _ = Describe("BMH Provisioning and Annotation Management", func() {
 	var (
 		specName       = "provisioning-ops"
+		secretName     = "bmc-credentials"
 		namespace      *corev1.Namespace
 		cancelWatches  context.CancelFunc
 		bmcUser        string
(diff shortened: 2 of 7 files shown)


test/e2e/common.go
```diff
@@ -153,21 +154,77 @@ func (c *Config) GetVariable(varName string) string {
...
 		key := types.NamespacedName{Namespace: input.Bmh.Namespace, Name: input.Bmh.Name}
 		g.Expect(input.Client.Get(ctx, key, &bmh)).To(Succeed())
-		g.Expect(bmh.Status.Provisioning.State).To(Equal(input.State))
+
+		currentStatus := bmh.Status.Provisioning.State
+
+		// Check if the current state matches any of the undesired states
+		if isUndesiredState(currentStatus, input.UndesiredStates) {
+			g.Expect(fmt.Sprintf("Encountered undesired state: %s", currentStatus)).ToNot(Succeed())
+			return
```
lentzi90, line 188: This is not checking temporary undesired states. Eventually will just keep going after the return unless you tell it to `StopTrying`.
```suggestion
			StopTrying(fmt.Sprintf("BMH is in an unexpected state: %s", currentStatus)).Now()
```

test/e2e/common.go
```diff
@@ -153,21 +154,77 @@ func (c *Config) GetVariable(varName string) string {
...
+			return true
+		}
+		g.Expect(err).NotTo(HaveOccurred())
+
+		currentStatus := bmh.Status.Provisioning.State
+
+		// If the BMH is found, check for undesired states.
+		if isUndesiredState(currentStatus, input.UndesiredStates) {
+			g.Expect(fmt.Sprintf("Encountered undesired state: %s", currentStatus)).ToNot(Succeed())
+			return false
```
lentzi90, line 221:
```suggestion
			StopTrying(fmt.Sprintf("BMH is in an unexpected state: %s", currentStatus)).Now()
```

lentzi90 (commented):
/lgtm

test/e2e/basic_provisioning_test.go
```diff
@@ -80,7 +69,7 @@ var _ = Describe("BMH Provisioning and Annotation Management", func() {
 				BootMACAddress: bootMacAddress,
 			},
 		}
-		err = clusterProxy.GetClient().Create(ctx, &bmh)
+		err := clusterProxy.GetClient().Create(ctx, &bmh)
 		Expect(err).NotTo(HaveOccurred())
 
 		By("Waiting for the BMH to become available")
```
mboukhalfa, line 75: UndesiredStates skipped down here on purpose

test/e2e/basic_ops_test.go
```diff
@@ -47,18 +48,7 @@ var _ = Describe("basic", func() {
...
-				"username": bmcUser,
-				"password": bmcPassword,
-			},
-		}
-		err := clusterProxy.GetClient().Create(ctx, &bmcCredentials)
-		Expect(err).NotTo(HaveOccurred())
+		CreateBMHCredentialsSecret(ctx, clusterProxy.GetClient(), namespace.Name, secretName, bmcUser, bmcPassword)
 
 		By("creating a BMH")
 		bmh := metal3api.BareMetalHost{
```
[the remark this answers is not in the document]
maxrantil: Do you mean "creating a BMH" into a function?

test/e2e/basic_provisioning_test.go
maxrantil: Excuse me, but I don't get what you mean?

test/e2e/basic_ops_test.go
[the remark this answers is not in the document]
lentzi90: I'm not convinced this would be a good idea. We would basically need parameters for everything since we are testing all configurations for the BareMetalHosts. Then it is easier to just create it directly

test/e2e/basic_provisioning_test.go
lentzi90: I guess he means that we do not use `UndesiredStates` here. It should be fine. We do not need to check all the time that the BMHs don't do weird things like jump from registering to deprovisioning and then available :joy: 
It is very good that we can check in some cases, like when we are supposed to skip deprovisioning, but here I don't see much need for it.

mboukhalfa: Yeah that what I thought that maybe the function fail to cache registering then if it is some other state then we fail but agree it is not that important as it will fail at the end

test/e2e/basic_ops_test.go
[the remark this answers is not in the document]
mboukhalfa: @maxrantil  Yeah that what I was thinking so far seems we are mostly just changing annotations if we can pass them as parameters would reduce a bit the duplication but probably in other cases this becomes parameters for everything as @lentzi90 stated.

Issue #1372, opened 2023-09-28
Add e2e test for automated disk cleaning


lentzi90:
We should test the [automated disk cleaning](https://book.metal3.io/bmo/automated_cleaning) in e2e.

Rozzii:
This ticket has been labeled as stale multiple times, I am moving this to frozen.
/lifecycle frozen

PR #1449, opened 2023-11-17
WIP: 🌱 Test live-ISO provisioning in e2e and confirm successful boot

### LINKED ISSUE #1429: Test live-ISO i e2e
**User Story**

As a developer I would like to ensure that virtuamedia/live-ISO provisioning works by testing it in the e2e tests.

**Detailed Description**

Check [this](https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/test/e2e/live_iso_test.go) for how it was implemented in CAPM3. This test makes more sense in BMO and we should implement it here. (Afterwards we can discuss if we should also remove it from CAPM3.)
My hope is that we can implement this test similar to how we already support ipmi and redfish by just changing the BMC details, but I suspect there is more to it. At least we need an ISO...

/kind feature

maxrantil:
Fixes #1429


test/e2e/live_iso_test.go
```diff
@@ -0,0 +1,143 @@
...
+			Creator:   clusterProxy.GetClient(),
+			ClientSet: clusterProxy.GetClientSet(),
+			Name:      fmt.Sprintf("%s-%s", specName, util.RandomString(6)),
+			LogFolder: artifactFolder,
+		})
+	})
+
+	It("should provision a BMH with live ISO and then deprovision it", func() {
+		By("Creating a secret with BMH credentials")
+		bmcCredentials := corev1.Secret{
```
dtantsur, line 48: nit: the BMC handling will happen in all tests in a similar fashion, we should probably move it to a common place

test/e2e/live_iso_test.go
```diff
@@ -0,0 +1,143 @@
...
+		bmh := metal3api.BareMetalHost{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      specName,
+				Namespace: namespace.Name,
+				Annotations: map[string]string{
+					metal3api.InspectAnnotationPrefix:   "disabled",
+					metal3api.HardwareDetailsAnnotation: hardwareDetails,
+				},
+			},
+			Spec: metal3api.BareMetalHostSpec{
```
dtantsur, line 71: You may want to disable cleaning as well to avoid wasting time on it (assuming there is already a test with cleaning enabled).

test/e2e/live_iso_test.go
```diff
@@ -0,0 +1,143 @@
...
+				BMC: metal3api.BMCDetails{
+					Address:         liveISOAddress,
+					CredentialsName: "bmc-credentials",
+				},
+				BootMode:       metal3api.Legacy,
+				BootMACAddress: bootMacAddress,
+				Image: &metal3api.Image{
+					URL:          liveISOImageURL,
+					Checksum:     "",
+					ChecksumType: "",
```
dtantsur, line 82: nit: two empty fields are redundant

test/e2e/live_iso_test.go
```diff
@@ -0,0 +1,143 @@
...
+					metal3api.HardwareDetailsAnnotation: hardwareDetails,
+				},
+			},
+			Spec: metal3api.BareMetalHostSpec{
+				Online: true,
+				BMC: metal3api.BMCDetails{
+					Address:         liveISOAddress,
+					CredentialsName: "bmc-credentials",
+				},
+				BootMode:       metal3api.Legacy,
```
dtantsur, line 77: Why so? Both Metal3 and Ironic default to UEFI, I'd rather avoid using something non-default in tests.

test/e2e/live_iso_test.go
```diff
@@ -0,0 +1,143 @@
...
+					ChecksumType: "",
+					DiskFormat:   StringPtr("live-iso"),
+				},
+			},
+		}
+
+		err = clusterProxy.GetClient().Create(ctx, &bmh)
+		Expect(err).NotTo(HaveOccurred())
+
+		By("Waiting for BMH to transition to Provisioning state")
```
dtantsur, line 91: When cleaning is off, this is going to be quite brief (essentially, the time it takes to download your ISO). 1 minutes may be too long.

test/e2e/live_iso_test.go
lentzi90: Good point! I have an issue open (https://github.com/metal3-io/baremetal-operator/issues/1372) for testing disk cleaning, but I realize now that by default it has of course already cleaned the disks in the provisioning test we have :smile:

test/e2e/live_iso_test.go
maxrantil: Agree, have a look: [here](https://github.com/metal3-io/baremetal-operator/pull/1448)

test/e2e/live_iso_test.go
maxrantil: Thanks for pointing this out, I got it from another test that was made earlier so that's why I took it. I'll take it away.

lentzi90: It is probably my fault. I think I copied this straight from some old metal3-dev-env template :see_no_evil:

maxrantil:
This issue is now implemented with SSH instead and can be found [here](https://github.com/metal3-io/baremetal-operator/pull/1453)

Issue #1443, opened 2023-11-13
E2e: enable parallel testing


lentzi90:
**User Story**

As a developer I would like to run the e2e tests in parallel for faster feedback.

**Detailed Description**

Currently we just create 1 VM to be used as BMH in the e2e tests. This obviously means that we cannot run tests in parallel.
The tests do not require that much resources and we would easily be able to run more of them in parallel if we can create a system for multiple VMs and BMHs.

Side note: We have previously discussed a related but different topic offline. I want to clarify this so there is no mistake. This issue is about running the tests in parallel, NOT running the same tests with multiple different protocols (e.g. redfish and ipmi). The intention here is to provide a pool of BMC/BMHs that can be used by the ginkgo processes when running multiple in parallel. This means that we would most likely not have control over what test is executed with what BMC/BMH.
End of side note.

I suggest to restructure the e2e config so that we have a list of BMCs instead of just one (similar to the list of images).

```yaml
bmcs:
- user: foo
  password: bar
  address: ipmi://1.2.3.4
```

From the list, we can then pick one BMC per ginkgo process by using the [process index](https://pkg.go.dev/github.com/onsi/ginkgo/v2#GinkgoParallelProcess). This can be done in the `BeforeSuite` as described [here](https://onsi.github.io/ginkgo/#patterns-for-parallel-integration-specs). The idea would be to create N VMs and then run ginkgo with N processes.

The more tests we add, the greater the need for this becomes.

/kind feature

PR #1451, opened 2023-11-20
🌱 Add BMO e2e upgrade test

### LINKED ISSUE #1430: BMO e2e upgrade test
**User Story**

As a developer I would like to test upgrading BMO in e2e fashion to ensure that the upgrade paths function as expected.

**Detailed Description**

We have proper releases of BMO now and should support upgrades from one to the next. The goal with this issue is to test these upgrade paths. For example, we need to ensure that it is possible to upgrade from the last release to the latest code (that would become the next release). We currently have a test like this [in CAPM3](https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/test/e2e/upgrade_baremetal_operator.go) that can be used for inspiration.

This is a different kind of test than the other e2e tests we have so far. Here we are focusing on making changes to BMO itself rather than the BMHs. We will still need to do some operations with the BMHs to ensure that BMO is functioning before/after the upgrade.
Because of this, it will be tricky to keep the same flexibility that we try to achieve in the rest of the tests. For example, we MUST be able to change the BMO image and CRDs in order to do the upgrade. We may not need to touch the rest of the config though, so it *may* still be possible to use an existing cluster with pre-applied configuration in order to run these tests. I would say we do not have to support this use case though if it becomes too problematic.

/kind feature

mquhuy:
Fixes #1430


test/e2e/config/ironic.yaml
```diff
@@ -32,6 +32,7 @@ variables:
   IMAGE_URL: "http://192.168.222.1/cirros-0.6.2-x86_64-disk.img"
   IMAGE_CHECKSUM: "c8fc807773e5354afe61636071771906"
   CERT_MANAGER_VERSION: "v1.13.0"
+  RUN_UPGRADE_TESTS: "true"
```
lentzi90, line 35: I don't think we should control what tests to run in this way. We can use `GINKGO_SKIP` to skip specific tests if needed.

test/e2e/config/fixture.yaml
```diff
@@ -25,6 +25,7 @@ variables:
   IMAGE_URL: "http://192.168.222.1/cirros-0.6.2-x86_64-disk.img"
   IMAGE_CHECKSUM: "c8fc807773e5354afe61636071771906"
   CERT_MANAGER_VERSION: "v1.13.0"
+  RUN_UPGRADE_TESTS: "false"
```
lentzi90, line 28: Why do we need to skip it?

test/e2e/upgrade_test.go
lentzi90, line 1: This is a good start but it needs some work! Now we do not check that BMO is working at all before or after. In fact we do not even have Ironic installed so there is no way it can work. I would like some kind of test that it is actually working, e.g. create a BMH and make it `available`, then upgrade BMO and check that it is possible to provision the BMH.

hack/ci-e2e.sh
```diff
@@ -41,8 +43,8 @@ virsh -c qemu:///system attach-interface --domain minikube --mac="52:54:00:6c:3c
 minikube stop
 minikube start
 
-# Load the BMO e2e image into it
-minikube image load quay.io/metal3-io/baremetal-operator:e2e
+# Load the local BMO images into it
```
lentzi90, line 46:
```suggestion
# Load the local BMO image into it
```

hack/ci-e2e.sh
```diff
@@ -122,6 +124,24 @@ docker run --name image-server-e2e -d \
...
+  git clone --branch "${BMO_RELEASE_BRANCH}" --single-branch "${BMO_ORIGIN_ADDRESS}" "${BMO_RELEASE_ROOT}"
+fi
+
+export BMO_UPGRADE_FROM_KUSTOMIZATION="${BMO_RELEASE_ROOT}/config/overlays/e2e"
+
+if [[ ! -d "${BMO_UPGRADE_FROM_KUSTOMIZATION}" ]]; then
+  mkdir -p "${BMO_UPGRADE_FROM_KUSTOMIZATION}"
+  cp -r "${BMO_E2E_KUSTOMIZATION}/"* "${BMO_UPGRADE_FROM_KUSTOMIZATION}/"
+  sed -i "s/newTag: e2e/newTag: ${BMO_RELEASE_BRANCH}/" "${BMO_UPGRADE_FROM_KUSTOMIZATION}/kustomization.yaml"
+fi
```
lentzi90, line 143: Clever! We may need to change this a bit when basic-auth and TLS is added but it should be doable, don't worry about it.

test/e2e/config/fixture.yaml
mquhuy: Because it'd require the release manifests. Do we really need to run upgrade test with fixture setup?

test/e2e/config/ironic.yaml
mquhuy: Yes. But my point here is that we should never run upgrade tests with fixture config. We can always add GINKGO_SKIP if we want to.

test/e2e/upgrade_test.go
mquhuy: Thank you. I see the issue now (was mostly copying this part from capm3 e2e). Luckily ironic is already there, so perhaps this requirement is not very difficult to achieve.

hack/ci-e2e.sh
mquhuy: Thank you. I just realized, however, that the sed should be outside of the if case :sweat_smile:

test/e2e/config/ironic.yaml
lentzi90: But my point is why should we not? As the test is now, I don't see why it would not work with fixture. In fact I think fixture is the only place it can work currently without Ironic

test/e2e/config/fixture.yaml
lentzi90: Maybe we do not need but why should we not? The fixture tests are very fast and cheap to run. If we can run the upgrade test for them I don't see a reason to skip them

mquhuy: Thank you. The reason I didn't include upgrade test in fixture was because we don't run `ci-e2e.sh` before running fixture, so there wouldn't be any `BMO_KUSTOMIZE_UPGRADE_FROM`. But maybe we can arrange some other solution for that.

lentzi90: Ah that makes sense, and this is actually why the fixture tests are so good! They show how things would go if running outside of CI in a way. So yes, something else is needed in this case. The user would be expected to provide `BMO_KUSTOMIZE_UPGRADE_FROM` and for fixture we would need to do the same. I think the solution for now would be to add something to the github workflow. Perhaps, copy the repo to a temp folder and checkout an older release?

hack/ci-e2e.sh
```diff
@@ -11,6 +11,9 @@
 set -eux
 
 REPO_ROOT=$(realpath "$(dirname "${BASH_SOURCE[0]}")/..")
+BMO_IMAGE_NAME="quay.io/metal3-io/baremetal-operator"
+
+export E2E_CONF_FILE="${REPO_ROOT}/test/e2e/config/ironic.yaml"
```
maxrantil, line 16: is this needed? Can't we take it away?

hack/ci-e2e.sh
```diff
@@ -126,6 +136,8 @@ docker run --name image-server-e2e -d \
 set +e
 
 # Run the e2e tests
+export E2E_CONF_FILE="${REPO_ROOT}/test/e2e/config/ironic.yaml"
+export USE_EXISTING_CLUSTER="true"
```
maxrantil, line 140: These 2 lines are like this by default, isn't it?

mquhuy: Yes. The assumption is that since you run `ci-e2e.sh`, this is what you want. Of course it's just to be extra careful, nothing would happen in CI if we take it away.

hack/ci-e2e.sh
mquhuy: Yes. Initially it wasn't needed because we have the default in Makefile, but now we also use this env var in the `hack/e2e/ensure_kustomization_paths.sh`, it should be explicit.

hack/e2e/ensure_kustomization_paths.sh
lentzi90, line 1: I think I understand what you are trying to do here. Something like this, right?

1. Find the BMO_KUSTOMIZATION used for the tests
2. Clone the repo and checkout an older version
3. Use the same BMO_KUSTOMIZATION from the older version if it exists, otherwise copy it
4. Set an older image to match the older version

There are a few issues with this though. Reading the value from the e2e config can easily break because the config is supposed to be overridable with env vars. We should not read the config outside of the go code. In these scripts, we should rather set values for the e2e config than read from it.

The BMO_KUSTOMIZATION we use in CI is in the repo, but we cannot really trust that this is the case in all situations. I think we should rather have a separate BMO_UPGRADE_FROM_KUSTOMIZATION. Then users/developers can also set their own. For CI we need to create that kustomization and point it to use an older version of the manifests. What I mean with this is that we should not script it, just add another overlay `config/overlays/e2e-upgrade-from` or something like this.

For example it could look like this:

```yaml
kind: Kustomization

resources:
- https://github.com/metal3-io/baremetal-operator/config?ref=release-0.4

# Add any special config needed. Probably same as we have in overlays/e2e
```

test/e2e/e2e_suite_test.go
```diff
@@ -73,7 +73,7 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 	e2eConfig = LoadE2EConfig(configPath)
 
 	if useExistingCluster {
-		kubeconfigPath = os.Getenv("KUBECONFIG")
+		kubeconfigPath = os.Getenv("KUBECONFIG_E2E")
```
lentzi90, line 76: I don't think we should change this. It will be very surprising that we do not use the standard environment variable.

test/e2e/upgrade_test.go
```diff
@@ -0,0 +1,407 @@
...
+				Creator:   upgradeClusterProxy.GetClient(),
+				ClientSet: upgradeClusterProxy.GetClientSet(),
+				Name:      fmt.Sprintf("%s-%s", specName, util.RandomString(6)),
+				LogFolder: artifactFolder,
+			})
+		}
+	})
+
+	It("Should upgrade BMO to latest version", func() {
+		if e2eConfig.GetVariable("DEPLOY_IRONIC") != "false" {
```
lentzi90, line 306: Why do we skip this if not deploying Ironic? This makes no sense to me

lentzi90 (commented):
Good progress!
I hope all my comments are not discouraging!

test/e2e/e2e_suite_test.go
mquhuy: Hmm. That's true, but I don't know if it's supported in CAPI framework. I can investigate.
Anw, could you elaborate why not using standard env var is a bad thing here?

hack/e2e/ensure_kustomization_paths.sh
```diff
@@ -0,0 +1,36 @@
...
+BMO_E2E_KUSTOMIZATION=$(realpath "${REPO_ROOT}/test/e2e/${BMO_E2E_KUSTOMIZATION_PATH}")
+if [[ ! -d "${BMO_E2E_KUSTOMIZATION}" ]]; then
+  echo "Kustomization path not exists: ${BMO_E2E_KUSTOMIZATION}"
+  exit 2
+fi
+
+BMO_E2E_KUSTOMIZATION_RELATIVE_PATH=$(realpath --relative-to="${REPO_ROOT}" "${BMO_E2E_KUSTOMIZATION}")
+
+if [[ ! -d "${BMO_RELEASE_ROOT}" ]]; then
+  git clone --branch "${BMO_RELEASE_BRANCH}" --single-branch "${BMO_ORIGIN_ADDRESS}" "${BMO_RELEASE_ROOT}"
```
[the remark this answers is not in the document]
mquhuy: Yes, right. I forgot about the `git worktree` syntax. Will change that.

hack/e2e/ensure_kustomization_paths.sh
```diff
@@ -0,0 +1,36 @@
...
+BMO_RELEASE_ROOT="/tmp/bmo-${BMO_RELEASE_BRANCH}"
+BMO_ORIGIN_ADDRESS="https://github.com/metal3-io/baremetal-operator.git"
+
+BMO_E2E_KUSTOMIZATION_PATH=$(grep BMO_KUSTOMIZATION "${E2E_CONF_FILE}" | awk -F: '{ print $2 }' | tr -d '" ')
+if [[ "${BMO_E2E_KUSTOMIZATION_PATH}" == "" ]]; then
+  echo "BMO_KUSTOMIZATION not defined in ${E2E_CONF_FILE}"
+  exit 2
+fi
+
+BMO_E2E_KUSTOMIZATION=$(realpath "${REPO_ROOT}/test/e2e/${BMO_E2E_KUSTOMIZATION_PATH}")
```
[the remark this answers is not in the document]
mquhuy: Thanks. I didn't know that.

hack/ci-e2e.sh
```diff
@@ -11,6 +11,9 @@
 set -eux
 
 REPO_ROOT=$(realpath "$(dirname "${BASH_SOURCE[0]}")/..")
+BMO_IMAGE_NAME="quay.io/metal3-io/baremetal-operator"
```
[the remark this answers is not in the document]
mquhuy: Ah, yes. Initially I intended to use this var for the UPGRADE_FROM image as well, but turned out we don't need to fetch it beforehands, minikube can do that automatically.

hack/ci-e2e.sh
```diff
@@ -25,24 +28,31 @@ export PATH="${PATH}:/usr/local/go/bin"
...
+export KUBECONFIG_E2E_UPGRADE="/tmp/kubeconfig-bmo-e2e-upgrade"
+
+# Set up minikube
+for profile in "bmo-e2e" "bmo-e2e-upgrade"; do
+  export KUBECONFIG="/tmp/kubeconfig-${profile}"
+  minikube start --profile $profile --driver=kvm2
+
+  # Attach baremetal-e2e interface to minikube with specific mac.
+  # This will give minikube a known reserved IP address that we can use for Ironic
+  virsh -c qemu:///system attach-interface --domain "${profile}" --mac="52:54:00:6c:3c:01" \
```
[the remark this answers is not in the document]
mquhuy: Hmm. In my understanding we never have an BMH in both clusters, so this should be ok.

hack/e2e/ensure_kustomization_paths.sh
mquhuy: Thank you. My issue with this is that we would fix the UPGRADE_FROM version. Is it a requirement that we will test other releases, e.g. release-0.3?

test/e2e/e2e_suite_test.go
lentzi90: It has nothing to do with CAPI. It is standard in Kubernetes itself, e.g. `kubectl` use it and most other CLIs that deal with the Kubernetes API, like `clusterctl`, `cmctl` and `argocd`.
Since this has become a standard that so many use, I would as a user expect that the same variable is used also here. Using the standard variable means less need for documenting the behavior since it "works as expected". If we use a custom variable it must definitely be documented.

If you need an example for how to implement this with contexts, check CAPO.
- [Here](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/fc7ba6edd13fe05dec9431c643c710a45dbf4f31/test/e2e/data/e2e_conf.yaml#L166) is the variable for what context to use
- [This is how they use it](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/fc7ba6edd13fe05dec9431c643c710a45dbf4f31/test/e2e/shared/cluster.go#L72)

hack/ci-e2e.sh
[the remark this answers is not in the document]
lentzi90: Yes that is expected since we upgrade from an existing release that already has a public image. The reason we need the `e2e` image normally is because we must run the tests with the very latest code from the current PR :slightly_smiling_face:

hack/e2e/ensure_kustomization_paths.sh
mquhuy: And also it seems kustomize doesn't like uses of url/abspath in resources.

hack/ci-e2e.sh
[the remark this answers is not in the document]
lentzi90: That may be true for now, but it is not something we can count on in the future. I'm also afraid that this will make the connection to Ironic quite unstable. If both minikubes have the same IP and Mac then probably BMO will send requests to both of them. One is working and the other is just sitting idle without Ironic even running. So BMO would get connection refused every time a request goes to the wrong minikube

hack/e2e/ensure_kustomization_paths.sh
lentzi90: We may want to test other releases also, but for now I think it is enough if we focus on release-0.4. If we need release-0.3 later we can just add a kustomization for that as well.

> And also it seems kustomize doesn't like uses of url/abspath in resources.

Can you explain what you mean? What error do you get? This kustomization works fine for me

mquhuy: Ah. My bad. Your exact syntax did work :smile: 
I was using the url from github, `kustomize` was able to download the file, but failed to catch the subsequent resources that file was referring to.

test/e2e/upgrade_test.go
mquhuy: Hmm. Why would we want it if we wouldn't provision anyway?

test/e2e/e2e_suite_test.go
mquhuy: Hmm. Sorry I'm still not following :sweat_smile: The example you showed seems to just check if the context is correct, and does nothing if it's not.
In our case, we will have to switch back and forth between the contexts.

My point regarding CAPI is that `framework.NewClusterProxy()` just takes a kubeconfig and doesn't care about the context.
And don't we get non-default kubeconfig files anyway if we create clusters with `bootstrap.CreateKindBootstrapClusterAndLoadImages()`?

test/e2e/upgrade_test.go
lentzi90: Deploying Ironic has nothing to do with if we provision or not. It is possible for the user to USE_EXISTING_CLUSTER=true + DEPLOY_IRONIC=false so that they can set up their own cluster and Ironic and then run the tests against it.
Then we have the fixture tests, which is running the tests without Ironic completely, since BMO runs in test-mode. Then we don't deploy Ironic but we still go through all tests and we provision the BMHs. It will of course not do anything "real", but it is still a test that BMO behaves correctly and makes changes to the BMHs.

test/e2e/e2e_suite_test.go
lentzi90: > In our case, we will have to switch back and forth between the contexts.

Good point. It may be easier with two different kubeconfigs.

> And don't we get non-default kubeconfig files anyway if we create clusters with bootstrap.CreateKindBootstrapClusterAndLoadImages()?

The point is not about where the file is, it is that `KUBECONFIG` is a standard environment variable. I think it is better that we keep it as is. For the new kubeconfig used for upgrade, you can add another variable. However, this *must* be in the e2e config, just like all other variables we add.
I would also suggest that if this "BMO upgrade" variable is not set, we should fallback to `KUBECONFIG` also in this case. It will probably not work in most cases, but it would be the expected "default". In some cases it could even work, e.g. if the user was running only the BMO upgrade test.

test/e2e/upgrade_test.go
mquhuy: What you said is true in general, but in this case isn't the point to check if the BMO can be upgraded from an older version? For the fixture test, we wouldn't be able to check if we can provision the BMH after upgrade anyway, so imo adding the BMH in the first place is a bit overkill, we should handle that in other tests. (I wouldn't mind adding it though if you think we should do so)

I think `USE_EXISTING_CLUSTER=true + DEPLOY_IRONIC=false` should not be allowed. That would make things too complicated. There are certain tasks (like provisioning) that wouldn't work without an Ironic, and we'd need separate configs for: "I don't have an ironic, but I don't need to install one, just ignore tests that requires an ironic", and "I have ironic already, don't attempt to install, just perform all the tests".

lentzi90: There is no difference between fixture and ironic for the tests. We just run the same tests with BMO in test mode or BMO + Ironic. BMO in test mode can provision BMHs and that is what we test. Of course it is just fake when BMO is in test mode but that is beside the point.

> I think USE_EXISTING_CLUSTER=true + DEPLOY_IRONIC=false should not be allowed.

Why? In the tests you do not need to think about how Ironic is deployed. Just assume that it is there and working. If a user specifies `DEPLOY_IRONIC=false` it is up to them to make sure it works. They could be running Ironic outside the cluster, or we could be running with BMO in test mode, which does not require Ironic to begin with.

> There are certain tasks (like provisioning) that wouldn't work without an Ironic

All tasks will require Ironic in the real world, when actually interacting with baremetal servers. That does not mean that we must always deploy Ironic, or that Ironic is needed in all cases. We can run BMO in test mode for example, or Ironic can be set up manually outside the cluster.

test/e2e/e2e_suite_test.go
mquhuy: That makes senses, but a bit tricky to implement. `ci-e2e.sh` needs to know this config to be able to populate the cluster, but we've also discussed that it shouldn't read the config file. I think we need a new solution for this thing :-?

test/e2e/upgrade_test.go
mquhuy: Got it. I forgot that we have the bmo test mode in fixture. Anw, my main point was that since this test was to check if BMO could be upgraded, we don't necessarily need to add all details in the fixture test (which could run for a while), but I'm not against running everything either.

test/e2e/e2e_suite_test.go
lentzi90: Think about `ci-e2e.sh` as "the user". It can create the cluster and configure the tests to use it. It does not need to know where the kubeconfig is, because it creates it and decides where to put it. In pseudo code

```
# ci-e2e.sh
create_bmo_upgrade_cluster()
save_kubeconfig(/tmp/bmo-upgrade-kubeconfig.yaml)

export BMO_UPGRADE_KUBECONFIG=/tmp/bmo-upgrade-kubeconfig.yaml
```

```yaml
# e2e-config.yaml
variables:
  BMO_UPGRADE_KUBECONFIG: "overridden by environment variable"
```

```
# In test code
if config:= GetVariable(BMO_UPGRADE_KUBECONFIG)
  use_kubeconfig(config)
else
  use_kubeconfig(KUBECONFIG if defined, otherwise ~/.kube/config)

```

test/e2e/upgrade_test.go
lentzi90: It is much harder to make special cases. Just run all tests but with different config for fixture. It does not hurt to test BMO upgrades in the fixture tests, in fact it could be very useful! We get a very quick test for checking if changes to the CRDs work and if BMO's webhook can handle conversions between new/old CRD versions.

test/e2e/config/ironic.yaml
```diff
@@ -49,3 +60,5 @@ intervals:
   default/wait-deprovisioning: ["1m", "10ms"]
   default/wait-deleted: ["20s", "10ms"]
   default/wait-secret-deletion: ["1m", "1s"]
+  upgrade/wait-provisioning: ["1m", "1s"]
+  upgrade/wait-provisioned: ["10m", "1s"]
```
lentzi90, line 64: Why not reuse `default/wait-provisioning` and `default/wait-provisioned`? There should not be a difference in the time required. I would rather use those existing values and update them if needed

test/e2e/e2e_suite_test.go
```diff
@@ -47,6 +47,8 @@ var (
 	// clusterProvider manages provisioning of the cluster to be used for the e2e tests.
 	// Please note that provisioning will be skipped if e2e.use-existing-cluster is provided.
 	clusterProvider bootstrap.ClusterProvider
+
+	kubeconfigPath string
```
lentzi90, line 51: This is a bit dangerous to make global. We update it if/when creating a kind cluster and this then affects the variable across all tests. For now I don't think it is an issue since we use the clusterproxy instead of the kubeconfig in most tests. However, I don't want to rely on luck here. There could be tests in the future when we use the kubeconfig. So please put it back as a local variable and use a separate local variable for the BMO upgrade test. Then they won't affect each other.

test/e2e/upgrade_test.go
```diff
@@ -0,0 +1,408 @@
...
+		bmoUpgradeFromImage := clusterctl.ContainerImage{
+			Name:         bmoUpgradeFromImageName,
+			LoadBehavior: clusterctl.MustLoadImage,
+		}
+
+		if useExistingCluster {
+			k8sUpgradeContextName := e2eConfig.GetVariable("KUBECONFIG_UPGRADE_CONTEXT_NAME")
+			By(fmt.Sprintf("Using kubeconfig context: %s for upgrade tests", k8sUpgradeContextName))
+			err := SwitchKubeconfigContext(kubeconfigPath, k8sUpgradeContextName)
+			Expect(err).NotTo(HaveOccurred())
```
lentzi90, line 210: Please check here also if the variable is set or not, and don't change the context if the variable is not set.

test/e2e/upgrade_test.go
```diff
@@ -0,0 +1,408 @@
...
+			}
+			if upgradeClusterProvider != nil {
+				upgradeClusterProvider.Dispose(ctx)
+			}
+		}
+		if useExistingCluster {
+			k8sContextName := e2eConfig.GetVariable("KUBECONFIG_DEFAULT_CONTEXT_NAME")
+			By(fmt.Sprintf("Switch back to kubeconfig context: %s", k8sContextName))
+			err := SwitchKubeconfigContext(kubeconfigPath, k8sContextName)
+			Expect(err).NotTo(HaveOccurred())
```
lentzi90, line 404: Shouldn't we always switch the context here, not just when using existing cluster?
I think it would be even better to check what context was used before the test and switch back to that, instead of switching to what the variable is. If the variable is not set this breaks.

test/e2e/upgrade_test.go
```diff
@@ -0,0 +1,408 @@
...
+			err := SwitchKubeconfigContext(kubeconfigPath, k8sUpgradeContextName)
+			Expect(err).NotTo(HaveOccurred())
+		} else {
+			By("Creating a separate cluster for upgrade tests")
+			upgradeClusterProvider = bootstrap.CreateKindBootstrapClusterAndLoadImages(ctx, bootstrap.CreateKindBootstrapClusterAndLoadImagesInput{
+				Name:   "bmo-e2e-upgrade",
+				Images: append(e2eConfig.Images, bmoUpgradeFromImage),
+			})
+			Expect(upgradeClusterProvider).ToNot(BeNil(), "Failed to create a cluster")
+			kubeconfigPath = clusterProvider.GetKubeconfigPath()
```
lentzi90, line 218: Note! This was using the wrong cluster!
```suggestion
			kubeconfigPath = upgradeClusterProvider.GetKubeconfigPath()
```

I noticed in the fixture test, that it just updated cert-manager, BMO and Ironic instead of deploying from scratch. See https://github.com/metal3-io/baremetal-operator/actions/runs/7033795294/job/19140414038?pr=1451#step:4:59

lentzi90 (commented):
Very good progress!
The real test with Ironic seem to work great! The fixture test was using the wrong cluster though :open_mouth:

test/e2e/config/ironic.yaml
mquhuy: In my trials, there were a bunch of times these waits are not sufficient, so I added those. Maybe I should update the values to the default ones instead?

test/e2e/upgrade_test.go
```diff
@@ -0,0 +1,408 @@
...
+				Getter:     upgradeClusterProxy.GetClient(),
+				Deployment: bmoDeployment,
+			}, e2eConfig.GetIntervals("default", "wait-deployment")...)
+			// Set up log watcher
+			framework.WatchDeploymentLogsByName(ctx, framework.WatchDeploymentLogsByNameInput{
+				GetLister:  upgradeClusterProxy.GetClient(),
+				Cache:      upgradeClusterProxy.GetCache(ctx),
+				ClientSet:  upgradeClusterProxy.GetClientSet(),
+				Deployment: bmoDeployment,
+				LogPath:    filepath.Join(artifactFolder, "logs", bmoDeployment.GetNamespace()),
```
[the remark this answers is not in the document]
mquhuy: Ah ha. When I made this I assumed the name of the test would be in the (test) namespace. Forgot that was not the case for bmo/ironic namespace.

test/e2e/e2e_suite_test.go
```diff
@@ -77,6 +78,9 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 		if kubeconfigPath == "" {
 			kubeconfigPath = os.Getenv("HOME") + "/.kube/config"
 		}
+		k8sContextName := e2eConfig.GetVariable("KUBECONFIG_DEFAULT_CONTEXT_NAME")
+		err := SwitchKubeconfigContext(kubeconfigPath, k8sContextName)
+		Expect(err).NotTo(HaveOccurred())
```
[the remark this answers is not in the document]
mquhuy: Yeah that makes senses. It would make the errors a bit more difficult to spot if user forgets to set the kubeconfig into correct context beforehands, but perhaps that's a bit of an odd case.

test/e2e/upgrade_test.go
mquhuy: I was a bit scared that if this is not set, then the same cluster will be used for both upgrade test and the other tests, which will basically screw up the whole thing without any error returned. But well, perhaps that's on the user xD

test/e2e/config/ironic.yaml
```diff
@@ -47,8 +61,8 @@ intervals:
   default/wait-namespace-deleted: ["10m", "1s"]
   ironic/wait-deployment: ["10m", "2s"]
   default/wait-power-state: ["10m", "100ms"]
-  default/wait-provisioning: ["20s", "1s"]
-  default/wait-provisioned: ["5m", "1s"]
+  default/wait-provisioning: ["1m", "1s"]
+  default/wait-provisioned: ["10m", "1s"]
```
maxrantil, line 65: Does it take more then 5 minutes to get into provisioned state sometimes? My tests has been around 2-3 minutes.

mquhuy: Well. In my trials things get a bit slower in the second cluster. Initially I used a separate wait time for the upgrade tests, but decided to just update the default waits after discussing with Lennart.
I'd say 10m maybe a bit overkill, but 5m would be too short.

test/e2e/upgrade_test.go
```diff
@@ -0,0 +1,413 @@
...
+			if kubeconfigPath == "" {
+				kubeconfigPath = os.Getenv("HOME") + "/.kube/config"
+			}
+			var err error
+			originalKubeConfigContext, err = GetCurrentKubeconfigContextName(kubeconfigPath)
+			Expect(err).NotTo(HaveOccurred())
+			k8sUpgradeContextName := e2eConfig.GetVariable("KUBECONFIG_UPGRADE_CONTEXT_NAME")
+			if (k8sUpgradeContextName != originalKubeConfigContext) && (k8sUpgradeContextName != "") {
+				By(fmt.Sprintf("Using kubeconfig context: %s for upgrade tests", k8sUpgradeContextName))
+				err = SwitchKubeconfigContext(kubeconfigPath, k8sUpgradeContextName)
```
lentzi90, line 212: We cannot switch contexts like this in the tests. It will break if we run tests in parallel (since all tests use the currently selected context). I'm sorry I guided you towards context, but I think we need to use separate kubeconfigs.

Makefile
```diff
@@ -51,7 +51,7 @@ GINKGO_TIMEOUT ?= 2h
 GINKGO_POLL_PROGRESS_AFTER ?= 60m
 GINKGO_POLL_PROGRESS_INTERVAL ?= 5m
 E2E_CONF_FILE ?= $(ROOT_DIR)/test/e2e/config/ironic.yaml
-USE_EXISTING_CLUSTER ?= true
+USE_EXISTING_CLUSTER ?= false
```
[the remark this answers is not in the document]
lentzi90: Please update the [readme](https://github.com/metal3-io/baremetal-operator/blob/main/test/e2e/README.md) to reflect this

test/e2e/config/ironic.yaml
```diff
@@ -14,11 +16,23 @@ images:
...
   IRONIC_KUSTOMIZATION: "../../ironic-deployment/overlays/e2e"
+  KUBECONFIG_DEFAULT_CONTEXT_NAME: "bmo-e2e"
+
+  # These settings are for the separate cluster that are used for upgrade tests
+  UPGRADE_DEPLOY_IRONIC: "true"
+  UPGRADE_DEPLOY_BMO: "true"
+  UPGRADE_DEPLOY_CERT_MANAGER: "true"
+  BMO_UPGRADE_FROM_KUSTOMIZATION: "../../config/overlays/e2e-release-0.4"
+  BMO_RELEASE_BRANCH: "release-0.4"
+  UPGRADE_IRONIC_KUSTOMIZATION: "../../ironic-deployment/overlays/e2e-upgrade"
```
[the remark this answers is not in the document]
mquhuy: Hmm. I didn't want to use `IRONIC_UPGRADE_FROM_KUSTOMIZATION`, because we (for now) are not upgrading ironic in this test. And using the same one that we use for the main cluster doesn't work either, since we use different ip for the baremetal network (which changes the ip used by ironic, defined in `ironic.env`).

The name `UPGRADE_IRONIC_KUSTOMIZATION` means that it's the same as `IRONIC_KUSTOMIZATION`, but for the upgrade cluster. Do you think `IRONIC_UPGRADE_FROM_KUSTOMIZATION` is acceptable and better here?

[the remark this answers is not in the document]
lentzi90: > And using the same one that we use for the main cluster doesn't work either, since we use different ip for the baremetal network (which changes the ip used by ironic, defined in ironic.env).

Good point, I did not think about this. But how does it work then when we upgrade BMO? Don't we switch to the other Ironic then?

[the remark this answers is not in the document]
mquhuy: > Good point, I did not think about this. But how does it work then when we upgrade BMO? Don't we switch to the other Ironic then?

Ah yes. I didn't think about that either. It does look like it switches to the other ironic :laughing: Let's see if I could find a way around this env file.

hack/ci-e2e.sh
```diff
@@ -26,90 +27,98 @@ export PATH="${PATH}:/usr/local/go/bin"
...
+    --model virtio --source "${profile}" --type network --config
+
+  # Restart minikube to apply the changes
+  minikube stop --profile "${profile}"
+  KUBECONFIG="${kubeconfig}" minikube start --profile "${profile}"
+
+  # Load the local BMO image into it
+  minikube --profile "${profile}" image load "${BMO_IMAGE}"
+
+  VM_NAME="bmo-e2e-${idx}"
```
lentzi90, line 62: Could we have a separate name for the VM used for the upgrade? Now they look so similar so it is easy to think that both are used with all tests.

hack/e2e/bmo-e2e-net.xml
```diff
@@ -1,15 +1,15 @@
...
       <port start='1024' end='65535'/>
     </nat>
   </forward>
-  <bridge name='metal3'/>
+  <bridge name='bmo-e2e'/>
   <ip address='192.168.222.1' netmask='255.255.255.0'>
     <dhcp>
       <range start='192.168.222.2' end='192.168.222.199'/>
-      <host mac='52:54:00:6c:3c:01' name='minikube' ip='192.168.222.199'/>
+      <host mac='52:54:00:6c:3c:01' name='ironic-0' ip='192.168.222.199'/>
```
lentzi90, line 12: Why change this? The host is minikube, not Ironic. Ironic just happens to be running in minikube and using this address.

hack/e2e/bmo-e2e-upgrade-net.xml
```diff
@@ -0,0 +1,17 @@
...
+  <forward mode='nat'>
+    <nat>
+      <port start='1024' end='65535'/>
+    </nat>
+  </forward>
+  <bridge name='bmo-e2e-upgrade'/>
+  <ip address='192.168.223.1' netmask='255.255.255.0'>
+    <dhcp>
+      <range start='192.168.223.2' end='192.168.223.199'/>
+      <host mac='52:54:00:6c:3c:02' name='ironic-1' ip='192.168.223.199'/>
```
lentzi90, line 12: I would suggest minikube-upgrade or something like this here instead.
```suggestion
      <host mac='52:54:00:6c:3c:02' name='minikube-upgrade' ip='192.168.223.199'/>
```

hack/e2e/ironic-e2e.env
lentzi90, line 1: Why is this moved?

ironic-deployment/overlays/e2e/.gitignore
lentzi90, line 1: Should not be needed.

test/e2e/common.go
```diff
@@ -359,3 +363,106 @@ func IsBootedFromDisk(client *ssh.Client) (bool, error) {
...
+	kustomization := input.Kustomization
+	ironicEnvFile := input.IronicEnvFile
+	clusterProxy := input.ClusterProxy
+	if ironicEnvFile != "" {
+		dst := filepath.Join(kustomization, "ironic.env")
+		err = CopyFile(ironicEnvFile, dst)
+		if err != nil {
+			return err
+		}
+	}
```
lentzi90, line 431: Why? Just let the file be in the kustomization so it can be applied directly.

test/e2e/common.go
```diff
@@ -359,3 +363,106 @@ func IsBootedFromDisk(client *ssh.Client) (bool, error) {
...
+	ClusterProxy        framework.ClusterProxy
+	WaitForDeployment   bool
+	WatchDeploymentLogs bool
+	DeploymentName      string
+	DeployNS            string
+	LogPath             string
+	WaitIntervals       []interface{}
+}
+
+func InstallIronicBMO(ctx context.Context, input *InstallIronicBMOInput) error {
```
lentzi90, line 420: I don't like this function. It does strange things without explanation (copying files). It is very specific (ironic/bmo) instead of generic (e.g. apply and wait for deployment). There is no doc string.
Don't get me wrong, I can see the usefulness of a function like this, but it should then be generic and well documented.

test/e2e/common.go
```diff
@@ -359,3 +363,106 @@ func IsBootedFromDisk(client *ssh.Client) (bool, error) {
 
 	return isDisk, nil
 }
+
+func DeploymentRolledOut(ctx context.Context, clusterProxy framework.ClusterProxy, name string, namespace string, desiredGeneration int64) bool {
```
lentzi90, line 367: Please add a doc string to this

test/e2e/config/fixture.yaml
```diff
@@ -11,17 +13,29 @@ images:
   loadBehavior: tryLoad
 
 variables:
+  # These settings are for the "default" cluster, which is used for all tests except upgrade ones
   DEPLOY_IRONIC: "false"
   DEPLOY_BMO: "true"
   DEPLOY_CERT_MANAGER: "true"
   BMO_KUSTOMIZATION: "../../config/overlays/fixture"
+  IRONIC_ENV_FILE: ""
```
lentzi90, line 21: This should not be needed.
```suggestion
```

test/e2e/config/fixture.yaml
```diff
@@ -11,17 +13,29 @@ images:
...
+  BMC_ADDRESS: ipmi://192.168.222.1:16230
+  BOOT_MAC_ADDRESS: "00:60:2f:31:81:01"
+
+  # These settings are for the separate cluster that are used for upgrade tests
+  UPGRADE_DEPLOY_IRONIC: "false"
+  UPGRADE_DEPLOY_BMO: "true"
+  UPGRADE_DEPLOY_CERT_MANAGER: "true"
+  UPGRADE_BMO_KUSTOMIZATION_FROM: "../../config/overlays/fixture-release-0.4"
+  UPGRADE_IRONIC_ENV_FILE: ""
```
lentzi90, line 30:
```suggestion
```

test/e2e/config/ironic.yaml
```diff
@@ -14,21 +16,38 @@ images:
...
   DEPLOY_IRONIC: "true"
   DEPLOY_BMO: "true"
   DEPLOY_CERT_MANAGER: "true"
   BMO_KUSTOMIZATION: "../../config/overlays/e2e"
   IRONIC_KUSTOMIZATION: "../../ironic-deployment/overlays/e2e"
-  EXPECTED_HOST_NAME: "bmo-e2e-0"
-  # Test credentials. The tests will create a BMH with these.
-  BMC_USER: admin
-  BMC_PASSWORD: password
+  IRONIC_ENV_FILE: "../../hack/e2e/ironic-e2e.env"
```
lentzi90, line 25: Should not be needed
```suggestion
```

test/e2e/config/ironic.yaml
```diff
@@ -14,21 +16,38 @@ images:
...
   BOOT_MAC_ADDRESS: "00:60:2f:31:81:01"
+  EXPECTED_HOST_NAME: "bmo-e2e-0"
+
+  # These settings are for the separate cluster that are used for upgrade tests
+  UPGRADE_DEPLOY_IRONIC: "true"
+  UPGRADE_DEPLOY_BMO: "true"
+  UPGRADE_DEPLOY_CERT_MANAGER: "true"
+  UPGRADE_BMO_KUSTOMIZATION_FROM: "../../config/overlays/e2e-release-0.4"
+  UPGRADE_KUBECONFIG_FILE: "/tmp/kubeconfig-e2e-upgrade"
+  UPGRADE_IRONIC_ENV_FILE: "../../hack/e2e/ironic-e2e-upgrade.env"
```
lentzi90, line 40:
```suggestion
```

test/e2e/e2e_suite_test.go
```diff
@@ -120,60 +122,36 @@ var _ = SynchronizedBeforeSuite(func() []byte {
...
-			Deployment: ironicDeployment,
-		}, e2eConfig.GetIntervals("ironic", "wait-deployment")...)
-		// Set up log watcher
-		framework.WatchDeploymentLogsByName(ctx, framework.WatchDeploymentLogsByNameInput{
-			GetLister:  clusterProxy.GetClient(),
-			Cache:      clusterProxy.GetCache(ctx),
-			ClientSet:  clusterProxy.GetClientSet(),
-			Deployment: ironicDeployment,
-			LogPath:    filepath.Join(artifactFolder, "logs", ironicDeployment.GetNamespace()),
+		err := InstallIronicBMO(ctx, &InstallIronicBMOInput{
```
lentzi90, line 125: If you want to refactor this (and I suggest not to do it in this PR), please make the function more generic and get rid of the env file input. Just apply the kustomization as it is.

lentzi90 (commented):
Alright! Now the logs look great and the test is doing what I was expecting/hoping! Great job!
I still have a lot of comments though :sweat_smile: 

In general, please try to keep it simple! Refactoring here and there makes it harder to review and understand. Many small PRs are easier and faster to get approved.
From the increased complexity (and resource requirements!) I'm now honestly wondering if we should split out the upgrade into its own `ci-e2e-upgrade.sh`. :thinking: 
With the structure in this PR it will be very hard and complex to do https://github.com/metal3-io/baremetal-operator/issues/1443 and we will also get a big merge conflict with #1483 
What do you think? Would it make sense to have a separate file and separate job for upgrades?

test/e2e/e2e_suite_test.go
mquhuy: The env file is needed, because we need different ips for different clusters. BMO (and later on, ironic) can be upgraded from one version to another, but the ip should stay the same. Otoh, the same BMO version can be installed in both clusters, but the ips need to be different.
Not having the env file configurable will require us to have different kustomizations for all of these cases, which I find too much for just simple changes.

hack/ci-e2e.sh
mquhuy: Sure. I think that's a good idea.

config/overlays/e2e-release-0.4/.gitignore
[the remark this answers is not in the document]
mquhuy: This is just so that `git status` does not show any changes after the test. Is that not an issue?

config/overlays/e2e/.gitignore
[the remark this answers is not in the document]
mquhuy: Please see my other comment

config/overlays/e2e-release-0.4/.gitignore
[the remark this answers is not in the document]
tuminoid: Test modifying committed file is the real issue.

hack/e2e/bmo-e2e-net.xml
mquhuy: Thanks. Since I didn't see any usage of this name, I was thinking naming it like this to make it less easy to mess up the ironic.env file. I'm not too keen on it though.

hack/e2e/ironic-e2e.env
mquhuy: It's so that we use the same env file name for BMO and ironic, so it's easier to have the env file copied in `installIronicBMO()`. Not sure the original reasons why they are different, but I suppose we can do it in e2e?

test/e2e/common.go
mquhuy: It's related to the ip address needed by the cluster. Please see my more detailed explanation in another reply.

test/e2e/common.go
mquhuy: My bad. The design and naming of this function is affected by the one in CAPM3 e2e, so I had an impression that it's a common design choice and needn't to be explained further. I'll add some docs.

Other than that reason, I thought of making more generic function, but as we've already a function to install cert manager, I think having one for BMO and Ironic is not so bad a design? This allows us to add some specific logic only needed for our usecase, and after all, this is meant to serve our e2e tests, so it doesn't need to be so generic it can be reused elsewhere.

config/overlays/e2e-release-0.4/.gitignore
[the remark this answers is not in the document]
mquhuy: This `ironic.env` file is actually not committed. It was committed in another kustomize, but imo can be safely removed.

mquhuy:
> Alright! Now the logs look great and the test is doing what I was expecting/hoping! Great job! I still have a lot of comments though 😅
> 
> In general, please try to keep it simple! Refactoring here and there makes it harder to review and understand. Many small PRs are easier and faster to get approved. From the increased complexity (and resource requirements!) I'm now honestly wondering if we should split out the upgrade into its own `ci-e2e-upgrade.sh`. 🤔 With the structure in this PR it will be very hard and complex to do #1443 and we will also get a big merge conflict with #1483 What do you think? Would it make sense to have a separate file and separate job for upgrades?

Thank you. I don't have any preferences. Separating the setup is surely the less complex choice and can make stuff faster. Anw, by "separate job", did you mean jenkins job? If we can make sure we never run upgrade test together with the other test, then I'm okay with it. I reckon that would make `ci-e2e-upgrade.sh` not necessary as well, as we can use the same `ci-e2e.sh` script to setup the upgrade test, as long as we don't use that cluster for anything else.

hack/e2e/bmo-e2e-net.xml
lentzi90: Please change it back. This represents the hostname of whatever is using this IP (minikube). Ironic just happens to run in minikube also.

test/e2e/e2e_suite_test.go
lentzi90: IMO different kustomizations is much, much better. By reusing this one file you save a few lines, but at the same time you increase the complexity and basically require developers to read the code *instead* of the configuration.

- How do you know what is used before upgrade?
  - You find the kustomization that is used. It does not have any env file...
  - You check the code for how it is installed.
  - You find that it copies the file from some other place.
  - You find the actual configuration used.
- How do you know what is used after upgrade? Repeat the steps.
- How do you know what is used in non-upgrade tests? Repeat the steps.

See how much effort goes into finding out these very simple things?

Compare now with "wasting" maybe 50 lines by duplicating the file:

- How do you know what is used before upgrade?
  - You find the kustomization that is used. There is your answer.
- Same process for after upgrade and non-upgrade tests.

At the same time, this allows developers and users to use it in a flexible way. Maybe we want to try using a completely different config before and after upgrade (different IP)? Maybe we want the exact same config (basically no-op "upgrade")? We cannot know all the different ways this will be used which is why flexibility is so important.

To summarize: I find all this complex behavior and code "too much for just simple changes" :wink: 
Static config is not dangerous. DRY is not always the answer. I would say KISS is much more important.

lentzi90:
> Separating the setup is surely the less complex choice and can make stuff faster. Anw, by "separate job", did you mean jenkins job? If we can make sure we never run upgrade test together with the other test, then I'm okay with it. I reckon that would make ci-e2e-upgrade.sh not necessary as well, as we can use the same ci-e2e.sh script to setup the upgrade test, as long as we don't use that cluster for anything else.

Yes, separate jenkins job! We can definitely use the same script. Then all we need is to pass in `GINKGO_FOCUS` or `GINKGO_SKIP` I guess? From jenkins we would just run the same script with different tests selected based on these variables.

test/e2e/common.go
lentzi90: Alright, I get it. Unfortunately CAPM3 is not a good "role model", so please don't take too much inspiration from it :sweat_smile: Especially everything related to how we deploy Ironic/BMO in there is just bad because it is heavily integrated and influenced by metal3-dev-env and the `deploy.sh` script. We should not reimplement that but rather try to learn from those mistakes and make something better here.

With that out of way, I think the only issue here is really the handling of that env file. It is "premature optimization" and not needed. If you just get rid of that, this function can be completely generic, useful and nice! :slightly_smiling_face:

test/e2e/e2e_suite_test.go
mquhuy: Just want to make a comment here that while I agree with everything you mentioned, it's worth to note that if we decide to duplicate the kustomization, we don't simply do it once. In this case, we would need two new kustomizations for BMO in the upgrade cluster, and one for ironic (+one more when we have ironic upgrade test). Imo that's not simple enough that we can say KISS was kept. 
Anyway, fortunately since we're going with separating the job to two, this dilemma does not need to happen anymore :smile:

lentzi90: It is literally just a config file with <10 lines. That is KISS if anything. If the config is the same before and after upgrade, then you can even make a common overlay/component that adds the config and you would get just a kustomization.yaml that "assembles" the config with the rest of the manifests.

It is not an issue any more for CI but I'm trying to explain that this is exactly how I would do it in a real situation. Let's say I want to run this in a homelab. I definitely do not add code for copying and changing files back and forth to get the minimum possible line count. What I would do is to create a separate config for each environment. Nice and clean.

test/e2e/config/ironic-upgrade.yaml
lentzi90, line 1: Not sure if it is worth it to have a separate file for this. Is it more than just the `*DEPLOY_*` variables that differ?
If it is just them I think we can just override them in `ci-e2e.sh` (e.g. `export DEPLOY_BMO=false`).

lentzi90 (commented):
This looks pretty good now!
Could you set up a pipeline for this upgrade job also? Then we could test it from this PR before we merge :slightly_smiling_face:

test/e2e/config/ironic-upgrade.yaml
mquhuy: Yes, then in that case do you think we should keep the `BMO_E2E_TEST_TYPE`?

mquhuy:
> This looks pretty good now! Could you set up a pipeline for this upgrade job also? Then we could test it from this PR before we merge 🙂

Sure. I added the pipeline in these commits:
https://github.com/metal3-io/project-infra/pull/617
https://gerrit.nordix.org/c/infra/cicd/+/20387/1/jjb/metal3/baremetal-operator/jjb-bmo-e2e.yaml

And I updated this PR with your suggested changes: the two jobs are now differentiated by `GINKGO_FOCUS`. If "upgrade" is in the focus list, it will be the only test, otherwise it will not be tested.

hack/ci-e2e.sh
```diff
@@ -17,6 +17,21 @@ cd "${REPO_ROOT}" || exit 1
 # BMO_E2E_EMULATOR can be set to either "vbmc" or "sushy-tools"
 BMO_E2E_EMULATOR=${BMO_E2E_EMULATOR:-"sushy-tools"}
 
+case "${GINKGO_FOCUS:-}" in
+  *upgrade*)
+    echo "'upgrade' found in env var GINKGO_FOCUS. Will ONLY run upgrade tests."
+    export GINKGO_FOCUS="upgrade"
```
lentzi90, line 23: We should not change the variable. If someone has a reason to set this to something more than just upgrade, we just accept it.
```suggestion
    echo "'upgrade' found in env var GINKGO_FOCUS. Will ONLY deploy 'upgrade from' versions of BMO and Ironic."
```

lentzi90 (commented):
Awesome job @mquhuy !
Thanks for the patience through all those comments :sweat_smile: 
/lgtm

config/overlays/e2e-release-0.4/ironic.env
```diff
@@ -0,0 +1,10 @@
+HTTP_PORT=6180
```
kashifest, line 1: The name of the directory could be ironic-release-0.4 or e2e-ironic-release-0.4, otherwise having e2e- in one and fixture- in another seems a bit out of convention.

config/overlays/fixture-release-0.4/kustomization.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
kashifest, line 1: Same comment, the name of the directory could be e2e-fixture-release-0.4 or we can keep it as it is and then add ironic- in the other directory name

hack/ci-e2e.sh
```diff
@@ -17,6 +17,18 @@ cd "${REPO_ROOT}" || exit 1
 # BMO_E2E_EMULATOR can be set to either "vbmc" or "sushy-tools"
 BMO_E2E_EMULATOR=${BMO_E2E_EMULATOR:-"sushy-tools"}
 
+case "${GINKGO_FOCUS:-}" in
+  *upgrade*)
+    export DEPLOY_IRONIC="false"
+    export DEPLOY_BMO="false"
```
kashifest, line 23: I dont understand this, by default we dont deploy BMO even in upgrade test?

kashifest (commented):
Few questions/comments in line

config/overlays/e2e-release-0.4/ironic.env
mquhuy: The name of these directories are to aligned with the equivalent from `main`: `fixtures` and `e2e`. I think your naming convention makes senses, but do you think we should change these original names (`fixtures` and `e2e`) as well? What about you @lentzi90 ?

hack/ci-e2e.sh
mquhuy: The upgrade test has its own logic to install all these components, and they follow the `UPGRADE_DEPLOY_*` configs instead. The reason for that is because we use a different BMO version for upgrade test than the rest of the suite.
In the `fixtures` setup, we use two separate clusters for this purpose, so we could run all tests in the same run, but in `ci-e2e.sh`, we have only 1 cluster, so if we run upgrade test, we don't install anything in the BeforeSuite and leave these installations to the test setup.

config/overlays/fixture-release-0.4/kustomization.yaml
mquhuy: My answer to this one would be the same as in the other kustomization, so let's continue the discussion there.

config/overlays/e2e-release-0.4/ironic.env
lentzi90: I don't have a strong preference. It looks alright like this to me.
My reasoning when I named the folders `e2e` and `fixture` is that they are used for the e2e and fixture tests respectively. So I think it also makes sense then with e2e-release-04 and fixture-release-0.4

kashifest: Ok. I dont have a strong opinion either. This is certainly not a blocker for the tests to go in.

/approve

PR #1453, opened 2023-11-20
🌱E2e live iso ssh implementation

### LINKED ISSUE #1429: Test live-ISO i e2e

maxrantil:
Fixes #1429

### DIFF
--- a/test/e2e/common.go
+++ b/test/e2e/common.go
@@ -8,6 +8,8 @@ import (
 	"strings"
 	"time"
 
+	"golang.org/x/crypto/ssh"
+
 	"github.com/pkg/errors"
 	"gopkg.in/yaml.v2"
 
@@ -18,6 +20,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	metal3api "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
+
+	capm3_e2e "github.com/metal3-io/cluster-api-provider-metal3/test/e2e"
 
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/cluster-api/test/framework"
@@ -295,4 +299,63 @@ func CreateBMHCredentialsSecret(ctx context.Context, client client.Client, secre
 	}
 
 	return nil
+}
+
+func executeSSHCommand(client *ssh.Client, command string) (string, error) {
+	session, err := client.NewSession()
+	if err != nil {
+		return "", fmt.Errorf("failed to create SSH session: %v", err)
+	}
+	defer session.Close()
+
+	output, err := session.CombinedOutput(command)
+	if err != nil {
+		return "", fmt.Errorf("failed to execute command '%s': %v", command, err)
+	}
+
+	return string(output), nil
+}
+
+// HasRootOnDisk parses the output from 'df -h' and checks if the root filesystem is on a disk (as opposed to tmpfs).
+func HasRootOnDisk(output string) bool {
+	lines := strings.Split(output, "\n")
+
+	for _, line := range lines[1:] { // Skip header line
+		if line == "" {
+			continue
+		}
+
+		fields := strings.Fields(line)
+		if len(fields) < 6 {
+			continue // Skip malformed lines
+		}
+
+		if fields[5] == "/" && !strings.Contains(fields[0], "tmpfs") {
+			return true // Found a non-tmpfs root filesystem
+		}
+	}
+
+	return false
+}
+
+// IsBootedFromDisk checks if the system, accessed via the provided ssh.Client, is booted
+// from a disk. It executes the 'df -h' command on the remote system to analyze the filesystem
+// layout. In the case of a disk boot, the output includes a disk-based root filesystem
+// (e.g., '/dev/vda1'). Conversely, in the case of a Live-ISO boot, the primary filesystems
+// are memory-based (tmpfs).
+func IsBootedFromDisk(client *ssh.Client) (bool, error) {
+	cmd := "df -h"
+	output, err := executeSSHCommand(client, cmd)
+	if err != nil {
+		return false, fmt.Errorf("error executing 'df -h': %w", err)
+	}
+
+	isDisk := HasRootOnDisk(output)
+	if isDisk {
+		capm3_e2e.Logf("System is booted from a disk.")
+	} else {
+		capm3_e2e.Logf("System is booted from a live ISO.")
+	}
+
+	return isDisk, nil
 }
--- a/test/e2e/live_iso_test.go
+++ b/test/e2e/live_iso_test.go
@@ -1,0 +1,161 @@
+package e2e
+
+import (
+	"context"
+	"fmt"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"golang.org/x/crypto/ssh"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/pointer"
+	"sigs.k8s.io/cluster-api/test/framework"
+	"sigs.k8s.io/cluster-api/util"
+	"sigs.k8s.io/cluster-api/util/patch"
+
+	metal3api "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
+
+	capm3_e2e "github.com/metal3-io/cluster-api-provider-metal3/test/e2e"
+)
+
+var _ = Describe("Live-ISO", func() {
+	var (
+		specName       = "live-iso-ops"
+		secretName     = "bmc-credentials"
+		namespace      *corev1.Namespace
+		cancelWatches  context.CancelFunc
+		bmcUser        string
+		bmcPassword    string
+		bmcAddress     string
+		bootMacAddress string
+		imageURL       string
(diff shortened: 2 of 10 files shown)


maxrantil:
How I check if the node it booting from `Live-ISO` or not it with the help of the `df -h` cmd.
The two different outputs is:

When booting from a Live-ISO (loaded into memory), the output is as follows:
```
Filesystem                Size      Used Available Use% Mounted on
/dev                      1.9G         0      1.9G   0% /dev
tmpfs                     1.9G         0      1.9G   0% /dev/shm
tmpfs                     1.9G     88.0K      1.9G   0% /run
```
In this case, you will not see a root filesystem (/) associated with a disk like /dev/vda1. Instead, the primary filesystems are memory-based (tmpfs).


When booting from a normal ISO (booting from disk), the output includes a disk-based root filesystem:
```
Filesystem                Size      Used Available Use% Mounted on
/dev                      1.9G         0      1.9G   0% /dev
/dev/vda1                18.9G     43.2M     18.3G   0% /
tmpfs                     1.9G         0      1.9G   0% /dev/shm
tmpfs                     1.9G     88.0K      1.9G   0% /run
```
Here, `/dev/vda1` represents the disk-based root filesystem, indicating a normal ISO boot.

By examining the `df -h` output, specifically looking for the presence or absence of a disk-based root filesystem, you can confirm the boot method of the node.

test/e2e/live_iso_test.go
```diff
@@ -0,0 +1,163 @@
...
+		}, e2eConfig.GetIntervals(specName, "wait-provisioning")...)
+
+		By("Waiting for the BMH to become provisioned")
+		WaitForBmhInProvisioningState(ctx, WaitForBmhInProvisioningStateInput{
+			Client: clusterProxy.GetClient(),
+			Bmh:    bmh,
+			State:  metal3api.StateProvisioned,
+		}, e2eConfig.GetIntervals(specName, "wait-provisioned")...)
+
+		By("Verifying the node booted from live ISO image")
```
lentzi90, line 106: This should be inside the if statement.

test/e2e/live_iso_test.go
```diff
@@ -0,0 +1,163 @@
...
+					ssh.Password("gocubsgo"),
+				},
+				HostKeyCallback: ssh.InsecureIgnoreHostKey(), // #nosec G106
+			}
+
+			// Establish an SSH connection
+			var client *ssh.Client
+
+			Eventually(func() error {
+				client, err = ssh.Dial("tcp", "192.168.222.122:22", config)
```
lentzi90, line 123: The IP needs to be configurable or auto-detected.

test/e2e/live_iso_test.go
```diff
@@ -0,0 +1,163 @@
...
+
+		By("Waiting for the BMH to become provisioned")
+		WaitForBmhInProvisioningState(ctx, WaitForBmhInProvisioningStateInput{
+			Client: clusterProxy.GetClient(),
+			Bmh:    bmh,
+			State:  metal3api.StateProvisioned,
+		}, e2eConfig.GetIntervals(specName, "wait-provisioned")...)
+
+		By("Verifying the node booted from live ISO image")
+		// This is to bypass the fixture test only
```
lentzi90, line 107:
```suggestion
		// The ssh check is not possible in all situations (e.g. fixture) so it can be skipped
```

test/e2e/common.go
```diff
@@ -221,3 +225,61 @@ func buildKustomizeManifest(source string) ([]byte, error) {
...
+	} else {
+		capm3_e2e.Logf("System is booted from a live ISO.")
+	}
+
+	return isDisk, nil
+}
+
+func StringPtr(s string) *string {
+	return &s
+}
```
lentzi90, line 285: Let's not reinvent the wheel. We can use https://pkg.go.dev/k8s.io/utils/pointer

```go
stringPointer := pointer.String("my-string-pointer")
```

test/e2e/common.go
```diff
@@ -221,3 +225,61 @@ func buildKustomizeManifest(source string) ([]byte, error) {
...
+
+		if fields[5] == "/" && !strings.Contains(fields[0], "tmpfs") {
+			return true // Found a non-tmpfs root filesystem
+		}
+	}
+
+	return false
+}
+
+func IsBootedFromDisk(client *ssh.Client) (bool, error) {
```
lentzi90, line 266: Please add a doc string

test/e2e/common.go
```diff
@@ -221,3 +225,61 @@ func buildKustomizeManifest(source string) ([]byte, error) {
...
+	output, err := session.CombinedOutput(command)
+	if err != nil {
+		return "", fmt.Errorf("failed to execute command '%s': %v", command, err)
+	}
+
+	return string(output), nil
+}
+
+// ParseFileSystemOutput parses the given filesystem output string and checks if booted from disk
+func ParseFileSystemOutput(output string) bool {
```
lentzi90, line 245: I would suggest renaming to clarify the output
```suggestion
// HasRootOnDisk parses the output from df -h and checks if the root filesystem is on a disk (as opposed to tmpfs).
func HasRootOnDisk(dfOutput string) bool {
```

test/e2e/live_iso_test.go
```diff
@@ -0,0 +1,163 @@
...
+			defer func() {
+				if client != nil {
+					client.Close()
+				}
+			}()
+			isDisk, err := IsBootedFromDisk(client)
+			Expect(err).NotTo(HaveOccurred(), "Error in verifying boot mode")
+			Expect(isDisk).To(Equal(false), "Error booting from disk when live ISO is expected")
+		} else {
+			capm3_e2e.Logf("Fixture system is booted from a live ISO.")
```
lentzi90, line 136: Let's not make false claims (if we run fixture, nothing at all is booted). It is better to be explicit here and warn that the check was skipped
```suggestion
			capm3_e2e.Logf("WARNING: Skipping ssh check since SSH_CHECK_PROVISIONED != true")
```

lentzi90 (commented):
Great progress!
The ssh check will be useful also in other tests!

test/e2e/live_iso_test.go
maxrantil: Very good point

test/e2e/common.go
maxrantil: Is what I now implemented similar to what you had in mind?

test/e2e/live_iso_test.go
maxrantil: Added variables in ironic config for it.

test/e2e/live_iso_test.go
```diff
@@ -0,0 +1,159 @@
...
+				Annotations: map[string]string{
+					metal3api.InspectAnnotationPrefix:   "disabled",
+					metal3api.HardwareDetailsAnnotation: hardwareDetails,
+				},
+			},
+			Spec: metal3api.BareMetalHostSpec{
+				Online: true,
+				BMC: metal3api.BMCDetails{
+					Address:         bmcAddress,
+					CredentialsName: "bmc-credentials",
```
adilGhaffarDev, line 66:
```suggestion
					CredentialsName: secretName,
```

test/e2e/live_iso_test.go
```diff
@@ -0,0 +1,159 @@
...
+				},
+				Image: &metal3api.Image{
+					URL:        e2eConfig.GetVariable("IMAGE_URL"),
```
adilGhaffarDev, line 69:
```suggestion
					URL:        imageURL,
```
Maybe move `e2eConfig.GetVariable("IMAGE_URL")` to `BeforeEach`.

test/e2e/live_iso_test.go
```diff
@@ -0,0 +1,161 @@
...
+
+		// The ssh check is not possible in all situations (e.g. fixture) so it can be skipped
+		if e2eConfig.GetVariable("SSH_CHECK_PROVISIONED") == "true" {
+			By("Verifying the node booted from live ISO image")
+
+			// Set up SSH client configuration
+			config := &ssh.ClientConfig{
+				User: "cirros",
+				Auth: []ssh.AuthMethod{
+					ssh.Password("gocubsgo"),
```
lentzi90, line 107: Note for the future: We should make this configurable and also support ssh keys. However, I think it is good enough for now since the ssh check can be disabled if needed.
I think we should work on this in a separate PR where we also add the ssh check to the other provisioning test.

lentzi90 (commented):
/lgtm

test/e2e/live_iso_test.go
maxrantil: Sounds good, I can do an issue on it.

kashifest (commented):
/approve

test/e2e/live_iso_test.go
maxrantil: Created [here](https://github.com/metal3-io/baremetal-operator/issues/1473)

PR #1456, opened 2023-11-22
🌱 Add github action to trigger container images build

mquhuy:
This PR adds an action to trigger the bmo and keepalived container builds whenever a PR is merged in this repo.

### DIFF
--- a/.github/workflows/build-images-action.yml
+++ b/.github/workflows/build-images-action.yml
@@ -1,0 +1,41 @@
+name: build-images-action
+on:
+  push:
+    branches:
+      - 'main'
+      - 'release-*'
+    tags:
+      - 'v*'
+permissions: {}
+jobs:
+  build:
+    name: Build container images
+    runs-on: ubuntu-latest
+    if: github.repository == 'metal3-io/baremetal-operator'
+    permissions:
+      contents: read
+    steps:
+      - name: build bmo image
+        uses: toptal/jenkins-job-trigger-action@137fff703dd260b52b53d3ba1960396415abc568 # 1.0.2
+        with:
+          jenkins_url: "https://jenkins.nordix.org/"
+          jenkins_user: "metal3.bot@gmail.com"
+          jenkins_token: ${{ secrets.JENKINS_TOKEN }}
+          job_name: "metal3_baremetal-operator_container_image_building"
+          job_params: |
+            {
+              "BUILD_CONTAINER_IMAGE_GIT_REFERENCE": "${{ github.ref }}"
+            }
+          job_timeout: "1000"
+      - name: build keepalived image
+        uses: toptal/jenkins-job-trigger-action@137fff703dd260b52b53d3ba1960396415abc568 # 1.0.2
+        with:
+          jenkins_url: "https://jenkins.nordix.org/"
+          jenkins_user: "metal3.bot@gmail.com"
+          jenkins_token: ${{ secrets.JENKINS_TOKEN }}
+          job_name: "metal3_keepalived_container_image_building"
+          job_params: |
+            {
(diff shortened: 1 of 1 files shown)


.github/workflows/build-images-action.yml
```diff
@@ -0,0 +1,43 @@
...
+      - name: build keepalived image
+        uses: toptal/jenkins-job-trigger-action@137fff703dd260b52b53d3ba1960396415abc568 # 1.0.2
+        with:
+          jenkins_url: "https://jenkins.nordix.org/"
+          jenkins_user: "metal3.bot@gmail.com"
+          jenkins_token: ${{ secrets.JENKINS_TOKEN }}
+          job_name: "metal3_keepalived_container_image_building"
+          job_params: |
+            {
+              "BUILD_CONTAINER_IMAGE_NAME": "keepalived",
```
tuminoid, line 40: How does this job know which Dockerfile to build for keepalived?

mquhuy: The data is stored in this file: https://github.com/Nordix/metal3-dev-tools/blob/main/scripts/files/container_image_names.json

And logic is handled here: https://github.com/Nordix/metal3-dev-tools/blob/main/scripts/build-container-image.sh#L105

mquhuy: I think you have a good argument, but there're a couple of reasons that I still prefer a centralized database of repose:
1. I don't like the idea of putting repo address as parameter, which is error-prone imo. Any small typo could fail the job, while we actually still have cases when we need to run the builds manually.
2. Having no centralized file means we will allow build and push of any repos, not just the ones we care about. Anyone on Nordix jenkins can trigger it, it can build anything, and all will be pushed to our repository in quay.

tuminoid: 1. Typo anywhere would make it fail. You can probably get it via GH reference, same as branch anyways.
2. I don't think they can run builds. You need group membership to launch them. As anon, I cannot access any build option. If it was possible, our current jobs could be equally abused as they're all parametrized.

mquhuy: 1. Yes, but typo in url is generally harder to check. And I think the bigger problem than failed jobs is jobs that do the wrong things, and generally more params = more errors. That's of course not a problem with github triggered runs, but we still have manual builds. It can happen that we paste one url while were meant to use another: say we want to build bmo but mistakenly use url from capm3 --> BMO image will be pushed with capm3 content, quite difficult to check.
2. I agree, I'm not sure regarding of jenkins permissions, it's likely that only us can trigger those jobs. Anyway, my points regarding the "any image" still checks.

tuminoid: 1. We won't allow pushing images from manual builds, so that is non issue.

tuminoid: This was cleared off-line. 

We do not do manual or custom builds that are built to Quay. We define the tag patterns in the workflow file, and the typical list is:
- main/latest for main branch updates
- release-* tags for release branch updates
- v* tags or capm3-v* tags for releases

We do keep X number of builds that are timestamped + sha encoded for the branch builds.

We do not want to build lightweight tags like:
- apis/v*

And we definitely do not support building and pushing images based on user input that might mess up with the above mentioned "official" images.

tuminoid (commented):
/lgtm

tuminoid:
/override test-centos-e2e-integration-main test-ubuntu-integration-main 
Not relevant.

kashifest (commented):
/approve

tuminoid:
Disabled build triggers in Quay for BMO and keepalived. It should now all flow thru this workflow + Jenkins.

PR #1459, opened 2023-11-27
🌱 improve hack/verify-release.sh draft release note handling

tuminoid:
Github API does not allow fetching draft release notes by tag, only by release id. Add the required code to fetch that, and also do not bail out if it fails, just skip the tests that require the data.

Also fix osv-scanner exiting the script before printing the vulnerabilities. It returns failure when vulns are found.

Same has been implemented in CAPM3 and IPAM. This needs to be cherry-picked to release-0.4 and release-0.3 branches.

### DIFF
--- a/hack/verify-release.sh
+++ b/hack/verify-release.sh
@@ -59,6 +59,8 @@ REGISTRY="quay.io"
 
 # if the given tag doesn't exist, we run only pre-tag checks
 TAG_EXISTS=""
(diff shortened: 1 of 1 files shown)


tuminoid:
/override test-centos-e2e-integration-main 
/override test-ubuntu-integration-main 
Save the trees.

lentzi90 (commented):
/lgtm

PR #1462, opened 2023-11-27
🌱 Delete "ok-to-test" label from dependabot PRs

peppi-lotta:
Dependabot recently is creating quite a few unnecessary PRs which need to be closed manually, having ok-to-test creates a lot of unneeded tests. Removing the label would allow adding labels selectively by reviewers and save some test bandwidth

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -22,8 +22,6 @@ updates:
       - dependency-name: "k8s.io/*"
     commit-message:
       prefix: ":seedling:"
-    labels:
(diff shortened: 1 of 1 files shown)


kashifest:
Here the title would need a proper icon as we talked offline, you can find an appropriate icon from here https://github.com/metal3-io/baremetal-operator/blob/main/CONTRIBUTING.md#contributing-a-patch for this patch. Once it is there the PR verifier should pass

kashifest:
/hold
Until comments are addressed.

PR #1463, opened 2023-11-28
🌱 Delete "ok-to-test" label from dependabot PRs

peppi-lotta:
Dependabot recently is creating quite a few unnecessary PRs which need to be closed manually, having ok-to-test creates a lot of unneeded tests. Removing the label would allow adding labels selectively by reviewers and save some test bandwidth

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,8 +9,6 @@ updates:
       interval: "monthly"
     commit-message:
       prefix: ":seedling:"
-    labels:
-      - "ok-to-test"
     # Go
   - package-ecosystem: "gomod"
     directory: "/"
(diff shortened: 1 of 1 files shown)


tuminoid:
> /test-centos-e2e-integration-main /test-ubuntu-integration-main

GH workflow files are not tested by integration tests, btw, don't bother triggering them. We will override them after approval.

tuminoid (commented):
/lgtm

This is what we agreed with @kashifest. Automatic `ok-to-test` creates massive spikes in CI during dependabot PR creation spree, and it handles the grouping of `k8s.io/*` deps so badly that it usually ends up creating, and rebasing, a dozen of them, while we eventually merge one or two of them. We will hence disable the automatic ok-to-test, and can revisit it after the dependabot team gets the package grouping config properly implemented.

PR #1469, opened 2023-11-30
🌱 Fix ensure-minikube script to allow minikube versions greater than MINIMUM_MINIKUBE_VERSION

maxrantil:
Right now ensure_minikube.sh gives an error while e2e tests locally even when the minikube version is greater than MINIMUM_MINIKUBE_VERSION which is wrong. This PR is fixing that.


kashifest (commented):
/lgtm

PR #1479, opened 2023-12-05
🌱 sync cert-manager version with preloaded images in e2e

tuminoid:
This allows e2e test to use preloaded images and not waste time to do double download.


lentzi90 (commented):
/lgtm

PR #1480, opened 2023-12-05
✨ Add an explicit way to auto-detect ChecksumType

dtantsur:
In preparation for the eventual removal of MD5 checksums, add a new
ChecksumType "auto" that instructs IPA to detect the type

Requires IPA from the 2023.2 release series. Users with older IPA
should provide the algorithm explicitly.

### DIFF
--- a/apis/metal3.io/v1alpha1/baremetalhost_types.go
+++ b/apis/metal3.io/v1alpha1/baremetalhost_types.go
@@ -450,7 +450,7 @@ const (
 )
 
 // ChecksumType holds the algorithm name for the checksum
-// +kubebuilder:validation:Enum=md5;sha256;sha512
+// +kubebuilder:validation:Enum=md5;sha256;sha512;auto
 type ChecksumType string
 
 const (
@@ -462,6 +462,9 @@ const (
 
 	// SHA512 checksum type
 	SHA512 ChecksumType = "sha512"
+
+	// Automatically detect
+	AutoChecksum ChecksumType = "auto"
 )
 
 // Image holds the details of an image either to provisioned or that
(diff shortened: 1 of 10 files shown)


Rozzii (approved):
/lgtm

lentzi90 (commented):
/lgtm

zaneb:
I think I disagree with this change.

This is not about removing MD5, since it's still possible to specify MD5 explicitly.

The is about changing the type field from required to optional for SHA sums. That means that the same v1alpha1 CR may fail to work on older versions of Metal³ - or, as you acknowledge in the description, on the _same_ version of Metal³ with different versions of IPA. Essentially this is starting to leak Ironic API changes through into the BMH.

If we want to remove the type field in v1beta1 (which I'm fine with) we can still translate back to v1alpha1 when required in the webhook by using the same logic that Ironic does (checking the length of the sum) to determine the algorithm. As far as I can see, this change is not required to do that.

dtantsur:
> we can still translate back to v1alpha1 when required in the webhook by using the same logic that Ironic does (checking the length of the sum)

We cannot, this logic is not trivial, and I don't want to replicate it there (it's not just length, there are many formats of downloadable checksums).

@zaneb would you be more okay with an explicit `auto` type as I initially planned? Otherwise, we'll have nothing to translate v1beta1 into.

dtantsur:
Okay, this should no longer break the backward compatibility with older IPA.

zaneb (commented):
Happy with this :+1: 
/approve

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -473,8 +476,9 @@ type Image struct {
 	// Checksum is the checksum for the image.
 	Checksum string `json:"checksum,omitempty"`
 
-	// ChecksumType is the checksum algorithm for the image.
-	// e.g md5, sha256, sha512
+	// ChecksumType is the checksum algorithm for the image, e.g md5, sha256 or sha512.
+	// The special value "auto" can be used to detech the algorithm from the checksum.
```
lentzi90, line 480: nit
```suggestion
	// The special value "auto" can be used to detect the algorithm from the checksum.
```

hardys:
> /cc @hardys
> in case you're affected

Thanks for the heads-up 👍 

/lgtm

lentzi90 (commented):
/lgtm

PR #1494, opened 2023-12-18
🌱 E2E: Have a common function to install BMO and Ironic

mquhuy:
This PR refactors and uses a common function to install ironic and bmo. This will be useful later should we have to install other stuff with kustomize.

### DIFF
--- a/test/e2e/common.go
+++ b/test/e2e/common.go
@@ -15,6 +15,7 @@ import (
 
 	. "github.com/onsi/gomega"
 
+	v1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


test/e2e/common.go
```diff
@@ -359,3 +360,56 @@ func IsBootedFromDisk(client *ssh.Client) (bool, error) {
...
+	ClusterProxy        framework.ClusterProxy
+	WaitForDeployment   bool
+	WatchDeploymentLogs bool
+	DeploymentName      string
+	DeployNS            string
+	LogPath             string
+	WaitIntervals       []interface{}
+}
+
+func BuildAndInstallDeployment(ctx context.Context, input *BuildAndInstallDeploymentInput) error {
```
lentzi90, line 375: Could you please add a doc string?

I think we can make it slightly more generic still, which should make it useful for even more things. We are really building and applying a kustomization, so I think changing the name would make sense.
```suggestion
func BuildAndApplyKustomization(ctx context.Context, input *BuildAndInstallDeploymentInput) error {
```

test/e2e/common.go
```diff
@@ -359,3 +360,56 @@ func IsBootedFromDisk(client *ssh.Client) (bool, error) {
...
+	manifest, err := buildKustomizeManifest(kustomization)
+	if err != nil {
+		return err
+	}
+	err = clusterProxy.Apply(ctx, manifest)
+	if err != nil {
+		return err
+	}
+
+	ironicBMODeployment := &v1.Deployment{
```
lentzi90, line 388:
```suggestion
	deployment := &v1.Deployment{
```

Does this work if no name or namespace is specified (if we do not ask for wait and watch)? If not, I would like to put it inside an if statement so it is only executed when really needed.

test/e2e/common.go
```diff
@@ -359,3 +360,56 @@ func IsBootedFromDisk(client *ssh.Client) (bool, error) {
...
 	return isDisk, nil
 }
+
+type BuildAndInstallDeploymentInput struct {
+	Kustomization       string
+	ClusterProxy        framework.ClusterProxy
+	WaitForDeployment   bool
+	WatchDeploymentLogs bool
+	DeploymentName      string
+	DeployNS            string
```
lentzi90, line 370:
```suggestion
	DeploymentNamespace            string
```

test/e2e/e2e_suite_test.go
```diff
@@ -116,64 +114,41 @@ var _ = SynchronizedBeforeSuite(func() []byte {
...
+
 	if e2eConfig.GetVariable("DEPLOY_IRONIC") != "false" {
 		// Install Ironic
 		By("Installing Ironic")
 		kustomization := e2eConfig.GetVariable("IRONIC_KUSTOMIZATION")
-		manifest, err := buildKustomizeManifest(kustomization)
-		Expect(err).NotTo(HaveOccurred())
-		err = clusterProxy.Apply(ctx, manifest)
+		err := BuildAndInstallDeployment(ctx, &BuildAndInstallDeploymentInput{
+			Kustomization:       kustomization,
```
lentzi90, line 124: nit: we can skip the extra variable here I think.
```suggestion
			Kustomization:       e2eConfig.GetVariable("IRONIC_KUSTOMIZATION"),
```

lentzi90 (commented):
Nice!

test/e2e/common.go
mquhuy: Thanks. I thought of something similar in the first place, but the "watch" stuff would look very weird if what we install doesn't have any deployments :sweat_smile: 
Maybe we can make it work with some explanation in the docs? I agree that later on we could use this for other things as well.

lentzi90: Yeah but we can actually use it for stuff without deployment already as long as we do not enable wait/watch :slightly_smiling_face: 
Explanations in the doc string (for either the function or the struct) is always good!

lentzi90 (commented):
Nice!
/lgtm

lentzi90 (commented):
/lgtm

maxrantil:
Nice, I like :)

test/e2e/common.go
```diff
@@ -359,3 +360,95 @@ func IsBootedFromDisk(client *ssh.Client) (bool, error) {
...
+
+	// Path to store the deployment logs
+	LogPath string
+
+	// Intervals to use in checking and waiting for the deployment
+	WaitIntervals []interface{}
+}
+
+func (input *BuildAndApplyKustomizeInput) validate() error {
+	if !input.WaitForDeployment && !input.WatchDeploymentLogs {
```
kashifest, line 393: Should it be `&&` or `||`, perhaps I am misunderstanding the use case but I would appreciate a comment here in the code.

kashifest (commented):
Nice Improvement, one question inline. otherwise lgtm

test/e2e/common.go
mquhuy: It should be &&. Only if both of them are false then we don't need to validate the other inputs. I'll add the comment

kashifest:
/approve
/lgtm 
Since it got removed while adding the comment

PR #1511, opened 2024-01-09
📖 fix previous release detection in make release-notes

tuminoid:
We're overlapping our version numbers with really old version numbers, ie. capm3-v0.4.1 vs v0.4.1 when using grep without line-start ^. Fix the issue, so the correct previous tag is used in release automation.


lentzi90 (commented):
Thanks!
/lgtm

smoshiur1237 (approved):
LGTM

PR #1515, opened 2024-01-09
🌱 Bump controller-runtime, k8s/*, CAPM3/test and other related go modules

kashifest:
Bump
- sigs.k8s.io/controller-runtime to v0.16.3
- k8s.io/* to v0.28.5
- github.com/metal3-io/cluster-api-provider-metal3/test to v1.6.0

MetricsBindAddress is deprecated. We now use metricsServer options to accomodate the bindAddress


tuminoid:
Can you mention in the description why the metrics is removed?

kashifest:
> Can you mention in the description why the metrics is removed?

Done

tuminoid (commented):
/lgtm

main.go
```diff
@@ -197,9 +194,15 @@ func main() {
...
+	var watchNamespaces map[string]cache.Config
+	if watchNamespace != "" {
+		watchNamespaces = map[string]cache.Config{
+			watchNamespace: {},
+		}
+	}
+
 	ctrlOpts := ctrl.Options{
-		Scheme:             scheme,
-		MetricsBindAddress: metricsBindAddr,
```
zaneb, line 202: The ability to configure the metrics server hasn't been removed, it has just moved to the [`Metrics`](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.16.3/pkg/manager/manager.go#L218) field: https://github.com/kubernetes-sigs/controller-runtime/blob/v0.16.3/pkg/metrics/server/server.go#L64-L67

kashifest: It has been changed from the flag to a Metrics server options,  we can configure it using the metrics  server option field as you pointed out . My plan is to rather introduce secure serving in a followup as we Did in CAPM3 https://github.com/metal3-io/cluster-api-provider-metal3/commit/f067b5b0c67a8f01dc47a34bed54683d811e68f1. Wdyt?

zaneb: We're not using this flag downstream, but this change will break anybody who is (break as in BMO won't start, not just metrics are unavailable). It seems to me like the follow-up PR would be the best place to discuss how to handle that, rather than breaking it unnecessarily as a side-effect of bumping controller-runtime.

kashifest: ok, fair enough, I have changed the code now to accomodate the same flag with metricsServer options. Anyhow, keep in mind that, it should be breaking change when we introduce secure serving.

zaneb:
Thanks!
/approve

tuminoid (commented):
/lgtm

PR #1520, opened 2024-01-18
🌱 Test virtualmedia booting

### LINKED ISSUE #1512: E2E: Test virtualmedia boot
Test virtualmedia boot in the e2e tests. We have a live-iso test already [here](https://github.com/metal3-io/baremetal-operator/blob/main/test/e2e/live_iso_test.go), but this just sets the disk format. The idea with the live-iso test that we have in CAPM3 is to test virtualmedia booting with a live-iso and that would be good to test here as well.

I think we have two alternatives:

1. Make the live-iso test "special". It would require redfish-virtualmedia or skip the test
2. Make all tests work with virtualmedia. Just like we currently test with IPMI and redfish, this would just be another option. In this case the live-iso test stays like it is now but we may have to adapt other tests to work with virtualmedia (e.g. disk cleaning could be strange).

I would prefer option 2.

lentzi90:

Fixes #1512


tuminoid:
/metal3-bmo-e2e-test-pull  
I wonder if this works as people are trying to use it this way for sure....

tuminoid:
Now this should work?

lentzi90:
Prow should also be able to help with commands...
/test ?

test/e2e/live_iso_test.go
```diff
@@ -30,7 +32,17 @@ var _ = Describe("Live-ISO", Label("required", "live-iso"), func() {
...
 
 	BeforeEach(func() {
-		imageURL = e2eConfig.GetVariable("IMAGE_URL")
+		// Check what kind of BMC we are dealing with
+		// It may be *possible* to boot a live-ISO over (i)PXE, but there are severe limitations.
+		// Therefore we skip the test if it doesn't support ISO preprovisioning images.
+		// See https://docs.openstack.org/ironic/latest/admin/ramdisk-boot.html
+		accessDetails, err := metal3bmc.NewAccessDetails(bmc.Address, false)
+		Expect(err).NotTo(HaveOccurred())
+		if !accessDetails.SupportsISOPreprovisioningImage() {
```
lentzi90, line 41: I could use a second opinion here. Does it make sense to skip the test for ipmi and redfish? Is this the right thing to check?
Based on the docs I linked in the comment above, I think it isn't really worth it to test live-ISO in non-virtualmedia cases and the test currently fails for them. This is why I added the skip.

hack/ci-e2e.sh
```diff
@@ -106,16 +107,42 @@ export IMAGE_URL="http://${IP_ADDRESS}/${IMAGE_FILE}"
...
   -p 80:8080 \
   -v "${IMAGE_DIR}:/usr/share/nginx/html" nginxinc/nginx-unprivileged
 
-# Generate the key pair
+# Generate ssh key pair for verifying provisioned BMHs
 ssh-keygen -t ed25519 -f "${IMAGE_DIR}/ssh_testkey" -q -N ""
 
+# Build an ISO image with baked ssh key
+# See https://www.system-rescue.org/scripts/sysrescue-customize/
+# We use the systemrescue ISO and their script for customizing it.
```
lentzi90, line 124: I'm open to suggestions here. Should we build the ISO in a different way? Use a different image to start from?

lentzi90:
/cc @dtantsur @Rozzii 
I would appreciate your input on this. Especially if you think it makes sense to skip the test for ipmi and redfish as I do now.
My plan for virtualmedia going forward would be that the e2e tests run with redfish and redfish-virtualmedia on all PRs. Periodic tests would also run ipmi with VBMC so we test a different protocol.

test/e2e/live_iso_test.go
dtantsur: Technically, you can boot an ISO via iPXE. We added this to BMO purely accidentally, but when a validation was added that blocked it, some customers complained. Apparently, people use this in production.

hack/ci-e2e.sh
dtantsur: In the Ironic CI, we use http://tinycorelinux.net/15.x/x86/release/Core-current.iso, but I don't know how to embed the SSH key there. If sysrescue does not take ages to build, it could be an option.

test/e2e/live_iso_test.go
```diff
@@ -68,8 +80,9 @@ var _ = Describe("Live-ISO", Label("required", "live-iso"), func() {
 					URL:        imageURL,
 					DiskFormat: pointer.String("live-iso"),
 				},
-				BootMode:       metal3api.Legacy,
-				BootMACAddress: bmc.BootMacAddress,
+				BootMode:              metal3api.Legacy,
```
dtantsur, line 83: Sigh.

lentzi90: One thing at a time. I have added https://github.com/metal3-io/baremetal-operator/issues/1607 for it :wink:

test/e2e/live_iso_test.go
lentzi90: But would you say we need to test it? It seems we would need some special ISO to get something useful with iPXE.

> This feature, when utilized with the ipxe boot_interface, will only allow a kernel and ramdisk to be booted from the supplied ISO file. Any additional contents, such as additional ramdisk contents or installer package files will be unavailable after the boot of the Operating System. Operators wishing to leverage this functionality for actions such as OS installation should explore use of the standard ramdisk deploy_interface along with the instance_info/kernel_append_params setting to pass arbitrary settings such as a mirror URL for the initial ramdisk to load data from. This is a limitation of iPXE and the overall boot process of the operating system where memory allocated by iPXE is released.

tuminoid:
Good work! Not having been involved enough in BMO testing, someone else can flag it. LGTM.

test/e2e/live_iso_test.go
elfosardo: IMHO it's probably not worth testing it at this stage
we could add a TODO for the future as a reminder in case this become more impactful, but I doubt it

elfosardo (commented):
/lgtm

PR #1522, opened 2024-01-18
🌱 Install golangci-lint directly instead of building it using go-modules

kashifest:
as a benefit we get rid of the hashicorp dependencies from `hack/tools/go.mod`.


Makefile
```diff
@@ -331,9 +340,10 @@ go-version: ## Print the go version we use to compile our binaries and images
 ## --------------------------------------
 
 .PHONY: clean
-clean: ## Remove all temporary files and folders
+clean: ## Remove all temporary files,folders and tools
```
tuminoid, line 343:
```suggestion
clean: ## Remove all temporary files, directories and tools
```
Missing space + folder is something on a desktop

tuminoid (commented):
One tiny nit.

maxrantil:
Whaow! That's a lot of cleaning up lines :) LGTM

Makefile
```diff
@@ -120,13 +128,14 @@ test-e2e: $(GINKGO) ## Run the end-to-end tests
 .PHONY: linters
 linters: lint generate-check fmt-check
 
-tools/bin/golangci-lint: hack/tools/go.mod
-	cd hack/tools; go build -o $(abspath $@) github.com/golangci/golangci-lint/cmd/golangci-lint
+$(GOLANGCI_LINT):
```
tuminoid, line 131: Can you add phony target for $(GOLANGCI_LINT_BIN) so it is easier to call as is? Now you need to call `make $(pwd)/hack/tools/bin/golangci-lint` to make it.

tuminoid: I can also add it in follow-up, if not feeling like it.

kashifest: I will add it here in a while

kashifest: Done

lentzi90 (commented):
/lgtm

Makefile
tuminoid: Thanks. 

/lgtm

honza:
This is great.

/approve

PR #1523, opened 2024-01-19
🌱 E2E Upgrade test: Add more BMO versions to upgrade from

mquhuy:
This PR adds more BMO config versions to the E2E upgrade tests, so that we can test from either of `release-0.3`, `release-0.4` and `release-0.5` to `main` (with both `ironic` and `fixture` config).
It also changes the default "from" version from `release-0.4` to `release-0.5`.

e.g. To run upgrade test from `release-0.3` to `main` with `ironic` config:
```bash
export GINKGO_FOCUS="upgrade"
export UPGRADE_BMO_KUSTOMIZATION_FROM=config/overlays/e2e-release-0.3
./hack/ci-e2e.sh
```

Edit: The upgrade-from version can now be seen from the test output:
```
 STEP: Installing BMO from ../../config/overlays/e2e-release-0.5 on the upgrade cluster
```
And from the logs
```
test/e2e/_artifacts/logs/baremetal-operator-system-upgrade/
├── bmo-e2e
│   └── baremetal-operator-controller-manager
│       ├── baremetal-operator-controller-manager-696c96dbdc-ng2qh
│       │   ├── kube-rbac-proxy-log-metadata.json
│       │   ├── kube-rbac-proxy.log
│       │   ├── manager-log-metadata.json
│       │   └── manager.log
│       └── baremetal-operator-controller-manager-7dd8b7c4b4-9mlhg
│           ├── kube-rbac-proxy-log-metadata.json
│           ├── kube-rbac-proxy.log
│           ├── manager-log-metadata.json
│           └── manager.log
├── bmo-e2e-release-0.5
│   └── baremetal-operator-controller-manager
│       ├── baremetal-operator-controller-manager-696c96dbdc-ng2qh
│       │   ├── kube-rbac-proxy-log-metadata.json
│       │   ├── kube-rbac-proxy.log
│       │   ├── manager-log-metadata.json
│       │   └── manager.log
│       └── baremetal-operator-controller-manager-7dd8b7c4b4-9mlhg
│           ├── kube-rbac-proxy-log-metadata.json
│           ├── kube-rbac-proxy.log
│           ├── manager-log-metadata.json
│           └── manager.log
└── ironic
    └── ironic-754996654b-ftkxq
        ├── ironic-httpd-log-metadata.json
        ├── ironic-httpd.log
        ├── ironic-inspector-log-metadata.json
        ├── ironic-inspector.log
        ├── ironic-log-metadata.json
        ├── ironic-log-watch-log-metadata.json
        ├── ironic-log-watch.log
        └── ironic.log
```


lentzi90 (commented):
This looks good already, but I have one request: would it be possible to make it visible in the logs what version we test?
Maybe just printing the exact overlay that is used? We could also print the image tag or something like that.

mquhuy:
> This looks good already, but I have one request: would it be possible to make it visible in the logs what version we test? Maybe just printing the exact overlay that is used? We could also print the image tag or something like that.

Thanks Lennart. Sorry it took a bit long, I almost forgot about it.
I've updated the PR with reference to the version in logs. Details in PR description. Please let me know if something else is still missing xD

/metal3-bmo-e2e-optional-test

config/overlays/e2e-release-0.3/kustomization.yaml
```diff
@@ -0,0 +1,45 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: baremetal-operator-system
+resources:
+  - https://github.com/metal3-io/baremetal-operator/config/tls?ref=release-0.3
```
tuminoid, line 5: Please use consistent k8s formatting, ie. without leading spaces for the lists. You have both now in this file.

test/e2e/upgrade_test.go
```diff
@@ -233,15 +234,19 @@ var _ = Describe("BMO Upgrade", func() {
 		}
 
 		if e2eConfig.GetVariable("UPGRADE_DEPLOY_BMO") != "false" {
-			By("Installing BMO on the upgrade cluster")
+			bmoKustomization := e2eConfig.GetVariable("UPGRADE_BMO_KUSTOMIZATION_FROM")
+			// Get the index of the last "release" word, which typically indicates the start of the BMO version name
+			lastReleaseWordIndex := strings.LastIndex(bmoKustomization, "release")
+			BmoFromVer := bmoKustomization[lastReleaseWordIndex:]
+			By(fmt.Sprintf("Installing BMO with version: %s on the upgrade cluster", BmoFromVer))
```
lentzi90, line 241: This basically adds a requirement on the kustomization folder name. I would rather have something simpler that doesn't care what you name them. Perhaps just print the path?
```suggestion
			bmoKustomization := e2eConfig.GetVariable("UPGRADE_BMO_KUSTOMIZATION_FROM")
			By(fmt.Sprintf("Installing BMO from %s on the upgrade cluster", bmoKustomization))
```

config/overlays/e2e-release-0.3/kustomization.yaml
mquhuy: Thanks. I've fixed it.

test/e2e/upgrade_test.go
mquhuy: Hmm. Okay. How about using the Base path instead? Having a full path in the log is fine, but I also want to have a nice way to notice it in the artifacts.

lentzi90: Oh I didn't notice it was used for the artifacts. Thanks for pointing it out!
I think the base path sounds good for artifacts, in the log I would say the more we include the better :D

mquhuy: Sounds good. I'll update it real quick

mquhuy:
> /lgtm Thanks for this! I think that it would be nice eventually to refactor it so that we can run the upgrade test for all different versions, one after the other. But let's do that in a follow up. The reason I think it would be nice is to reduce the number of jobs so it is easier to keep track of them and also to understand what we test.

Thank you. Yes, I was thinking about that idea the other day. Having a separate job for every upgrade test is too much to handle, but the "problem" with running them in sequence is that to run any upgrade test (including ironic upgrade), the cluster should be cleaned up to its initial state (no ironic nor bmo), so we need to test if cleaning up a cluster is reliable (i.e. if the cluster is really "clean" afterwards). I guess it needs more investigating, and I might start doing it together with the ironic upgrade test.

tuminoid (commented):
Does the inspector config here work with the https://github.com/metal3-io/baremetal-operator/pull/1544 ?

config/overlays/e2e-release-0.3/ironic.env
```diff
@@ -0,0 +1,10 @@
+HTTP_PORT=6180
+PROVISIONING_IP=192.168.222.199
+DEPLOY_KERNEL_URL=http://192.168.222.199:6180/images/ironic-python-agent.kernel
+DEPLOY_RAMDISK_URL=http://192.168.222.199:6180/images/ironic-python-agent.initramfs
+IRONIC_ENDPOINT=https://192.168.222.199:6385/v1/
+IRONIC_INSPECTOR_ENDPOINT=https://192.168.222.199:5050/v1/
+CACHEURL=http://192.168.222.199/images
+IRONIC_FAST_TRACK=true
+IRONIC_KERNEL_PARAMS=console=ttyS0
+IRONIC_INSPECTOR_VLAN_INTERFACES=all
```
dtantsur, line 10: Will this consume ironic-image from main? If so, you need to set `USE_IRONIC_INSPECTOR=true`.

mquhuy:
@tuminoid I cannot say for sure, but if inspector removal doesn't work with these kustomization, then it will fail on e2e as well. I think we can merge this PR, and fix everything later when/if it fails.
@dtantsur Thanks. May I ask does it mean BMO cannot work without ironic inspector, or it means the ironic image from main needs to adapt?

config/overlays/e2e-release-0.3/ironic.env
lentzi90: Yes, we use the same ironic image for all jobs currently and that is the latest. Please add it @mquhuy

lentzi90: Wait a second, this is for BMO. In #1544 you did not touch those, do we need to?
- BMO: https://github.com/metal3-io/baremetal-operator/blob/main/config/overlays/e2e/ironic.env
- Ironic: https://github.com/metal3-io/baremetal-operator/blob/main/ironic-deployment/overlays/e2e/ironic_bmo_configmap.env

Does BMO need to have the variable or is it only relevant for Ironic?

dtantsur: BMO requires Inspector before 0.5.0. Does it answer your question?

PR #1527, opened 2024-01-22
 🌱 Enable parallel testing in E2E

### LINKED ISSUE #1443: E2e: enable parallel testing

mquhuy:
Fixes #1443

### DIFF
--- a/hack/ci-e2e.sh
+++ b/hack/ci-e2e.sh
@@ -38,6 +38,7 @@ export PATH="${PATH}:/usr/local/go/bin"
 "${REPO_ROOT}/hack/e2e/ensure_htpasswd.sh"
 # CAPI test framework uses kubectl in the background
 "${REPO_ROOT}/hack/e2e/ensure_kubectl.sh"
+"${REPO_ROOT}/hack/e2e/ensure_yq.sh"
 
 # Build the container image with e2e tag (used in tests)
 IMG=quay.io/metal3-io/baremetal-operator:e2e make docker
@@ -59,39 +60,20 @@ minikube start
 # Load the BMO e2e image into it
 minikube image load quay.io/metal3-io/baremetal-operator:e2e
 
-# Create libvirt domain
-VM_NAME="bmo-e2e-0"
-export BOOT_MAC_ADDRESS="00:60:2f:31:81:01"
-
-"${REPO_ROOT}/tools/bmh_test/create_vm.sh" "${VM_NAME}" "${BOOT_MAC_ADDRESS}"
-
 # This IP is defined by the network we created above.
 IP_ADDRESS="192.168.222.1"
 
-# These variables are used by the tests. They override variables in the config file.
-# These are the VBMC defaults (used since we did not specify anything else for `vbmc add`).
-export BMC_USER=admin
-export BMC_PASSWORD=password
-
 if [[ "${BMO_E2E_EMULATOR}" == "vbmc" ]]; then
-	# VBMC variables
-  VBMC_PORT="16230"
-  export BMC_ADDRESS="ipmi://${IP_ADDRESS}:${VBMC_PORT}"
-
   # Start VBMC
   docker run --name vbmc --network host -d \
     -v /var/run/libvirt/libvirt-sock:/var/run/libvirt/libvirt-sock \
     -v /var/run/libvirt/libvirt-sock-ro:/var/run/libvirt/libvirt-sock-ro \
     quay.io/metal3-io/vbmc
 
-  # Add BMH VM to VBMC
-  "${REPO_ROOT}/tools/bmh_test/vm2vbmc.sh" "${VM_NAME}" "${VBMC_PORT}"
 
 elif [[ "${BMO_E2E_EMULATOR}" == "sushy-tools" ]]; then
   # Sushy-tools variables
   SUSHY_EMULATOR_FILE="${REPO_ROOT}"/test/e2e/sushy-tools/sushy-emulator.conf
-  SUSHY_PORT="8000"
-  export BMC_ADDRESS="redfish+http://${IP_ADDRESS}:${SUSHY_PORT}/redfish/v1/Systems/${VM_NAME}"
 
   # Start sushy-tools
   docker run --name sushy-tools -d --network host \
@@ -104,6 +86,13 @@ else
   echo "Invalid e2e emulator specified: ${BMO_E2E_EMULATOR}"
   exit 1
 fi
+
+export E2E_BMCS_CONF_FILE="${REPO_ROOT}/test/e2e/config/bmcs-${BMO_E2E_EMULATOR}.yaml"
+"${REPO_ROOT}/hack/create_bmcs.sh" "${E2E_BMCS_CONF_FILE}" baremetal-e2e
+
+# Set the number of ginkgo processes to the number of BMCs
+n_vms=$(yq '. | length' "${E2E_BMCS_CONF_FILE}")
+export GINKGO_NODES="${n_vms}"
 
 # Image server variables
 CIRROS_VERSION="0.6.2"
@@ -152,6 +141,7 @@ echo "IRONIC_HTPASSWD=$(htpasswd -n -b -B "${IRONIC_USERNAME}" "${IRONIC_PASSWOR
   "${IRONIC_OVERLAY}/ironic-htpasswd"
 echo "INSPECTOR_HTPASSWD=$(htpasswd -n -b -B "${IRONIC_INSPECTOR_USERNAME}" \
   "${IRONIC_INSPECTOR_PASSWORD}")" > "${IRONIC_OVERLAY}/ironic-inspector-htpasswd"
+
 
 # We need to gather artifacts/logs before exiting also if there are errors
 set +e
--- a/hack/create_bmcs.sh
+++ b/hack/create_bmcs.sh
@@ -1,0 +1,35 @@
+#!/usr/bin/env bash
+#
+# This script reads BMC information in a config file and prepare VMs 
+# whose info match those config
+#
+set -eux
+
+REPO_ROOT=$(realpath "$(dirname "${BASH_SOURCE[0]}")/..")
+cd "${REPO_ROOT}"
+
+CONFIG_FILE=$1
+NETWORK=${2:-"baremetal-e2e"}
+
+readarray -t BMCS < <(yq e -o=j -I=0 '.[]' "${CONFIG_FILE}")
+
+for bmc in "${BMCS[@]}"; do
+  address=$(echo "${bmc}" | jq -r '.address')
+  bootMacAddress=$(echo "${bmc}" | jq -r '.bootMacAddress')
+  hostName=$(echo "${bmc}" | jq -r '.hostName')
+  ipAddress=$(echo "${bmc}" | jq -r '.ipAddress')
+
+  # Add the the VM to the network host list
+  virsh -c qemu:///system net-update "${NETWORK}" add-last ip-dhcp-host \
+    "<host mac='${bootMacAddress}' name='${hostName}' ip='${ipAddress}' />" \
+    --live --config --parent-index 0
+
+  # Create VM
+  "${REPO_ROOT}/tools/bmh_test/create_vm.sh" "${hostName}" "${bootMacAddress}"
+
+  # Add BMH VM to VBMC
+  if [[ "${address}" =~ "ipmi://" ]]; then
+    vbmc_port="${address##*:}"
+    "${REPO_ROOT}/tools/bmh_test/vm2vbmc.sh" "${hostName}" "${vbmc_port}"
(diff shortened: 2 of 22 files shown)


hack/ci-e2e.sh
```diff
@@ -58,38 +58,35 @@ minikube start
 minikube image load quay.io/metal3-io/baremetal-operator:e2e
 
 # Create libvirt domain
-VM_NAME="bmo-e2e-0"
-export BOOT_MAC_ADDRESS="00:60:2f:31:81:01"
+for vm_idx in $(seq 0 1); do
```
tuminoid, line 61:
```suggestion
for vm_idx in 0 1; do
```
Two item list does not warrant seq and a subshell.

hack/ci-e2e.sh
```diff
@@ -58,38 +58,35 @@ minikube start
...
-
   # Start VBMC
   docker run --name vbmc --network host -d \
     -v /var/run/libvirt/libvirt-sock:/var/run/libvirt/libvirt-sock \
     -v /var/run/libvirt/libvirt-sock-ro:/var/run/libvirt/libvirt-sock-ro \
     quay.io/metal3-io/vbmc
 
-  # Add BMH VM to VBMC
-  "${REPO_ROOT}/tools/bmh_test/vm2vbmc.sh" "${VM_NAME}" "${VBMC_PORT}"
+  for vm_idx in $(seq 0 1); do
```
tuminoid, line 78:
```suggestion
  for vm_idx in 0 1; do
```
Same

test/e2e/config/ironic-vbmc.yaml
```diff
@@ -0,0 +1,73 @@
+images:
+# Use locally built e2e images
+- name: quay.io/metal3-io/baremetal-operator:e2e
+  loadBehavior: tryLoad
+- name: quay.io/metal3-io/ironic:e2e
+  loadBehavior: tryLoad
+# Save some time and network by using cached images if available
+- name: quay.io/metal3-io/baremetal-operator:release-0.4
```
tuminoid, line 8: Why `release-0.4`, it doesn't even have the BMO e2e?

test/e2e/config/ironic-vbmc.yaml
```diff
@@ -0,0 +1,73 @@
...
+  DEPLOY_BMO: "true"
+  DEPLOY_CERT_MANAGER: "true"
+  BMO_KUSTOMIZATION: "../../config/overlays/e2e"
+  IRONIC_KUSTOMIZATION: "../../ironic-deployment/overlays/e2e"
+
+  # These settings are for the separate cluster that are used for upgrade tests
+  UPGRADE_DEPLOY_IRONIC: "true"
+  UPGRADE_DEPLOY_BMO: "true"
+  UPGRADE_DEPLOY_CERT_MANAGER: "true"
+  UPGRADE_BMO_KUSTOMIZATION_FROM: "../../config/overlays/e2e-release-0.4"
```
tuminoid, line 46: We do upgrade from 0.4? Why not from 0.5?

tuminoid (commented):
Minor nits, and also a question.

test/e2e/config/ironic-vbmc.yaml
mquhuy: These two new files are direct replacement of the `ironic.yaml`, so the content was just copied over. Using another upgrade from version is another story. Please check https://github.com/metal3-io/baremetal-operator/pull/1523

test/e2e/config/ironic-vbmc.yaml
mquhuy: It's for upgrade test. We will install this version, then upgrade it to main.

hack/ci-e2e.sh
mquhuy: Thanks. My idea was to set any number of parallel tests, but still having a bit of dilemma.

hack/ci-e2e.sh
```diff
@@ -105,6 +87,19 @@ else
   exit 1
 fi
 
+export E2E_BMCS_CONF_FILE="${REPO_ROOT}/test/e2e/config/bmcs-${BMO_E2E_EMULATOR}.yaml"
+"${REPO_ROOT}/hack/create_bmcs.sh" "${E2E_BMCS_CONF_FILE}" baremetal-e2e
+
+# Set the number of ginkgo processes to the smaller number in
+# the number of host CPUs and the number of BMCs
+n_cpus=$(nproc)
```
lentzi90, line 95: Not sure this is a good idea. Each VM we create has 2 vcpu so we would quickly use up all the CPU I think :smile:

test/e2e/bmc.go
```diff
@@ -0,0 +1,27 @@
...
+
+// BMC defines a BMH to use in the tests
+type BMC struct {
+	User           string `yaml:"user,omitempty"`
+	Password       string `yaml:"password,omitempty"`
+	Address        string `yaml:"address,omitempty"`
+	BootMacAddress string `yaml:"bootMacAddress,omitempty"`
+	HostName       string `yaml:"hostName,omitempty"`
+	IpAddress      string `yaml:"ipAddress,omitempty"`
+	SSHPort        string `yaml:"sshPort,omitempty"`
```
lentzi90, line 18: Could you add doc strings for all of these? I think especially the IP address and ssh port we should point out that they are optional and used only for the ssh check.

hack/ci-e2e.sh
mquhuy: Yeah, perhaps just use all the VMs then? Or do you think we should limit it to 2 only?

test/e2e/bmc.go
mquhuy: Sure!

hack/ci-e2e.sh
lentzi90: Using all the VMs sounds good. I would also be happy with not doing anything at all, i.e. just pass through. Then we can just set the variable before calling the script

lentzi90:
How do you pick which BMC to use for which ginkgo worker?
I would expect that we need to do some kind of selection [here](https://github.com/metal3-io/baremetal-operator/blob/8bab491f9deb21ca5e01f94ec110ff979c9d89f4/test/e2e/e2e_suite_test.go#L154-L164). Probably using [GinkgoParallelProcess](https://pkg.go.dev/github.com/onsi/ginkgo/v2#GinkgoParallelProcess) to pick an index in the list of BMCs

mquhuy:
> How do you pick which BMC to use for which ginkgo worker? I would expect that we need to do some kind of selection [here](https://github.com/metal3-io/baremetal-operator/blob/8bab491f9deb21ca5e01f94ec110ff979c9d89f4/test/e2e/e2e_suite_test.go#L154-L164). Probably using [GinkgoParallelProcess](https://pkg.go.dev/github.com/onsi/ginkgo/v2#GinkgoParallelProcess) to pick an index in the list of BMCs

Thanks. I am now doing that in the tests. But yes, it should be better to handle it in the place you mentioned.

hack/create_bmcs.sh
```diff
@@ -0,0 +1,35 @@
...
+#
+set -eux
+
+REPO_ROOT=$(realpath "$(dirname "${BASH_SOURCE[0]}")/..")
+cd "${REPO_ROOT}" || exit 1
+
+CONFIG_FILE=$1
+NETWORK=${2:-"baremetal-e2e"}
+
+readarray -t BMCs < <(yq e -o=j -I=0 '.[]' "${CONFIG_FILE}")
```
tuminoid, line 14:
```suggestion
readarray -t BMCS < <(yq e -o=j -I=0 '.[]' "${CONFIG_FILE}")
```
Not a fan of mixed-case variables

hack/create_bmcs.sh
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
+#
+# This script reads BMC information in a config file and prepare VMs 
+# whose info match those config
+#
+set -eux
+
+REPO_ROOT=$(realpath "$(dirname "${BASH_SOURCE[0]}")/..")
+cd "${REPO_ROOT}" || exit 1
```
tuminoid, line 9:
```suggestion
cd "${REPO_ROOT}"
```
Unnecessary as we have `set -e` which will exit on error.

hack/e2e/ensure_yq.sh
```diff
@@ -0,0 +1,24 @@
...
+
+set -eux
+
+USR_LOCAL_BIN="/usr/local/bin"
+YQ_VERSION="v4.40.5"
+
+# Check if yq tool is installed and install it if not
+verify_yq()
+{
+    if ! [ -x "$(command -v yq)" ]; then
```
tuminoid, line 11:
```suggestion
    if ! [[ -x "$(command -v yq)" ]]; then
```

tuminoid (commented):
Few nits.

mquhuy:
> Few nits.

Thanks a lot. I fixed them now.

lentzi90 (commented):
Ok I'm happy with this if you can get the tests to pass!
I guess it is probably a matter of tweaking the intervals or we are a bit resource constrained in the Jenkins worker...

mquhuy:
> Ok I'm happy with this if you can get the tests to pass! I guess it is probably a matter of tweaking the intervals or we are a bit resource constrained in the Jenkins worker...

Thank you. I reckon using 2 VMs might be more likely to succeed, I just want to test if 3 VMs is feasible. I'll test it out a couple of times more to figure out.

lentzi90:
For reference, the workers we currently use have 8 CPU and 16 GB memory. We can change this later if we want :slightly_smiling_face:

mquhuy:
> For reference, the workers we currently use have 8 CPU and 16 GB memory. We can change this later if we want 🙂

@lentzi90 I guess it explained the random flakes in the tests. 3 VMs would take 6 CPUs, so it means we have to run 3 threads for the tests on 2 remaining CPUs.
Switching to 2 VMs made the tests pass. We can increase it later if we want though. A 2-VM setup take about 25mins, while 3-VM one takes 20 mins.

Btw, could you help me override the integration tests?

lentzi90 (commented):
/lgtm

I don't have privileges to override context in this repo unfortunately

tuminoid:
/override test-centos-e2e-integration-main  test-ubuntu-integration-main

test/e2e/config/ironic.yaml
```diff
@@ -22,29 +22,17 @@ variables:
...
-  BMC_ADDRESS: "redfish+http://192.168.222.1:8000/redfish/v1/Systems/bmo-e2e-0"
-  BOOT_MAC_ADDRESS: "00:60:2f:31:81:01"
-  EXPECTED_HOST_NAME: "bmo-e2e-0"
-
   IMAGE_URL: "http://192.168.222.1/cirros-0.6.2-x86_64-disk.img"
   IMAGE_CHECKSUM: "c8fc807773e5354afe61636071771906"
   CERT_MANAGER_VERSION: "v1.13.1"
   SSH_CHECK_PROVISIONED: "true"
-  IP_BMO_E2E_0: "192.168.222.122"
-  SSH_PORT: "22"
```
maxrantil, line 47: Do all tests pass without `SSH_PORT` ? It is used when checking to verify the node's boot source. Maybe it used 22 by default. But I think it would be good to still have it because the variable is used in the go code and someone might want to use another port sometime?

maxrantil: Oh NVM! I see that you have changed the code.

maxrantil:
LGTM

kashifest:
This is cool. Great work @mquhuy 
/approve

PR #1530, opened 2024-01-24
🌱 Set USE_IRONIC_INSPECTOR explicitly (to true for now)

dtantsur:
This is to avoid changing everything (ironic-image and BMO) in a lockstep.


iurygregory:
LGTM

PR #1532, opened 2024-01-24
🌱 Allow BMC address to be updated 

MahnoorAsghar:
Allow BMC address to be updated if the BMH is in the Registering state or is detached.

This allows the BMC address to be updated if there are changes in the BMC network, or if a mistake was made when setting it the first time.

### DIFF
--- a/apis/metal3.io/v1alpha1/baremetalhost_validation.go
+++ b/apis/metal3.io/v1alpha1/baremetalhost_validation.go
@@ -75,8 +75,12 @@ func (host *BareMetalHost) validateChanges(old *BareMetalHost) []error {
 		errs = append(errs, err...)
 	}
 
-	if old.Spec.BMC.Address != "" && host.Spec.BMC.Address != old.Spec.BMC.Address {
-		errs = append(errs, errors.New("BMC address can not be changed once it is set"))
+	if old.Spec.BMC.Address != "" &&
+		host.Spec.BMC.Address != old.Spec.BMC.Address &&
+		host.Status.OperationalStatus != OperationalStatusDetached &&
+		host.Status.Provisioning.State != StateRegistering {
+		errs = append(errs, errors.New("BMC address can not be changed if the BMH is not "+
+			"in the Registering state, or if the BMH is not detached."))
 	}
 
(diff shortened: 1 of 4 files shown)


apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -75,7 +75,10 @@ func (host *BareMetalHost) validateChanges(old *BareMetalHost) []error {
 		errs = append(errs, err...)
 	}
 
-	if old.Spec.BMC.Address != "" && host.Spec.BMC.Address != old.Spec.BMC.Address {
+	if old.Spec.BMC.Address != "" &&
+		host.Spec.BMC.Address != old.Spec.BMC.Address &&
+		host.Status.OperationalStatus != OperationalStatusDetached &&
+		host.Status.Provisioning.State != StateRegistering {
 		errs = append(errs, errors.New("BMC address can not be changed once it is set"))
```
elfosardo, line 82: not sure this message is still valid as we can now update the BMC address

MahnoorAsghar: I have corrected the message now

derekhiggins:
@MahnoorAsghar I don't see the bmc address of the node in ironic getting updated, does that happen as a result of the validation passing?

MahnoorAsghar:
> @MahnoorAsghar I don't see the bmc address of the node in ironic getting updated, does that happen as a result of the validation passing?

Very good question, Derek. There are two cases for a BMC address update: 
- Updating the address when a BMH is in the Registering state: for that, I have now added the changes needed to update the address in Ironic.
- Updating the address when the BMH is detached, the node is deleted from Ironic and when the detachment annotation is removed, the node is re-created in Ironic with the new BMC address.

tuminoid:
Wakey wakey Jenkins…

tuminoid:
It seems this PR is in some Jenkins black hole. We've seen this before couple times, we're trying to figure out who to fix the situation.

At least the e2e workflow passed now.

tuminoid:
One last test as it worked immediately on https://github.com/metal3-io/baremetal-operator/pull/1543

PR #1537, opened 2024-02-01
🐛 Fix VM selection in clean-e2e.sh

### LINKED ISSUE #1531: BMO e2e cleanup potentially leaving VMs?

See https://jenkins.nordix.org/view/Metal3/job/metal3-bmo-e2e-test-pull/214/

```
[2024-01-24T11:43:09.124Z] + virt-install --connect qemu:///system --name bmo-e2e-0 --description 'Virtualized BareMetalHost' --osinfo=ubuntu-lts-latest --ram=4096 --vcpus=2 --disk size=20 --graphics=none --console pty,target_type=serial --serial file,path=/var/log/libvirt/qemu/bmo-e2e-0-serial0.log --xml ./devices/serial/@type=pty --xml ./devices/serial/log/@file=/var/log/libvirt/qemu/bmo-e2e-0-serial0.log --xml ./devices/serial/log/@append=on --pxe --network network=baremetal-e2e,mac=00:60:2f:31:81:01 --noautoconsole

[2024-01-24T11:43:09.124Z] ERROR    The MAC address '00:60:2f:31:81:01' is in use by another virtual machine. (Use --check mac_in_use=off or --check all=off to override)

script returned exit code 1
```

My suspicion is that there is an issue with the cleanup script that somehow left a VM behind from the previous job. Then when this job started we got the conflict.
I could easily be mistaken though. Perhaps minikube somehow happened to get exactly the same MAC as we are giving the BMH VM? :exploding_head: 

Edit:
I think I found the issue now!
https://github.com/metal3-io/baremetal-operator/blob/53281e6ee4010971059f9ba85dd402083d9c4cad/hack/clean-e2e.sh#L11
This is not using the `qemu:///system` socket so it will not find the VMs.
I don't think we should select all VMs either. Somehow I missed that while reviewing.
We should instead do as [here](https://github.com/metal3-io/baremetal-operator/blob/main/tools/bmh_test/clean_local_bmh_test_setup.sh). In fact we should probably reuse that script and just take the prefix for what VMs to delete from a variable.

/kind bug

mquhuy:
Fixes #1531

This PR takes the `tools/bmh_test/clean_local_bmh_test_setup.sh` into use:
- Removed `-e` as many operations could fail, but the script should keep going

### DIFF
--- a/hack/clean-e2e.sh
+++ b/hack/clean-e2e.sh
@@ -8,17 +8,7 @@ docker rm -f vbmc
 docker rm -f image-server-e2e
 docker rm -f sushy-tools
 
-virsh_vms=$(virsh list --name --all)
-
-for vm in ${virsh_vms}; do
-  if [[ "${vm}" =~ "bmo-e2e-" ]]; then
-    virsh -c qemu:///system destroy --domain "${vm}"
-    virsh -c qemu:///system undefine --domain "${vm}" --nvram --remove-all-storage
-  fi
-done
-
-virsh -c qemu:///system net-destroy baremetal-e2e
-virsh -c qemu:///system net-undefine baremetal-e2e
+"${REPO_ROOT}/tools/bmh_test/clean_local_bmh_test_setup.sh" "^bmo-e2e-"
 
 rm -rf "${REPO_ROOT}/test/e2e/_artifacts"
 rm -rf "${REPO_ROOT}"/artifacts-*
--- a/tools/bmh_test/clean_local_bmh_test_setup.sh
+++ b/tools/bmh_test/clean_local_bmh_test_setup.sh
@@ -1,24 +1,24 @@
 #!/usr/bin/env bash
 
-set -eux
+set -ux
 
+BMH_NAME_REGEX="${1:-^bmh-test-}"
 # Get a list of all virtual machines
-VM_LIST=$(virsh -c qemu:///system list --all --name | grep '^bmh-test-') || true
+VM_LIST=$(virsh -c qemu:///system list --all --name | grep "${BMH_NAME_REGEX}")
 
 if [[ -n "${VM_LIST}" ]]; then
     # Loop through the list and delete each virtual machine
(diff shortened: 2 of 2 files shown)


maxrantil:
LGTM

hack/clean-e2e.sh
```diff
@@ -4,21 +4,10 @@ REPO_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
 cd "${REPO_ROOT}" || exit 1
 
 minikube delete
-docker rm -f vbmc
```
lentzi90, line 7: Could you add this back? It feels weird to delete sushy-tools here but not vbmc.
In the long run we can move all of it to the other script but for now I think we can keep it here and just fix the bug.

lentzi90 (commented):
Great! Just one small comment

hack/clean-e2e.sh
mquhuy: Sure. I removed it since it felt weird to me that we clean it in both places while it may not even be there, but no big issue.

lentzi90 (commented):
/lgtm

peppi-lotta:
LGTM

tools/bmh_test/clean_local_bmh_test_setup.sh
```diff
@@ -1,24 +1,24 @@
...
         virsh -c qemu:///system undefine --domain "${vm_name}" --remove-all-storage
-        kubectl delete baremetalhost "${vm_name}" || true
+        kubectl delete baremetalhost "${vm_name}"
     done
 else
     echo "No virtual machines found. Skipping..."
 fi
 
 # Clear vbmc
-docker stop vbmc
```
kashifest, line 20: whats the harm if we stop it first, why are we removing this? Is there a possibility that it might not exist?

lentzi90: The `-f` makes it unnecessary. We remove it no matter if it runs or not

PR #1543, opened 2024-02-02
:seedling: Reduce GH fixture test GINKGO_NODES to 1

mquhuy:
We're facing many failures in GH-action-based E2E fixture tests. This is likely due to the github VM is too small and cannot handle two gingko threads.

### DIFF
--- a/.github/workflows/e2e-fixture-test.yml
+++ b/.github/workflows/e2e-fixture-test.yml
@@ -35,4 +35,5 @@ jobs:
       env:
         E2E_CONF_FILE: ${{ github.workspace }}/test/e2e/config/fixture.yaml
         USE_EXISTING_CLUSTER: "false"
+        GINKGO_NODES: 1
       run: make test-e2e


tuminoid (commented):
LGTM, but

/cc @lentzi90

tuminoid:
/retitle :seedling: Reduce GH fixture test GINKGO_NODES to 1

lentzi90 (commented):
Let's try to figure out if/why this is an issue first. The tests should not be heavy when running with BMO in test mode. Perhaps the test mode is just not handling concurrency well?

tuminoid:
This might actually be bug in BMO as found by @lentzi90 . Let's investigate that a bit more, and this might not be needed.
/hold

lentzi90:
Yup, I can put it here also so it is more obvious what we are discussing.
Normally the live-ISO test takes <40 seconds in the fixture job. When we hit this particular issue, it times out after 300 seconds!
I have seen this happen when going from `deprovisioning` to `available` and also from `provisioning` to `provisioned`. I have a feeling we are dealing with a subtle bug rather than performance issues

tuminoid:
Random test.

tuminoid (commented):
/lgtm
/unhold

lentzi90:
/cc @kashifest 
Could you approve? We believe this could stabilize the fixture tests a bit

PR #1544, opened 2024-02-02
⚠️ Remove Inspector from all deployments scripts

dtantsur:
With this change, the scripts will no longer install ironic-inspector,
relying on the ironic's built-in inspection instead.

Note that BMO itself still supports Inspector as the inspection backend,
the change only affects deploy.sh and Kustomization configs.


kashifest:
I think for this to pass the centos tests, we need to get rid of some hard assumption on BMO kustomization on dev env side, I can see from the logs that BMO is complaining about `"Could not contact ironic-inspector for version discovery: Unable to find a version discovery document at http://172.22.0.2:5050, the service is unavailable or misconfigured. Required version range (any - any), version hack disabled.","current":"inspect failed","target":"manageable"` . Also I see this https://github.com/metal3-io/metal3-dev-env/blob/main/03_launch_mgmt_cluster.sh#L69 in dev-ev which we shouldnt need anymore after this PR right ?

dtantsur:
https://github.com/metal3-io/metal3-dev-env/pull/1346

lentzi90:
I would also be interested in the optional tests

dtantsur:

Connectivity issues

lentzi90:
> > I would also be interested in the optional tests /test metal3-bmo-e2e-test-optional-pull
> 
> This test is now trying to find `/home/metal3ci/slave_root/workspace/metal3-bmo-e2e-test-optional-pull/config/overlays/e2e-release-0.4/ironic-inspector-username` , so possibly deploying BMO from release-0.4 branch and upgrading to latest, so could it be that this PR needs to also make sure that when we deploy older branches the inspector configs are present?

Yeah I think the issue is that `ci-e2e.sh` is generating credentials for both overlays (`e2e` and `e2e-releas-0.4`). The 0.4 overlay still needs them and fails when we do not generate them. Just leaving that part of the script as is should solve it.

hack/ci-e2e.sh
```diff
@@ -118,29 +118,19 @@ IRONIC_OVERLAY="${REPO_ROOT}/ironic-deployment/overlays/e2e"
...
 echo "${IRONIC_USERNAME}" > "${BMO_OVERLAY}/ironic-username"
 echo "${IRONIC_PASSWORD}" > "${BMO_OVERLAY}/ironic-password"
-echo "${IRONIC_INSPECTOR_USERNAME}" > "${BMO_OVERLAY}/ironic-inspector-username"
-echo "${IRONIC_INSPECTOR_PASSWORD}" > "${BMO_OVERLAY}/ironic-inspector-password"
 
 BMO_UPGRADE_FROM_OVERLAY="${REPO_ROOT}/config/overlays/e2e-release-0.4"
 echo "${IRONIC_USERNAME}" > "${BMO_UPGRADE_FROM_OVERLAY}/ironic-username"
 echo "${IRONIC_PASSWORD}" > "${BMO_UPGRADE_FROM_OVERLAY}/ironic-password"
-echo "${IRONIC_INSPECTOR_USERNAME}" > "${BMO_UPGRADE_FROM_OVERLAY}/ironic-inspector-username"
-echo "${IRONIC_INSPECTOR_PASSWORD}" > "${BMO_UPGRADE_FROM_OVERLAY}/ironic-inspector-password"
```
lentzi90, line 133: This part is still needed since we have inspector in release-0.4

lentzi90: I was wrong. The issue was not what I thought.

For reference, the BMO e2e tests currently work like this
- Ironic is deployed from [this](https://github.com/metal3-io/baremetal-operator/blob/main/ironic-deployment/overlays/e2e/kustomization.yaml), which is using [this ironic manifest](https://github.com/metal3-io/baremetal-operator/blob/main/ironic-deployment/base/ironic.yaml) that has the image `quay.io/metal3-io/ironic` (latest).
- BMO is deployed from [this](https://github.com/metal3-io/baremetal-operator/blob/main/config/overlays/e2e/kustomization.yaml) normally, which is using a freshly built BMO image every time.
- BMO is deployed from [this](https://github.com/metal3-io/baremetal-operator/blob/main/config/overlays/e2e-release-0.4/kustomization.yaml) for upgrade tests, which is using `quay.io/metal3-io/baremetal-operator:release-0.4`.

I saw the optional test fail with 

```
lstat /home/metal3ci/slave_root/workspace/metal3-bmo-e2e-test-optional-pull/config/overlays/e2e-release-0.4/ironic-inspector-username: no such file or directory"
```

and thought that we still had inspector there, but of course the test is just upgrading BMO, Ironic stays the same throughout.

So the *actual* reason it failed is just that the BMO e2e release-0.4 overlay is expecting the files to be there. You just have to remove this:

https://github.com/metal3-io/baremetal-operator/blob/ab253c0f3ee643b9b7a883c0b0b4cc3a4485bbd6/config/overlays/e2e-release-0.4/kustomization.yaml#L41-L44

dtantsur: Yeah, good point. But that only works with BMO 0.5 and newer, the previous versions require Inspector.

lentzi90: Ok I think I understand. So for release-0.4 we would need to deploy inspector and that means we need a separate ironic overlay. We can do that in the same way as the BMO release-0.4 overlay, i.e. point to the release branch where we still have inspector.
https://github.com/metal3-io/baremetal-operator/blob/3cd8e3871eb623a47a835904bd7d4cf4b0073707/config/overlays/e2e-release-0.4/kustomization.yaml#L5

lentzi90: CC @mquhuy this is relevant for you also if you start working on ironic upgrades.

dtantsur: So, which upgrades do we test now on master? If only from 0.5, we can proceed with your earlier suggestion.

mquhuy: We are actually planning to have multiple BMO upgrade tests as in #1451 . Is it possible for `release-0.5` and BMO `main` to use `Inspector` still? In that case I think it would be easier to just keep `inspector` for upgrade tests, until we no longer support release-0.3 and release-0.4

lentzi90: @mquhuy we should test going from having inspector to not having it. That is the most valuable test!
For now we do not touch Ironic so all whatever we deploy to start with is what you use also after BMO upgrade, that means we will have to use inspector for at least upgrades from 0.4 and 0.3. I think the way forward is to create one overlay for ironic with inspector and one without inspector. For release-0.5 upgrade we can use the one without inspector then.

lentzi90: @dtantsur yes for now we only upgrade from 0.5 so you can completely remove inspector. (It was 0.4 until recently which is probably why the test failed on this PR. If you rebase I think it will be fine.)

mquhuy: Yes, it makes sense

mquhuy:
Upgrade test is in the optional e2e test

dtantsur:
Huh, BMO is trying to mount the inspector credentials.. Which are not there.

dtantsur:
All passed \o/

config/README.md
```diff
@@ -21,9 +21,8 @@ in `hack/ci-e2e.sh`, and put them directly in the `e2e` overlays.
    - **basic-auth** - Enable basic authentication. Note that the basic-auth
       component is missing the actual credentials. This is on purpose, to make
       sure that the user is setting the password instead of using sample
-      credentials. The required secrets are:
+      credentials. The required secrets is called:
```
mquhuy, line 24: secret?

lentzi90 (commented):
/lgtm

dtantsur:
/cc @kashifest @zaneb 

Could either of you approve?

PR #1548, opened 2024-02-06
🐛 Fix an indentation error in kustomization file regarding ironic configmap generator

maxrantil:
As title describes


lentzi90 (commented):
/lgtm

PR #1549, opened 2024-02-08
🌱 Allow BMC address to be updated

MahnoorAsghar:
Allow BMC address to be updated if the BMH is in the Registering state or is detached.

This allows the BMC address to be updated if there are changes in the BMC network, or if a mistake was made when setting it the first time.

### DIFF
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -3,6 +3,8 @@ package ironic
 import (
 	"fmt"
 	"net"
+	"reflect"
+	"regexp"
 	"strings"
 	"time"
 
@@ -443,6 +445,26 @@ func (p *ironicProvisioner) ValidateManagementAccess(data provisioner.Management
 
 		updater.SetTopLevelOpt("name", ironicNodeName(p.objectMeta), ironicNode.Name)
 
+		var bmcAddressChanged bool
+		newAddress := make(map[string]interface{})
+		ironicAddress := make(map[string]interface{})
+		reg := regexp.MustCompile("_address$")
+		for key, value := range driverInfo {
+			if reg.MatchString(key) {
+				newAddress[key] = value
+				break
+			}
+		}
+		for key, value := range ironicNode.DriverInfo {
+			if reg.MatchString(key) {
+				ironicAddress[key] = value
+				break
+			}
+		}
+		if !reflect.DeepEqual(newAddress, ironicAddress) {
+			bmcAddressChanged = true
+		}
+
 		// When node exists but has no assigned port to it by Ironic and actuall address (MAC) is present
 		// in host config and is not allocated to different node lets try to create port for this node.
 		if p.bootMACAddress != "" {
@@ -472,9 +494,11 @@ func (p *ironicProvisioner) ValidateManagementAccess(data provisioner.Management
 		}
 
 		// The actual password is not returned from ironic, so we want to
-		// update the whole DriverInfo only if the credentials have changed
-		// otherwise we will be writing on every call to this function.
-		if credentialsChanged {
+		// update the whole DriverInfo only if the credentials or BMC address
+		// has changed, otherwise we will be writing on every call to this
+		// function.
+		if credentialsChanged || bmcAddressChanged {
+			p.log.Info("Updating driver info because the credentials and/or the BMC address changed")
 			updater.SetTopLevelOpt("driver_info", driverInfo, ironicNode.DriverInfo)
 		}
 
--- a/pkg/provisioner/ironic/validatemanagementaccess_test.go
+++ b/pkg/provisioner/ironic/validatemanagementaccess_test.go
@@ -892,6 +892,121 @@ func TestValidateManagementAccessMalformedBMCAddress(t *testing.T) {
 	assert.Equal(t, "failed to parse BMC address information: failed to parse BMC address information: parse \"<ipmi://192.168.122.1:6233>\": first path segment in URL cannot contain colon", result.ErrorMessage)
 }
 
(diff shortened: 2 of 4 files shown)


kashifest:
Finally jenkins woke up

tuminoid:
For the record, here is the original one with reviews and discussions: https://github.com/metal3-io/baremetal-operator/pull/1532

pkg/provisioner/ironic/ironic.go
```diff
@@ -472,9 +503,11 @@ func (p *ironicProvisioner) ValidateManagementAccess(data provisioner.Management
...
 		// The actual password is not returned from ironic, so we want to
-		// update the whole DriverInfo only if the credentials have changed
-		// otherwise we will be writing on every call to this function.
-		if credentialsChanged {
+		// update the whole DriverInfo only if the credentials or BMC address
+		// has changed, otherwise we will be writing on every call to this
+		// function.
+		if credentialsChanged || bmcAddressChanged {
+			p.log.Info("Updating driver info because the credentials and/or the BMC address changed")
 			updater.SetTopLevelOpt("driver_info", driverInfo, ironicNode.DriverInfo)
```
dtantsur, line 511: This is calling the PATCH API to set the node's driver_info field to the value of `driverInfo` variable. Which does not seem to contain your changes (you're updating `ironicNode.DriverInfo`).

pkg/provisioner/ironic/validatemanagementaccess_test.go
```diff
@@ -892,6 +892,64 @@ func TestValidateManagementAccessMalformedBMCAddress(t *testing.T) {
 	assert.Equal(t, "failed to parse BMC address information: failed to parse BMC address information: parse \"<ipmi://192.168.122.1:6233>\": first path segment in URL cannot contain colon", result.ErrorMessage)
 }
 
+func TestValidateManagementAccessUpdateBMCAddress(t *testing.T) {
```
dtantsur, line 895: Could you add another test to see what happens if a user also changes the driver part (e.g. from IPMI to Redfish)? (I wonder if we really want to allow that.. probably?)

pkg/provisioner/ironic/ironic.go
```diff
@@ -443,6 +445,35 @@ func (p *ironicProvisioner) ValidateManagementAccess(data provisioner.Management
 
 		updater.SetTopLevelOpt("name", ironicNodeName(p.objectMeta), ironicNode.Name)
 
+		var bmcAddressChanged bool
+		var newAddressKey string
+		var newAddressValue interface{}
+		var ironicAddressKey string
+		var ironicAddressValue interface{}
+		for key, value := range driverInfo {
+			if found, _ := regexp.MatchString("address", key); found {
```
dtantsur, line 454: This is not very robust. At the very least, I'd use something like "_address$". Ideally, let the bmcAccess figure that out.

pkg/provisioner/ironic/ironic.go
```diff
@@ -443,6 +445,35 @@ func (p *ironicProvisioner) ValidateManagementAccess(data provisioner.Management
...
+		for key, value := range ironicNode.DriverInfo {
+			if found, _ := regexp.MatchString("address", key); found {
+				ironicAddressKey = key
+				ironicAddressValue = value
+				break
+			}
+		}
+		if (newAddressKey != ironicAddressKey) ||
+			(!reflect.DeepEqual(newAddressValue, ironicAddressValue)) {
+			delete(ironicNode.DriverInfo, ironicAddressKey)
```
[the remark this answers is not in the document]
dtantsur: I suspect it works by accident. Once you get here, you know there is a change, and you know that driverInfo already contains the right value (it's created this way). So you can probably drop lines 469-473, and everything will still work.

pkg/provisioner/ironic/validatemanagementaccess_test.go
MahnoorAsghar: If the user changes the driver part, the node's driver_info field is updated accordingly (with the new values and defaults), but the node's 'driver' and interfaces are left as it is, which can lead to en error state in Ironic if they end up incorrect. 
Whether we should allow this or not is indeed a good question. The user might have to responsible to ensure that all the node fields are correct, or BMO can update the entire node object as part of this change as well.

pkg/provisioner/ironic/ironic.go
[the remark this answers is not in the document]
MahnoorAsghar: Yes, it does work like you explained. I have dropped the lines 469-473 and it works just fine.

pkg/provisioner/ironic/ironic.go
MahnoorAsghar: I've changed the regex and tried to reduce the number of variables needed. Does it look better?

pkg/provisioner/ironic/ironic.go
MahnoorAsghar: The changes are already present in the driverInfo variable, like you noted above.

pkg/provisioner/ironic/ironic.go
```diff
@@ -443,6 +445,26 @@ func (p *ironicProvisioner) ValidateManagementAccess(data provisioner.Management
...
+		}
+		for key, value := range ironicNode.DriverInfo {
+			if reg.MatchString(key) {
+				ironicAddress[key] = value
+				break
+			}
+		}
+		if !reflect.DeepEqual(newAddress, ironicAddress) {
+			bmcAddressChanged = true
+		}
```
dtantsur, line 466: nit:

 bmcAddressChanged := !reflect.DeepEqual(newAddress, ironicAddress)

PR #1553, opened 2024-02-13
🌱 Add ironic and bmo cleanup for upgrade BMO E2E tests

mquhuy:
This PR adds Ironic and BMO to the cleanup phase in upgrade tests. This is needed as a preparation for multiple upgrade tests (multiple versions of BMO upgrade + ironic upgrade), as the test has to start from a clean state (with no Ironic nor BMO).
We also move installation of Ironic and BMO to the `It{}`, as it's more related to the test itself, rather than a preparation.

Notice that this change has allowed the test to be ran multiple times using the same cluster, as followed:
```
export GINKGO_FOCUS=upgrade
hack/ci-e2e.sh # first time
export UPGRADE_DEPLOY_CERT_MANAGER=false
<extra exports normally handled by ci-e2e.sh>
make test-e2e # As many times as needed
```
We could apply this to the required e2e tests as well if running multiple times is a requirement. That's not the main focus of the PR though.

### DIFF
--- a/test/e2e/upgrade_test.go
+++ b/test/e2e/upgrade_test.go
@@ -175,13 +175,12 @@ var _ = Describe("BMO Upgrade", func() {
 		specName               = "upgrade"
 		secretName             = "bmc-credentials"
 		namespace              *corev1.Namespace
-		bmoIronicNamespace     string
+		bmoIronicNamespace     = "baremetal-operator-system"
 		upgradeClusterProvider bootstrap.ClusterProvider
 		upgradeClusterProxy    framework.ClusterProxy
 		bmh                    metal3api.BareMetalHost
 	)
 	BeforeEach(func() {
-		bmoIronicNamespace = "baremetal-operator-system"
 		var kubeconfigPath string
 
 		if useExistingCluster {
@@ -203,6 +202,7 @@ var _ = Describe("BMO Upgrade", func() {
 		framework.TryAddDefaultSchemes(scheme)
 		metal3api.AddToScheme(scheme)
 		upgradeClusterProxy = framework.NewClusterProxy("bmo-e2e-upgrade", kubeconfigPath, scheme)
+
 		if e2eConfig.GetVariable("UPGRADE_DEPLOY_CERT_MANAGER") != "false" {
 			By("Installing cert-manager on the upgrade cluster")
 			cmVersion := e2eConfig.GetVariable("CERT_MANAGER_VERSION")
@@ -215,12 +215,15 @@ var _ = Describe("BMO Upgrade", func() {
 			err = checkCertManagerAPI(upgradeClusterProxy)
 			Expect(err).NotTo(HaveOccurred())
 		}
+	})
 
+	It("Should upgrade BMO to latest version", func() {
 		if e2eConfig.GetVariable("UPGRADE_DEPLOY_IRONIC") != "false" {
 			// Install Ironic
+			ironicKustomization := e2eConfig.GetVariable("IRONIC_KUSTOMIZATION")
 			By("Installing Ironic on the upgrade cluster")
-			err := BuildAndApplyKustomize(ctx, &BuildAndApplyKustomizeInput{
-				Kustomization:       e2eConfig.GetVariable("IRONIC_KUSTOMIZATION"),
+			err := BuildAndApplyKustomization(ctx, &BuildAndApplyKustomizationInput{
+				Kustomization:       ironicKustomization,
 				ClusterProxy:        upgradeClusterProxy,
 				WaitForDeployment:   true,
 				WatchDeploymentLogs: true,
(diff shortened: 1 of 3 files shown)


test/e2e/upgrade_test.go
```diff
@@ -172,16 +173,16 @@ const hardwareDetailsRelease04 = `
...
-		upgradeClusterProxy    framework.ClusterProxy
-		bmh                    metal3api.BareMetalHost
+		specName                  = "upgrade"
+		secretName                = "bmc-credentials"
+		namespace                 *corev1.Namespace
+		bmoIronicNamespace        *corev1.Namespace
+		upgradeClusterProvider    bootstrap.ClusterProvider
+		upgradeClusterProxy       framework.ClusterProxy
+		bmh                       metal3api.BareMetalHost
+		_, bmoIronicCancelWatches = context.WithCancel(ctx)
```
lentzi90, line 183:
```suggestion
		bmoIronicCancelWatches context.CancelFunc
```

test/e2e/upgrade_test.go
```diff
@@ -203,6 +204,14 @@ var _ = Describe("BMO Upgrade", func() {
 		framework.TryAddDefaultSchemes(scheme)
 		metal3api.AddToScheme(scheme)
 		upgradeClusterProxy = framework.NewClusterProxy("bmo-e2e-upgrade", kubeconfigPath, scheme)
+
+		bmoIronicNamespace, bmoIronicCancelWatches = framework.CreateNamespaceAndWatchEvents(ctx, framework.CreateNamespaceAndWatchEventsInput{
```
lentzi90, line 208: It will be strange if we create the namespace if the user asked to not deploy BMO/Ironic. Why do we need to create it? Is it to watch the events?

test/e2e/upgrade_test.go
```diff
@@ -344,6 +352,7 @@ var _ = Describe("BMO Upgrade", func() {
 
 	AfterEach(func() {
 		cleanup(ctx, upgradeClusterProxy, namespace, cancelWatches, e2eConfig.GetIntervals("default", "wait-namespace-deleted")...)
+		cleanup(ctx, upgradeClusterProxy, bmoIronicNamespace, bmoIronicCancelWatches, e2eConfig.GetIntervals("default", "wait-namespace-deleted")...)
```
lentzi90, line 355: This will only delete the namespace and everything in it, but cluster scoped resources will stay. I think we should cleanup also the CRDs. How about we instead do something like `kubectl delete -k /path/to/upgrade/kustomization`? That should delete everything we added or modified when applying the kustomization so it should be pretty clean.

mquhuy: Thanks. I thought of something similar to that, but CAPI test framework does not currently support deletion of the resources, so doing so would require a bit bigger changes. 
Do we have any cluster scoped resources other than CRDs? IMO keeping the CRDs should be acceptable.

test/e2e/upgrade_test.go
lentzi90: Ok so we don't really have to create the namespace, just clean it up, right? I think it would be better to not create it. We can still delete it after the test.
If we create it like this we will get an error if the user already created it (using existing cluster and already deployed Ironic).

test/e2e/upgrade_test.go
mquhuy: Hmm. Yes you're right. Now I realized there might be nasty issues that we fail to tests if CRDs are left there.
I'll come up with a solution based on `kubectl remove -k` wow.

test/e2e/upgrade_test.go
mquhuy: What I was aiming to is to be able to delete it before the test as well, since the cleanup might fail after the previous test. But yes, might be creating it separately is not necessary, I simply did that to be able to use your cleanup() function.

test/e2e/upgrade_test.go
mquhuy: I couldn't find any better solution than using `exec` for this :sweat_smile: It was based on CAPI implementation for kubectlApply, so imo should be ok for us, though I don't particularly love it :sweat:

test/e2e/upgrade_test.go
```diff
@@ -247,6 +254,10 @@ var _ = Describe("BMO Upgrade", func() {
 				WaitIntervals:       e2eConfig.GetIntervals("default", "wait-deployment"),
 			})
 			Expect(err).NotTo(HaveOccurred())
+			DeferCleanup(func() {
+				By(fmt.Sprintf("Removing BMO from %s on the upgrade cluster", bmoKustomization))
+				BuildAndRemoveKustomize(ctx, bmoKustomization, upgradeClusterProxy)
```
lentzi90, line 259: This is a bit confusing and dangerous. The variable `bmoKustomization` is used both for the old and upgraded BMO kustomization so it is easy to make a mistake when reasoning about it. Could you change it so we have separate variables?

Also, it seems strange to delete the old kustomization after the upgrade. At that point it should be the upgraded BMO we delete. If there is an issue and the test fails early, we should anyway stop. I don't think there is much point trying to clean up at that point.

mquhuy: I think the point here is to clean up so the next test can start, so no matter at what stage the test stops, the cluster will be restored to its original state. If the upgrade was done, then we would delete the upgraded deployment, but even in case, for some reason, it was skipped, then we still cleanup the upgrade-from deployment. Of course, it would mean we may do a bit extra in our cleanup, but imo that's acceptable, and shouldn't cause any harm (similarly, we always clean up both `vbmc` and `sushy-tools` containers, even though only one of them should exist at one point).

In the latest patch I renamed the param `bmoKustomization` in one place as you suggested. I also changed all reference of `kustomize` in function names to `kustomization`, for consistency.

lentzi90: Thanks the kustomization functions and variables look much better now! I don't think the `defer` makes sense in tests though. We should use the `AfterEach` instead, that is what it is for and where we clean up everything else from a test. This way we also make sure that we can collect relevant logs before we remove the containers.

If you really want to do the double deletion of both the old and upgraded I can accept that, but please note that it clutters the logs with a lot of errors and the test should anyway not continue if there was an error. If the upgrade from 0.4 fails for example then we stop and collect all data we can. No more tests should run at that point so cleaning up or not does not matter.

mquhuy: Thank you! My intention is to add ironic Upgrade into the same suite, so imo full cleanup, even doubling ones, is necessary.

I don't understand, however, why `DeferCleanup` is not suitable? It's [listed as a way to clean up in Ginkgo docs](https://onsi.github.io/ginkgo/#spec-cleanup-aftereach-and-defercleanup), and compared to `AfterEach`, I think it is safer in this case: we only need to make sure to put one `DeferCleanup` after each installation, and we're fine. Otherwise, to put these cleanups into `AfterEach`, we would have to keep track of what was installed and what was not (for e.g. `ironic` is there, but it was preinstalled so shouldn't be cleaned, etc.)

I don't think missing logs is an issue either. `DeferCleanup` should be triggered only after the `It` block has ended. In our case it means the trigger phase will be similar to as if we put that cleanup in `AfterEach`. (In fact, [it runs AFTER `AfterEach`](https://github.com/onsi/ginkgo/issues/1022), so it should be safe even if we want to put any extra log collection in `AfterEach`)

lentzi90: Ah I learned something new. Thanks for that! I was not aware of how this DeferCleanup worked. This looks good to me now :)

lentzi90 (commented):
/lgtm

kashifest (commented):
/approve

PR #1556, opened 2024-02-14
🌱 E2E: Fix basic auth config

lentzi90:

The variables used for setting basic auth credentials were not exported, meaning that envsubst did not pick them up. This caused the auth config to have empty values.

Fixes #


tuminoid (commented):
/lgtm

maxrantil:
LGTM

mboukhalfa (approved):

mquhuy:
lgtm

kashifest:
> The variables used for setting basic auth credentials were not exported, meaning that envsubst did not pick them up. This caused the auth config to have empty values.

Shouldn't the test fail/complain in such case ?

lentzi90:
> Shouldn't the test fail/complain in such case ?

My understanding is that empty auth config for Ironic is the same as disabling basic auth. That does not make the tests fail since everything still works, it is just insecure. BMO was all the time configured with credentials but those were not needed since Ironic didn't ask for any.

PR #1557, opened 2024-02-14
🌱 E2E: Remove unused variables

lentzi90:

For legacy reasons, the configmaps used to configure BMO and Ironic were identical. The same configmap could then be used by both. This is not done in practice in the e2e tests so it does not make sense to keep BMO related variables in the configmap for Ironic and vice versa. This commit removes the irrelevant variables from the config files.

Fixes #


mquhuy:
LGTM. This is a nice change!

kashifest (commented):
/approve
Good clean up, less confusion

PR #1561, opened 2024-02-20
🌱 E2E: Run multiple BMO upgrade tests in sequence

mquhuy:
This PR will run multiple BMO upgrade tests from multiple BMO versions (kustomizations), in sequence. 

The log tree for the upgrade set was also modified to be more identifiable.
```
test/e2e/_artifacts/                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       
├── bmo-upgrade-from-e2e-release-0-3                                                                                                                                                                                                                                                                                                                                                                                                                                                                       
│   ├── logs                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               
│   │   ├── bmo-from-e2e-release-0-3                                                                                                                                                                                                                                                                                                                                                                                                                                                                       
│   │   │   └── baremetal-operator-controller-manager                                                                                                                                                                                                                                                                                                                                                                                                                                                      
│   │   │       ├── baremetal-operator-controller-manager-696c96dbdc-lr9vc                                                                                                                                                                                                                                                                                                                                                                                                                                 
│   │   │       │   ├── kube-rbac-proxy-log-metadata.json                                                                                                                                                                                                                                                                                                                                                                                                                                                  
│   │   │       │   ├── kube-rbac-proxy.log                                                                                                                                                                                                                                                                                                                                                                                                                                                                
│   │   │       │   ├── manager-log-metadata.json                                                                                                                                                                                                                                                                                                                                                                                                                                                          
│   │   │       │   └── manager.log                                                                                                                                                                                                                                                                                                                                                                                                                                                                        
│   │   │       ├── baremetal-operator-controller-manager-6d8c55778b-8lk2h                                                                                                                                                                                                                                                                                                                                                                                                                                 
│   │   │       │   ├── kube-rbac-proxy-log-metadata.json                                                                                                                                                                                                                                                                                                                                                                                                                                                  
│   │   │       │   ├── kube-rbac-proxy.log                                                                                                                                                                                                                                                                                                                                                                                                                                                                
│   │   │       │   ├── manager-log-metadata.json                                                                                                                                                                                                                                                                                                                                                                                                                                                          
│   │   │       │   └── manager.log                                                                                                                                                                                                                                                                                                                                                                                                                                                                        
│   │   │       ├── baremetal-operator-controller-manager-7dd8b7c4b4-slg8g                                                                                                                                                                                                                                                                                                                                                                                                                                 
│   │   │       │   ├── kube-rbac-proxy-log-metadata.json                                                                                                                                                                                                                                                                                                                                                                                                                                                  
│   │   │       │   ├── kube-rbac-proxy.log                                                                                                                                                                                                                                                                                                                                                                                                                                                                
│   │   │       │   ├── manager-log-metadata.json                                                                                                                                                                                                                                                                                                                                                                                                                                                          
│   │   │       │   └── manager.log                                                                                                                                                                                                                                                                                                                                                                                                                                                                        
│   │   │       ├── baremetal-operator-controller-manager-7dd8b7c4b4-t9g5l                                                                                                                                                                                                                                                                                                                                                                                                                                 
│   │   │       │   ├── kube-rbac-proxy-log-metadata.json                                                                                                                                                                                                                                                                                                                                                                                                                                                  
│   │   │       │   ├── kube-rbac-proxy.log                                                                                                                                                                                                                                                                                                                                                                                                                                                                
│   │   │       │   ├── manager-log-metadata.json                                                                                                                                                                                                                                                                                                                                                                                                                                                          
│   │   │       │   └── manager.log                                                                                                                                                                                                                                                                                                                                                                                                                                                                        
│   │   │       ├── baremetal-operator-controller-manager-7dd8b7c4b4-vxn6g                                                                                                                                                                                                                                                                                                                                                                                                                                 
│   │   │       │   ├── kube-rbac-proxy-log-metadata.json                                                                                                                                                                                                                                                                                                                                                                                                                                                  
│   │   │       │   ├── kube-rbac-proxy.log                                                                                                                                                                                                                                                                                                                                                                                                                                                                
│   │   │       │   ├── manager-log-metadata.json                                                                                                                                                                                                                                                                                                                                                                                                                                                          
│   │   │       │   └── manager.log                                                                                                                                                                                                                                                                                                                                                                                                                                                                        
│   │   │       └── baremetal-operator-controller-manager-896c6d5b9-78gtq                                                                                                                                                                                                                                                                                                                                                                                                                                  
│   │   │           ├── kube-rbac-proxy-log-metadata.json                                                                                                                                                                                                                                                                                                                                                                                                                                                  
│   │   │           ├── kube-rbac-proxy.log                                                                                                                                                                                                                                                                                                                                                                                                                                                                
│   │   │           ├── manager-log-metadata.json                                                                                                                                                                                                                                                                                                                                                                                                                                                          
│   │   │           └── manager.log                                                                                                                                                                                                                                                                                                                                                                                                                                                                        
│   │   ├── bmo-upgrade-main                                                                                                                                                                                                                                                                                                                                                                                                                                                                               
│   │   │   └── baremetal-operator-controller-manager                                                                                                                                                                                                                                                                                                                                                                                                                                                      
│   │   │       ├── baremetal-operator-controller-manager-696c96dbdc-lr9vc                                                                                                                                                                                                                                                                                                                                                                                                                                 
│   │   │       │   ├── kube-rbac-proxy-log-metadata.json                                                                                                                                                                                                                                                                                                                                                                                                                                                  
│   │   │       │   ├── kube-rbac-proxy.log                                                                                                                                                                                                                                                                                                                                                                                                                                                                
│   │   │       │   ├── manager-log-metadata.json                                                                                                                                                                                                                                                                                                                                                                                                                                                          
│   │   │       │   └── manager.log                                                                                                                                                                                                                                                                                                                                                                                                                                                                        
│   │   │       ├── baremetal-operator-controller-manager-6d8c55778b-8lk2h                                                                                                                                                                                                                                                                                                                                                                                                                                 
│   │   │       │   ├── kube-rbac-proxy-log-metadata.json                                                                                                                                                                                                                                                                                                                                                                                                                                                  
│   │   │       │   ├── kube-rbac-proxy.log                                                                                                                                                                                                                                                                                                                                                                                                                                                                
│   │   │       │   ├── manager-log-metadata.json                                                                                                                                                                                                                                                                                                                                                                                                                                                          
│   │   │       │   └── manager.log                                                                                                                                                                                                                                                                                                                                                                                                                                                                        
│   │   │       ├── baremetal-operator-controller-manager-7dd8b7c4b4-slg8g                                                                                                                                                                                                                                                                                                                                                                                                                                 
│   │   │       │   ├── kube-rbac-proxy-log-metadata.json                                                                                                                                                                                                                                                                                                                                                                                                                                                  
│   │   │       │   ├── kube-rbac-proxy.log                                                                                                                                                                                                                                                                                                                                                                                                                                                                
│   │   │       │   ├── manager-log-metadata.json                                                                                                                                                                                                                                                                                                                                                                                                                                                          
│   │   │       │   └── manager.log                                                                                                                                                                                                                                                                                                                                                                                                                                                                        
│   │   │       ├── baremetal-operator-controller-manager-7dd8b7c4b4-t9g5l                                                                                                                                                                                                                                                                             
│   │   │       │   ├── kube-rbac-proxy-log-metadata.json                                                                                                                                                                                                                                                                                              
│   │   │       │   ├── kube-rbac-proxy.log                                                                                                                                                                                                                                                                                                            
│   │   │       │   ├── manager-log-metadata.json                                                                                                                                                                                                                                                                                                      
│   │   │       │   └── manager.log                                                                                                                                                                                                                                                                                                                    
│   │   │       ├── baremetal-operator-controller-manager-7dd8b7c4b4-vxn6g                                                                                                                                                                                                                                                                             
│   │   │       │   ├── kube-rbac-proxy-log-metadata.json                                                                                                                                                                                                                                                                                              
│   │   │       │   ├── kube-rbac-proxy.log                                                                                                                                                                                                                                                                                                            
│   │   │       │   ├── manager-log-metadata.json                                                                                                                                                                                                                                                                                                      
│   │   │       │   └── manager.log                                                                                                                                                                                                                                                                                                                    
│   │   │       └── baremetal-operator-controller-manager-896c6d5b9-78gtq                                                                                                                                                                                                                                                                              
│   │   │           ├── kube-rbac-proxy-log-metadata.json                                                                                                                                                                                                                                                                                              
│   │   │           ├── kube-rbac-proxy.log                                                                                                                                                                                                                                                                                                            
│   │   │           ├── manager-log-metadata.json                                                                                                                                                                                                                                                                                                      
│   │   │           └── manager.log                                                                                                                                                                                                                                                                                                                    
│   │   └── ironic                                                                                                                                                                                                                                                                                                                                     
│   │       └── ironic                                                                                                                                                                                                                                                                                                                                 
│   │           ├── ironic-76997c6c6-cflm5                                                                                                                                                                                                                                                                                                             
│   │           │   ├── ironic-httpd-log-metadata.json                                                                                                                                                                                                                                                                                                 
│   │           │   ├── ironic-httpd.log                                                                                                                                                                                                                                                                                                               
│   │           │   ├── ironic-inspector-log-metadata.json                                                                                                                                                                                                                                                                                             
│   │           │   ├── ironic-inspector.log                                                                                                                                                                                                                                                                                                           
│   │           │   ├── ironic-log-metadata.json                                                                                                                                                                                                                                                                                                       
│   │           │   ├── ironic-log-watch-log-metadata.json                                                                                                                                                                                                                                                                                             
│   │           │   ├── ironic-log-watch.log                                                                                                                               
│   │           │   └── ironic.log                                                                                                                                         
│   │           ├── ironic-76997c6c6-s4qbk                                                                                                                                 
│   │           │   ├── ironic-httpd-log-metadata.json                                                                                                                     
│   │           │   ├── ironic-httpd.log                                                                                                                                   
│   │           │   ├── ironic-inspector-log-metadata.json                                                                                                                 
│   │           │   ├── ironic-inspector.log                                                                                                                               
│   │           │   ├── ironic-log-metadata.json                                                                                                                           
│   │           │   ├── ironic-log-watch-log-metadata.json                                                                                                                 
│   │           │   ├── ironic-log-watch.log                                                                                                                               
│   │           │   └── ironic.log                                                                                                                                         
│   │           └── ironic-76997c6c6-vv27n                                                                                                                                 
│   │               ├── ironic-httpd-log-metadata.json                                                                                                                     
│   │               ├── ironic-httpd.log                                                                                                                                   
│   │               ├── ironic-inspector-log-metadata.json                                                                                                                 
│   │               ├── ironic-inspector.log                                                                                                                               
│   │               ├── ironic-log-metadata.json                                                                                                                           
│   │               ├── ironic-log-watch-log-metadata.json                                                                                                                 
│   │               ├── ironic-log-watch.log                                                                                                                               
│   │               └── ironic.log                                                                                                                                         
│   └── resources                                                                                                                                                          
│       └── upgrade-v6bm0w                                                                                                                                                 
│           └── events.log                                                                                                                                                 
├── bmo-upgrade-from-e2e-release-0-4                                         
....
```


test/e2e/config/fixture.yaml
```diff
@@ -23,7 +23,9 @@ variables:
   UPGRADE_DEPLOY_IRONIC: "false"
   UPGRADE_DEPLOY_BMO: "true"
   UPGRADE_DEPLOY_CERT_MANAGER: "true"
-  UPGRADE_BMO_KUSTOMIZATION_FROM: "../../config/overlays/fixture-release-0.5"
+  UPGRADE_BMO_KUSTOMIZATION_FROM: "../../config/overlays/fixture-release-0.3 ../../config/overlays/fixture-release-0.4 ../../config/overlays/fixture-release-0.5"
```
lentzi90, line 26: I think it is time to turn this into proper structured config. It is becoming too messy. Could you take a look at how this looks for CAPI and try to do something similar?

test/e2e/upgrade_test.go
```diff
@@ -190,19 +193,26 @@ var _ = Describe("BMO Upgrade", func() {
 			}
 		} else {
 			By("Creating a separate cluster for upgrade tests")
+			upgradeClusterName = fmt.Sprintf("bmo-e2e-upgrade-%d", GinkgoParallelProcess())
```
lentzi90, line 196: So the clusters will all be named the same when running with one worker? :thinking: 
This does not feel right. I would prefer to name it something based on the version it is going to test.

test/e2e/upgrade_test.go
```diff
@@ -217,153 +227,150 @@ var _ = Describe("BMO Upgrade", func() {
...
-			err := BuildAndApplyKustomization(ctx, &BuildAndApplyKustomizationInput{
-				Kustomization:       ironicKustomization,
-				ClusterProxy:        upgradeClusterProxy,
-				WaitForDeployment:   true,
-				WatchDeploymentLogs: true,
-				DeploymentName:      "ironic",
-				DeploymentNamespace: bmoIronicNamespace,
-				LogPath:             filepath.Join(artifactFolder, "logs", fmt.Sprintf("%s-%s", bmoIronicNamespace, specName)),
-				WaitIntervals:       e2eConfig.GetIntervals("default", "wait-deployment"),
+	for i := range bmoFromKustomizations {
```
lentzi90, line 230: This is not a great pattern. We end up using indices all over the place and it becomes hard to understand and read the code.
Could you instead break out the test spec into a separate function that takes a struct with config as input? Then we just call that function for each test we want to run. ([Example](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/1cb5d5329f967e90ed891386f598cb266e8daa13/test/e2e/suites/e2e/clusterctl_upgrade_test.go#L33-L100))

test/e2e/upgrade_test.go
mquhuy: Hmm. May I ask why? I don't think the cluster name is that important, we only named it "bmo-e2e-upgrade" before. The cluster name is not logged anywhere else, and the clusters themselves will be cleaned afterwards. IMO the only reason it needs to be unique is so that the tests can be ran in parallel.

Making the name based on version requires somewhat hacky change. And the names will get messy real fast once we introduce ironic upgrade tests.

test/e2e/config/fixture.yaml
mquhuy: Yep. Good idea! I'll have a look

test/e2e/upgrade_test.go
mquhuy: Yes, thanks for the idea. I kinda like it to be in a loop though, but having a function is also nice. Extra useful since we will be able to use it for ironic upgrade as well.

test/e2e/upgrade_test.go
lentzi90: Ok, it was not as bad as I thought. I remembered how the cluster name is the folder where the logs are stored in CAPI/CAPM3 tests. It is fine like this :slightly_smiling_face:

test/e2e/upgrade_test.go
mquhuy: Thanks Lennart. And sorry for not getting back to you sooner.

I think using the same function for both of upgrade cases is possible, and maybe even preferable, as most of the steps are identical. That's up for discussion though, but in this PR I will focus on making the BMO upgrade first.

Regarding the loop, using DescribeTable is a good idea, and I think I will take that into use, but my reason for looping was that I don't want to pre-define the number of test cases, but rather based on the number of kustomizations provided by the user. Do you happen to have a good idea how to solve that, or maybe my expectation was just to unrealistic?

mquhuy: I think I've come up with a good solution. Could you please have a look?

test/e2e/upgrade_test.go
```diff
@@ -216,154 +230,164 @@ var _ = Describe("BMO Upgrade", Label("optional", "upgrade"), func() {
 			Expect(err).NotTo(HaveOccurred())
 		}
 	})
+	DescribeTable("Upgrade",
+		func(ctx context.Context, input *BMOIronicUpgradeInput) {
```
lentzi90, line 234: This function is huge. It would be nice to define it separately just for better readability. A proper description of what exactly it is supposed to do is also necessary.
I know I have linked many times to CAPI and I will do it again here. Please check [how they describe](https://github.com/kubernetes-sigs/cluster-api/blob/5245bac52be028369285a0ee04205d073a6b864f/test/e2e/clusterctl_upgrade.go#L151-L180) the upgrade spec. I want to see something similar here. It is especially important to describe how/if there are differences between BMO and Ironic since the same function is used for both!

lentzi90 (commented):
Great work!
I think this is almost ready to merge but would like to see some better documentation/comments first.

I also want to suggest that you consider if we should actually include Ironic at this point. Why not make this a BMO upgrade only? It is possible that we later discover some issue or want to do things in a different way for Ironic. Coupling them together in the same upgrade spec like this will make these things harder to fix later.

test/e2e/e2e_config.go
```diff
@@ -30,6 +30,15 @@ const (
...
 )
 
+type BMOIronicUpgradeInput struct {
+	DeployIronic               bool   `yaml:"deployIronic,omitempty"`
+	DeployBMO                  bool   `yaml:"deployBMO,omitempty"`
+	InitBMOKustomization       string `yaml:"initBMOKustomization,omitempty"`
+	InitIronicKustomization    string `yaml:"initIronicKustomization,omitempty"`
+	UpgradeEntityKustomization string `yaml:"upgradeEntityKustomization,omitempty"`
+	UpgradeEntityName          string `yaml:"upgradeEntityName,omitempty"`
+}
```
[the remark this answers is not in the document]
mquhuy: Thanks Lennart. Yes, I'll definitely add some docs to these input fields.

Regarding the BMO and Ironic separation, my plan is to add an Ironic upgrade test right after this PR is in. (Currently we can simply add the case configs to `ironic.yaml`, and the tests will run). I'd suggest that we have BMO and Ironic upgrades in the same function for now, we can always split them up later if there's different needs. Is it ok for you?

test/e2e/upgrade_test.go
mquhuy: Sure. I'll add some description! Thank you for pointing out the CAPI example!
I couldn't find any way to use an external func with `DescribeTable` (CAPI E2E has several `Describe` blocks calling the same function, which IMO is less nicer than `DescribeTable`). To be honest, I don't think that was an issue as it's currently also a function (without a name), but of course it would be nice. Would you have an example of how that could be done, which I could follow?

lentzi90: It should be as simple as

```go
RunUpgradeTest(ctx context.Context, input *BMOIronicUpgradeInput) {
// ...
}

DescribeTable("Upgrade", RunUpgradeTest, ...)
```

test/e2e/e2e_config.go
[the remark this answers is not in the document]
lentzi90: It is ok :+1:

test/e2e/upgrade_test.go
mquhuy: Yes, but some of variables used in that function (like the namespace or clusterProxy) are defined inside `BeforeEach`.

lentzi90: Oh! Then that is the issue! We should not rely on variables from the context "outside" the function in that way. Either add what is needed to the BMOIronicUpgradeInput or move them inside the function.

mquhuy: My bad, `namespace` is not defined in `BeforeEach`, but it and `cancelWatches` are defined inside the `RunUpgradeTest()`, and will be used later in `AfterEach()`.

mquhuy: Adding this cleanup() inside `AfterEach` is kinda compulsory, because it should happen before BMO cleanup. If we add a `DeferCleanup()` right after namespace creation, BMO cleanup will happen first, as BMO upgrade happens after namespace creation. Otoh, if we put the namespace cleanup in a `DeferCleanup()` after BMO upgrade, it won't get run if something breaks the test before it gets to BMO upgrade :-?

mquhuy: I may have found a solution. I will try to implement it.

lentzi90 (approved):
I'm happy with this now!
/approve

ironic-deployment/overlays/e2e-with-inspector/kustomization.yaml
tuminoid, line 1: Formatting is not consistent in this file. Half of lists are indented and half are not. Use no-indentation for lists k8s formatting.

tuminoid (commented):
Nit.

test/e2e/upgrade_test.go
mquhuy: Function separated to its own place

test/e2e/e2e_config.go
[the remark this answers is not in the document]
mquhuy: Docs added

ironic-deployment/overlays/e2e-with-inspector/ironic-patch.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: ironic
+spec:
+  template:
+    spec:
+      containers:
+        - name: ironic-dnsmasq
```
tuminoid, line 9: Sorry, missed this last time. This needs to be indented properly as well.

mquhuy: Done

PR #1562, opened 2024-02-21
🌱 release-0.5: harden stable branch against inspector removal

dtantsur:
- Use ironic-image release-24.0 which still has Inspector support
- Explicitly enable Inspector in all environment files


lentzi90 (commented):
/lgtm
I think it makes sense to pin ironic like this

PR #1566, opened 2024-02-28
🌱 Use real Context and make provisioners context-aware

dtantsur:
There is no compelling reason to use TODO/Background: we have context available in reconcilers.

GopherCloud 2.0.0-beta2 will require context in its calls, so also pass the context to the provisioner.


kashifest (commented):
/approve

PR #1567, opened 2024-02-28
🌱 Drop dependency on x/exp/slices

dtantsur:
Go 1.21 already has what we need in the standard library.


lentzi90:
I cannot reproduce the test failure locally. My best guess is something was not properly cleaned up between runs.

lentzi90 (commented):
/lgtm

go.mod
```diff
@@ -60,6 +59,7 @@ require (
 	github.com/spf13/cobra v1.7.0 // indirect
 	github.com/spf13/pflag v1.0.5 // indirect
 	go.uber.org/multierr v1.11.0 // indirect
+	golang.org/x/exp v0.0.0-20230522175609-2e198f4a06a1 // indirect
```
kashifest, line 62: Which module is still dependant on this?

dtantsur: I'm not sure how to figure that out. Nothing in our code, I believe.

kashifest: `go mod graph | grep "golang.org/x/exp"` should give us the module which is using this.

kashifest: ok its coming from 

`k8s.io/apiextensions-apiserver@v0.28.3` and 
`sigs.k8s.io/controller-runtime@v0.16.3`, so we cant really get rid of this entry.

/approve

dtantsur: TIL `go mod graph`!

PR #1569, opened 2024-02-29
⚠️ Update to gophercloud 2.0.0-beta2

dtantsur:
Although the current version is specified as 1.5.1, it's actually
a commit somewhere in the 2.0 branch. This change brings us to
a more specific version.

2.0.0-beta2 introduced a requirement on Context for all calls.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,11 +1,11 @@
 module github.com/metal3-io/baremetal-operator
 
-go 1.21
+go 1.21.6
 
 require (
 	github.com/go-logr/logr v1.4.1
 	github.com/google/safetext v0.0.0-20230106111101-7156a760e523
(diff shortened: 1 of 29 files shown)


go.mod
```diff
@@ -1,11 +1,11 @@
 module github.com/metal3-io/baremetal-operator
 
-go 1.21
+go 1.21.6
```
kashifest, line 3: Any specific reason to pin it to patch version?

dtantsur: This is apparently the new way Go does it starting with 1.21. I haven't done this change - `go mod tidy` does it (based on Gophercloud's go.mod, I believe).

kashifest: Interesting , I didnt face it when I bumped it to go 1.21, I have to check it again.

kashifest (commented):
/approve

PR #1585, opened 2024-03-04
🌱 Bump go to v1.21.7

Sunnatillo:
Golang 1.20 is out of support in Feb 2024. Bumping golang image only in dockerfile, but not go.mod because it will break library consumers.


tuminoid (commented):
/lgtm

kashifest (commented):
/approve

PR #1602, opened 2024-03-06
🌱 deploy.sh to go code to fix gosec related issues

maxrantil:
This PR transitions the functionality of the `deploy.sh` script into Go, aiming to fix gosec related issues. ~~There is still one exception related to gosec that remains within the `pipeCommands` function:~~

```
OLD
```

~~A env var, `KUBECTL_ARGS`, introduces a potential security risk. The `validateCmd` function is currently the best approach to ador the option to choose kubeconfigdressing this issue~~


tuminoid:
> A env var, KUBECTL_ARGS, introduces a potential security risk. The validateCmd function is currently the best approach to addressing this

What kind of legit arguments is passed there?  Something like `--foo --bar=3` ? If so, then you should be validating versus legit patterns like `--\w` and `--\w=\w`. It is very hard to try look for unsafe characters, there are so many ways around it, via UTF etc.

maxrantil:
This is an example from `metal3-dev-env/tests/roles/run_tests/vars/main.yml`:
`KUBECTL_ARGS: "--kubeconfig=/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"`

Also in `baremetal-operator/docs/deploying.md` is says:
`- KUBECTL_ARGS : Additional arguments to kubectl apply`

From my testing (setting up with dev-env) there has been nothing set to `KUBECTL_ARGS`

tuminoid:
> This is an example from `metal3-dev-env/tests/roles/run_tests/vars/main.yml`: `KUBECTL_ARGS: "--kubeconfig=/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"`
> 
> Also in `baremetal-operator/docs/deploying.md` is says: `- KUBECTL_ARGS : Additional arguments to kubectl apply`
> 
> From my testing (setting up with dev-env) there has been nothing set to `KUBECTL_ARGS`

Then definitely it is worth doing the regex validation the way I suggested. There aren't many legit ways to format legit params.

dtantsur:
Do you want to remove the old deploy.sh to avoid confusion?

tools/deploy.go
```diff
@@ -0,0 +1,591 @@
...
+
+	// This is to keep the current behavior of using the ironic.env file for the configmap
+	ironicEnvSrc := filepath.Join(scriptDir, "config", "default", "ironic.env")
+	ironicEnvDst := filepath.Join(tempBMOOverlay, "ironic.env")
+	copyFile(ironicEnvSrc, ironicEnvDst)
+
+	execCommand(kustomizePath, "edit", "add", "configmap", "ironic", "--behavior=create", "--from-env-file=ironic.env")
+
+	kubectlBaseCmd := []string{"kubectl", "apply"}
+	kubectlBaseCmd = prepareKubectlCmdArgs(kubectlBaseCmd, kubectlArgs)
```
adilGhaffarDev, line 334: Can we use k8s client go libraries for kubectl related commands?

kashifest:
> Do you want to remove the old deploy.sh to avoid confusion?

Thats the ultimate goal, for the time being they can coexist until we test the go code properly.

tools/deploy.go
maxrantil: Thanks for the feedback, I'll look into it!

maxrantil:
I'm currently updating the code to incorporate the k8s go-client. However, I'm unsure about the necessary options for `KUBECTL_ARGS`. Is specifying `KUBECTL_ARGS="--kubeconfig=/path/to/custom-kubeconfig.yaml"` sufficient, or should we support additional options? For example, should we include options like `KUBECTL_ARGS="--namespace=my-namespace"` and more? It is not that straight forward to add every possible option with k8s go-client.

kashifest:
> I'm currently updating the code to incorporate the k8s go-client. However, I'm unsure about the necessary options for `KUBECTL_ARGS`. Is specifying `KUBECTL_ARGS="--kubeconfig=/path/to/custom-kubeconfig.yaml"` sufficient, or should we support additional options? For example, should we include options like `KUBECTL_ARGS="--namespace=my-namespace"` and more? It is not that straight forward to add every possible option with k8s go-client.

Can you elaborate a bit more what do you mean my specifying options? Would you push the code so that we can check what is it that we may need?

tools/deploy.go
```diff
@@ -0,0 +1,711 @@
...
+
+// getKubeconfigPath determines the kubeconfig path from KUBECTL_ARGS, KUBECONFIG_PATH, or defaults to ~/.kube/config.
+func getKubeconfigPath() string {
+	kubectlArgs := os.Getenv("KUBECTL_ARGS")
+	kubeconfigPath := ""
+
+	kubeconfigPrefix := "--kubeconfig="
+	if strings.Contains(kubectlArgs, kubeconfigPrefix) {
+		if strings.Contains(kubectlArgs, " ") {
+			log.Fatalf("Error: $KUBECTL_ARGS accepts only one argument.")
```
tuminoid, line 283:
```suggestion
			log.Fatalf("Error: KUBECTL_ARGS accepts only one argument.")
```

tools/deploy.go
```diff
@@ -0,0 +1,711 @@
...
+		}
+		kubeconfigPath = strings.TrimPrefix(kubectlArgs, kubeconfigPrefix)
+	} else if kubectlArgs != "" {
+		log.Fatalf("Error: Invalid format in $KUBECTL_ARGS. Expected format: '--kubeconfig=/path/to/kubeconfig'.")
```
tuminoid, line 287:
```suggestion
		log.Fatalf("Error: Invalid format in KUBECTL_ARGS. Expected format: '--kubeconfig=/path/to/kubeconfig'.")
```

tools/deploy.go
```diff
@@ -0,0 +1,711 @@
...
+}
+
+// getKubeconfigPath determines the kubeconfig path from KUBECTL_ARGS, KUBECONFIG_PATH, or defaults to ~/.kube/config.
+func getKubeconfigPath() string {
+	kubectlArgs := os.Getenv("KUBECTL_ARGS")
+	kubeconfigPath := ""
+
+	kubeconfigPrefix := "--kubeconfig="
+	if strings.Contains(kubectlArgs, kubeconfigPrefix) {
+		if strings.Contains(kubectlArgs, " ") {
```
tuminoid, line 282: I'd like to see whitelisting rather than blacklisting.  For example, check it matches `--kubeconfig=[a-zA-Z/.]+` and then verify its a file and its readable.

tools/deploy.go
```diff
@@ -0,0 +1,711 @@
...
+
+	err = applyYAML(yamlOutputPath)
+	if err != nil {
+		return fmt.Errorf("failed to apply YAML: %v", err)
+	}
+
+	return nil
+}
+
+func copyFile(src, dst string) error {
```
tuminoid, line 432: What is different from `os.Copy()`?

tools/deploy.go
```diff
@@ -0,0 +1,711 @@
...
+	return nil
+}
+
+func replaceAllPlaceholdersInFile(filePath, placeholder, newValue string) error {
+	input, err := os.ReadFile(filePath)
+	if err != nil {
+		return fmt.Errorf("failed to read file %s: %v", filePath, err)
+	}
+
+	output := strings.ReplaceAll(string(input), placeholder, newValue)
```
tuminoid, line 474: So if the file has `PLACEHOLDER` and `PLACEHOLDER_ARGS` and you replace `PLACEHOLDER` with `foo` then you end up with `foo` and `foo_ARGS`, and then you cannot replace `PLACEHOLDER_ARGS` later...

tools/deploy.go
```diff
@@ -0,0 +1,711 @@
...
+		return fmt.Errorf("failed to write file %s: %v", filePath, err)
+	}
+
+	return nil
+}
+
+func deployIronic(tempIronicOverlay, scriptDir, kustomizePath string, deployKeepalivedFlag bool) error {
+	revertFunc, err := changeDirTemporarily(tempIronicOverlay)
+	if err != nil {
+		log.Fatalf("Failed to change directory %s: %v", tempIronicOverlay, err)
```
tuminoid, line 486: This falls thru, do we want to continue if we cannot change dirs?

tools/deploy.go
```diff
@@ -0,0 +1,711 @@
...
+	ironicBMOConfigmap := filepath.Join(tempIronicOverlay, "ironic_bmo_configmap.env")
+	copyFile(ironicBMOConfigmapSource, ironicBMOConfigmap)
+
+	restartContainerCertificateUpdated := GetEnvOrDefault("RESTART_CONTAINER_CERTIFICATE_UPDATED", "false")
+	err = updateOrAppendKeyValueInFile(ironicBMOConfigmap, "RESTART_CONTAINER_CERTIFICATE_UPDATED", restartContainerCertificateUpdated)
+	if err != nil {
+		return err
+	}
+
+	ironicHostIP := GetEnvOrDefault("IRONIC_HOST_IP", "")
```
tuminoid, line 507: Is empty string legit value for `IRONIC_HOST_IP` ? If not, then error out.

tools/deploy.go
```diff
@@ -0,0 +1,711 @@
...
+
+	if !deployBMOFlag && !deployIronicFlag {
+		fmt.Println("ERROR: At least one of -b (BMO) or -i (Ironic) must be specified for deployment.")
+		usage()
+		os.Exit(1)
+	}
+
+	exePath, err := os.Executable()
+	if err != nil {
+		log.Fatalf("Failed to determine executable path: %v", err)
```
tuminoid, line 612: This falls thru again after logging a fatal.

tools/deploy.go
```diff
@@ -0,0 +1,711 @@
...
+	scriptDir := filepath.Join(filepath.Dir(exePath), "..")
+
+	ironicBasicAuthComponent := filepath.Join(scriptDir, "ironic-deployment", "components", "basic-auth")
+	tempBMOOverlay := filepath.Join(scriptDir, "config", "overlays", "temp")
+	tempIronicOverlay := filepath.Join(scriptDir, "ironic-deployment", "overlays", "temp")
+
+	// Cleaning up directories before making new ones
+	err = os.RemoveAll(tempBMOOverlay)
+	if err != nil {
+		log.Fatalf("Failed to remove temp BMO overlay: %v", err)
```
tuminoid, line 623: Same here and below.

tools/deploy.go
```diff
@@ -0,0 +1,711 @@
...
+
+	// Generate credentials as needed
+	ironicDataDir := GetEnvOrDefault("IRONIC_DATA_DIR", "/opt/metal3/ironic/")
+	ironicAuthDir := filepath.Join(ironicDataDir, "auth")
+
+	// If usernames and passwords are unset, read them from file or generate them
+	if deployBasicAuthFlag {
+		ironicUsername, err := getEnvOrFile("IRONIC_USERNAME", filepath.Join(ironicAuthDir, "ironic-username"), 12)
+		if err != nil {
+			log.Fatalf("Error retrieving Ironic username: %v", err)
```
tuminoid, line 650: And here everywhere.

tuminoid:
Please check if you can use built-ins or de facto libraries instead of DIY. Also, logging a fatal and continuing is weird. If you can fall thru, then its not a fatal but a warning, or you error out if its really fatal.

maxrantil:
Thanks @tuminoid  for the review! Will look into these comments today

tools/deploy.go
tuminoid: Never mind, Fatalf exits too.

tools/deploy.go
```diff
@@ -0,0 +1,711 @@
...
+
+	err = writeLines(updatedLines, filePath)
+	if err != nil {
+		return fmt.Errorf("failed to write updated lines to file %s: %v", filePath, err)
+	}
+
+	return nil
+}
+
+func splitYAMLDocuments(yamlContent []byte) [][]byte {
```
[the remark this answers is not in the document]
maxrantil: I did change it now, thanks!

tools/deploy.go
```diff
@@ -0,0 +1,711 @@
...
+		execCommand(kustomizePath, "edit", "add", "secret", "ironic-credentials", "--from-file=username=ironic-username", "--from-file=password=ironic-password")
+	}
+
+	if deployTLSFlag {
+		execCommand(kustomizePath, "edit", "add", "component", "../../components/tls")
+	}
+}
+
+// helper function to read lines from a file.
+func readLines(filePath string) ([]string, error) {
```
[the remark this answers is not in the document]
maxrantil: I changed to more lib functions now, thanks for the feedback!

config/README.md
```diff
@@ -1,7 +1,7 @@
 # Kustomizations for Baremetal Operator
 
 This folder contains kustomizations for the Baremetal Operator. They have
-traditionally been used through the [deploy.sh](../tools/deploy.sh) script,
+traditionally been used through the [deploy.go](../tools/deploy.go) program,
```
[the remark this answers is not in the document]
maxrantil: :) removed

tools/deploy.go
maxrantil: This code is now gone. I am using `sed` with exec.Command instead.

tools/deploy.sh
```diff
@@ -97,6 +97,7 @@ fi
 
 MARIADB_HOST_IP="${MARIADB_HOST_IP:-"127.0.0.1"}"
 KUBECTL_ARGS="${KUBECTL_ARGS:-""}"
+echo "KUBECTL_ARGS HERE: ${KUBECTL_ARGS}"
```
tuminoid, line 100: This should be removed.

tools/deploy.go
tuminoid: I don't still see why you're not using `os.Copy()`?

tools/deploy.go
tuminoid, line 1: Many of the functions miss documentation.

tools/deploy.go
```diff
@@ -0,0 +1,684 @@
...
+		if ok && exiterr.ExitCode() == 1 {
+			err = appendToFile(ironicBMOConfigmap, fmt.Sprintf("RESTART_CONTAINER_CERTIFICATE_UPDATED=%s\n", restartContainerCertificateUpdated))
+			if err != nil {
+				return err
+			}
+		} else {
+			return fmt.Errorf("error checking for RESTART_CONTAINER_CERTIFICATE_UPDATED: %v", err)
+		}
+	} else {
+		execCommand("sh", "-c", fmt.Sprintf("sed -i 's/\\(RESTART_CONTAINER_CERTIFICATE_UPDATED\\).*/\\1=%s/' '%s'", restartContainerCertificateUpdated, ironicBMOConfigmap))
```
tuminoid, line 478: Not a fan of running external sed for changing a file content. Did not mean to go from a bit flawed `ReplaceAll()` implementation to running commands, especially when these have the exact same issue.

Please revert back to a function that reads in the file, does the replacement (a bit more carefully), and then writes it back.

tuminoid (commented):
Much better, but still needs some fixes.

tools/deploy.go
maxrantil: I need file descriptors for `io.Copy()`. I now use io.Copy() in my `copyFile` function.

tools/deploy.go
maxrantil: Will work on this next. I thought that some of them were self-explanatory, but will make comments anyway then.

tools/deploy.go
maxrantil: I now changed it to use `strings.Split` to get the exact match in the replacement. I think this will solve the issue you noticed, right?

tools/deploy.go
tuminoid: golint should complain about that, maybe Kashif hasn't enabled it yet due lot of missing documentation.

tuminoid: Thanks for adding them.

lentzi90 (commented):
Added a few comments below.
In general, this feels too much like a "translation" of the deploy script instead of an adaption to go. Since we are not directly trying to replace deploy.sh (it will be around for some time until we have tested this enough), there is no need to keep the exact same behavior either. We can, and should, do things better where possible so we don't end up re-implementing "bad practices" just to keep things similar.

Let me give a few examples and ideas around what I mean.

1. The deploy script is directly tied to the repository. It interacts with files in the repo and cannot be run without having the repo available. This is quite common in scripts, but not so much in go binaries. Think about how `clusterctl` works for example, or `cmctl` for cert-manager. I think we can improve in this area. For example, get the necessary kustomizations from github directly  by default and let the user specify the folder if they want that instead (and have the repo locally).
2. Can we use libraries instead of calling `htpasswd` and similar?
3. Find common tasks and implement functions so that they can be reused in multiple places. We already have a need to build kustomizations and apply them in the e2e tests. Find the common ground and make use of it!
4. The script makes use of `kustomize edit` to create a kustomization, but this may not be the best way to do it in go. I think we should at least consider just writing the kustomize file directly. We can use go templates and snippets of plain text that we write out to create the final file.

If I take all of this and try to envision what the go tool would look like, I get something like this made up CLI:

```console
$ bmoctl --help
The Bare Metal Operator CLI is used to generate configuration and kustomizations for Ironic and Bare Metal Operator and to install them in a Kubernetes cluster.

Usage:
  bmoctl [command]

Commands:
  generate    Generate kustomizations for BMO and/or Ironic (sub commands bmo and ironic)
              The kustomizations are written to a folder where they can be modified further or stored for later.
  init        Generate and apply kustomizations to initialize BMO and Ironic

Flags: basic auth, tls, etc...
```

Don't try to implement this. It is just a vision to give you an idea about what is possible. Keep it in mind and think about how we can reasonably improve on the deploy.sh script when re-doing it in go. :slightly_smiling_face:

maxrantil:
Thanks @lentzi90, this is great stuff! I will look into it today.

tools/deploy.go
```diff
@@ -0,0 +1,736 @@
...
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/restmapper"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+// GetEnvOrDefault returns the value of the environment variable key if it exists
+// and is non-empty. Otherwise it returns the provided default value.
+func GetEnvOrDefault(key, defaultValue string) string {
```
mquhuy, line 32: I'd love if we could get rid of using env vars, or at least have an option to pass a config file directly to this script. For traditional reason we've always been tied up to env vars, and they have been very difficult to track, especially for newbies.

Of course changing it here doesn't mean we could get rid of the env vars, but it could be the the first step. In dev-env, for now we can add an ansible job to generate the config file based on the env vars we have, before running this script. For general users, it'd make more senses to ask them creating their config file than to populate all the env vars.

maxrantil: Good idea, I will see if I find time to look into this.
@lentzi90 do you agree on doing this?

tools/deploy.go
```diff
@@ -0,0 +1,736 @@
...
+	_, err := rand.Read(b)
+	if err != nil {
+		return "", fmt.Errorf("failed to generate random string: %v", err)
+	}
+
+	return base64.RawURLEncoding.EncodeToString(b)[:length], nil
+}
+
+// readOrCreateFile reads the content of a file, or creates it with random content if it doesn't exist.
+func readOrCreateFile(path string, length int) (string, error) {
```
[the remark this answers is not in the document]
maxrantil: I added your suggested flow and logging - so now the user will know what was used.

tools/deploy.go
mquhuy: Thanks Max. For the record, it's pretty fine for me if you don't implement this in the current PR. It looks like an improvement that deserves its own scope and user story.

maxrantil:
/ test-centos-e2e-integration-main

lentzi90:
Closing this in favor of https://github.com/metal3-io/baremetal-operator/pull/1669
/close

PR #1603, opened 2024-03-06
🌱 Stop setting IRONIC_FAST_TRACK by default

dtantsur:
It's true by default and most operators will not want to change it.
Definitely not without understanding what it does.


kashifest:
how about removing them from run_local_ironic.sh script for ubuntu based tests https://github.com/metal3-io/baremetal-operator/blob/main/tools/run_local_ironic.sh#L84 and https://github.com/metal3-io/baremetal-operator/blob/main/tools/run_local_ironic.sh#L104

dtantsur:
@kashifest I thought someone might find an ability to override this variable useful still (this script is targeting developers after all, not just normal admins). If you don't think so, I can remove it.

lentzi90 (commented):
/approve

kashifest:
> @kashifest I thought someone might find an ability to override this variable useful still (this script is targeting developers after all, not just normal admins). If you don't think so, I can remove it.

Good point, I am just thinking that this script is also used in our ubuntu based CI tests, and if it would create confusion / different results in the test, however as long as it defaults it to true the tests should be identical if I am understanding correctly. I am fine either way.

Lets have others opinions  as well cc @lentzi90 @Rozzii

lentzi90:
It will be identical as you say. I think it is good to have the option to override in the scripts, but there is not much point setting this variable explicitly in the tests.

kashifest:
One more thing that comes into my mind, we have one more place where we explicitly set it to true https://github.com/metal3-io/metal3-dev-env/blob/main/03_launch_mgmt_cluster.sh#L159 irrespective of if it ironic is run as a deployment or run locally

dtantsur:
@kashifest created https://github.com/metal3-io/metal3-dev-env/pull/1360

PR #1609, opened 2024-03-08
🌱 bump shellcheck to v0.10.0

tuminoid:
Shellcheck had its first release since 2022, so bump it.

/hold
needs project-infra change too.


lentzi90 (commented):
/lgtm

PR #1622, opened 2024-03-14
🌱 Improve deploy.sh to differentiate between build and run of kustomize

CJCShadowsan:

Currently deploy.sh tries to use ${SCRIPTDIR} in its KUSTOMIZE variable which breaks calling it via the Makefile to build kustomize in some cases (it certainly does on my system!). To fix, I added another variable, KUSTOMIZE_BUILD which points to exactly what the Makefile wants to see.

Fixes #

### DIFF
--- a/tools/deploy.sh
+++ b/tools/deploy.sh
@@ -110,7 +110,8 @@ mkdir -p "${TEMP_BMO_OVERLAY}"
 mkdir -p "${TEMP_IRONIC_OVERLAY}"
 
 KUSTOMIZE="${SCRIPTDIR}/tools/bin/kustomize"
-make -C "$(dirname "$0")/.." "${KUSTOMIZE}"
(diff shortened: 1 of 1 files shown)


tools/deploy.sh
```diff
@@ -110,7 +110,8 @@ mkdir -p "${TEMP_BMO_OVERLAY}"
 mkdir -p "${TEMP_IRONIC_OVERLAY}"
 
 KUSTOMIZE="${SCRIPTDIR}/tools/bin/kustomize"
```
elfosardo, line 112: maybe can be done in a follow up but if you invert the lines between KUSTOMIZE and KUSTOMIZE_BUILD you can reuse the latter to build the former

elfosardo:
/approve
not sure I can approve here :/

kashifest (commented):
/approve

PR #1594, opened 2024-03-05
:sparkles: Attach non-bootable iso to BareMetalHost

hroyrh:
**What this PR does / why we need it**:
Add new `DataImage` CRD which allows attachment/detachment of non-bootable iso to BareMetalHost, based on [this proposal](https://github.com/metal3-io/metal3-docs/blob/072349dac347b9ae39c94f1aaf2ec750db642a0a/design/baremetal-operator/bmh_non-bootable_iso.md).

PR #1623, opened 2024-03-15
:sparkles: Add nonbootable iso api

hroyrh:
This PR defines the API for attaching and detaching non-bootable iso image. The Controller implementation for the API is in progress in this pr : https://github.com/metal3-io/baremetal-operator/pull/1594


apis/metal3.io/v1alpha1/dataimage_types.go
```diff
@@ -0,0 +1,82 @@
...
+
+// Contains the count of errors and the last error message.
+type DataImageError struct {
+	Count   int    `json:"count"`
+	Message string `json:"message"`
+}
+
+// DataImageSpec defines the desired state of DataImage.
+type DataImageSpec struct {
+	// Important: Run "make" to regenerate code after modifying this file
```
dtantsur, line 38: nit: please remove

apis/metal3.io/v1alpha1/dataimage_types.go
```diff
@@ -0,0 +1,82 @@
...
+
+	// AttachedImage shows the currently attached image
+	AttachedImage *AttachedImageReference `json:"attachedImage,omitempty"`
+
+	// Error contains errors related to DataImage attachment/detachment
+	Error *DataImageError `json:"error,omitempty"`
+}
+
+//+kubebuilder:object:root=true
+//+kubebuilder:resource:shortName=dimg
```
dtantsur, line 59: It's going to be a very rarely used resource, let's maybe not have a short name for it?

dtantsur:
The previous PR modifies a few more API-related files, such as config/base/crds/kustomization.yaml, a sample or RBAC roles. Could you add them too?

hroyrh:
@dtantsur I have updated the base kustomization config, and added cainjection, webhook and sample dataimage resource file. I've left the RBAC configs since they depend on the controller and it will be more appropriate to include them in the other pr.

PR #1624, opened 2024-03-18
⚠️ Remove ibmc and idrac drivers

dtantsur:
Both drivers (in case of idrac - its wsman implementation) has been
unmaintained for years, same for their underlying libraries ibmcclient [1]
and dracclient [2]. Ironic has deprecated them in the 2024.1 cycle.

Dell users should use idrac-redfish:// and idrac-virtualmedia:// instead.
iBMC users should try using the generic Redfish drivers and report any
bugs to the Ironic team and any standard compatibility problems directly
to Huawei.

[1] https://github.com/IamFive/python-ibmcclient
[2] https://opendev.org/openstack/python-dracclient


honza:
Do we need so kind of deprecation notice?

/approve

tuminoid:
> Do we need some kind of deprecation notice?

Wouldn't any deprecation notice needed to be in previous release already?

dtantsur:
I hoped that my ML thread would serve as a deprecation notice. Given the state of the drivers, I did not want to make the process too long.

Issue #1626, opened 2024-03-20
Split BMO e2e tests into separate spec and test files


lentzi90:
This is a bit more advanced. If you are new to go and/or ginkgo it may not be the best to start with.

The e2e tests are currently written as plain tests, each test in its own file. To make the structure clearer and the code reusable, we should split them up so that we have separate (non-test) files with procedure implementations. Each test can then call an implementation with a specific configuration.
For example, we could have one implementation of an upgrade test. This could then be called twice to test upgrades from both v0.4 and v0.5 to main.

**What you need to do**

1. Pick a test from [here](https://github.com/metal3-io/baremetal-operator/tree/main/test/e2e)
2. Identify the relevant input and what the test does
3. Implement the test separately and create an input struct with all relevant input
4. Change the original test to call the new implementation

**Tasks**

- [ ] [Basic ops](https://github.com/metal3-io/baremetal-operator/blob/main/test/e2e/basic_ops_test.go)
- [ ] [External inspection](https://github.com/metal3-io/baremetal-operator/blob/main/test/e2e/external_inspection_test.go)
- [ ] [Inspection](https://github.com/metal3-io/baremetal-operator/blob/main/test/e2e/inspection_test.go)
- [ ] [Provision](https://github.com/metal3-io/baremetal-operator/blob/main/test/e2e/provisioning_and_annotation_test.go) (I think the [live-iso test](https://github.com/metal3-io/baremetal-operator/blob/main/test/e2e/live_iso_test.go) could be another call to this with different config.)
- [ ] [Re-inspection](https://github.com/metal3-io/baremetal-operator/blob/main/test/e2e/re_inspection_test.go)
- [ ] [Upgrade](https://github.com/metal3-io/baremetal-operator/blob/main/test/e2e/upgrade_test.go)

jackjii79:
Hi @lentzi90 
I made some refactor against the `upgrade_test`, but i m not sure if my changes would align with the expectation consider this is my first PR in this repo, could you please take a preview about https://github.com/metal3-io/baremetal-operator/pull/1637 ?

PR #1637, opened 2024-03-27
🌱 Separate upgrade plan from upgrade test

jackjii79:

Generalize upgrade test procedure by extracting common procedure as independent flow to be reusable for more complex upgrading test scenarios.
 
Fixes https://github.com/metal3-io/baremetal-operator/issues/1626


jackjii79:
> like this



> Thanks for working on this!
> 
> I see many great things here, but I do think it goes a bit too far in what can be changed in the test. If I understand this correctly, it would be possible to even swap out the BMOUpgradePlanner to some custom implementation that could do the upgrade in a completely different way. What I would like to see is that this part (what BMOUpgradePlanner does) would be the test spec itself, similar to [this in CAPI](https://github.com/kubernetes-sigs/cluster-api/blob/3a16912194a2cdac514ae753a57737ef589541b6/test/e2e/clusterctl_upgrade.go#L180-L680). It would then be called [like this](https://github.com/kubernetes-sigs/cluster-api/blob/3a16912194a2cdac514ae753a57737ef589541b6/test/e2e/clusterctl_upgrade_test.go#L40-L111).
> 
> There seems to be some issue with the code currently. Could you take a look at why the E2E fixture test failed and try to fix that?

thanks @lentzi90 to the hint, i get a much clearer picture now, will make some updates

jackjii79:
@lentzi90 refactor is ready to review, however i have some problem setting up the minikube with kvm driver in my local environment to test, so i do not have a chance to run the e2e in my local to give it a shot in advance. The `make test-e2e` is also not easy to use in my local as some testing logic is trying to post requests to the service and i do not know which service/port might need port forward before even the e2e started.

jackjii79:
@lentzi90 Should we proceed with this PR ?

PR #1641, opened 2024-03-28
🌱 check for skipCleanup before running cleanup of test resources

### LINKED ISSUE #1629: Skip cleanup after each e2e test when SKIP_RESOURCE_CLEANUP is set
Currently we [skip cleanup](https://github.com/metal3-io/baremetal-operator/blob/6c823f4f8e0eb917bc7607c0acef055c4b070a94/test/e2e/e2e_suite_test.go#L190) of the management cluster if the variable is set but each individual test still does a cleanup.
Make all tests respect the variable so that no cleanup is done when it is set to true.

For each test [here](https://github.com/metal3-io/baremetal-operator/tree/main/test/e2e), look for the [cleanup](https://github.com/metal3-io/baremetal-operator/blob/17a286385a191563cb8368ca1e4c827d4d17c35c/test/e2e/basic_ops_test.go#L113) and make it so that it is skipped if asked for.

NeerajNagure:
Checks whether the skipCleanup variable is true or not.If true,it skips the cleanup of the test resources

Fixes #1629

### DIFF
--- a/test/e2e/basic_ops_test.go
+++ b/test/e2e/basic_ops_test.go
@@ -110,6 +110,8 @@ var _ = Describe("basic", Label("required", "basic"), func() {
 	})
 
 	AfterEach(func() {
-		cleanup(ctx, clusterProxy, namespace, cancelWatches, e2eConfig.GetIntervals("default", "wait-namespace-deleted")...)
+		if !skipCleanup {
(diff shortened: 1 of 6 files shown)


lentzi90 (commented):
Thank you for the PR!
/approve

tuminoid (commented):
/lgtm

Thanks for the contribution!

PR #1659, opened 2024-04-08
🌱 Disable name suffix hash

sathieu:

We need deterministic resource names to ease diffing between versions.

Also, we plan to install bmo using Helm, and test that the Helm chart produces the same output as the reference implementation (which uses kustomize).

### DIFF
--- a/ironic-deployment/components/keepalived/kustomization.yaml
+++ b/ironic-deployment/components/keepalived/kustomization.yaml
@@ -4,6 +4,9 @@ kind: Component
 patches:
 - path: keepalived_patch.yaml
 
+generatorOptions:
+  disableNameSuffixHash: true
+
 configMapGenerator:
 - envs:
   - ironic_bmo_configmap.env
--- a/ironic-deployment/default/kustomization.yaml
+++ b/ironic-deployment/default/kustomization.yaml
@@ -4,6 +4,10 @@ namespace: baremetal-operator-system
 namePrefix: baremetal-operator-
 resources:
 - ../base
+
+generatorOptions:
+  disableNameSuffixHash: true
+
 configMapGenerator:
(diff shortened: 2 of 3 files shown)


elfosardo:
give it another chance to trigger the job, but I think this needs to be removed

sathieu:
@elfosardo The CI is passing now. What is the next step?

tuminoid:
/cc @lentzi90 
Do you foresee issues with this?
/cc @dtantsur

lentzi90:
The reason for having a suffix hash is that you automatically get a rollout when the secret or configMap changes, because a change means a new secret or configMap is created. Any changes will then definitely be picked up directly instead of relying on the processes in the containers to reload the data eventually if at all.
I think we should make sure that BMO/Ironic actually do pick up the changes when not using the suffix hash. Otherwise we will be hard to debug behavior when users change the configMap but the containers are not restarted.

Could you explain what issues with diffing you are currently having? In my experience `kubectl diff -k /path/to/kustomization` works perfectly well for this, although it hides the secret data in the output. For secrets I'm mostly interested in if they changed or not. For configMaps we get the full diff.

Regarding the comparison of helm and kustomize output I just want to mention that the configMap and secrets are probably not something you should be comparing. They contain credentials and environment specific values. I would expect them to be different between every environment out there. Honestly I do not like that we provide "defaults" for the configMap because people sometimes assume that they can use it off the shelf without changes.

sathieu:
Thanks @lentzi90 I understand better now. In Helm, we use an annotation containing configMap or secret data, hashed, to force rollout.

Given this, I'll close this PR and keep maintaining this patch myself for https://gitlab.com/kubitus-project/external-helm-charts/metal3/.

PR #1669, opened 2024-04-15
:sparkles: Add Golang-based deploy-cli

mquhuy:
This package does the jobs that are currently handled by `deploy.sh`. It will help installing BMO, Ironic with suitable configuration.

This PR has been successfully tested here: https://github.com/metal3-io/metal3-dev-env/pull/1447

Co-author: @maxrantil


kashifest:
cc @adilGhaffarDev @lentzi90 @Rozzii @kashifest

mquhuy:
> cc @adilGhaffarDev @lentzi90 @Rozzii @kashifest

/cc @adilGhaffarDev @lentzi90 @Rozzii @kashifest
Just so that you guys are added to reviewers list

adilGhaffarDev:
Nice work @mquhuy 🎊  . Can we add some docs regarding how someone can use deploy-cli?

mquhuy:
> Nice work @mquhuy 🎊 . Can we add some docs regarding how someone can use deploy-cli?

Sure. I think @maxrantil already added some docs (in this PR), but please let me know if you would like to see it somewhere else as well.

Rozzii:
I hope I will have some time to review this in detail in the future !

config/README.md
```diff
@@ -1,8 +1,15 @@
 # Kustomizations for Baremetal Operator
 
 This folder contains kustomizations for the Baremetal Operator. They have
-traditionally been used through the [deploy.sh](../tools/deploy.sh) script,
+been used through the [deploy-cli](../tools/deploy-cli) library,
```
lentzi90, line 4: nit: "They can be used". This is new stuff, not something we have been doing :smile: 
I would also leave the note about deploy.sh, because that is indeed how they have been used traditionally (and still are in metal3-dev-env).

ironic-deployment/README.md
```diff
@@ -1,8 +1,14 @@
 # Kustomizations for Ironic
 
 This folder contains kustomizations for Ironic. They are mainly used
-through the [deploy.sh](../tools/deploy.sh) script, which takes care of
+through the [deploy-cli](../tools/deploy-cli) library, which takes care of
```
lentzi90, line 4: Same as above

tools/deploy-cli/go.mod
lentzi90, line 1: I don't have a strong opinion on this, but I think we should at least consider reusing the go module under `hack/tools`. If/when the deploy-cli has proven itself worthy, we can move it to its own module or keep it under the main module.

tools/deploy-cli/templates/ironic-kustomize.tpl
```diff
@@ -0,0 +1,80 @@
...
+{{/* The keepalived component has its own configmap, but we are overriding depending on environment here so we must replace it.*/}}
+{{ if .DeployKeepAlived }}
+- behavior: replace
+{{ else }}
+- behavior: create
+{{ end }}
+  envs:
+  - ironic_bmo_configmap.env
+  name: ironic-bmo-configmap
+{{/* This configMap is solely so that we can replace the IP placeholders */}}
```
lentzi90, line 40: If we have the value in the ironic_bmo_configmap.env we can use that instead. This is fine like it is though, just wanted to mention it in case we can simplify.

tools/deploy-cli/templates/ironic_bmo_configmap_env.tpl
```diff
@@ -0,0 +1,16 @@
...
+PROVISIONING_INTERFACE=ironicendpoint
+{{ else }}
+PROVISIONING_INTERFACE=eth2
+IRONIC_INSPECTOR_VLAN_INTERFACES=all
+{{ end }}
+DHCP_RANGE=172.22.0.10,172.22.0.100
+DEPLOY_KERNEL_URL=http://172.22.0.2:6180/images/ironic-python-agent.kernel
+DEPLOY_RAMDISK_URL=http://172.22.0.2:6180/images/ironic-python-agent.initramfs
+IRONIC_ENDPOINT=http://172.22.0.2:6385/v1/
+CACHEURL=http://172.22.0.1/images
```
lentzi90, line 13: I don't think we can have these hard coded. It will only work in CI. I think the best solution is to provide a flag (e.g. `--ironic-env-file`) that can be used to give the whole file instead of templating. We can have some default template but it will never be able to support all possible configurations.

tools/deploy-cli/deploy-cli.go
```diff
@@ -0,0 +1,324 @@
...
+	}
+	defer resp.Body.Close()
+
+	// Write the body to file
+	_, err = io.Copy(out, resp.Body)
+	return err
+}
+
+// EnsureFileExists checks if a file exists at path, creating it with random content if it doesn't.
+func EnsureFileExists(varName, path string, length int) (bool, error) {
```
lentzi90, line 103: I would be extremely surprised as a user of this function. Ensuring that the file exists should not imply that it is filled with random data. This should be called something else (maybe `EnsureCredentials`) or perhaps rather split up in two: ensure the file is there and ensure that it has some random content.
We should probably also validate the content when the file does exist.
Finally the input seems a bit random. `varName` is just used in a log message.

tools/deploy-cli/deploy-cli.go
```diff
@@ -0,0 +1,324 @@
...
+	content, err := ReadFileContent(varName, filePath)
+	if err != nil {
+		return "", err
+	}
+
+	return content, nil
+}
+
+// BuildKustomizeManifest builds a provided kustomize overlays to output, same as `kustomize build`
+func BuildKustomizeManifest(source string) ([]byte, error) {
```
lentzi90, line 158: Don't we have this in the e2e module already? Let's reuse it if we have it already and otherwise add it so that we can reuse it.

tools/deploy-cli/deploy-cli.go
```diff
@@ -0,0 +1,324 @@
...
+	inputFile := "templates/bmo-kustomize.tpl"
+	kustomizeFile := filepath.Join(tempBMOOverlay, "kustomization.yaml")
+	if err := RenderEmbedTemplateToFile(data.TemplateFiles, inputFile, kustomizeFile, data); err != nil {
+		return err
+	}
+	ironicEnvSrc := "https://raw.githubusercontent.com/metal3-io/baremetal-operator/main/config/default/ironic.env"
+	ironicEnvDst := filepath.Join(tempBMOOverlay, "ironic.env")
+	if err := DownloadFile(ironicEnvSrc, ironicEnvDst); err != nil {
+		return err
+	}
```
lentzi90, line 223: This does not make sense. That file will only work for one specific configuration. The user must be able to override. We should do it similar to Ironic above with a template and then a way for the user to use their own.

tools/deploy-cli/deploy-cli.go
```diff
@@ -0,0 +1,324 @@
...
+
+	return deployWithKustomizeAndApply(tempBMOOverlay)
+}
+
+// getKubeconfigPath determines the kubeconfig path from KUBECTL_ARGS, KUBECONFIG_PATH, or defaults to ~/.kube/config.
+func getKubeconfigPath() (string, error) {
```
lentzi90, line 229: I think we have the same in e2e. Let's reuse the code or make it reusable if necessary!

tools/deploy-cli/deploy-cli.go
```diff
@@ -0,0 +1,324 @@
...
+	if clusterProxy == nil {
+		return nil, fmt.Errorf("failed to create cluster proxy")
+	}
+	return &clusterProxy, nil
+}
+
+// deployWithKustomizeAndApply first generates the YAML configuration by running Kustomize build on the overlay directory,
+// then outputs the generated YAML to a temp file within the overlay dir, and then applies
+// the configuration to the Kubernetes cluster specified in the kubeconfig file.
+func deployWithKustomizeAndApply(overlayPath string) error {
```
lentzi90, line 275: This also feels like something very close to what we have in e2e already. Reuse if possible, or explain why it is not.

tools/deploy-cli/deploy-cli.go
```diff
@@ -0,0 +1,324 @@
...
+	fmt.Println(`Usage : deploy [options]
+Options:
+	-h:	show this help message
+	-b:	deploy BMO
+	-i:	deploy Ironic
+	-t:	deploy with TLS enabled
+	-n:	deploy without authentication
+	-k:	deploy with keepalived
+	-m:	deploy with mariadb (requires TLS enabled)`)
+}
```
lentzi90, line 324: Please add at least required environment variables (IRONIC_HOST_IP), preferably also optional variables.

lentzi90 (commented):
First pass done. I have not tested it yet, but there were quite a lot of comments already from reading so I will post them now and hopefully return to test things in a couple of days.

mquhuy:
Thank you @lentzi90.

I agree with most of what you said, but I think it's worth mentioning that the current behavior of this package is what it is because I was trying to mimic the `deploy.sh`. I have a lot of improvement ideas for it, but IMO these may come in later PRs.
Do you still think it's worthwhile to implement the suggested changes (non-hardcoded config, required env files, etc.) already in this PRs?

tools/deploy-cli/go.mod
mquhuy: Imo that would just unnecessarily complicate the implementation. The reason I splited it to a new package in the first place was due to mismatching dependencies. Also, since this works on its own binary, implementing it inside a bigger package would just increase the binary size.

tools/deploy-cli/templates/ironic-kustomize.tpl
mquhuy: Not sure I understood your point. Could you please elaborate?

tools/deploy-cli/templates/ironic_bmo_configmap_env.tpl
mquhuy: I don't have an opinion about it, but imo this would live together just fine with the said `--ironic-env-file` flag. (We can use this one only for CI, for e.g.). This hardcoded template simply is what `deploy.sh` does.

tools/deploy-cli/deploy-cli.go
mquhuy: Yes. I forgot about this, the one in e2e is currently not public, so I copied it over and made a mental note to make it public later before removing this copy.

tools/deploy-cli/deploy-cli.go
mquhuy: I agree. But again this is to follow `deploy.sh` behavior. I think we could improve it later, unless you want to do that now?

tools/deploy-cli/deploy-cli.go
mquhuy: Thanks. I have not gone through this one very carefully, I think we may not even need it, I'll check.

tools/deploy-cli/deploy-cli.go
mquhuy: Thanks. I'll do that

tools/deploy-cli/deploy-cli.go
mquhuy: Hmm. What are those required variables?
IMO there are just too many of them, we shouldn't make them flags. In this version, we take them from the environment, just like `deploy.sh` does. In later improvement, we could add a config file, so that user could provide their own vars.

tools/deploy-cli/go.mod
Rozzii: I am siding with @mquhuy on this one, I also think we shouldn't mix this small tool together with to rest of the BMO go code. My reasoning is that this is just some deployment automation not a mandatory part of BMO.

tools/deploy-cli/deploy-cli.go
lentzi90: I don't mean to add flags for them, I just mean that we should mention at least the required environment variables. If I'm not mistaken it will immediately exit with an error if you don't set IRONIC_HOST_IP.

tools/deploy-cli/templates/ironic_bmo_configmap_env.tpl
lentzi90: Yes it is what deploy.sh does, but it is actually not what we use in CI. We have an [ugly hack in metal3-dev-env](https://github.com/metal3-io/metal3-dev-env/blob/af8d1765c7d66640377ba7632419b3c3409ce018/03_launch_mgmt_cluster.sh#L142-L162) where we replace this file before we run the deploy script. :see_no_evil: 
With the template and hard-coded values built in to the binary this will no longer be possible. So I think we must at least add a flag to override and preferably also allow overriding each of the values with environment variables.
Is it possible to do the equivalent of `DHCP_RANGE=${DHCP_RANGE:-172.22.0.10,172.22.0.100}`?

tools/deploy-cli/go.mod
lentzi90: > we shouldn't mix this small tool together with to rest of the BMO go 

This I agree with, but I think it would fit perfectly well in `hack/tools`.

> Imo that would just unnecessarily complicate the implementation. The reason I splited it to a new package in the first place was due to mismatching dependencies.

I would want to know more about this. BMO seems to have kustomize v4 still. Perhaps that is the whole issue and it would work well if we just bump it to v5?

> Also, since this works on its own binary, implementing it inside a bigger package would just increase the binary size.

AFAIK this is not true. When you build a specific binary that does not include everything in the module.

IF we still decide to go with a separate module just for this, then we must also ensure that it is included properly in [linting](https://github.com/metal3-io/baremetal-operator/blob/2eb3da0fa44ddf30f57d62027a5932f4a7364401/Makefile#L139-L144) etc...

It should not be a big deal to have this under `hack/tools`. Creating a new module is IMO something to think carefully about, not something to do just because we had difficulties with the dependencies.

lentzi90:
> Do you still think it's worthwhile to implement the suggested changes (non-hardcoded config, required env files, etc.) already in this PRs?

I think it is worth making this usable before merging and getting rid of "strange" code. What I mean with this is that we should
1. Have some way of either overriding the configmaps or actually setting the values in them
2. Not download a random file from the internet. This is IMO surprising behavior and we should not need the code for doing this at all.
3. Document (in the help output) what required variables there are.

Why I think this should be done before merging is to avoid "abandoned" code. We may have good intentions to fix this later, but there is no guarantee that it will ever happen. We could be distracted or pulled into other projects and then two years later someone will find this and wonder why it was made in this way and if it is safe to change.

tools/deploy-cli/templates/ironic-kustomize.tpl
mquhuy: Thank you! It is more elegant that way. I was just having an impression that the config map values can only come from either a file or literals, not both. Happy to be wrong though xD

tools/deploy-cli/deploy-cli.go
mquhuy: Yes, then we're on the same page. I'm just having a hard time deciding which env vars should be required xD. 
My initial intention was that we will gradually add those required variables as we see fit, but of course we could start with something now. :-?

tools/deploy-cli/templates/ironic_bmo_configmap_env.tpl
lentzi90: Yes, defaults would be very nice!

tools/deploy-cli/templates/ironic-kustomize.tpl
lentzi90: It is possible to combine all in one. You can have a configmap with all three `envs`,  `files` and `litterals` :slightly_smiling_face:

tools/deploy-cli/deploy-cli.go
lentzi90: I think it would be as simple as this: If you can run the command successfully without the env var then it is not required. So try running it and see what would be needed. I noticed [this place](https://github.com/metal3-io/baremetal-operator/blob/ed781396e524255850e4ddc54c27e2cdcd7079eb/tools/deploy-cli/deploy-cli.go#L190-L192) where we fail if the variable is empty. There could be more that I missed though.

tools/deploy-cli/go.mod
mquhuy: >I would want to know more about this. BMO seems to have kustomize v4 still. Perhaps that is the whole issue and it would work well if we just bump it to v5?
Thanks, that could be the issue, but just to be clear, I did not mean that I split to a new package just because of the mismatched dependencies. It was just one of the reasons. I think @Rozzii explained it quite well, I don't see the point in mixing this tool with any other code, as it works on its own and has no architecture relationship with neither BMO nor `hack/tools` code.

>It should not be a big deal to have this under hack/tools. Creating a new module is IMO something to think carefully about, not something to do just because we had difficulties with the dependencies.

This for sure, but again the main reason for this split is because (1) it works independently, and (2) it doesn't require anything from `hack/tools`, and vice-verse. 

Also, I don't fully get the purpose of the `hack/tools` module, it seems to be a workaround for something whose at least wiki link is outdated (https://github.com/metal3-io/baremetal-operator/blob/main/hack/tools/tools.go#L4). Even the `release/notes.go` doesn't seem to belong there (it has `package main`). IMO it's better to just remove `hack/tools` go module? Or we would need to reconstruct this module if we wanted to use it for something else.

lentzi90: The hack/tools module is just a bunch of completely unrelated tools that can be useful for developing. The deploy-cli seems to fit perfectly well there. :wink: 
There is kustomize for handling manifests, controller-gen for creating CRDs and deploy-cli for deploying the result.
We may not need to have hack/tools for kustomize and controller-gen since we just consume the upstream projects directly, so from that perspective they are fine to be removed. However, having them there is a way to be explicit about what development dependencies we have.

Rozzii (commented):
Thanks @mquhuy 
In principle it looks good to me, we will see how robust the new code when we take it into use in e.g. dev-env . As a first iteration it looks good to me.

It would be a nice next step to create unit tests for this.

/lgtm

tuminoid:
I has open comments etc as well, but in general I don't want this in BMO 0.7 yet, IMO better merge it only after we've cut the release, so it can have a cycle to be tested, polished and integrated.
/hold

Let me know if you disagree.

config/README.md
```diff
@@ -3,6 +3,33 @@
 This folder contains kustomizations for the Baremetal Operator. They have
 traditionally been used through the [deploy.sh](../tools/deploy.sh) script,
 which takes care of generating the necessary config for basic-auth and TLS.
+
+Experimentally, in `deploy.sh` stead, you can use the new golang-based
```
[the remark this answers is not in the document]
mquhuy: You are correct the two phrases have same meaning, it's not a typo though.

hack/tools/deploy-cli/main.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	if deployBMOFlag {
+		err = dc.deployBMO()
+		if err != nil {
+			log.Fatalf("Failed to deploy BMO: %v", err)
+		}
+	}
+
+	// dc.cleanup()
```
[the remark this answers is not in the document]
mquhuy: Nope. Thanks, I missed that.

hack/tools/deploy-cli/deploy-cli.go
```diff
@@ -0,0 +1,456 @@
...
+
+		d.IronicOverlay = ironicOverlay
+	}
+
+	ironicOverlay := d.IronicOverlay
+	log.Printf("Installing ironic with kustomize overlay: %s", ironicOverlay)
+
+	if d.IronicBMOConfigMapEnvFile != "" {
+		log.Printf("Using custom ironic_bmo_configmap.env file: %s", d.IronicBMOConfigMapEnvFile)
+		CopyFile(d.IronicBMOConfigMapEnvFile, filepath.Join(ironicOverlay, "ironic_bmo_configmap.env"))
```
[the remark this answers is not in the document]
mquhuy: Yes. That is correct. I'll make a change

ironic-deployment/README.md
```diff
@@ -1,8 +1,36 @@
 # Kustomizations for Ironic
 
 This folder contains kustomizations for Ironic. They are mainly used
-through the [deploy.sh](../tools/deploy.sh) script, which takes care of
-generating the necessary config for basic-auth and TLS.
+traditionally been used through the [deploy.sh](../tools/deploy.sh) script,
+which takes care of generating the necessary config for basic-auth and TLS.
+
+Experimentally, in `deploy.sh` stead, you can use the new golang-based
```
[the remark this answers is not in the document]
mquhuy: Same answer :D Anw as "instead of" might be more familiar to majority of people, I'll change it.

mquhuy:
> I has open comments etc as well, but in general I don't want this in BMO 0.7 yet, IMO better merge it only after we've cut the release, so it can have a cycle to be tested, polished and integrated. /hold
> 
> Let me know if you disagree.

I'm not sure what you mean by "cut the release"?

IMO waiting for a whole release to merge this is a waste of time. It's passed all CI tests, so if we don't merge it and let people use it, how can we further test it then? We still keep `deploy.sh`, and use it in CI, so this new package is n't gonna harm anything. It's not a part of BMO code either.

config/README.md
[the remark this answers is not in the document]
NymanRobin: Cool, I had simply never heard the expression before but looked it up now! Everyday you learn

tuminoid:
> > I has open comments etc as well, but in general I don't want this in BMO 0.7 yet, IMO better merge it only after we've cut the release, so it can have a cycle to be tested, polished and integrated. /hold
> > Let me know if you disagree.
> 
> I'm not sure what you mean by "cut the release"?
> 
> IMO waiting for a whole release to merge this is a waste of time. It's passed all CI tests, so if we don't merge it and let people use it, how can we further test it then? We still keep `deploy.sh`, and use it in CI, so this new package is n't gonna harm anything. It's not a part of BMO code either.
> 
> Edit: Sorry now I think I got your point. If it just means we will wait till after 0.7 release, then we merge this, I wouldn't mind. Anw, as I said, it wouldn't hurt to have it in 0.7 either.

Yes, I meant 0.7, but if its not replacing deploy.sh, and deploy.sh is still the documented way, then it doesn't matter so much. 

I was rather thinking that we make this one replace deploy.sh  after we've cut 0.7, so it gets tested in CI for a release cycle as primary tool, and there wouldn't be two tools to maintain.

How does that sound?

mquhuy:
> I was rather thinking that we make this one replace deploy.sh after we've cut 0.7, so it gets tested in CI for a release cycle as primary tool, and there wouldn't be two tools to maintain.
> 
> How does that sound?

Yep that would be great! We will have to do that change in dev-env though, so doesn't matter if we merge this PR or not.

tuminoid:
> > I was rather thinking that we make this one replace deploy.sh after we've cut 0.7, so it gets tested in CI for a release cycle as primary tool, and there wouldn't be two tools to maintain.
> > How does that sound?
> 
> Yep that would be great! We will have to do that change in dev-env though, so doesn't matter if we merge this PR or not.

Yes I'm aware of that. However we need to maintain this in 0.7 if we merge it before cutting release, so I think we are agreeing on merging only after 0.7?

mquhuy:
> Yes I'm aware of that. However we need to maintain this in 0.7 if we merge it before cutting release, so I think we are agreeing on merging only after 0.7?

Yeah let's merge after 0.7 then

kashifest (commented):
/approve

hack/tools/go.mod
```diff
@@ -3,16 +3,20 @@ module github.com/metal3-io/baremetal-operator/tooling
 go 1.22
 
 require (
+	golang.org/x/crypto v0.26.0
+	sigs.k8s.io/cluster-api/test v1.7.5
```
tuminoid, line 7: we bumped CAPI to 1.7.6 recently, might bump this as well

hack/tools/go.mod
```diff
@@ -27,37 +31,35 @@ require (
...
 	google.golang.org/protobuf v1.33.0 // indirect
 	gopkg.in/evanphx/json-patch.v4 v4.12.0 // indirect
 	gopkg.in/inf.v0 v0.9.1 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
-	k8s.io/api v0.27.16 // indirect
-	k8s.io/apiextensions-apiserver v0.27.7 // indirect
-	k8s.io/apimachinery v0.27.16 // indirect
-	k8s.io/klog/v2 v2.90.1 // indirect
+	k8s.io/api v0.29.3 // indirect
```
tuminoid, line 55: These are at v0.29.9

ironic-deployment/README.md
```diff
@@ -1,8 +1,36 @@
...
+which takes care of generating the necessary config for basic-auth and TLS.
+
+Experimentally, instead of `deploy.sh`, you can use the new golang-based
+[deploy-cli](../hack/tools/deploy-cli) library,
+which, at the moment, handles everything `deploy.sh` does. You can either:
+
+- Run the package with `go run`:
+
+```shell
+cd $BMO_ROOT/hack/tools/deploy-cli
```
tuminoid, line 14: Where is BMO_ROOT defined? I think it would be better to just that in BMO repository, cd `hack/tools/...`

ironic-deployment/README.md
```diff
@@ -1,8 +1,36 @@
...
+go run *.go
+```
+
+- Otherwise, build the package to a static binary:
+
+```shell
+cd $BMO_ROOT
```
tuminoid, line 21: If we have previous in place, we don't have to repeat $BMO_ROOT everywhere

hack/tools/deploy-cli/templates/bmo-kustomize-bmopath.tpl
```diff
@@ -0,0 +1,29 @@
...
+  - username=ironic-username
+  - password=ironic-password
+  type: Opaque
+{{ end }}
+configMapGenerator:
+- name: ironic
+  behavior: create
+  envs:
+  - ironic.env
+
```
tuminoid, line 29: unnecessary empty line

config/README.md
```diff
@@ -3,6 +3,33 @@
 This folder contains kustomizations for the Baremetal Operator. They have
 traditionally been used through the [deploy.sh](../tools/deploy.sh) script,
 which takes care of generating the necessary config for basic-auth and TLS.
+
```
tuminoid, line 6: Not loving that we actually have two copies of this to maintain. Can we put this in one place only?

tuminoid (commented):
Some nits

Rozzii:
In general the structure and the approach looks good. 
It is a quite a bit of code to just verify by looking at it so I guess we will see the issues and edge cases after we start using this in CI or in local dev-env.
I am willing to put lgtm on this after the nits and branch conflict has been resolved

config/README.md
mquhuy: Yes. But could we do that in another pr?

tuminoid: You need to push rebase etc so why can't we have a readme fixed in this PR? It is a minor fix.

mquhuy: Because the two README has been there before this, so IMO it doesn't make sense to consolidate them here. And we might need to discuss how to have common README for both of them as well.

mquhuy:
> In general the structure and the approach looks good. It is a quite a bit of code to verify the workflow just by looking at it so I guess we will see the issues and edge cases after we start using this in CI or in local dev-env. I am willing to put lgtm on this after the nits and branch conflict has been resolved

Thank you. Of course we can only make sure how well it works after it's in CI, but at least we were able to test it in this PR: https://github.com/metal3-io/metal3-dev-env/pull/1447

tuminoid (commented):
/lgtm

I'll put a hold until the final run of tests in the dev-env PR passes.
/hold

mquhuy:
Tests passed
/unhold

PR #1671, opened 2024-04-16
🌱 Bump golang to v1.22

kashifest:
This PR bumps the golang version to v1.22 in different go modules, tests and Dockerfile.

### DIFF
--- a/.github/workflows/e2e-fixture-test.yml
+++ b/.github/workflows/e2e-fixture-test.yml
@@ -28,6 +28,15 @@ jobs:
     steps:
     - uses: actions/checkout@0ad4b8fadaa221de15dcec353f45205ec38ea70b # v4.1.4
 
+    - name: Calculate go version
+      id: vars
+      run: echo "go_version=$(make go-version)" >> $GITHUB_OUTPUT
+
+    - name: Set up Go
+      uses: actions/setup-go@0c52d547c9bc32b1aa3301fd7a9cb496313a4491 # v5.0.0
+      with:
+        go-version: ${{ steps.vars.outputs.go_version }}
+
     - name: Build BMO e2e Docker Image
       env:
         IMG: quay.io/metal3-io/baremetal-operator:e2e
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -27,7 +27,7 @@ jobs:
     - name: Install go
       uses: actions/setup-go@0c52d547c9bc32b1aa3301fd7a9cb496313a4491 # v5.0.0
       with:
-        go-version: '1.21'
+        go-version: '1.22'
     - name: Generate release notes
       run: |
         make release-notes
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -56,7 +56,7 @@ linters:
   fast: true
 linters-settings:
   gosec:
-    go: "1.21"
+    go: "1.22"
     severity: medium
     confidence: medium
     concurrency: 8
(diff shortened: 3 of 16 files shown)


kashifest:
/hold
This requires more discussion and project infra changes, this PR should not go in now until we have decided to bump the golang to v1.22

kashifest:
@dtantsur @honza @zaneb when do you think we would be ready to bump golang to v1.22 for main branch in BMO, once we decide I will push changes here and in other metal3 repo's main branches accordingly

tuminoid:
/retest
Checking github action retriggering via prow.

dtantsur:
Do we have a need to bump it? Doing so complicates local development for everyone who is not on the bleeding edge like me.

```
$ go version
go version go1.21.9 linux/amd64
```

We're also going to force the bump on any downstream consumers, whether they're ready or not.

tuminoid:
Golang 1.21 end of support date is 2024-08-08, so we need to bump to it sooner or later. Since we need to maintain our releases for roughly a year, we'll need to bump to it release branches as well (we did the same for 1.20 -> 1.21 too when it went EOS).

So we might hold off for a while, but we need to bump it before next minor anyways. The later we do it, the less testing it gets, so there is trade-off.

tuminoid:
Also, we already have dependencies we cannot bump, because they have moved to 1.22, like https://github.com/metal3-io/cluster-api-provider-metal3/pull/1617
There will be more and more the closer we move to EOS.

kashifest:
Yes, I agree with @tuminoid , the sooner we do it, the better, projects like k8s itself, CAPI etc have already bumped it to 1.22 in main branch, I was waiting for BMO to be ready to make it in all metal3 main branches where applicable.

dtantsur:
> Golang 1.21 end of support date is 2024-08-08, so we need to bump to it sooner or later.

I'm not following the logic. I thought the directive in go.mod only states the minimum version and users are welcome to use newer ones? My objection holds: we're pushing the change even on users of distributions that will maintain Go beyond its EOS date.

If/when we're forced to make the change because of a dependency bump, it's fine by me.

tuminoid:
> > Golang 1.21 end of support date is 2024-08-08, so we need to bump to it sooner or later.
> 
> I'm not following the logic. I thought the directive in go.mod only states the minimum version and users are welcome to use newer ones? My objection holds: we're pushing the change even on users of distributions that will maintain Go beyond its EOS date.
> 
> If/when we're forced to make the change because of a dependency bump, it's fine by me.

You're correct, go.mod bump is not strictly necessary. However, when Go 1.21 reaches EOS, is there an actual reason to keep it? Users should be using Go 1.22 at that point and it is only 2 months away. We do have some deps requiring it lined up, so it will be soon at least partially forced on us as well, so IMO it is cleaner to also switch directive to `go 1.22` right away when we merge.

Rozzii:
The lack of go 1.22 blocks this PR https://github.com/metal3-io/baremetal-operator/pull/1751

dtantsur:
> is there an actual reason to keep it?

The distributions don't move so quickly as we'd wish. Also, bumping Go versions should not be taken easily: I've just spent a whole morning on an enormous patch to move one of our components to Go 1.22 because the old controller-runtime segfaulted on it.

If we're forced by gophercloud, so be it. But I don't want such bumps (in go.mod specifically, not in workflows) to become a standard practice.

lentzi90 (commented):
/approve

PR #1681, opened 2024-04-17
:seedling: Remove capm3_e2e dependency from bmo E2E

mquhuy:
BMO E2E is using some functions from CAPM3-E2E. This likely causes circular imports in the future.


kashifest (commented):
This is a very good approach indeed.

/approve

tuminoid (commented):
/lgtm

PR #1683, opened 2024-04-18
🌱 remove variables specific to ironic from bmo env file

sathieu:

I found it hard to understand where the variables are used. This is a tentative to clean this up a bit.

Warning: PR created after reading the code. I may have missed something!


Rozzii:
These are used within the Ironic Pod and manage Ironic's config @sathieu

sathieu:
ok. thanks

PR #1616, opened 2024-03-12
🌱 Mount ironic HTPASSWD as volume instead of environment variable

MahnoorAsghar:
Security baselines such as CIS do not recommend using secrets as environment variables, but using files instead. Therefore, the IRONIC_HTPASSWD and INSPECTOR_HTPASSWD will now be mounted as volumes instead of environment variables.

PR #1685, opened 2024-04-18
Mount ironic HTPASSWD as volume instead of environment variable

MahnoorAsghar:
Security baselines do not recommend using secrets as environment variables but using files instead. Therefore, the IRONIC_HTPASSWD and INSPECTOR_HTPASSWD will now be mounted as volumes instead of environment variables.

### DIFF
--- a/tools/run_local_ironic.sh
+++ b/tools/run_local_ironic.sh
@@ -169,11 +169,18 @@ if [[ -r "${IPXE_KEY_FILE}" ]]; then
 fi
 
 BASIC_AUTH_MOUNTS=""
-IRONIC_HTPASSWD=""
+IRONIC_HTPASSWD_USERNAME_FILE="${IRONIC_DATA_DIR}/auth/ironic-username"
+IRONIC_HTPASSWD_PASSWORD_FILE="${IRONIC_DATA_DIR}/auth/ironic-password"
+set +x
 if [ -n "$IRONIC_USERNAME" ]; then
-     IRONIC_HTPASSWD="$(htpasswd -n -b -B "${IRONIC_USERNAME}" "${IRONIC_PASSWORD}")"
-     IRONIC_HTPASSWD="--env HTTP_BASIC_HTPASSWD=${IRONIC_HTPASSWD} --env IRONIC_HTPASSWD=${IRONIC_HTPASSWD}"
+     echo "$IRONIC_USERNAME" > "${IRONIC_HTPASSWD_USERNAME_FILE}"
+     IRONIC_HTPASSWD_USERNAME="-v ${IRONIC_HTPASSWD_USERNAME_FILE}:/auth/ironic/username"
 fi
+if [ -n "$IRONIC_PASSWORD" ]; then
+     echo "${IRONIC_PASSWORD}" > "${IRONIC_HTPASSWD_PASSWORD_FILE}"
+     IRONIC_HTPASSWD_PASSWORD="-v ${IRONIC_HTPASSWD_PASSWORD_FILE}:/auth/ironic/password"
+fi
+set -x
 
 sudo mkdir -p "$IRONIC_DATA_DIR/html/images"
 # Locally supplied IPA images are imported here when the environment variables are set accordingly.
@@ -192,6 +199,7 @@ fi
 
 "$SCRIPTDIR/tools/remove_local_ironic.sh"
 
+set +x
 if [ "$IRONIC_USE_MARIADB" = "true" ]; then
     # set password for mariadb
     mariadb_password=$(echo "$(date;hostname)"|sha256sum |cut -c-20)
@@ -199,6 +207,7 @@ if [ "$IRONIC_USE_MARIADB" = "true" ]; then
 else
     IRONIC_MARIADB_PASSWORD=
 fi
+set -x
 
 POD=""
 
@@ -235,9 +244,10 @@ sudo "${CONTAINER_RUNTIME}" run -d --net host --privileged --name dnsmasq \
 # https://github.com/metal3-io/ironic-image/blob/main/scripts/runhttpd
 # shellcheck disable=SC2086
 sudo "${CONTAINER_RUNTIME}" run -d --net host --privileged --name httpd \
-     ${POD} ${CERTS_MOUNTS} ${BASIC_AUTH_MOUNTS} ${IRONIC_HTPASSWD} \
+     ${POD} ${CERTS_MOUNTS} ${BASIC_AUTH_MOUNTS} \
      --env-file "${IRONIC_DATA_DIR}/ironic-vars.env" \
-     -v "$IRONIC_DATA_DIR:/shared" --entrypoint /bin/runhttpd "${IRONIC_IMAGE}"
+     -v "${IRONIC_DATA_DIR}:/shared" ${IRONIC_HTPASSWD_USERNAME} \
+     ${IRONIC_HTPASSWD_PASSWORD} --entrypoint /bin/runhttpd "${IRONIC_IMAGE}"
 
 if [ "$IRONIC_USE_MARIADB" = "true" ]; then
     # https://github.com/metal3-io/mariadb-image/blob/main/runmariadb
@@ -252,10 +262,11 @@ fi
 # https://github.com/metal3-io/ironic-image/blob/main/scripts/runironic
 # shellcheck disable=SC2086
 sudo "${CONTAINER_RUNTIME}" run -d --net host --privileged --name ironic \
-     ${POD} ${CERTS_MOUNTS} ${BASIC_AUTH_MOUNTS} ${IRONIC_HTPASSWD} \
+     ${POD} ${CERTS_MOUNTS} ${BASIC_AUTH_MOUNTS} \
      --env-file "${IRONIC_DATA_DIR}/ironic-vars.env" \
      ${IRONIC_MARIADB_PASSWORD} --entrypoint /bin/runironic \
-     -v "$IRONIC_DATA_DIR:/shared" "${IRONIC_IMAGE}"
+     -v "$IRONIC_DATA_DIR:/shared" ${IRONIC_HTPASSWD_USERNAME} \
+     ${IRONIC_HTPASSWD_PASSWORD} "${IRONIC_IMAGE}"
 
 # Start ironic-endpoint-keepalived
 # shellcheck disable=SC2086
--- a/ironic-deployment/components/basic-auth/auth.yaml
+++ b/ironic-deployment/components/basic-auth/auth.yaml
@@ -8,8 +8,14 @@ spec:
       containers:
       - name: ironic
         envFrom:
-        # This is the htpassword matching the ironic password
-        - secretRef:
-            name: ironic-htpasswd
         - configMapRef:
             name: ironic-bmo-configmap
+        volumeMounts:
+        - name: ironic-htpasswd
+          mountPath: "/auth/ironic"
+          readOnly: true
+      volumes:
+      # This is the htpassword matching the ironic password
+      - name: ironic-htpasswd
+        secret:
+          secretName: ironic-htpasswd
--- a/ironic-deployment/overlays/basic-auth_tls/basic-auth_tls.yaml
+++ b/ironic-deployment/overlays/basic-auth_tls/basic-auth_tls.yaml
@@ -8,7 +8,13 @@ spec:
       containers:
       - name: ironic-httpd
         envFrom:
-        - secretRef:
-            name: ironic-htpasswd
         - configMapRef:
             name: ironic-bmo-configmap
+        volumeMounts:
+        - name: ironic-htpasswd
+          mountPath: "/auth/ironic"
+          readOnly: true
+      volumes:
+      - name: ironic-htpasswd
+        secret:
+          secretName: ironic-htpasswd
--- a/tools/deploy.sh
+++ b/tools/deploy.sh
@@ -148,8 +148,8 @@ if [[ "${DEPLOY_BASIC_AUTH}" == "true" ]]; then
     fi
 
     if [[ "${DEPLOY_IRONIC}" == "true" ]]; then
-        echo "IRONIC_HTPASSWD=$(htpasswd -n -b -B "${IRONIC_USERNAME}" "${IRONIC_PASSWORD}")" > \
-        "${TEMP_IRONIC_OVERLAY}/ironic-htpasswd"
+        echo "${IRONIC_USERNAME}" > "${TEMP_IRONIC_OVERLAY}/ironic-username"
+        echo "${IRONIC_PASSWORD}" > "${TEMP_IRONIC_OVERLAY}/ironic-password"
     fi
 fi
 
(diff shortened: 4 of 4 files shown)


MahnoorAsghar:
(This PR is work continued from https://github.com/metal3-io/baremetal-operator/pull/1616)

lentzi90 (commented):
This looks good, but there are a few more places that needs changes to get this in the e2e tests. Can you check the `e2e*` folders [here](https://github.com/metal3-io/baremetal-operator/tree/main/ironic-deployment/overlays)? For example [this](https://github.com/metal3-io/baremetal-operator/blob/d0378b411b68fdd16391dd29be29b9edb6110974/ironic-deployment/overlays/e2e/kustomization.yaml#L21-L26) should change. The secrets for the e2e tests are generated [here](https://github.com/metal3-io/baremetal-operator/blob/d0378b411b68fdd16391dd29be29b9edb6110974/hack/ci-e2e.sh#L169-L191) which may also need some changes.

mboukhalfa:
I think the github branch protection here is not yet update 
the test triggers now are :
```
/test metal3-centos-e2e-integration-test-main
/test metal3-ubuntu-e2e-integration-test-main
```
The approver can skip the trigger with you do not need to consider this once we update soon:
`/override test-centos-e2e-integration-main`

hack/ci-e2e.sh
```diff
@@ -176,17 +176,17 @@ for overlay in "${BMO_OVERLAYS[@]}"; do
 done
 
 for overlay in "${IRONIC_OVERLAYS[@]}"; do
-  echo "IRONIC_HTPASSWD=$(htpasswd -n -b -B "${IRONIC_USERNAME}" "${IRONIC_PASSWORD}")" > \
-    "${overlay}/ironic-htpasswd"
+  echo "${IRONIC_USERNAME}" > "${overlay}/ironic-username"
+  echo "${IRONIC_PASSWORD}" > "${overlay}/ironic-password"
```
lentzi90, line 180: I'm confused by this. We used to put an htpasswd here, which contained the username in plain text and the password hashed by bcrypt. That means the password is not directly readable but still verifiable. With this change we have both username and password as plain text which is definitely not best practice.
Is it really the intention to mount them as plain text? Why?

ironic-deployment/overlays/e2e-release-24.0-with-inspector/kustomization.yaml
```diff
@@ -26,12 +26,16 @@ images:
 secretGenerator:
 - name: ironic-htpasswd
   behavior: create
-  envs:
-  - ironic-htpasswd
+  files:
+  - username=ironic-username
+  - password=ironic-password
```
lentzi90, line 31: The secret is called `ironic-htpasswd` because it used to contain ah [htpasswd formatted file](https://httpd.apache.org/docs/2.4/programs/htpasswd.html). Now it just contains the username and password as plain text. If we really intend to have them as plain text we should change the name to avoid confusion.

lentzi90 (changes requested):
/hold
I don't think we should change from htpasswd to plaintext passwords!
/cc @tuminoid

ironic-deployment/overlays/e2e-release-24.0-with-inspector/kustomization.yaml
MahnoorAsghar: Didn't it just contain an htpasswd string generated by the command? Basically ironic-image will now use the same command to generate the htpsasswd itself, so it needs the ironic username and password as input. Since it will eventually be used to generate the same htpasswd, I think keeping the name the same makes sense.

hack/ci-e2e.sh
MahnoorAsghar: I was reading about kustomize secretGeneration using files [here](https://kubernetes.io/docs/tasks/configmap-secret/managing-secret-using-kustomize/#:~:text=In%20all%20cases%2C%20you%20don%27t%20need%20to%20base64%20encode%20the%20values.) Mounting the password as plain text does seem wrong, though. Will ironic-image be able to generate an htpasswd with the ironic-password if its given a hash of the password? It could, but it wouldn't be correct, but more secure overall? What do you think?

ironic-deployment/overlays/e2e-release-24.0-with-inspector/kustomization.yaml
lentzi90: Clarified offline, commenting for closure.
The issue is that we then give ironic access to the password in plain text, which is more information than it needs. The hash is enough for its purpose so it is better to "limit" the access level to that :slightly_smiling_face:

hack/ci-e2e.sh
lentzi90: Clarified offline, commenting for closure.
We should essentially revert https://github.com/metal3-io/ironic-image/pull/482/files so that ironic does not generate the htpasswd. Instead, we put that directly in the secret and mount it as a file that ironic can use.

dtantsur: Sorry, but absolutely no. Mounting secrets is the normal and documented way to provide information into pods. The pod is responsible for any fancy format it uses. Passing htpasswd in any form is just increasing the load on users trying to install ironic (and provides for some terrible code in ironic-standalone-operator), especially since ironic does not accept any htpasswd (only certain ciphers).

ironic-deployment/overlays/e2e-release-24.0-with-inspector/kustomization.yaml
dtantsur: See above. This is fake security: if you have read access to the Ironic pod, you have all BMC credentials and all configdrives. In that cases, knowing the API password is not overly useful anymore.

hack/ci-e2e.sh
dtantsur: Oh, and another consideration. We're working on liveness/readiness checks that do more than just make sure the API exists. These will require credentials. I left a TODO until Mahnoor's work is done: https://github.com/metal3-io/ironic-image/blob/main/scripts/ironic-probe.j2#L7-L9

dtantsur: On the bright side, when we split away configuration as you're planning @lentzi90, the secret will only be mounted for the short duration of the init container.

lentzi90: Fair enough! I understand the reasoning. It was just hard to understand from the initial description that hinted at *better* security. Going from environment variable -> mounted file I can see as some kind of improvement to security, but to then at the same time get rid of bcrypt raised red flags for me.

I'm fine storing it as plain text. However, I must then insist that we rename the secret. It cannot be named `ironic-htpasswd` "because ironic will make an htpasswd out of it". It will be too confusing.

ironic-deployment/overlays/e2e-release-24.0-with-inspector/kustomization.yaml
lentzi90: Agreed. @MahnoorAsghar please change the PR so it is clear we are going to *change* the ironic credentials to be plain text *and* we are going to mount the secrets instead of consuming them as environment variables. My confusion came from the description that we are going to mount *htpasswd* as a volume which we do not.

hack/ci-e2e.sh
dtantsur: Yeah, no disagreement on the naming part.

ironic-deployment/overlays/e2e-release-24.0-with-inspector/kustomization.yaml
lentzi90, line 1: Do we have the necessary changes in the ironic-image at 24.0?
I triggered the optional job that should test this to make sure

lentzi90 (commented):
/approve

lentzi90:
/hold
Looks like we need the old way for the older image still. It should be enough to revert the changes in the the release-24.0 kustomization and adapt the ci-e2e.sh script so it generates the htpasswd for that folder. The rest can use the new way.

ironic-deployment/base/ironic.yaml
```diff
@@ -169,6 +175,9 @@ spec:
       volumes:
       - name: ironic-data-volume
         emptyDir: {}
+      - name: ironic-credentials
+        secret:
+          secretName: ironic-credentials
```
lentzi90, line 180: Why do we need this in the `base`? (I'm open to make basic-auth mandatory and part of the base, but then I think we should get rid of the basic-auth component.)
I'm afraid this will make it impossible to deploy without credentials, which can be convenient in development.

MahnoorAsghar: In trying to resolve the 'metal3-bmo-e2e-test-optional-pull' optional job, I edited this file and forgot to change it back. I have removed the credentials from it now.

lentzi90 (commented):
Great job!
/hold cancel

hack/ci-e2e.sh
zaneb: Hey, sorry to reopen this discussion, but I think there is potentially a better way of doing this that still meets all the future requirements (for clarity, I mean do these steps instead of any changes that are in progress or have merged):

- Leave the startironic logic as it was for now
- Mount the htpasswd from the secret in the same location that the startironic script would have saved it instead of passing it as an env var. (This solves the problem without breaking backward compatibility for any existing deployments that pass it as an env var.)
- In future, add a separate init container setup script that takes the plaintext password (mounted from a secret) and generates an htpasswd file saved in the correct location for startironic to read it
- Further in the future, generate a second username/password in the init container for use with by the readiness check, and add that into the same htpasswd file. Save _this_ password as plaintext to pass through to the ironic container for the use of the readiness check. (There's no authz in ironic, but this is marginally more secure as it gets regenerated every time the pod starts.)

I think that would avoid breaking backward compat, while still allowing us to make things easier to deploy over time (even without ironic-operator), and not blocking us from adding readiness checks either.

lentzi90: Sounds reasonable to me! Do you mean that we should revert https://github.com/metal3-io/ironic-image/pull/482 then?

hack/ci-e2e.sh
```diff
@@ -176,13 +176,20 @@ for overlay in "${BMO_OVERLAYS[@]}"; do
 done
 
 for overlay in "${IRONIC_OVERLAYS[@]}"; do
-  echo "IRONIC_HTPASSWD=$(htpasswd -n -b -B "${IRONIC_USERNAME}" "${IRONIC_PASSWORD}")" > \
-    "${overlay}/ironic-htpasswd"
+  if [[ "${overlay}" = "${REPO_ROOT}/ironic-deployment/overlays/e2e-release-24.0-with-inspector" || "${overlay}" = "${REPO_ROOT}/ironic-deployment/overlays/e2e-with-inspector" ]]; then
```
tuminoid, line 179:
```suggestion
  if [[ "${overlay}" = "${REPO_ROOT}/ironic-deployment/overlays/e2e-release-24.0-with-inspector" ]] || [[ "${overlay}" = "${REPO_ROOT}/ironic-deployment/overlays/e2e-with-inspector" ]]; then
```

tools/run_local_ironic.sh
```diff
@@ -169,11 +169,18 @@ if [[ -r "${IPXE_KEY_FILE}" ]]; then
...
 
 BASIC_AUTH_MOUNTS=""
-IRONIC_HTPASSWD=""
+IRONIC_CRED_USERNAME_FILE="${IRONIC_DATA_DIR}/auth/ironic-username"
+IRONIC_CRED_PASSWORD_FILE="${IRONIC_DATA_DIR}/auth/ironic-password"
+set +x
 if [ -n "$IRONIC_USERNAME" ]; then
-     IRONIC_HTPASSWD="$(htpasswd -n -b -B "${IRONIC_USERNAME}" "${IRONIC_PASSWORD}")"
-     IRONIC_HTPASSWD="--env HTTP_BASIC_HTPASSWD=${IRONIC_HTPASSWD} --env IRONIC_HTPASSWD=${IRONIC_HTPASSWD}"
+     echo "$IRONIC_USERNAME" > "${IRONIC_CRED_USERNAME_FILE}"
```
tuminoid, line 176:
```suggestion
     echo "${IRONIC_USERNAME}" > "${IRONIC_CRED_USERNAME_FILE}"
```

tuminoid (commented):
Some nits.

hack/ci-e2e.sh
zaneb: Sorry, s/startironic/[runironic](https://github.com/metal3-io/ironic-image/blob/main/scripts/runironic)/. i.e. the ironic-image container when used to run ironic (as opposed to when it is used to run httpd, or ironic-inspector, or dnsmasq).

> - We all agree to mount the htpasswd under /auth/ironic/htpasswd in ironic-image (ironic-image now no longer accepts the IRONIC_HTPASSWD environment variable and reads from the file instead).

I was thinking both.

- If the user passes nothing, then `/auth/ironic/htpasswd` doesn't exist
- If the user passes IRONIC_HTPASSWD then it gets written to `/auth/ironic/htpasswd`
- If the user doesn't pass IRONIC_HTPASSWD but mounts a file at `/auth/ironic/htpasswd` then it's already there

MahnoorAsghar: So we just provide a new way in ironic-image for htpasswd to be provided (via file), and security-conscious consumers can use the new way. Does that fully solve the inherent complaint about ironic-image being insecure? Maybe it addresses it sufficiently?

dtantsur: I don't think we can make the change without a compatibility layer still. The htpasswd is put into `/etc/ironic/htpasswd` and `/etc/ironic/htpasswd-rpc` conditionally and some options are set. What you suggest @zaneb makes certain sense, but it requires reworking runironic either way.

@MahnoorAsghar we probably don't want to have 2 ways to pass htpasswd long-term. IRONIC_HTPASSWD will need to be kept for backward compatibility for some time.

MahnoorAsghar: So for now, we can keep both options, and eventually phase out the environment variable IRONIC_HTPASSWD.

zaneb: > Does that fully solve the inherent complaint about ironic-image being insecure? Maybe it addresses it sufficiently?

I think the complaint is about how we expose data from a Secret (i.e. as an environment variable), not how ironic-image receives it. So as long as we switch to the new method (i.e. mounting as file), I think the complaints go away.

zaneb: > it requires reworking runironic either way

Yes, that's fine, TBH I haven't looked into any of the details. The mount location could even be somewhere different (as long as it is clearly documented) and we copy the file to where it is needed at startup if it exists, or generate the file where it is needed from the env vars if they exist. Whatever implementation makes sense.

tuminoid:
/retest 
CI should be better.

tuminoid:
Seems like an actual issue with the PR, unless it is timing out due CI slowness on the background.

tools/run_local_ironic.sh
```diff
@@ -170,10 +170,15 @@ fi
 
 BASIC_AUTH_MOUNTS=""
 IRONIC_HTPASSWD=""
```
elfosardo, line 172: I wonder if we need this as we don't use it undefined anymore

tools/run_local_ironic.sh
```diff
@@ -170,10 +170,15 @@ fi
...
+IRONIC_HTPASSWD_FILE="${IRONIC_DATA_DIR}/auth/ironic-htpasswd"
+IRONIC_HTPASSWD_MOUNT=""
+set +x
 if [ -n "$IRONIC_USERNAME" ]; then
      IRONIC_HTPASSWD="$(htpasswd -n -b -B "${IRONIC_USERNAME}" "${IRONIC_PASSWORD}")"
-     IRONIC_HTPASSWD="--env HTTP_BASIC_HTPASSWD=${IRONIC_HTPASSWD} --env IRONIC_HTPASSWD=${IRONIC_HTPASSWD}"
+     echo "${IRONIC_HTPASSWD}" > "${IRONIC_HTPASSWD_FILE}"
```
elfosardo, line 178: you could probably just redirect the output of the htpasswd command directly into the file and remove the need of the IRONIC_HTPASSWD variable entirely

MahnoorAsghar: Removed the variable

tools/run_local_ironic.sh
MahnoorAsghar: Fixed

PR #1705, opened 2024-04-29
🐛 Demote frequent routine logging messages to debug level

dtantsur:
Having these on each reconcile makes logs quite unreadable.

Also make the PreprovisioningImage message more useful by including the
actual image.


kashifest (commented):
/approve

PR #1717, opened 2024-05-02
Refactoring: split update_opts.go and updateopts_test.go

dtantsur:
update_opts.go contains helpers for Ironic PATCH API that are not tied
to the provisioner and thus are better haused in the clients module
(renamed to match the existing convention).

updateopts_test.go is a mix of tests for update_opts.go and a few
provision-specific functions. The former are migrated to clients as
well, the latter - to the more suitable provision_test.py.


dtantsur:
/retitle 🌱 Refactoring: split update_opts.go and updateopts_test.go

honza:
Seems reasonable to me

/approve

PR #1725, opened 2024-05-10
:seedling:  Add more ctrl options flags

honza:
We add a few command-line flags to the baremetal-operator binary to configure more `ctrl.Options{}`.

### DIFF
--- a/main.go
+++ b/main.go
@@ -23,6 +23,7 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
+	"time"
 
 	metal3api "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
 	metal3iocontroller "github.com/metal3-io/baremetal-operator/controllers/metal3.io"
@@ -125,6 +126,9 @@ func main() {
 	var restConfigQPS float64
 	var restConfigBurst int
 	var controllerConcurrency int
+	var leaseDurationSeconds string
+	var renewDeadlineSeconds string
+	var retryPeriodSeconds string
 
 	// From CAPI point of view, BMO should be able to watch all namespaces
 	// in case of a deployment that is not multi-tenant. If the deployment
@@ -168,6 +172,11 @@ func main() {
 			"Insecure values: "+strings.Join(tlsCipherInsecureValues, ", ")+".")
 	flag.IntVar(&controllerConcurrency, "controller-concurrency", 0,
 		"Number of CRs of each type to process simultaneously")
+
+	flag.StringVar(&leaseDurationSeconds, "lease-duration-seconds", os.Getenv("LEASE_DURATION_SECONDS"), "Leader election duration in seconds.")
+	flag.StringVar(&renewDeadlineSeconds, "renew-deadline-seconds", os.Getenv("RENEW_DEADLINE_SECONDS"), "Leader election renew deadline duration in seconds.")
+	flag.StringVar(&retryPeriodSeconds, "retry-period-seconds", os.Getenv("RETRY_PERIOD_SECONDS"), "Leader election retry period in seconds.")
+
 	flag.Parse()
 
 	logOpts := zap.Options{}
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -210,16 +221,47 @@ func main() {
...
 			TLSOpts: tlsOptionOverrides,
 		}),
-		LeaderElection:          enableLeaderElection,
-		LeaderElectionID:        leaderElectionID,
-		LeaderElectionNamespace: leaderElectionNamespace,
-		HealthProbeBindAddress:  healthAddr,
+		LeaderElection:                enableLeaderElection,
+		LeaderElectionID:              leaderElectionID,
+		LeaderElectionNamespace:       leaderElectionNamespace,
+		LeaderElectionReleaseOnCancel: leaderElectionReleaseOnCancel,
```
zaneb, line 227: TBH I'm struggling to think of a reason we couldn't set this to true unconditionally.

The [docs](https://github.com/kubernetes-sigs/controller-runtime/blob/479b723944e34ae42c9911fe01228ff34eb5ca81/pkg/manager/manager.go#L197-L201) say:

> This requires the binary to immediately end when the Manager is stopped, otherwise this setting is unsafe.

As far as I can tell we never stop the manager and leave other stuff running, so there should be no downside to always setting this.

Even the terminology here ("on cancel") is referring to internal golang stuff that doesn't really make sense as a CLI flag.

main.go
```diff
@@ -168,6 +174,11 @@ func main() {
 			"Insecure values: "+strings.Join(tlsCipherInsecureValues, ", ")+".")
 	flag.IntVar(&controllerConcurrency, "controller-concurrency", 0,
 		"Number of CRs of each type to process simultaneously")
+
+	flag.StringVar(&leaseDuration, "lease-duration", os.Getenv("LEASE_DURATION"), "Golang duration, e.g. '26s'")
```
zaneb, line 178: Help strings could be something like `Leader election lease duration with units, e.g. "15s"`

Ideally we should use the [defaults](https://github.com/kubernetes-sigs/controller-runtime/blob/479b723944e34ae42c9911fe01228ff34eb5ca81/pkg/manager/manager.go#L210-L221) (which are 15s, 10s, and 2s, respectively) as the examples.

main.go
```diff
@@ -210,16 +221,47 @@ func main() {
...
+		LeaderElectionReleaseOnCancel: leaderElectionReleaseOnCancel,
+		HealthProbeBindAddress:        healthAddr,
 		Cache: cache.Options{
 			ByObject:          secretutils.AddSecretSelector(nil),
 			DefaultNamespaces: watchNamespaces,
 		},
 	}
 
+	if leaseDuration != "" {
+		dur, err := time.ParseDuration(leaseDuration)
```
zaneb, line 236: Looking at the defaults, it's tempting to skip the parsing and make the command line argument `-lease-duration-seconds`. Nobody is going to want a duration in years or even hours. It could be a float if they really want to get down to the milliseconds.

zaneb (commented):
/approve

main.go
honza: If we have `leaseDurationSeconds`, it'll create weird temporaries like:

```go
leaseDurationSecondsDuration := time.Duration(leaseDurationSeconds) * time.Second
ctrlOptions.LeaseDuration = &leaseDurationSecondsDuration
```

But more annoyingly, we'll have to parse the env vars to `int`s. So this tidy line:

```go
flag.StringVar(&renewDeadline, "renew-deadline", os.Getenv("RENEW_DEADLINE"), "Golang duration, e.g. '26s'")
```

would become something more involved because `os.Getenv` returns a string but the default must be a number.

main.go
honza: Done

zaneb:
@honza did you forget to push your local changes?

main.go
zaneb: > it'll create weird temporaries

TBH I would take that one extra line of code over complicated user documentation.

> we'll have to parse the env vars to ints

`flag.IntVar()` is only a convenience. If you wanted to keep it as `StringVar` and continue to do the parsing to int/float here yourself I don't think there's anything stopping you.

honza:
> did you forget to push your local changes?

Nope.  But I did push them to the wrong branch... :disappointed:

main.go
honza: I switched it to using seconds, left it as a `StringVar` and added `strconv.ParseInt`.

main.go
```diff
@@ -210,16 +219,49 @@ func main() {
...
+		}
+
+		duration := time.Second * time.Duration(seconds)
+		ctrlOpts.LeaseDuration = &duration
+	}
+
+	if renewDeadlineSeconds != "" {
+		seconds, err := strconv.ParseInt(renewDeadlineSeconds, 10, 16)
+		if err != nil {
+			setupLog.Error(err, "failed to parse duration")
```
dtantsur, line 247: s/duration/renew deadline/ ?

main.go
```diff
@@ -210,16 +219,49 @@ func main() {
...
+		}
+
+		duration := time.Second * time.Duration(seconds)
+		ctrlOpts.RenewDeadline = &duration
+	}
+
+	if retryPeriodSeconds != "" {
+		seconds, err := strconv.ParseInt(retryPeriodSeconds, 10, 16)
+		if err != nil {
+			setupLog.Error(err, "failed to parse duration")
```
dtantsur, line 258: s/duration/retry period/ ?

honza: Done

PR #1728, opened 2024-05-14
🌱 E2E: Run as github workflow

lentzi90:

Run the E2E tests as github workflows. This adds one reusable workflow (test-e2e.yml) that can be called by other workflows, in this PR test-e2e-pull.yml and test-e2e-periodic.yml. This way we avoid duplication and can have different configuration for pull requests and periodics.

- test-e2e-periodic.yml runs the e2e suite for redfish, redfish-virtualmedia and ipmi
- test-e2e-pull.yml runs the e2e suite for redfish-virtualmedia and ipmi

Fixes #

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -1,0 +1,47 @@
+name: E2E Test
+
+on:
+  workflow_call:
+    inputs:
+      bmc-protocol:
+        required: true
+        type: string
+      runner:
+        type: string
+        default: "ubuntu-latest-4-cores"
+
+permissions: {}
+
+jobs:
+  test:
+    name: e2e test
+    runs-on: ${{ inputs.runner }}
+
+    permissions:
+      contents: read
+
+    steps:
+    - uses: actions/checkout@0ad4b8fadaa221de15dcec353f45205ec38ea70b # v4.1.4
+
+    - name: Install libvirt
+      run: |
+          sudo apt-get update
+          sudo apt-get install -y libvirt-daemon-system qemu-kvm virt-manager
+
+    - name: Run BMO e2e Tests
+      env:
+        BMC_PROTOCOL: ${{ inputs.bmc-protocol }}
+      # We need a new shell to pick up the new group. That is why we do the sudo -s -u $USER ...
+      # Remove the pre-installed go version. We install the exact version we need.
+      run: |
+        sudo usermod -a -G libvirt $USER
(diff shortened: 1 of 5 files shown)


tuminoid:
/hold
This is a test using CNCF shared resources.

mquhuy:
Unless there's constraints resource-wise, this is a very good improvement imo. Anyway, I'm wondering if there's a way to override this test? I know it runs automatically, but I'm talking about situations when it failed for some other reason, which we know we could ignore.

Otherwise, /lgtm

.github/workflows/e2e-test.yml
```diff
@@ -0,0 +1,62 @@
...
+  test:
+    name: e2e test
+    runs-on: ubuntu-latest-8-cores
+    strategy:
+      fail-fast: false
+      matrix:
+        bmc-protocol:
+        # - redfish
+        - redfish-virtualmedia
+        # - ipmi
```
kashifest, line 30: shall we comment out ipmi and test how is it behaving ?

kashifest:
> /hold This is a test using CNCF shared resources.

CNCF service desk has replied:

`Hi Kashif Nizam Khan, correct, since you’re in GHE you don’t have to worry about those billable hours. We just need to be good custodians as a community about how much we consume.`

I feel we can proceed with switching BMO e2e on github large runners since these are free resource for us. So this PR is ready to go in IMO.

/hold cancel

.github/workflows/e2e-test.yml
lentzi90: In the current tests we run both redfish-virtualmedia and ipmi on PRs and all on periodics. I commented out all but one here to conserve resources while testing. I'll push an update with periodic job and the same protocols tested as we have now :slightly_smiling_face:

kashifest:
I think we need to ensure the golang version here is correct as well, check how I fixed it here https://github.com/metal3-io/baremetal-operator/commit/e76ad63e469a96ae46f3f7ccd0e94f159b398542

lentzi90:
Alright it seems to be working now.
I first tried the same as for the fixture tests (using an action to setup go). Unfortunately, our `ensure-go.sh` script did not seem to pick that up properly. I think the issue is that it gets confused by multiple installed versions (one from the action and one pre-installed on the runner).
So instead I ended up deleting the pre-installed go binary and switch the PATH so that this is preferred above everything else.

PR #1745, opened 2024-05-22
:seedling: explicitly drop pr approver workflow permissions

tuminoid:
Drop all permissions on top-level. We're already giving out actions: write on the job level.

This is noted in the OpenSSF scorecard as well.


lentzi90 (commented):
/approve

PR #1747, opened 2024-05-24
🌱 remove kubesec workflow

tuminoid:
Kubesec action has not been maintained for years and is broken. Kubesec itself is maintained, and we'll see if we reintroduce it as prow test instead of action later, or if we have enough coverage from existing tools.


lentzi90 (commented):
/lgtm

PR #1753, opened 2024-05-30
:seedling: Remove hack/tools go module

mquhuy:
The hack/tools go module seems to exist just so we could install `kustomize` and `controller-gen`. These can be installed now with `go install`, and we have already installed `golangci_lint` that way.


lentzi90 (commented):
/lgtm

mquhuy:
/override metal3-bmo-e2e-test-pull
This is not relevant

smoshiur1237 (commented):
LGTM

honza:
The advantage of our current approach is that the tooling is bundled together with the software and this is especially helpful in scenarios where vendoring is required.  Everything can be built with what's already available in the repo and without downloading further packages and installing more tools.

hack/verify-release.sh
tuminoid, line 1: On L492 there is a comment about hack/tools, remove that as well.

Makefile
```diff
@@ -21,8 +21,13 @@ TOOLS_DIR := hack/tools
 TOOLS_BIN_DIR := $(abspath $(TOOLS_DIR)/$(BIN_DIR))
```
tuminoid, line 21: TOOLS_DIR on L20 isn't used for anything else than this definition, remove it.

kashifest:
> The advantage of our current approach is that the tooling is bundled together with the software and this is especially helpful in scenarios where vendoring is required. Everything can be built with what's already available in the repo and without downloading further packages and installing more tools.

@honza do I understand correctly that you have use cases for this for vendoring downstream ?

tuminoid:
/hold

Let's finish discussion on vendoring.

lentzi90:
@honza I understand the vendoring use-case, but I wonder if you actually need to vendor hack/tools? It seems to me that [golangci-lint](https://github.com/metal3-io/baremetal-operator/blob/7a329fe76c40db3b61a250bec647ecc5d15a545d/Makefile#L133-L134) already breaks this anyway.

honza:
We build our images in CI without internet access so this would break us.  I'm happy to work around this but I just wanted to point out the implications in case others had similar requirements.

mquhuy:
Hi @honza, we have discussed a bit and decided that we will not merge this PR for the time being. Could you please let us know if/when you no longer need this module? :)

honza:
> Could you please let us know if/when you no longer need this module? :)

Sure thing, thanks!

Rozzii:
/remove-lifecycle stale
FYI @honza

Rozzii:
/remove-lifecycle stale
@mquhuy @honza

Rozzii:
@honza do you still need this module?
/remove-lifecycle stale

PR #1754, opened 2024-05-30
🌱 shellcheck only files

tuminoid:
Make shellcheck consider only files.

If branch name ended with .sh or .sh/something, it was found on the filesystem as something ending with .sh, and shellcheck picked it up, and linting failed obviously.


kashifest (commented):
/lgtm

PR #1765, opened 2024-06-04
🌱 Add periodic jobs for optional tests and release branches

lentzi90:

This adds more github workflows, namely periodic jobs for the release branches with the required e2e tests and periodic jobs for the optional e2e tests for main and release branches.

Fixes #


.github/workflows/e2e-test.yml
```diff
@@ -31,12 +39,13 @@ jobs:
...
       env:
         BMC_PROTOCOL: ${{ inputs.bmc-protocol }}
+        GINKGO_FOCUS: "${{ inputs.ginkgo-focus }}"
       # We need a new shell to pick up the new group. That is why we do the sudo -s -u $USER ...
       # Remove the pre-installed go version. We install the exact version we need.
       run: |
         sudo usermod -a -G libvirt $USER
         sudo rm /usr/bin/go
-        sudo -s -u $USER bash ${{ github.workspace }}/hack/ci-e2e.sh
+        sudo -s -u $USER --preserve-env bash ${{ github.workspace }}/hack/ci-e2e.sh
```
lentzi90, line 48: Note: This was a mistake in the previous PR. The env vars were not passed through. By adding --preserve-env we make sure that the environment variables really are available in the script.

.github/workflows/e2e-test-optional-periodic.yml
```diff
@@ -0,0 +1,27 @@
+name: Periodic E2E Test Optional
+
+on:
+  schedule:
+    # Run every day at 04:20 UTC (it is recommended to avoid running at the start of the hour)
+    - cron: '20 4 * * *'
+  pull_request:
```
kashifest, line 7: This would mean that the periodic would run on PRs as well as we see in the current run, I dont think we want that.

/hold

lentzi90: Thanks, good catch! This was in a test commit to make sure that they work. I managed to forget it :sweat: 
Dropping it now an repushing

lentzi90:
Test commit removed.

kashifest:
/hold cancel
looks good now

PR #1767, opened 2024-06-04
:seedling: Add new go library to create new bmc and bmh

### LINKED ISSUE #1560: E2E: Make go tool for creating BMH VMs
**User Story**

As a developer I would like to have a go module for creating virtualized BMHs, so that this can be easily reused and consumed between repos and scripts.

**Detailed Description**

Similar to [make-virt-host](https://github.com/metal3-io/baremetal-operator/tree/main/cmd/make-virt-host), convert the [scripts](https://github.com/metal3-io/baremetal-operator/blob/main/hack/create_bmcs.sh) that we use into a go tool that can be reused for e2e tests, tilt setup and (in the future) CAPM3. The currently used scripts work well but are tricky to reuse. It is doable for e2e and tilt so far but it would be much nicer to have a proper go module for it. That module could then be used also by projects building on BMO, e.g. CAPM3, not to mention downstream projects and "testing environments".

At a high level, I want to be able to
1. `go install` a binary for creating VMs that can be used as BMHs
2. import a go module to do the same in go code
3. specify some basic parameters for how to configure the VMs (amount of VMs to create, CPU, RAM, disk size, MAC, etc). Some kind of config file would probably make sense for this.

I think this is the library we should use: https://gitlab.com/libvirt/libvirt-go-module

/kind feature

mquhuy:
Fixes #1560


test/createVM/main.go
```diff
@@ -0,0 +1,298 @@
...
+	"bytes"
+	"embed"
+	"flag"
+	"fmt"
+	"os"
+	"text/template"
+
+	"github.com/dypflying/go-qcow2lib/qcow2"
+	bmoe2e "github.com/metal3-io/baremetal-operator/test/e2e"
+	"libvirt.org/go/libvirt"
```
Rozzii, line 13: Could you please use test this syntax?:
 
```suggestion
	"github.com/libvirt/libvirt-go-module"
```

test/go.mod
```diff
@@ -15,6 +16,7 @@ require (
 	k8s.io/apimachinery v0.29.3
 	k8s.io/klog/v2 v2.110.1
 	k8s.io/utils v0.0.0-20231127182322-b307cd553661
+	libvirt.org/go/libvirt v1.10003.0
```
Rozzii, line 19:
```suggestion
	github.com/libvirt/libvirt-go-module v1.10003.0
```

test/go.sum
```diff
@@ -487,6 +489,8 @@ k8s.io/kube-openapi v0.0.0-20231010175941-2dd684a91f00 h1:aVUu9fTY98ivBPKR9Y5w/A
 k8s.io/kube-openapi v0.0.0-20231010175941-2dd684a91f00/go.mod h1:AsvuZPBlUDVuCdzJ87iajxtXuR9oktsTctW/R9wwouA=
 k8s.io/utils v0.0.0-20231127182322-b307cd553661 h1:FepOBzJ0GXm8t0su67ln2wAZjbQ6RxQGZDnzuLcrUTI=
 k8s.io/utils v0.0.0-20231127182322-b307cd553661/go.mod h1:OLgZIPagt7ERELqWJFomSt595RzquPNLL48iOWgYOg0=
+libvirt.org/go/libvirt v1.10003.0 h1:LEoawzuggD6IL5R/XtnBE8wWJx49i7UZ1HcB7p9glwE=
+libvirt.org/go/libvirt v1.10003.0/go.mod h1:1WiFE8EjZfq+FCVog+rvr1yatKbKZ9FaFMZgEqxEJqQ=
```
Rozzii, line 493:
```suggestion
github.com/libvirt/libvirt-go-module v1.10003.0 h1:LEoawzuggD6IL5R/XtnBE8wWJx49i7UZ1HcB7p9glwE=
github.com/libvirt/libvirt-go-module v1.10003.0/go.mod h1:1WiFE8EjZfq+FCVog+rvr1yatKbKZ9FaFMZgEqxEJqQ=
```

Rozzii (changes requested):
Could we try this syntax? Might solve the automatic download issue.

test/go.sum
mquhuy: I tried and that syntax doesn't work because the module name is `libvirt.org/go/libvirt` (as defined by https://github.com/libvirt/libvirt-go-module/blob/master/go.mod#L1)

```
go: github.com/libvirt/libvirt-go-module@upgrade (v1.10003.0) requires github.com/libvirt/libvirt-go-module@v1.10003.0: parsing go.mod:
        module declares its path as: libvirt.org/go/libvirt
                but was required as: github.com/libvirt/libvirt-go-module
```

Even if the syntax is ok, that wouldn't change the module content. The github module is just a read-only fork of the gitlab one.

mquhuy:
> Could we try this syntax? Might solve the automatic download issue.

I tried and that syntax doesn't work because the module name is libvirt.org/go/libvirt (as defined by https://github.com/libvirt/libvirt-go-module/blob/master/go.mod#L1)

go: github.com/libvirt/libvirt-go-module@upgrade (v1.10003.0) requires github.com/libvirt/libvirt-go-module@v1.10003.0: parsing go.mod:
        module declares its path as: libvirt.org/go/libvirt
                but was required as: github.com/libvirt/libvirt-go-module
Even if the syntax is ok, that wouldn't change the module content. The github module is just a read-only fork of the gitlab one.

Rozzii:
> > Could we try this syntax? Might solve the automatic download issue.
> 
> Sorry, not following, what automatic download issue are we talking about? I tried and that syntax didn't work because the module name is libvirt.org/go/libvirt (as defined by https://github.com/libvirt/libvirt-go-module/blob/master/go.mod#L1)
> 
> ```
> go: github.com/libvirt/libvirt-go-module@upgrade (v1.10003.0) requires github.com/libvirt/libvirt-go-module@v1.10003.0: parsing go.mod:
>         module declares its path as: libvirt.org/go/libvirt
>                 but was required as: github.com/libvirt/libvirt-go-module
> ```
> 
> Even if the syntax is ok, that wouldn't change the module content. The github module is just a read-only fork of the gitlab one.

I don't want to change the content I want to solve the issue with the dependency.
If you check the error logs you can see that a bunch of dependencies are being pulled  and I don't know why the libvirt is not pulled and discovered automatically. You tried to create a custom image just to have this library but we never do it for any go library go toolchain supposed to download them on it's own. So I thought it has some issue resolving the "libvirt.org/go/libvirt" but I was clearly wrong.
image

Rozzii:
> > > Could we try this syntax? Might solve the automatic download issue.
> > 
> > 
> > Sorry, not following, what automatic download issue are we talking about? I tried and that syntax didn't work because the module name is libvirt.org/go/libvirt (as defined by https://github.com/libvirt/libvirt-go-module/blob/master/go.mod#L1)
> > ```
> > go: github.com/libvirt/libvirt-go-module@upgrade (v1.10003.0) requires github.com/libvirt/libvirt-go-module@v1.10003.0: parsing go.mod:
> >         module declares its path as: libvirt.org/go/libvirt
> >                 but was required as: github.com/libvirt/libvirt-go-module
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Even if the syntax is ok, that wouldn't change the module content. The github module is just a read-only fork of the gitlab one.
> 
> I don't want to change the content I want to solve the issue with the dependency. If you check the error logs you can see that a bunch of dependencies are being pulled and I don't know why the libvirt is not pulled and discovered automatically. You tried to create a custom image just to have this library but we never do it for any go library go toolchain supposed to download them on it's own. So I thought it has some issue resolving the "libvirt.org/go/libvirt" but I was clearly wrong. image The golang ci linter says this> `could not import libvirt.org/go/libvirt (-: # libvirt.org/go/libvirt` and this was that started to make me think that this is just some path issue or something but looks like simply changing the url wouldn't help.

Okay, never mind.... we  are not using just plain go generate we use controller-gen

Rozzii:
Okay I have now get where go-libvirt-module specifies the distro package dependencies..... I am getting closer and closer to the opinion that we should just have a custom go toolchain container where we can add all these dependencies similar thing could happen in the future for other go libraries too...

kashifest:
/approve
I liked the overall idea hence and since the tests passed I am approving. I couldn't walk through every tiny details in the code hence I am putting a hold. Please unhold when you theres a couple of more reviews or you have lgtm. 

/hold

Rozzii (commented):
/lgtm

Rozzii (approved):

PR #1770, opened 2024-06-05
🌱 Workflow pipeline with unit tests

lentzi90:
**What this PR does / why we need it**:

This adds a unit test workflow and restructures the github workflows
so that we run golangci-lint and unit tests first. Only when they have
succeded will the pipeline continue with e2e tests.
The goal is to conserve resources and avoid running pointless jobs.
The commit also removes unnecessary permissions and avoids running the
jobs if the PR is a draft.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

PR #1772, opened 2024-06-05
🌱 [backport] Workflow pipeline with unit tests

lentzi90:
Manual backport of https://github.com/metal3-io/baremetal-operator/pull/1770

This adds a unit test workflow and restructures the github workflows so that we run golangci-lint and unit tests first. Only when they have succeded will the pipeline continue with e2e tests. The goal is to conserve resources and avoid running pointless jobs. The commit also removes unnecessary permissions and avoids running the jobs if the PR is a draft.


kashifest (commented):
/approve

PR #1776, opened 2024-06-08
🌱 Fix Set image tag for release branches and tags

### LINKED ISSUE #1722: Publish release artifacts with manifests

As seen in https://github.com/metal3-io/metal3-docs/pull/414 we are not setting the image tag in the [manifests](https://github.com/metal3-io/baremetal-operator/blob/d3376d8eba3e224a3c440d4009d3596aa37a6906/config/base/manager.yaml#L26). This is fine for the main branch, but it is not ideal for release branches and tags.
~~I think we should set the image tag always on release branches to the same as the release branch and for tags to the exact tag.
Otherwise a user trying to apply manifests from `release-0.6` for example, will end up with the latest container image from `main`.~~

We should publishing release artifacts where the correct image tag is used. Currently we have [a default config](https://github.com/metal3-io/baremetal-operator/blob/main/config/default/ironic.env) mixed in with the manifests. This does not make sense to include in the release artifacts IMO. We should handle it the same as [the secrets](https://github.com/metal3-io/baremetal-operator/blob/61c9dc108e7ab592b47702a5f05989d9bec5d450/config/components/basic-auth/kustomization.yaml#L4-L16). I.e. the manifests expect the configmap to exist but it is up to the user to create/add it in some way.

With that fixed I think we can do release artifacts and set the image there.

When using manifests from a specific branch or tag, the corresponding container image should be used.
We should probably include this in the release process.

/kind bug

7h3-3mp7y-m4n:
✨ 

This PR solves image tags, by adding a verification check and also setting the tag automatically 

**Which issue(s) this PR fixes
Fixes #1722

### DIFF
--- a/docs/releasing.md
+++ b/docs/releasing.md
@@ -57,7 +57,10 @@ or if using existing repository, verify your intended remote is set to
   `git push origin release-0.x` to create it
 - If creating a new patch release, use existing branch `release-0.x`:
   `git checkout origin/release-0.x`
-
+- Before creating the release tag, update the image tag in the manifests to match the release version. Run the following command, replacing `<RELEASE_TAG>` with the appropriate tag.
+```bash
+make set-manifest-image-bmo TAG=<RELEASE_TAG> 
+```
 ### Tags
 
 First we create a primary release tag, that triggers release note creation and
--- a/hack/verify-release.sh
+++ b/hack/verify-release.sh
@@ -193,6 +193,34 @@ _version_check()
     [[ "${min_version}" == $(echo -e "${min_version}\n${version}" | sort -s -t. -k 1,1 -k 2,2n -k 3,3n | head -n1) ]]
 }
 
+check_image_tag() {
+    if [ -z "$1" ]; then
+        echo "Error: No image tag provided"
+        exit 1
+    fi
+    TAG=$1
+    MANIFEST_FILES=$(find template -type f -name '*.yaml')
+    ERROR=0
+
+    echo "Checking the image tag :TAG"
+
+    for FILE in $MANIFEST_FILES; do
+        if ! grep -q "image: .*/baremetal-operator:$TAG" "$FILE"; then
+            echo "Error: Image tag in $FILE is not set to $TAG"
+            ERROR=1
+        fi
+    done
+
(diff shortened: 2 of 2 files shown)


tuminoid:
/ok-to-test

Thanks for the PR! Please set the PR type to `:seedling` in the title. I'll review this shortly.

7h3-3mp7y-m4n:
Sorry for the deal, as I see `markdownlint` and `shellcheck` are the recheck ones, I will raise a PR soon with a fix.

7h3-3mp7y-m4n:
@tuminoid  can you tell me what I am missing here? cause I'm confused

tuminoid:
Let's run a retest so we can have fresh logs, the old logs have expired already.

/retest

tuminoid:
> @tuminoid can you tell me what I am missing here? cause I'm confused

Please check the details links of shellcheck and markdownlint for details what is failing.

For the markdownlink checker failure, I'm not sure what that is complaining about as it just says "action failed" with no logs, I'll look into that.

7h3-3mp7y-m4n:
> > @tuminoid can you tell me what I am missing here? cause I'm confused
> 
> Please check the details links of shellcheck and markdownlint for details what is failing.
> 
> For the markdownlink checker failure, I'm not sure what that is complaining about as it just says "action failed" with no logs, I'll look into that.

Thanks @tuminoid  I'll look into that

7h3-3mp7y-m4n:
oh, my bad,  I'll fix it , I thought that it has been merged , but my bad , I'll fix the `ci` fails

7h3-3mp7y-m4n:
> I actually wrote some review comments here, and then realized this is not the right thing to do, and discussed with @kashifest and @adilGhaffarDev what we should actually be doing. We don't want to have them committed to the repo, but we should set the tags in release artifacts. So I filed [issue](https://github.com/metal3-io/baremetal-operator/issues/2089) to actually start including some release artifacts in BMO, such as manifests, where we do need to set the version.
> 
> Let's continue the discussion there what we need to include, especially considering IRSO will take Ironic manifests out of BMO shortly.
> 
> This PR is not going to be merged, but thanks @7h3-3mp7y-m4n for raising the issue. /hold

for sure!

Issue #1785, opened 2024-06-14
All Metal3 Centos E2E main tests fail with #1685


tuminoid:
**What steps did you take and what happened:**
https://github.com/metal3-io/baremetal-operator/pull/1685 was merged, and since then all Metal3 Centos based e2e tests on main branch have failed. If the PR is reverted, they work.

**What did you expect to happen:**
Centos e2e succeeds.

**Anything else you would like to add:**
Ubuntu variants pass (given that #1780 is merged to fix one issue), so this is isolated to Centos. 

**Environment:**
Dev-env / CI, e2e integration, e2e feature, e2e ephemeral, bml e2e periodics all fail.
All PR jobs with centos-e2e-integration-main fail

See https://jenkins.nordix.org/view/Metal3%20Periodic/job/metal3-periodic-centos-e2e-integration-test-main/87/ or any other periodic centos main job.

- Baremetal Operator version: main
- Environment (metal3-dev-env or other): dev-env / CI

/kind bug

tuminoid:
/triage accepted

/cc @dtantsur @elfosardo @MahnoorAsghar @mboukhalfa @Rozzii @kashifest 
FYI

tuminoid:
Notable difference in BMO logs is 

```
"level":"info","ts":1718358988.3083067,"logger":"provisioner.ironic","msg":"error caught while checking endpoint, will retry","host":"metal3~node-0","endpoint":"https://172.22.0.2:6385/v1/","error":"Expected HTTP response code [200 300] when accessing [GET https://172.22.0.2:6385/v1/], but got 503 instead: <!DOCTYPE HTML PUBLIC \"-//IETF//DTD HTML 2.0//EN\">\n<html><head>\n<title>503 Service Unavailable</title>\n</head><body>\n<h1>Service Unavailable</h1>\n<p>The server is temporarily unable to service your\nrequest due to maintenance downtime or capacity\nproblems. Please try again later.</p>\n</body></html>"}
{"level":"info","ts":1718358988.3096807,"logger":"controllers.BareMetalHost","msg":"provisioner is not ready","baremetalhost":{"name":"node-0","namespace":"metal3"},"RequeueAfter:":30}
[...]

PR #1786, opened 2024-06-14
:bug: Fix incorrect htpasswd file generation

Rozzii:
This commit:
  - removes the unnecessary parts of the htpasswd file generation and clears up the content of the file

The reason:
   - The original command was designed to generate an env variable but the new ironic-image logic expects the content  to be suitable for a htpasswd file this the variable name and the "=" is not needed.

Fixes: #1785


tuminoid (commented):
It worked on my manual test :+1: 

/lgtm

kashifest:
/approve
thanks for the fix and debugging @Rozzii @tuminoid

PR #1790, opened 2024-06-19
🌱 limit running BMO e2e periodics to metal3-io repo

tuminoid:
Otherwise workflows try to run on forks as well, which might not have CNCF (or any matching runners) and they would just keep scheduling and eventually after 30 hours get cancelled and maintainers get spammed.


kashifest (commented):
Absolutely makes sense. 
/approve
/lgtm

PR #1791, opened 2024-06-19
:seedling: Always upload test artifacts, even if the tests fail

mquhuy:
By default, if a step in github action fails, the subsequent steps won't be run. This is not ideal in case of running e2e tests, as we would like the tests artifacts to be uploaded, even if the tests fail.

This PR enables test artifacts upload for failed tests.

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -45,6 +45,7 @@ jobs:
         sudo -s -u $USER --preserve-env bash ${{ github.workspace }}/hack/ci-e2e.sh
 
     - name: Upload artifacts
+      if: ${{ !cancelled() }}
       uses: actions/upload-artifact@65462800fd760344b1a7b4382951275a0abb4808 # v4.3.3
       with:
         name: artifacts-${{ inputs.bmc-protocol }}.tar.gz


.github/workflows/e2e-test.yml
```diff
@@ -45,6 +45,7 @@ jobs:
         sudo -s -u $USER --preserve-env bash ${{ github.workspace }}/hack/ci-e2e.sh
 
     - name: Upload artifacts
+      if: always()
```
tuminoid, line 48: Per https://docs.github.com/en/actions/learn-github-actions/expressions#status-check-functions
it should be

```suggestion
      if: ${{ always() }}
```
?

mquhuy: Well, I'm pretty sure `always()` will work too, but as the syntax was stated like that, why not follow?

Rozzii (commented):
/lgtm

.github/workflows/e2e-test.yml
tuminoid: I can see few of those too in the ecosystem, but most are using the syntax with curlys, like in the documentation, and in Metal3 we have used the curly syntax, so good for consistency as well.

tuminoid (approved):
`!cancelled()` is better than `always()`, good addition.

Rozzii (commented):
/lgtm

kashifest (commented):
/approve

PR #1793, opened 2024-06-25
🐛  Fix HFC to execute updates

iurygregory:
Currently any changes in HostFirmwareComponents won't trigger the firmware update on it.
This commit aims to fix the issues we have observed while testing with real hardware.

**What this PR does / why we need it**: metal3-io/metal3-docs#364

### DIFF
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -1164,6 +1164,20 @@ func (r *BareMetalHostReconciler) actionPreparing(prov provisioner.Provisioner, 
 		return recordActionFailure(info, metal3api.PreparationError, provResult.ErrorMessage)
 	}
 
+	if hfcDirty && started {
+		hfcStillDirty, err := r.saveHostFirmwareComponents(prov, info, hfc)
+		if err != nil {
+			return actionError{errors.Wrap(err, "could not save the host firmware components")}
+		}
+
+		if hfcStillDirty {
+			info.log.Info("going to update the host firmware components")
+			if err := r.Status().Update(info.ctx, hfc); err != nil {
+				return actionError{errors.Wrap(err, "failed to update hostfirmwarecomponents status")}
+			}
+		}
+	}
+
 	if bmhDirty && started {
 		info.log.Info("saving host provisioning settings")
 		_, err := saveHostProvisioningSettings(info.host, info)
@@ -1740,6 +1754,32 @@ func saveHostProvisioningSettings(host *metal3api.BareMetalHost, info *reconcile
 	return dirty, nil
 }
 
+func (r *BareMetalHostReconciler) saveHostFirmwareComponents(prov provisioner.Provisioner, info *reconcileInfo, hfc *metal3api.HostFirmwareComponents) (dirty bool, err error) {
+	dirty = false
+	if reflect.DeepEqual(hfc.Status.Updates, hfc.Spec.Updates) {
+		info.log.Info("not saving hostFirmwareComponents information since is not necessary")
+		return dirty, nil
+	}
+
+	info.log.Info("saving hostFirmwareComponents information", "spec updates", hfc.Spec.Updates, "status updates", hfc.Status.Updates)
+
+	hfc.Status.Updates = make([]metal3api.FirmwareUpdate, len(hfc.Spec.Updates))
+	for i := range hfc.Spec.Updates {
+		hfc.Spec.Updates[i].DeepCopyInto(&hfc.Status.Updates[i])
+	}
+
+	// Retrieve new information about the firmware components stored in ironic
+	components, err := prov.GetFirmwareComponents()
+	if err != nil {
+		info.log.Error(err, "failed to get new information for firmware components in ironic")
+		return dirty, err
+	}
+	hfc.Status.Components = components
+	dirty = true
+
+	return dirty, nil
+}
+
 func (r *BareMetalHostReconciler) createHostFirmwareComponents(info *reconcileInfo) error {
 	// Check if HostFirmwareComponents already exists
 	hfc := &metal3api.HostFirmwareComponents{}
@@ -1755,18 +1795,31 @@ func (r *BareMetalHostReconciler) createHostFirmwareComponents(info *reconcileIn
 
 			// Set bmh as owner, this makes sure the resource is deleted when bmh is deleted
 			if err = controllerutil.SetControllerReference(info.host, hfc, r.Scheme()); err != nil {
-				return errors.Wrap(err, "could not set bmh as controller")
+				return errors.Wrap(err, "could not set bmh as controller for hostFirmwareComponents")
 			}
 			if err = r.Create(info.ctx, hfc); err != nil {
 				return errors.Wrap(err, "failure creating hostFirmwareComponents resource")
 			}
 
 			info.log.Info("created new hostFirmwareComponents resource")
-		} else {
-			// Error reading the object
-			return errors.Wrap(err, "could not load hostFirmwareComponents resource")
+			return nil
 		}
+		// Error reading the object
+		return errors.Wrap(err, "could not load hostFirmwareComponents resource")
 	}
+	// Necessary in case the CRD is created manually.
+
+	if !ownerReferenceExists(info.host, hfc) {
+		if err := controllerutil.SetControllerReference(info.host, hfc, r.Scheme()); err != nil {
+			return errors.Wrap(err, "could not set bmh as controller for hostFirmwareComponents")
+		}
+		if err := r.Update(info.ctx, hfc); err != nil {
+			return errors.Wrap(err, "failure updating hostFirmwareComponents resource")
+		}
+
+		return nil
+	}
+
 	return nil
 }
 
@@ -1851,15 +1904,6 @@ func (r *BareMetalHostReconciler) getHostFirmwareComponents(info *reconcileInfo)
 
 	// Check if there are Updates in the Spec that are different than the Status
 	if meta.IsStatusConditionTrue(hfc.Status.Conditions, string(metal3api.HostFirmwareComponentsChangeDetected)) {
-		// Check if the status have been populated
-		if len(hfc.Status.Updates) == 0 {
-			return false, nil, errors.New("host firmware status updates not available")
-		}
-
-		if len(hfc.Status.Components) == 0 {
-			return false, nil, errors.New("host firmware status components not available")
-		}
-
 		if meta.IsStatusConditionTrue(hfc.Status.Conditions, string(metal3api.HostFirmwareComponentsValid)) {
 			info.log.Info("hostFirmwareComponents indicating ChangeDetected", "namespacename", info.request.NamespacedName)
 			return true, hfc, nil
--- a/apis/metal3.io/v1alpha1/hostfirmwarecomponents_types.go
+++ b/apis/metal3.io/v1alpha1/hostfirmwarecomponents_types.go
@@ -57,7 +57,7 @@ type HostFirmwareComponentsStatus struct {
 	// Updates is the list of all firmware components that should be updated
 	// they are specified via name and url fields.
 	// +optional
-	Updates []FirmwareUpdate `json:"updates"`
+	Updates []FirmwareUpdate `json:"updates,omitempty"`
 
 	// Components is the list of all available firmware components and their information.
 	Components []FirmwareComponentStatus `json:"components,omitempty"`
--- a/controllers/metal3.io/host_state_machine.go
+++ b/controllers/metal3.io/host_state_machine.go
@@ -465,6 +465,14 @@ func (hsm *hostStateMachine) handleAvailable(info *reconcileInfo) actionResult {
 		return actionComplete{}
 	}
 
+	// Check if hostFirmwareComponents have changed
+	if dirty, _, err := hsm.Reconciler.getHostFirmwareComponents(info); err != nil {
+		return actionError{err}
+	} else if dirty {
+		hsm.NextState = metal3api.StatePreparing
+		return actionComplete{}
+	}
+
 	// ErrorCount is cleared when appropriate inside actionManageAvailable
 	actResult := hsm.Reconciler.actionManageAvailable(hsm.Provisioner, info)
 	if _, complete := actResult.(actionComplete); complete {
--- a/controllers/metal3.io/hostfirmwarecomponents_controller.go
+++ b/controllers/metal3.io/hostfirmwarecomponents_controller.go
@@ -122,7 +122,6 @@ func (r *HostFirmwareComponentsReconciler) Reconcile(ctx context.Context, req ct
 
 	// Fetch the HostFirmwareComponents
 	hfc := &metal3api.HostFirmwareComponents{}
-	info := &rhfcInfo{ctx: ctx, log: reqLogger, hfc: hfc, bmh: bmh}
 	if err = r.Get(ctx, req.NamespacedName, hfc); err != nil {
 		// The HFC resource may have been deleted
 		if k8serrors.IsNotFound(err) {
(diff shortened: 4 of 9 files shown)


tuminoid:
We probably want to to cherry-pick this to release-0.6 when done, if HFC doesn't work?

iurygregory:
@tuminoid correct

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1740,6 +1755,34 @@ func saveHostProvisioningSettings(host *metal3api.BareMetalHost, info *reconcile
 	return dirty, nil
 }
 
+func (r *BareMetalHostReconciler) saveHostFirmwareComponents(prov provisioner.Provisioner, info *reconcileInfo, hfc *metal3api.HostFirmwareComponents) (dirty bool, err error) {
+	dirty = false
+	if !reflect.DeepEqual(hfc.Status.Updates, hfc.Spec.Updates) {
```
dtantsur, line 1760: Let's have an early return here instead of a huge block.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1164,6 +1164,21 @@ func (r *BareMetalHostReconciler) actionPreparing(prov provisioner.Provisioner,
 		return recordActionFailure(info, metal3api.PreparationError, provResult.ErrorMessage)
 	}
 
+	if hfcDirty && started {
+		info.log.Info("saving host firmware components")
```
dtantsur, line 1168: This is misleading: we don't yet know if we're going to save anything.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1740,6 +1755,34 @@ func saveHostProvisioningSettings(host *metal3api.BareMetalHost, info *reconcile
 	return dirty, nil
 }
 
+func (r *BareMetalHostReconciler) saveHostFirmwareComponents(prov provisioner.Provisioner, info *reconcileInfo, hfc *metal3api.HostFirmwareComponents) (dirty bool, err error) {
+	dirty = false
+	if !reflect.DeepEqual(hfc.Status.Updates, hfc.Spec.Updates) {
+		info.log.Info("HostFirmwareComponents Status", "updates", hfc.Status.Updates)
```
dtantsur, line 1761: If you want to log something here, let's log both old and new updates, as well as provide a clearer message.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1740,6 +1755,34 @@ func saveHostProvisioningSettings(host *metal3api.BareMetalHost, info *reconcile
...
+
+		hfc.Status.Updates = make([]metal3api.FirmwareUpdate, len(hfc.Spec.Updates))
+		for i := range hfc.Spec.Updates {
+			hfc.Spec.Updates[i].DeepCopyInto(&hfc.Status.Updates[i])
+		}
+
+		info.log.Info("re assigned status updates")
```
dtantsur, line 1768: Please remove

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1740,6 +1755,34 @@ func saveHostProvisioningSettings(host *metal3api.BareMetalHost, info *reconcile
...
+		info.log.Info("New HostFirmwareComponents Status", "updates", hfc.Status.Updates)
```
dtantsur, line 1769: If you address my comment on line 1761, this will not be needed.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1740,6 +1755,34 @@ func saveHostProvisioningSettings(host *metal3api.BareMetalHost, info *reconcile
...
+
+		// Retrieve new information about the firmware components stored in ironic
+		components, err := prov.GetFirmwareComponents()
+		if err != nil {
+			info.log.Info("Failed to get new information for firmware components in ironic")
```
dtantsur, line 1774: This is not an Info, it's an Error

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1740,6 +1755,34 @@ func saveHostProvisioningSettings(host *metal3api.BareMetalHost, info *reconcile
...
+			return dirty, err
+		}
+		hfc.Status.Components = components
+		t := metav1.Now()
```
dtantsur, line 1778: Per Zane's comment, it's not needed.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1740,6 +1755,34 @@ func saveHostProvisioningSettings(host *metal3api.BareMetalHost, info *reconcile
...
+		hfc.Status.LastUpdated = &t
+		info.log.Info("Updates for HostFirmwareComponents have changed")
```
dtantsur, line 1780: Redundant

dtantsur (changes requested):

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1755,17 +1795,24 @@ func (r *BareMetalHostReconciler) createHostFirmwareComponents(info *reconcileIn
...
+			return nil
 		}
+		// Error reading the object
+		return errors.Wrap(err, "could not load hostFirmwareComponents resource")
+	}
+	// Necessary in case the CRD is manually created.
+	if err := controllerutil.SetControllerReference(info.host, hfc, r.Scheme()); err != nil {
+		return errors.Wrap(err, "could not set bmh as controller for hostFirmwareComponents")
+	}
+	if err := r.Update(info.ctx, hfc); err != nil {
```
dtantsur, line 1814: Is it going to get called on every reconciliation?

iurygregory: you are probably right, let me do a test  having the Update condition nested with the previous if in case the err is nil

iurygregory: Done, I just tested the change.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1740,6 +1754,32 @@ func saveHostProvisioningSettings(host *metal3api.BareMetalHost, info *reconcile
 	return dirty, nil
 }
 
+func (r *BareMetalHostReconciler) saveHostFirmwareComponents(prov provisioner.Provisioner, info *reconcileInfo, hfc *metal3api.HostFirmwareComponents) (dirty bool, err error) {
+	dirty = false
+	if reflect.DeepEqual(hfc.Status.Updates, hfc.Spec.Updates) {
+		info.log.Info("Not Saving HostFirmwareComponents Information since is not necessary")
```
dtantsur, line 1760: nit: inconsistent case in log messages (I think we use lower case except for CRD's - check other examples)

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1755,18 +1795,28 @@ func (r *BareMetalHostReconciler) createHostFirmwareComponents(info *reconcileIn
...
-			return errors.Wrap(err, "could not load hostFirmwareComponents resource")
+			return nil
 		}
+		// Error reading the object
+		return errors.Wrap(err, "could not load hostFirmwareComponents resource")
 	}
+	// Necessary in case the CRD is created manually.
+	err := controllerutil.SetControllerReference(info.host, hfc, r.Scheme())
+	if err == nil {
+		if err := r.Update(info.ctx, hfc); err != nil {
```
dtantsur, line 1813: I still feel somewhat uneasy calling Update every time. I see other places use `if !ownerReferenceExists(...)`, maybe use that as well?

iurygregory: oh like we have for DataImage, I will give a try and test in my setup

controllers/metal3.io/baremetalhost_controller.go
iurygregory: I did that because of  saveHostProvisioningSettings , I will push a change to address.

iurygregory:
Done, I've tested again in my setup before pushing the changes here.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1755,18 +1795,31 @@ func (r *BareMetalHostReconciler) createHostFirmwareComponents(info *reconcileIn
 
 			// Set bmh as owner, this makes sure the resource is deleted when bmh is deleted
 			if err = controllerutil.SetControllerReference(info.host, hfc, r.Scheme()); err != nil {
```
zaneb, line 1797: Controller references are exclusive, and almost never what you want. Is there a reason we are doing this instead of a regular owner reference?

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1755,18 +1795,31 @@ func (r *BareMetalHostReconciler) createHostFirmwareComponents(info *reconcileIn
...
-			return errors.Wrap(err, "could not load hostFirmwareComponents resource")
+			return nil
 		}
+		// Error reading the object
+		return errors.Wrap(err, "could not load hostFirmwareComponents resource")
 	}
+	// Necessary in case the CRD is created manually.
+
+	if !ownerReferenceExists(info.host, hfc) {
+		if err := controllerutil.SetControllerReference(info.host, hfc, r.Scheme()); err != nil {
```
zaneb, line 1813: see above

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1851,15 +1904,6 @@ func (r *BareMetalHostReconciler) getHostFirmwareComponents(info *reconcileInfo)
 
 	// Check if there are Updates in the Spec that are different than the Status
 	if meta.IsStatusConditionTrue(hfc.Status.Conditions, string(metal3api.HostFirmwareComponentsChangeDetected)) {
```
zaneb, line 1906: In theory we should check that the condition matches the current Generation ([example](https://github.com/metal3-io/baremetal-operator/blob/main/controllers/metal3.io/baremetalhost_controller.go#L670)) so we know the data is not out of date.
Not a huge deal though. (Bigger worry is actually that we miss that the Spec has been copied to the Status - which won't bump the generation - which would result in us doing the update again.)

zaneb (commented):
I left some comments, but clearly this is a huge improvement so it may be worth merging and following up on those afterwards.

controllers/metal3.io/baremetalhost_controller.go
dtantsur: An HFC object is tightly coupled with a BareMetalHost (it's the same Ironic Node), so I think it's justified.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1740,6 +1754,32 @@ func saveHostProvisioningSettings(host *metal3api.BareMetalHost, info *reconcile
...
+
+	info.log.Info("saving hostFirmwareComponents information", "spec updates", hfc.Spec.Updates, "status updates", hfc.Status.Updates)
+
+	hfc.Status.Updates = make([]metal3api.FirmwareUpdate, len(hfc.Spec.Updates))
+	for i := range hfc.Spec.Updates {
+		hfc.Spec.Updates[i].DeepCopyInto(&hfc.Status.Updates[i])
+	}
+
+	// Retrieve new information about the firmware components stored in ironic
+	components, err := prov.GetFirmwareComponents()
```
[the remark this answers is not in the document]
iurygregory: yeah, this is the problem I'm planning to fix in a separate PR, still trying to figure out how

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1164,6 +1164,20 @@ func (r *BareMetalHostReconciler) actionPreparing(prov provisioner.Provisioner,
 		return recordActionFailure(info, metal3api.PreparationError, provResult.ErrorMessage)
 	}
 
+	if hfcDirty && started {
+		hfcStillDirty, err := r.saveHostFirmwareComponents(prov, info, hfc)
```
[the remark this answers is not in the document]
iurygregory: if there is a failure we shouldn't retry I would say, it can be a bad firmware (Dell for example has firmware separate for each model, if I use the firmware of an R750 in R640 it complains and fails)

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1740,6 +1754,32 @@ func saveHostProvisioningSettings(host *metal3api.BareMetalHost, info *reconcile
...
+		info.log.Info("not saving hostFirmwareComponents information since is not necessary")
+		return dirty, nil
+	}
+
+	info.log.Info("saving hostFirmwareComponents information", "spec updates", hfc.Spec.Updates, "status updates", hfc.Status.Updates)
+
+	hfc.Status.Updates = make([]metal3api.FirmwareUpdate, len(hfc.Spec.Updates))
+	for i := range hfc.Spec.Updates {
+		hfc.Spec.Updates[i].DeepCopyInto(&hfc.Status.Updates[i])
+	}
```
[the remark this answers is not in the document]
iurygregory: ack, in the follow-up to track the update of components information I will update this.

controllers/metal3.io/baremetalhost_controller.go
iurygregory: I will do some tests in the follow-up I'm working on.

controllers/metal3.io/baremetalhost_controller.go
zaneb: Controller references aren't a value judgement on how tightly coupled objects are, they're to solve a specific problem between ReplicaSet and Pod that doesn't exist here.
See https://github.com/kubernetes/design-proposals-archive/blob/acc25e14ca83dfda4f66d8cb1f1b491f26e78ffe/api-machinery/controller-ref.md

controllers/metal3.io/baremetalhost_controller.go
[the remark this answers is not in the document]
zaneb: Or it could be a dropped connection or a failed write.
We shouldn't report that we've done something if we haven't done it.
If the user provides the wrong firmware we should keep trying until they realise and stop doing that.

zaneb:
/approve
This is strictly better than what we have. Should probably hold off on the backport though.

controllers/metal3.io/baremetalhost_controller.go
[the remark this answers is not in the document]
dtantsur: Yeah, unless we invent a way for Ironic to say "this will never work, don't try again", the current trend is to retry.

controllers/metal3.io/baremetalhost_controller.go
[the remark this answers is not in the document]
dtantsur: Zane is right. But I'm also wondering why the HFC controller does not update the components afterwards.

dtantsur (approved):
I'm fine with a follow-up. Thank you Iury.

/lgtm

controllers/metal3.io/baremetalhost_controller.go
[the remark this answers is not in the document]
zaneb: Hmm, good point. I thought at one point we deleted that from the HFC controller, but indeed it's still there.

controllers/metal3.io/baremetalhost_controller.go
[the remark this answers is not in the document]
iurygregory:
```
		if bmhDirty {
			info.log.Info("handling cleaning error in controller")
			clearHostProvisioningSettings(info.host)
		}
		if hfcDitry {
			clearHostFirmwareComponentsUpdates(hfc)
		}
		return recordActionFailure(info, metal3api.PreparationError, provResult.ErrorMessage)
	}
```
Something like this?

controllers/metal3.io/baremetalhost_controller.go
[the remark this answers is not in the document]
iurygregory: Yup, I've waited some hours to see if the HFC controller would attemp to update, but it didn't..,my best guess is that I need to change the conditions when calling `updateHostFirmware` in the `Reconcile` for `HostFirmwareComponentsReconciler`

PR #1795, opened 2024-06-28
:seedling: Use basic-checks image in generate and gomod tests

mquhuy:
As a continuation of https://github.com/metal3-io/project-infra/pull/797, we use basic-checks image in local tests too for BMO.

### DIFF
--- a/hack/generate.sh
+++ b/hack/generate.sh
@@ -32,6 +32,7 @@ else
         --volume "${PWD}:${WORKDIR}:rw,z" \
         --entrypoint sh \
         --workdir "${WORKDIR}" \
-        docker.io/golang:1.22 \
+        quay.io/metal3-io/basic-checks:golang-1.22 \
+        --pull=always \
         "${WORKDIR}"/hack/generate.sh "$@"
 fi
--- a/hack/gomod.sh
+++ b/hack/gomod.sh
@@ -41,6 +41,7 @@ else
         --volume "${PWD}:{WORKDIR}:ro,z" \
         --entrypoint sh \
         --workdir "${WORKDIR}" \
-        docker.io/golang:1.22 \
+        quay.io/metal3-io/basic-checks:golang-1.22 \
+        --pull=always \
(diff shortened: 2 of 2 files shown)


tuminoid (commented):
/lgtm

These work, but it appears we have some Docker related mount option issues in some of the hack scripts, that prevent returning success right now when run locally. I'll take a look at those.

/hold
Until project-infra PR is merged.

tuminoid:
Also, let's merge this first https://github.com/metal3-io/baremetal-operator/pull/1796 so this can be tested locally. Changing the docker image here doesn't mean anything from CI perspective, so it needs a manual test.

Rozzii (commented):
/lgtm

hack/generate.sh
```diff
@@ -32,6 +32,6 @@ else
         --volume "${PWD}:${WORKDIR}:rw,z" \
         --entrypoint sh \
         --workdir "${WORKDIR}" \
-        docker.io/golang:1.22 \
+        quay.io/metal3-io/basic-checks:golang-1.22 \
```
tuminoid, line 35:
```suggestion
        --pull=always \
        quay.io/metal3-io/basic-checks:golang-1.22 \
```
Otherwise locally we never will get updates made to that image after the first pull is done. For golang image it is not meaningful as its just bug fixes, but here if we add a new package here, it wouldn't show up locally unless manually forcing cleanup/repull.

hack/gomod.sh
```diff
@@ -41,6 +41,6 @@ else
         --volume "${PWD}:{WORKDIR}:ro,z" \
         --entrypoint sh \
         --workdir "${WORKDIR}" \
-        docker.io/golang:1.22 \
+        quay.io/metal3-io/basic-checks:golang-1.22 \
```
tuminoid, line 44: same here

Rozzii (commented):
/lgtm

tuminoid (commented):
/unhold
/lgtm

dtantsur (approved):

PR #1796, opened 2024-06-28
🌱 fix hack scripts to run locally again

tuminoid:
Fix hack scripts to run locally again. We have some Git permissions issues, formatting issues, deprecated linters, bad Go configs, custom mount points etc. Now it should be cleaner, runs on Linux and Mac equally (zsh can't handle "${@}" for example)...


kashifest (commented):
/approve
/cc @honza

hack/generate.sh
```diff
@@ -2,35 +2,35 @@
...
+IS_CONTAINER="${IS_CONTAINER:-false}"
+ARTIFACTS="${ARTIFACTS:-/tmp}"
 CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-podman}"
 
 if [ "${IS_CONTAINER}" != "false" ]; then
-  eval "$(go env)"
-  cd "${GOPATH}"/src/github.com/metal3-io/baremetal-operator
-  export XDG_CACHE_HOME="/tmp/.cache"
+    # we need to tell git its OK to use dir owned by someone else
+    git config --global safe.directory /workdir
```
elfosardo, line 15: the change looks ok
just wondering if we could define a variable instead of hardcoding /workdir since it's used in at least 2 files

tuminoid: Sure, changed it in all linters as it was mentioned at least 3 times each file in Docker command alone.

PR #1797, opened 2024-06-28
🌱 Optimize the dependabot configuration and add release branch configurations

kashifest:
This PR add the following:

- Adds newly introduced multi-directories feature in dependabot. With this hopefully all the go modules would be bumped in one PR for a dependancy. Also this PR optimizes the dependabot configs a bit by adding a group for k8s.io dependancy bump and allowing patch versions for the ignore rules.

- Also add release branch configurations for dependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,50 +3,105 @@
 
 version: 2
 updates:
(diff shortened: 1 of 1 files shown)


tuminoid (commented):
/lgtm

PR #1816, opened 2024-07-04
🌱 schedule dependabot on certain day, and add ok-to-test automatically

tuminoid:
BMO schedule:
- GH actions: saturday
- Gomod: sunday

Automatically add labels:
- ok-to-test

We have dynamic Prow scheduling now and proper requests, we should not suffer from load failures in very basic tests anymore. BMO is scheduled for weekend as it automatically runs BMO e2e etc, which are bit more heavy than the Prow-only linters in CAPM3/IPAM. Also it means they are tested by Monday morning to easier merging.

Fix some formatting and move blocks around so they're consistent with release branches.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,10 +8,13 @@ updates:
   directory: "/" # Location of package manifests
   schedule:
     interval: "monthly"
+    day: "saturday"
   target-branch: main
   commit-message:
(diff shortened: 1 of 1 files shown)


kashifest (commented):
/approve

Rozzii (commented):
/lgtm

PR #1824, opened 2024-07-25
🌱 Bump golang to 1.22 in release 0.6

### LINKED ISSUE #1803: Bump BMO 0.6 release branch to Go 1.22
Release branch release-0.6 has Golang 1.21.x in use. Golang 1.21 support ends in August 2024, when Golang 1.23 is released.

BMO 0.6 is in support for quite a while still, so we need to bump it to Go 1.22 before next patch release.

NymanRobin:
This is done since golang 1.21 is deprecating in August 2024

Fixes #1803

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -27,7 +27,7 @@ jobs:
     - name: Install go
       uses: actions/setup-go@cdcb36043654635271a94b9a6d1392de5bb323a7 # v5.0.1
       with:
-        go-version: '1.21'
+        go-version: '1.22'
     - name: Generate release notes
       run: |
         make release-notes
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -45,7 +45,7 @@ linters:
 
 linters-settings:
   gosec:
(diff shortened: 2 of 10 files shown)


NymanRobin:
Here is the corresponding project-infra change: https://github.com/metal3-io/project-infra/pull/823
That will be needed before this will pass the tests

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/metal3-io/baremetal-operator
 
-go 1.21
+go 1.22
```
dtantsur, line 3: This is not necessary unless a dependency is forcing us to.

dtantsur (changes requested):
Still is my regular request to avoid unnecessary changes to go.mod, especially on stable branches.

go.mod
NymanRobin: Sure, I see the concern of unnecessary changes on stable branches, should be fixed now!

tuminoid:
@dtantsur I think your concern was fixed, can you remove the "request changes" review, as it is blocking merging?

PR #1825, opened 2024-07-26
🐛 Fix HFC Status

iurygregory:
This commit fixes the following cases for HFC:
- Status Components Information using the newer information that ironic contains.
- Clear the Status Updates in case of a failed fiirmware update, so it can retry.

Also change the condition as a follow-up to baremetal-operator#1793

**What this PR does / why we need it**: metal3-io/metal3-docs#364,

### DIFF
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -1161,6 +1161,10 @@ func (r *BareMetalHostReconciler) actionPreparing(prov provisioner.Provisioner, 
 			info.log.Info("handling cleaning error in controller")
 			clearHostProvisioningSettings(info.host)
 		}
+		if hfcDirty {
+			info.log.Info("handling cleaning error during firmware update")
+			hfc.Status.Updates = nil
+		}
 		return recordActionFailure(info, metal3api.PreparationError, provResult.ErrorMessage)
 	}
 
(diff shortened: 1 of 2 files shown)


controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1287,6 +1291,11 @@ func clearHostProvisioningSettings(host *metal3api.BareMetalHost) {
 	host.Status.Provisioning.Firmware = nil
 }
 
+// clearHostFirmwareComponentsUpdates removes the values related to Updates from status.
+func clearHostFirmwareComponentsUpdates(hfc *metal3api.HostFirmwareComponents) {
```
dtantsur, line 1295: nit: This function seems too trivial to be a function

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1902,8 +1904,8 @@ func (r *BareMetalHostReconciler) getHostFirmwareComponents(info *reconcileInfo)
 		return false, nil, nil
 	}
 
-	// Check if there are Updates in the Spec that are different than the Status
-	if meta.IsStatusConditionTrue(hfc.Status.Conditions, string(metal3api.HostFirmwareComponentsChangeDetected)) {
+	// Check if the condition matches the current Generation to know if the data is not out of date.
+	if readyCond := meta.FindStatusCondition(hfc.Status.Conditions, string(metal3api.HostFirmwareComponentsChangeDetected)); readyCond != nil && readyCond.Status == metav1.ConditionTrue && readyCond.ObservedGeneration == hfc.Generation {
 		if meta.IsStatusConditionTrue(hfc.Status.Conditions, string(metal3api.HostFirmwareComponentsValid)) {
```
dtantsur, line 1909: I assume we don't need the same generation check here because all conditions are updated together, right?

iurygregory: correct

PR #1838, opened 2024-08-05
🌱 fix dependabot config on api directory location

tuminoid:
Config has mistakenly listed `/api` as a gomod ecosystem directory, but the correct directory in BMO is `/apis`.


kashifest (commented):
/approve

PR #1861, opened 2024-08-06
:seedling: Move container build to gh runner

mquhuy:
This assumes https://github.com/metal3-io/project-infra/pull/833 was merged, and that `QUAY_IO_USERNAME` and `QUAY_IO_TOKEN` secrets exist, for pushing the image to `quay.io`.

An example of complete workflow (except for the push part):
https://github.com/metal3-io/baremetal-operator/actions/runs/10280659565/job/28448467379?pr=1861

/hold to wait for the secrets.

### DIFF
--- a/.github/workflows/build-images-action.yml
+++ b/.github/workflows/build-images-action.yml
@@ -1,5 +1,8 @@
 name: build-images-action
 
+permissions:
+  contents: read
+
 on:
   push:
     branches:
@@ -8,49 +11,25 @@ on:
     tags:
     - 'v*'
 
-permissions: {}
-
 jobs:
-  build:
-    name: Build container images
-    runs-on: ubuntu-latest
-
-    permissions:
-      contents: read
-
-    if: github.repository == 'metal3-io/baremetal-operator'
-    steps:
-    - name: build bmo image
-      uses: toptal/jenkins-job-trigger-action@137fff703dd260b52b53d3ba1960396415abc568 # 1.0.2
-      with:
-        jenkins_url: "https://jenkins.nordix.org/"
-        jenkins_user: "metal3.bot@gmail.com"
-        jenkins_token: ${{ secrets.JENKINS_TOKEN }}
-        job_name: "metal3_baremetal-operator_container_image_building"
-        job_params: |
-          {
-            "BUILD_CONTAINER_IMAGE_GIT_REFERENCE": "${{ github.ref }}"
-          }
-        job_timeout: "1000"
-    - name: build keepalived image
-      uses: toptal/jenkins-job-trigger-action@137fff703dd260b52b53d3ba1960396415abc568 # 1.0.2
-      with:
-        jenkins_url: "https://jenkins.nordix.org/"
-        jenkins_user: "metal3.bot@gmail.com"
-        jenkins_token: ${{ secrets.JENKINS_TOKEN }}
-        job_name: "metal3_keepalived_container_image_building"
-        job_params: |
-          {
-            "BUILD_CONTAINER_IMAGE_GIT_REFERENCE": "${{ github.ref }}"
-          }
-        job_timeout: "1000"
-    - name: Slack Notification on Failure
-      if: ${{ failure() }}
-      uses: rtCamp/action-slack-notify@4e5fb42d249be6a45a298f3c9543b111b02f7907 # 2.3.0
-      env:
-        SLACK_TITLE: 'GitHub Action Failed in ${{ github.repository }}'
-        SLACK_COLOR: '#FF0000'
-        SLACK_MESSAGE: 'The GitHub Action workflow failed for baremetal operator image build.'
-        SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
-        SLACK_CHANNEL: metal3-github-actions-notify
-        SLACK_USERNAME: metal3-github-actions-notify
+  build_bmo:
+    name: Build BMO container image
+    uses: metal3-io/project-infra/.github/workflows/container-image-build.yml@main
+    with:
+      image-name: "baremetal-operator"
+      pushImage: true
+    secrets:
+      QUAY_USERNAME: ${{ secrets.QUAY_USERNAME }}
+      QUAY_PASSWORD: ${{ secrets.QUAY_PASSWORD }}
+      SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
+  build_keepalived:
+    name: Build keepalived container image
+    uses: metal3-io/project-infra/.github/workflows/container-image-build.yml@main
+    with:
+      image-name: "keepalived"
+      dockerfile-directory: resources/keepalived-docker
+      pushImage: true
+    secrets:
+      QUAY_USERNAME: ${{ secrets.QUAY_USERNAME }}
+      QUAY_PASSWORD: ${{ secrets.QUAY_PASSWORD }}
+      SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}


.github/workflows/build-images-action.yml
```diff
@@ -1,49 +1,106 @@
 name: build-images-action
 
 on:
-  push:
-    branches:
-    - 'main'
-    - 'release-*'
-    tags:
-    - 'v*'
+  pull_request
```
tuminoid: Put the original triggers back

tuminoid (commented):
Since we're doing such repetive steps, consider using similar approach as we have in IPAM with the osv-scanner workflow, per image. This would avoid duplicating the code and config.

mquhuy:
> Since we're doing such repetive steps, consider using similar approach as we have in IPAM with the osv-scanner workflow, per image. This would avoid duplicating the code and config.

Thanks. That's interesting! I'll give it a try

tuminoid:
Much nicer! Did you test this or is it impossible with PR + PR combo?

We need to put in the project-infra PR first https://github.com/metal3-io/project-infra/pull/833

Are the secrets in place now?

mquhuy:
> Much nicer! Did you test this or is it impossible with PR + PR combo?
> 
> We need to put in the project-infra PR first [metal3-io/project-infra#833](https://github.com/metal3-io/project-infra/pull/833)
> 
> Are the secrets in place now?

Exactly. The project-infra PR needs to be in before this one.
I temporarily changed this PR back to be triggered by PR, and uses the Nordix branch to illustrate how it works.

The secrets are not in place. I don't have the permission to add them.

tuminoid:
> Exactly. The project-infra PR needs to be in before this one. I temporarily changed this PR back to be triggered by PR, and uses the Nordix branch to illustrate how it works.

OK, but I'm not seeing the GH workflow running here?

> The secrets are not in place. I don't have the permission to add them.

Do they go to org level or repo level?

mquhuy:
> OK, but I'm not seeing the GH workflow running here?
They should be running here. I saw them in the list of test at the end of this page. Otherwise you can click on the "Action" tab at the beginning of the page. They are listed as `build-images-action / Build keepalived container image / job (pull_request)` and `build-images-action / Build BMO container image / job (pull_request)`.

> Do they go to org level or repo level?
Previously we could use org level for jenkins credentials, so I think org-level secrets work well here. Otherwise we have to add secrets to all repos.

tuminoid:
> > OK, but I'm not seeing the GH workflow running here?
> 
> They should be running here. I saw them in the list of test at the end of this page. Otherwise you can click on the "Action" tab at the beginning of the page. They are listed as `build-images-action / Build keepalived container image / job (pull_request)` and `build-images-action / Build BMO container image / job (pull_request)`.

OK, got it. I saw them earlier in the checks list too, but the run in  Actions tab looks really clean.

> > Do they go to org level or repo level?
> 
> Previously we could use org level for jenkins credentials, so I think org-level secrets work well here. Otherwise we have to add secrets to all repos.

/cc @kashifest 
We need you for this.

.github/workflows/build-images-action.yml
```diff
@@ -1,56 +1,27 @@
...
-  push:
-    branches:
-    - 'main'
-    - 'release-*'
-    tags:
-    - 'v*'
-
-permissions: {}
+  pull_request
+  # push:
```
tuminoid, line 5: Let's wait with removing these until secrets are in place and we can see the push working too, otherwise LGTM.

tuminoid:
> > > Do they go to org level or repo level?
> > 
> > 
> > Previously we could use org level for jenkins credentials, so I think org-level secrets work well here. Otherwise we have to add secrets to all repos.
> 
> /cc @kashifest We need you for this.

Actually now that we call the workflow from project-infra, I'd think the secrets could be only there and not in org level. We could try that before adding to org level.

mquhuy:
> Actually now that we call the workflow from project-infra, I'd think the secrets could be only there and not in org level. We could try that before adding to org level.

IMO that wouldn't work. According to [1], _In the reusable workflow, use the inputs and secrets keywords to define inputs or secrets that will be passed from a caller workflow._, but we can try.

[1] https://docs.github.com/en/actions/using-workflows/reusing-workflows#using-inputs-and-secrets-in-a-reusable-workflow

tuminoid:
> > Actually now that we call the workflow from project-infra, I'd think the secrets could be only there and not in org level. We could try that before adding to org level.
> 
> IMO that wouldn't work. According to [1], _In the reusable workflow, use the inputs and secrets keywords to define inputs or secrets that will be passed from a caller workflow._, but we can try.
> 
> [1] https://docs.github.com/en/actions/using-workflows/reusing-workflows#using-inputs-and-secrets-in-a-reusable-workflow

In that case, we'll put them directly into org. No reason to assume documentation is wrong.

mquhuy:
Note for reviewers: Unfortunately, we aren't able to test the image pushes in this PR, since according to [this doc](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow), the secrets are not available for forks (and PRs creating from forks). We will need to merge the PR to see if it really works.

tuminoid (commented):
/lgtm

tuminoid:
/cc @kashifest 
Please approve this, we need to merge this to see if the push works. Secrets are not available to PRs.

mquhuy:
@tuminoid Could you please re-lgtm? I accidentally force pushed this PR when I tested it upstream.

.github/workflows/build-images-action.yml
```diff
@@ -8,49 +8,24 @@ on:
     tags:
     - 'v*'
 
-permissions: {}
```
tuminoid, line 11: Put this back.

.github/workflows/build-images-action.yml
```diff
@@ -8,49 +8,24 @@ on:
...
-
 jobs:
-  build:
-    name: Build container images
-    runs-on: ubuntu-latest
-
-    permissions:
```
tuminoid, line 18: And if you need permissions, put this back as well.

.github/workflows/build-images-action.yml
mquhuy: Unfortunately that wouldn't work. The wf would fail with:

```
The workflow is not valid. .github/workflows/build-images-action.yml (Line: 15, Col: 3): Error calling workflow 'Nordix/metal3-project-infra/.github/workflows/container-image-build.yml@mquhuy/change-container-image-build-wf-permissions'. The nested job 'job' is requesting 'contents: read', but is only allowed 'contents: none'.
```

.github/workflows/build-images-action.yml
mquhuy: That wouldn't work either, the job was supposed to be taken as a whole, not as a step in the bigger wf:

```
The workflow is not valid. .github/workflows/build-images-action.yml (Line: 21, Col: 5): Unexpected value 'uses' .github/workflows/build-images-action.yml (Line: 22, Col: 5): Unexpected value 'with'
```

https://github.com/metal3-io/baremetal-operator/actions/runs/10300436232

.github/workflows/build-images-action.yml
mquhuy: https://github.com/metal3-io/baremetal-operator/actions/runs/10300338972

mquhuy: Or maybe we could put `permissions: contents: read` as the default instead of `permissions: {}`?

tuminoid: OK, it seems that is the working option. Let's do that

.github/workflows/build-images-action.yml
tuminoid: OK, let's go with global read first, and iterate from that if it is improvable.

.github/workflows/build-images-action.yml
```diff
@@ -1,5 +1,8 @@
 name: build-images-action
 
+permission:
```
tuminoid, line 3:
```suggestion
permissions:
```

tuminoid (commented):
/lgtm

PR #1868, opened 2024-08-09
:seedling: Make container build only run for metal3.io repo

mquhuy:
The container build github workflow should be triggered only if it's running on the upstream.


Rozzii (commented):
/lgtm

kashifest (commented):
/approve

PR #1879, opened 2024-08-13
:seedling: Fix syntax errors in generate.sh and gomod.sh

mquhuy:
`--pull=always` seems to be causing issues for both docker and podman:

```
docker run --rm --env IS_CONTAINER=TRUE --volume /home/ubuntu/baremetal-operator:/workdir:ro,z --entrypoint sh --workdir /workdir quay.io/metal3-io/basic-checks:golang-1.22 --pull=always /workdir/hack/gomod.sh
sh: 0: Illegal option --
```

Moving the `--pull=always \` line to before other options solved the issue. Also fixed the missing `$` in directory reference in `gomod.sh`.


tuminoid (commented):
/lgtm

This was my bad originally, not sure how the patch was applied like that. It only affects local runs, so thanks for noticing.

kashifest (commented):
/approve

Issue #1253, opened 2023-04-17
`saveHostStatus` triggers even on unchanged object


s3rj1k:
```
func (r *BareMetalHostReconciler) saveHostStatus(host *metal3v1alpha1.BareMetalHost) error {
	t := metav1.Now()
	host.Status.LastUpdated = &t

	return r.Status().Update(context.TODO(), host)
}
```

There is an edge case when `saveHostStatus` is triggered even when no actual objects changes are made, the only change to object is updating `LastUpdated` filed.

It would be better to have this field updated only if something in status was actually changed.

Current behavior adds unneeded reconciles for external watchers that follow BMH changes, this becomes noticeable on big clusters with lots for BMH.

s3rj1k:
@dtantsur What do you think?

dtantsur:
From a quick look at the source code, we do seem to only call `saveHostStatus` when the action returns a `dirty` flag. Maybe we sometimes do it incorrectly. Do you have specific examples when it happens @s3rj1k?

s3rj1k:
@Rozzii What information is needed?

s3rj1k:
> Do you have specific examples when it happens @s3rj1k?

It affected me noticeably during the pre-provision phase (in big cluster), when machines are getting linked with hosts, host where changing constantly.

Rozzii:
@s3rj1k I would like to have steps to reproduce the issue, as far as I understand from your description this happens when there are a lot of new machines are getting registered  but from that I can't tell what exactly is causing the issue.

s3rj1k:
> @s3rj1k I would like to have steps to reproduce the issue, as far as I understand from your description this happens when there are a lot of new machines are getting registered but from that I can't tell what exactly is causing the issue.

I am not sure what extra steps you need to reproduce this, the issue of NOP updates is there all the time, just check the code, updates are unconditional, every reconcile (in not provisioned states) time is getting updated.

Rozzii:
I will go ahead and accept this, although I am not that familiar with this part of the code, the issue makes sense to me and no one else have reacted.
/accept
/remove-lifecycle stale
/lifecycle frozen

zaneb:
> I am not sure what extra steps you need to reproduce this, the issue of NOP updates is there all the time, just check the code, updates are unconditional, every reconcile (in not provisioned states) time is getting updated.

[...]

PR #1881, opened 2024-08-13
🌱Add test for baremetalhost controller updateEventHandler

troy0820:

~~When using this method, the `saveHostStatus` will requeue the object and create more reconciles for the object when the only thing that has changed is the lastupdated field in the status.  This leads to multiple reconciles because this happens during each run through the reconciler.~~

This adds a test to the baremetalhost controller to validate that the saveHostStatus will not requeue the object and create more reconciles when lastUpdated is updated.

~~Fixes https://github.com/metal3-io/baremetal-operator/issues/1253~~

### DIFF
--- a/controllers/metal3.io/baremetalhost_controller_test.go
+++ b/controllers/metal3.io/baremetalhost_controller_test.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"testing"
+	"time"
 
 	metal3api "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
 	"github.com/metal3-io/baremetal-operator/pkg/hardwareutils/bmc"
@@ -1664,6 +1665,8 @@ func TestProvisionerIsReady(t *testing.T) {
 }
 
 func TestUpdateEventHandler(t *testing.T) {
+	now := metav1.Now()
+	later := metav1.Now().Add(time.Second * 20)
 	cases := []struct {
 		name            string
 		event           event.UpdateEvent
@@ -1765,6 +1768,30 @@ func TestUpdateEventHandler(t *testing.T) {
 			},
 
 			expectedProcess: true,
+		},
+		{
+			name: "do-not-requeue-when-only-status-lastupdated-are-different",
+			event: event.UpdateEvent{
+				ObjectOld: &metal3api.BareMetalHost{ObjectMeta: metav1.ObjectMeta{
+					Generation:  0,
+					Finalizers:  []string{},
+					Annotations: map[string]string{},
+				},
+					Status: metal3api.BareMetalHostStatus{
+						LastUpdated: &metav1.Time{Time: now.Time},
+					},
+				},
+				ObjectNew: &metal3api.BareMetalHost{ObjectMeta: metav1.ObjectMeta{
+					Generation:  0,
+					Finalizers:  []string{},
+					Annotations: map[string]string{},
+				},
+					Status: metal3api.BareMetalHostStatus{
+						LastUpdated: &metav1.Time{Time: later},
+					}},
+			},
+
+			expectedProcess: false,
 		},
 	}
 


troy0820:
Actually it looks like it may be in here:
```go
_, oldOK := e.ObjectOld.(*metal3api.BareMetalHost)
	_, newOK := e.ObjectNew.(*metal3api.BareMetalHost)
	if !(oldOK && newOK) {
		// The thing that changed wasn't a host, so we
		// need to assume that we must update. This
		// happens when, for example, an owned Secret
		// changes.
		return true
	}

```
Because the status will not be the same and then it will return true and not process the rest of the handler.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -2080,7 +2080,7 @@ func (r *BareMetalHostReconciler) updateEventHandler(e event.UpdateEvent) bool {
 	}
 
 	// Discard updates that did not increase the resource Generation (such as on Status.LastUpdated), except for the finalizers or annotations
-	if reflect.DeepEqual(e.ObjectNew.GetFinalizers(), e.ObjectOld.GetFinalizers()) && reflect.DeepEqual(e.ObjectNew.GetAnnotations(), e.ObjectOld.GetAnnotations()) {
+	if reflect.DeepEqual(e.ObjectNew.GetFinalizers(), e.ObjectOld.GetFinalizers()) && reflect.DeepEqual(e.ObjectNew.GetAnnotations(), e.ObjectOld.GetAnnotations()) || oldBMH.Status.LastUpdated != newBMH.Status.LastUpdated {
 		return false
```
troy0820, line 2084: This seems wrong, in the test below, we would want that event to come through because the finalizer is different from old/new, this prevents it because of the or. We can move this to a new line and isolate that.

troy0820: Added test and removed the logic on the line with the false condition.  Now this will be isolated from the event where the LastUpdated was modified but nothing else changed.

tuminoid:
Thanks for the contribution! Let's squash those commits up.

Let's get some eyes on this.
/cc @kashifest @Rozzii @dtantsur @elfosardo

kashifest:
I am trying to understand this patch a bit, the original issue pointed by @s3rj1k was an edge case where code updated the status/BMH object even when there was no actual change in status, but it seems this patch is ignoring updating the object even when theres is any update on the status? Am I understanding it correctly?

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -2079,6 +2079,11 @@ func (r *BareMetalHostReconciler) updateEventHandler(e event.UpdateEvent) bool {
 		return true
 	}
 
+	// The assumption that this will not always be filled out is why this check is not appended to the below block
+	if oldBMH.Status.LastUpdated != nil && newBMH.Status.LastUpdated != nil && oldBMH.Status.LastUpdated != newBMH.Status.LastUpdated && reflect.DeepEqual(oldBMH.Spec, newBMH.Spec) && reflect.DeepEqual(oldBMH.ObjectMeta, newBMH.ObjectMeta) {
```
zaneb, line 2083: Why only skip reconcile when LastUpdated has changed?

Naively I'd have thought we'd want to skip it _even more_ when LastUpdated has **not** changed.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -2079,6 +2079,11 @@ func (r *BareMetalHostReconciler) updateEventHandler(e event.UpdateEvent) bool {
...
+		return false
+	}
+
 	// Discard updates that did not increase the resource Generation (such as on Status.LastUpdated), except for the finalizers or annotations
 	if reflect.DeepEqual(e.ObjectNew.GetFinalizers(), e.ObjectOld.GetFinalizers()) && reflect.DeepEqual(e.ObjectNew.GetAnnotations(), e.ObjectOld.GetAnnotations()) {
```
zaneb, line 2088: Isn't this already a rather awkward way of saying `if reflect.DeepEqual(oldBMH.ObjectMeta, newBMH.ObjectMeta) { return false }`?
i.e. this already captures everything that the new test above would capture, and then some.

troy0820:
> I am trying to understand this patch a bit, the original issue pointed by @s3rj1k was an edge case where code updated the status/BMH object even when there was no actual change in status, but it seems this patch is ignoring updating the object even when theres is any update on the status? Am I understanding it correctly?

The object is updated but it doesn’t requeue the object in the event that the status last changed gets updated during a saveStatus. This was supposed to be handled by the predicate generation changed but seems to slip through.

zaneb (commented):
I'm not sure this has any effect.

It's probably safe as long as we always remember to return Requeue:true whenever we need to requeue.
I think we do, but I also thought that we avoid writing unchanged statuses (as Dmitry [pointed out](https://github.com/metal3-io/baremetal-operator/issues/1253#issuecomment-1549695960)), yet according the the bug report we do not. (OTOH, I'm not sure whether to believe this, because when asked where this happens the [response](https://github.com/metal3-io/baremetal-operator/issues/1253#issuecomment-1647654820) was "check the code", and yet upon checking the code it's [clearly fine](https://github.com/metal3-io/baremetal-operator/blob/main/controllers/metal3.io/baremetalhost_controller.go#L233-L237).)

controllers/metal3.io/baremetalhost_controller.go
troy0820: The above check from this patch would make sure that everything in ObjectMeta is equal including what is not explicitly set with the Finalizers and/or the Annotations, e.g: Labels, OwnerRefs, ResourceVersion, Generation, etc.  

The check on line 2088 seems to only care about the Finalizers and Annotations which would create requeues with Label changes, ownerRefs, etc which would be valid. 

The check I am putting in makes sure that the only change that is different is that the LastUpdated on the status is changed without anything else being different, therefor not requeuing.

troy0820:
Running the test that I added as part of this change works without the patch included.  So I can dismiss this change and leave the test but we probably should close the issue if it is fixed.

controllers/metal3.io/baremetalhost_controller_test.go
```diff
@@ -1766,6 +1769,54 @@ func TestUpdateEventHandler(t *testing.T) {
 
 			expectedProcess: true,
 		},
+		{
+			name: "do-requeue-when-finalizers-are-different",
```
andfasano, line 1773: Isn't already covered by https://github.com/metal3-io/baremetal-operator/pull/1881/files#diff-20c1a976622b8d54cf98349e3e6219778fa00e89ed669f1bb12cd778326469bbR1733?

controllers/metal3.io/baremetalhost_controller_test.go
```diff
@@ -1766,6 +1769,54 @@ func TestUpdateEventHandler(t *testing.T) {
...
+				},
+					Status: metal3api.BareMetalHostStatus{
+						LastUpdated: &metav1.Time{Time: later},
+					}},
+			},
+
+			expectedProcess: true,
+		},
+		{
+			name: "do-not-requeue-when-status-lastupdated-are-different",
```
andfasano, line 1797: Just my two cents, but given that there is already a test for [base case](https://github.com/metal3-io/baremetal-operator/pull/1881/files#diff-20c1a976622b8d54cf98349e3e6219778fa00e89ed669f1bb12cd778326469bbR1693), I would find a better name something like `do-not-requeue-when-only-status-lastupdated-are-different`

troy0820: I can update the name

controllers/metal3.io/baremetalhost_controller_test.go
troy0820: I can remove this but this was to simulate that this would requeue despite the LastUpdated in the status.  This wouldn't do anything other than prove that this still requeues and not get swallowed by the false requeue from the LastUpdated in the status.

controllers/metal3.io/baremetalhost_controller.go
zaneb: We basically have:

```
if foo && bar && baz &&quux {
  return false
}
if foo {
  return false
}
return true
```

When put like that, it's pretty easy to see that the first check isn't doing anything.

troy0820: This change is now outdated and only includes the test to verify that the change implemented to address the root issue is resolved. 

The thought behind it was the first check would include the other changes that don’t get captured in the second check which is why the status is not included but only the spec/metadata and the LastUpdated.

troy0820: Resolving as not relevant to the review.

controllers/metal3.io/baremetalhost_controller.go
troy0820: > Running the test that I added as part of this change works without the patch included. So I can dismiss this change and leave the test but we probably should close the issue if it is fixed.

This removes the included patch but leaves the test.  Resolving

zaneb:
Reading over the bug report, the complaint was that we were doing no-op updates in the first place (although there is no evidence of this yet), thus causing extra work for _other_ controllers that were watching the BMH.
There was no suggestion that BMO itself was doing too many reconciles after such an update. Extra reconciles for that **and** a much broader superset of cases were already eliminated by #747.

I'm fine with adding a test though. We could even test that _all_ updates to the Status are ignored.

controllers/metal3.io/baremetalhost_controller_test.go
troy0820: Resolving is the first test is removed from patch set.

controllers/metal3.io/baremetalhost_controller_test.go
troy0820: The name is updated.  I am resolving.

Rozzii (commented):
/lgtm

tuminoid:
/cc @kashifest @zaneb @dtantsur @elfosardo 
Can we get an approver for this?

PR #1888, opened 2024-08-14
🌱 fix verify-release.sh to understand go directive better

tuminoid:
Currently it fails to do the vulnerability scan if the go directive has golang patch version in it, like "1.22.3", as it replaces "1.22" with "1.22.5" it finds from Dockerfile, resulting to "1.22.5.3" which blows up.


smoshiur1237 (commented):
LGTM

kashifest (commented):
/approve

PR #1889, opened 2024-08-14
✨ Prevent BMH from transitioning when an invalid BIOS setting is detected

tliu2021:
Currently, the installation process ignores invalid BIOS settings, which can result in a system booting without the desired configuration. This commit ensures that the process halts when an invalid BIOS setting is detected, allowing the user to correct the issue before proceeding, thereby preventing unintended configurations.

Fixes #

### DIFF
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -53,6 +53,7 @@ const (
 	preprovImageRetryDelay        = time.Minute * 5
 	provisionerNotReadyRetryDelay = time.Second * 30
 	subResourceNotReadyRetryDelay = time.Second * 60
+	configurationErrRetryDelay    = time.Second * 90
 	clarifySoftPoweroffFailure    = "Continuing with hard poweroff after soft poweroff fails. More details: "
 	hardwareDataFinalizer         = metal3api.BareMetalHostFinalizer + "/hardwareData"
 )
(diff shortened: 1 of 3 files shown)


controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1128,8 +1129,14 @@ func (r *BareMetalHostReconciler) actionPreparing(prov provisioner.Provisioner,
 	hfsDirty, hfs, err := r.getHostFirmwareSettings(info)
 
 	if err != nil {
-		// wait until hostFirmwareSettings are ready
-		return actionContinue{subResourceNotReadyRetryDelay}
+		switch {
+		case errors.As(err, &ConfigurationError{}):
+			// wait for the user to correct the error
+			return actionContinue{configurationErrRetryDelay}
```
dtantsur, line 1135: Why is it different from what we had before? You're only changing the retry delay for 1 to 1.5 minutes. Did you intend to you something else than actionContinue?

tliu2021: Previously, the operation did not return an error, and there was no retry delay (see line 1892 below).

tliu2021:
This behavior needs to be triggered based on a label in the HFS or BMH.

controllers/metal3.io/baremetalhost_controller.go
dtantsur: I think on line 1132 the error would be returned in any case.

PR #1892, opened 2024-08-15
:seedling: Make kustomize apply function not dependent on ginkgo

mquhuy:
The goal is to make this function usable for other packages, which may not use ginkgo

### DIFF
--- a/test/e2e/common.go
+++ b/test/e2e/common.go
@@ -328,8 +328,10 @@ func (input *BuildAndApplyKustomizationInput) validate() error {
 // BuildAndApplyKustomization takes input from BuildAndApplyKustomizationInput. It builds the provided kustomization
 // and apply it to the cluster provided by clusterProxy.
 func BuildAndApplyKustomization(ctx context.Context, input *BuildAndApplyKustomizationInput) error {
-	Expect(input.validate()).To(Succeed())
 	var err error
+	if err = input.validate(); err != nil {
+		return err
+	}
 	kustomization := input.Kustomization
 	clusterProxy := input.ClusterProxy
 	manifest, err := BuildKustomizeManifest(kustomization)


mquhuy:
/cc @kashifest @tuminoid 

I would like to merge this before 0.7 release, so that this function could be used in `deploy-cli`.

tuminoid:
> /cc @kashifest @tuminoid
> 
> I would like to merge this before 0.7 release, so that this function could be used in `deploy-cli`.

You're suggesting using code from e2e tests in deploy, which is non-test code? If that is what you're after, then this code should live somewhere else.

kashifest (commented):
/approve

mquhuy:
> > /cc @kashifest @tuminoid
> > I would like to merge this before 0.7 release, so that this function could be used in `deploy-cli`.
> 
> You're suggesting using code from e2e tests in deploy, which is non-test code? If that is what you're after, then this code should live somewhere else.

Agreed. But currently we don't have anywhere more suitable for this code. I don't think it's a blocker though, we can decide on having common package which can be imported by both test and non-test later.

tuminoid:
> > > /cc @kashifest @tuminoid
> > > I would like to merge this before 0.7 release, so that this function could be used in `deploy-cli`.
> > 
> > 
> > You're suggesting using code from e2e tests in deploy, which is non-test code? If that is what you're after, then this code should live somewhere else.
> 
> Agreed. But currently we don't have anywhere more suitable for this code. I don't think it's a blocker though, we can decide on having common package which can be imported by both test and non-test later.

This change is fine by itself as it just removes Ginkgo calls, but do not use it in deploy-cli without relocating it elsewhere first. We should not be importing test code into prod code.

tuminoid (commented):
/lgtm
/unhold

PR #1894, opened 2024-08-15
:seedling: Bump docker/docker in test/ to v25.0.6

mquhuy:
Bump to fix this issue:
https://osv.dev/vulnerability/GO-2024-3005


tuminoid:
This is not going to work. Docker major versions are not compatible without code changes, hence I did not bump it earlier as we don't use the vulnerable functionality.

/hold

PR #1932, opened 2024-09-03
🌱 remove release-0.5 from dependabot config

tuminoid:
Remove release-0.5 from dependabot config as we're no longer maintaining it.


kashifest (commented):
/approve

PR #1936, opened 2024-09-04
🌱 Fix minor issues in the API help strings

dtantsur:
Follow-up to commit 9a30c0de44ff7534b568000bb3f818a8d168af82.

### DIFF
--- a/apis/metal3.io/v1alpha1/baremetalhost_types.go
+++ b/apis/metal3.io/v1alpha1/baremetalhost_types.go
@@ -233,7 +233,7 @@ const (
 	StateInspecting ProvisioningState = "inspecting"
 
 	// StatePoweringOffBeforeDelete means we are in the process of
-	// powering off the node before it's deleted.
+	// powering off the host before it's deleted.
 	StatePoweringOffBeforeDelete ProvisioningState = "powering off before delete"
 
 	// StateDeleting means we are in the process of cleaning up the host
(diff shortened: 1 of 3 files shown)


tuminoid (commented):
/lgtm

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -833,7 +832,7 @@ type BareMetalHostStatus struct {
 	// The last error message reported by the provisioning subsystem.
 	ErrorMessage string `json:"errorMessage"`
 
-	// Whether or not the host is currently powered on. This field may get
+	// The currently detected power state of the host. This field may get
 	// briefly out of sync with the actual state of the hardware while
 	// provisioning processes are running.
```
zaneb, line 837: I meant we could say 'currently detected' _instead_ of all the stuff about being briefly out of sync.
I feel like it goes without saying that we are in a distributed system and detection is not instananeous.

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -405,8 +405,8 @@ type BareMetalHostSpec struct {
 	// +kubebuilder:validation:Pattern=`[0-9a-fA-F]{2}(:[0-9a-fA-F]{2}){5}`
 	BootMACAddress string `json:"bootMACAddress,omitempty"`
 
-	// Should the host be powered on? Changing this value will trigger
-	// a change in power state of the host.
+	// Should the host be powered on? If the host is not currently being
+	// provisioned, its power state will be forced to match this value.
```
zaneb, line 409: Consider saying "once the host is in a steady state", since there are other states where this is ignored (e.g. inspecting, preparing).

apis/metal3.io/v1alpha1/baremetalhost_types.go
dtantsur: I'd rather keep the second part here because it's a common source of confusion.

tuminoid (commented):
/lgtm

Issue #1383, opened 2023-10-10
E2e resource collection


lentzi90:
**User Story**

As a developer I would like to collect BMHs and other relevant resources from e2e tests in order to understand failures better.

**Detailed Description**

The e2e tests currently collects BMO and Ironic logs along with events from the namespaces where the test resources are created.
This is already quite good, but in some cases it would be useful to get the complete BMH object. Perhaps other custom resources could also be useful. I would say we should even try to get the state from Ironic's point of view (e.g. `baremetal node list`).
See [this function](https://github.com/kubernetes-sigs/cluster-api/blob/90311aeabd7a1d8f6f47e14259b4c3d34d2b1b07/test/framework/alltypes_helpers.go#L120) in CAPI e2e. We can probably reuse it.

- [ ] Collect BareMetalHosts
- [ ] Collect other [custom resources](https://github.com/metal3-io/baremetal-operator/tree/main/config/crd/bases) relevant to BMO
- [ ] Collect Ironic state

/kind feature

Rozzii:
I don't remember whether someone has done or is doing anything related to this. Does anyone have any info on this ? Maybe stuff you have done or are doing overlaps with this? @mquhuy @smoshiur1237 @Sunnatillo @huutomerkki

mquhuy:
> I don't remember whether someone has done or is doing anything related to this. Does anyone have any info on this ? Maybe stuff you have done or are doing overlaps with this? @mquhuy @smoshiur1237 @Sunnatillo @huutomerkki

I think @huutomerkki is working on some items of this list.

lentzi90:
/reopen
Only the last point was fixed in the linked PR

lentzi90:
Ok now I see. There was another PR that fixed the two other points: https://github.com/metal3-io/baremetal-operator/pull/1939
/close

PR #1939, opened 2024-09-06
🌱 Add resource dumping to e2e tests

huutomerkki:
This PR adds custom resource collection/dumping to BMO e2e tests. The collection is done at AfterEach of each test case.

It currently collects all CRDs and related CRs, including BMHs, but adding a filter if we one need should be easy.

It also calls framework.DumpAllResources as that was mentioned in the linked issue, but that doesn't seem to do much. I'm not sure if it should be included.

This PR is related to #1383 and fulfills two of the subtasks.

### DIFF
--- a/test/go.mod
+++ b/test/go.mod
@@ -13,6 +13,7 @@ require (
 	golang.org/x/crypto v0.26.0
 	gopkg.in/yaml.v2 v2.4.0
 	k8s.io/api v0.29.8
+	k8s.io/apiextensions-apiserver v0.29.8
 	k8s.io/apimachinery v0.29.8
 	k8s.io/klog/v2 v2.110.1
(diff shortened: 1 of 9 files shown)


test/go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	golang.org/x/crypto v0.26.0
 	gopkg.in/yaml.v2 v2.4.0
 	k8s.io/api v0.29.8
+	k8s.io/apiextensions-apiserver v0.29.3
```
tuminoid, line 16: Use same version as other k8s.io deps, `v0.29.8`

tuminoid (commented):
Nit

kashifest (commented):
/approve

Rozzii (commented):
/lgtm

PR #1954, opened 2024-09-12
✨ Add redfish-uefihttp driver

derekhiggins:
Add new interface to allow UefiHttp boot.

Design Doc
https://github.com/metal3-io/metal3-docs/blob/main/design/baremetal-operator/uefi-http-boot.md

Needs to be used along with
https://github.com/metal3-io/ironic-image/pull/561
https://github.com/metal3-io/metal3-dev-env/pull/1457

### DIFF
--- a/pkg/hardwareutils/bmc/access_test.go
+++ b/pkg/hardwareutils/bmc/access_test.go
@@ -433,6 +433,18 @@ func TestStaticDriverInfo(t *testing.T) {
 		},
 
 		{
+			Scenario:   "redfish uefi http boot",
+			input:      "redfish-uefihttp+https://192.168.122.1",
+			needsMac:   true,
+			driver:     "redfish",
+			bios:       "",
+			boot:       "redfish-https",
+			firmware:   "redfish",
+			management: "",
+			power:      "",
+		},
+
+		{
 			Scenario:   "idrac redfish",
 			input:      "idrac-redfish://192.168.122.1",
 			needsMac:   true,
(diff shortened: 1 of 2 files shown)


pkg/hardwareutils/bmc/access_test.go
```diff
@@ -432,6 +432,18 @@ func TestStaticDriverInfo(t *testing.T) {
 			power:      "",
 		},
 
+		{
+			Scenario:   "redfish uefi http boot",
+			input:      "redfish-https+https://192.168.122.1",
```
derekhiggins, line 437: hmm, according to the design doc I should have called this redfish-uefihttp
https://github.com/metal3-io/metal3-docs/blob/main/design/baremetal-operator/uefi-http-boot.md

derekhiggins:
/hold
see inline comment

pkg/hardwareutils/bmc/access_test.go
dtantsur: Yeah, let's use that name: redfish-https+https will be confused with redfish+https for sure

Rozzii (commented):
/lgtm

pkg/hardwareutils/bmc/access_test.go
tuminoid: Does PR title needs update then as well?

Rozzii (commented):
/lgtm

pkg/hardwareutils/bmc/access_test.go
Rozzii: Yeah I was wondering the same should the title be Add "redfish-uefihttp" interface?
Or should we refer to it as  Add supportfor  "redfish-http" boot interface?

PR #1963, opened 2024-09-16
:seedling: Cleanup unused ironic:e2e tryload

mquhuy:
`ironic:e2e` isn't an image we build or have, so trying to load it is just confusing, and wasteful too.


tuminoid (commented):
/lgtm

PR #1966, opened 2024-09-17
🌱 Adding ipxe builder to init container

adilGhaffarDev:
Adding ipxe builder to ironic init container

Fixes #


mquhuy:
So this ipxe-builder will run as an init container for ironic? Why does it have its own overlays?

Rozzii:
> So this ipxe-builder will run as an init container for ironic? Why does it have its own overlays?

Yes it will run as init container, the idea is that it is not mandatory if user is fine with the default iPXE firmware that this init container is not needed.

ironic-deployment/overlays/ipxe-builder-tls/kustomization.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: baremetal-operator-system
+resources:
+- https://github.com/metal3-io/baremetal-operator/config/namespace?ref=release-0.8&timeout=120s
+- https://github.com/metal3-io/baremetal-operator/ironic-deployment/base?ref=release-0.8&timeout=120s
```
mquhuy, line 6: I suppose we want to have ironic-tls here as well?

mquhuy (commented):
I think the overlays names should be something like `with-ipxe-builder`, instead of `ipxe-builder`. That overlay should install ironic with `ipxe-builder` as an init container, not just the ipxe-builder itself.

I left an inline comment. Currently this overlay should give us tls on ipxe-builder, but not on ironic. Not sure if that's what we want?

mquhuy:
> > So this ipxe-builder will run as an init container for ironic? Why does it have its own overlays?
> 
> Yes it will run as init container, the idea is that it is not mandatory if user is fine with the default iPXE firmware that this init container is not needed.

Thanks for the explanation. I was confused by the overlays names.

ironic-deployment/overlays/ipxe-builder-tls/kustomization.yaml
adilGhaffarDev: fixed

adilGhaffarDev: also changed the name of the folder

Rozzii:
FYI: I have started reviewing this, looks good but I have not finished yet.

tuminoid:
LGTM to me

Rozzii (commented):
/approve

PR #1689, opened 2024-04-19
✨ Host live updates - proof of concept

dtantsur:

PR #1969, opened 2024-09-20
✨ Add reference to HostUpdatePolicy in Servicing.

rhjanders:

This PR enables BMO to run Ironic servicing operations (such as applying firmware settings changes - or in the future firmware updates to already provisioned nodes). Servicing is an opt-in feature and is controlled by creation of a HostUpdatePolicy for a node with attributes indicating the desire to make changes to firmware configuration onReboot.

This is a partial implementation of https://github.com/metal3-io/metal3-docs/blob/main/design/baremetal-operator/host-live-updates.md (please note only firmware settings changes are currently supported, firmware update support will be added next).

### DIFF
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -889,7 +889,6 @@ func (r *BareMetalHostReconciler) registerHost(prov provisioner.Provisioner, inf
 
 	// Create the hostFirmwareSettings resource with same host name/namespace if it doesn't exist
 	// Create the hostFirmwareComponents resource with same host name/namespace if it doesn't exist
-	// Set owner reference on hostUpdatePolicy resource if not set
 	if info.host.Name != "" {
 		if !info.host.DeletionTimestamp.IsZero() {
 			info.log.Info(fmt.Sprintf("will not attempt to create new hostFirmwareSettings and hostFirmwareComponents in %s", info.host.Namespace))
@@ -901,10 +900,6 @@ func (r *BareMetalHostReconciler) registerHost(prov provisioner.Provisioner, inf
 			if err = r.createHostFirmwareComponents(info); err != nil {
 				info.log.Info("failed creating hostfirmwarecomponents")
 				return actionError{errors.Wrap(err, "failed creating hostFirmwareComponents")}
-			}
-			if err = r.hostUpdatePolicySetOwnerReference(info); err != nil {
-				info.log.Info("failed setting owner reference on hostupdatepolicy")
-				return actionError{errors.Wrap(err, "failed setting owner reference on hostUpdatePolicy")}
 			}
 		}
 	}
@@ -1362,7 +1357,7 @@ func (r *BareMetalHostReconciler) actionDeprovisioning(prov provisioner.Provisio
 	return actionComplete{}
 }
 
-func (r *BareMetalHostReconciler) checkServicing(prov provisioner.Provisioner, info *reconcileInfo) (result actionResult, isServicing bool) {
+func (r *BareMetalHostReconciler) doServiceIfNeeded(prov provisioner.Provisioner, info *reconcileInfo, hup *metal3api.HostUpdatePolicy) (result actionResult, isServicing bool) {
 	servicingData := provisioner.ServicingData{}
 
 	// (NOTE)janders: since Servicing is an opt-in feature that requires HostUpdatePolicy to be created and set to onReboot
@@ -1371,12 +1366,6 @@ func (r *BareMetalHostReconciler) checkServicing(prov provisioner.Provisioner, i
 	var fwDirty bool
 	var hfsDirty bool
 	var liveFirmwareSettingsAllowed bool
-
-	hup := &metal3api.HostUpdatePolicy{}
-	err := r.Get(info.ctx, info.request.NamespacedName, hup)
-	if err != nil {
-		return actionError{fmt.Errorf("unable to fetch HostUpdatePolicy, aborting servicing. Error: %w", err)}, false
-	}
 
 	if hup != nil {
 		liveFirmwareSettingsAllowed = (hup.Spec.FirmwareSettings == metal3api.HostUpdatePolicyOnReboot)
@@ -1456,11 +1445,9 @@ func (r *BareMetalHostReconciler) manageHostPower(prov provisioner.Provisioner, 
 	if hwState.PoweredOn != nil && *hwState.PoweredOn != info.host.Status.PoweredOn {
 		info.log.Info("updating power status", "discovered", *hwState.PoweredOn)
 		info.host.Status.PoweredOn = *hwState.PoweredOn
-		targetOperationalStatus := metal3api.OperationalStatusOK
-		if info.host.Status.OperationalStatus == metal3api.OperationalStatusServicing {
-			targetOperationalStatus = metal3api.OperationalStatusServicing
+		if info.host.Status.OperationalStatus == metal3api.OperationalStatusError || info.host.Status.ErrorType == metal3api.PowerManagementError {
+			clearError(info.host)
 		}
-		clearErrorWithStatus(info.host, targetOperationalStatus)
 		return actionUpdate{}
 	}
 
@@ -1481,11 +1468,16 @@ func (r *BareMetalHostReconciler) manageHostPower(prov provisioner.Provisioner, 
 			return actionContinue{}
 		}
 	}
+	hup, err := r.hostUpdatePolicySetOwnerReference(info)
+	if err != nil {
+		info.log.Info("failed setting owner reference on hostupdatepolicy")
+		return actionError{errors.Wrap(err, "failed setting owner reference on hostUpdatePolicy")}
+	}
 
 	servicingAllowed := isProvisioned && !info.host.Status.PoweredOn && desiredPowerOnState
 	if servicingAllowed || info.host.Status.OperationalStatus == metal3api.OperationalStatusServicing || info.host.Status.ErrorType == metal3api.ServicingError {
-		result, isServicing := r.checkServicing(prov, info)
-		if result != nil && (result.Dirty() || isServicing) {
+		result, _ := r.doServiceIfNeeded(prov, info, hup)
+		if result != nil {
 			return result
 		}
 	}
@@ -1962,7 +1954,7 @@ func (r *BareMetalHostReconciler) createHostFirmwareSettings(info *reconcileInfo
 	return nil
 }
 
-func (r *BareMetalHostReconciler) hostUpdatePolicySetOwnerReference(info *reconcileInfo) error {
+func (r *BareMetalHostReconciler) hostUpdatePolicySetOwnerReference(info *reconcileInfo) (policy *metal3api.HostUpdatePolicy, err error) {
 	// NOTE(janders) the goal here is to ensure that the controller reads the hup resource and adds OwnerReference to it
 	hup := &metal3api.HostUpdatePolicy{}
 	if err := r.Get(info.ctx, info.request.NamespacedName, hup); err != nil {
@@ -1972,23 +1964,23 @@ func (r *BareMetalHostReconciler) hostUpdatePolicySetOwnerReference(info *reconc
 			// garbage collected. For additional cleanup logic use
 			// finalizers.  Return and don't requeue
 
-			return nil
+			return nil, nil
 		}
 		// Error reading the object
-		return fmt.Errorf("could not load hostUpdatePolicy resource due to %w", err)
+		return nil, fmt.Errorf("could not load hostUpdatePolicy resource due to %w", err)
 	}
 	if !ownerReferenceExists(info.host, hup) {
 		if err := controllerutil.SetOwnerReference(info.host, hup, r.Scheme()); err != nil {
-			return fmt.Errorf("could not set bmh as owner for hostUpdatePolicy due to %w", err)
+			return hup, fmt.Errorf("could not set bmh as owner for hostUpdatePolicy due to %w", err)
 		}
 		if err := r.Update(info.ctx, hup); err != nil {
-			return fmt.Errorf("failure updating hostUpdatePolicy resource due to %w", err)
+			return hup, fmt.Errorf("failure updating hostUpdatePolicy resource due to %w", err)
 		}
 
-		return nil
+		return hup, nil
 	}
 
-	return nil
+	return hup, nil
 }
 
 // Get the stored firmware settings if there are valid changes.
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -1357,7 +1357,7 @@ func (r *BareMetalHostReconciler) actionDeprovisioning(prov provisioner.Provisio
 	return actionComplete{}
 }
 
-func (r *BareMetalHostReconciler) doServiceIfNeeded(prov provisioner.Provisioner, info *reconcileInfo, hup *metal3api.HostUpdatePolicy) (result actionResult, isServicing bool) {
+func (r *BareMetalHostReconciler) doServiceIfNeeded(prov provisioner.Provisioner, info *reconcileInfo, hup *metal3api.HostUpdatePolicy) (result actionResult) {
 	servicingData := provisioner.ServicingData{}
 
 	// (NOTE)janders: since Servicing is an opt-in feature that requires HostUpdatePolicy to be created and set to onReboot
@@ -1382,7 +1382,7 @@ func (r *BareMetalHostReconciler) doServiceIfNeeded(prov provisioner.Provisioner
 		var err error
 		hfsDirty, hfs, err = r.getHostFirmwareSettings(info)
 		if err != nil {
-			return actionError{fmt.Errorf("could not determine updated settings: %w", err)}, false
+			return actionError{fmt.Errorf("could not determine updated settings: %w", err)}
 		}
 		if hfsDirty {
 			servicingData.ActualFirmwareSettings = hfs.Status.Settings
@@ -1395,41 +1395,47 @@ func (r *BareMetalHostReconciler) doServiceIfNeeded(prov provisioner.Provisioner
 	// Even if settings are clean, we need to check the result of the current servicing.
 	if !dirty && info.host.Status.OperationalStatus != metal3api.OperationalStatusServicing && info.host.Status.ErrorType != metal3api.ServicingError {
 		// If nothing is going on, return control to the power management.
-		return nil, false
+		return nil
 	}
 
+	// FIXME(janders/dtantsur): this implementation may lead to a scenario where if we never actually
+	// succeed before leaving this state (e.g. by deprovisioning) we lose the signal that the
+	// update didn't actually happen. This is deemed an acceptable risk for the moment since it is only
+	// going to impact a small subset of Firmware Settings implementations.
+	currentError := info.host.Status.ErrorType
 	provResult, started, err := prov.Service(servicingData, dirty,
-		info.host.Status.ErrorType == metal3api.ServicingError)
+		currentError == metal3api.ServicingError)
 	if err != nil {
-		return actionError{fmt.Errorf("error servicing host: %w", err)}, false
+		return actionError{fmt.Errorf("error servicing host: %w", err)}
 	}
 	if provResult.ErrorMessage != "" {
 		result = recordActionFailure(info, metal3api.ServicingError, provResult.ErrorMessage)
-		return result, true
+		return result
 	}
-	if started && clearErrorWithStatus(info.host, metal3api.OperationalStatusServicing) {
-		if fwDirty {
-			info.host.Status.Provisioning.Firmware = info.host.Spec.Firmware.DeepCopy()
-		}
+
+	if clearErrorWithStatus(info.host, metal3api.OperationalStatusServicing) {
 		dirty = true
+	}
+
+	if started && fwDirty {
+		info.host.Status.Provisioning.Firmware = info.host.Spec.Firmware.DeepCopy()
 	}
 
 	if provResult.Dirty {
 		result := actionContinue{provResult.RequeueAfter}
 		if dirty {
-			return actionUpdate{result}, true
+			return actionUpdate{result}
 		}
-		return result, true
+		return result
 	}
 
 	// Servicing is finished at this point, clean up operational status
 	if clearErrorWithStatus(info.host, metal3api.OperationalStatusOK) {
-		// We need to give the HostFirmwareSettings controller some time to
-		// catch up with the changes, otherwise we risk starting the same
-		// operation again.
-		return actionUpdate{actionContinue{delay: subResourceNotReadyRetryDelay}}, true
+		// FIXME(janders/dtantsur): this can be racy. We should consider
+		// using a generation number to decide if we start servicing or not.
+		return actionUpdate{actionContinue{delay: subResourceNotReadyRetryDelay}}
 	}
-	return nil, false
+	return nil
 }
 
 // Check the current power status against the desired power status.
@@ -1445,7 +1451,7 @@ func (r *BareMetalHostReconciler) manageHostPower(prov provisioner.Provisioner, 
 	if hwState.PoweredOn != nil && *hwState.PoweredOn != info.host.Status.PoweredOn {
 		info.log.Info("updating power status", "discovered", *hwState.PoweredOn)
 		info.host.Status.PoweredOn = *hwState.PoweredOn
-		if info.host.Status.OperationalStatus == metal3api.OperationalStatusError || info.host.Status.ErrorType == metal3api.PowerManagementError {
+		if info.host.Status.OperationalStatus == metal3api.OperationalStatusError && info.host.Status.ErrorType == metal3api.PowerManagementError {
 			clearError(info.host)
 		}
 		return actionUpdate{}
@@ -1456,6 +1462,8 @@ func (r *BareMetalHostReconciler) manageHostPower(prov provisioner.Provisioner, 
 	provState := info.host.Status.Provisioning.State
 	// Normal reboots only work in provisioned states, changing online is also possible for available hosts.
 	isProvisioned := provState == metal3api.StateProvisioned || provState == metal3api.StateExternallyProvisioned
+	// FIXME(janders/dtantsur) it would be preferrable to pass in state as an argument
+	// however this falls outside the scope of this specific change.
 
 	if !info.host.Status.PoweredOn {
 		if _, suffixlessAnnotationExists := info.host.Annotations[metal3api.RebootAnnotationPrefix]; suffixlessAnnotationExists {
@@ -1468,7 +1476,7 @@ func (r *BareMetalHostReconciler) manageHostPower(prov provisioner.Provisioner, 
 			return actionContinue{}
 		}
 	}
-	hup, err := r.hostUpdatePolicySetOwnerReference(info)
+	hup, err := r.acquireHostUpdatePolicy(info)
 	if err != nil {
 		info.log.Info("failed setting owner reference on hostupdatepolicy")
 		return actionError{errors.Wrap(err, "failed setting owner reference on hostUpdatePolicy")}
@@ -1476,7 +1484,7 @@ func (r *BareMetalHostReconciler) manageHostPower(prov provisioner.Provisioner, 
 
 	servicingAllowed := isProvisioned && !info.host.Status.PoweredOn && desiredPowerOnState
 	if servicingAllowed || info.host.Status.OperationalStatus == metal3api.OperationalStatusServicing || info.host.Status.ErrorType == metal3api.ServicingError {
-		result, _ := r.doServiceIfNeeded(prov, info, hup)
+		result := r.doServiceIfNeeded(prov, info, hup)
 		if result != nil {
 			return result
 		}
@@ -1954,7 +1962,7 @@ func (r *BareMetalHostReconciler) createHostFirmwareSettings(info *reconcileInfo
 	return nil
 }
 
-func (r *BareMetalHostReconciler) hostUpdatePolicySetOwnerReference(info *reconcileInfo) (policy *metal3api.HostUpdatePolicy, err error) {
+func (r *BareMetalHostReconciler) acquireHostUpdatePolicy(info *reconcileInfo) (policy *metal3api.HostUpdatePolicy, err error) {
 	// NOTE(janders) the goal here is to ensure that the controller reads the hup resource and adds OwnerReference to it
 	hup := &metal3api.HostUpdatePolicy{}
 	if err := r.Get(info.ctx, info.request.NamespacedName, hup); err != nil {
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -1403,6 +1403,11 @@ func (r *BareMetalHostReconciler) doServiceIfNeeded(prov provisioner.Provisioner
 	// update didn't actually happen. This is deemed an acceptable risk for the moment since it is only
 	// going to impact a small subset of Firmware Settings implementations.
 	currentError := info.host.Status.ErrorType
+	if info.host.Status.OperationalStatus != metal3api.OperationalStatusServicing {
+		info.host.Status.OperationalStatus = metal3api.OperationalStatusServicing
+		return actionUpdate{}
+	}
+
 	provResult, started, err := prov.Service(servicingData, dirty,
 		currentError == metal3api.ServicingError)
 	if err != nil {
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -321,6 +321,8 @@ func recordActionFailure(info *reconcileInfo, errorType metal3api.ErrorType, err
 		metal3api.InspectionError:              "InspectionError",
 		metal3api.ProvisioningError:            "ProvisioningError",
 		metal3api.PowerManagementError:         "PowerManagementError",
+		metal3api.PreparationError:             "PreparationError",
+		metal3api.ServicingError:               "ServicingError",
 	}[errorType]
 
 	counter := actionFailureCounters.WithLabelValues(eventType)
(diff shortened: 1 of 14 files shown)


pkg/provisioner/ironic/servicing.go
```diff
@@ -0,0 +1,124 @@
...
+				Interface: nodes.InterfaceBIOS,
+				Step:      "apply_configuration",
+				Args: map[string]interface{}{
+					"settings": newSettings,
+				},
+			},
+		)
+	}
+
+	// TODO: Add service steps for firmware updates
```
dtantsur, line 38: TODO needs to be resolved, although you can do it in the next change.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1355,6 +1362,93 @@ func (r *BareMetalHostReconciler) actionDeprovisioning(prov provisioner.Provisio
...
+	// (NOTE)janders: since Servicing is an opt-in feature that requires HostUpdatePolicy to be created and set to onReboot
+	// set below booleans to false by default and change to true based on policy settings
+
+	servicingData.LiveFirmwareUpdateAllowed = false
+	servicingData.LiveFirmwareSettingsAllowed = false
+
+	hup := &metal3api.HostUpdatePolicy{}
+	err := r.Get(info.ctx, info.request.NamespacedName, hup)
+	if err != nil {
+		return actionError{fmt.Errorf("unable to fetch HostUpdatePolicy, aborting servicing. Error: %w", err)}, false
```
dtantsur, line 1380: This will fail if the policy does not exist.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1355,6 +1357,87 @@ func (r *BareMetalHostReconciler) actionDeprovisioning(prov provisioner.Provisio
...
+	if hup != nil {
+		if hup.Spec.FirmwareSettings == metal3api.HostUpdatePolicyOnReboot {
+			servicingData.LiveFirmwareSettingsAllowed = true
+		}
+		if hup.Spec.FirmwareUpdates == metal3api.HostUpdatePolicyOnReboot {
+			servicingData.LiveFirmwareUpdateAllowed = true
+		}
+	}
+
+	if servicingData.LiveFirmwareUpdateAllowed {
```
dtantsur, line 1381: Btw, the way code is written, you don't need to store the *Allowed variables in servicingData: you fully use them locally.

pkg/provisioner/ironic/servicing.go
iurygregory: Yeah, I will push a PR for this using this one as base.

pkg/provisioner/provisioner.go
```diff
@@ -107,6 +107,15 @@ type PrepareData struct {
 	TargetFirmwareComponents []metal3api.FirmwareUpdate
 }
 
+type ServicingData struct {
+	FirmwareConfig              *metal3api.FirmwareConfig
+	TargetFirmwareSettings      metal3api.DesiredSettingsMap
+	ActualFirmwareSettings      metal3api.SettingsMap
+	LiveFirmwareUpdateAllowed   bool
+	LiveFirmwareSettingsAllowed bool
```
dtantsur, line 115: These two fiels are not necessary since they are only used locally. You can just create two local boolean variables in checkServicing.

controllers/metal3.io/baremetalhost_controller_test.go
```diff
@@ -795,6 +796,63 @@ func TestRebootWithSuffixedAnnotation(t *testing.T) {
...
+	host.Status.PoweredOn = true
+	host.Status.Provisioning.State = metal3api.StateProvisioned
+	host.Spec.Online = true
+	host.Spec.Image = &metal3api.Image{URL: "foo", Checksum: "123"}
+	host.Spec.Image.URL = "foo"
+	host.Spec.Firmware = &metal3api.FirmwareConfig{
+		VirtualizationEnabled: ptr.To(true),
+	}
+	host.Status.Provisioning.Image.URL = "foo"
+
```
dtantsur, line 814: A policy object must be created here, otherwise the test will never pass.

controllers/metal3.io/baremetalhost_controller_test.go
```diff
@@ -795,6 +796,63 @@ func TestRebootWithSuffixedAnnotation(t *testing.T) {
...
+	r := newTestReconciler(host)
```
dtantsur, line 815: I suspect you need to pass the HostUpdatePolicy object as the 2nd argument here.

controllers/metal3.io/baremetalhost_controller_test.go
rhjanders: Agreed, we just haven't figured out how to pass it in yet. The comment below suggests a promising path. In terms of creating the HUP object, do I need to do something similar to this: https://github.com/metal3-io/baremetal-operator/blob/main/controllers/metal3.io/baremetalhost_controller_test.go#L2319-L2356 ?

pkg/provisioner/provisioner.go
rhjanders: Agreed, will apply this suggestion in the next revision.

controllers/metal3.io/baremetalhost_controller_test.go
dtantsur: Yes. This test is actually very similar to what you need: it also creates another object in addition to the BMH.

rhjanders: thank you Dmitry, will try this later tonight or tomorrow

controllers/metal3.io/baremetalhost_controller_test.go
```diff
@@ -795,6 +797,80 @@ func TestRebootWithSuffixedAnnotation(t *testing.T) {
...
+	host.Spec.Image.URL = "foo"
+	host.Spec.Firmware = &metal3api.FirmwareConfig{
+		VirtualizationEnabled: ptr.To(true),
+	}
+	host.Status.Provisioning.Image.URL = "foo"
+	hup := &metal3api.HostUpdatePolicy{
+			TypeMeta: metav1.TypeMeta{
+				Kind: "HostUpdatePolicy",
+				APIVersion: "metal3.io/v1alpha1",
+		},
```
dtantsur, line 819: nit: I suspect TypeMeta is redundant (implied by the structure type)

controllers/metal3.io/baremetalhost_controller_test.go
```diff
@@ -795,6 +797,80 @@ func TestRebootWithSuffixedAnnotation(t *testing.T) {
...
+			Namespace: namespace,
+		},
+		Spec: metal3api.HostUpdatePolicySpec{
+			FirmwareSettings: metal3api.HostUpdatePolicyOnReboot,
+			FirmwareUpdates: metal3api.HostUpdatePolicyOnReboot,
+		},
+	}
+	// hup.ResourceVersion = ""
+	r := newTestReconciler(host, hup)
+	err := r.Update(context.TODO(), hup)
```
dtantsur, line 831: nit: I'm not sure you need to explicit create the object after adding it to the test reconciler.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1355,6 +1357,70 @@ func (r *BareMetalHostReconciler) actionDeprovisioning(prov provisioner.Provisio
...
+func (r *BareMetalHostReconciler) checkServicing(prov provisioner.Provisioner, info *reconcileInfo, hup *metal3api.HostUpdatePolicy) (result actionResult, isServicing bool) {
+	servicingData := provisioner.ServicingData{}
+
+	var dirty bool
+
+	// (NOTE)janders: since Servicing is an opt-in feature that requires HostUpdatePolicy to be created and set to onReboot
+	// set below booleans to false by default and change to true based on policy settings
+
+	var LiveFirmwareSettingsAllowed bool
+	LiveFirmwareSettingsAllowed = false
```
dtantsur, line 1369: nit: redundant (false by default)

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1355,6 +1357,70 @@ func (r *BareMetalHostReconciler) actionDeprovisioning(prov provisioner.Provisio
...
+
+	if hup != nil {
+		if hup.Spec.FirmwareSettings == metal3api.HostUpdatePolicyOnReboot {
+			LiveFirmwareSettingsAllowed = true
```
dtantsur, line 1373: nit: 

LiveFirmwareSettingsAllowed = (hup.Spec.FirmwareSettings == metal3api.HostUpdatePolicyOnReboot)

controllers/metal3.io/baremetalhost_controller_test.go
```diff
@@ -795,6 +797,80 @@ func TestRebootWithSuffixedAnnotation(t *testing.T) {
...
+func TestRebootWithServicing(t *testing.T) {
+	host := newDefaultHost(t)
+	host.Annotations = make(map[string]string)
+	host.Annotations[metal3api.RebootAnnotationPrefix] = ""
+	host.Status.PoweredOn = true
+	host.Status.Provisioning.State = metal3api.StateProvisioned
+	host.Spec.Online = true
+	host.Spec.Image = &metal3api.Image{URL: "foo", Checksum: "123"}
+	host.Spec.Image.URL = "foo"
+	host.Spec.Firmware = &metal3api.FirmwareConfig{
```
dtantsur, line 811: Looking at the recent version of checkServicing, is this field even handled any more? Remember: there are two firmware settings API, I think you're only handling HFS now?

rhjanders: Good pickup. Does this mean I need to add an HFS CR mockup as well?

rhjanders: I tried a couple things but still without luck - if you have time I'd appreciate pointers. I tried different things (some still present but commented out) but I'm still failing going over 25 iterations.

dtantsur: You don't have this part https://github.com/metal3-io/baremetal-operator/pull/1689/files#diff-a69d9029388ab766ed36b32180145f52785a9d4a153775510dbddfa928a72e1cR1344-R1347 from my PoC patch. Without it, setting `Spec.Firmware` does nothing regardless of servicing and HUP.

dtantsur: (and you probably need another test for HFS case)

rhjanders: w/r/t to missing parts - that was a massive stuff-up and misunderstanding. Once I dropped that commit, with the short assignment fix and with the understanding of old firmware settings API I finally managed to get things to pass locally. Let's see what CI thinks.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1355,6 +1357,84 @@ func (r *BareMetalHostReconciler) actionDeprovisioning(prov provisioner.Provisio
...
 
+func (r *BareMetalHostReconciler) checkServicing(prov provisioner.Provisioner, info *reconcileInfo, hup *metal3api.HostUpdatePolicy) (result actionResult, isServicing bool) {
+	servicingData := provisioner.ServicingData{}
+
+	// (NOTE)janders: since Servicing is an opt-in feature that requires HostUpdatePolicy to be created and set to onReboot
+	// set below booleans to false by default and change to true based on policy settings
+
+	var fwDirty bool
+	var hfsDirty bool
+	var LiveFirmwareUpdatesAllowed bool
```
dtantsur, line 1368: nit: local variables start with lower case (I'm not sure why the linter does not complain)

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1355,6 +1357,84 @@ func (r *BareMetalHostReconciler) actionDeprovisioning(prov provisioner.Provisio
...
+	var LiveFirmwareSettingsAllowed bool
+
+	if hup != nil {
+		LiveFirmwareUpdatesAllowed = (hup.Spec.FirmwareUpdates == metal3api.HostUpdatePolicyOnReboot)
+		LiveFirmwareSettingsAllowed = (hup.Spec.FirmwareSettings == metal3api.HostUpdatePolicyOnReboot)
+	}
+
+	if LiveFirmwareUpdatesAllowed {
```
dtantsur, line 1376: This is not firmware updates, it should be grouped with other firmware settings

controllers/metal3.io/baremetalhost_controller_test.go
```diff
@@ -795,6 +796,75 @@ func TestRebootWithSuffixedAnnotation(t *testing.T) {
...
+	)
+
+	// make sure we don't go into another reboot
+	tryReconcile(t, r, host,
+		func(host *metal3api.BareMetalHost, result reconcile.Result) bool {
+			return host.Status.PoweredOn
+		},
+	)
+}
+
```
dtantsur, line 867: Let's have a unit test that shows that without a HUP object, servicing does not happen even if there are changes.

pkg/provisioner/provisioner.go
dtantsur: Not resolved, the variables should be deleted from here

iurygregory:
LGTM, thanks for working on it @rhjanders

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1355,6 +1362,66 @@ func (r *BareMetalHostReconciler) actionDeprovisioning(prov provisioner.Provisio
 	return actionComplete{}
 }
 
+func (r *BareMetalHostReconciler) checkServicing(prov provisioner.Provisioner, info *reconcileInfo) (result actionResult, isServicing bool) {
```
zaneb: This function does more than check.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1355,6 +1362,66 @@ func (r *BareMetalHostReconciler) actionDeprovisioning(prov provisioner.Provisio
...
+	// Even if settings are clean, we need to check the result of the current servicing.
+	if !dirty && info.host.Status.OperationalStatus != metal3api.OperationalStatusServicing && info.host.Status.ErrorType != metal3api.ServicingError {
+		// If nothing is going on, return control to the power management.
+		return nil, false
+	}
+
+	provResult, started, err := prov.Service(servicingData, dirty,
+		info.host.Status.ErrorType == metal3api.ServicingError)
+	if err != nil {
+		return actionError{fmt.Errorf("error servicing host: %w", err)}, false
```
zaneb, line 1404: Currently this is ignored as if you had returned nil.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1355,6 +1362,66 @@ func (r *BareMetalHostReconciler) actionDeprovisioning(prov provisioner.Provisio
...
+	}
+	if provResult.ErrorMessage != "" {
+		result = recordActionFailure(info, metal3api.ServicingError, provResult.ErrorMessage)
+		return result, true
+	}
+	if started && clearErrorWithStatus(info.host, metal3api.OperationalStatusServicing) {
+		if fwDirty {
+			info.host.Status.Provisioning.Firmware = info.host.Spec.Firmware.DeepCopy()
```
zaneb, line 1412: Updating as soon as we've started rather than when we've finished means that if we never actually succeed before leaving this state (e.g. by deprovisioning) we lose the signal that the update didn't actually happen.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1355,6 +1362,66 @@ func (r *BareMetalHostReconciler) actionDeprovisioning(prov provisioner.Provisio
...
+			return actionUpdate{result}, true
+		}
+		return result, true
+	}
+
+	// Servicing is finished at this point, clean up operational status
+	if clearErrorWithStatus(info.host, metal3api.OperationalStatusOK) {
+		// We need to give the HostFirmwareSettings controller some time to
+		// catch up with the changes, otherwise we risk starting the same
+		// operation again.
```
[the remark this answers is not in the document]
dtantsur: Then we need to rethink the whole architecture of having separate controllers for HFS, HFC, DataImage and so on (in fact, a similar race is reported for DataImages now).

controllers/metal3.io/baremetalhost_controller.go
dtantsur: This is how it works for non-servicing case too. Unfortunately, for this older firmware approach, we have no way to learn from Ironic if it was applied or not.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1355,6 +1362,66 @@ func (r *BareMetalHostReconciler) actionDeprovisioning(prov provisioner.Provisio
...
+	provResult, started, err := prov.Service(servicingData, dirty,
+		info.host.Status.ErrorType == metal3api.ServicingError)
+	if err != nil {
+		return actionError{fmt.Errorf("error servicing host: %w", err)}, false
+	}
+	if provResult.ErrorMessage != "" {
+		result = recordActionFailure(info, metal3api.ServicingError, provResult.ErrorMessage)
+		return result, true
+	}
+	if started && clearErrorWithStatus(info.host, metal3api.OperationalStatusServicing) {
```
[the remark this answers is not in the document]
dtantsur: If we do this, we cannot check for `info.host.Status.ErrorType` on line 1402.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1386,9 +1461,13 @@ func (r *BareMetalHostReconciler) manageHostPower(prov provisioner.Provisioner,
 		}
 	}
 
-	provState := info.host.Status.Provisioning.State
-	// Normal reboots only work in provisioned states, changing online is also possible for available hosts.
-	isProvisioned := provState == metal3api.StateProvisioned || provState == metal3api.StateExternallyProvisioned
+	servicingAllowed := isProvisioned && !info.host.Status.PoweredOn && desiredPowerOnState
+	if servicingAllowed || info.host.Status.OperationalStatus == metal3api.OperationalStatusServicing || info.host.Status.ErrorType == metal3api.ServicingError {
+		result, isServicing := r.checkServicing(prov, info)
+		if result != nil && (result.Dirty() || isServicing) {
```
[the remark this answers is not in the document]
dtantsur: It could be an artefact of an earlier iteration of this PR. On a fresh sight, `result != nil` should be enough.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1386,9 +1461,13 @@ func (r *BareMetalHostReconciler) manageHostPower(prov provisioner.Provisioner,
...
+			return result
+		}
+	}
 
 	desiredReboot, desiredRebootMode := hasRebootAnnotation(info, !isProvisioned)
```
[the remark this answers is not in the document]
dtantsur: It does not matter: in this case Ironic orchestrates the reboot itself (which is a problem of its own, but let's leave it for another day).

controllers/metal3.io/baremetalhost_controller.go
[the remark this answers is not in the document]
dtantsur: Okay, we can probably cache restartOnFailure in a local variable before we clear the error. That will do, I guess?

controllers/metal3.io/baremetalhost_controller.go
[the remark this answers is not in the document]
dtantsur: The alternative, I guess, it to pass the HFS object all the way here and update its status directly. It's not great either (responsibility violation). But I guess better then just rescheduling and hoping for better.

pkg/provisioner/provisioner.go
rhjanders: apologies, I must have accidentally added them back in while rebasing. Looks like they are gone now.

pkg/provisioner/ironic/servicing.go
```diff
@@ -0,0 +1,124 @@
...
+			},
+		)
+	}
+	result, err = operationComplete()
+	return
+}
+
+func (p *ironicProvisioner) Service(data provisioner.ServicingData, unprepared, restartOnFailure bool) (result provisioner.Result, started bool, err error) {
+	if !p.availableFeatures.HasServicing() {
+		result, err = transientError(fmt.Errorf("servicing not supported: requires API version 1.87, available is 1.%d", p.availableFeatures.MaxVersion))
```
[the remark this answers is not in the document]
rhjanders: Done

controllers/metal3.io/baremetalhost_controller.go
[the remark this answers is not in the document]
rhjanders: Done

controllers/metal3.io/baremetalhost_controller.go
rhjanders: renamed to doServiceIfNeeded()

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1368,12 +1435,20 @@ func (r *BareMetalHostReconciler) manageHostPower(prov provisioner.Provisioner,
 	if hwState.PoweredOn != nil && *hwState.PoweredOn != info.host.Status.PoweredOn {
 		info.log.Info("updating power status", "discovered", *hwState.PoweredOn)
 		info.host.Status.PoweredOn = *hwState.PoweredOn
-		clearError(info.host)
+		targetOperationalStatus := metal3api.OperationalStatusOK
+		if info.host.Status.OperationalStatus == metal3api.OperationalStatusServicing {
+			targetOperationalStatus = metal3api.OperationalStatusServicing
```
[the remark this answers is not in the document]
rhjanders: I tried to address this together with https://github.com/metal3-io/baremetal-operator/pull/1969#discussion_r1805797177 , please let me know if I got it right, if not we'll iterate over it

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1368,12 +1435,20 @@ func (r *BareMetalHostReconciler) manageHostPower(prov provisioner.Provisioner,
...
+		}
+		clearErrorWithStatus(info.host, targetOperationalStatus)
```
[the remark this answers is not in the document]
rhjanders: hopefully addressed together with https://github.com/metal3-io/baremetal-operator/pull/1969#discussion_r1805822355

controllers/metal3.io/baremetalhost_controller.go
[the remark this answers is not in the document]
zaneb: I'm confused now, because line 1477 says that we only start the service mode if the power state is off but we wanted it to be on. AFAIK that only occurs during a hard reboot.

controllers/metal3.io/baremetalhost_controller.go
iurygregory: If I understood correctly, we just need the return here
```
return actionError{fmt.Errorf("error servicing host: %w", err)}
```
Is this a correct assumption @zaneb ?

zaneb: Yep, due to https://github.com/metal3-io/baremetal-operator/pull/1969#discussion_r1806419875 we no longer need to return multiple values from this function at all.

controllers/metal3.io/baremetalhost_controller.go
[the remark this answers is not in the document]
dtantsur: Ah, you're probably right. No, soft reboot is totally fine and potentially even desired to prevent data loss. From that point on, Ironic will take care of the process.

controllers/metal3.io/baremetalhost_controller.go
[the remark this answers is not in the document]
rhjanders: discussion consensus: use generation number to decide if we start servicing or not. Error handling can be followed from manual cleaning / preparing state etc.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1368,12 +1435,20 @@ func (r *BareMetalHostReconciler) manageHostPower(prov provisioner.Provisioner,
...
 		return actionUpdate{}
 	}
 
 	desiredPowerOnState := info.host.Spec.Online
 
+	provState := info.host.Status.Provisioning.State
+	// Normal reboots only work in provisioned states, changing online is also possible for available hosts.
+	isProvisioned := provState == metal3api.StateProvisioned || provState == metal3api.StateExternallyProvisioned
```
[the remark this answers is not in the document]
rhjanders: consensus: follow up material potentially.

controllers/metal3.io/baremetalhost_controller.go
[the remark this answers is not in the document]
rhjanders: TODO: check and make sure it's soft poweroff not soft reboot.

controllers/metal3.io/baremetalhost_controller.go
rhjanders: TODO(janders): should be resolved in latest commit, make sure it stays this way after I reapply the other fix

controllers/metal3.io/baremetalhost_controller.go
rhjanders: at minimum, document at code, can be a follow-up material.

controllers/metal3.io/baremetalhost_controller.go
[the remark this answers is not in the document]
rhjanders: adding a FIXME note for this also

controllers/metal3.io/baremetalhost_controller.go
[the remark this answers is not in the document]
rhjanders: comment from Dmitry: We don't even have a provisioner call for rebooting - so it shouldn't be of concern.

iurygregory:
LGTM

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1355,6 +1357,86 @@ func (r *BareMetalHostReconciler) actionDeprovisioning(prov provisioner.Provisio
...
+		return result
+	}
+
+	if fwDirty {
+		info.host.Status.Provisioning.Firmware = info.host.Spec.Firmware.DeepCopy()
+	}
+
+	if provResult.Dirty {
+		result := actionContinue{provResult.RequeueAfter}
+		if dirty {
```
zaneb, line 1425: The actual thing we want to check here is whether we need to write the BMH. Writes occur on line 1406 and line 1420, but `dirty` could be true if `hfsDirty` is true even if nothing is actually updated.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1355,6 +1357,86 @@ func (r *BareMetalHostReconciler) actionDeprovisioning(prov provisioner.Provisio
...
+		// If nothing is going on, return control to the power management.
+		return nil
+	}
+
+	// FIXME(janders/dtantsur): this implementation may lead to a scenario where if we never actually
+	// succeed before leaving this state (e.g. by deprovisioning) we lose the signal that the
+	// update didn't actually happen. This is deemed an acceptable risk for the moment since it is only
+	// going to impact a small subset of Firmware Settings implementations.
+	currentError := info.host.Status.ErrorType
+	if clearErrorWithStatus(info.host, metal3api.OperationalStatusServicing) {
```
zaneb, line 1406: I think this is too early to clear an error. We should do it after the check for `provResult.ErrorMessage != ""` around line 1418.

It's also arguably too late for an attempt to set the status to Servicing in the non-error case, because we'll still only write it once servicing has already started.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1355,6 +1357,86 @@ func (r *BareMetalHostReconciler) actionDeprovisioning(prov provisioner.Provisio
...
+		if dirty {
+			return actionUpdate{result}
+		}
+		return result
+	}
+
+	// Servicing is finished at this point, clean up operational status
+	if clearErrorWithStatus(info.host, metal3api.OperationalStatusOK) {
+		// FIXME(janders/dtantsur): this can be racy. We should consider
+		// using a generation number to decide if we start servicing or not.
```
zaneb, line 1434: +1
This is a _big_ delay that we will have to wait before we can turn the power back on.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1862,7 +1961,7 @@ func (r *BareMetalHostReconciler) createHostFirmwareSettings(info *reconcileInfo
 	return nil
 }
 
-func (r *BareMetalHostReconciler) hostUpdatePolicySetOwnerReference(info *reconcileInfo) error {
+func (r *BareMetalHostReconciler) hostUpdatePolicySetOwnerReference(info *reconcileInfo) (policy *metal3api.HostUpdatePolicy, err error) {
```
zaneb, line 1964: nit: can probably come up with a more descriptive name for this function now that its responsibilities have grown.
We use `AcquireSecret()` to mean getting a Secret and setting the owner reference.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1355,6 +1357,86 @@ func (r *BareMetalHostReconciler) actionDeprovisioning(prov provisioner.Provisio
...
+
+	// FIXME(janders/dtantsur): this implementation may lead to a scenario where if we never actually
+	// succeed before leaving this state (e.g. by deprovisioning) we lose the signal that the
+	// update didn't actually happen. This is deemed an acceptable risk for the moment since it is only
+	// going to impact a small subset of Firmware Settings implementations.
+	currentError := info.host.Status.ErrorType
+	if clearErrorWithStatus(info.host, metal3api.OperationalStatusServicing) {
+		dirty = true
+	}
+	provResult, _, err := prov.Service(servicingData, dirty,
```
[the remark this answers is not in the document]
rhjanders: Reverting this and utilising started to address the comment below.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1355,6 +1357,86 @@ func (r *BareMetalHostReconciler) actionDeprovisioning(prov provisioner.Provisio
...
+		currentError == metal3api.ServicingError)
+	if err != nil {
+		return actionError{fmt.Errorf("error servicing host: %w", err)}
+	}
+	if provResult.ErrorMessage != "" {
+		result = recordActionFailure(info, metal3api.ServicingError, provResult.ErrorMessage)
+		return result
+	}
+
+	if fwDirty {
```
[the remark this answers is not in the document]
rhjanders: Done

controllers/metal3.io/baremetalhost_controller.go
rhjanders: I changed the order as per the first line of the comment. Unsure about the second part - will check in with Dmitry.

controllers/metal3.io/baremetalhost_controller.go
rhjanders: renamed hostUpdatePolicySetOwnerReference() to acquireHostUpdatePolicy()

controllers/metal3.io/baremetalhost_controller.go
zaneb: Remember that writes to the k8s API only happen after we return from this function. So if you want to put it into `OperationalStatusServicing` _before_ servicing starts then you'd need to do something like:

```
    if info.host.Status.OperationalStatus != metal3api.OperationalStatusServicing {
        info.host.Status.OperationalStatus = metal3api.OperationalStatusServicing
        return actionUpdate{}
    }
```

Just changing the order has no effect in that respect.

dtantsur: If we do it this way, we'll lose the error information, and it will never be passed to `prov.Service`.

dtantsur: I stand corrected. If we don't call `clearErrorWithStatus`, we'll have `ErrorType` still set. We just need to remember to unset it after the successful call.

Maybe we need to return `clearErrorWithStatus` to where it was before, but also set the status to servicing explicitly in the way that Zane suggested?

controllers/metal3.io/baremetalhost_controller.go
rhjanders: we'll try address this one in a follow-up change

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1355,6 +1357,92 @@ func (r *BareMetalHostReconciler) actionDeprovisioning(prov provisioner.Provisio
...
+	if !dirty && info.host.Status.OperationalStatus != metal3api.OperationalStatusServicing && info.host.Status.ErrorType != metal3api.ServicingError {
+		// If nothing is going on, return control to the power management.
+		return nil
+	}
+
+	// FIXME(janders/dtantsur): this implementation may lead to a scenario where if we never actually
+	// succeed before leaving this state (e.g. by deprovisioning) we lose the signal that the
+	// update didn't actually happen. This is deemed an acceptable risk for the moment since it is only
+	// going to impact a small subset of Firmware Settings implementations.
+	currentError := info.host.Status.ErrorType
```
dtantsur, line 1405: nit: you no longer need to store this because you no longer clear ErrorType before calling Service.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1355,6 +1357,92 @@ func (r *BareMetalHostReconciler) actionDeprovisioning(prov provisioner.Provisio
...
+		result = recordActionFailure(info, metal3api.ServicingError, provResult.ErrorMessage)
+		return result
+	}
+
+	if clearErrorWithStatus(info.host, metal3api.OperationalStatusServicing) {
+		dirty = true
+	}
+
+	if started && fwDirty {
+		info.host.Status.Provisioning.Firmware = info.host.Spec.Firmware.DeepCopy()
```
dtantsur, line 1426: You probably need `dirty = true` here

dtantsur:
/close

Superseded by https://github.com/metal3-io/baremetal-operator/pull/2041

PR #1974, opened 2024-09-27
:bug: Remove dataImage finalizer if BMH is missing

hroyrh:
This fixes the issue of dataImage resource not being deleted, if the corresponding BMH is deleted without detaching the dataImage.

### DIFF
--- a/controllers/metal3.io/dataimage_controller.go
+++ b/controllers/metal3.io/dataimage_controller.go
@@ -36,8 +36,9 @@ import (
 )
 
 const (
-	dataImageRetryDelay  = time.Second * 60
-	dataImageUpdateDelay = time.Second * 30
+	dataImageRetryDelay          = time.Second * 60
+	dataImageUpdateDelay         = time.Second * 30
+	dataImageUnmanagedRetryDelay = time.Second * 20
 )
 
 // DataImageReconciler reconciles a DataImage object.
@@ -129,7 +130,7 @@ func (r *DataImageReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 
 	if hasDetachedAnnotation(bmh) {
 		reqLogger.Info("the host is detached, not running reconciler")
-		return ctrl.Result{Requeue: true, RequeueAfter: unmanagedRetryDelay}, nil
+		return ctrl.Result{Requeue: true, RequeueAfter: dataImageUnmanagedRetryDelay}, nil
 	}
 
 	// If the reconciliation is paused, requeue
--- a/controllers/metal3.io/dataimage_controller.go
+++ b/controllers/metal3.io/dataimage_controller.go
@@ -111,7 +111,13 @@ func (r *DataImageReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	if err := r.Get(ctx, req.NamespacedName, bmh); err != nil {
 		// There might not be any BareMetalHost for the DataImage
 		if k8serrors.IsNotFound(err) {
-			reqLogger.Info("bareMetalHost not found for the dataImage")
+			reqLogger.Info("bareMetalHost not found for the dataImage, remove finalizer if it exists")
+			di.Finalizers = utils.FilterStringFromList(
+				di.Finalizers, metal3api.DataImageFinalizer)
+
+			if err := r.Update(ctx, di); err != nil {
+				return ctrl.Result{Requeue: true, RequeueAfter: dataImageRetryDelay}, fmt.Errorf("failed to update resource after remove finalizer, %w", err)
+			}
 			return ctrl.Result{}, nil
 		}
 


eranco74 (commented):
/lgtm

dtantsur:
This is an interesting edge case. Is it because of a forced deletion of the BMH?

hroyrh:
> This is an interesting edge case. Is it because of a forced deletion of the BMH?

Do you mean removing BMH's finalizer ? I tested with both - removing the finalizer and waiting for BMH to deprovision, after triggering BMH delete by running `kubectl delete bmh node-name`. In both cases, the DataImage was left orphaned.

dtantsur:
> waiting for BMH to deprovision, after triggering BMH delete by running kubectl delete bmh node-name

At least this case should be handled by the BMH controller, not the DataImage controller, I think. ALso, don't we disconnect the ISO on a "normal" deprovisioning?

hroyrh:
> At least this case should be handled by the BMH controller, not the DataImage controller, I think. ALso, don't we disconnect the ISO on a "normal" deprovisioning?

Why does that make more sense, won't a force deletion of BMH ( by removing the finalizer ) leave the dataImage orphaned in that case ?
I need to verify whether ISO disconnection happens on a "normal" deprovisioning.

hroyrh:
> Your PR handles forced deletion, but I don't think it handles the normal deprovisioning case? The BMH controller must decide what to do with DataImages it owns. Either remove the finalizer (esp. when the node is detached) or disconnect the ISO first?

Makes sense, I will work on the detached and normal delete process. What about the forced deletion scenario, should I keep the current DataImage controller fix in place ?

eranco74:
> ested with both - removing the finalizer and waiting for BMH to deprovision, after triggering BMH delete by running `kub`

not forced, BMH deletion isn't blocked on dataImage existence or attachment.
You can delete a BMH attached to dataImage, but then you end up with a dataImage you can't delete

dtantsur:
> What about the forced deletion scenario, should I keep the current DataImage controller fix in place ?

I think it's a fair to remove the finalizer if the controller sees that the owner BMH is gone.

mquhuy:
lgtm, but could we have a couple of e2e tests related to this?

iurygregory:
lgtm, maybe a separate PR with the e2e would be ok (I do think it would be important)

hroyrh:
Right, I can add the e2e tests in a separate PR.

elfosardo:
/approve
let's add the e2e tests as soon as this merges

tuminoid:
> Right, I can add the e2e tests in a separate PR.

https://github.com/metal3-io/baremetal-operator/issues/2053 so they're not forgotten, or at least are tracked.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -520,6 +526,11 @@ func (r *BareMetalHostReconciler) actionDeleting(prov provisioner.Provisioner, i
 		"timestamp", info.host.DeletionTimestamp,
 	)
 
+	// Deleting DataImage if it exists for the BMH
+	if err := r.removeFinalizerDataImage(info); err != nil {
+		info.log.Error(err, "Failed to delete DataImage")
```
dtantsur, line 531: Why do we ignore the error here? I'm fine with ignoring HTTP 404 (in case somebody did a nasty force-deletion), but it seems that ignoring any errors may result with orphaned DataImage.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -593,6 +604,11 @@ func hasCustomDeploy(host *metal3api.BareMetalHost) bool {
 
 // detachHost() detaches the host from the Provisioner.
 func (r *BareMetalHostReconciler) detachHost(prov provisioner.Provisioner, info *reconcileInfo) actionResult {
+	// Remove DataImage finalizer if the host has been detached
+	if err := r.removeFinalizerDataImage(info); err != nil {
+		info.log.Info("Failed to remove DataImage finalizer, %w", "Error", err)
```
dtantsur, line 609: Same question.

Also, %w does nothing, there is no string interpolation here.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1415,7 +1433,9 @@ func (r *BareMetalHostReconciler) manageHostPower(prov provisioner.Provisioner,
 		"reboot process", desiredPowerOnState != info.host.Spec.Online)
 
 	if desiredPowerOnState {
-		if isProvisioned {
+		// If the host is provisioned/externallyProvisioned or if its in
+		// the available state after deprovisioning.
```
dtantsur, line 1437: Won't this cause the image attachment to also happen in the available state?

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1493,6 +1513,21 @@ func (r *BareMetalHostReconciler) handleDataImageActions(prov provisioner.Provis
 		return actionContinue{}
 	}
 
+	bmhProvState := info.host.Status.Provisioning.State
```
dtantsur, line 1516: nit: unnecessary variable

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1415,7 +1433,9 @@ func (r *BareMetalHostReconciler) manageHostPower(prov provisioner.Provisioner,
 		"reboot process", desiredPowerOnState != info.host.Spec.Online)
 
 	if desiredPowerOnState {
-		if isProvisioned {
+		// If the host is provisioned/externallyProvisioned or if its in
+		// the available state after deprovisioning.
+		if isProvisioned || isAvailable {
```
dtantsur, line 1438: nit: I'm not sure we can enter this function in any other state.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1493,6 +1513,21 @@ func (r *BareMetalHostReconciler) handleDataImageActions(prov provisioner.Provis
 		return actionContinue{}
 	}
 
+	bmhProvState := info.host.Status.Provisioning.State
+	isBMHAvailable := bmhProvState == metal3api.StateAvailable
```
dtantsur, line 1517: nit: I'd prefer this function to have an argument like `detachOnly bool` and keep the provision state logic at the caller side (where you already check for the Available state).

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1547,10 +1582,20 @@ func (r *BareMetalHostReconciler) handleDataImageActions(prov provisioner.Provis
...
+		if isBMHAvailable {
+			detachReason = "BMH Deprovision finished, node in available state"
+		}
+		if hostDeleteTriggered {
+			detachReason = "BMH requested to be deleted"
+		}
+		info.log.Info("Requesting DataImage detach", "Reason", detachReason)
 		if attachedURL != "" || dirty {
-			info.log.Info("Detaching DataImage as it was deleted")
+			info.log.Info("Detaching DataImage, Reason = %w", "Reason", detachReason)
```
dtantsur, line 1598: The %w part will be logged verbatim, there is no string interpolation here.

controllers/metal3.io/baremetalhost_controller.go
dtantsur: Thinking more about it, I don't believe using Available is the right thing to do. The image won't be disconnected until a reboot, which will happen.. who knows when. I think we should start deprovisioning with deleting a data image.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1547,10 +1582,20 @@ func (r *BareMetalHostReconciler) handleDataImageActions(prov provisioner.Provis
...
 
-	if deleteDataImage {
-		info.log.Info("DataImage requested for deletion")
+	if deleteDataImage || detachDataImage {
+		detachReason := ""
+		if deleteDataImage {
+			detachReason = "Delete DataImage"
+		}
+		if isBMHAvailable {
+			detachReason = "BMH Deprovision finished, node in available state"
```
dtantsur, line 1591: Are we going to leave a stale DataImage object in this case? If yes, it will behave unpredictably on the next deployment because it will still have a non-empty URL.

dtantsur (changes requested):
Some comments inline.

On top of that, I think this PR tries to solve at least 3 related but distinct problem (removing DataImage on detachment, on deletion, fixing orphaned DataImage objects). I'd rather have 3 PRs.

controllers/metal3.io/baremetalhost_controller.go
hroyrh: You mean just deleting the data image and not detaching it, right ?

controllers/metal3.io/baremetalhost_controller.go
hroyrh: Makes sense, I will make the changes.

controllers/metal3.io/baremetalhost_controller.go
hroyrh: Good point, I will also set the `spec.url: ""`.

dtantsur (approved):

PR #1980, opened 2024-10-02
:seedling: Revert "use ubuntu-latest rather than hosted 4-core runner"

tuminoid:
This reverts commit 6aa2b8e71727c1e550e99cafe727e9b0f15dc1e0.

ubuntu-latest only has 14GB disk, while ubuntu-latest-4-cores has 140GB. Our optional e2e tests need more disk for the VM disks than ubuntu-latest can offer, hence reverting.


Rozzii (commented):
/lgtm

PR #1981, opened 2024-10-03
🌱 Create new file with status and inspect annotations merged

SafeEHA:
This PR merges status Annotations and Inspect Annotations file into one file now titled annotations in Baremetal-operator docs:

This issue fixes [#314](https://github.com/metal3-io/metal3-docs/issues/314) in metal3-docs repo

@Rozzii @dtantsur  kindly review this, I'll also like to confirm if this file should be moved to metal3-docs repo?

### DIFF
--- a/docs/annotations.md
+++ b/docs/annotations.md
@@ -1,0 +1,121 @@
+# Status and Inspect Annotations
+
+## Status Annotation
+Baremetalhost's(BMH) _Status_ sub-resource contain a handful of critical data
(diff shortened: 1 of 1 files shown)


tuminoid:
I think the point was not moving them around in here, but to move these to metal3-docs to be part of user-guide, and then point there from here.
/hold

SafeEHA:
Thanks for pointing this out @dtantsur, i'll do just that now

PR #1983, opened 2024-10-04
🌱 Fix minikube restart in e2e by trying in a loop

kashifest:
This is a workaround for a long existing issue in minikube which we opened long ago but it seems it has not gone away. Here is the issue for more detail https://github.com/kubernetes/minikube/issues/14456

### DIFF
--- a/hack/ci-e2e.sh
+++ b/hack/ci-e2e.sh
@@ -70,7 +70,9 @@ virsh -c qemu:///system attach-interface --domain minikube --mac="52:54:00:6c:3c
 
 # Restart minikube to apply the changes
 minikube stop
-minikube start
+## Following loop is to avoid minikube restart issue
+## https://github.com/kubernetes/minikube/issues/14456
+while ! minikube start; do sleep 30; done
 
 # Load the BMO e2e image into it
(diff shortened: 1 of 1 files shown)


hack/ci-e2e.sh
```diff
@@ -70,7 +70,15 @@ virsh -c qemu:///system attach-interface --domain minikube --mac="52:54:00:6c:3c
 
 # Restart minikube to apply the changes
 minikube stop
-minikube start
+## Following loop is to avoid minikube restart issue
+## https://github.com/kubernetes/minikube/issues/14456
+while /bin/true; do
+    minikube_error=0
+    minikube start || minikube_error=1
```
tuminoid, line 77: This is simpler with just 
```suggestion
    if minikube start; then break; fi
```
Also adding a sleep 5 or sleep 30 (whatever is appropriate) is good, instead of just spamming. Some exit condition in case minikube doesn't recover would also be nice, until we deem quitting after timeout (45mins or so) is OK.

tuminoid: Or even

```suggestion
    while ! minikube start; do sleep 30; done
```

kashifest: agree, this is simpler and less spammed, done

PR #1991, opened 2024-10-11
🌱 make CAPI a pattern in dependabot config

tuminoid:
Make CAPI a pattern in dependabot config, so cluster-api and cluster-api/test packages get updated with single dependabot PR.

Same has been implemented in CAPM3 and IPAM as well.


kashifest (commented):
/approve

PR #2027, opened 2024-10-31
🌱 Move TLS 1.3 from hardcoded default value to a config flag

smoshiur1237:
Move TLS 1.3 from hardcoded default value to a config flag.

This aligns BMO implementation to CAPM3/IPAM implementation while keeping TLS 1.3 as default.

### DIFF
--- a/config/base/manager.yaml
+++ b/config/base/manager.yaml
@@ -23,6 +23,7 @@ spec:
             - /baremetal-operator
           args:
             - --enable-leader-election
+            - --tls-min-version=TLS13
           image: quay.io/metal3-io/baremetal-operator
           imagePullPolicy: Always
           env:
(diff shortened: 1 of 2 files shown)


tuminoid (commented):
/retitle 🌱 Move TLS 1.3 from hardcoded default value to a config flag

This aligns BMO implementation to CAPM3/IPAM implementation while keeping TLS 1.3 as default.
 
/lgtm

mquhuy:
/retest

lgtm, but I think the PR content should be changed. Here the default value is changed to TLS 1.2, and TLS 1.3 is set as a parameter inside the container image.

tuminoid:
> /retest
> 
> lgtm, but I think the PR content should be changed. Here the default value is changed to TLS 1.2, and TLS 1.3 is set as a parameter inside the container image.

I've fixed summary and retitled it to be more exact.

PR #2028, opened 2024-11-01
:seedling:  Fetch Ironic node list after each e2e test

### LINKED ISSUE #1383: E2e resource collection

mquhuy:
Fixes #1383

There will be a new file called "ironic-nodes.json" in the testcase log directory.

### DIFF
--- a/test/e2e/common.go
+++ b/test/e2e/common.go
@@ -3,7 +3,13 @@ package e2e
 import (
 	"bytes"
 	"context"
+	"crypto/tls"
+	"encoding/base64"
+	"encoding/json"
 	"fmt"
+	"io"
+	"net"
+	"net/http"
 	"os"
 	"path"
 	"path/filepath"
@@ -499,7 +505,7 @@ func dumpCRDS(ctx context.Context, cli client.Client, artifactFolder string) {
 }
 
 // DumpResources dumps resources related to BMO e2e tests as YAML.
-func DumpResources(ctx context.Context, clusterProxy framework.ClusterProxy, namespace string, artifactFolder string) {
+func DumpResources(ctx context.Context, e2eConfig *Config, clusterProxy framework.ClusterProxy, namespace string, artifactFolder string) {
 	// CAPI resources.
 	framework.DumpAllResources(ctx, framework.DumpAllResourcesInput{
 		Lister:    clusterProxy.GetClient(),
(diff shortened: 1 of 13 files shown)


test/e2e/common.go
```diff
@@ -507,4 +517,77 @@ func DumpResources(ctx context.Context, clusterProxy framework.ClusterProxy, nam
...
+
+func DumpIronicNodes(ctx context.Context, artifactFolder string) {
+	ironicProvisioningIP, exists := os.LookupEnv("IRONIC_PROVISIONING_IP")
+	if !exists {
+		Logf("Running fixture tests, skip fetching ironic nodes.")
+		return
+	}
+	ironicURL := fmt.Sprintf("https://%s/v1/nodes", net.JoinHostPort(ironicProvisioningIP, "6385"))
+	username := os.Getenv("IRONIC_USERNAME")
+	password := os.Getenv("IRONIC_PASSWORD")
```
lentzi90, line 531: It is not so nice to read environment variables so "late"/"deep" in the code. Usually things like this would be handled immediately at the start of execution of a program, to initialize data structures and configuration (just like we do in [e2e_suite_test.go](https://github.com/metal3-io/baremetal-operator/blob/main/test/e2e/e2e_suite_test.go). Then this data would be explicitly passed through as arguments so it is clear what input each function has to work with.
Now that this function reads it from environment variables, it looks a bit like magic. It doesn't take any credentials or even an endpoint as input but still knows how to get the ironic nodes.

I would want it to work more like [PerformSSHBootCheck](https://github.com/metal3-io/baremetal-operator/blob/1c72ffe4ae8b05067a85b40ed925478a657b31a1/test/e2e/common.go#L276). It takes all information it needs as arguments and it uses the e2e config for variables. That way the user gets unified way of configuring things. They can decide if they want environment variables or config files or both.

Could you add these variables to the e2e config and make this function access them that way?

test/e2e/common.go
```diff
@@ -507,4 +517,77 @@ func DumpResources(ctx context.Context, clusterProxy framework.ClusterProxy, nam
...
+	client := &http.Client{
+		Transport: &http.Transport{TLSClientConfig: tlsConfig},
+	}
+
+	// Create the request
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, ironicURL, http.NoBody)
+	if err != nil {
+		Errorf("Failed to create request: %v", err)
+		return
+	}
```
lentzi90, line 546:
```suggestion
	Expect(err).NotTo(HaveOccurred(), "Failed to create request")
```

test/e2e/common.go
```diff
@@ -507,4 +517,77 @@ func DumpResources(ctx context.Context, clusterProxy framework.ClusterProxy, nam
...
+	// Set basic auth header
+	auth := base64.StdEncoding.EncodeToString([]byte(username + ":" + password))
+	req.Header.Add("Authorization", "Basic "+auth)
+
+	// Make the request
+	resp, err := client.Do(req)
+	if err != nil {
+		Errorf("Failed to send request: %v", err)
+		return
+	}
```
lentzi90, line 557:
```suggestion
	Expect(err).NotTo(HaveOccurred(), "Failed to send request")
```

test/e2e/common.go
```diff
@@ -507,4 +517,77 @@ func DumpResources(ctx context.Context, clusterProxy framework.ClusterProxy, nam
...
+
+	if resp.StatusCode != http.StatusOK {
+		Errorf("Request failed with status: %d", resp.StatusCode)
+		return
+	}
```
lentzi90, line 562:
```suggestion
	Expect(resp.StatusCode).To(Equal(http.StatusOK), "Unexpected http status code")
```

test/e2e/common.go
```diff
@@ -507,4 +517,77 @@ func DumpResources(ctx context.Context, clusterProxy framework.ClusterProxy, nam
...
+	defer resp.Body.Close()
+	// Read and output the response
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		Errorf("Failed to read response body: %v", err)
+		return
+	}
```
lentzi90, line 569:
```suggestion
	Expect(err).NotTo(HaveOccurred(), "Failed to read response body")
```

test/e2e/common.go
```diff
@@ -507,4 +517,77 @@ func DumpResources(ctx context.Context, clusterProxy framework.ClusterProxy, nam
...
+
+	var logOutput bytes.Buffer
+
+	// Format the JSON with indentation
+	err = json.Indent(&logOutput, body, "", "    ")
+	if err != nil {
+		Errorf("Error formatting JSON: %v", err)
+		return
+	}
```
lentzi90, line 578:
```suggestion
	Expect(err).NotTo(HaveOccurred(), "Error formatting JSON")
```

test/e2e/common.go
```diff
@@ -507,4 +517,77 @@ func DumpResources(ctx context.Context, clusterProxy framework.ClusterProxy, nam
...
+
+	file, err := os.Create(path.Join(artifactFolder, "ironic-nodes.json"))
+	if err != nil {
+		Errorf("Error creating file: %v", err)
+		return
+	}
```
lentzi90, line 584:
```suggestion
	Expect(err).NotTo(HaveOccurred(), "Error creating file")
```

test/e2e/common.go
```diff
@@ -507,4 +517,77 @@ func DumpResources(ctx context.Context, clusterProxy framework.ClusterProxy, nam
...
+	defer file.Close()
+
+	// Write indented JSON to file
+	_, err = file.Write(logOutput.Bytes())
+	if err != nil {
+		Errorf("Error writing JSON to file: %v", err)
+		return
+	}
```
lentzi90, line 592:
```suggestion
	Expect(err).NotTo(HaveOccurred(), "Error writing JSON to file")
```

test/e2e/common.go
```diff
@@ -447,6 +453,10 @@ func Logf(format string, a ...interface{}) {
 	fmt.Fprintf(GinkgoWriter, "INFO: "+format+"\n", a...)
 }
 
+func Errorf(format string, a ...interface{}) {
+	fmt.Fprintf(GinkgoWriter, "ERROR: "+format+"\n", a...)
+}
+
```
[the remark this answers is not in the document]
mquhuy: The reason I don't want to use Expect() was because I don't think it's worth it to stop the whole process if this fetch function fails. `Errorf()` acts as a way of getting the error log without stopping the test.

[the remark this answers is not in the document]
lentzi90: Then it is not an error. Just use `Logf` :slightly_smiling_face:

test/e2e/common.go
mquhuy: Thanks. I've now implemented it that way. Could you check and see if the code looks like what you have in mind?

test/e2e/e2e_config.go
```diff
@@ -127,7 +127,7 @@ func (c *Config) Validate() error {
 
 // GetIntervals returns the intervals to be applied to a Eventually operation.
 // It searches for [spec]/[key] intervals first, and if it is not found, it searches
-// for default/[key]. If also the default/[key] intervals are not found,
+// for default/[kgy]. If also the default/[key] intervals are not found,
```
lentzi90, line 130: What? :sweat_smile: 
```suggestion
// for default/[key]. If also the default/[key] intervals are not found,
```

lentzi90 (commented):
Great!
/approve

test/e2e/e2e_config.go
mquhuy: :sweat_smile: Huh. I must have messed up some vim key at some point :sweat_smile: Chang incoming.

Rozzii (commented):
/lgtm

PR #2041, opened 2024-11-05
✨ Add reference to HostUpdatePolicy in Servicing with HFS Support

iurygregory:

This PR enables BMO to run Ironic servicing operations (such as applying firmware settings changes - or in the future firmware updates to already provisioned nodes). Servicing is an opt-in feature and is controlled by creation of a HostUpdatePolicy for a node with attributes indicating the desire to make changes to firmware configuration onReboot.

This is a partial implementation of metal3-io/metal3-docs@e260d42304a5822816ebc1752ab6c0818b9ee589  (please note only firmware settings changes are currently supported, firmware update support will be added next).

### DIFF
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -1361,7 +1361,7 @@ func (r *BareMetalHostReconciler) actionDeprovisioning(prov provisioner.Provisio
 	return actionComplete{}
 }
 
-func (r *BareMetalHostReconciler) doServiceIfNeeded(prov provisioner.Provisioner, info *reconcileInfo, hup *metal3api.HostUpdatePolicy) (result actionResult, isServicing bool) {
+func (r *BareMetalHostReconciler) doServiceIfNeeded(prov provisioner.Provisioner, info *reconcileInfo, hup *metal3api.HostUpdatePolicy) (result actionResult) {
 	servicingData := provisioner.ServicingData{}
 
 	// (NOTE)janders: since Servicing is an opt-in feature that requires HostUpdatePolicy to be created and set to onReboot
@@ -1386,7 +1386,7 @@ func (r *BareMetalHostReconciler) doServiceIfNeeded(prov provisioner.Provisioner
 		var err error
 		hfsDirty, hfs, err = r.getHostFirmwareSettings(info)
 		if err != nil {
-			return actionError{fmt.Errorf("could not determine updated settings: %w", err)}, false
+			return actionError{fmt.Errorf("could not determine updated settings: %w", err)}
 		}
 		if hfsDirty {
 			servicingData.ActualFirmwareSettings = hfs.Status.Settings
@@ -1399,41 +1399,47 @@ func (r *BareMetalHostReconciler) doServiceIfNeeded(prov provisioner.Provisioner
 	// Even if settings are clean, we need to check the result of the current servicing.
 	if !dirty && info.host.Status.OperationalStatus != metal3api.OperationalStatusServicing && info.host.Status.ErrorType != metal3api.ServicingError {
 		// If nothing is going on, return control to the power management.
-		return nil, false
+		return nil
 	}
 
+	// FIXME(janders/dtantsur): this implementation may lead to a scenario where if we never actually
+	// succeed before leaving this state (e.g. by deprovisioning) we lose the signal that the
+	// update didn't actually happen. This is deemed an acceptable risk for the moment since it is only
+	// going to impact a small subset of Firmware Settings implementations.
+	currentError := info.host.Status.ErrorType
 	provResult, started, err := prov.Service(servicingData, dirty,
-		info.host.Status.ErrorType == metal3api.ServicingError)
+		currentError == metal3api.ServicingError)
 	if err != nil {
-		return actionError{fmt.Errorf("error servicing host: %w", err)}, false
+		return actionError{fmt.Errorf("error servicing host: %w", err)}
 	}
 	if provResult.ErrorMessage != "" {
 		result = recordActionFailure(info, metal3api.ServicingError, provResult.ErrorMessage)
-		return result, true
+		return result
 	}
-	if started && clearErrorWithStatus(info.host, metal3api.OperationalStatusServicing) {
-		if fwDirty {
-			info.host.Status.Provisioning.Firmware = info.host.Spec.Firmware.DeepCopy()
-		}
+
+	if clearErrorWithStatus(info.host, metal3api.OperationalStatusServicing) {
 		dirty = true
+	}
+
+	if started && fwDirty {
+		info.host.Status.Provisioning.Firmware = info.host.Spec.Firmware.DeepCopy()
 	}
 
 	if provResult.Dirty {
 		result := actionContinue{provResult.RequeueAfter}
 		if dirty {
-			return actionUpdate{result}, true
+			return actionUpdate{result}
 		}
-		return result, true
+		return result
 	}
 
 	// Servicing is finished at this point, clean up operational status
 	if clearErrorWithStatus(info.host, metal3api.OperationalStatusOK) {
-		// We need to give the HostFirmwareSettings controller some time to
-		// catch up with the changes, otherwise we risk starting the same
-		// operation again.
-		return actionUpdate{actionContinue{delay: subResourceNotReadyRetryDelay}}, true
+		// FIXME(janders/dtantsur): this can be racy. We should consider
+		// using a generation number to decide if we start servicing or not.
+		return actionUpdate{actionContinue{delay: subResourceNotReadyRetryDelay}}
 	}
-	return nil, false
+	return nil
 }
 
 // Check the current power status against the desired power status.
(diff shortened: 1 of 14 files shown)


controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1355,6 +1357,92 @@ func (r *BareMetalHostReconciler) actionDeprovisioning(prov provisioner.Provisio
...
+		if err != nil {
+			return actionError{fmt.Errorf("could not determine updated settings: %w", err)}
+		}
+		if hfsDirty {
+			servicingData.ActualFirmwareSettings = hfs.Status.Settings
+			servicingData.TargetFirmwareSettings = hfs.Spec.Settings
+		}
+	}
+
+	dirty := fwDirty || hfsDirty
```
dtantsur, line 1393: Per comment from @zaneb https://github.com/metal3-io/baremetal-operator/pull/1969#discussion_r1815900322, I think we should not assign `dirty` here. Maybe rename this new variable to `hasChanges` or something like that and let `dirty` to be assigned on lines 1421 and 1426.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1355,6 +1357,92 @@ func (r *BareMetalHostReconciler) actionDeprovisioning(prov provisioner.Provisio
...
+	if clearErrorWithStatus(info.host, metal3api.OperationalStatusServicing) {
+		dirty = true
+	}
+
+	if started && fwDirty {
+		info.host.Status.Provisioning.Firmware = info.host.Spec.Firmware.DeepCopy()
+		dirty = true
+	}
+
+	if provResult.Dirty {
```
dtantsur, line 1429: Similarly, I suspect this logic is not 100% correct (sorry for that!). I think we need something along the lines of

```
if dirty {
    return actionUpdate{result}
} else if provResult.Dirty {
}
```

This way we won't forget to return `actionUpdate` if the provisioner returns a clean result but there are changes to the BMH.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -895,10 +901,6 @@ func (r *BareMetalHostReconciler) registerHost(prov provisioner.Provisioner, inf
 				info.log.Info("failed creating hostfirmwarecomponents")
 				return actionError{errors.Wrap(err, "failed creating hostFirmwareComponents")}
 			}
-			if err = r.hostUpdatePolicySetOwnerReference(info); err != nil {
-				info.log.Info("failed setting owner reference on hostupdatepolicy")
-				return actionError{errors.Wrap(err, "failed setting owner reference on hostUpdatePolicy")}
-			}
```
zaneb, line 901: I think we want to continue to do this as early as possible (i.e. during registration).

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1385,10 +1480,19 @@ func (r *BareMetalHostReconciler) manageHostPower(prov provisioner.Provisioner,
...
+	hup, err := r.acquireHostUpdatePolicy(info)
+	if err != nil {
+		info.log.Info("failed setting owner reference on hostupdatepolicy")
+		return actionError{errors.Wrap(err, "failed setting owner reference on hostUpdatePolicy")}
+	}
 
-	provState := info.host.Status.Provisioning.State
-	// Normal reboots only work in provisioned states, changing online is also possible for available hosts.
-	isProvisioned := provState == metal3api.StateProvisioned || provState == metal3api.StateExternallyProvisioned
+	servicingAllowed := isProvisioned && !info.host.Status.PoweredOn && desiredPowerOnState
```
zaneb, line 1489: nit: there is already an `if !info.host.Status.PoweredOn` block above. We could move all of this new code inside that.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1355,6 +1357,91 @@ func (r *BareMetalHostReconciler) actionDeprovisioning(prov provisioner.Provisio
...
+	}
+	if provResult.ErrorMessage != "" {
+		result = recordActionFailure(info, metal3api.ServicingError, provResult.ErrorMessage)
+		return result
+	}
+
+	dirty := false
+	if clearErrorWithStatus(info.host, metal3api.OperationalStatusServicing) {
+		dirty = true
+	}
```
zaneb, line 1423: nit:

```
dirty := clearErrorWithStatus(info.host, metal3api.OperationalStatusServicing)
```

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1355,6 +1357,91 @@ func (r *BareMetalHostReconciler) actionDeprovisioning(prov provisioner.Provisio
...
+	if started && fwDirty {
+		info.host.Status.Provisioning.Firmware = info.host.Spec.Firmware.DeepCopy()
+		dirty = true
+	}
+
+	if dirty {
+		return actionUpdate{}
+	} else if provResult.Dirty {
+		return actionContinue{provResult.RequeueAfter}
+	}
```
zaneb, line 1434: nit:

```
result := actionContinue{delay: provResult.RequeueAfter}
if dirty {
    return actionUpdate{result}
} else if provResult.Dirty {
    return result
}
```

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1355,6 +1357,91 @@ func (r *BareMetalHostReconciler) actionDeprovisioning(prov provisioner.Provisio
...
+		info.host.Status.OperationalStatus = metal3api.OperationalStatusServicing
+		return actionUpdate{}
+	}
+
+	provResult, started, err := prov.Service(servicingData, hasChanges,
+		info.host.Status.ErrorType == metal3api.ServicingError)
+	if err != nil {
+		return actionError{fmt.Errorf("error servicing host: %w", err)}
+	}
+	if provResult.ErrorMessage != "" {
```
zaneb, line 1415: We need to clear `info.host.Status.Provisioning.Firmware` at this point to signal that it was not, in fact, updated (since we copy the target status into it as soon as we start trying).

zaneb (commented):
Looking pretty solid.

controllers/metal3.io/baremetalhost_controller.go
iurygregory: oh ok! I was wondering about this and asked @dtantsur , we though we wouldn't need the result variable. I will update

controllers/metal3.io/baremetalhost_controller.go
iurygregory: You mean adding a call for `acquireHostUpdatePolicy` here correct? (since we changed the name `hostUpdatePolicySetOwnerReference` and we only call in `managePower`. 
The HostUpdatePolicy is not automatically created, so I'm wondering if we really need during registration 🤔 
@dtantsur do you have thoughts on it?

controllers/metal3.io/baremetalhost_controller.go
iurygregory: Something like https://paste.opendev.org/show/bAWRmUqzLC5hrAbS8Uec/ ?

controllers/metal3.io/baremetalhost_controller.go
iurygregory: Changed to 
```
	resultAction := actionContinue{delay: provResult.RequeueAfter}
	if dirty {
		return actionUpdate{resultAction}
	} else if provResult.Dirty {
		return resultAction
	}
```

For some reason the return `actionUpdate` was complaining that the type of result was `actionResult` not `actionContinue`

controllers/metal3.io/baremetalhost_controller.go
iurygregory: added info.host.Status.Provisioning.Firmware = nil

controllers/metal3.io/baremetalhost_controller.go
iurygregory: added
```
			if _, err = r.acquireHostUpdatePolicy(info); err != nil {
				info.log.Info("failed setting owner reference on hostupdatepolicy")
				return actionError{errors.Wrap(err, "failed setting owner reference on hostUpdatePolicy")}
			}
```

controllers/metal3.io/baremetalhost_controller.go
dtantsur: Yeah, because the derived type is specific, not interface, so the 2nd assignment is not valid. Honestly, I don't agree with Zane on this one, but I also won't fight about it.

controllers/metal3.io/baremetalhost_controller.go
dtantsur: I think it's a fair call: we need to ensure that even user-created objects have the right owner reference, and we should not wait with that until the power action.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1385,10 +1480,19 @@ func (r *BareMetalHostReconciler) manageHostPower(prov provisioner.Provisioner,
 			return actionContinue{}
 		}
 	}
+	hup, err := r.acquireHostUpdatePolicy(info)
+	if err != nil {
+		info.log.Info("failed setting owner reference on hostupdatepolicy")
+		return actionError{errors.Wrap(err, "failed setting owner reference on hostUpdatePolicy")}
+	}
```
dtantsur, line 1487: If we do this early, we don't need to do it here.

controllers/metal3.io/baremetalhost_controller.go
dtantsur: The annotation has to be deleted before servicing, and you no longer need acquireHostUpdatePolicy if you do it on registration.

dtantsur: Okay, I guess it won't hurt to call it here as well.

iurygregory:
I've tested the code trying to update a firmware setting, it works updating the setting but the OperationalStatus for the BMH stays `servicing` and doesn't move to `OK`

@zaneb @dtantsur any thoughts on this? Seems to be related to the comment's I've addressed in the last commit, but no idea at the moment.

iurygregory:
Moved back to the old logic, since  for the `servicingAllowed`, since we still need to account

controllers/metal3.io/baremetalhost_controller.go
zaneb: It wasn't _wrong_ before, but it would have requeued immediately. Doing it this way means it doesn't requeue until after the delay specified by the provisioner.

PR #2043, opened 2024-11-06
🌱 Introduce inspection delay for the fixture provisioner

lentzi90:

The fixture provisioner did have a delay previously, but it was removed as part of some work on reducing access to sensitive information. Now that we are running tests with the fixture provisioner, this has become an issue, since inspection happens so fast that it is hard to detect.

This commit introduces artificial delay without accessing any extra information.

Fixes #


mquhuy:
/lgtm
Thank you for making this. This issue's been a major part of our flaky fixture CI.

dtantsur:
/approve

I'm not really fond of relying on timing here, but at least it improves the situation.

PR #2044, opened 2024-11-06
✨  Allow Firmware Updates via Servicing with HostUpdatePolicy

iurygregory:
**What this PR does / why we need it**:   This a follow-up to add support for Firmware Updates via HostUpdatePolicy, finalizing the implementation of metal3-io/metal3-docs@e260d42


rhjanders:
Thank you Iury. It passes CI and addresses in-code TODOs so
/lgtm
(let me try if I have reviewer rights upstream).
If this works it would be get another pair of eyes on it, too, since I am still very new to this codebase.

rhjanders:
> @rhjanders: adding LGTM is restricted to approvers and reviewers in OWNERS files.

well, clearly I do not :)

pkg/provisioner/provisioner.go
```diff
@@ -108,10 +108,10 @@ type PrepareData struct {
...
 
 type ServicingData struct {
-	FirmwareConfig         *metal3api.FirmwareConfig
-	TargetFirmwareSettings metal3api.DesiredSettingsMap
-	ActualFirmwareSettings metal3api.SettingsMap
-	// TargetFirmwareComponents []metal3api.FirmwareUpdate
+	FirmwareConfig           *metal3api.FirmwareConfig
+	TargetFirmwareSettings   metal3api.DesiredSettingsMap
+	ActualFirmwareSettings   metal3api.SettingsMap
+	TargetFirmwareComponents []metal3api.FirmwareUpdate
```
dtantsur, line 114: Where is this field set? I think you need to update the controller to populate it somewhere.

pkg/provisioner/ironic/servicing.go
```diff
@@ -35,9 +35,22 @@ func (p *ironicProvisioner) buildServiceSteps(bmcAccess bmc.AccessDetails, data
 		)
 	}
 
-	// TODO: Add service steps for firmware updates
+	newUpdates := p.getFirmwareComponentsUpdates(data.TargetFirmwareComponents)
```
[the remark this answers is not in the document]
iurygregory: So, if we are doing servicing the API already supports firmware updates (1.87),  so I didn't think it was necessary.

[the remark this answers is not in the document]
dtantsur: Right, yeah, you cannot have servicing and not have firmware updates.

pkg/provisioner/provisioner.go
iurygregory: Now I'm puzzled, I was sure I added 
```
	hfcDirty, hfc, err := r.getHostFirmwareComponents(info)

	if err != nil {
		// wait until hostFirmwareComponents are ready
	}
	if hfcDirty {
		prepareData.TargetFirmwareComponents = hfc.Spec.Updates
	}
```

Let me see why is not showing here 😅

iurygregory: Done, and tested  with virtual BMH  =)

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1396,9 +1398,17 @@ func (r *BareMetalHostReconciler) doServiceIfNeeded(prov provisioner.Provisioner
 			servicingData.ActualFirmwareSettings = hfs.Status.Settings
 			servicingData.TargetFirmwareSettings = hfs.Spec.Settings
 		}
+
```
dtantsur, line 1401: You need a separate block that checks hup.Spec.FirmwareUpdates, similarly to how hup.Spec.FirmwareSettings defines liveFirmwareSettingsAllowed.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1422,6 +1432,9 @@ func (r *BareMetalHostReconciler) doServiceIfNeeded(prov provisioner.Provisioner
 	}
 	if provResult.ErrorMessage != "" {
 		info.host.Status.Provisioning.Firmware = nil
+		if hfcDirty {
+			hfc.Status.Updates = nil
```
dtantsur, line 1436: I don't think this change is going to be saved. On line 1434 we're relying on the higher-level controller code to save the BMH but nothing is gong to save the HFC after you return.

Changing the HFC here runs a risk that we override the error message with a transient failure from the unability to update the HFC. I wonder if we can do it earlier? Is there any harm in wiping the current Updates after line 1426 (provided that hfcDirty is true)?

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1433,6 +1446,21 @@ func (r *BareMetalHostReconciler) doServiceIfNeeded(prov provisioner.Provisioner
 		dirty = true
 	}
 
+	if hfcDirty && started {
+		hfcStillDirty, err := r.saveHostFirmwareComponents(prov, info, hfc)
```
dtantsur, line 1450: nit: you can reuse hfcDirty since you no longer need its original value

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1433,6 +1446,21 @@ func (r *BareMetalHostReconciler) doServiceIfNeeded(prov provisioner.Provisioner
...
+		if err != nil {
+			return actionError{errors.Wrap(err, "could not save the host firmware components")}
+		}
+
+		if hfcStillDirty {
+			info.log.Info("going to update the host firmware components")
```
dtantsur, line 1456: This is not super helpful, let's log the new components or drop the message.

dtantsur (changes requested):

controllers/metal3.io/baremetalhost_controller.go
iurygregory: oh you are right, we have `firmwareSettings` and `firmwareUpdates` in the Spec for the HUP!
I will update

controllers/metal3.io/baremetalhost_controller.go
iurygregory: Thanks!

controllers/metal3.io/baremetalhost_controller.go
iurygregory: So maybe we have the same problem in https://github.com/metal3-io/baremetal-operator/blob/main/controllers/metal3.io/baremetalhost_controller.go#L1183 ? 

`Changing the HFC here runs a risk that we override the error message with a transient failure from the unability to update the HFC. I wonder if we can do it earlier? Is there any harm in wiping the current Updates after line 1426 (provided that hfcDirty is true)?`

I would say tehre is a harm if we try to change the `Spec.Updates` from the `hfc` because we call the `r.saveHostFirmwareComponents(prov, info, hfc)` on L 1450

controllers/metal3.io/baremetalhost_controller.go
iurygregory: removing since previously we would call the `saveHostFirmwareComponents` and in case of changes to be save we will log the components.

controllers/metal3.io/baremetalhost_controller.go
zaneb: Yes, that line is for sure doing nothing and so is this one. You'd have to call `saveHostFirmwareComponents()` after for it to have any effect.

> Changing the HFC here runs a risk that we override the error message with a transient failure

The error message will still be there when you come back.

iurygregory: > Yes, that line is for sure doing nothing and so is this one. 

`that` line are you referring to what exactly?  `this one` I'm assuming you are talking about the `hfc.Status.Updates = nil`  L1440.

> You'd have to call `saveHostFirmwareComponents()` after for it to have any effect.

Ok, so we will need it twice the `saveHostFirmwareComponents()` because we have the return in the if block `	if provResult.ErrorMessage != "" {`

iurygregory: Thanks, I've updated! to include the `r.Status().Update(info.ctx, hfc)`

lentzi90 (commented):
/lgtm

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1398,7 +1401,18 @@ func (r *BareMetalHostReconciler) doServiceIfNeeded(prov provisioner.Provisioner
 		}
 	}
 
-	hasChanges := fwDirty || hfsDirty
+	if liveFirmwareUpdatesAllowed {
+		var errhfc error
```
dtantsur, line 1405: nit: you could call this `err`, there is not much harm in that

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1422,6 +1436,11 @@ func (r *BareMetalHostReconciler) doServiceIfNeeded(prov provisioner.Provisioner
 	}
 	if provResult.ErrorMessage != "" {
 		info.host.Status.Provisioning.Firmware = nil
+		if hfcDirty {
+			hfc.Status.Updates = nil
+			// Probably safe to ignore any errors, since we will return result based on an error
```
dtantsur, line 1441: I don't quite get this comment. It's not safe to ignore errors: this way we will not retry Status.Update next time, and `Status.Updates` will keep whatever value is has. I think what @zaneb meant was that if we return actionError here, on the next iteration the provisioner.Service call will still return an ErrorMessage, so we'll enter this code again.

On top of that, we should not call Status.Update if the value is already `nil`.

dtantsur (changes requested):

controllers/metal3.io/baremetalhost_controller.go
iurygregory: Done

controllers/metal3.io/baremetalhost_controller.go
iurygregory: Done, thanks!

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1433,6 +1453,20 @@ func (r *BareMetalHostReconciler) doServiceIfNeeded(prov provisioner.Provisioner
...
 	}
 
+	if hfcDirty && started {
+		hfcDirty, err = r.saveHostFirmwareComponents(prov, info, hfc)
+		if err != nil {
+			return actionError{errors.Wrap(err, "could not save the host firmware components")}
+		}
+
+		if hfcDirty {
+			if err := r.Status().Update(info.ctx, hfc); err != nil {
```
dtantsur, line 1463: Please remind me, why are we doing this? It looks like the status changes are actually in the HFC object, not in the BMH. And the BMH will be saved by the higher level controller code anyway.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1433,6 +1453,20 @@ func (r *BareMetalHostReconciler) doServiceIfNeeded(prov provisioner.Provisioner
...
+				return actionError{errors.Wrap(err, "failed to update hostfirmwarecomponents status")}
+			}
+		}
+		dirty = true
```
dtantsur, line 1467: The BMH is no longer dirty: you've just saved its status (see above). Maybe you actually only need this `dirty = true` but do not need the Update above?

controllers/metal3.io/baremetalhost_controller.go
iurygregory: When calling the saveHostFirmwareComponents we change the information there, so the call here is to indeed update the CR

controllers/metal3.io/baremetalhost_controller.go
iurygregory: agree that the BMH is not dirty, but unsure if I don't need the Update above

controllers/metal3.io/baremetalhost_controller.go
dtantsur: Right, I somehow thought you were updating the BMH

controllers/metal3.io/baremetalhost_controller.go
iurygregory: Done

iurygregory:
` e2e-test (ipmi) ` Error: No files were found with the provided path: test/e2e/_artifacts. No artifacts will be uploaded. 🤔

dtantsur (approved):

PR #2045, opened 2024-11-07
🐛 Fix the logic that determines previous git tag

huutomerkki:

This PR changes the way previous tag is determined to only include those tags that have been merged into the current tree.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -328,7 +328,7 @@ mod: ## Clean up go module settings
 ## --------------------------------------
 RELEASE_TAG ?= $(shell git describe --abbrev=0 2>/dev/null)
 RELEASE_NOTES_DIR := releasenotes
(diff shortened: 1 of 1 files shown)


tuminoid (changes requested):
Technically the right change, but practically not working. For example, we don't have v0.6.0 tag in main at all, nor we have v0.8.0, so if we check what the command now picks for v0.9.0, it'll compare it to v0.5.0.

Issue here is that when we make minor RC releases, they're already made from release branches. Thus the tags do not appear in main, and are not considered when making next minor.

tuminoid:
Also, https://github.com/metal3-io/ip-address-manager/pull/734 in IPAM will change the release creation logic, and will implement the same in Go code, so it makes no sense to implement this here.

PR #2048, opened 2024-11-10
:bug: Make custom deploy equivalent to image deploy

zaneb:
If we have to recreate the provisioner's state during deprovisioning, we must
first call Adopt() to ensure that something will actually happen to deprovision it.
This is the case regardless of provisioning method, but previously only
happened when an image was provisioned. In practice, the ironic provisioner
was attempting to adopt in cases where it was not originally intended to, and
subsequent changes mean it is never necessary.

If an externally provisioned host has a provisioning image set and then
the externallyProvisioned flag is removed, we should not begin an inspection
(which would involve rebooting the host), but instead move directly to
the provisioned state. This is also the case regardless of
provisioning method, but previously only happened when an image was
provisioned.

### DIFF
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -1309,7 +1309,7 @@ func clearHostProvisioningSettings(host *metal3api.BareMetalHost) {
 }
 
 func (r *BareMetalHostReconciler) actionDeprovisioning(prov provisioner.Provisioner, info *reconcileInfo) actionResult {
-	if info.host.Status.Provisioning.Image.URL != "" {
+	if info.host.Status.Provisioning.Image.URL != "" || info.host.Status.Provisioning.CustomDeploy != nil {
 		// Adopt the host in case it has been re-registered during the
 		// deprovisioning process before it completed
 		provResult, err := prov.Adopt(
--- a/apis/metal3.io/v1alpha1/baremetalhost_types.go
+++ b/apis/metal3.io/v1alpha1/baremetalhost_types.go
@@ -1036,6 +1036,10 @@ func (host *BareMetalHost) WasProvisioned() bool {
 		// We have an image provisioned.
 		return true
 	}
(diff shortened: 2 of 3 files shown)


controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1309,7 +1309,7 @@ func clearHostProvisioningSettings(host *metal3api.BareMetalHost) {
 }
 
 func (r *BareMetalHostReconciler) actionDeprovisioning(prov provisioner.Provisioner, info *reconcileInfo) actionResult {
-	if info.host.Status.Provisioning.Image.URL != "" {
+	if info.host.Status.Provisioning.Image.URL != "" || info.host.Status.Provisioning.CustomDeploy != nil {
```
zaneb, line 1312: @dtantsur I think we also have to make a change to https://github.com/metal3-io/baremetal-operator/blob/af93711c586ed708e26217cf7a74caa1826a456e/pkg/provisioner/ironic/ironic.go#L1106-L1109 for this to have any effect, but in the case of CustomDeploy the `instance_info` is likely to be empty either way. I presume the place where it is getting cleared is https://opendev.org/openstack/ironic/src/branch/stable/2024.2/ironic/conductor/manager.py#L1149. There doesn't seem to be any corresponding code in ironic to clear the `deploy_interface`, so I don't know how we can distinguish being in the `manageable` state after a previous failure to clean from being in the `manageable` state after recreating the node in a fresh database. Do you have any ideas?

zaneb (commented):
/hold

controllers/metal3.io/baremetalhost_controller.go
zaneb: On closer inspection, that code wasn't doing anything anyway, so I removed it.

tuminoid:
It seems Prow is having issues. Failures are non PR related.

dtantsur:
/approve

Looks valid to me.

PR #2060, opened 2024-11-19
🐛 Update HFC Status when handling error in actionPreparing

iurygregory:
When a provisioning error happens and firmware updates are involved, we should only set the Status Updates of the HFC to nil if they are not nil.


tuminoid:
> /ok-to-test

Btw, ok-to-test is not needed for Metal3 members.

PR #2063, opened 2024-11-20
:seedling: use uuidgen to generate random username/password

tuminoid:
Using /dev/urandom has entropy issue. It can run out especially in CI, leaving jobs stuck. We have used uuidgen in BMO e2e for long time now and it works great. These usernames/passwords just need to be anything random to make sure nothing is hardcoded, so uuidgen works great for that anyways.


lentzi90 (commented):
/approve

PR #2069, opened 2024-11-20
:seedling: Bump Go to 1.23.3 and golangci-lint to 1.60.3

tuminoid:
Bump Go to 1.23.3 and golangci-lint to 1.60.3 which supports Go 1.23.

Ignore couple new warnings that are generated, as there is no immediate good options.

Cleanup golangci-lint config from deprecated configs.

Include timeouts where needed, golangci-lint 1.60 is defaulting to real short runs and failing otherwise.

Fixes:  #2009

PR #2070, opened 2024-11-20
:seedling: Bump Go to 1.23.3 and golangci-lint to 1.60.3

tuminoid:
Bump Go to 1.23.3 and golangci-lint to 1.60.3 which supports Go 1.23.

Ignore couple new warnings that are generated, as there is no immediate good options.

Cleanup golangci-lint config from deprecated configs.

Include timeouts where needed, golangci-lint 1.60 is defaulting to real short runs and failing otherwise.

Manual cherry-pick of #2069 due conflict in golangci-lint config.

/hold
until #2069 is merged, so they can be merged as identical.

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -3,7 +3,7 @@ name: golangci-lint
 on:
   workflow_call:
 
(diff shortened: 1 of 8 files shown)


lentzi90 (commented):
/lgtm

PR #2087, opened 2024-11-28
🌱 Remove deprecated ControllerManagerConfig

hex2dec:

The ComponentConfig has been deprecated in the controller-runtime since v0.15.0. Remove this component config because it is no longer used by the baremetal operator.


lentzi90 (commented):
/approve

PR #2088, opened 2024-11-28
🐛 Fix listing ironic node's ports with empty mac address

hex2dec:

Fix listing ironic node's ports with empty mac address. Call the listing ports API with empty mac address is same as listing all ironic node ports.

Fixes #


lentzi90 (commented):
I don't know this area well enough to review.
/cc @dtantsur

pkg/provisioner/ironic/ironic.go
```diff
@@ -139,14 +139,15 @@ func (p *ironicProvisioner) validateNode(ironicNode *nodes.Node) (errorMessage s
 func (p *ironicProvisioner) listAllPorts(address string) ([]ports.Port, error) {
 	var allPorts []ports.Port
 
-	opts := ports.ListOpts{
-		Fields: []string{"node_uuid"},
+	// Don't call the ironic api when the port mac address is empty.
```
dtantsur, line 142: Under what conditions do we provide an empty MAC address here? If it's a user error, shouldn't this call return an error?

zaneb: When we can't find a node by name, we look for it by port. The address is the bootMACAddress, which is optional. So I think returning an empty list is indeed the right thing here.

dtantsur: Honestly, I'd rather not call this function with empty address.

hex2dec: I will update this commit.

pkg/provisioner/ironic/ironic.go
```diff
@@ -249,35 +249,37 @@ func (p *ironicProvisioner) findExistingHost(bootMACAddress string) (ironicNode
 		}
 	}
 
-	// Try to load the node by port address
-	p.log.Info("looking for existing node by MAC", "MAC", bootMACAddress)
-	allPorts, err := p.listAllPorts(bootMACAddress)
+	// If mac address is present in host config, try to load the node by this port with mac address.
+	if bootMACAddress != "" {
```
zaneb, line 253: nit: an early return here would be easier to read.

hex2dec: a good coding tip 😃 . From the `findExistingHost` function, it supports multiple methods to find an existing node, and other methods may be added in the future, **the early return is not very good at showing the order in which all methods were tried**. Therefore, it is recommended to use the method to find an node by port **only** when the value of bootMACAddress is present.

PR #2102, opened 2024-12-05
 ✨ Remove kube-rbac-proxy and use controller run-time provided WithAuthenticationAndAuthorization filter instead 

kashifest:
❗ Following resources have been removed as part of this PR: 

- ClusterRole: baremetal-operator-metrics-reader
- ClusterRole: baremetal-operator-proxy-role
- ClusterRoleBinding: baremetal-operator-proxy-rolebinding
- Service: baremetal-operator-controller-manager-metrics-service

Short description: kube-rbac-proxy was historically used to protect the metrics endpoint. However, its usage has been discontinued in Kubebuilder. The default scaffold now leverages the WithAuthenticationAndAuthorization feature provided by Controller-Runtime. This change is trigerred from the fact that the image gcr.io/kubebuilder/kube-rbac-proxy is deprecated and will become unavailable sometime from early 2025.

Fixes: https://github.com/metal3-io/baremetal-operator/issues/2091


kashifest:
> /hold Please add this to the description so it is picked up in the release notes. I wonder if we should also mark it as ✨

Done

tuminoid:
@kashifest shall we do manual cherry-pick? 0.8 support extends beyond rbac-proxy deprecation period.

kashifest:
> @kashifest shall we do manual cherry-pick? 0.8 support extends beyond rbac-proxy deprecation period.

Good point, I will push it.

config/render/capm3.yaml
```diff
@@ -2618,16 +2562,6 @@ spec:
...
-      - args:
-        - --secure-listen-address=0.0.0.0:8443
-        - --upstream=http://127.0.0.1:8085/
-        - --logtostderr=true
-        - --v=10
-        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.16.0
-        name: kube-rbac-proxy
-        ports:
-        - containerPort: 8443
-          name: https
```
camilamacedo86, line 2630: Where are the RBAC roles now required for controller runtime?
I think you forgot those.

lentzi90: Oh! @kashifest we need to test this and probably add some new RBAC

kashifest: Indeed! I have put a PR for that https://github.com/metal3-io/baremetal-operator/pull/2116, I haven't tested them yet but those seem to me the missing RBACs

PR #2136, opened 2024-12-18
:bug: bump cert-manager to v1.12.14 and x/crypto to v0.31.0

tuminoid:
https://osv.dev/GO-2024-3282
Called.

https://osv.dev/GO-2024-3321
Not called. Patching while at it to silence the scanners.


Rozzii (commented):
/lgtm

PR #2144, opened 2024-12-19
:seedling: Fix BMO fixture-release-0.8 tag names

mquhuy:
Fixture overlay for release-0.8 was wrongly tagged with release-0.9 ref. Putting back the correct tags.


tuminoid (commented):
/lgtm

PR #2146, opened 2024-12-20
🌱 vulnerability fixes  github.com/docker/docker and golang.org/x/net

adilGhaffarDev:
vulnerability fixes
- https://osv.dev/GO-2024-3333 
- https://osv.dev/GO-2024-3005

Fixes #


apis/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/metal3-io/baremetal-operator/apis
 
-go 1.21
+go 1.22.8
```
tuminoid, line 3: These vulns are not called, so they're irrelevant. In release branch we should not bump go without good reason.

tuminoid (changes requested):
Find minimal set without go bump if possible, otherwise just close.

adilGhaffarDev:
> Find minimal set without go bump if possible, otherwise just close.

not possible closing it

PR #2154, opened 2024-12-24
🌱 Refactor and rename ValidateManagementAccess

dtantsur:
Validating access is not its primary responsibility: it's main job is to
make sure the Ironic node exists and is correctly configured. For a lack
of a perfect name, I'm calling it Register here.

The renamed method is moved to a new file and split into more manageable
pieces. In the future, we can even unit-test them separately.


lentzi90 (commented):
/approve

Rozzii (commented):
/lgtm

PR #2161, opened 2024-12-30
🌱 Add intervals while checking metrics service

kashifest:
Currently the periodic optional tests are failing to check if metrics service is available or not. Perhaps we need to wait for a while for the service to be available.


Rozzii (commented):
/approve

PR #2165, opened 2025-01-02
🐛 Fix metrics service validation

kashifest:
Since the `cmd` block is outside the `eventually` block, when the same `cmd` object is used multiple times, we see `exec: Stdout already set` error. Moving it inside the `eventually` block should solve it.


Rozzii (commented):
/approve

PR #2170, opened 2025-01-07
🐛  Fix BMO optional test by not checking metrics service

kashifest:
BMO optional test is not deploying BMO via e2e_suite_test.go but that is done separately in upgrade test. Hence the check is failing since we are not deploying BMO in e2e suite test but we are checking for metrics service to be up. This PR makes sure that metrics service is only checked when BMO is deployed.


lentzi90 (commented):
/approve

PR #2190, opened 2025-01-16
:bug: Add validation against cross-namespace secret references

### LINKED ISSUE #1965: Remove namespace field from secrets used by BMH API
The BareMetalHost CRD allows the userData, metaData, and networkData for
the provisioned host to be specified as links to k8s Secrets. There are
fields for both the Name and Namespace of the Secret, meaning that the
baremetal-operator will read a Secret from any namespace. A user with
access to create or edit a BareMetalHost can thus exfiltrate a Secret
from another namespace by using it as e.g. the userData for provisioning
some host (note that this need not be a real host, it could be a VM
somewhere).

We should fail with an error when a namespace is specified in the Secret
reference and it does not match the namespace of the BMH itself. We
should reject attempts to set such a configuration in the validating
webhook. And we should remove the namespace field in the next revision
of the API.

https://github.com/metal3-io/baremetal-operator/security/advisories/GHSA-pqfh-xh7w-7h3p is limiting the secrets to same namespace now so API cleanup and validating webhook are just improvements/cleanup to the issue.

**Environment:**

- Baremetal Operator version: 0.8.0
- Environment (metal3-dev-env or other): any

@zaneb @dtantsur

mquhuy:
Add to BMH creation a validation, which checks if any secret references is from namespaces other than BMH's. This is to prevent the secret leakage when a user references secrets from other ns just to get the secrets out.

Fixes #1965

### DIFF
--- a/apis/metal3.io/v1alpha1/baremetalhost_validation.go
+++ b/apis/metal3.io/v1alpha1/baremetalhost_validation.go
@@ -12,6 +12,9 @@ import (
 
 	"github.com/google/uuid"
 	"github.com/metal3-io/baremetal-operator/pkg/hardwareutils/bmc"
+	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
 var (
@@ -33,6 +36,8 @@ func (host *BareMetalHost) validateHost() []error {
 			errs = append(errs, err)
 		}
 	}
+
+	errs = append(errs, host.validateCrossNamespaceSecretReferences()...)
 
 	if raidErrors := validateRAID(host.Spec.RAID); raidErrors != nil {
 		errs = append(errs, raidErrors...)
@@ -308,4 +313,40 @@ func validateRebootAnnotation(rebootAnnotation string) error {
 	}
 
 	return nil
+}
+
+// validateCrossNamespaceSecretReferences validates that a SecretReference does not refer to a Secret
+// in a different namespace than the host resource.
+func validateCrossNamespaceSecretReferences(hostNamespace, hostName, fieldName string, ref *corev1.SecretReference) error {
+	if ref != nil &&
+		ref.Namespace != "" &&
+		ref.Namespace != hostNamespace {
+		return k8serrors.NewForbidden(
+			schema.GroupResource{
+				Group:    "metal3.io",
+				Resource: "baremetalhosts",
+			},
+			hostName,
+			fmt.Errorf("%s: cross-namespace Secret references are not allowed", fieldName),
+		)
+	}
+	return nil
+}
+
+// validateCrossNamespaceSecretReferences checks all Secret references in the BareMetalHost spec
+// to ensure they do not reference Secrets from other namespaces. This includes userData,
+// networkData, and metaData Secret references.
+func (host *BareMetalHost) validateCrossNamespaceSecretReferences() []error {
+	secretRefs := map[*corev1.SecretReference]string{
+		host.Spec.UserData:    "userData",
+		host.Spec.NetworkData: "networkData",
+		host.Spec.MetaData:    "metaData",
+	}
+	errs := []error{}
+	for ref, fieldName := range secretRefs {
+		if err := validateCrossNamespaceSecretReferences(host.Namespace, host.Name, fieldName, ref); err != nil {
+			errs = append(errs, err)
+		}
+	}
+	return errs
 }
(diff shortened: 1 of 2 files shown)


dtantsur (dismissed):

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -309,3 +314,34 @@ func validateRebootAnnotation(rebootAnnotation string) error {
...
+	if ref != nil &&
+		ref.Namespace != "" &&
+		ref.Namespace != hostNamespace {
+		return k8serrors.NewForbidden(
+			schema.GroupResource{
+				Group:    "metal3.io",
+				Resource: "baremetalhosts",
+			},
+			hostName,
+			fmt.Errorf("%s: cross-namespace Secret references are not allowed", fieldName),
```
tuminoid, line 328: Should we print the offending/expected namespace here?

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -309,3 +314,34 @@ func validateRebootAnnotation(rebootAnnotation string) error {
 
 	return nil
 }
+
+func validateCrossNamespaceSecretReferences(hostNamespace, hostName, fieldName string, ref *corev1.SecretReference) error {
```
tuminoid, line 318: Document the function

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -309,3 +314,34 @@ func validateRebootAnnotation(rebootAnnotation string) error {
...
+				Resource: "baremetalhosts",
+			},
+			hostName,
+			fmt.Errorf("%s: cross-namespace Secret references are not allowed", fieldName),
+		)
+	}
+	return nil
+}
+
+func (host *BareMetalHost) validateCrossNamespaceSecretReferences() []error {
```
tuminoid, line 334: Document the function.

apis/metal3.io/v1alpha1/baremetalhost_validation_test.go
```diff
@@ -678,6 +679,97 @@ func TestValidateCreate(t *testing.T) {
...
+					MetaData: &corev1.SecretReference{
+						Name:      "test-secret3",
+						Namespace: "different-namespace3",
+					},
+				},
+			},
+			oldBMH:    nil,
+			wantedErr: "baremetalhosts.metal3.io \"test\" is forbidden: userData: cross-namespace Secret references are not allowed", // Should catch at least one error
+		},
+		{
```
tuminoid, line 750: Add a test where namespace is nil, if that is ever allowed/possible to be nil.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -309,3 +314,34 @@ func validateRebootAnnotation(rebootAnnotation string) error {
 
 	return nil
 }
+
+func validateCrossNamespaceSecretReferences(hostNamespace, hostName, fieldName string, ref *corev1.SecretReference) error {
+	if ref != nil &&
+		ref.Namespace != "" &&
```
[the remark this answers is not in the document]
mquhuy: Namespace is of type string, it's never nil.

apis/metal3.io/v1alpha1/baremetalhost_validation_test.go
mquhuy: As said namespace cannot be nil.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
mquhuy: I don't think we should. And what would be the expected ns in this case? The secret's or the BMH's?

apis/metal3.io/v1alpha1/baremetalhost_validation.go
mquhuy: Hmm. OK I will, but imo these functions are quite self-explanatory, and the other validate...() func in this files don't seem to have docstring either.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
tuminoid: Since there is only one allowed namespace, maybe it does not make too much sense. Printing the offending namespace wouldn't hurt IMO, but I'm fine either way.

apis/metal3.io/v1alpha1/baremetalhost_validation.go
tuminoid: We are missing the linter for that, but want to enable it like in CAPM3, so no need to add future work. It is good to document them right away, it is the proper thing to do.

tuminoid (commented):
/lgtm

kashifest (commented):
/approve

PR #2191, opened 2025-01-16
🐛 Set e2e timeout to 1.5h

### LINKED ISSUE #2189: BMO e2e periodics run for maximum of 6 hours when stuck
When BMO e2e periodics get stuck in docker/minikube issue for example, they don't (always) exit, but they time out in 6 hours (GitHub maximum). Normal run is less than 0.5 hour. We should be good citizens and timeout at 1 hour, to free up resources/not generate unnecessary bill.

edit: there are optional combos that take over 1 hour, so timeout could be 1.5h. My bad.

image

/kind bug

mquhuy:
Save resources in case the test stucks.

Fixes #2189

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -22,6 +22,7 @@ jobs:
   test:
     name: E2E test
     runs-on: ${{ inputs.runner }}
+    timeout-minutes: 90
 
     steps:
     - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2


.github/workflows/e2e-test.yml
```diff
@@ -22,6 +22,7 @@ jobs:
   test:
     name: E2E test
     runs-on: ${{ inputs.runner }}
+    timeout-minutes: 60
```
tuminoid, line 25:
```suggestion
    timeout-minutes: 90
```
Sorry, there are some e2e combos that take more than hour, namely in optional periodics.

Rozzii (dismissed):
I also thin 90 min is more realistic (based on experience).
Please change the timeout to 90 min.

tuminoid (commented):
/retitle 🐛 Set e2e timeout to 1.5h
/lgtm

kashifest (commented):
/approve

PR #2174, opened 2025-01-09
:bug: Adding empty checksumType checks on BMH  to automatically detect hashing algorithm

fracappa:
Previously, if no checksum type was specified into the BMH CR, it automatically handled it as MD5, causing errors to whom used different hashing algorithms. 

**What this PR does / why we need it**:
It adds checks on the length of the checksum string to understand which hashing algorithm when the BMH CRs do not have an explicit `.spec.image.checksumType` field.

**Which issue(s) this PR fixes**:
Fixes #2172

PR #2192, opened 2025-01-16
Auto Checksum if checksumType is empty

### LINKED ISSUE #2172: Empty checksumType in BMH CRs is automatically defaulted to MD5
Hi everyone.

As a user I would like to report that BMH CR creation with empty `spec.image.checksumType`  is defaulted with MD5.

The line I'm referring to is this [one](https://github.com/metal3-io/baremetal-operator/blob/0633e68862f7d225072722f65f4bf9c7abb12ec2/apis/metal3.io/v1alpha1/baremetalhost_types.go#L1134) .

As you imagine, this causes errors if users have SHA256 or SHA512 as checksum. 

 I don't know if this is a desiderata, but I thought it would be nice if there would be some checks on the checksum string length to understand the hashing algorithm.
 
 
 Thanks for your attention.
 
 Regards,
 Francesco

/kind feature

mquhuy:
In case the checksumType is not provided, let ironic decide the checksum type, instead of setting it to MD5.

Fixes #2172


mquhuy:
/close
on favor of https://github.com/metal3-io/baremetal-operator/pull/2174

PR #2207, opened 2025-01-21
[WIP] :seedling: Allow vbmctl to create VM with multiple network interfaces

mquhuy:
vbmctl currently creates VM with 1 network interface. This is enough for BMO E2E, but to be able to have VMs that work for other purposes, having the flexibility to have multiple networks is required.


Rozzii (commented):
In general makes sense, will check after it passes the tests.

Rozzii:
/remove-lifecycle stale 
we have discussed offline that someone has to pick this up, most likely folks who will work on the dev-env decoupling @smoshiur1237 @Sunnatillo or maybe someone else?

sirpaleet:
Rebased, will start working on this.

sirpaleet:
Still working on this, focusing on other things atm

test/e2e/config/bmcs-fixture.yaml
```diff
@@ -5,7 +5,15 @@
   password: password
   address: "redfish+http://192.168.222.1:8000/redfish/v1/Systems/bmo-e2e-0"
   bootMacAddress: "00:60:2f:31:81:01"
+  networks:
+  - name: "baremetal-e2e"
+    macAddress: "00:60:2f:31:81:01"
+    ipAddress: "192.168.222.122"
```
lentzi90: Can we leave it out for fixture? Then we keep the file smaller and people don't get confused thinking that the fixture tests can actually use this. Plus, we test that the configuration is optional.

test/e2e/bmc.go
```diff
@@ -35,6 +46,8 @@ type BMC struct {
 	IPAddress string `yaml:"ipAddress,omitempty"`
 	// RootDeviceHints provides guidance for where to write the disk image.
 	RootDeviceHints metal3api.RootDeviceHints `yaml:"rootDeviceHints,omitempty"`
+	// Optional. Not needed for E2E tests
```
lentzi90:
```suggestion
    // Networks describes the network interfaces that should be added to the VM representing this BMH.
	// Optional. Not needed for E2E tests
```

test/e2e/bmc.go
```diff
@@ -10,6 +10,17 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
+type Network struct {
+	// Name of the network
```
lentzi90:
```suggestion
	// Name of the libvirt network
```

test/e2e/bmc.go
```diff
@@ -10,6 +10,17 @@ import (
...
+	NetworkName string `yaml:"name,omitempty"`
+	// MacAddress of the network
```
lentzi90:
```suggestion
	// MacAddress of the interface connected to the network
```

test/e2e/bmc.go
```diff
@@ -10,6 +10,17 @@ import (
...
+	MacAddress string `yaml:"macAddress,omitempty"`
+	// IPAddress of the network
```
lentzi90:
```suggestion
	// IPAddress to reserve for the MAC address in the network.
```

test/vbmctl/templates/VM.xml.tpl
```diff
@@ -25,7 +25,8 @@
     <disk type='file' device='disk'>
       <driver name='qemu' type='qcow2'/>
       <source file='{{ .PoolPath }}/{{ .Name }}-1.qcow2'/>
-      <target dev='vda' bus='virtio'/>
+      <target dev='sda' bus='scsi'/>
+      <address type='drive' controller='0' bus='0' target='0' unit='0'/>
```
lentzi90: This does not seem related. Please put it in a separate PR

test/vbmctl/templates/VM.xml.tpl
```diff
@@ -72,19 +73,19 @@
...
     </interface>
+    {{ end }}
     <serial type='pty'>
-      <log file='/var/log/libvirt/qemu/{{ .Name }}-serial0.log' append='on'/>
       <target type='isa-serial' port='0'>
         <model name='isa-serial'/>
       </target>
     </serial>
     <console type='pty'>
-      <log file='/var/log/libvirt/qemu/{{ .Name }}-serial0.log' append='on'/>
```
lentzi90, line 87: These changes are not related either. I do not think we should change the log files.

lentzi90 (commented):
This seems to be quite close to working, but there are some unrelated changes that I think makes this unnecessarily complicated. If we just focus on the network changes it should be fairly simple to get it working and merged

sirpaleet:
Thanks for the input! I also see that there are some changes made in #2590 conflicting with this as well. Yes this does not work as is but I'll look into all the comments :)

test/vbmctl/templates/VM.xml.tpl
sirpaleet: I don't know what the logic behind this was initially so I'll remove it then

test/vbmctl/templates/VM.xml.tpl
sirpaleet: Not sure about this either so I'll try without then

sirpaleet:
Still not working but recommended changes added, tested and seemingly no additional errors at least in the limited tests and some even disappeared. Will see however.

test/vbmctl/main.go
```diff
@@ -187,81 +211,66 @@ func CreateLibvirtVM(conn *libvirt.Connect, name, networkName, macAddress string
...
 
 func main() {
-	var name = flag.String(
-		"name", "BMH-0", "The name of the VM to create")
-	var networkName = flag.String(
-		"network-name", "baremetal-e2e", "The name of the network that the new VM should be attached to")
-	var macAddress = flag.String(
-		"mac-address", "00:60:2f:31:81:01", "Mac address of the VM on the network")
-	var ipAddress = flag.String(
-		"ip-address", "", "IP address of the VM on the network")
```
lentzi90, line 247: I would prefer to keep these flags. We don't use them in CI, but I hope that they will one day be useful for users

sirpaleet: At least the compiler complains about this not being used, and linter won't let it pass either. I'll comment them out but leave there?

lentzi90: Probably because the code that was using them is also removed below. You need to add that back also. There should be a few lines making use of these variables if no config file is specified.

sirpaleet: Hmm I think I have to modify it some then, but you think it would be good to have a fallback cofig if no yaml file is specified? I am not sure why it was first removed, but I could ask Huy and see if he remembers.

sirpaleet: According to Huy:

"I think the reasons these were deleted was that it would be more complex to keep them than to remove them. Now that you are in theory able to have an unlimited number of interfaces, you will need to modify the network+address flags so that they allow an unlimited numbers of network interfaces, and also they should be paired so that an ip doesn't get assigned to a wrong interface. It's doable but I just didn't feel the need, as the PR was getting big and we didn't use these configs anyway.
 
If you want to keep them, a suggestion is to merge these flags into one, and allow the new flag to be put multiple times. I think some new param also needs to be put into that new flag (like interface name, etc.). "

So would it be better to keep these flags and do the work to change the flags/do a new one?

lentzi90: I think there is a misunderstanding here. The config file and the other flags do not have to support the same thing. These flags are there for simple manual usage. The config file is more convenient for advanced use-cases.
What I am trying to say is that you don't need to support multiple interfaces for anything other than the config file. If users want that, they have to use the config file. If other flags are used, then only a single interface is supported.

sirpaleet: This needs some organizing then but I'll try to make it so that both ways work!

test/vbmctl/main.go
```diff
@@ -273,12 +282,12 @@ func main() {
 
 	for _, bmc := range bmcs {
 		if bmc.IPAddress != "" {
```
sirpaleet: From what I understand is that this is only defined if it is explicitly stated in the `ip-address` flag (EDIT: or with the `configFile` in case only one network is wanted), and with multiple networks these may be explicitly stated inside the Network struct in the `configFile`, and thus optional, which is why using the current logic in `CreateLibvirtVMWithReservedAddress` probably will not work (it assumes that all ip addressess are explicitly stated, even though `bmc.IPAddress` is only given a value when one network is wanted). I think I'll have to move the run through this explicit statement for loop on L215 to the main function or create a helper to check this, and then call the `CreateLibvirtVMWithReservedAddress` based on each case --Input? (not tested)

sirpaleet: Or it could probably be through the `configFile` as well but assuming that people might not understand to always specify the `IPAddress` when they also want to create all the multiple networks with reserved addresses as well, so they might just use the networks structure and only that. ... Right?

sirpaleet: @lentzi90 any opinions on this? I am doing quite a lot of restructuring on this file specifically. Is the multiple networks/one network an either/or situation, or should the users have the option to specify both?

lentzi90: Sorry I don't understand what you are asking. Do you mean if they should be able to combine both config and individual flags? I would say no. That was not possible before and probably should not be.
Or are you wondering about default values? Some of the flags have default values, so the user does not have to specify anything for them and they will still get that value. The IP address is not mandatory. We should not require that every VM always has a reserved IP.

On a high level, this is how I would want it to work:

Use-case A) As a developer I want to easily create VMs to use as BMHs, either manually or through tilt or similar.
- Create a single VM with all defaults: `vbmctl`
- Create another VM with different name and MAC: `vbmctl --name bmh-b --mac-address 00:60:2f:31:81:02`

Use-case B) For advanced scenarios, I want to create complex VMs with multiple network interfaces.
- Use config file: `vbmctl --yaml-source-file advanced.yaml`

sirpaleet: Yes! Sorry I formatted the questions  a bit scrambly. The flags part makes sense and I understand that a bit clearer now, thanks!

Mainly, if the way to format both a singular network and the netwoks structure at the same time is possible and should be accounted for. Or with the upper level `ipAddress` and `bootMacAddress` specified and no `networks` or the `ipAddress` or `macAddress` under it, and the opposite (aka. the either/or situation is assumed to be used). 

Additionally, should all `networks` have the option to have / not have the `ipAddress` (the way it works now I think)?

Either way, the current way this is implemented always assumes that the way to specify if you want a reserved IP is via the upper level `ipAddress` value, and not the network-specific ones.

lentzi90: Ok I see! I think we do not need to support reserved addresses with the config file or with multiple networks. At least for CAPM3 e2e tests it should not be needed. However, if it turns out to be easier to implement it that way, I would not have an issue with it.

There is no need to support single network (flags) and the networks structure (config) at the same time. If a config file is given, just ignore the other flags.

sirpaleet: Hmm, well if the reserved IP addresses are not needed should I just take away the option to specify it under the "networks" at all – or leave as is so that it is possible to specify / not, if wanted? Including the possibility now isn't a big change since the way this was originally done seems to take that into account.

Taking away the specification would only leave the network's name & macAddress to be speficied by the user. But would that make the whole setup less complicated then? I'll push a WIP on this now again and explain on that more.

test/vbmctl/main.go
```diff
@@ -229,7 +254,41 @@ func CreateLibvirtVMWithReservedIPAddress(conn *libvirt.Connect, macAddress, nam
...
+		network, err := conn.LookupNetworkByName(networkName)
+		if err != nil {
+			return err
+		}
+		err = updateNetwork(network, bmc.BootMacAddress, bmc.Name, bmc.IPAddress)
+		if err != nil {
+			return err
+		}
+
+	} else {
```
sirpaleet: This apparently a linter problem, but my editor adds this automatically when saving and this is used elsewhere too -- fixed

test/vbmctl/main.go
```diff
@@ -272,13 +331,16 @@ func main() {
 	defer conn.Close()
 
 	for _, bmc := range bmcs {
-		if bmc.IPAddress != "" {
-			if err = CreateLibvirtVMWithReservedIPAddress(conn, bmc.BootMacAddress, bmc.Name, bmc.IPAddress, *networkName); err != nil {
+		// If IPAddress on upper level is specified or if there are
+		// multiple networks, there might need to be a VM with reserved
+		// address/addresses.
+		if bmc.IPAddress != "" || len(bmc.Networks) != 0 {
```
sirpaleet: For this version, spedicying an IPAddress on each network is possible, and since the function was "CreateLibvirtVM..." all network checks are done there for each case. The function is capitalized so I am assuming there is a reason for that (used somewhere else later?) so this does what it is named as and not just for one network.

test/vbmctl/main.go
```diff
@@ -140,28 +173,30 @@ func CreateVolume(conn *libvirt.Connect, volumeName, poolName, poolPath string,
...
-		Network    string
-		MacAddress string
-		PoolPath   string
-	}{
-		Name:       name,
-		Network:    networkName,
-		MacAddress: macAddress,
-		PoolPath:   poolPath,
+	var data Host
+	if len(bmc.Networks) == 0 {
```
sirpaleet: This now works with either one networkName specified/default, or the networks structure. So it is not erasing the option to have only a singular network specified.

test/vbmctl/main.go
```diff
@@ -229,7 +254,41 @@ func CreateLibvirtVMWithReservedIPAddress(conn *libvirt.Connect, macAddress, nam
...
+		}
+		err = updateNetwork(network, bmc.BootMacAddress, bmc.Name, bmc.IPAddress)
+		if err != nil {
+			return err
+		}
+
+	} else {
+		for _, net := range bmc.Networks {
+			// Checking if this network has IP specified.
+			if bmc.IPAddress != "" {
```
sirpaleet: The network-specific IP is checked here: uhh needs edit it seems :D -- fixed

test/e2e/config/bmcs-redfish-virtualmedia.yaml
```diff
@@ -4,6 +4,10 @@
   bootMacAddress: "00:60:2f:31:81:01"
   name: "bmo-e2e-0"
   ipAddress: "192.168.222.122"
+  networks:
+  - name: "baremetal-e2e"
+    macAddress: "00:60:2f:31:81:01"
+    ipAddress: "192.168.222.122"
```
lentzi90, line 12: Is this needed, or can we skip it?

test/e2e/config/bmcs-redfish-virtualmedia.yaml
```diff
@@ -12,5 +16,9 @@
   bootMacAddress: "00:60:2f:31:81:02"
   name: "bmo-e2e-1"
   ipAddress: "192.168.222.123"
+  networks:
+  - name: "baremetal-e2e"
+    macAddress: "00:60:2f:31:81:02"
+    ipAddress: "192.168.222.123"
```
lentzi90, line 24: Same question here

test/e2e/config/bmcs-ipmi.yaml
```diff
@@ -4,6 +4,10 @@
   bootMacAddress: "00:60:2f:31:81:01"
   name: "bmo-e2e-0"
   ipAddress: "192.168.222.122"
+  networks:
+  - name: "baremetal-e2e"
+    macAddress: "00:60:2f:31:81:01"
+    ipAddress: "192.168.222.122"
```
lentzi90, line 12: Can we remove this?

test/e2e/config/bmcs-ipmi.yaml
```diff
@@ -12,5 +16,9 @@
   bootMacAddress: "00:60:2f:31:81:02"
   name: "bmo-e2e-1"
   ipAddress: "192.168.222.123"
+  networks:
+  - name: "baremetal-e2e"
+    macAddress: "00:60:2f:31:81:02"
+    ipAddress: "192.168.222.123"
```
lentzi90, line 24: And this?

test/e2e/config/bmcs-redfish.yaml
```diff
@@ -4,6 +4,10 @@
   bootMacAddress: "00:60:2f:31:81:01"
   name: "bmo-e2e-0"
   ipAddress: "192.168.222.122"
+  networks:
+  - name: "baremetal-e2e"
+    macAddress: "00:60:2f:31:81:01"
+    ipAddress: "192.168.222.122"
```
lentzi90, line 12: Can we remove this?

test/e2e/config/bmcs-redfish.yaml
```diff
@@ -12,5 +16,9 @@
   bootMacAddress: "00:60:2f:31:81:02"
   name: "bmo-e2e-1"
   ipAddress: "192.168.222.123"
+  networks:
+  - name: "baremetal-e2e"
+    macAddress: "00:60:2f:31:81:02"
+    ipAddress: "192.168.222.123"
```
lentzi90, line 24: And this?

lentzi90 (commented):
This looks quite ok now.
It is not so nice that we have duplicate information (e.g. IP address and MAC address under networks and also directly in the BMC struct), but this is really a symptom of us mixing two different things together. We can clean it up separately, as I mentioned, by completely separating the vbmctl config from the e2e BMC config.
For the e2e config we would have only the BMC struct (without networks) and for vbmctl we would have a new struct that included the networks.

test/e2e/config/bmcs-redfish-virtualmedia.yaml
sirpaleet: Hmm I am assuming these are just to see that the structure works, now that they aren't a necessity and the old way can be used also, I'll take them off. I still would need to test that the current way works but I can do that locally now!

sirpaleet:
Additionally, many of the local scripts assume that the network that is used is always "baremetal-e2e" and only that. I might need to add a note on that in case someone intends to use them with this multiple networks structure now.

sirpaleet:
> For the e2e config we would have only the BMC struct (without networks) and for vbmctl we would have a new struct that included the networks.

Hmm, I can look at starting this today. Either way I think most of the rest of the progress for this on my part will happen today, and someone would need to pick up on where I leave this. I'll do my best to have this not be a total mess!

EDIT: Did not get to this yet, was fixing errors first

test/vbmctl/templates/VM.xml.tpl
```diff
@@ -72,11 +72,21 @@
       <target chassis='6' port='0x15'/>
       <address type='pci' domain='0x0000' bus='0x00' slot='0x02' function='0x5'/>
     </controller>
-    <interface type='network'>
+    {{ if not .Networks }}
```
sirpaleet: Fixing this so that the old way can be used here too

test/vbmctl/main.go
```diff
@@ -140,28 +174,30 @@ func CreateVolume(conn *libvirt.Connect, volumeName, poolName, poolPath string,
...
+			Name:       bmc.Name,
+			Network:    networkName,
+			MacAddress: bmc.BootMacAddress,
+			PoolPath:   poolPath,
+		}
+	} else {
+		data = Host{
+			Name:     bmc.Name,
+			Networks: bmc.Networks,
+			PoolName: poolName,
```
sirpaleet: I am not quite sure why this was added here initially, in place of the macAddress, needs to be tested

test/e2e/bmc.go
```diff
@@ -10,6 +10,17 @@ import (
...
+type Network struct {
+	// Name of the libvirt network.
+	NetworkName string `yaml:"name,omitempty"`
+	// MacAddress of the interface connected to the network.
+	MacAddress string `yaml:"macAddress,omitempty"`
+	// IPAddress to reserve for the MAC address in the network.
+	IPAddress string `yaml:"ipAddress,omitempty"`
+}
+
+type Networks []Network
```
sirpaleet: Maybe `yaml:"networks,omitempty"` Here? -- does this mean it is not optional now without it? EDIT: On the "Networks" within the struct (L48) I commented wrong line

lentzi90:
/retitle 🌱 vbmctl: support multiple network interfaces

Alright we have passing tests and I have confirmed locally that it works also with multiple interfaces, with or without reserved IPs.

One thing that was a bit surprising to me was that each DHCP host entry has to have a unique name. I would have thought that it is enough with unique MAC. The name is picked up as hostname by IPA, so it becomes a bit strange if we have multiple numbered interfaces that one of them becomes hostname. Anyway, we should not rely on this for setting the hostname. I changed the re-inspection test to just check that the hostname changed, instead of comparing it against the BMC name.

Another thing I want to mention is the reserved addresses. I commented earlier and asked if we can remove them. That should be possible to do, but I opted to keep them for now to avoid introducing more changes. We now have duplicate information for the network in the BMC config file. This is because everything under `networks` is for vbmctl. This is not read by the e2e tests. We should eventually split this out to its own file to avoid mixing things up.

/cc @adilGhaffarDev

Rozzii (commented):
It i passing the tests, I see there are improvement requests from @lentzi90  but let's do those in a followup PR.
/approve

smoshiur1237 (commented):
LGTM

PR #2209, opened 2025-01-21
🌱 Switch e2e to kind

### LINKED ISSUE #1783: [flake] Docker restart fails in BMO e2e
### Which jobs are flaking?

BMO e2e tests

### Which tests are flaking?

Randomly any of e2e variants

### Since when has it been flaking?

Since moving to GH actions.

### Jenkins link

https://github.com/metal3-io/baremetal-operator/actions/runs/9477201984/job/26111334840

```console
n 12 05:29:25 minikube systemd[1]: docker.service: Deactivated successfully.
Jun 12 05:29:25 minikube systemd[1]: Stopped Docker Application Container Engine.
Jun 12 05:29:25 minikube systemd[1]: Starting Docker Application Container Engine...
Jun 12 05:29:25 minikube dockerd[796]: time="2024-06-12T05:29:25.668001724Z" level=info msg="Starting up"
Jun 12 05:30:25 minikube dockerd[796]: failed to start daemon: failed to dial "/run/containerd/containerd.sock": failed to dial "/run/containerd/containerd.sock": context deadline exceeded
Jun 12 05:30:25 minikube systemd[1]: docker.service: Main process exited, code=exited, status=1/FAILURE
Jun 12 05:30:25 minikube systemd[1]: docker.service: Failed with result 'exit-code'.
Jun 12 05:30:25 minikube systemd[1]: Failed to start Docker Application Container Engine.
```

_No response_

### Label(s) to be applied

/kind flake

lentzi90:

Minikube is having troubles starting sometimes.
It was nice to work with since the VM could easily be attached to the same network as the BMH VMs, but it is possible to work around that also with kind.

It also removes overlays that are no longer used and the upgrade tests for ironic-24.0 as these are not supported/needed anymore. With that we also get rid of the last inspector code from the scripts :tada: 

Fixture tests were only uploading artifacts on success. New they upload also on failure, just like e2e.

I have also tried stabilizing the fixture tests. It seems like there is some condition when namespace deletion takes a long time. From the BMO logs I saw that BMO was trying to create some resources in the terminating namespace. In an attempt to avoid this I added BMH deletion to the cleanup before deleting the namespace. That way at least BMO should be aware what is coming. Not sure if that is the root cause of the slow namespace deletion though.

Fixes #1783


config/overlays/e2e-release-0.4/ironic.env
tuminoid, line 1: Can we clean this up?

config/overlays/e2e-release-0.5/ironic.env
tuminoid, line 1: This too.

ironic-deployment/overlays/e2e-release-24.0-with-inspector/ironic_bmo_configmap.env
tuminoid, line 1: And this entire directory.

tuminoid (commented):
Cleanup suggestion. Less work to update those.

ironic-deployment/overlays/e2e-release-24.0-with-inspector/ironic_bmo_configmap.env
lentzi90: Looks like we are still using this one here: https://github.com/metal3-io/baremetal-operator/blob/23c54cfdfffe3f2e9f8c544a07c8b02921f3490d/test/e2e/config/ironic.yaml#L85
The rest were all good to go though :+1:

lentzi90:
Alright I think this is ready for review. It became a bit larger than planned... Some of it can definitely be split into separate PRs if you wish.
I have run every test the is locally (including fixture and optional upgrade jobs) and all have passed. The fixture test is actually what I am most unsure about. It was quite flaky so I am not completely sure if I got it stable with this. The flakiness is anyway unrelated to this PR as far as I can tell.

test/e2e/config/ironic.yaml
```diff
@@ -2,6 +2,8 @@ images:
 # Use locally built e2e images
 - name: quay.io/metal3-io/baremetal-operator:e2e
   loadBehavior: tryLoad
+# - name: quay.io/metal3-io/ironic:local
```
tuminoid, line 5: Why have this added but commented out?

tuminoid (commented):
One nit.

Also, @kashifest brought up https://github.com/medyagh/setup-minikube as a way to set up minikube in a GH action.  I did not check it out myself if it would actually even be flexible enough for our needs. Did you check that?

test/e2e/config/ironic.yaml
lentzi90: Oh I forgot about it!
@Rozzii you added the option to use a locally built ironic image recently. This is basically all that is required for it as far as I understand. If you uncomment this, the image is loaded. The rest is handled by changing the manifests directly or by pointing to another kustomization to deploy below.
Is this acceptable?

Previous process:
- Set `LOAD_LOCAL_IRONIC=true` and build the image
- Change e2e config to deploy ironic from [e2e-local-ironic](https://github.com/metal3-io/baremetal-operator/tree/main/ironic-deployment/overlays/e2e-local-ironic)

Process after this PR:
- Build the image and comment out these lines in e2e config
- Change e2e config to deploy ironic from e2e-local-ironic

Rozzii: If we move to kind then the image transfer is not needed anymore  as kind runs on the host VM so in that context I am fine with the changes.
 
I am fine with doing it via the config file although I prefer env variable based config because I am a bit worried that it is easy to accidentally commit changes to ironic.yaml.

Rozzii (commented):
/lgtm

tuminoid (commented):
/lgtm

:rocket: let's get rid of the minikube/docker flake!
Thank you @lentzi90 for taking time to do this.

PR #2221, opened 2025-01-28
:seedling: Fix ensure_kind.sh filename

mquhuy:
In ensure_kind.sh, the downloaded binary file's name is kind-linux-amd64, so we should run `sudo install kind-linux-amd64` instead of `sudo install kind`.


tuminoid (commented):
Good catch. We normally have 0.26.0 anyways, so this path is not triggered, but I manually tested that it would fail, in case we'd need to get newer kind installed.

/lgtm

lentzi90 (commented):
/approve

PR #2223, opened 2025-01-28
🌱 E2E vbmctl and ssh boot check refactor

lentzi90:

This refactors and simplifies vbmctl by:
- updating doc strings
- renaming functions and variables
- reusing one libvirt connection instead of creating a new for every operation
- moving the ssh port to the e2e config, since vbmctl cannot influence it anyway
- adding network name to the BMC config to unify it with the vbmctl flags

This all comes from trying to switch e2e to UEFI. It has grown too large so to avoid I monster PR, I'm now trying to push one piece at a time.

There is still a lot that could be done for vbmctl. In particular, reusing the same config struct for the tests and for creating the VMs is not ideal. We end up entangling the way we emulate bare metal with the actual test code, which should not care, nor know about it. It is also easy to get confused when some of the fields are ignored by either part of the system.

### DIFF
--- a/test/e2e/bmc.go
+++ b/test/e2e/bmc.go
@@ -9,28 +9,28 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
-// BMC defines a BMH to use in the tests.
+// BMC defines connection details for a baseboard management controller
+// and other details needed for creating a virtual machine related to it.
 type BMC struct {
-	// BMC initial username
+	// User is the username for accessing the BMC.
 	User string `yaml:"user,omitempty"`
-	// BMC initial password
+	// Password is the password for accessing the BMC.
 	Password string `yaml:"password,omitempty"`
-	// BMC initial address
+	// Address of the BMC, e.g. "redfish-virtualmedia+http://192.168.222.1:8000/redfish/v1/Systems/bmo-e2e-1".
 	Address string `yaml:"address,omitempty"`
-	// BMC Mac address
+	// BootMacAddress is the MAC address of the VMs network interface.
 	BootMacAddress string `yaml:"bootMacAddress,omitempty"`
-	// The Hostname of the node, which will be read into BMH object
-	HostName string `yaml:"hostName,omitempty"`
-	// The IP address of the node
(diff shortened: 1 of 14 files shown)


tuminoid:
/test metal3-ubuntu-e2e-integration-main
Just in case ... :sweat_smile:

test/e2e/bmc.go
```diff
@@ -9,28 +9,28 @@ import (
...
+	// This will be paired with the MAC address in the DHCP configuration.
+	// Example: 192.168.222.122
 	IPAddress string `yaml:"ipAddress,omitempty"`
-	// Optional. Only needed if e2eConfig variable
-	// SSH_CHECK_PROVISIONED is true
-	SSHPort string `yaml:"sshPort,omitempty"`
 }
 
-func LoadBMCConfig(configPath string) (*[]BMC, error) {
+func LoadBMCConfig(configPath string) ([]BMC, error) {
```
mquhuy, line 33: I'm wondering why this change? Isn't using a pointer better for speed and memory usage?

mquhuy (commented):
Thanks. This is imo a good improvement. Just have a small question regarding the implementation details.

Otherwise lgtm.

test/e2e/bmc.go
lentzi90: Well, a slice is already using a pointer actually. There may be cases when it is useful to still use a pointer to a slice, but in practice I have very rarely come across situations when this is done. It is easier, less confusing and practically no difference for what we do here.

Here is an excellent answer about it:

> A slice, logically, is a struct of three fields: a pointer to the slice's backing array which actually contains the slice's elements, and two ints: the capacity of the slice and its length. When you pass around and assign a slice value, these struct values are copied: a pointer and two integers. As you can see, when you pass a slice value around the backing array is not copied—only a pointer to it.

https://stackoverflow.com/questions/72978660/is-it-correct-to-use-slice-as-item-because-slice-is-by-default-pointer

mquhuy: Thanks. 
/lgtm

PR #2227, opened 2025-01-30
🐛  Do not create HFC for non-redfish BMH

iurygregory:
Currently we are creating HFC for all BMH.
HFC only works for redfish based hardware, so we should only create when we detect that is redfish is in use.


tuminoid (commented):
Can we have test added for this?  @iurygregory

PR #2229, opened 2025-01-31
✨ Add support for Disablepoweroff

derekhiggins:
Add support for disable_power_off
as described in https://github.com/metal3-io/metal3-docs/pull/478


iurygregory:
Seems like a CI issue 
```
dial tcp: lookup proxy.golang.org on 10.96.0.10:53: server misbehaving
```

PR #2238, opened 2025-02-07
:bug: Set ownerReference for dataImage during BMH Reconcile

hroyrh:
Add the ownerReference for dataImage as soon as its created, so that in case of forced BMH deletion the dataImage is not left orphaned.

### DIFF
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -141,6 +141,20 @@ func (r *BareMetalHostReconciler) Reconcile(ctx context.Context, request ctrl.Re
 		}
 	}
 
+	// If DataImage exists, add its ownerReference
+	dataImage := &metal3api.DataImage{}
+	err = r.Get(ctx, request.NamespacedName, dataImage)
+	if !(err != nil || ownerReferenceExists(host, dataImage)) {
+		if err := controllerutil.SetControllerReference(host, dataImage, r.Scheme()); err != nil {
+			return ctrl.Result{}, fmt.Errorf("could not set bmh as controller, %w", err)
+		}
+		if err := r.Update(ctx, dataImage); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failure updating dataImage status, %w", err)
+		}
+
+		return ctrl.Result{Requeue: true}, nil
+	}
+
 	hostData, err := r.reconcileHostData(ctx, host, request)
 	if err != nil {
 		return ctrl.Result{}, errors.Wrap(err, "Could not reconcile host data")
@@ -1632,18 +1646,6 @@ func (r *BareMetalHostReconciler) handleDataImageActions(prov provisioner.Provis
 		}
 		// Error reading the object - requeue the request.
 		return actionError{fmt.Errorf("could not load dataImage, %w", err)}
-	}
-
-	// Set ControllerReference to DataImage
-	if !ownerReferenceExists(info.host, dataImage) {
-		if err := controllerutil.SetControllerReference(info.host, dataImage, r.Scheme()); err != nil {
-			return actionError{fmt.Errorf("could not set bmh as controller, %w", err)}
-		}
-		if err := r.Update(info.ctx, dataImage); err != nil {
-			return actionError{fmt.Errorf("failure updating dataImage status, %w", err)}
-		}
-
-		return actionContinue{}
 	}
 
 	// Update reconciliation timestamp for dataImage


controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -141,6 +141,20 @@ func (r *BareMetalHostReconciler) Reconcile(ctx context.Context, request ctrl.Re
 		}
 	}
 
+	// If DataImage exists, add its ownerReference
```
zaneb, line 144: nit: Could this be a function we call, instead of dropping more code into the main Reconcile function?

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -141,6 +141,20 @@ func (r *BareMetalHostReconciler) Reconcile(ctx context.Context, request ctrl.Re
...
+	err = r.Get(ctx, request.NamespacedName, dataImage)
+	if !(err != nil || ownerReferenceExists(host, dataImage)) {
+		if err := controllerutil.SetControllerReference(host, dataImage, r.Scheme()); err != nil {
+			return ctrl.Result{}, fmt.Errorf("could not set bmh as controller, %w", err)
+		}
+		if err := r.Update(ctx, dataImage); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failure updating dataImage status, %w", err)
+		}
+
+		return ctrl.Result{Requeue: true}, nil
```
zaneb, line 155: I don't see any need to return early here.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -141,6 +141,20 @@ func (r *BareMetalHostReconciler) Reconcile(ctx context.Context, request ctrl.Re
...
 
+	// If DataImage exists, add its ownerReference
+	dataImage := &metal3api.DataImage{}
+	err = r.Get(ctx, request.NamespacedName, dataImage)
+	if !(err != nil || ownerReferenceExists(host, dataImage)) {
+		if err := controllerutil.SetControllerReference(host, dataImage, r.Scheme()); err != nil {
+			return ctrl.Result{}, fmt.Errorf("could not set bmh as controller, %w", err)
+		}
+		if err := r.Update(ctx, dataImage); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failure updating dataImage status, %w", err)
```
zaneb, line 152: This seems pretty dangerous, as any failure at all to do with the DataImage (e.g. missing RBAC permissions) will now prevent the BMH itself from ever getting reconciled.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -141,6 +141,20 @@ func (r *BareMetalHostReconciler) Reconcile(ctx context.Context, request ctrl.Re
 		}
 	}
 
+	// If DataImage exists, add its ownerReference
+	dataImage := &metal3api.DataImage{}
+	err = r.Get(ctx, request.NamespacedName, dataImage)
+	if !(err != nil || ownerReferenceExists(host, dataImage)) {
+		if err := controllerutil.SetControllerReference(host, dataImage, r.Scheme()); err != nil {
```
zaneb, line 148: We should never set a Controller reference. Use `SetOwnerReference` always and everywhere.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -141,6 +141,20 @@ func (r *BareMetalHostReconciler) Reconcile(ctx context.Context, request ctrl.Re
...
+			return ctrl.Result{}, fmt.Errorf("could not set bmh as controller, %w", err)
```
zaneb, line 149: This is unnecessarily dangerous, because things we have no control over (like e.g. something else sets an owner reference on the DataImage with the controller bit set) will prevent the BMH itself from ever getting reconciled.

controllers/metal3.io/baremetalhost_controller.go
hroyrh: Makes sense. Would ignoring and continuing make sense here, since the failure might potentially block bmh reconciling from happening ?

controllers/metal3.io/baremetalhost_controller.go
hroyrh: Looking at go docs, `SetControllerReference` triggers reconciliation for owner object when controlled object changes - is that the only issue or are there some other side effects as well which are not desirable in this case ?

controllers/metal3.io/baremetalhost_controller.go
hroyrh: Makes sense, I think raising an error message in the logs and continuing makes sense in this scenario.

controllers/metal3.io/baremetalhost_controller.go
hroyrh: Good idea, I will move this to a separate function.

controllers/metal3.io/baremetalhost_controller.go
zaneb: Owner reference is enough to trigger reconciliation iff you have `Owns()` (e.g. https://github.com/hroyrh/baremetal-operator/blob/0efbb139a9dfe3b5011780bd725968f7b76c91ff/controllers/metal3.io/baremetalhost_controller.go#L2294) in the controller setup.
A Controller reference is just an owner reference with the `controller` flag set. This makes the reference mutually exclusive. It exists to [solve a specific use case for ReplicaSet](https://github.com/kubernetes/design-proposals-archive/blob/acc25e14ca83dfda4f66d8cb1f1b491f26e78ffe/api-machinery/controller-ref.md) that ~nobody else has.

PR #2243, opened 2025-02-12
:bug: bump golang to 1.23.6

tuminoid:
This is to get rid of https://osv.dev/GO-2025-3447, where code path is called, but the vulnerability is only on ppc64, so it is not exploitable.


Rozzii (commented):
/lgtm

PR #2244, opened 2025-02-12
:seedling: pin osv-scanner image in verify-release.sh

tuminoid:
osv-scanner:latest image is built from unpublished v2 branch now, which does not support same command line args as v1. Pin it to v1.9.2 for now and bump the command line usage to v2 compatible when it is published.


lentzi90 (commented):
/approve

PR #2249, opened 2025-02-13
:bug: Add BMH OwnerReference for DataImage right after its creation 

hroyrh:

This is followup to https://github.com/metal3-io/baremetal-operator/pull/2238, fixing issues raised in the previous pr : https://github.com/metal3-io/baremetal-operator/pull/2238/files/0efbb139a9dfe3b5011780bd725968f7b76c91ff

Also updates the order of checks in the dataImage reconcile ( as per Zane's [comment here](https://github.com/metal3-io/baremetal-operator/pull/2249#discussion_r1955816469) )  :
1. paused annotation
2. owner reference
3. finalizer
4. detached host

### DIFF
--- a/controllers/metal3.io/dataimage_controller.go
+++ b/controllers/metal3.io/dataimage_controller.go
@@ -31,6 +31,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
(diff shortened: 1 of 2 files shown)


controllers/metal3.io/dataimage_controller.go
```diff
@@ -140,6 +142,18 @@ func (r *DataImageReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		return ctrl.Result{Requeue: false}, nil
 	}
 
+	// If DataImage exists, add its ownerReference
```
zaneb, line 145: I think the best order to check stuff in would be:

1. paused annotation
2. owner reference
3. finalizer
4. detached host

hroyrh: Done

honza:
The patch does more than the description suggests.  It would be great if we could update the commit message with more details.  Also, consider squashing the two commits (the beauty of git is that we can pretend we got it right on the first try).

PR #2222, opened 2025-01-28
🌱 Switch e2e to kind

lentzi90:
**What this PR does / why we need it**:

This is a manual backport of https://github.com/metal3-io/baremetal-operator/pull/2209 and includes https://github.com/metal3-io/baremetal-operator/pull/2221

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

PR #2276, opened 2025-03-03
🌱 E2E: Use kind instead of minikube

lentzi90:

Minikube is having troubles starting sometimes.
It was nice to work with since the VM could easily be attached to the same network as the BMH VMs, but it is possible to work around that also with kind.

This is a manual backport of (parts of) https://github.com/metal3-io/baremetal-operator/pull/2222

Fixes #


ironic-deployment/overlays/e2e-release-24.0-with-inspector/ironic_bmo_configmap.env
```diff
@@ -1,6 +1,6 @@
 HTTP_PORT=6180
-PROVISIONING_IP=192.168.222.199
-CACHEURL=http://192.168.222.199/images
-IRONIC_FAST_TRACK=true
```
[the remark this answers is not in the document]
lentzi90: I removed it to reduce the differences between 0.8 and main. Removing it does not disable fast track because it defaults to true in the ironic-image

hack/e2e/ensure_kind.sh
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
+
+set -eux
+
+USR_LOCAL_BIN="/usr/local/bin"
+MINIMUM_KIND_VERSION=v0.26.0
```
tuminoid, line 6: We have kind 0.23.0 in go.mod, but here we want 0.26.0, is that on purpose?

lentzi90: That is a mistake. I didn't pay enough attention when cherry-picking. Now that I think about it I am not sure we need to even install kind. :thinking: 
With minikube we created the cluster before the test started. Now we create it as part of the test suite itself, which makes this script redundant as far as I can tell. I can push a PR on main to see if we can remove it.

lentzi90: Checking here: https://github.com/metal3-io/baremetal-operator/pull/2277

tuminoid: I added cherry-pick to 0.9 too, then with this we have all branches covered.

hack/clean-e2e.sh
```diff
@@ -3,7 +3,6 @@
 REPO_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
 cd "${REPO_ROOT}" || exit 1
 
-minikube delete
```
tuminoid, line 6: Why don't we delete the kind cluster here anymore?

lentzi90: It is cleaned up [here](https://github.com/metal3-io/baremetal-operator/blob/f672c9a8c42607206f427350dcde636f762d1404/test/e2e/e2e_suite_test.go#L323-L331)

tuminoid (commented):
/lgtm

kashifest (commented):
/approve

PR #2278, opened 2025-03-05
:sparkles:  Add read only mode for local ironic run

Rozzii:
This PR:
 - Adds an option to the local ironic deployment script to run the containers with a read-only file system
 - Adds additional mounts for the Ironic containers regardless if a container runs with a read-only file system or not

This commit is needed because otherwise developers would be unable to run the local ironic deployment with read-only root file system.

Related to: https://github.com/metal3-io/ironic-image/issues/613
IRSO and BMO ironic-deployment overlays will need this config also.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,7 @@ go.work.sum
 *.tmp
 .DS_Store
 .zed*
+*.swp
 
 # Tilt files.
 .tiltbuild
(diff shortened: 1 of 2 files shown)


tuminoid (commented):
Not against cleaning the bash here, but can you do it in separate commit or in separate PR as it is making super hard to identify the actual change here.

Rozzii:
/remove-lifecycle stale
No no, at this point everything is in place for this to work.

tuminoid:
Should we abandon this due IRSO being the way to install Ironic now?

Rozzii:
I can move this deployment method to dev-env , if BMO will only host IrSO based deployments.
/remove-lifecycle stale

Rozzii:
/remove-lifecycle stale
This will most likely go to dev-env directly.

PR #2302, opened 2025-03-11
🌱 E2E: Fix kustomization for Ironic with inspector

lentzi90:

The overlays for the older releases had some issues with the patches and values after the switch to kind. This fixes that.

Fixes #

/hold
Using a temporary commit to run the optional tests.

### DIFF
--- a/config/overlays/e2e-release-0.4/ironic.env
+++ b/config/overlays/e2e-release-0.4/ironic.env
@@ -1,4 +1,4 @@
-DEPLOY_KERNEL_URL=http://192.168.222.199:6180/images/ironic-python-agent.kernel
-DEPLOY_RAMDISK_URL=http://192.168.222.199:6180/images/ironic-python-agent.initramfs
-IRONIC_ENDPOINT=https://192.168.222.199:6385/v1/
-IRONIC_INSPECTOR_ENDPOINT=https://192.168.222.199:5050/v1/
+DEPLOY_KERNEL_URL=http://192.168.222.1:6180/images/ironic-python-agent.kernel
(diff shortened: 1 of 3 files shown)


lentzi90:
Optional tests passed: https://github.com/metal3-io/baremetal-operator/actions/runs/13783341172?pr=2302
Removing test commit now.

config/overlays/e2e-release-0.4/ironic.env
tuminoid, line 1: 0.4 is out of testing too, by far, do we need to keep the file here at all?

lentzi90: That is a good point. We may need to reconsider what jobs to run on the release branches.
The thing is, we configure the workflow to run periodically on main, 0.9 and 0.8. But in the optional tests, we have all the upgrades also, so we are running the upgrade tests that exists on main, 0.9 and 0.8. The upgrade tests that we have on 0.8 are going all the way back to 0.4.

Probably we should run upgrade jobs only on main, and make sure we there test upgrades form all versions we want to test. This would mean disabling/removing the upgrade jobs (i.e. BMO e2e optional) for the release branches

tuminoid: Since you've made the fix already, let's merge this and then have another PR to do what you proposed and/or limiting the range of upgrades tested.

IMO we should only be testing upgrades starting from the last tested release (0.6), the maintained releases (0.8, 0.9) and main. Anything out of maintenance and testing is dead IMO.

tuminoid (commented):
/lgtm

kashifest (commented):
/approve

PR #2317, opened 2025-03-17
:bug: bump x/net, x/crypto and x/oauth2

tuminoid:
Bump some mods to get rid of some called/uncalled vulnerabilities.


Rozzii (commented):
/approve

PR #2321, opened 2025-03-17
:warning: Only accept HTTPHeadersRef in same namespace

tuminoid:
The BMCEventSubscription spec uses a Secret reference for the HTTPHeadersRef field. This allows the user to specify any namespace. To prevent leaking Secrets, this commit restricts the namespace to the same namespace that the BMCEventSubscription is in.


Rozzii (commented):
/approve

Rozzii:
It was agreed prior to the PR by security team.
/lgtm
So I lgtm and approve in one go.

PR #2323, opened 2025-03-17
:rocket: Release v0.9.1

tuminoid:
Release v0.9.1.


Rozzii (commented):
/approve
/lgtm
Adding both because of the special situation as it is related to CVE.

PR #2325, opened 2025-03-17
:bug: fix broken release process

tuminoid:
Release notes generation was more or less completely broken, and release workflow failed to create release notes.


tuminoid:
Release team
/cc @Rozzii @peppi-lotta 

PTAL if I missed any portion of the release process patches. At least the go parts now do the right thing, and workflow is fixed based on what I noticed, I did v0.8.1 and v0.9.1 releases on GH side manually as the workflow failed.

Rozzii (commented):
/approve

tuminoid:
> PTAL if I missed any portion of the release process patches. At least the go parts now do the right thing, and workflow is fixed based on what I noticed, I did v0.8.1 and v0.9.1 releases on GH side manually as the workflow failed.

/hold
To be sure it doesn't go in too early.

PR #2326, opened 2025-03-18
:seedling: group all github action bumps into single PR

tuminoid:
Group all GitHub Action bumps into single bump PR. We cannot really test them anyways, so the review process is more or less eyeballing the release notes and Github provided compatibility numbers.

In addition, having many action bump "pollutes" our release notes with user-irrelevant seedling PRs.


Rozzii (commented):
/approve

PR #2330, opened 2025-03-19
🌱 Reduce logging noise from status reconstruction

dtantsur:
Don't announce reconstructing the status when no information is present.
Reduce the priority of the message when no information is present.


tuminoid (commented):
/lgtm

PR #2331, opened 2025-03-20
:seedling: bump tj-actions/changed-files to v46.0.1

tuminoid:
Since the exploit code was added and then removed to this action, it looks like they moved all the v45 tags to point to same updated commit. This makes Dependabot confused. Let's bump it to latest release v46.0.1.


Rozzii (commented):
/approve

PR #2335, opened 2025-03-21
🌱 Bump golangci-lint to v1.64.7

kashifest:
This PR also bumps golangci-lint-action to 6.5.2. While doing that it removes some of the old unsupported golangci-lint settings.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,5 +1,5 @@
 run:
-  deadline: 10m
+  go: "1.23"
   build-tags:
   - e2e
   - vbmctl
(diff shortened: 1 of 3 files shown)


.golangci.yaml
```diff
@@ -1,5 +1,5 @@
 run:
-  deadline: 10m
+  go: "1.23"
```
tuminoid, line 2: Hmm, did we not remove these go version configs elsewhere?

kashifest: in other repos, it was here already, this was missing in bmo it seems.

tuminoid (commented):
/lgtm

tuminoid (commented):
/lgtm cancel

This has broken in CAPM3 and IRSO, let's fix those and take the fixes into this too

tuminoid:
Linter issue was not related to this PR, but golangci-lint issue.
/lgtm

PR #2336, opened 2025-03-21
:sparkles: Use standard GetVirtualMedia API for checking dataImage status

hroyrh:
Use the newly introduced [virtual media get api](https://github.com/gophercloud/gophercloud/pull/3303) to get the current dataImage status.

### DIFF
--- a/controllers/metal3.io/baremetalhost_controller.go
+++ b/controllers/metal3.io/baremetalhost_controller.go
@@ -1664,21 +1664,15 @@ func (r *BareMetalHostReconciler) handleDataImageActions(prov provisioner.Provis
 	// given constant ?
 	dataImageRetryBackoff := max(dataImageUpdateDelay, calculateBackoff(dataImage.Status.Error.Count))
 
-	// Check if any attach/detach action is pending or failed to attach
-	// We are assuming that the action will have completed by the time
-	// this reconcile is called ( after the delay specified in the previous
-	// action)
-	// TODO(hroyrh) : update this once vmedia.get api is available
-	isNodeBusy, nodeError := prov.IsDataImageReady()
-	if isNodeBusy {
-		info.log.Info("Node is busy, requeuing")
+	// Check if dataImage is attached to the node or not
+	// Given that this is a synchronous call to Ironic, should we add
+	// a longer wait ?
+	isImageAttached, getVmediaError := prov.GetDataImageStatus()
+	if getVmediaError != nil {
+		info.log.Error(getVmediaError, "Error fetching Virtual Media details")
 
-		// In case the last node error was not nil for dataimage,
-		// update message and counter
-		if nodeError != nil {
-			info.log.Info("DataImage action failed", "Error", nodeError.Error())
-
-			dataImage.Status.Error.Message = nodeError.Error()
+		if !errors.Is(getVmediaError, provisioner.ErrNodeIsBusy) {
+			dataImage.Status.Error.Message = getVmediaError.Error()
 			dataImage.Status.Error.Count++
 
 			if err := r.Status().Update(info.ctx, dataImage); err != nil {
@@ -1689,29 +1683,31 @@ func (r *BareMetalHostReconciler) handleDataImageActions(prov provisioner.Provis
 		return actionContinue{dataImageRetryBackoff}
 	}
 
-	// Is the current dataImage status valid
-	dirty := false
-
-	// In case the last node error was not nil for dataimage,
-	// update message and counter
-	if nodeError != nil {
-		info.log.Info("DataImage not ready", "Error", nodeError.Error())
-
-		dataImage.Status.Error.Message = nodeError.Error()
-		dataImage.Status.Error.Count++
-		dirty = true
-	}
-
 	deleteDataImage := !dataImage.DeletionTimestamp.IsZero()
 
 	requestedURL := dataImage.Spec.URL
 
+	// We will assume that the attached url is correct
+	// In case of failed attach, this variable will be overridden
 	attachedURL := dataImage.Status.AttachedImage.URL
+
+	// If there is no attached image, we will override the attachedURL
+	// so that further actions are handled accordingly
+	if !isImageAttached && attachedURL != "" {
+		dataImage.Status.AttachedImage.URL = ""
+
+		// Update DataImage Status
+		if err := r.Status().Update(info.ctx, dataImage); err != nil {
+			return actionError{fmt.Errorf("failed to update DataImage status, %w", err)}
+		}
+
+		return actionContinue{}
+	}
 
 	if deleteDataImage {
 		info.log.Info("DataImage requested for deletion")
-		if attachedURL != "" || dirty {
-			info.log.Info("Detaching DataImage as it was deleted")
+		if isImageAttached {
+			info.log.Info("Detaching DataImage as its deletion has been requested")
 			err := r.detachDataImage(prov, info, dataImage)
 			if err != nil {
 				return actionError{fmt.Errorf("failed to detach, %w", err)}
@@ -1726,7 +1722,7 @@ func (r *BareMetalHostReconciler) handleDataImageActions(prov provisioner.Provis
 		return nil
 	}
 
-	if requestedURL != attachedURL || dirty {
+	if requestedURL != attachedURL {
 		info.log.Info("DataImage change detected")
 		if attachedURL != "" {
 			info.log.Info("Detaching DataImage")
@@ -1754,21 +1750,15 @@ func (r *BareMetalHostReconciler) handleDataImageActions(prov provisioner.Provis
 		}
 	}
 
-	// Clear dataImage errors if nodeError is nil
-	if !dirty {
-		dataImage.Status.Error.Message = ""
-		dataImage.Status.Error.Count = 0
-	}
+	// Clear dataImage errors as relevant error cases were already handled
+	dataImage.Status.Error.Message = ""
+	dataImage.Status.Error.Count = 0
 
 	if err := r.Status().Update(info.ctx, dataImage); err != nil {
 		return actionError{fmt.Errorf("failed to update DataImage status, %w", err)}
 	}
 
-	info.log.Info("Updated DataImage Status after handling attachment/detachment")
-
-	if dirty {
-		return actionContinue{dataImageRetryBackoff}
-	}
+	info.log.Info("Updated DataImage Status triggered after handling attachment/detachment")
 
 	return nil
 }
@@ -1801,9 +1791,13 @@ func (r *BareMetalHostReconciler) attachDataImage(prov provisioner.Provisioner, 
 		return fmt.Errorf("failed to attach dataImage, %w", err)
 	}
 
-	// Update attached.URL here, we will mark it dirty in case any node errors
-	// are encountered
+	// Update the attached URL here assuming this operation will succeed
+	// In case of failure, this will be overridden in next reconcile
+	// We have to do this, as there is no other way to make sure the
+	// attached image url since the virtual media get api always returns
+	// the same standard url
 	dataImage.Status.AttachedImage.URL = dataImage.Spec.URL
+
 	// Error updating DataImage Status
 	if err := r.Status().Update(info.ctx, dataImage); err != nil {
 		return fmt.Errorf("failed to update DataImage status, %w", err)
@@ -1825,14 +1819,6 @@ func (r *BareMetalHostReconciler) detachDataImage(prov provisioner.Provisioner, 
 		}
 
 		return fmt.Errorf("failed to detach dataImage, %w", err)
-	}
-
-	// Update attached.URL here, we will mark it dirty in case any node errors
-	// are encountered
-	dataImage.Status.AttachedImage.URL = ""
-	// Error updating DataImage Status
-	if err := r.Status().Update(info.ctx, dataImage); err != nil {
-		return fmt.Errorf("failed to update DataImage status, %w", err)
 	}
 
 	return nil
--- a/controllers/metal3.io/dataimage_controller.go
+++ b/controllers/metal3.io/dataimage_controller.go
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"time"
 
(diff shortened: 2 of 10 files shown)


go.mod
```diff
@@ -5,7 +5,7 @@ go 1.23.7
 require (
 	github.com/go-logr/logr v1.4.2
 	github.com/google/safetext v0.0.0-20230106111101-7156a760e523
-	github.com/gophercloud/gophercloud/v2 v2.6.0
+	github.com/gophercloud/gophercloud/v2 v2.0.1-0.20250311123527-1031a8731f40
```
dtantsur, line 8: This is suspicious, are you switching to the v3 (master) branch? We need to stay on v2 for the time being. If your patch is not in a v2 release, ask the gophercloud team.

hroyrh: I don't think its the v3 master branch, I did `go get gophercloud/v2@commit` so it should be based on v2 ? Still, its a good idea to include it as part of a release, I will reach out to the gophercloud team.

iurygregory: We should probably change, they added to v2 4 days ago see  https://github.com/gophercloud/gophercloud/pull/3335

hroyrh: Done, thanks Iury!

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1664,54 +1664,48 @@ func (r *BareMetalHostReconciler) handleDataImageActions(prov provisioner.Provis
...
-
 	deleteDataImage := !dataImage.DeletionTimestamp.IsZero()
 
 	requestedURL := dataImage.Spec.URL
 
+	// We will assume that the attached url is correct
+	// In case of failed attach, this variable will be overridden
 	attachedURL := dataImage.Status.AttachedImage.URL
 
+	// If there is no attached image, we will override the atachedURL
```
elfosardo, line 1692: nit attachedURL

elfosardo (changes requested):
Thanks Himanshu, it looks good, just one thing, are the changes in go.mod and go.sum related to this change feature?
if not we should probably not update them

controllers/metal3.io/baremetalhost_controller.go
hroyrh: Done

iurygregory:
We need to update gophercloud, so the changes in go.mod and go.sum are valid.
/lgtm

tuminoid:
> We need to update gophercloud, so the changes in go.mod and go.sum are valid. /lgtm

2.7.0 is now out, which has been updated here, nice.

elfosardo:
/lgtm
thanks!

pkg/provisioner/ironic/ironic.go
```diff
@@ -1760,34 +1761,60 @@ func (p *ironicProvisioner) RemoveBMCEventSubscriptionForNode(subscription metal
...
-		return true, fmt.Errorf("ironic version=%d doesn't support DataImage API, needs version>=1.89", p.availableFeatures.MaxVersion)
+// Uses the Ironic Virtual Media Get API which synchronously fetches the
+// virtual media details for the given node.
+// We return only the bool isImageAttached because the url in the response
+// body is always the same - based on the node uuid, so it is not useful
+// for any comparison purpose.
+func (p *ironicProvisioner) GetDataImageStatus() (isImageAttached bool, err error) {
+	// Check if Ironic API version supports Virtual Media Get API
+	// Needs version >= 1.93
+	if !p.availableFeatures.HasVirtualMediaGetAPI() {
```
dtantsur, line 1772: This is okay for a rarely used feature. If it was something more core, we'd need to gracefully fall back to the previous approach.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1760,34 +1761,60 @@ func (p *ironicProvisioner) RemoveBMCEventSubscriptionForNode(subscription metal
...
 		return true, err
 	}
 
-	isNodeBusy = node.Reservation != ""
+	// If the node is busy, return error, since the get request will anyways
+	// fail
+	// TODO(hroyrh): is there a better way to do this ?
+	isNodeBusy := node.Reservation != ""
+	if isNodeBusy {
+		return false, fmt.Errorf("node is busy, retry GetVirtualMedia request in some time")
```
dtantsur, line 1786: Why do we need to treat this as an error now? Having reservation is normal while the request is being processed. This is going to populate Status.Error unnecessary.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1760,34 +1761,60 @@ func (p *ironicProvisioner) RemoveBMCEventSubscriptionForNode(subscription metal
...
+	}
+
+	err = nodes.GetVirtualMedia(p.ctx, p.client, p.nodeID).Err
+	if err != nil {
+		return false, err
+	}
+
+	body := nodes.GetVirtualMedia(p.ctx, p.client, p.nodeID).Body
```
dtantsur, line 1794: Please don't do the request twice, you can use the result of the previous request (and probably the Extract function to get body and error at the same time)

pkg/provisioner/ironic/ironic.go
```diff
@@ -1760,34 +1761,60 @@ func (p *ironicProvisioner) RemoveBMCEventSubscriptionForNode(subscription metal
...
 
-	// In case the error node encountered was related to something else
-	// TODO(hroyrh) : Is this check valid ?
-	if !strings.Contains(node.LastError, "attach") && !strings.Contains(node.LastError, "detach") {
-		return isNodeBusy, nil
+	data, err := json.Marshal(body)
```
dtantsur, line 1796: All this JSON logic must be handled by GopherCloud, not us. You need to use Extract.

dtantsur (changes requested):

controllers/metal3.io/dataimage_controller.go
```diff
@@ -187,26 +187,30 @@ func (r *DataImageReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
-	// Is the current dataImage status valid
-	dirty := false
-	// In case the last node error was not nil for dataimage,
+	// In case there was an error fetching vmedia details
 	// upadate message and counter
-	if nodeError != nil {
-		di.Status.Error.Message = nodeError.Error()
+	if vmediaGetError != nil {
+		di.Status.Error.Message = vmediaGetError.Error()
 		di.Status.Error.Count++
```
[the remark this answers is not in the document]
hroyrh: But both of them have a get vmedia request before, so an error might come up in either or both case.

pkg/provisioner/ironic/ironic.go
hroyrh: Do you mean, in case of api version >= 1.89 and < 1.93, we should fall back to "IsDataImageReady" logic ?

pkg/provisioner/ironic/ironic.go
hroyrh: We don't have an extract function for virtualmedia in gophercloud yet, but I will get the result into a single call.

pkg/provisioner/ironic/ironic.go
hroyrh: Same as above, can we do it once I have added the extract function in gophercloud ?

pkg/provisioner/ironic/ironic.go
hroyrh: Yes, but in case of nodeBusy, we won't know the actual status of virtual media so returning error here to retry. I will see if we can do it without updating the Status.Error .

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1664,54 +1664,48 @@ func (r *BareMetalHostReconciler) handleDataImageActions(prov provisioner.Provis
...
-
-			if err := r.Status().Update(info.ctx, dataImage); err != nil {
-				return actionError{fmt.Errorf("failed to update DataImage status, %w", err)}
-			}
+	// Check if dataImage is attached to the node or not
+	// Given that this is a synchronous call to Ironic, should we add
+	// a longer wait ?
+	isImageAttached, getVmediaError := prov.GetDataImageStatus()
+	if getVmediaError != nil {
+		info.log.Info("Error fetching Virtual Media details", "Error", getVmediaError)
```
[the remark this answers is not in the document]
hroyrh: Done

pkg/provisioner/ironic/ironic.go
hroyrh: I have made changes to check for nodeBusy error and not update the Status in that scenario.

pkg/provisioner/ironic/ironic.go
iurygregory: Are we ok with going with json.Marshal while we add the Extract to gophercloud?

controllers/metal3.io/host_state_machine_test.go
```diff
@@ -1403,7 +1403,7 @@ func (p *mockProvisioner) GetFirmwareComponents() (components []metal3api.Firmwa
 	return components, nil
 }
 
-func (m *mockProvisioner) IsDataImageReady() (isNodeBusy bool, nodeError error) {
+func (p *mockProvisioner) GetDataImageStatus() (isImageAttached bool, err error) {
```
iurygregory, line 1406: I'm wondering if renaming is ok here, since it was called IsDataImageReady before..

hroyrh: I think the two functions served similar but distinct roles. "IsDataImageReady" was meant to capture node errors and guess if the virtual media attach/detach was successful, while "GetDataImageStatus" gets the actual current status for virtual media.

controllers/metal3.io/dataimage_controller.go
```diff
@@ -187,26 +187,34 @@ func (r *DataImageReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+	// In case there was an error fetching vmedia details
 	// upadate message and counter
-	if nodeError != nil {
-		di.Status.Error.Message = nodeError.Error()
-		di.Status.Error.Count++
-		dirty = true
+	if vmediaGetError != nil {
+		reqLogger.Error(vmediaGetError, "failed to fetch Virtual Media details")
+
+		if vmediaGetError.Error() != "node is busy" {
```
dtantsur, line 197: Same here

controllers/metal3.io/host_state_machine_test.go
dtantsur: Provisioner API is not public, at least for now. Renaming is fine.

pkg/provisioner/ironic/ironic.go
dtantsur: In theory. In practice, this feature is rarely used, and in OpenShift we couple BMO and Ironic versions, so I don't suggest we complicated the code further.

pkg/provisioner/ironic/ironic.go
dtantsur: Let's at least leave a FIXME

dtantsur: Also... why are we doing a Marschal followed by Unmarschall? What is Body even?

pkg/provisioner/ironic/ironic.go
```diff
@@ -1760,34 +1761,62 @@ func (p *ironicProvisioner) RemoveBMCEventSubscriptionForNode(subscription metal
...
+	if err != nil {
+		p.log.Info("GetDataImage, failed to marshal response body", "Error", err)
+		return false, err
+	}
+
+	var jsonData []map[string]interface{}
```
dtantsur, line 1804: Let's create a list of structures, not maps. We know how the response is supposed to look.

pkg/provisioner/ironic/ironic.go
iurygregory: result := nodes.GetVirtualMedia(p.ctx, p.client, p.nodeID)

returns the following:

{{[map[image: inserted:false media_types:[Floppy USBStick]] map[image:http://<address>/redfish/boot-12345678-aaaa-5856-bfb6-6aaaacd3dd11.iso inserted:true media_types:[CD DVD]]] 0 map[Content-Length:[212] Content-Type:[application/json] Date:[Fri, 28 Feb 2025 01:26:50 GMT] Openstack-Request-Id:[req-...] X-Openstack-Ironic-Api-Maximum-Version:[1.95] X-Openstack-Ironic-Api-Minimum-Version:[1.1] X-Openstack-Ironic-Api-Version:[1.93]] <nil>}}

So yeah we probably don't need the Unmarshall, seem unnecessary, we are already able to access Body and Err, so it goes down having the proper structs to handle the information.

tuminoid:
Are we planning to take this in 0.10 or can we leave it after branching?

pkg/provisioner/ironic/ironic.go
hroyrh: I've used VirtualMedia struct to capture the body data, will submit a patch to gophercloud to move it there.

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1664,21 +1664,15 @@ func (r *BareMetalHostReconciler) handleDataImageActions(prov provisioner.Provis
...
-
-			dataImage.Status.Error.Message = nodeError.Error()
+	// Check if dataImage is attached to the node or not
+	// Given that this is a synchronous call to Ironic, should we add
+	// a longer wait ?
+	isImageAttached, getVmediaError := prov.GetDataImageStatus()
+	if getVmediaError != nil {
+		info.log.Error(getVmediaError, "Error fetching Virtual Media details")
+
+		if getVmediaError.Error() != "node is busy" {
```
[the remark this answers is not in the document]
hroyrh: I have created a const NodeIsBusy for this error, did you mean something else ?

pkg/provisioner/ironic/ironic.go
```diff
@@ -1760,34 +1761,62 @@ func (p *ironicProvisioner) RemoveBMCEventSubscriptionForNode(subscription metal
...
 	}
 
-	isNodeBusy = node.Reservation != ""
+	// If the node is busy, return error, since the get request will anyways
+	// fail
+	// TODO(hroyrh): is there a better way to do this ?
+	isNodeBusy := node.Reservation != ""
+	if isNodeBusy {
+		p.log.Info("node is busy, retry GetVirtualMedia request in some time", "nodeid", node.UUID)
+		return false, fmt.Errorf("node is busy")
```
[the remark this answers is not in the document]
hroyrh: I have created a const NodeIsBusy for this error, did you mean something else ?

controllers/metal3.io/baremetalhost_controller.go
```diff
@@ -1689,29 +1683,31 @@ func (r *BareMetalHostReconciler) handleDataImageActions(prov provisioner.Provis
...
+	// so that further actions are handled accordingly
+	if !isImageAttached && attachedURL != "" {
+		dataImage.Status.AttachedImage.URL = ""
+
+		// Update DataImage Status
+		if err := r.Status().Update(info.ctx, dataImage); err != nil {
+			return actionError{fmt.Errorf("failed to update DataImage status, %w", err)}
+		}
+
+		return actionContinue{dataImageRetryBackoff}
```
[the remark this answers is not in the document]
hroyrh: Makes sense, I have removed the delay.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1760,34 +1761,62 @@ func (p *ironicProvisioner) RemoveBMCEventSubscriptionForNode(subscription metal
...
 		return true, err
 	}
 
-	isNodeBusy = node.Reservation != ""
+	// If the node is busy, return error, since the get request will anyways
+	// fail
+	// TODO(hroyrh): is there a better way to do this ?
+	isNodeBusy := node.Reservation != ""
+	if isNodeBusy {
+		p.log.Info("node is busy, retry GetVirtualMedia request in some time", "nodeid", node.UUID)
```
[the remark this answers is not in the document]
hroyrh: Done.

pkg/provisioner/ironic/ironic.go
hroyrh: I've created a struct for VirtualMedia, will move it to gophercloud next.

dtantsur:
> Are we planning to take this in 0.10 or can we leave it after branching?

This feature is not on the default path, it shouldn't be a large regression risk if that's what worries you.

pkg/provisioner/ironic/ironic.go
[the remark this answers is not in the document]
dtantsur: Not really. The Go way of checking errors is `errors.Is`/`errors.As`, not comparing error messages. If you need prior art, ErrNeedsPreprovisioningImage is handled this way in the controller.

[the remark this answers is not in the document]
hroyrh: Fixed

dtantsur (approved):

elfosardo (approved):

Issue #2072, opened 2024-11-21
Increase number of enabled golangci-lint linters 


tuminoid:
BMO has 40+ linters enabled in our config whitelist. Golangci-lint has 100+ linters. Enable more linters to increase coverage and code quality.

/kind ci

Rozzii:
Let's propose new linters under a github discussion.
/triage accepted

tuminoid:
/unassign @sandeepkunusoth 
/assign @tuminoid 

Please work on sub-tasks, this is just coordination parent. There is one left.

tuminoid:
All linters done. Thank you all who contributed!
/close

PR #2350, opened 2025-03-26
:seedling: update the used linters list

tuminoid:
We have [agreed a list of linters](https://github.com/orgs/metal3-io/discussions/2098) to be enabled in all of our Go repos. Sync that list to BMO and leave disabled those that do not pass.

We could enable 12 more linters without any failures. We have 14 linters commented out.

I will make sub-issues of https://github.com/metal3-io/baremetal-operator/issues/2072 for the remaining items.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -12,13 +12,24 @@ linters:
   - bodyclose
   #- containedctx
   - copyloopvar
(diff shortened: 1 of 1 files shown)


lentzi90 (commented):
/lgtm

PR #2374, opened 2025-04-02
:seedling: bump golang to v1.23.8 and x/net to v0.38.0

tuminoid:
Bump golang to v1.23.8 due CVE-2025-22871.


build/Dockerfile
lentzi90, line 1: What happened here? Did git just take the link and made a hard copy of the file?
/hold

tuminoid: Good catch. Apparently if you use `sed -i` over symlink, it will make in-place edit with hard copy.

Fixed.
/unhold

lentzi90 (commented):
/approve

PR #2381, opened 2025-04-04
:bug: Fix few errors when making beta release notes

peppi-lotta:
**What this PR does / why we need it**:  When running make release-notes for v0.10.0-beta.0 I encountered a few problems. In addition to "Merge pull request" pattern there was a "Merge commit from fork" pattern. I added this as an if so that notes.go doesn't break if this patter occurs. 

I also changed the order of warning printing so that it matches the order in previous pre-releases.

### DIFF
--- a/hack/tools/release/notes.go
+++ b/hack/tools/release/notes.go
@@ -213,9 +213,18 @@ func run() int {
 		}
 		body = fmt.Sprintf("- %s", body)
 
-		_, err := fmt.Sscanf(c.merge, "Merge pull request %s from %s", &prNumber, &fork)
-		if err != nil {
-			log.Fatalf("Error parsing merge commit message: %v", err)
+		if strings.HasPrefix(c.merge, "Merge pull request") {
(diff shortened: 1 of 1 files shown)


tuminoid:
> **What this PR does / why we need it**: When running make release-notes for v0.10.0-beta.0 I encountered a few problems. In addition to "Merge pull request" pattern there was a "Merge commit from fork" pattern. I added this as an if so that notes.go doesn't break if this patter occurs.

As discussed offline, this is due merging the security advisory https://github.com/metal3-io/baremetal-operator/security/advisories/GHSA-c98h-7hp9-v9hq as it is no ordinary PR, but directly merging the fix from a private fork. We merged that on main, but did not cut release from main, so we did not hit the issue back then. On release-0.9 and release-0.8 we did a cherry-pick PR instead, which is again a PR that it could handle.

tuminoid (commented):
/lgtm

Tested, and this works for cutting BMO 0.10.0-beta.0. Handles security advisory and also moves the beta notice in proper place. Thanks @peppi-lotta !

lentzi90 (commented):
/approve

PR #2382, opened 2025-04-04
🌱 Enable dupl linter in golangci-lint

andriisoldatenko:

fixes https://github.com/metal3-io/baremetal-operator/issues/2351

> Note: IMO it make sense to disable this linter for test files, cause a little bit duplication in tests is acceptable.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -14,7 +14,7 @@ linters:
   - copyloopvar
   - decorder
   - dogsled
-  #- dupl
+  - dupl
   - dupword
   - durationcheck
   - errcheck
@@ -142,6 +142,7 @@ issues:
     - gci
     - goconst
     - gosec
+    - dupl
   - path: _test\.go
     linters:
     - errcheck


tuminoid:
> > **_NOTE:_** IMO it make sense to disable this linter for test files, cause a little bit duplication in tests is acceptable.

Depending on duplication. We had this discussion earlier in IRSO when fixing `dupl` there, and it ended up making tests much more readable and increased code reusability.

Can you give specific examples what you think is acceptable duplication?

andriisoldatenko:
> > > **_NOTE:_** IMO it make sense to disable this linter for test files, cause a little bit duplication in tests is acceptable.
> 
> Depending on duplication. We had this discussion earlier in IRSO when fixing `dupl` there, and it ended up making tests much more readable and increased code reusability.
> 
> Can you give specific examples what you think is acceptable duplication?

yeah you are absolutely right, q is do you want to check tests as well and fix current duplication?
these https://github.com/metal3-io/baremetal-operator/blob/2e96d2a796f8014224a6bf0d72679d4c4cff3119/controllers/metal3.io/host_state_machine_test.go#L141-L165

and these 

https://github.com/metal3-io/baremetal-operator/blob/2e96d2a796f8014224a6bf0d72679d4c4cff3119/controllers/metal3.io/host_state_machine_test.go#L188-L212

tuminoid:
In my opinion that should be a parameterized test function. I think there is like 2 line diff between the snippets?

`dupl` might product huge number of complaints, so IMO it is perfectly fine to split fixing it to multiple PRs, to keep refactoring per PR at readable/understandable level, if that helps.

andriisoldatenko:
already done previously - in 190644b9

PR #2383, opened 2025-04-04
🌱 Enable testifylint linter in golangci-lint

andriisoldatenko:

resolves https://github.com/metal3-io/baremetal-operator/issues/2363

### DIFF
--- a/apis/metal3.io/v1alpha1/hostfirmwarecomponents_types_test.go
+++ b/apis/metal3.io/v1alpha1/hostfirmwarecomponents_types_test.go
@@ -85,7 +85,7 @@ func TestValidateHostFirmwareComponents(t *testing.T) {
 
 			err := hostFirmwareComponents.ValidateHostFirmwareComponents()
 			if err == nil {
-				assert.Equal(t, tc.ExpectedError, "")
+				assert.Equal(t, "", tc.ExpectedError)
 			} else {
 				assert.Equal(t, tc.ExpectedError, err.Error())
 			}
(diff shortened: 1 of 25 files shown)


controllers/metal3.io/baremetalhost_controller_test.go
```diff
@@ -539,34 +539,34 @@ func TestDoNotAddSecretFinalizersDuringDelete(t *testing.T) {
 	// Let the host reach the available state before deleting it
 	waitForProvisioningState(t, r, host, metal3api.StateAvailable)
 	err := doDeleteHost(host, r)
-	assert.NoError(t, err)
+	require.NoError(t, err)
```
adilGhaffarDev, line 542: do we have to change all `asserts` to `require` ? is it something that Linter is forcing?

tuminoid: It is not changing all of them, just the ones `NoError`. `NoError` is expected to be `fatal` failure that will ruin the rest of test run anyways, if it the assertion fails, so require just calls it a day earlier?

apis/metal3.io/v1alpha1/hostfirmwarecomponents_types_test.go
```diff
@@ -85,7 +85,7 @@ func TestValidateHostFirmwareComponents(t *testing.T) {
 
 			err := hostFirmwareComponents.ValidateHostFirmwareComponents()
 			if err == nil {
-				assert.Equal(t, tc.ExpectedError, "")
+				assert.Equal(t, "", tc.ExpectedError)
```
dtantsur, line 88: Ugh, this is a weird code (not your fault, just pointing out)

dtantsur:
/approve

But please see one comment on the import reordering

PR #2389, opened 2025-04-07
:seedling: make osv-scanner scan run daily

tuminoid:
Now that osv-scanner v2 is configured to announce any findings in Slack, we should be running it daily as weekly is too sparse and we usually fix issues sooner than the weekly job notices them.


lentzi90 (commented):
/approve

PR #2405, opened 2025-04-14
🌱 Refactoring: move ./controllers to ./internal/controller

dtantsur:
Per the recent KubeBuilder format, controllers go under internal
together with webhooks (and the directory uses singular).


Rozzii (commented):
/lgtm

zaneb:
We have a controller that exists largely for the purpose of people using it as a library: https://github.com/metal3-io/baremetal-operator/blob/main/internal/controller/metal3.io/preprovisioningimage_controller.go
Doesn't moving it under `internal` prevent them ([us](https://github.com/openshift/image-customization-controller/blob/main/cmd/controller/main.go#L38)) from doing that?

Does anything else keep up with changes in kubebuilder? My experience of it was that it was effectively a one-time thing.

PR #2415, opened 2025-04-23
:bug: Make firmware update handle only a subset of `Spec.Updates`

winiciusallan:
When an operator wants to update bare metal firmware, it must pass to `HostFirmwareComponents.Spec` a list of components and URIs for the update. In the old behavior, the controller will try to update the whole list, but it is desired to update only the firmware that was not updated. This PR fixes this passing to `TargetComponents` only the subset of unupdated components.

### DIFF
--- a/internal/controller/metal3.io/baremetalhost_controller.go
+++ b/internal/controller/metal3.io/baremetalhost_controller.go
@@ -1133,6 +1133,22 @@ func (r *BareMetalHostReconciler) matchProfile(info *reconcileInfo) (dirty bool,
 	return
 }
 
+func getUpdatesDifference(specUpdates []metal3api.FirmwareUpdate, statusUpdates []metal3api.FirmwareUpdate) []metal3api.FirmwareUpdate {
+	diff := []metal3api.FirmwareUpdate{}
+	// Mapping already updated components
+	updated := make(map[string]string, len(statusUpdates))
+	for _, s := range statusUpdates {
+		updated[s.Component] = s.URL
+	}
+
+	for _, firmware := range specUpdates {
+		if _, ok := updated[firmware.Component]; !ok || firmware.URL != updated[firmware.Component] {
+			diff = append(diff, firmware)
+		}
+	}
+	return diff
+}
+
 func (r *BareMetalHostReconciler) actionPreparing(prov provisioner.Provisioner, info *reconcileInfo) actionResult {
 	info.log.Info("preparing")
 
@@ -1178,7 +1194,12 @@ func (r *BareMetalHostReconciler) actionPreparing(prov provisioner.Provisioner, 
 		return actionContinue{subResourceNotReadyRetryDelay}
 	}
 	if hfcDirty {
-		prepareData.TargetFirmwareComponents = hfc.Spec.Updates
+		// Handle only Firmware Component that it is in hfc.Spec.Updates but not in hfc.Status.Updates.
+		if hfc.Status.Updates != nil {
+			prepareData.TargetFirmwareComponents = getUpdatesDifference(hfc.Spec.Updates, hfc.Status.Updates)
+		} else {
+			prepareData.TargetFirmwareComponents = hfc.Spec.Updates
+		}
 	}
 
 	provResult, started, err := prov.Prepare(prepareData, bmhDirty || hfsDirty || hfcDirty,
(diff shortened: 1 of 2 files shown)


internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -1178,7 +1178,22 @@ func (r *BareMetalHostReconciler) actionPreparing(prov provisioner.Provisioner,
...
+				updated[s.Component] = s.URL
+			}
+
+			for _, firmware := range hfc.Spec.Updates {
+				if _, ok := updated[firmware.Component]; !ok || firmware.URL != updated[firmware.Component] {
+					prepareData.TargetFirmwareComponents = append(prepareData.TargetFirmwareComponents, firmware)
+				}
+			}
+		} else {
+			prepareData.TargetFirmwareComponents = hfc.Status.Updates
```
dtantsur, line 1195: This is just setting to `nil`, right? Please be explicit.

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -1178,7 +1178,22 @@ func (r *BareMetalHostReconciler) actionPreparing(prov provisioner.Provisioner,
 		return actionContinue{subResourceNotReadyRetryDelay}
 	}
 	if hfcDirty {
-		prepareData.TargetFirmwareComponents = hfc.Spec.Updates
+		// Handle only Firmware Component that it is in hfc.Spec.Updates but not in hfc.Status.Updates.
+		if hfc.Status.Updates != nil {
+			// Mapping already updated components
+			updated := make(map[string]string)
```
dtantsur, line 1184: nit: `make(map[string]string, len(hfc.Status.Updates))`

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -1426,7 +1441,22 @@ func (r *BareMetalHostReconciler) doServiceIfNeeded(prov provisioner.Provisioner
 			return actionError{fmt.Errorf("could not determine firmware components: %w", err)}
 		}
 		if hfcDirty {
-			servicingData.TargetFirmwareComponents = hfc.Spec.Updates
+			// Handle only Firmware Component that it is in hfc.Spec.Updates but not in hfc.Status.Updates.
+			if hfc.Status.Updates != nil {
```
dtantsur, line 1445: Could you extract this code to a helper function to avoid repeating it twice? Bonus: you can unit-test it then.

internal/controller/metal3.io/baremetalhost_controller.go
winiciusallan: Done

internal/controller/metal3.io/baremetalhost_controller.go
winiciusallan: Done

internal/controller/metal3.io/baremetalhost_controller.go
winiciusallan: For sure. Unit test added too.

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -1178,7 +1194,12 @@ func (r *BareMetalHostReconciler) actionPreparing(prov provisioner.Provisioner,
 		return actionContinue{subResourceNotReadyRetryDelay}
 	}
 	if hfcDirty {
-		prepareData.TargetFirmwareComponents = hfc.Spec.Updates
+		// Handle only Firmware Component that it is in hfc.Spec.Updates but not in hfc.Status.Updates.
+		if hfc.Status.Updates != nil {
+			prepareData.TargetFirmwareComponents = getUpdatesDifference(hfc.Spec.Updates, hfc.Status.Updates)
+		} else {
+			prepareData.TargetFirmwareComponents = nil
```
iurygregory, line 1201: So, here I don't think we should set to `nil`
If someone manually creates the HFC with `Spec.Updates` set, the status is not yet populated so `Status.Updates` is `nil` , the `prepareData.TargetFirmwareComponents` is what we will be sending to ironic to execute, if we say it's `nil` we won't be executing any firmware update, that is why before we had always set
`prepareData.TargetFirmwareComponents = hfc.Spec.Updates` so I think this should be in the else part

Does it makes sense to you @dtantsur ? I also think the same should apply to the servicing code below.

iurygregory (changes requested):

internal/controller/metal3.io/baremetalhost_controller.go
dtantsur: You're right. Initially, this was an assignment to `hfc.Status.Updates`, which is known to be nil here, so I simply ran with it. But now I understand that this code should probably assign to `hfc.Spec.Updates` instead.

winiciusallan: Done

iurygregory (commented):
/lgtm

iurygregory (approved):

PR #2426, opened 2025-04-28
:rocket: Release v0.10.1

peppi-lotta:
**What this PR does / why we need it**: Release v0.10.1 (where Ironic is pinned)

### DIFF
--- a/releasenotes/v0.10.1.md
+++ b/releasenotes/v0.10.1.md
@@ -1,0 +1,10 @@
+<!-- markdownlint-disable no-inline-html line-length -->
+# Changes since v0.10.0
+
+## :bug: Bug Fixes
+
+- Pin ironic to version 29.0 (#2419)
+
+The image for this release is: v0.10.1
+
+_Thanks to all our contributors!_ 😊


lentzi90 (commented):
/approve

tuminoid (commented):
Do we really need this release? We don't release artifacts for BMO and were removing the BMO commit pinning in dev-env and will point to release branch where the fix is present?

lentzi90:
Probably not so important, but if there are users who use tagged commits only (instead of the release branch) then this would still be useful I think

releasenotes/v0.10.1.md
```diff
@@ -0,0 +1,10 @@
+<!-- markdownlint-disable no-inline-html line-length -->
+# Changes since v0.10.0
+
+## :seedling: Others
+
+- Pin ironic to version 29.0 (#2419)
```
tuminoid, line 6: Let's make it a bug, if we release a patch version for it, per release policy.

tuminoid (commented):
/lgtm

PR #2436, opened 2025-05-02
:seedling: Enable tagliatelle linter

### LINKED ISSUE #2362: Enable "tagliatelle" linter in golangci-lint
Enable "tagliatelle" linter in golangci-lint. Fix any warnings enabling it causes.

/area ci
/help
/triage accepted

erjavaskivuori:
**What this PR does / why we need it**: Enabled tagliatelle linter

Fixes #2362

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -64,7 +64,7 @@ linters:
   - rowserrcheck
   - staticcheck
   - stylecheck
-  #- tagliatelle
+  - tagliatelle
   - testifylint
   - thelper
   - tparallel
@@ -124,6 +124,10 @@ linters-settings:
     - unnecessaryDefer
     - whyNoLint
     - wrapperFunc
+  tagliatelle:
+    case:
+      rules:
+        json: goCamel
 issues:
   exclude-dirs:
   - mock*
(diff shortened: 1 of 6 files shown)


lentzi90 (commented):
This is great work and shows that the API is a bit inconsistent.
Unfortunately, it is very much a breaking change. I don't think we can do this even in a minor release. We need a new API version completely.
/hold

erjavaskivuori:
> This is great work and shows that the API is a bit inconsistent. Unfortunately, it is very much a breaking change. I don't think we can do this even in a minor release. We need a new API version completely. /hold

Thanks @lentzi90! We discussed about this with @kashifest and anticipated that this might be the case.

tuminoid (commented):
Yup, this is probably something we want to rather just nolint the existing API variable names until we publish new API version. We don't want to have more of them added, so I'd go for nolint + enabling linter. WDYT @kashifest , @lentzi90 ?

lentzi90:
I agree. Enabling and adding nolint is good so we avoid adding more. We could also get started on an issue for next API version where we can list all the things we want to fix/change

tuminoid:
It is still running ipmi tests, so rebase is needed. @erjavaskivuori

PR #2437, opened 2025-05-03
[WIP]🐛 Fixing DisablePowerOff backward compatibility

adilGhaffarDev:
/hold
testing
This fix is making sure that BMO works with ironic where DisablePowerOff is not implemented.
Without the fix we get following error:
`          "faultstring": "Schema error for node: Additional properties are not allowed ('disable_power_off' was unexpected)",`

Fixes #


adilGhaffarDev:
tested it locally with IRONIC_TAG="v26.0.1" it works fine.
Testing in PR here: https://github.com/metal3-io/metal3-dev-env/pull/1527

PR #2448, opened 2025-05-06
:seedling: Add e2e tests for release 0.10

peppi-lotta:
**What this PR does / why we need it**: Add e2e tests for release 0.10


lentzi90 (commented):
Should we remove some of the oldest tests at the same time or are we doing it separately? I started working on something like that in https://github.com/metal3-io/baremetal-operator/pull/2434 to check the BMO <-> Ironic version compatibility

peppi-lotta:
> Should we remove some of the oldest tests at the same time or are we doing it separately? I started working on something like that in #2434 to check the BMO <-> Ironic version compatibility

I was thinking of removing the old tests in a separate PR

test/e2e/config/ironic.yaml
```diff
@@ -106,24 +108,31 @@ bmoIronicUpgradeSpecs:
...
-# Upgrade Ironic 24.1 -> latest | BMO 0.9
+# Upgrade BMO 0.10 -> main
 - deployIronic: true
   deployBMO: true
-  initBMOKustomization: "../../config/overlays/e2e-release-0.9"
+  initBMOKustomization: "../../config/overlays/e2e-release-0.10"
+  initIronicKustomization: "../../ironic-deployment/overlays/e2e"
+  upgradeEntityKustomization: "../../config/overlays/e2e"
+  upgradeEntityName: "bmo"
+# Upgrade Ironic 24.1 -> latest | BMO 0.10
```
lentzi90, line 118: This will fail because BMO 0.10 only works with Ironic 29 or newer.
Could you make it so that we keep BMO at 0.9 and do Ironic 24.1 -> 29.0?
Same for the rest, except the last one where we can do BMO at 0.10 and Ironic 29 -> latest

lentzi90 (commented):
We need to decide what to do with the upgrade tests here since BMO 0.10 won't work with Ironic <29.
@adilGhaffarDev what do you think? Should we stick to BMO 0.9 as I suggested below?

lentzi90 (commented):
/approve

.github/workflows/e2e-test-periodic-release0.10.yml
tuminoid, line 1: File name is incorrect, should be ending `-release-0.10.yml`

PR #2467, opened 2025-05-20
remove optional e2e badge link

tuminoid:
We don't run optional e2e on release branches, hence there is no badge and the link is broken.


tuminoid:
0.8 is tested only, so let's not actually fix this there.
/close

Issue #2465, opened 2025-05-19
Confusing behavior on removing `externallyProvisioned: true`


dtantsur:
I'm not sure we ever intended for the `externallyProvisioned` field to be reset to `false`, but right now it explicitly causes inspection to start. Unfortunately, it cannot possibly succeed with the Ironic node in the `active` state:

```
action \"inspecting\" failed: hardware inspection failed: unexpected provision state active","errorVerbose":"unexpected provision state active
```

If inspection is disabled, the next state is `preparing`, which I suspect will fail the same way for the same reason.

We need to decide on a sane behavior in this case and probably add the scenario to e2e tests.

/kind bug

dtantsur:
On thinking more about it, I think removing `externallyProvisioned` should result in a normal `provisioned` state, which then can be re-inspected or whatever.

Rozzii:
We have discussed this during community meeting, there are multiple options:
Let's just document this behavior and if someone is interested in the community then feel free to fix this.
/triage accepted
/help

zaneb:
Hmmm, yes we adopt the node in externallyProvisioned state, so it shows as active. When somebody removes the flag, we need to delete (i.e. deprovision) in ironic.
I don't think going to the Provisioned state is the right thing though, because there's no way to get back to inspecting from there if no inspection was ever done on the host (and inspection is not disabled).

dtantsur:
> I don't think going to the Provisioned state is the right thing though, because there's no way to get back to inspecting from there if no inspection was ever done on the host (and inspection is not disabled).

Then we need to enforce (with the webhook) that the user also removes image and customDeploy. Otherwise, they may expect the end result to be Provisioned.

... they may still expect it, I guess, but that's on them, right?

zaneb:
> the user also removes image and customDeploy. Otherwise, they may expect the end result to be Provisioned.

To be fair, the _end_ result _will_ be Provisioned, but the host will go through inspection/cleaning to get there, which may very much not be what they expected 😄. Now that you mention it, I feel like we must have discussed this ExternallyProvisioned->Provisioned thing at some point in the past.

[...]

PR #2472, opened 2025-05-20
⚠️ Prevent changing ExternallyProvisioned in the webhook

dtantsur:
Changing ExternallyProvisioned does not actually work. Furthermore, it
results in a host hopelessly stuck in the "inspecting" state (in my
testing, I could not even delete it). Until we sort out this issue, let
us prevent users from getting into a problem.

Workaround for: #2465
Includes:
- https://github.com/metal3-io/baremetal-operator/pull/2471


tuminoid (commented):
/cc @lentzi90 @Rozzii

lentzi90 (commented):
Sounds reasonable to me
/approve

PR #2454, opened 2025-05-12
🌱 E2E: Temporary workaround for connection issues

lentzi90:
**What this PR does / why we need it**:

IPA is trying to use the wrong URL when we do not set IRONIC_BASE_URL.
Configure iptables to allow connections between kind and VMs.
This makes redfish-virtualmedia green again.

Disable IPMI tests on PRs to unblock.
We still need a proper solution for the connection issues.

This also cleans up some minor things:
- Include go in the path before running the ensure script
- Remove download of unused ISO file

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Related to but does not properly fix https://github.com/metal3-io/baremetal-operator/issues/2456

PR #2476, opened 2025-05-22
🌱 E2E: Override IRONIC_HTTP_URL

### LINKED ISSUE #2456: All E2E tests are red
### Which jobs are failing?

All periodic E2E jobs
- https://github.com/metal3-io/baremetal-operator/actions/workflows/e2e-test-periodic-main.yml
- https://github.com/metal3-io/baremetal-operator/actions/workflows/e2e-test-optional-periodic.yml
- https://github.com/metal3-io/baremetal-operator/actions/workflows/e2e-test-periodic-release-0.10.yml
- https://github.com/metal3-io/baremetal-operator/actions/workflows/e2e-test-periodic-release-0.9.yml
- https://github.com/metal3-io/baremetal-operator/actions/workflows/e2e-test-periodic-release-0.8.yml

### Which tests are failing?

Practically all (except really basic things like BMH without BMC credentials -> unmanaged). Anything involving inspection or provisioning fails. Even external inspection.

### Since when has it been failing?

The e2e workflows have been 🔴 since May 9th.

### Jenkins link

_No response_

There is a communication issue between Ironic and IPA.
I have gotten inspection to work in https://github.com/metal3-io/baremetal-operator/pull/2454 by setting the IRONIC_BASE_URL variable, but provisioning still fails. There is something wrong with (the configuration of?) Ironic or IPA or how the agent token is handled.
Ironic logs:

```
[   13.334467] ironic-python-agent[480]: 2025-05-12 07:46:14.439 480 WARNING ironic_python_agent.agent [-] The agent token has already been retrieved. IPA may not operate as intended and the deployment may fail depending on settings in the ironic deployment.
```

There are no suspicious commits in BMO around the time (May 9th) when this started happening. There are no commits in ironic-image around that time (it was last built on May 6th).

### Label(s) to be applied

/kind failing-test
One or more /area label. See https://github.com/metal3-io/baremetal-operator/labels for the list of labels.

lentzi90:

This fixes PXE booting when Ironic cannot be reach directly from the BMHs. Ironic runs in the 172.18.0.x network (inside kind), but the relevant ports are exposed on the host. This way the VMs can reach Ironic at 192.168.222.1.

This commit also updates the e2e README.md, which still had the minikube configuration.

Fixes #2456


lentzi90:
Alright I think I have a working config. We need to specify `IRONIC_HTTP_URL` so that the ipxe config ends up with the correct IP (192.168.222.1). Otherwise, it will try to use the IP of the kind node (172.18.0.x). This is why virtualmedia worked but not ipmi.

I tried using IRONIC_EXTERNAL_IP first. That did not have any effect on the ipxe config. I am also still confused about why `IRONIC_BASE_URL` is needed. The *effect* it has is that it makes IPA use the correct URL to send back inspection data. But I thought that was the whole point of `IRONIC_EXTERNAL_CALLBACK_URL` that we already set. Now I am not sure what that does. Maybe we do not need it anymore?
CC @dtantsur and @Rozzii since you may have insights into these variables

dtantsur:
Do we have an idea why it broke all of the sudden?

I'm somewhat worried about testing with settings that don't really resemble a production deployment. Also, these hacks won't be possible with IrSO.

> I tried using IRONIC_EXTERNAL_IP first. That did not have any effect on the ipxe config.

Yep, EXTERNAL variables only affect virtual media, they were added for deployments over WAN.

lentzi90:
This is still the same issue as in https://github.com/metal3-io/baremetal-operator/issues/2456#issuecomment-2886469301.
It broke because the github runners bumped docker from 26 to 28. Probably the security issues that you linked to.
In https://github.com/metal3-io/baremetal-operator/pull/2454 we just got virtualmedia back to green (and disabled IPMI tests on PRs). IPMI and redfish are still red in the periodics: https://github.com/metal3-io/baremetal-operator/actions/workflows/e2e-test-periodic-main.yml They should go green with this PR, and we enable IPMI testing on PRs again.

I can see how this may not be an ideal configuration to test, but changing that will take some more time and should probably be planned a bit better. We could go back to minikube with the libvirt driver, but the setup we used was very flaky. We could roll our own management VM but that sounds like a lot of work. We could run ironic directly on the host in a container but that is again further from a working IrSO configuration.

Honestly, this configuration isn't so bad. We forward ports into ironic in the kind cluster. That requires overriding the IPs that ironic itself detects, but this is supposed to be supported in the image, right? It may not be the "end game" for how to run ironic, more like a bootstrap setup.
I'm a bit confused over why some of the variables doesn't work. That may be because I have misunderstood how or for what they are. For example, I don't get IRONIC_EXTERNAL_IP, IRONIC_EXTERNAL_HTTP_URL and IRONIC_EXTERNAL_CALLBACK_URL. They don't seem to help even though it sounds like they are for exactly this purpose.
I expected IPA to use IRONIC_EXTERNAL_CALLBACK_URL when sending back inspection data but no, I had to set IRONIC_BASE_URL for that.

lentzi90:
Now that I think about it, I wonder if there was a change in IPA that broken the IRONIC_EXTERNAL_CALLBACK_URL...
I noticed this a while back:
> When the new Ironic built-in inspection is used, ipa-inspection-callback-url can now be automatically derived from ipa-api-url. In this case, inspection will be enabled if the ipa-inspection-collectors option is set.

From https://docs.openstack.org/releasenotes/ironic-python-agent/2024.1.html

Is it possible that IPA now ignores the callback URL and just uses ipa-api-url? Perhaps the older docker version masked this for us so we didn't notice?
The only place we use IRONIC_EXTERNAL_CALLBACK_URL as far as I know is in the ironic config:
https://github.com/metal3-io/ironic-image/blob/22de03ea9086639c96399a5535972a81a6b96a92/ironic-config/ironic.conf.j2#L130
We set the ipa-inspection-callback-url based on IRONIC_BASE_URL, and that is at least used in the inspector.ipxe, maybe other places also:
https://github.com/search?q=repo%3Ametal3-io%2Fironic-image%20INSPECTOR_EXTRA_ARGS&type=code

pierrecregut:
@lentzi90 IRONIC_EXTERNAL_CALLBACK_URL may be related with https://bugs.launchpad.net/ironic/+bug/2101173 that should be solved by https://review.opendev.org/c/openstack/ironic/+/949521 I filled after a slack discussion with @dtantsur   . If you can easily have a patched ironic image it may worth trying.

lentzi90:
> @lentzi90 IRONIC_EXTERNAL_CALLBACK_URL may be related with https://bugs.launchpad.net/ironic/+bug/2101173 I filled after a slack discussion with @dtantsur It should be solved by https://review.opendev.org/c/openstack/ironic/+/949521 . If you can easily have a patched ironic image it may worth trying.

Thanks for linking! I guess that will allow us to get rid of the IRONIC_BASE_URL then maybe?

Rozzii (commented):
/approve

dtantsur:
> That may be because I have misunderstood how or for what they are.

Which is one of the biggest reasons I started IrSO :)

PR #2477, opened 2025-05-22
✨ Allow disabling drivers that require a provisioning network

dtantsur:
It's a common mistake among my customers to use drivers like `ipmi`,
`redfish` (as opposed to e.g. `redfish-virtualmedia`) without configuring
a provisioning network.

This change adds a new environment variable to tell BMO that no
provisioning network is available.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -46,6 +46,10 @@ BMO_CONCURRENCY value lower than the requested PROVISIONING_LIMIT. Default is 20
 image for nodes that use IPv6. In dual stack environments, this can be
 used to tell Ironic which IP version it should set on the BMC.
 
+`PROVISIONING_NETWORK_DISABLED` -- Set to `true` if your deployment does not
+feature a provisioning network. This option disables drivers that require a
+provisioning network (such as IPMI).
+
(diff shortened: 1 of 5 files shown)


docs/configuration.md
```diff
@@ -46,6 +46,10 @@ BMO_CONCURRENCY value lower than the requested PROVISIONING_LIMIT. Default is 20
 image for nodes that use IPv6. In dual stack environments, this can be
 used to tell Ironic which IP version it should set on the BMC.
 
+`PROVISIONING_NETWORK_DISABLED` -- Set to `true` if your deployment does not
```
s3rj1k, line 49: I would suggest expanding on `PROVISIONING_NETWORK` term, Redfish can also use `provisioning network` but idea here is to facilitate deprecation of IPMI is I understand it

s3rj1k: ok, so maybe list should be expanded to have also other drivers as example, to make the term a bit more clearer.

s3rj1k: yea, lets just change that to something like `drivers like IPMI, redfish but not redfish-virtualmedia`

PR #2504, opened 2025-06-03
🌱 Bump golang to 1.24.3

Rozzii:
- gomods are still using v1.23.x
- Dockerfile and prow jobs are bumped to use v1.24.3


tuminoid (commented):
/lgtm

PR #2514, opened 2025-06-09
:seedling: Enable nilnil linter, fix warnings

### LINKED ISSUE #2360: Enable "nilnil" linter in golangci-lint
Enable "nilnil" linter in golangci-lint. Fix any warnings enabling it causes.

/area ci
/help
/triage accepted

sirpaleet:
Enable nilnil linter and fix warnings flagged by it. Add .vscode/ directory to gitignore.

Fixes #2360

### DIFF
--- a/internal/controller/metal3.io/baremetalhost_controller.go
+++ b/internal/controller/metal3.io/baremetalhost_controller.go
@@ -709,7 +709,7 @@ func getHostArchitecture(host *metal3api.BareMetalHost) string {
 func (r *BareMetalHostReconciler) getPreprovImage(info *reconcileInfo, formats []metal3api.ImageFormat) (*provisioner.PreprovisioningImage, error) {
 	if formats == nil {
 		// No image build requested
-		return nil, nil
+		return nil, nil //nolint:nilnil
 	}
 
 	if len(formats) == 0 {
@@ -2047,7 +2047,7 @@ func (r *BareMetalHostReconciler) acquireHostUpdatePolicy(info *reconcileInfo) (
 			// garbage collected. For additional cleanup logic use
 			// finalizers.  Return and don't requeue
 
-			return nil, nil
+			return nil, nil //nolint:nilnil
 		}
 		// Error reading the object
 		return nil, fmt.Errorf("could not load hostUpdatePolicy resource due to %w", err)
@@ -2150,7 +2150,7 @@ func (r *BareMetalHostReconciler) getHostStatusFromAnnotation(host *metal3api.Ba
 	annotations := host.GetAnnotations()
 	content := []byte(annotations[metal3api.StatusAnnotation])
 	if annotations[metal3api.StatusAnnotation] == "" {
-		return nil, nil
+		return nil, errors.New("status annotation not found")
 	}
 	objStatus, err := unmarshalStatusAnnotation(content)
 	if err != nil {
@@ -2163,7 +2163,7 @@ func (r *BareMetalHostReconciler) getHostStatusFromAnnotation(host *metal3api.Ba
 func (r *BareMetalHostReconciler) getHardwareDetailsFromAnnotation(host *metal3api.BareMetalHost) (*metal3api.HardwareDetails, error) {
 	annotations := host.GetAnnotations()
 	if annotations[metal3api.HardwareDetailsAnnotation] == "" {
-		return nil, nil
+		return nil, nil //nolint:nilnil
 	}
 	objHardwareDetails := &metal3api.HardwareDetails{}
 	decoder := json.NewDecoder(strings.NewReader(annotations[metal3api.HardwareDetailsAnnotation]))
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -232,6 +232,7 @@ func (p *ironicProvisioner) isAddressAllocatedToPort(address string) (bool, erro
 }
 
 // Look for an existing registration for the host in Ironic.
+// Returns nil, nil if no host exists with MAC.
 func (p *ironicProvisioner) findExistingHost(bootMACAddress string) (ironicNode *nodes.Node, err error) {
 	// Try to load the node by UUID
 	ironicNode, err = p.getNode()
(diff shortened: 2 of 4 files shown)


internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -709,7 +709,7 @@ func getHostArchitecture(host *metal3api.BareMetalHost) string {
 func (r *BareMetalHostReconciler) getPreprovImage(info *reconcileInfo, formats []metal3api.ImageFormat) (*provisioner.PreprovisioningImage, error) {
 	if formats == nil {
 		// No image build requested
-		return nil, nil
+		return nil, nil //nolint:nilnil
```
tuminoid, line 712: This could be handled otherwise if the check for if formats == nil should be at caller side at L820, and this function should fail at L715 for both formats == nil and len(formats) == 0.

It is also working as is.

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -2047,7 +2047,7 @@ func (r *BareMetalHostReconciler) acquireHostUpdatePolicy(info *reconcileInfo) (
 			// garbage collected. For additional cleanup logic use
 			// finalizers.  Return and don't requeue
 
-			return nil, nil
+			return nil, nil //nolint:nilnil
```
tuminoid, line 2050: This one seems legit, we actively choose to ignore it.

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -2163,7 +2163,7 @@ func (r *BareMetalHostReconciler) getHostStatusFromAnnotation(host *metal3api.Ba
 func (r *BareMetalHostReconciler) getHardwareDetailsFromAnnotation(host *metal3api.BareMetalHost) (*metal3api.HardwareDetails, error) {
 	annotations := host.GetAnnotations()
 	if annotations[metal3api.HardwareDetailsAnnotation] == "" {
-		return nil, nil
+		return nil, nil //nolint:nilnil
```
tuminoid, line 2166: This seems legit as well.

pkg/provisioner/ironic/ironic.go
```diff
@@ -266,7 +266,7 @@ func (p *ironicProvisioner) findExistingHost(bootMACAddress string) (ironicNode
 
 	if err != nil {
 		p.log.Info("failed to find an existing port with address", "MAC", bootMACAddress)
-		return nil, nil //nolint:nilerr
+		return nil, nil //nolint:nilerr,nilnil
```
tuminoid, line 270: This is legit as well. It concludes it doesn't have a node by the MAC and can register it.

pkg/provisioner/ironic/ironic.go
```diff
@@ -291,7 +291,7 @@ func (p *ironicProvisioner) findExistingHost(bootMACAddress string) (ironicNode
 	p.log.Info("port with address doesn't exist", "MAC", bootMACAddress)
 	// Either the node was never created or the Ironic database has
 	// been dropped.
-	return nil, nil
+	return nil, fmt.Errorf("node with MAC %s not found", bootMACAddress)
```
tuminoid: This is not an error. If the node isn't found, it can be created which happens on nil nil return.

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -2150,7 +2150,7 @@ func (r *BareMetalHostReconciler) getHostStatusFromAnnotation(host *metal3api.Ba
 	annotations := host.GetAnnotations()
 	content := []byte(annotations[metal3api.StatusAnnotation])
 	if annotations[metal3api.StatusAnnotation] == "" {
-		return nil, nil
+		return nil, errors.New("status annotation not found")
```
tuminoid, line 2153: Similar code as L2166, and caller does not even handle the case where Err != nil. It does have earlier check that it has some status, so this will probably not work or will never trigger.

internal/controller/metal3.io/baremetalhost_controller.go
lentzi90: You mean legit as in we should do something about it? Or legit as in we should have the nolint?

sirpaleet: From what I understood, legit as in should have nolint.

tuminoid: Meant legit nolint indeed. Not a major fan of function design like these, so I might be wrong too, so PTAL.

pkg/provisioner/ironic/ironic.go
lentzi90: I agree. @sirpaleet please put this back as `nil, nil` and add nolint.
Could you also add to the doc string at the top of the function that it returns `nil, nil` if no host with this MAC exists? Just to clarify that it is not considered an error.

internal/controller/metal3.io/baremetalhost_controller.go
lentzi90: Yeah this will never trigger as the code looks now, but I do think it is better this way.

internal/controller/metal3.io/baremetalhost_controller.go
sirpaleet: Working on this now: len(formats) == 0 works for nil values also, so is it okay to not have the format == nil check on the caller side al all? Just using len(formats) == 0 would catch this before anything is done. It will return an imagebuilderror with this (unless that is not acceptable for formats == nil, then I will do another check to define the error type).

sirpaleet: Just tested my suggestion does not work :dancing_men:  Will edit

sirpaleet: Tested your suggestion. This would work if the tests didn't also test on formats == nil value and the test requires the value to be Nil Upon output. Is it better to change the tests so that they expect an error value (like imagebuilderror) or just leave the formats == nil value check as was?

lentzi90: I would say just leave it for now

tuminoid: Yeah, it is working as is, so let's not fix that now.

lentzi90 (commented):
/approve

tuminoid (commented):
/lgtm

PR #2516, opened 2025-06-10
:bug: don't wait for dataImage detachment when the host is detached and requested for deletion

hroyrh:

When the host is in detached and a deletion request is raised, then associated dataImage should also get deleted. Right now, the dataImage reconciliation stops when the host is in detached state and hence the host gets stuck in deletion.

### DIFF
--- a/internal/controller/metal3.io/dataimage_controller.go
+++ b/internal/controller/metal3.io/dataimage_controller.go
@@ -163,8 +163,22 @@ func (r *DataImageReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		return ctrl.Result{Requeue: true}, nil
 	}
 
-	// If the associated BMH is detached, keep requeuing till the annotation is removed
 	if hasDetachedAnnotation(bmh) {
+		// In detached state, if dataImage deletion is requested, then we remove the
+		// finalizer without checking for dataImage attached status
+		if !di.DeletionTimestamp.IsZero() {
+			reqLogger.Info("dataImage deletion requested in detached state, removing finalizer")
+
+			di.Finalizers = utils.FilterStringFromList(
+				di.Finalizers, metal3api.DataImageFinalizer)
+
+			if err := r.Update(ctx, di); err != nil {
+				return ctrl.Result{Requeue: true, RequeueAfter: dataImageRetryDelay}, fmt.Errorf("failed to update resource after remove finalizer, %w", err)
+			}
+			return ctrl.Result{}, nil
+		}
+
+		// If the associated BMH is detached, keep requeuing till the annotation is removed
 		reqLogger.Info("the host is detached, not running reconciler")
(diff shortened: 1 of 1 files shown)


internal/controller/metal3.io/host_state_machine.go
```diff
@@ -240,6 +240,18 @@ func (hsm *hostStateMachine) checkInitiateDelete(log logr.Logger) bool {
...
+		if hsm.Host.OperationalStatus() == metal3api.OperationalStatusDetached {
+			if delayDeleteForDetachedHost(hsm.Host) {
+				log.Info("Delaying detached host deletion")
+				deleteDelayedForDetached.Inc()
+				return false
+			}
+			// We cannot power off a detached host.  Skip to delete.
+			hsm.NextState = metal3api.StateDeleting
+		} else {
+			hsm.NextState = metal3api.StatePoweringOffBeforeDelete
```
hroyrh, line 256: Do we need to powerOffBeforeDelete for externallyProvisioned host or should we directly delete it in all cases, whether the detached annotation is present of not ?

iurygregory: I think we would need, because ironic still responsible for handling the power actions 
@dtantsur thoughts on this?

hroyrh: Yes, but for externallyProvisioned host, is it desirable to power off the host before deleting it from bmo, since the host will remain "provisioned", it's just we won't be managing the power of the host anymore.

tuminoid:
Squash commits please.

internal/controller/metal3.io/host_state_machine.go
Rozzii: If we would like to stay consistent with the logic of deleting a regularly provisioned BMH then we have to actually deprovision it like on L244 instead of just turning it off .

Rozzii (commented):
In general the PR makes sense, but the handling of the detached + externally provisioned node might need a tiny change.

internal/controller/metal3.io/host_state_machine.go
hroyrh: I don't think we are supposed to deprovision an enternallyProvisioned host, right ?

Rozzii: Indeed, my bad sorry, thought that it can be deprovisioned but seems we have decided that at least for now there is no option to deprovision it. I think the missing piece here is that there is still no functionality implemented to move an externally provisioned node to provisioned state.

Anyways I will approve the PR.

Rozzii (commented):
/approve

internal/controller/metal3.io/host_state_machine.go
hroyrh: Right, and even if we had the functionality to move it to provisioned I think the user will still need to set `externallyProvisioned` to false.

PR #2521, opened 2025-06-13
:seedling: Enable variable shadowing check in govet linter

### LINKED ISSUE #2460: Enable variable shadowing check in govet linter
If we enable variable shadowing check in govet, we get a lot of errors. Variable shadowing can cause hard to debug issues.

```diff
diff --git a/.golangci.yaml b/.golangci.yaml
index 075a0062..0302721d 100644
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -124,6 +124,9 @@ linters-settings:
     - unnecessaryDefer
     - whyNoLint
     - wrapperFunc
+  govet:
+    enable:
+    - shadow
 issues:
   exclude-dirs:
   - mock*
```

leads to

```text
main.go:235:12: shadow: declaration of "err" shadows declaration at line 195 (govet)
		seconds, err := strconv.ParseInt(leaseDurationSeconds, 10, 16)
		         ^
main.go:246:12: shadow: declaration of "err" shadows declaration at line 195 (govet)
		seconds, err := strconv.ParseInt(renewDeadlineSeconds, 10, 16)
		         ^
main.go:257:12: shadow: declaration of "err" shadows declaration at line 195 (govet)
		seconds, err := strconv.ParseInt(retryPeriodSeconds, 10, 16)
		         ^
internal/controller/metal3.io/baremetalhost_controller.go:173:3: shadow: declaration of "err" shadows declaration at line 109 (govet)
		err := r.Update(ctx, host)
		^
internal/controller/metal3.io/baremetalhost_controller.go:1019:7: shadow: declaration of "err" shadows declaration at line 989 (govet)
			if err := r.Update(info.ctx, info.host); err != nil {
			   ^
internal/controller/metal3.io/baremetalhost_controller.go:1485:7: shadow: declaration of "err" shadows declaration at line 1476 (govet)
			if err := r.Status().Update(info.ctx, hfc); err != nil {
			   ^
internal/controller/metal3.io/baremetalhost_controller.go:1581:8: shadow: declaration of "err" shadows declaration at line 1534 (govet)
		hup, err := r.acquireHostUpdatePolicy(info)
		     ^
internal/controller/metal3.io/baremetalhost_controller.go:1813:6: shadow: declaration of "err" shadows declaration at line 1807 (govet)
		if err := r.Status().Update(info.ctx, dataImage); err != nil {
		   ^
internal/controller/metal3.io/baremetalhost_controller.go:1843:6: shadow: declaration of "err" shadows declaration at line 1837 (govet)
		if err := r.Status().Update(info.ctx, dataImage); err != nil {
		   ^
internal/controller/metal3.io/dataimage_controller.go:122:7: shadow: declaration of "err" shadows declaration at line 115 (govet)
			if err := r.Update(ctx, di); err != nil {
			   ^
internal/controller/metal3.io/preprovisioningimage_controller.go:85:6: shadow: declaration of "err" shadows declaration at line 74 (govet)
		if err := r.discardExistingImage(&img, log); err != nil {
		   ^
internal/controller/metal3.io/preprovisioningimage_controller.go:90:3: shadow: declaration of "err" shadows declaration at line 74 (govet)
		err := r.Update(ctx, &img)
		^
internal/controller/metal3.io/preprovisioningimage_controller.go:100:3: shadow: declaration of "err" shadows declaration at line 74 (govet)
		err := r.Update(ctx, &img)
		^
internal/controller/metal3.io/preprovisioningimage_controller.go:176:7: shadow: declaration of "err" shadows declaration at line 160 (govet)
			if err := r.discardExistingImage(img, log); err != nil {
			   ^
pkg/provisioner/ironic/ironic.go:1257:18: shadow: declaration of "err" shadows declaration at line 1227 (govet)
		if provResult, err := p.setUpForProvisioning(ironicNode, data); err != nil || provResult.Dirty || provResult.ErrorMessage != "" {
		               ^
pkg/provisioner/ironic/ironic.go:1261:16: shadow: declaration of "err" shadows declaration at line 1227 (govet)
		configDrive, err := p.getConfigDrive(data)
		             ^
pkg/provisioner/ironic/ironic.go:1290:18: shadow: declaration of "err" shadows declaration at line 1227 (govet)
		if provResult, err := p.setUpForProvisioning(ironicNode, data); err != nil || provResult.Dirty || provResult.ErrorMessage != "" {
		               ^
pkg/provisioner/ironic/ironic.go:1297:16: shadow: declaration of "err" shadows declaration at line 1227 (govet)
		configDrive, err := p.getConfigDrive(data)
		             ^
pkg/provisioner/ironic/ironic.go:1490:24: shadow: declaration of "err" shadows declaration at line 1465 (govet)
			_, success, result, err := p.tryUpdateNode(ironicNode, updater)
			                    ^
pkg/provisioner/ironic/clients/auth.go:46:8: shadow: declaration of "err" shadows declaration at line 43 (govet)
	if _, err := os.Stat(authPath); err != nil {
	      ^
pkg/provisioner/ironic/testserver/server.go:94:6: shadow: declaration of "response" shadows declaration at line 28 (govet)
		if response, ok := m.responsesByMethod[r.URL.Path][r.Method]; ok {
		   ^
pkg/provisioner/ironic/testserver/server.go:206:2: shadow: declaration of "defaultResponse" shadows declaration at line 33 (govet)
	defaultResponse := defaultResponse{
	^
INFO File cache stats: 21 entries of total size 302.4KiB 
INFO Memory: 440 samples, avg is 1359.4MB, max is 4326.0MB 
INFO Execution took 44.438255091s                 
make: *** [Makefile:142: lint] Error 1
```

/kind ci
/help

sirpaleet:
Enable variable shadowing check in govet linter. Fix warnings caused by running the linter.

Fixes #2460

### DIFF
--- a/hack/tools/deploy-cli/deploy-cli.go
+++ b/hack/tools/deploy-cli/deploy-cli.go
@@ -80,7 +80,7 @@ func (d *DeployContext) determineIronicAuth() error {
 	ironicAuthDir := filepath.Join(ironicDataDir, "auth")
 	ironicUsernameFile := filepath.Join(ironicAuthDir, "ironic-username")
 
-	if err := os.MkdirAll(ironicAuthDir, filePerm755); err != nil {
+	if err = os.MkdirAll(ironicAuthDir, filePerm755); err != nil {
 		return err
 	}
 
@@ -119,7 +119,6 @@ func (d *DeployContext) determineIronicAuth() error {
 // deployIronic configures the kustomize overlay for ironic
 // based on the configuration, then install ironic with that overlay.
 func (d *DeployContext) deployIronic() error {
-	var err error
 	ironicDataDir, err := d.GetEnvOrDefault("IRONIC_DATA_DIR")
 	if err != nil {
 		return err
@@ -127,7 +126,7 @@ func (d *DeployContext) deployIronic() error {
 	configMapFileName := "ironic_bmo_configmap.env"
 	configMapInDataDir := filepath.Join(ironicDataDir, configMapFileName)
 
-	if _, err := os.Stat(configMapInDataDir); err == nil && d.IronicBMOConfigMapEnvFile == "" {
+	if _, err = os.Stat(configMapInDataDir); err == nil && d.IronicBMOConfigMapEnvFile == "" {
 		log.Printf("Detected ironic_bmo_configmap.env file in IRONIC_DATA_DIR: %s. Taken into use.", configMapInDataDir)
 		d.IronicBMOConfigMapEnvFile = configMapInDataDir
 	}
@@ -149,14 +148,14 @@ func (d *DeployContext) deployIronic() error {
 
 		kustomizeFile := filepath.Join(ironicOverlay, "kustomization.yaml")
 
-		if err := RenderEmbedTemplateToFile(d.TemplateFiles, ironicKustomizeTpl, kustomizeFile, d); err != nil {
+		if err = RenderEmbedTemplateToFile(d.TemplateFiles, ironicKustomizeTpl, kustomizeFile, d); err != nil {
 			return err
 		}
 
 		ironicBMOConfigMapTpl := "templates/ironic_bmo_configmap_env.tpl"
 		ironicBMOConfigMapOutput := filepath.Join(ironicOverlay, configMapFileName)
 
-		if err := RenderEmbedTemplateToFile(d.TemplateFiles, ironicBMOConfigMapTpl, ironicBMOConfigMapOutput, d); err != nil {
+		if err = RenderEmbedTemplateToFile(d.TemplateFiles, ironicBMOConfigMapTpl, ironicBMOConfigMapOutput, d); err != nil {
 			return err
 		}
 
@@ -168,7 +167,7 @@ func (d *DeployContext) deployIronic() error {
 
 	if d.IronicBMOConfigMapEnvFile != "" {
 		log.Printf("Using custom ironic_bmo_configmap.env file: %s", d.IronicBMOConfigMapEnvFile)
-		if err := CopyFile(d.IronicBMOConfigMapEnvFile, filepath.Join(ironicOverlay, "ironic_bmo_configmap.env")); err != nil {
+		if err = CopyFile(d.IronicBMOConfigMapEnvFile, filepath.Join(ironicOverlay, "ironic_bmo_configmap.env")); err != nil {
 			return err
 		}
 	}
@@ -176,7 +175,8 @@ func (d *DeployContext) deployIronic() error {
 	username, password := d.IronicUsername, d.IronicPassword
 
 	if username != "" && password != "" {
-		ironicHtpasswd, err := GenerateHtpasswd(username, password)
+		var ironicHtpasswd string
+		ironicHtpasswd, err = GenerateHtpasswd(username, password)
 		if err != nil {
 			return err
 		}
@@ -200,7 +200,7 @@ func (d *DeployContext) deployBMO() error {
 	ironicEnvFileName := "ironic.env"
 	ironicEnvInDataDir := filepath.Join(ironicDataDir, ironicEnvFileName)
 
-	if _, err := os.Stat(ironicEnvInDataDir); err == nil && d.IronicEnvFile == "" {
+	if _, err = os.Stat(ironicEnvInDataDir); err == nil && d.IronicEnvFile == "" {
 		log.Printf("Detected ironic.env file in IRONIC_DATA_DIR: %s. Taken into use.", ironicEnvInDataDir)
 		d.IronicEnvFile = ironicEnvInDataDir
 	}
--- a/hack/tools/deploy-cli/main.go
+++ b/hack/tools/deploy-cli/main.go
@@ -117,7 +117,7 @@ func main() {
 		os.Exit(1)
 	}
 
-	if err := dc.determineIronicAuth(); err != nil {
+	if err = dc.determineIronicAuth(); err != nil {
 		log.Fatalf("Error retrieving Ironic username/password: %v", err)
 	}
 
--- a/internal/controller/metal3.io/baremetalhost_controller.go
+++ b/internal/controller/metal3.io/baremetalhost_controller.go
@@ -170,7 +170,7 @@ func (r *BareMetalHostReconciler) Reconcile(ctx context.Context, request ctrl.Re
 		)
 		host.Finalizers = append(host.Finalizers,
 			metal3api.BareMetalHostFinalizer)
-		err := r.Update(ctx, host)
+		err = r.Update(ctx, host)
 		if err != nil {
 			return ctrl.Result{}, fmt.Errorf("failed to add finalizer: %w", err)
 		}
@@ -1015,7 +1015,7 @@ func (r *BareMetalHostReconciler) actionInspecting(prov provisioner.Provisioner,
 		}
 
 		if dirty {
-			if err := r.Update(info.ctx, info.host); err != nil {
+			if err = r.Update(info.ctx, info.host); err != nil {
 				return actionError{fmt.Errorf("failed to update the host after inspection start: %w", err)}
 			}
 			return actionContinue{}
@@ -1067,10 +1067,10 @@ func (r *BareMetalHostReconciler) actionInspecting(prov provisioner.Provisioner,
 		if controllerutil.ContainsFinalizer(hardwareData, hardwareDataFinalizer) {
 			controllerutil.RemoveFinalizer(hardwareData, hardwareDataFinalizer)
 		}
-		if err := r.Update(info.ctx, hardwareData); err != nil {
+		if err = r.Update(info.ctx, hardwareData); err != nil {
 			return actionError{fmt.Errorf("failed to remove hardwareData finalizer: %w", err)}
 		}
-		if err := r.Client.Delete(info.ctx, hd); err != nil {
+		if err = r.Client.Delete(info.ctx, hd); err != nil {
 			return actionError{fmt.Errorf("failed to delete hardwareData: %w", err)}
 		}
 	}
@@ -1480,7 +1480,7 @@ func (r *BareMetalHostReconciler) doServiceIfNeeded(prov provisioner.Provisioner
 		info.host.Status.Provisioning.Firmware = nil
 		if hfcDirty && hfc.Status.Updates != nil {
 			hfc.Status.Updates = nil
-			if err := r.Status().Update(info.ctx, hfc); err != nil {
+			if err = r.Status().Update(info.ctx, hfc); err != nil {
 				return actionError{fmt.Errorf("failed to update hostfirmwarecomponents status: %w", err)}
 			}
 		}
@@ -1576,7 +1576,8 @@ func (r *BareMetalHostReconciler) manageHostPower(prov provisioner.Provisioner, 
 
 	servicingAllowed := isProvisioned && !info.host.Status.PoweredOn && desiredPowerOnState
 	if servicingAllowed || info.host.Status.OperationalStatus == metal3api.OperationalStatusServicing || info.host.Status.ErrorType == metal3api.ServicingError {
-		hup, err := r.acquireHostUpdatePolicy(info)
+		var hup *metal3api.HostUpdatePolicy
+		hup, err = r.acquireHostUpdatePolicy(info)
 		if err != nil {
 			info.log.Info("failed setting owner reference on hostupdatepolicy")
 			return actionError{fmt.Errorf("failed setting owner reference on hostUpdatePolicy: %w", err)}
@@ -1808,8 +1809,8 @@ func (r *BareMetalHostReconciler) attachDataImage(prov provisioner.Provisioner, 
 		dataImage.Status.Error.Count++
 		dataImage.Status.Error.Message = err.Error()
 		// Error updating DataImage Status
-		if err := r.Status().Update(info.ctx, dataImage); err != nil {
-			return fmt.Errorf("failed to update DataImage status, %w", err)
+		if errOnUpdate := r.Status().Update(info.ctx, dataImage); errOnUpdate != nil {
+			return fmt.Errorf("failed to update DataImage status, %w", errOnUpdate)
 		}
 
 		return fmt.Errorf("failed to attach dataImage, %w", err)
@@ -1838,8 +1839,8 @@ func (r *BareMetalHostReconciler) detachDataImage(prov provisioner.Provisioner, 
 		dataImage.Status.Error.Count++
 		dataImage.Status.Error.Message = err.Error()
 		// Error updating DataImage Status
-		if err := r.Status().Update(info.ctx, dataImage); err != nil {
-			return fmt.Errorf("failed to update DataImage status, %w", err)
+		if errOnUpdate := r.Status().Update(info.ctx, dataImage); errOnUpdate != nil {
+			return fmt.Errorf("failed to update DataImage status, %w", errOnUpdate)
 		}
 
 		return fmt.Errorf("failed to detach dataImage, %w", err)
--- a/main.go
+++ b/main.go
@@ -231,7 +231,8 @@ func main() {
 	}
 
 	if leaseDurationSeconds != "" {
-		seconds, err := strconv.ParseInt(leaseDurationSeconds, 10, 16)
+		var seconds int64
+		seconds, err = strconv.ParseInt(leaseDurationSeconds, 10, 16)
 		if err != nil {
 			setupLog.Error(err, "failed to parse duration")
 			os.Exit(1)
@@ -242,7 +243,8 @@ func main() {
 	}
 
 	if renewDeadlineSeconds != "" {
-		seconds, err := strconv.ParseInt(renewDeadlineSeconds, 10, 16)
+		var seconds int64
+		seconds, err = strconv.ParseInt(renewDeadlineSeconds, 10, 16)
 		if err != nil {
 			setupLog.Error(err, "failed to parse renew deadline")
 			os.Exit(1)
@@ -253,7 +255,8 @@ func main() {
 	}
 
 	if retryPeriodSeconds != "" {
-		seconds, err := strconv.ParseInt(retryPeriodSeconds, 10, 16)
+		var seconds int64
+		seconds, err = strconv.ParseInt(retryPeriodSeconds, 10, 16)
 		if err != nil {
 			setupLog.Error(err, "failed to parse retry period")
 			os.Exit(1)
--- a/pkg/provisioner/ironic/testserver/server.go
+++ b/pkg/provisioner/ironic/testserver/server.go
@@ -91,8 +91,8 @@ func (m *MockServer) Handler(pattern string, handlerFunc http.HandlerFunc) *Mock
 
 func (m *MockServer) buildHandler(_ string) func(http.ResponseWriter, *http.Request) {
 	handler := func(w http.ResponseWriter, r *http.Request) {
-		if response, ok := m.responsesByMethod[r.URL.Path][r.Method]; ok {
-			m.sendData(w, r, response.code, response.payload)
+		if resp, ok := m.responsesByMethod[r.URL.Path][r.Method]; ok {
+			m.sendData(w, r, resp.code, resp.payload)
 			return
 		}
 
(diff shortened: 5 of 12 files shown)


tuminoid:
This should be something Copilot can handle, let's test.

tuminoid:
> Comments suppressed due to low confidence (3)
> **test/e2e/e2e_suite_test.go:170**
> 
> * [nitpick] The variable name `clstrProxy` is abbreviated and may be unclear; consider a more descriptive name like `clusterProxyClient` or simply `proxy` to improve readability.
> 
> ```
> clstrProxy := framework.NewClusterProxy("bmo-e2e", kubeconfigPath, scheme)
> ```

I agree with variable name issue, use a proper name.

pkg/provisioner/ironic/testserver/server.go
```diff
@@ -91,8 +91,8 @@ func (m *MockServer) Handler(pattern string, handlerFunc http.HandlerFunc) *Mock
 
 func (m *MockServer) buildHandler(_ string) func(http.ResponseWriter, *http.Request) {
 	handler := func(w http.ResponseWriter, r *http.Request) {
-		if response, ok := m.responsesByMethod[r.URL.Path][r.Method]; ok {
-			m.sendData(w, r, response.code, response.payload)
+		if resp, ok := m.responsesByMethod[r.URL.Path][r.Method]; ok {
+			m.sendData(w, r, resp.code, resp.payload)
```
lentzi90, line 95: Why change this? I think it is better to keep the full word.

pkg/provisioner/ironic/testserver/server.go
```diff
@@ -203,7 +203,7 @@ func (m *MockServer) AddDefaultResponse(patternWithVars string, httpMethod strin
 	pattern := "^" + regexp.MustCompile("{(.[^}]*)}").ReplaceAllString(patternWithVars, "(?P<$1>.[^/]*)") + "$"
 	m.t.Logf("%s: adding default response for %s (%s) -> {%d, %s}", m.name, patternWithVars, pattern, code, payload)
 
-	defaultResponse := defaultResponse{
+	defaultResp := defaultResponse{
```
lentzi90, line 206: Here also I think it is better to keep it as it was

pkg/provisioner/ironic/testserver/server.go
```diff
@@ -212,7 +212,7 @@ func (m *MockServer) AddDefaultResponse(patternWithVars string, httpMethod strin
 		},
 	}
 
-	m.defaultResponses = append(m.defaultResponses, defaultResponse)
+	m.defaultResponses = append(m.defaultResponses, defaultResp)
```
lentzi90, line 215: Here also I think it is better to keep it as it was

test/e2e/e2e_suite_test.go
lentzi90, line 1: Please don't make changes without reason when enabling linters. It makes it harder to see what changed and why.
In this file I would like to have `clusterProxy` back as it was. I don't see a reason to change it to `proxy`.

sirpaleet: This was because of the shadowing check as well, the linter complains about shadowing on line 56 declaration, and just writing over the value seemed not wanted because the value for this line 56 variable was given for the first time in a consequential function. Where should I usually state individual reasons for changes like this, in case they are unclear?

pkg/provisioner/ironic/testserver/server.go
sirpaleet: This was because of shadowing over variable defined at line 28. "response" seems to be a struct in the same file so naming the variable also as such caused linter warnings.

pkg/provisioner/ironic/testserver/server.go
sirpaleet: This is also because of shadowing a struct in the same file, should I overwrite the name to indicate a variable or just ignore the  linter warning here?

pkg/provisioner/ironic/testserver/server.go
sirpaleet: Same reasons here

test/e2e/e2e_suite_test.go
lentzi90: Interesting! Line 56 is meant to be the declaration of the variable, as far as I understand. What if we put back clusterProxy and remove the : so that it is using the already declared variable? Does that work?
I guess there could be some reason why we shadow it here. Maybe to have the bootstrap proxy separate from the workload/management proxy? Anyway, this is exactly why this linter is good. Instead of shadowing, we should use separate names if that is needed. But let's first try just without the `:`.

> Where should I usually state individual reasons for changes like this, in case they are unclear?

In this case I would do it as a "self review". I.e. create the PR and then add a comment directly in the diff on why you changed it. Another option, is to put the comment directly in the code. E.g. "we cannot use `name` here because it was already used before" or something like that.

pkg/provisioner/ironic/testserver/server.go
lentzi90: Make sense! Thanks for explaining :slightly_smiling_face:

pkg/provisioner/ironic/testserver/server.go
lentzi90: Oh my. Well then it is ok :+1:

hack/tools/deploy-cli/deploy-cli.go
```diff
@@ -119,15 +119,14 @@ func (d *DeployContext) determineIronicAuth() error {
 // deployIronic configures the kustomize overlay for ironic
 // based on the configuration, then install ironic with that overlay.
 func (d *DeployContext) deployIronic() error {
-	var err error
```
sirpaleet, line 122: Removed unnecessary declaration, following renamings to avoid shadowing of err.

hack/tools/deploy-cli/deploy-cli.go
```diff
@@ -80,7 +80,7 @@ func (d *DeployContext) determineIronicAuth() error {
 	ironicAuthDir := filepath.Join(ironicDataDir, "auth")
 	ironicUsernameFile := filepath.Join(ironicAuthDir, "ironic-username")
 
-	if err := os.MkdirAll(ironicAuthDir, filePerm755); err != nil {
+	if err = os.MkdirAll(ironicAuthDir, filePerm755); err != nil {
```
sirpaleet, line 83: Overwrite of err value here seems reasonable since it is re-declared right after anyways.

hack/tools/deploy-cli/main.go
```diff
@@ -117,8 +117,8 @@ func main() {
 		os.Exit(1)
 	}
 
-	if err := dc.determineIronicAuth(); err != nil {
-		log.Fatalf("Error retrieving Ironic username/password: %v", err)
+	if errOnAuth := dc.determineIronicAuth(); errOnAuth != nil {
```
sirpaleet: Rename to avoid shadowing

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -107,6 +107,11 @@ func (info *reconcileInfo) publishEvent(reason, message string) {
 
 // Reconcile handles changes to BareMetalHost resources.
 func (r *BareMetalHostReconciler) Reconcile(ctx context.Context, request ctrl.Request) (result ctrl.Result, err error) {
+	var hostData ctrl.Result
```
sirpaleet: Declaring all err related variables here, to avoid err shadowing on line 178. All err re-declared on same scope as the first one can be just overwritten.

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -972,6 +977,11 @@ func updateRootDeviceHints(host *metal3api.BareMetalHost, info *reconcileInfo) (
 
 // Ensure we have the information about the hardware on the host.
 func (r *BareMetalHostReconciler) actionInspecting(prov provisioner.Provisioner, info *reconcileInfo) actionResult {
+	var err error
```
sirpaleet: Declaring err and related variables at top of function to avoid shadowing on line 1028 where it can be overwritten since the check always returns. Other cases use renaming to avoid shadowing and limiting to scope, because the err value might leak and not always return (however with nil value)

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -1067,11 +1077,11 @@ func (r *BareMetalHostReconciler) actionInspecting(prov provisioner.Provisioner,
 		if controllerutil.ContainsFinalizer(hardwareData, hardwareDataFinalizer) {
 			controllerutil.RemoveFinalizer(hardwareData, hardwareDataFinalizer)
 		}
-		if err := r.Update(info.ctx, hardwareData); err != nil {
-			return actionError{fmt.Errorf("failed to remove hardwareData finalizer: %w", err)}
+		if errOnUpdate := r.Update(info.ctx, hardwareData); errOnUpdate != nil {
```
sirpaleet: Example on the case of renaming value to avoid err unwanted nil value leakage, but maybe it is okay? Should this and others like be named back to "err" still?

test/e2e/e2e_suite_test.go
sirpaleet: Sorry about the clutter here. I did this and took away some unnecessary edits I made. I have added some comments but I am not too sure if I should still edit some of my name changes. Should I convert to draft and edit when I have time or is it okay to come back when the more time critical task is in better condition? It seems this needs more thought than I thought :''D.

lentzi90: No need to put it in draft, this is good :+1:

hack/tools/deploy-cli/deploy-cli.go
lentzi90: Yes this is good, we should do more this way. By removing the `:` we are no longer shadowing, we just reuse (explicitly) the same variable.

hack/tools/deploy-cli/deploy-cli.go
lentzi90: I think it would be better to remove the `:` where possible and just keep using `err`. It is the `:` that makes it "shadowing", since it means that we declare a new variable with the same name.
If we instead remove the `:`, we are explicitly reusing a variable. That is perfectly fine.

hack/tools/deploy-cli/main.go
lentzi90:
```suggestion
	if err = dc.determineIronicAuth(); errOnAuth != nil {
```

internal/controller/metal3.io/baremetalhost_controller.go
lentzi90: I'm not following. What do we leak?
I think it is fine to have separate names here. I also think it would be ok to reuse `err` (but not shadow it by using `:`). There is perhaps the risk that we later want to add some check of the original `err`, then the different names are needed. But I think it is ok either way here. It is quite clear what the code does (as long as we skip the shadowing).

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -1081,8 +1091,8 @@ func (r *BareMetalHostReconciler) actionInspecting(prov provisioner.Provisioner,
 	}
 
 	// either hardwareData was deleted above, or not found. We need to re-create it
-	if err := r.Client.Create(info.ctx, hd); err != nil {
-		return actionError{fmt.Errorf("failed to create hardwareData: %w", err)}
+	if errOnCreate := r.Client.Create(info.ctx, hd); errOnCreate != nil {
+		return actionError{fmt.Errorf("failed to create hardwareData: %w", errOnCreate)}
```
lentzi90: At this point, I don't think there is a need to use different names. There is no other errors that we could mix this up with.
```suggestion
	if err = r.Client.Create(info.ctx, hd); err != nil {
		return actionError{fmt.Errorf("failed to create hardwareData: %w", err)}
```

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -1502,8 +1512,8 @@ func (r *BareMetalHostReconciler) doServiceIfNeeded(prov provisioner.Provisioner
 		}
 
 		if hfcDirty {
-			if err := r.Status().Update(info.ctx, hfc); err != nil {
-				return actionError{fmt.Errorf("failed to update hostfirmwarecomponents status: %w", err)}
+			if errOnUpdate := r.Status().Update(info.ctx, hfc); errOnUpdate != nil {
+				return actionError{fmt.Errorf("failed to update hostfirmwarecomponents status: %w", errOnUpdate)}
```
lentzi90: This feels inconsistent. We are reusing `err` on line 1493 and 1509. Why not here?
```suggestion
			if err = r.Status().Update(info.ctx, hfc); err != nil {
				return actionError{fmt.Errorf("failed to update hostfirmwarecomponents status: %w", err)}
```

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -1576,10 +1588,10 @@ func (r *BareMetalHostReconciler) manageHostPower(prov provisioner.Provisioner,
 
 	servicingAllowed := isProvisioned && !info.host.Status.PoweredOn && desiredPowerOnState
 	if servicingAllowed || info.host.Status.OperationalStatus == metal3api.OperationalStatusServicing || info.host.Status.ErrorType == metal3api.ServicingError {
-		hup, err := r.acquireHostUpdatePolicy(info)
-		if err != nil {
+		hup, errOnAcquire := r.acquireHostUpdatePolicy(info)
+		if errOnAcquire != nil {
```
lentzi90: I would rather declare `hup` and reuse `err`. It is one extra line, but "less surprising" at least to me. If the error get a different name I start looking for why we need the original err later and things like that.
```suggestion
               var hup *metal3api.HostUpdatePolicy
		hup, err = r.acquireHostUpdatePolicy(info)
		if err != nil {
```

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -1808,8 +1820,8 @@ func (r *BareMetalHostReconciler) attachDataImage(prov provisioner.Provisioner,
 		dataImage.Status.Error.Count++
 		dataImage.Status.Error.Message = err.Error()
 		// Error updating DataImage Status
-		if err := r.Status().Update(info.ctx, dataImage); err != nil {
-			return fmt.Errorf("failed to update DataImage status, %w", err)
+		if errOnUpdate := r.Status().Update(info.ctx, dataImage); errOnUpdate != nil {
+			return fmt.Errorf("failed to update DataImage status, %w", errOnUpdate)
```
lentzi90, line 1813: This is good! :+1:  Here we are using the `err` after so we were really shadowing it before. Now it is much clearer that it is a different error here in between. Also same below :+1:

main.go
```diff
@@ -356,8 +363,8 @@ func main() {
 	}
 
 	setupLog.Info("starting manager")
-	if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
-		setupLog.Error(err, "problem running manager")
+	if errOnStart := mgr.Start(ctrl.SetupSignalHandler()); errOnStart != nil {
+		setupLog.Error(errOnStart, "problem running manager")
```
lentzi90:
```suggestion
	if err = mgr.Start(ctrl.SetupSignalHandler()); err != nil {
		setupLog.Error(err, "problem running manager")
```

internal/controller/metal3.io/baremetalhost_controller.go
sirpaleet: Okay, that was my original thought. I was just wondering whether to change back and overwrite because these checks specifically test against nil and always return if the value is anything else. And the original value was checked against nil before to only do this loop err==nil, too. So checking against the value would not really happen here since we always know it is nil. I was just wondering with these "clear-cut" cases whether moving a nil value on top of known nil etc would be okay?

internal/controller/metal3.io/baremetalhost_controller.go
sirpaleet: My logic here was that the check roes not return right after, which it does on the line 1493, which is why I considered that "safe" to overwrite. Line 1509 was previously overwritten so I did not change that.

internal/controller/metal3.io/baremetalhost_controller.go
lentzi90: Ah ok! Yes this is ok and an expected way of working. We just save the value from each check. There is no danger or controversy to overwrite the previous `nil`

internal/controller/metal3.io/baremetalhost_controller.go
lentzi90: Hmm I am not seeing it :thinking: 
We return on both 1516 and 1497, and both were previously shadowing now the first is reusing `err` and the second got a new name

sirpaleet: I will change back to err here too then :+1:

internal/controller/metal3.io/baremetalhost_controller.go
sirpaleet: For this the linter does not flag as shadowing even with :=, so I am actually leaving it as original

internal/controller/metal3.io/baremetalhost_controller.go
sirpaleet: This is also not flagged as shadowing for some reason, if i use err :=, wonder why. Leaving as original.

main.go
sirpaleet: This seems to also not be flagged even if it seems like shadowing, changing back to original

sirpaleet:
I am limiting my commit's scope to only include directly shadowing related changes and the previously discussed changes, pushing my pr again in a bit, and will add some comments.

hack/tools/deploy-cli/deploy-cli.go
```diff
@@ -119,15 +119,14 @@ func (d *DeployContext) determineIronicAuth() error {
...
-	var err error
 	ironicDataDir, err := d.GetEnvOrDefault("IRONIC_DATA_DIR")
 	if err != nil {
 		return err
 	}
 	configMapFileName := "ironic_bmo_configmap.env"
 	configMapInDataDir := filepath.Join(ironicDataDir, configMapFileName)
 
-	if _, err := os.Stat(configMapInDataDir); err == nil && d.IronicBMOConfigMapEnvFile == "" {
+	if _, errOnPath := os.Stat(configMapInDataDir); errOnPath == nil && d.IronicBMOConfigMapEnvFile == "" {
```
sirpaleet: changing variable name to avoid shadowing and moving non-nil err value

hack/tools/deploy-cli/deploy-cli.go
```diff
@@ -200,7 +200,7 @@ func (d *DeployContext) deployBMO() error {
 	ironicEnvFileName := "ironic.env"
 	ironicEnvInDataDir := filepath.Join(ironicDataDir, ironicEnvFileName)
 
-	if _, err := os.Stat(ironicEnvInDataDir); err == nil && d.IronicEnvFile == "" {
+	if _, errOnPath := os.Stat(ironicEnvInDataDir); errOnPath == nil && d.IronicEnvFile == "" {
```
sirpaleet: changing variable name to avoid shadowing and moving non-nil err value

test/e2e/e2e_suite_test.go
```diff
@@ -165,8 +166,8 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 
 	scheme := runtime.NewScheme()
 	framework.TryAddDefaultSchemes(scheme)
-	clusterProxy := framework.NewClusterProxy("bmo-e2e", kubeconfigPath, scheme)
-	Expect(clusterProxy).ToNot(BeNil(), "Failed to get a cluster proxy")
+	clusterProxy = framework.NewClusterProxy("bmo-e2e", kubeconfigPath, scheme)
+	Expect(clusterProxy).ToNot(BeNil(), "Failed to get a cluster clusterProxy")
```
lentzi90: nit: This line did not need to change. Probably changed just because of renaming clusterProxy -> proxy -> clusterProxy.

hack/tools/deploy-cli/deploy-cli.go
lentzi90: I think we should reuse `err` here. If `err` was non-nil, we return before we get here.
```suggestion
	if _, err = os.Stat(ironicEnvInDataDir); err == nil && d.IronicEnvFile == "" {
```

hack/tools/deploy-cli/deploy-cli.go
lentzi90: I would prefer to keep the `err` declaration on top and reuse here without the `:`. If `err` was non-nil we return before we get here.

lentzi90 (commented):
Great job!
Just two suggestions and a nit below

test/e2e/e2e_suite_test.go
sirpaleet: The original reason why I first changed this was because of shadowing, and then after that I changed it back to the original name. Adding : would shadow declaration on line 56 here. I think we discussed in a previous comment that just overwrite would be better here than changing the name.

lentzi90: Sorry it was unclear, I mean line 170. It has bit strange wording now "Failed to get a cluster clusterProxy".

sirpaleet: Aaa I see, did not even notice that. Thanks! Will edit now.

Rozzii (commented):
/lgtm

PR #2528, opened 2025-06-17
🌱 Migrate to golangci-lint v2

ldez:

Migrates to golangci-lint v2.

The configuration has been migrated with command line:

<details>

```console
$ golangci-lint migrate
WARN The configuration comments are not migrated.
WARN Details about the migration: https://golangci-lint.run/product/migration-guide/
╭───────────────────────────────────────────────────────────────────────────╮
│                                                                           │
│                               We need you!                                │
│                                                                           │
│ Donations help fund the ongoing development and maintenance of this tool. │
│  If golangci-lint has been useful to you, please consider contributing.   │
│                                                                           │
│                  Donate now: https://donate.golangci.org                  │
│                                                                           │
╰───────────────────────────────────────────────────────────────────────────╯
```

</details>

And then updated to restore comments, and limiting new changes.


.golangci.yaml
```diff
@@ -1,207 +1,227 @@
+version: "2"
 run:
   go: "1.24"
   build-tags:
-  - e2e
-  - vbmctl
+    - e2e
```
tuminoid: We use k8s style indents. Please don't reindent the whole file.

ldez: The file is not reindented.

The v2 configuration structure is just different than v1.

ldez: The file is not reindented: the v2 configuration structure differs from v1.

But I updated the file to unindent the arrays.

.golangci.yaml
```diff
@@ -63,149 +60,172 @@ linters:
...
+  enable:
+  - gci
+  - gofmt
+  - goimports
+  exclusions:
+    paths:
+    - zz_generated.*\.go$
+    - .*conversion.*\.go$
+    - mock*
+    - _test\.go
```
lentzi90: This `_test\.go` seems new. Is it so that `_test` files were excluded from formatters by default before and now we need to explicitly exclude them? Or maybe I am missing something

ldez: This is not new; this comes from the previous configuration.

`_test` was excluded from `gci` formatter.

https://github.com/metal3-io/baremetal-operator/blob/25236a322c8477c026c6c8f65c5e11c8edd7e583/.golangci.yaml#L147-L152


The exclusions configuration is different between linters and formatters because of the nature of the process.

But I can remove it.

ldez:
Note: I removed `goimports` because `gci` is doing the same job (and internally `gci` uses the `goimports` code).

lentzi90 (commented):
Thanks, this looks good to me!
Could you please squash the commits? We try to avoid multiple commits per PR

.github/workflows/golangci-lint.yml
```diff
@@ -34,8 +34,7 @@ jobs:
       with:
         go-version: ${{ steps.vars.outputs.go_version }}
     - name: golangci-lint-${{matrix.working-directory}}
-      uses: golangci/golangci-lint-action@55c2c1448f86e01eaae002a5a3a9624417608d84 # v6.5.2
+      uses: golangci/golangci-lint-action@4afd733a84b1f43292c63897423277bb7f4313a9 # v8.0.0
       with:
-        version: v1.64.7
+        version: v2.1
```
tuminoid:
```suggestion
        version: v2.1.6
```
Let's pin the latest

.golangci.yaml
```diff
@@ -63,154 +60,175 @@ linters:
...
+      path: .*(api|types)\/.*\/conversion.*\.go$
+      text: 'ST1016: methods on the same type should have the same receiver name'
+    - linters:
+      - tagliatelle
+      text: RAID|MAC|BMO
+    paths:
+    - zz_generated.*\.go$
+    - .*conversion.*\.go$
+    - mock*
+
```
tuminoid, line 235: This is extra linefeed,  or alternatively, add linefeed on L226.

tuminoid:
Asking Copilot review for this, curious.

.github/workflows/golangci-lint.yml
```diff
@@ -34,8 +34,7 @@ jobs:
...
         working-directory: ${{matrix.working-directory}}
-        args: --timeout=10m
```
tuminoid, line 41: Timeout is removed, bot is correct. Is there reason why we can't have it here or in config file?

tuminoid:
> Comments suppressed due to low confidence (2)
>
> **.golangci.yaml:223**
> 
> * There are two `issues:` sections at the top level of this YAML, causing the first block to be overridden. Merge or rename one of them so all settings are applied.
> 
> ```
> issues:
> ```

This is false.

> **.github/workflows/golangci-lint.yml:39**
> 
> * [nitpick] The `args: --timeout=10m` was removed; without it, the linter may use a shorter default timeout and time out on large codebases.

Bot is correct, is timeout not supported for the action nor the config file?

.github/workflows/golangci-lint.yml
ldez: I will ask you a question: Do you know (not guess) why the flag was set?

<details><summary>Answer to the previous question before reading this part</summary>

In golangci-lint v1, the default timeout was too short (1 minute), but for most large projects, the first run inside a CI takes longer than 1 minute.

So, everybody started to add `--timeout` (or the equivalent inside the configuration file), and so this has been propagated to a lot of projects, but without the meaning behind this element.

In golangci-lint v2, there is no default timeout, and there are internal improvements to avoid the need for the timeout.

In summary, your project doesn't need to set a timeout inside the CI.

</details>

tuminoid: Yes, we had cases where we needed the timeout as it got stuck until GH terminated workflow. Linter timeout is a safety net against this, and there is no harm having it.

ldez: I guess that the previous problems you encountered were related to memory because of the new Go version; this has been fixed in v2. And the timeout did not affect this; in this case, the timeout was ignored.

During golangci-lint v2 development, I kept the timeout option only for rare edge cases.
If you have a concrete example where the timeout was effectively used, I'm interested in.

tuminoid: It has been a while, so I'm not going to do gravedigging. Let's assume the issue has been solved and I'll let you know if we hit some timeout issues with v2.

tuminoid:
> /approve

Bot missed this, please re-add.

tuminoid (commented):
Thanks for contributing this.

/lgtm

PR #2529, opened 2025-06-18
:sparkles: Add namespace scoped mode

Sunnatillo:
This PR introduces namespace scoped mode for BMO. We can specify namespaces to watch for BMO by "WATCH_NAMESPACE" variable. 

- Makes main branch  e2e tests to run in namespaced scope.

Please read "namespace-scoped-mode.md" document before reviewing. 

Fixes #


kashifest:
@tuminoid @lentzi90 @Rozzii

tuminoid:
/cc @tuminoid 
I'll be taking a look shortly.

config/overlays/namespaced/kustomization.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+resources:
+- ../../base
+# The subjects and roleRef needs to be update here otherwise we lose name-prefix
+# This is of course not ideal if the name-prefix changes
+
+    # Add a namespace to watch
+
+
+    # Add a namespace to watch
```
lentzi90:
```suggestion
```

hack/tools/kustomize-namespace-annotator/go.mod
lentzi90, line 1: I think we should reuse hack/tools/go.mod for this.

hack/tools/kustomize-namespace-annotator/main.go
```diff
@@ -0,0 +1,103 @@
...
+				indexToRemove := removeIndices[i-1]
+				items = append(items[:indexToRemove], items[indexToRemove+1:]...)
+			}
+		}
+
+		// Ensure all namespaces are defined, as roles cannot be applied without existing namespaces
+		for _, namespace := range config.Values {
+			namespaceYaml := createNamespaceDefinition(namespace)
+			items = append(items, namespaceYaml)
+		}
```
lentzi90: It would be better to error out in my opinion. The user has probably made a mistake and it could be rally hard to figure out the issue if we transparently tries to fix it. For example, there could be a hard to spot typo in the namespace name. It is also possible that the user expects this namespace to be configured with labels, security policies, etc.

hack/tools/kustomize-namespace-annotator/main.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	p := framework.SimpleProcessor{Config: config, Filter: kio.FilterFunc(fn)}
+	cmd := command.Build(p, command.StandaloneDisabled, false)
+	command.AddGenerateDockerfile(cmd)
+	if err := cmd.Execute(); err != nil {
+		os.Exit(1)
+	}
+}
+
+func duplicateResourceOverNamespaces(roleConfig *yaml.RNode, config *ValueAnnotator) ([]*yaml.RNode, error) {
```
lentzi90: Please add a doc string that explains what this does

hack/tools/kustomize-namespace-annotator/main.go
```diff
@@ -0,0 +1,103 @@
...
+	"sigs.k8s.io/kustomize/kyaml/fn/framework/command"
+	"sigs.k8s.io/kustomize/kyaml/kio"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+type ValueAnnotator struct {
+	Values []string `yaml:"values" json:"values"`
+}
+
+func main() {
```
lentzi90: Please add some comment to explain what this does

hack/tools/kustomize-namespace-annotator/namespaceAnnotator
lentzi90, line 1: Remove this from git (and maybe add it to .gitignore)

hack/tools/kustomize-namespace-annotator/README.md
```diff
@@ -0,0 +1,29 @@
...
+in `config/overlays/namespaced/roles-ns-annotator.yaml`.
+
+To build NamespaceAnnotator plugin binary, run:
+
+```bash
+go build -build -o NamespaceAnnotator main.go
+```
+
+Once the binary  is ready, you can execute the command for the namespaced
+operator:
```
lentzi90: If I understand correctly, this is mixing container and binary way of executing the plugin. It shows how to build the binary but also talks about the (container?) "tag"?

docs/namespace-scoped-setup.md
```diff
@@ -0,0 +1,325 @@
...
+  kind: Namespace
+  metadata:
+    name: bar
+  ---
+  apiVersion: v1
+  kind: Namespace
+  metadata:
+    name: foo
+  ---
+  apiVersion: v11. **Change/Add Roles for each namespaces:**
```
lentzi90:
```suggestion
  apiVersion: v1
```

docs/namespace-scoped-setup.md
```diff
@@ -0,0 +1,325 @@
...
+
+4.**Change/Add RoleBindings for each namespaces:**
+
+  ```yaml
+  ---
+  apiVersion: rbac.authorization.k8s.io/v1
+  kind: RoleBinding
+  metadata:
+    name: baremetal-operator-manager-rolebinding
+    namespace: bar
```
lentzi90: Is it possible to skip the namespace for the Role and RoleBinding here and just add it through kustomize/kubectl? That was what I had in mind, i.e. we could provide these manifests and they can then easily be applied in any namespace that should be watched, like this:
```bash
kubectl -n foo -f bmo-namespaced-rbac.yaml
kubectl -n bar -f bmo-namespaced-rbac.yaml
kubectl -n baz -f bmo-namespaced-rbac.yaml
```

lentzi90 (commented):
Good progress @Sunnatillo !
I would honestly prefer to do it without the plugin first, just to get a feeling for how the manual way works. The plugin also has the limitation that it is mainly useful during initial setup. I would like to see how to work with the manifests in a dynamic environment.

For this reason I would also like to avoid making the e2e namespaces static. Perhaps it would turn into a mess, but this is the first thing I would like to try. I.e. apply the extra RBAC in the namespace as part of the test and modify BMO to include it. Modifying BMO will probably be the problematic part as that will disrupt the controller once for every test... I have probably asked this before, but is there a way to make the controller select namespaces based on labels instead? That would solve this part neatly!

config/overlays/namespaced/kustomization.yaml
Sunnatillo: Removed now

hack/tools/kustomize-namespace-annotator/main.go
Sunnatillo: Thanks I will address that.

hack/tools/kustomize-namespace-annotator/README.md
Sunnatillo: yes, forgot to align  README to the latest changes.

docs/namespace-scoped-setup.md
Sunnatillo: Yes, it is possible. This would be a manual work. 
By doing this we will not use kustomize plugin.

lentzi90: Good! The plugin is really useful I think, but it is nice to have an easy, manual way of doing it also. Some will prefer it that way for one reason or another :slightly_smiling_face:

docs/namespace-scoped-setup.md
Sunnatillo: done

hack/tools/kustomize-namespace-annotator/README.md
Sunnatillo: done

Makefile
```diff
@@ -225,6 +225,10 @@ manifests-generate: $(CONTROLLER_GEN)
 manifests-kustomize: $(KUSTOMIZE)
 	$< build config/default > config/render/capm3.yaml
 
+.PHONY: manifests-namespaced
+manifests-namespaced: manifests-generate $(KUSTOMIZE)
+	$(KUSTOMIZE) build config/overlays/namespaced --enable-alpha-plugins --enable-exec > config/render/capm3.yaml
```
tuminoid: annotator plugin should not be needed

config/overlays/e2e/kustomization.yaml
```diff
@@ -1,11 +1,12 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
-namespace: baremetal-operator-system
 resources:
 - ../basic-auth_tls
+- namespace-scoped-resources.yaml
```
tuminoid: We don't want massive rbac file, one Role and RoleBinding should be created without namespace, and then applied in loop for affected namespaces.

config/overlays/e2e/kustomization.yaml
```diff
@@ -19,6 +20,22 @@ patches:
   target:
     kind: Deployment
     name: controller-manager
+# Remove ClusterRole (namaspace scoped mode)
```
tuminoid: Make note of the migration plan, 0.12 namespaced is an option, 0.13 clusterroles are the option.

config/overlays/e2e/namespace-scoped-resources.yaml
tuminoid, line 1: we want only one role and one rolebinding here, without namespace, so they can be applied in a loop to each namespace.

config/overlays/e2e/roles-ns-annotator.yaml
tuminoid, line 1: we don't need this

docs/namespace-scoped-setup.md
tuminoid, line 1: update the documentation to the new simpler way of applying this, and also document the limitations and recovery if you screw up with the namespace.

hack/tools/kustomize-namespace-annotator/README.md
tuminoid, line 1: none of this is needed as discussed

tools/deploy.sh
```diff
@@ -223,8 +223,9 @@ if [[ "${DEPLOY_BMO}" == "true" ]]; then
     # This is to keep the current behavior of using the ironic.env file for the configmap
     cp "${SCRIPTDIR}/config/default/ironic.env" "${TEMP_BMO_OVERLAY}/ironic.env"
     ${KUSTOMIZE} edit add configmap ironic --behavior=create --from-env-file=ironic.env
+
     # shellcheck disable=SC2086
-    ${KUSTOMIZE} build "${TEMP_BMO_OVERLAY}" | kubectl apply ${KUBECTL_ARGS} -f -
+    ${KUSTOMIZE} build --enable-alpha-plugins --enable-exec "${TEMP_BMO_OVERLAY}" | kubectl apply ${KUBECTL_ARGS} -f -
```
tuminoid: plugin not needed

tuminoid (commented):
Summarizing as rough review comments what we discussed as a group in the meeting how this could be simplified.

config/overlays/namespaced/namespaced-manager-patch.yaml
```diff
@@ -0,0 +1,13 @@
...
+metadata:
+  name: controller-manager
+  namespace: system
+spec:
+  template:
+    spec:
+      containers:
+      - env:
+        - name: WATCH_NAMESPACE
+          value: basic-ops,external-inspection,externally-provisioned,inspection,live-iso-ops,provisioning-ops,re-inspection,automated-cleaning
```
tuminoid: This should probably be foo,bar,quo

config/overlays/namespaced/kustomization.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- ../../base
+- ns-foo.yaml
```
tuminoid: We're still not getting rid of these copies?

config/overlays/e2e/kustomization.yaml
```diff
@@ -1,11 +1,19 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
-namespace: baremetal-operator-system
 resources:
 - ../basic-auth_tls
+- ./namespaced/automated-cleaning-resources.yaml
```
tuminoid: What is wrong with a loop in e2e script that applies role/rolebinding to all relevant namespaces? We shouldn't force that in kustomize IMO?

@lentzi90 PTAL

Sunnatillo: I could not find the the loop way solution via kustomize. In helm we can do that, but in Kustomize by design declarative and not generative

lentzi90: I think we can do it two ways:
1. Loop in bash or go where you do something like
for ns in test_namespaces; kubectl create ns $ns && kubectl -n $ns apply -k namespaced; done
2. Declaratively define overlays for each and then pull them all in one for applying easily.

I think 2 is what you were sort of going for here, but with only one kustomization. We can reduce the amount of yaml a lot if you do separate overlays for each namespace. Like this:
```
├── basic-ops
│   ├── kustomization.yaml
│   └── namespace.yaml
├── live-iso
│   ├── kustomization.yaml
│   └── namespace.yaml
└── re-inspection
    ├── kustomization.yaml
    └── namespace.yaml
```

In each kustomization.yaml you then put something like this:

```yaml
kind: Kustomization
resources:
- ../../namespaced
- namespace.yaml

namespace: basic-ops
```

Now you don't have the full manifest but instead a kustomization that generates it when you need it. You can then put them all together in one kustomization:

```yaml
kind: Kustomization
resources:
- basic-ops
- live-iso
- re-inspection
- ...
```

Sunnatillo: I have done the changes you requested. But I have placed them inside component so it can be called for other tests when we include them.

test/e2e/common.go
```diff
@@ -218,6 +219,24 @@ func cleanup(ctx context.Context, clusterProxy framework.ClusterProxy, namespace
...
+	if !isNamespaced {
+		framework.DeleteNamespace(ctx, framework.DeleteNamespaceInput{
+			Deleter: clusterProxy.GetClient(),
+			Name:    namespace.Name,
+		})
+		WaitForNamespaceDeleted(ctx, WaitForNamespaceDeletedInput{
+			Getter:    clusterProxy.GetClient(),
+			Namespace: *namespace,
+		}, intervals...)
+	}
```
lentzi90, line 236: This needs an explanation I think. I.e. because of limitations in the controller-runtime, we cannot delete watched namespaces when the controller is running in namespaced mode (if I recall correctly)

test/e2e/inspection_test.go
```diff
@@ -51,6 +68,23 @@ var _ = Describe("Inspection", Label("required", "inspection"), func() {
 			Bmh:    bmh,
 			State:  metal3api.StateUnmanaged,
 		}, e2eConfig.GetIntervals(specName, "wait-unmanaged")...)
+
+		By("Delete BMH")
+		err = clusterProxy.GetClient().Delete(ctx, &bmh)
+		Expect(err).NotTo(HaveOccurred())
+
+		By("Waiting for the BMH to be deleted")
```
lentzi90, line 70: Why do we need to delete the BMH here when we have the Cleanup in AfterEach?

test/e2e/provisioning_and_annotation_test.go
```diff
@@ -253,12 +253,35 @@ var _ = Describe("Provision, detach, recreate from status and deprovision", Labe
 				Bmh:    bmh,
 				State:  metal3api.StateAvailable,
 			}, e2eConfig.GetIntervals(specName, "wait-available")...)
+
+			By("Delete BMH")
+			err = clusterProxy.GetClient().Delete(ctx, &bmh)
+			Expect(err).NotTo(HaveOccurred())
+
+			By("Waiting for the BMH to be deleted")
```
lentzi90, line 260: Here also, why?

test/e2e/re_inspection_test.go
```diff
@@ -107,12 +107,35 @@ var _ = Describe("Re-Inspection", Label("required", "re-inspection"), func() {
 		Expect(clusterProxy.GetClient().Get(ctx, key, &bmh)).To(Succeed())
 		// TODO(lentzi90): Hostname should not be determined or configured through BMC
 		Expect(bmh.Status.HardwareDetails.Hostname).To(Equal(bmc.Name))
+
+		By("Delete BMH")
+		err = clusterProxy.GetClient().Delete(ctx, &bmh)
+		Expect(err).NotTo(HaveOccurred())
+
+		By("Waiting for the BMH to be deleted")
```
lentzi90, line 114: And here

test/e2e/common.go
```diff
@@ -234,7 +253,12 @@ func WaitForBmhInPowerState(ctx context.Context, input WaitForBmhInPowerStateInp
 }
 
 func BuildKustomizeManifest(source string) ([]byte, error) {
-	kustomizer := krusty.MakeKustomizer(krusty.MakeDefaultOptions())
+	kustomizeOptions := krusty.MakeDefaultOptions()
+	// Enable plugins
+	kustomizeOptions.PluginConfig = kustomizetypes.MakePluginConfig(
+		kustomizetypes.PluginRestrictionsNone,
+		kustomizetypes.BploUseStaticallyLinked)
+	kustomizer := krusty.MakeKustomizer(kustomizeOptions)
```
lentzi90: We can remove this, right?

config/overlays/e2e/kustomization.yaml
```diff
@@ -1,11 +1,14 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
-namespace: baremetal-operator-system
 resources:
 - ../basic-auth_tls
 
+components:
+  - ../../components/namespaced
```
lentzi90: I think I understand the idea here now, but I would want to change it a bit. A component is nice if it can be reused, but now it has all the namespaces already hard-coded, which makes it hard to reuse (except for the e2e tests, which I guess was the idea).

Can we make the `namespaced` component generic, so that it only removes the ClusterRole(Bindings)? Then you should be able to use that both here and in `overlays/namespaced`.
Then we need to add the namespaces separately. That will be different for everyone so there is not much point making it a component. For e2e I would put them instead in an overlay and we can include it here and in the other e2e overlays, just as we do with `basic-auth_tls` above.

I hope I am not confusing you with this. The main point is to separate the truly generic part that everyone will need (ClusterRole patch) and the e2e specfic. We can have them as two separate components or overlays, that doesn't matter much to me. The important part is just to avoid mixing them together.

config/overlays/namespaced/kustomization.yaml
```diff
@@ -0,0 +1,24 @@
...
+- ns-bar.yaml
+- ns-quo.yaml
+
+patches:
+# Remove ClusterRoleBinding (namaspace scoped mode)
+- patch: |
+    apiVersion: rbac.authorization.k8s.io/v1
+    kind: ClusterRoleBinding
```
lentzi90: Use the component here so we don't need to add the patch, and reduce the `ns-*.yaml` manifests to just the namespace (no roles)

test/e2e/common.go
Sunnatillo: yes, we cannot delete namespace because of limitation in controller runtime. 
I will add explanation here.

test/e2e/inspection_test.go
Sunnatillo: During testing it keeps failing every time when trying to register bmh with error saying that mac address already has registered with another bmh.

test/e2e/common.go
Sunnatillo: ohh, yes, this is leftover from kustomize plugin implementation.

tools/deploy.sh
Sunnatillo: removed

test/e2e/inspection_test.go
lentzi90: Hmm strange. I don't understand why it would work if we delete here in the test but not in the AfterEach. We may need to dig deeper in this.

test/e2e/common.go
Sunnatillo: Added

test/e2e/inspection_test.go
Sunnatillo: We have discussed this. We run parallel tests. Those tests when reach AfterEach another test will already start, there is a time bmh is not deleted in first test and second test starts and bmh fails to register. 
Thus we decided to keep Delete BMH step in tests.

config/overlays/e2e/kustomization.yaml
Sunnatillo: I have implemented removal of ClusterRole and CLusterRolebinging as a component. Creation of Roles and RoleBindings are put in e2e overlay.

config/overlays/namespaced/kustomization.yaml
Sunnatillo: This is example overlay. This is not used anywhere. 
I left it as it is for users to clearly understand what we are changing. I can remove it as well.

config/overlays/namespaced/namespaced-manager-patch.yaml
Sunnatillo: fixed it now.

docs/namespace-scoped-setup.md
Sunnatillo: I have added recovery proicess and limitations

config/overlays/namespaced/kustomization.yaml
lentzi90: As an overlay, I think it is better to show how to use the component, but we already have the e2e overlay that can work as example to show that. I am fine with keeping a plain manifest example also, but then there isn't really much point in having the kustomization here.
I would suggest just removing this. We can document the whole process in the metal3-docs instead.

test/e2e/automated_cleaning_test.go
```diff
@@ -198,11 +198,28 @@ var _ = Describe("Automated cleaning", Label("required", "automated-cleaning"),
...
 	AfterEach(func() {
 		DumpResources(ctx, e2eConfig, clusterProxy, path.Join(artifactFolder, specName))
 		if !skipCleanup {
-			cleanup(ctx, clusterProxy, namespace, cancelWatches, e2eConfig.GetIntervals("default", "wait-namespace-deleted")...)
+			namespaced := e2eConfig.GetVariable("NAMESPACE_SCOPED")
+			isNamespaced, err := strconv.ParseBool(namespaced)
+			if err != nil {
+				// if error it sets variable to false
+				isNamespaced = false
+			}
```
lentzi90:
```suggestion
			isNamespaced := e2eConfig.GetBoolVariable("NAMESPACE_SCOPED")
```

test/e2e/inspection_test.go
```diff
@@ -26,12 +27,28 @@ var _ = Describe("Inspection", Label("required", "inspection"), func() {
...
 	)
 	BeforeEach(func() {
-		namespace, cancelWatches = framework.CreateNamespaceAndWatchEvents(ctx, framework.CreateNamespaceAndWatchEventsInput{
+
+		namespaced := e2eConfig.GetVariable("NAMESPACE_SCOPED")
+		isNamespaced, err := strconv.ParseBool(namespaced)
+		if err != nil {
+			// if error it sets variable to false
+			isNamespaced = false
+		}
```
lentzi90:
```suggestion
		isNamespaced := e2eConfig.GetBoolVariable("NAMESPACE_SCOPED")
```

test/e2e/inspection_test.go
```diff
@@ -125,12 +164,29 @@ var _ = Describe("Inspection", Label("required", "inspection"), func() {
...
 	AfterEach(func() {
 		DumpResources(ctx, e2eConfig, clusterProxy, path.Join(artifactFolder, specName))
 		if !skipCleanup {
-			cleanup(ctx, clusterProxy, namespace, cancelWatches, e2eConfig.GetIntervals("default", "wait-namespace-deleted")...)
+			namespaced := e2eConfig.GetVariable("NAMESPACE_SCOPED")
+			isNamespaced, err := strconv.ParseBool(namespaced)
+			if err != nil {
+				// if error it sets variable to false
+				isNamespaced = false
+			}
```
lentzi90:
```suggestion
			isNamespaced := e2eConfig.GetBoolVariable("NAMESPACE_SCOPED")
```

test/e2e/live_iso_test.go
```diff
@@ -136,12 +136,29 @@ var _ = Describe("Live-ISO", Label("required", "live-iso"), func() {
...
 	AfterEach(func() {
 		DumpResources(ctx, e2eConfig, clusterProxy, path.Join(artifactFolder, specName))
 		if !skipCleanup {
-			cleanup(ctx, clusterProxy, namespace, cancelWatches, e2eConfig.GetIntervals("default", "wait-namespace-deleted")...)
+			namespaced := e2eConfig.GetVariable("NAMESPACE_SCOPED")
+			isNamespaced, err := strconv.ParseBool(namespaced)
+			if err != nil {
+				// if error it sets variable to false
+				isNamespaced = false
+			}
```
lentzi90:
```suggestion
			isNamespaced := e2eConfig.GetBoolVariable("NAMESPACE_SCOPED")
```

test/e2e/provisioning_and_annotation_test.go
```diff
@@ -253,12 +253,29 @@ var _ = Describe("Provision, detach, recreate from status and deprovision", Labe
...
 		AfterEach(func() {
 			DumpResources(ctx, e2eConfig, clusterProxy, path.Join(artifactFolder, specName))
 			if !skipCleanup {
-				cleanup(ctx, clusterProxy, namespace, cancelWatches, e2eConfig.GetIntervals("default", "wait-namespace-deleted")...)
+				namespaced := e2eConfig.GetVariable("NAMESPACE_SCOPED")
+				isNamespaced, err := strconv.ParseBool(namespaced)
+				if err != nil {
+					// if error it sets variable to false
+					isNamespaced = false
+				}
```
lentzi90:
```suggestion
				isNamespaced := e2eConfig.GetBoolVariable("NAMESPACE_SCOPED")
```

test/e2e/re_inspection_test.go
```diff
@@ -107,12 +107,29 @@ var _ = Describe("Re-Inspection", Label("required", "re-inspection"), func() {
...
 	AfterEach(func() {
 		DumpResources(ctx, e2eConfig, clusterProxy, path.Join(artifactFolder, specName))
 		if !skipCleanup {
-			cleanup(ctx, clusterProxy, namespace, cancelWatches, e2eConfig.GetIntervals("default", "wait-namespace-deleted")...)
+			namespaced := e2eConfig.GetVariable("NAMESPACE_SCOPED")
+			isNamespaced, err := strconv.ParseBool(namespaced)
+			if err != nil {
+				// if error it sets variable to false
+				isNamespaced = false
+			}
```
lentzi90:
```suggestion
			isNamespaced := e2eConfig.GetBoolVariable("NAMESPACE_SCOPED")
```

docs/namespace-scoped-setup.md
```diff
@@ -0,0 +1,206 @@
...
+  BMO will monitor resources across all namespaces in the cluster.
+* **Watch a single namespace:** Set `WATCH_NAMESPACE` to the desired
+  namespace name (e.g., `WATCH_NAMESPACE=my-namespace`).
+* **Watch multiple namespaces:** Provide a comma-separated list of namespaces
+  (e.g., `WATCH_NAMESPACE=foo,bar,quo`).
+
+**Example: Configuring BMO to watch its own deployed namespace:**
+
+To configure the operator to automatically watch the same namespace it's
+deployed in, modify the `config/base/manager.yaml` file. Add the following
```
lentzi90: The example below is watching foo,bar and quo.
```suggestion
To configure the operator to automatically watch the same namespaces foo,
bar and quo, modify the `config/base/manager.yaml` file. Add the following
```

test/e2e/automated_cleaning_test.go
Sunnatillo: Done in all places

docs/namespace-scoped-setup.md
Sunnatillo: Done as well

config/overlays/namespaced/kustomization.yaml
Sunnatillo: Removed  namespaced overlay.

test/e2e/inspection_test.go
lentzi90: Missed this one. The rest is all good now :slightly_smiling_face:

lentzi90:
Can you please check inspection_test.go again. There is one left there

fmuyassarov (approved):
I didn't go through the e2e part and only kustomize related changes and it looks good to me. Good work @Sunnatillo .

Since we are here, I think as a follow up it would be good to revisit the whole structure of the kustomize directory. Because, it seems to me that we are not following the actual structure of the kustomize and there are some bits that can be reused and reduce the duplication and complexity. But of course, it has nothing to do with this PR.

tuminoid:
I'll find time to review this next week, holding meanwhile.
/hold

test/e2e/config/ironic.yaml
```diff
@@ -50,6 +50,8 @@ variables:
   IRONIC_PROVISIONING_IP: "localhost"
   IRONIC_PROVISIONING_PORT: "6385"
 
+  NAMESPACE_SCOPED: "true"
```
Rozzii, line 53: Does this mean that we will be running every regular, e2e in namespaced mode and with specifically looking at 1 namespace?
So we will be not testing either cluster scope watch or watching multiple individual namespaces?
If so, are there plans to test all the watch scope configurations?

lentzi90:
> In general its seems fine, as I understand we are not testing the "watch multiple specific namespaces" mode, I would assume that is a common multi tenant use case so it would be nice to have a tests for that too, or to make the existing tests work with a set of namespaces.
> 
> I might be missing something (quite a lot of files), so if there are tests for watching multiple specific namespaces then ignore my comments. If I am correct, am still not asking to add those tests to this PR, just to have an issue about it so we can do it later.

We actually do set the whole list of namespaces for the e2e tests. It can be seen in https://github.com/metal3-io/baremetal-operator/pull/2529/files#diff-ee1ecc521e8457e4bc906a1682d2d4fa93a9a3be5c3915019b7280c4c79df298
I was asking at some point that can we do it specific for each test, but it would be quite painful since this list cannot be updated without basically rolling out a new pod.

test/e2e/config/ironic.yaml
Sunnatillo: NAMESPACE_SCOPED variable means that bmo will not have clusterwide Clusterrole and Clusterrolebinding, instead it will have roles and rolebindings for each namespaces. 
We currently create Roles and Rolebindings for following namespaces that we use in our e2e tests:
`basic-ops,external-inspection,externally-provisioned,inspection,live-iso-ops,provisioning-ops,re-inspection,automated-cleaning`

When NAMESPACE_SCOPED true BMO watches multiple namespaces and not all.

docs/namespace-scoped-setup.md
```diff
@@ -0,0 +1,206 @@
...
+operator's permissions and scope for security reasons.
+
+## 1. Generating Namespace-Scoped Manifests
+
+To deploy BMO with namespace-specific configurations, you need to generate
+specialized Kubernetes manifests.
+
+**How to generate:**
+
+## 1\. Configuring BMO to Watch Specific Namespaces
```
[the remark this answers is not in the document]
tuminoid: Best practice would be not to have numbers in titles at all, but this nit-pick is correct in a sense that when it renders out it does have two times 1. before it goes to 2.

docs/namespace-scoped-setup.md
```diff
@@ -0,0 +1,206 @@
...
+  template:
+    spec:
+      containers:
+      - name: manager
+        env:
+        - name: WATCH_NAMESPACE
+          value: foo,bar,quo
+```
+
+## 2\. Create Roles and RoleBindings for each namespaces
```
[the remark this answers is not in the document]
tuminoid: Valid typo fix

docs/namespace-scoped-setup.md
```diff
@@ -0,0 +1,206 @@
...
+* **Watch a single namespace:** Set `WATCH_NAMESPACE` to the desired
+  namespace name (e.g., `WATCH_NAMESPACE=my-namespace`).
+* **Watch multiple namespaces:** Provide a comma-separated list of namespaces
+  (e.g., `WATCH_NAMESPACE=foo,bar,quo`).
+
+**Example: Configuring BMO to watch its own deployed namespace:**
+
+To configure the operator to automatically watch the same namespaces foo,
+bar and quo, modify the `config/base/manager.yaml` file. Add the following
+environment variable configuration under spec.template.spec.containers.env`:
```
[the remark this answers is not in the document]
tuminoid: Valid find as well

docs/namespace-scoped-setup.md
```diff
@@ -0,0 +1,206 @@
...
+      - name: manager
+        env:
+        - name: WATCH_NAMESPACE
+          value: foo,bar,quo
+```
+
+## 2\. Create Roles and RoleBindings for each namespaces
+
+When BMO operates in a namespace-scoped mode, you need to create `Roles` and
+`Rolebindings` for each namespaces that you added in WATCH_NAMESPACE in env
```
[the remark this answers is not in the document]
tuminoid: Valid, and I'd say `Role`s and `RoleBinding`s as those are the correct terms.

config/components/namespace-scoped/kustomization.yaml
```diff
@@ -0,0 +1,18 @@
...
+
+patches:
+# Remove ClusterRoleBinding (namaspace scoped mode)
+- patch: |
+    apiVersion: rbac.authorization.k8s.io/v1
+    kind: ClusterRoleBinding
+    metadata:
+      name: baremetal-operator-manager-rolebinding
+    $patch: delete
+# Remove ClusterRole (namaspace scoped mode)
```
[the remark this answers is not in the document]
tuminoid: Typos but in commented out line though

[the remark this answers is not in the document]
tuminoid: Same

tuminoid:
Fix the typos in readme, then let's unhold. There are basically only e2e fixes, and the actual change in the controller is near zero.

docs/namespace-scoped-setup.md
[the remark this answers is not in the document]
Sunnatillo: fixed

docs/namespace-scoped-setup.md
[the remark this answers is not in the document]
Sunnatillo: fixed

docs/namespace-scoped-setup.md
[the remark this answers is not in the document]
Sunnatillo: fixed

docs/namespace-scoped-setup.md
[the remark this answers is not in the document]
Sunnatillo: fixed

config/components/namespace-scoped/kustomization.yaml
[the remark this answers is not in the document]
Sunnatillo: fixed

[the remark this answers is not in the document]
Sunnatillo: fixed

Sunnatillo:
> Fix the typos in readme, then let's unhold. There are basically only e2e fixes, and the actual change in the controller is near zero.

Addressed the comments. if LGTY we can merge this

Sunnatillo:
/retest 
Live ISO test is broken in CI. Not related to the changes

lentzi90:
I would prefer not to override for large changes like this. We risk breaking the CI more that way.
We have reverted the PR that broke live-ISO so the tests should now pass.
/retest

lentzi90:
I think the test is failing because an inspection test was added in https://github.com/metal3-io/baremetal-operator/pull/2670. You will need to adapt

Rozzii (commented):
/lgtm

lentzi90:
Thanks, now it looks good!
I had one more idea, but let's do that separately. Just explaining here what I noticed. We now set the WATCH_NAMESPACE environment variable through a separate kustomize patch. This is a bit strange when we already have a configmap that we put all environment variables in. We should move it there to keep things consistent.
But as I said, let's not delay this anymore. We can do it separately

/hold cancel

tuminoid (commented):
/lgtm
/retitle :sparkles: Add namespace scoped mode

Let's merge this, but prepare to revert it if something unexpectedly breaks.

Issue #2100, opened 2024-12-05
Stop requiring DEPLOY_KERNEL/DEPLOY_RAMDISK


dtantsur:
With https://github.com/metal3-io/ironic-image/pull/452 in place for some time, Ironic has reasonable defaults for these values when paired with the default IPA-downloader. Not requiring them makes BMO installation simpler.

/kind feature

dtantsur:
/lifecycle frozen

This is a pretty important installation UX improvement.

PR #2537, opened 2025-06-24
🌱 Refactor setting various Ironic properties

dtantsur:
Currently, Ironic instance_info and properties fields are populated at
random either in most states or before deployment. While potentially
convenient, it makes it very hard to reason about the code.

Now, the logic is split into two parts:
1. configureNode (renamed from configureImages) writes fields that are
   considered properties of the node itself: CPU architecture, deploy
   images, capabilities, etc.
2. getInstanceUpdateOpts (merge of getImageUpdateOptsForNode and
   getUpdateOptsForNode) writes fields that are required for deployment
   and thus are properties of instance. This includes images, checksums,
   runtime capabilities. As an exception, root device hints fall under
   this category and thus are now set in instance_info, not properties.

Signed-off-by: Dmitry Tantsur <dtantsur@protonmail.com>

PR #2538, opened 2025-06-26
:sparkles:  Allow configuring different IPA images per architecture

diconico07:
This PR allows one to work with BareMetalHosts of different architectures while using kernel+ramdisk for preprovisioning.

It does this by adding new environment variables to configure the different URLs per target CPU architecture:
 - `DEPLOY_KERNEL_URL_<ARCH>`
 - `DEPLOY_RAMDISK_URL_<ARCH>`
 - `DEPLOY_ISO_URL_<ARCH>`
 - `DEPLOY_BOOTLOADER_URL_<ARCH>`

Non suffixed variables are used as defaults, if there is no architecture specific image(s) defined for the BMH CPU architecture.

It also adds management of the Ironic bootloader option in the operator, so that all elements used by Ironic to build the preprovisioning image are coming from the operator. This goes with the `DEPLOY_BOOTLOADER_URL` variable.

**Note for reviewers:**

The list of "supported" architecture is embedded here, so it will currently only checks for `x86_64` and `aarch64`, while we could potentially find arbitrary architectures by parsing the entire `os.Environ()` slice to lift that limitation I am not sure it would be relevant for now, and would make the code harder to read for that PR in my opinion.

### DIFF
--- a/internal/controller/metal3.io/baremetalhost_controller.go
+++ b/internal/controller/metal3.io/baremetalhost_controller.go
@@ -847,6 +847,7 @@ func (r *BareMetalHostReconciler) registerHost(prov provisioner.Provisioner, inf
 			PreprovisioningNetworkData: preprovisioningNetworkData,
 			HasCustomDeploy:            hasCustomDeploy(info.host),
 			DisablePowerOff:            info.host.Spec.DisablePowerOff,
+			CPUArchitecture:            getHostArchitecture(info.host),
 		},
 		credsChanged,
 		info.host.Status.ErrorType == metal3api.RegistrationError)
--- a/pkg/imageprovider/imageprovider.go
+++ b/pkg/imageprovider/imageprovider.go
@@ -20,6 +20,7 @@ type ImageData struct {
 type GeneratedImage struct {
 	ImageURL          string
 	KernelURL         string
+	BootloaderURL     string
 	ExtraKernelParams string
 }
 
--- a/pkg/provisioner/ironic/factory.go
+++ b/pkg/provisioner/ironic/factory.go
@@ -58,9 +58,10 @@ func (f *ironicProvisionerFactory) init(havePreprovImgBuilder bool) error {
 	f.log.Info("ironic settings",
 		"endpoint", ironicEndpoint,
 		"ironicAuthType", ironicAuth.Type,
-		"deployKernelURL", f.config.deployKernelURL,
-		"deployRamdiskURL", f.config.deployRamdiskURL,
-		"deployISOURL", f.config.deployISOURL,
+		"defaultDeployKernelURL", f.config.defaultDeployConfig.kernelURL,
+		"defaultDeployRamdiskURL", f.config.defaultDeployConfig.ramdiskURL,
+		"defaultDeployISOURL", f.config.defaultDeployConfig.ISOURL,
+		"defaultDeployBootloaderURL", f.config.defaultDeployConfig.bootloaderURL,
 		"liveISOForcePersistentBootDevice", f.config.liveISOForcePersistentBootDevice,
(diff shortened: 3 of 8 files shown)


lentzi90:
Some of the test failures seem to be caused by this PR. We did have unrelated issues yesterday though that also caused the e2e tests to fail. Could you rebase to get the fix for that? Then it will be clearer if there are issues with these changes

diconico07:
Yes, I did a mistake while refactoring before pushing, fixed it.

lentzi90:
Thanks!
The fixture test looks like a potential flake.
/retest

lentzi90:
I won't have time to review properly before going on vacation, but I want to say thank you for working on this!
It looks really useful and thought through. Hopefully there will be others around to review soon!

I wonder if we would be able to include ARM64 machines for testing in the e2e suite? :thinking: That would be really cool!
Not necessarily something that has to happen as part of this PR though.

diconico07:
Wanted to create the PR before going on vacation as well :smile: .

It would be nice to have multiple architectures in the e2e suite, but that would probably need some additional work on the ipa downloader image first, didn't do anything there yet as I'm using a custom one.
I didn't look into the e2e test suite inner workings, so I'm not sure of what would be needed to add ARM64 machine here.

Rozzii:
> Wanted to create the PR before going on vacation as well 😄 .
> 
> It would be nice to have multiple architectures in the e2e suite, but that would probably need some additional work on the ipa downloader image first, didn't do anything there yet as I'm using a custom one. I didn't look into the e2e test suite inner workings, so I'm not sure of what would be needed to add ARM64 machine here.

We have quota to run CI on ARM machines donated to CNCF by Oracle so that should be no issue. 

This is quite a considerable change though, there have been some work done on the ironic-image to make it arm friendly but this feature is a bit out of the blue for me, just before most of the contributors go on vacation.

I have nothing but support for making our stack work with ARM but I would like to have a bit wider discussion about this implementation. I need some time to think this approach through.  I will certainly bring this topic up on the next community meeting.

dtantsur:
/hold

I'm sorry, I see that this PR is pretty advanced already, but I think it's attacking the problem from the wrong end. We really need to move away from BMO accepting a ton of options that are blindly passed to Ironic. There is a tracking bug for that: https://github.com/metal3-io/baremetal-operator/issues/2100.

The actual place to pass these variables is ironic-image. https://github.com/metal3-io/ironic-image/pull/452 has started some work in this direction, although only for the downloader. I think we need these steps:

1. Fix #2100 and gradually phase out passing deploy kernel/initramfs/iso through BMO, except for the PreprovisioningImage code path.
2. Update IPA-downloader to link downloaded images with an arch-specific name (if not already - need checking)
3. Update ironic-image to detect downloaded images for several architectures (similarly to https://github.com/metal3-io/ironic-image/pull/452) and set the `*_per_arch` group of options (like this one  https://docs.openstack.org/ironic/latest/configuration/config.html#conductor.deploy_kernel_by_arch).
4. Update ironic-image to optionally accept the parameters you created for BMO in this PR.

I'm happy to help if this all is overwhelming (and since you're vacationing). I can probably take #2100 after I finish with #2558.

dtantsur:
I've started with a couple of items for #2100: https://github.com/metal3-io/baremetal-operator/pull/2559 and https://github.com/metal3-io/ironic-image/pull/709

diconico07:
@dtantsur no problem, I made this MR from a quick PoC to check the approach would work, I missed #2100 so I went the wrong direction.

diconico07:
Closing this one then, #2537 will ensure Ironic has the `cpu_arch` filled, so now everything should be in Ironic image and IPA downloader.
Will probably start working on that next week

PR #2539, opened 2025-06-26
🌱 E2E: Expose more BMC related fields

lentzi90:

This makes it possible to specify the rootDeviceHints in the "BMCs" config file for e2e tests. It is necessary when running the tests against real hardware that may have different disk devices.
It also adds bootMode and disableCertificateVerification that may be
needed depending on the hardware.

Fixes #

### DIFF
--- a/test/e2e/bmc.go
+++ b/test/e2e/bmc.go
@@ -6,6 +6,7 @@ package e2e
 import (
 	"os"
 
+	metal3api "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
 	"gopkg.in/yaml.v2"
 )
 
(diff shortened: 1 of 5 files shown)


lentzi90:
Failure does not seem to be related. It failed because the BMH went to deprovisioning despite the detach annotation. We have a sleep in place from before to workaround this. I suspect it was just a flake.
/retest

lentzi90:
/hold
I have realized that it may be better to properly expose the whole bmc field.

lentzi90:
/hold cancel

I have added all the relevant fields now I think. I realized there wasn't really much under `bmh.spec.bmc` but there were some other fields that may be needed.

kashifest (commented):
/approve

Issue #1729, opened 2024-05-17
Detach annotation does not apply immediately


lentzi90:
**What steps did you take and what happened:**

In the e2e test where we test the detach annotation, we had to [add a sleep](https://github.com/metal3-io/baremetal-operator/blob/e300eeddf594cf68a9491e29c89b52d5d94011e4/test/e2e/provisioning_and_annotation_test.go#L154-L160) before attempting to delete the BareMetalHost. Otherwise it would still go into deprovisioning.

**What did you expect to happen:**

As soon as the detach annotation is set, the BareMetalHost should be possible to delete without causing any change to its state.

**Anything else you would like to add:**

Here is what it looks like when the test fails because the BMH goes to deprovisioning:

```
[2024-05-17T03:11:49.176Z]   STEP: Waiting for the BMH to become provisioned @ 05/17/24 03:08:16.862
[2024-05-17T03:11:49.176Z]   STEP: Verifying the node booting from disk @ 05/17/24 03:10:57.201
[2024-05-17T03:11:49.176Z]   INFO: System is booted from a disk.
[2024-05-17T03:11:49.176Z]   STEP: Retrieving the latest BMH object @ 05/17/24 03:11:12.581
[2024-05-17T03:11:49.176Z]   STEP: Adding the detached annotation @ 05/17/24 03:11:12.619
[2024-05-17T03:11:49.176Z]   STEP: Saving the status to a JSON string @ 05/17/24 03:11:12.969
[2024-05-17T03:11:49.176Z]   STEP: Deleting the BMH @ 05/17/24 03:11:12.97
[2024-05-17T03:11:49.176Z]   STEP: Waiting for the BMH to be deleted @ 05/17/24 03:11:15.243
[2024-05-17T03:11:49.176Z]   [FAILED] in [It] - /home/metal3ci/workspace/metal3-bmo-e2e-test-periodic/test/e2e/common.go:101 @ 05/17/24 03:11:16.017
[2024-05-17T03:11:49.176Z]   INFO: Deleting namespace provisioning-ops-5sx97e
[2024-05-17T03:11:49.176Z]   << Timeline
[2024-05-17T03:11:49.176Z] 
[2024-05-17T03:11:49.176Z]   [FAILED] Told to stop trying after 0.771s.
[2024-05-17T03:11:49.176Z]   BMH provisioning-ops in namespace provisioning-ops-5sx97e should be deleted
[...]

PR #2541, opened 2025-06-26
🌱 e2e: wait for detachment to actually happen

dtantsur:
Sleeping 2 seconds may not be enough, and it's also not what we document
the users should do. Wait for operationalStatus to become detached.

Fixes: #1729 


lentzi90 (commented):
/approve

PR #2551, opened 2025-06-30
:book: update docs/releasing.md on dependabot configs

tuminoid:
Update docs/releasing.md on dependabot config changes while releasing new minors.

### DIFF
--- a/docs/releasing.md
+++ b/docs/releasing.md
@@ -146,6 +146,23 @@ BMO e2e is running as GitHub Actions. We need to add released branch and remove
 the non-maintained branches there, along with the suitable configurations with
 recently released Ironic-image releases as well in the fixtures.
 
+### Dependabot configuration
+
+In `main` branch, we need to update Dependabot configuration to allow updates
+to release branch dependencies and GitHub Workflows.
+
+If project dependencies or modules have not changed, we copy the previous release
+branch configuration and amend the `target-branch` to point to our new release
+branch. We also remove the EOL release branch at the same time, as updating
+`dependabot.yml` causes Dependabot to run the rules, ignoring the configured
+schedule.
+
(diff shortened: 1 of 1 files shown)


docs/releasing.md
```diff
@@ -146,6 +146,23 @@ BMO e2e is running as GitHub Actions. We need to add released branch and remove
 the non-maintained branches there, along with the suitable configurations with
 recently released Ironic-image releases as well in the fixtures.
 
+### Dependabot configuration
+
+In `main` branch, we need to update Dependabot configuration to allow updates
```
Rozzii: I would use either "you" as or "one, the release team etc..." instead of the "we". This "we" was introduced in the "Release artifacts" section and doesn't really fit into a technical document or to a instruction guide.

Rozzii (commented):
/cancel approve
please remove this "we" perspective from the text, IMO it does not fit this technical document and it is also inconsistently used.

docs/releasing.md
tuminoid: Done.

docs/releasing.md
```diff
@@ -146,6 +146,24 @@ BMO e2e is running as GitHub Actions. We need to add released branch and remove
 the non-maintained branches there, along with the suitable configurations with
 recently released Ironic-image releases as well in the fixtures.
 
+### Dependabot configuration
+
+In `main` branch, Dependabot configuration must be amended to allow updates
+to release branch dependencies and GitHub Workflows.
+
+If project dependencies or modules have not changed, previous release branch
+configuration can be copied and amend the `target-branch` to point to our new
```
dtantsur, line 155: "and amend" no longer passes grammatically, maybe use ", amending"?

Issue #2525, opened 2025-06-13
No handling of validation errors for image (e.g. missing checksum)


dtantsur:
For example, if you set `image.url` without a checksum, the validation will pass until Ironic fails with

```
  errorMessage: 'host validation error: Node 2a6456cd-f6cf-4927-b68e-82579cdb4ee7
    failed to validate deploy image info. Some parameters were missing. Missing are:
    [''instance_info.image_source'']'
```

I think that's because `setDirectDeployUpdateOptsForNode` does nothing when `imageData.GetChecksum()` returns `!ok`.

I'd expect the validation to happen somewhere earlier, but the only other place where `GetChecksum` is used is `ironicHasSameImage`, and it also does not have any error handling.

We need a much better error message. Image source is provided by the user, the checksum isn't.

/kind bug

PR #2558, opened 2025-07-02
🐛 Provide an error when required checksum is missing

dtantsur:
Otherwise, we end up producing an invalid InstanceInfo, and Ironic
rejects it with a confusing message.

Change GetChecksum to return an actual error instead of just ok.

Fixes: #2525 


dtantsur:
/hold 

Needs to be done on the webhook level as well

PR #2559, opened 2025-07-02
✨ Stop requiring DEPLOY_KERNEL/RAMDISK

dtantsur:
Ironic has global configuration that allows specifying them, even
depending on the architecture. Our ironic-image supports that when
IPA downloader is used (and should start supporting explicit variables
too).

Fixes: #2100 


dtantsur:
/retest

Gateway timeout from quay

PR #2566, opened 2025-07-16
:warning:  Enable removal of ExternallyProvisioned attribute from BareMetalHosts

### LINKED ISSUE #2465: Confusing behavior on removing `externallyProvisioned: true`

ethan-gallant:
It enables removing the ExternallyProvisioned attribute, this can be useful if you want to transition hosts from an externally provisioned source into being directly managed.

Fixes #2465

Signed-off-by: Ethan J. Gallant <ethan.gallant@gmail.com>

### DIFF
--- a/internal/controller/metal3.io/host_state_machine.go
+++ b/internal/controller/metal3.io/host_state_machine.go
@@ -408,6 +408,9 @@ func (hsm *hostStateMachine) handleRegistering(_ *reconcileInfo) actionResult {
 	// if the credentials change and the Host must be re-registered.
 	if hsm.Host.Spec.ExternallyProvisioned {
 		hsm.NextState = metal3api.StateExternallyProvisioned
+	} else if hsm.Host.Status.Provisioning.State == metal3api.StateExternallyProvisioned {
+		// Removing externallyManaged moves hosts to provisioned state
+		hsm.NextState = metal3api.StateProvisioned
 	} else if inspectionDisabled(hsm.Host) {
 		hsm.NextState = metal3api.StatePreparing
 	} else {
--- a/internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
+++ b/internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
@@ -109,10 +109,6 @@ func (webhook *BareMetalHost) validateChanges(oldObj *metal3api.BareMetalHost, n
 		errs = append(errs, errors.New("bootMACAddress can not be changed once it is set"))
 	}
 
-	if oldObj.Spec.ExternallyProvisioned != newObj.Spec.ExternallyProvisioned {
-		errs = append(errs, errors.New("externallyProvisioned can not be changed"))
-	}
-
 	return errs
 }
 
--- a/test/e2e/externally_provisioned_test.go
+++ b/test/e2e/externally_provisioned_test.go
@@ -18,6 +18,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/cluster-api/test/framework"
 	"sigs.k8s.io/cluster-api/util"
+	kclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 var _ = Describe("Create as externally provisioned, deprovision", Label("required", "provision", "deprovision"),
@@ -79,14 +80,28 @@ var _ = Describe("Create as externally provisioned, deprovision", Label("require
 			}, &bmh)
 			Expect(err).NotTo(HaveOccurred())
 
-			By("checking that the BMH was not inspected or deployed")
+			By("Checking that the BMH was not inspected or deployed")
 			Expect(bmh.Status.OperationHistory.Inspect.Start.IsZero()).To(BeTrue())
 			Expect(bmh.Status.OperationHistory.Provision.Start.IsZero()).To(BeTrue())
 
+			By("Removing the ExternallyProvisioned field")
+			bmh.Spec.ExternallyProvisioned = false
+
+			err = clusterProxy.GetClient().Update(ctx, &bmh)
+			Expect(err).NotTo(HaveOccurred())
+
+			Eventually(func(g Gomega) {
+				err = clusterProxy.GetClient().Get(ctx, kclient.ObjectKeyFromObject(&bmh), &bmh)
+				g.Expect(bmh.Status.Provisioning.State).To(Equal(metal3api.StateProvisioned))
+			}, 2*time.Second, 100*time.Millisecond).Should(Succeed())
+
 			By("Deleting the BMH")
-			// Wait for 2 seconds to allow time to confirm annotation is set
-			// TODO: fix this so we do not need the sleep
-			time.Sleep(2 * time.Second)
+			// Wait for the finalizer to be there, otherwise cleanup logic can't hold the resource
+			Eventually(func(g Gomega) {
+				err = clusterProxy.GetClient().Get(ctx, kclient.ObjectKeyFromObject(&bmh), &bmh)
+				g.Expect(err).NotTo(HaveOccurred())
+				g.Expect(bmh.Finalizers).To(ContainElement(metal3api.BareMetalHostFinalizer))
+			}, 2*time.Second, 100*time.Millisecond).Should(Succeed())
 
 			err = clusterProxy.GetClient().Delete(ctx, &bmh)
 			Expect(err).NotTo(HaveOccurred())
--- a/internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation_test.go
+++ b/internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation_test.go
@@ -970,14 +970,6 @@ func TestValidateUpdate(t *testing.T) {
 				TypeMeta: tm, ObjectMeta: om, Spec: metal3api.BareMetalHostSpec{BootMACAddress: "test-mac"}},
 			wantedErr: "bootMACAddress can not be changed once it is set",
 		},
-		{
-			name: "updateExternallyProvisioned",
-			newBMH: &metal3api.BareMetalHost{
-				TypeMeta: tm, ObjectMeta: om, Spec: metal3api.BareMetalHostSpec{}},
-			oldBMH: &metal3api.BareMetalHost{
-				TypeMeta: tm, ObjectMeta: om, Spec: metal3api.BareMetalHostSpec{ExternallyProvisioned: true}},
-			wantedErr: "externallyProvisioned can not be changed",
-		},
 	}
 
 	for _, tt := range tests {


internal/controller/metal3.io/host_state_machine.go
```diff
@@ -408,6 +408,9 @@ func (hsm *hostStateMachine) handleRegistering(_ *reconcileInfo) actionResult {
 	// if the credentials change and the Host must be re-registered.
 	if hsm.Host.Spec.ExternallyProvisioned {
 		hsm.NextState = metal3api.StateExternallyProvisioned
+	} else if hsm.Host.Status.Provisioning.State == metal3api.StateExternallyProvisioned {
```
Rozzii, line 411: Wouldn't this implementation move all externally provisioned BMH to "regular provisioned" state even if the user would add an unrelated annotation or the reconciliation would be triggered for whatever other reason?

It might be just a theoretical scenario but I would like to have an an additionl condition that checks that the "ExtarnallyProvisioned" filed is really set to false.

Rozzii (commented):
I think this topic would need some community consensus at least a lazy one. I don't remember at the moment why was it decided to block an externally provisioned node to transition to the regular provisioned state.

In principle I don't see why we shouldn't allow this transition but I am not using this feature on a daily basis so I am interested in the opinions of those who do might have more experience with the use-case, they might know what would be the side effects of allowing the proposed state transition.

test/e2e/externally_provisioned_test.go
```diff
@@ -79,14 +80,28 @@ var _ = Describe("Create as externally provisioned, deprovision", Label("require
 			}, &bmh)
 			Expect(err).NotTo(HaveOccurred())
 
-			By("checking that the BMH was not inspected or deployed")
+			By("Checking that the BMH was not inspected or deployed")
 			Expect(bmh.Status.OperationHistory.Inspect.Start.IsZero()).To(BeTrue())
 			Expect(bmh.Status.OperationHistory.Provision.Start.IsZero()).To(BeTrue())
```
Rozzii, line 85: I am not sure these two conditions mean that the BMH is still in registering as they would never be true in case of an externally provisioned BMH at least that would be my expectation. 

I assume you have run this test locally but I wonder how long does the BMH stay in "Registering" state that your implementation relies on.

How do you ensure that the BMH stays in registering state to allow the test code to do the bmh.Spec.ExternallyProvisioned field update? I am not a user of the externally provisioned feature so I might be missing something but I would have suspected the BMH to very quickly transition out of "Registering" and If that is true this test would be very prone to timing issues.

Rozzii:
I have changed this from bug to breaking change, and IMO this is a feature not a bug as blocking the transition between externally and regularly provisioned state was intentional and documented.
/kind feature

internal/controller/metal3.io/host_state_machine.go
ethan-gallant: Since this lands in an else block after the check for externally provisioned I don't believe that would be possible here. 

Happy to reformat it as well into a switch statement or similar as the conditional branches here are getting a bit wild already before the change.

test/e2e/externally_provisioned_test.go
ethan-gallant: I did a manual test however I had some issues getting the e2e tests running locally with `make test-e2e` as the e2e image from quay.io could not be loaded. 

I'll try again today as it seems that the tests did not align with what I saw manually testing. If you have any guidance that I might have missed getting the E2E working locally (on Linux / MacOS) that would be greatly appreciated.

internal/controller/metal3.io/host_state_machine.go
zaneb: I don't believe this does anything at all, because in `handleRegistering()` the `hsm.Host.Status.Provisioning.State` is, [by definition](https://github.com/metal3-io/baremetal-operator/blob/f5aa16bd2770b281603b00067170ec76374e155c/internal/controller/metal3.io/host_state_machine.go#L48), `StateRegistering`.

internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -109,10 +109,6 @@ func (webhook *BareMetalHost) validateChanges(oldObj *metal3api.BareMetalHost, n
 		errs = append(errs, errors.New("bootMACAddress can not be changed once it is set"))
 	}
 
-	if oldObj.Spec.ExternallyProvisioned != newObj.Spec.ExternallyProvisioned {
```
zaneb, line 112: This is a valid bug fix.
According to the state diagram, it is valid to remove the externallyProvisioned flag: https://github.com/metal3-io/baremetal-operator/blob/main/docs/baremetalhost-states.md
I have no idea why the webhook is preventing it.

zaneb (commented):
> this can be useful if you want to transition hosts from an externally provisioned source into being directly managed.

I was going to say that the question is why you need to do that.
An externally provisioned host is still fully managed in terms of power state.
The other thing you can do with a `provisioned` host is deprovision it... but that's what happens when you remove the externallyProvisioned flag anyway.

However, I see that somebody has prevented you from removing the flag in the webhook. That is a real bug in my opinion. Otherwise (e.g. with the webhook disabled) it already works correctly and as-documented as far as I know.

internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
ethan-gallant: The linked issue gives a bit more background, the webhook seems to be a bandaid fix to prevent the host from getting stuck in the provisioning state.

zaneb:
On further inspection, I see that the webhook change was added recently to prevent people encountering bug #2465.

internal/controller/metal3.io/host_state_machine.go
Rozzii: > Since this lands in an else block after the check for externally provisioned I don't believe that would be possible here.
> 
> Happy to reformat it as well into a switch statement or similar as the conditional branches here are getting a bit wild already before the change.

That is true , silly me :D .

But then there is the other thing zaneb pointed out.

dtantsur: The original bug is about the incorrect state in Ironic: once we get to Inspecting, BMO does not expect to find the Node in the `active` state in Ironic, but that's what actually happens. BMO needs to deprovision the host first. Then make sure that inspection still happen.

internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
dtantsur: To echo what I said in the bug: if removing externallyProvisioned unconditionally deprovisions the host, we need to make sure that `image` and `customDeploy` are not set.

Or we need to permit them to be set and keep the host `Provisioned`.

ethan-gallant: Looks like I may have misunderstood the issue and this will need a bit of additional work. 

Just to confirm (since I was not at the working group meetings) the desired behavior is:
- If the `Image` & `customDeploy` are set
    - Move the host to the `Provisioned` state since the host complies with the standard setup, it was just (as expected) `externallyProvisioned` 
- If the host does not have the `image` & `customDeploy` fields set
    - Deprovision the host since the host can not be managed in the standard fashion


Overall I'm wondering if the approach originally for the externallyProvisioned would have been to follow the cert-manager pattern and had a `ProvisionRequest` CR where some external system could fulfill the provision instead of hacking around the BareMetalHost CR.

Rozzii: _Overall I'm wondering if the approach originally for the externallyProvisioned would have been to follow the cert-manager pattern and had a ProvisionRequest CR where some external system could fulfill the provision instead of hacking around the BareMetalHost CR._


Just to reflect on this point, this would be a nice approach but in practice the system that has done the provisioning might not be cloud native or not even automated at all so a ProvisionRequest CR would provide some optional extra functionality, but it would still not cover all the use cases.

dtantsur: > Overall I'm wondering if the approach originally for the externallyProvisioned would have been to follow the cert-manager pattern and had a ProvisionRequest CR where some external system could fulfill the provision instead of hacking around the BareMetalHost CR.

Well, we should have had a separate CRD for provisioning for sure. The HostClaim work https://github.com/metal3-io/metal3-docs/pull/408 is a step in that direction but it does not address externally provisioned.

> Just to confirm (since I was not at the working group meetings) the desired behavior is

I think your summary is correct, except that: "If the `Image` **or** `customDeploy` are set" (one is enough).

internal/controller/metal3.io/host_state_machine.go
```diff
@@ -408,6 +408,9 @@ func (hsm *hostStateMachine) handleRegistering(_ *reconcileInfo) actionResult {
...
+		// Removing externallyManaged moves hosts to provisioned state
+		hsm.NextState = metal3api.StateProvisioned
```
dtantsur, line 413: See Zane's comment: this is a wrong place. You need to fix `handleExternallyProvisioned` instead.

test/e2e/externally_provisioned_test.go
```diff
@@ -79,14 +80,28 @@ var _ = Describe("Create as externally provisioned, deprovision", Label("require
...
 
+			By("Removing the ExternallyProvisioned field")
+			bmh.Spec.ExternallyProvisioned = false
+
+			err = clusterProxy.GetClient().Update(ctx, &bmh)
+			Expect(err).NotTo(HaveOccurred())
+
+			Eventually(func(g Gomega) {
```
dtantsur, line 93: Better use WaitForBmhInProvisioningState

test/e2e/externally_provisioned_test.go
```diff
@@ -79,14 +80,28 @@ var _ = Describe("Create as externally provisioned, deprovision", Label("require
...
+				err = clusterProxy.GetClient().Get(ctx, kclient.ObjectKeyFromObject(&bmh), &bmh)
+				g.Expect(bmh.Status.Provisioning.State).To(Equal(metal3api.StateProvisioned))
```
dtantsur, line 95: The way the test is written, the expected state is not Provisioned. Or even: it make become Provisioned briefly, then BMO detects that there is no `image` and no `customDeploy` and deprovisions.

PR #2567, opened 2025-07-17
:seedling: [WIP] Add e2e test for automated cleaning

### LINKED ISSUE #1372: Add e2e test for automated disk cleaning

erjavaskivuori:
**What this PR does / why we need it**: Add e2e test for automated cleaning

Fixes #1372


erjavaskivuori:
(Not ready for review, I just wanted to trigger the e2es)

kashifest:
@fmuyassarov @adilGhaffarDev @peppi-lotta PTAL

Rozzii (commented):
> **What this PR does / why we need it**: WIP! Add e2e test for automated cleaning
> 
> To verify that disk cleaning works as expected, the following steps are done in the test:
> 
>     * Provision the BMH
> 
>     * SSH into the BMH and create a test marker on the disk
> 
>     * Deprovision the BMH (this is when disk cleaning should occur)
> 
>     * Provision the BMH again
> 
>     * SSH into the BMH and check if the previously created marker is no longer present, indicating that the disk was successfully cleaned
> 
> 
> **Which issue(s) this PR fixes**: Fixes #1372


To properly test this you have to create at least 2 disks and and put the marker on both as provisioning also destroys the partition table and the file systems  on the disk where the node image will land, a automated cleaning does the same for the other disks too.

I will check the actual cleaning process that BMO initiates because it partially relies on default cleanup but it has some other firmware and bios related conditions too.

Rozzii:
I have checked, I think for now we are fine to just test the default disk metadata cleanup with the 2 disks, we don't really have good firmware update test environment yet and bios/EFI is also not updated as it is a fixed firmware used by qemu+kvm . RAID enabled setups also cleaned up a bit differently but I don't think we have RAID e2e either.

erjavaskivuori:
> I have checked, I think for now we are fine to just test the default disk metadata cleanup with the 2 disks, we don't really have good firmware update test environment yet and bios/EFI is also not updated as it is a fixed firmware used by qemu+kvm . RAID enabled setups also cleaned up a bit differently but I don't think we have RAID e2e either.

Thank you, I am on it!

erjavaskivuori:
/cc @fmuyassarov @adilGhaffarDev @peppi-lotta @Rozzii

This is now ready for review I think

adilGhaffarDev:
code wise looking good to me, nice work. I would tag @lentzi90 to check this.
I am not sure how fixture tests are working, but seems like this test not testing anything when it comes to fixture tests.
should we just exclude this from fixture tests?

kashifest (commented):
/lgtm

test/e2e/automated_cleaning_test.go
```diff
@@ -0,0 +1,247 @@
...
+		}, e2eConfig.GetIntervals(specName, "wait-provisioning")...)
+
+		By("Waiting for the BMH to become provisioned")
+		WaitForBmhInProvisioningState(ctx, WaitForBmhInProvisioningStateInput{
+			Client: clusterProxy.GetClient(),
+			Bmh:    bmh,
+			State:  metal3api.StateProvisioned,
+		}, e2eConfig.GetIntervals(specName, "wait-provisioned")...)
+
+		By("Deprovisioning the BMH to trigger automated cleaning")
```
lentzi90: I think this could be time sensitive. When the BMH becomes provisioned, it only means that the disk image was written and it is now booting from it. So if we want to be sure that the cloud-init script has finished we need something more, otherwise we could start deprovisioning before it ever had time to put anything on the disks.

test/e2e/automated_cleaning_test.go
```diff
@@ -0,0 +1,247 @@
...
+
+		// Only create second disk for environments with Ironic (e.g. not fixture)
+		if e2eConfig.GetVariable("DEPLOY_IRONIC") == "true" {
+			By("Creating second disk for the VM")
+			if err = CreateAndAttachSecondDisk(ctx, bmc.Name); err != nil {
+				Fail(fmt.Sprintf("Failed to create and attach second disk: %v", err))
+			}
+		} else {
+			Logf("Skipping second disk creation because Ironic is not deployed")
+		}
```
lentzi90: This is clever, and works well in CI with VMs, however, we cannot create extra disks like this on real hardware 😅 and I would really like these tests to work for that scenario also.
Could you instead move this so that we do it with the vbmctl before the test starts? I get that it will be "wasteful" to add disks to all, even when this test is not going to run, but I don't think it is a big deal.

erjavaskivuori:
> code wise looking good to me, nice work. I would tag @lentzi90 to check this. I am not sure how fixture tests are working, but seems like this test not testing anything when it comes to fixture tests. should we just exclude this from fixture tests?

I added skipping this test for fixture tests and removed the related if-elses. Also reverted renaming the `SSH_CHECK_PROVISIONED` variable since it's now again only used for checking provisioning state.

test/e2e/automated_cleaning_test.go
```diff
@@ -0,0 +1,240 @@
...
+		bmh := metal3api.BareMetalHost{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      specName,
+				Namespace: namespace.Name,
+				Annotations: map[string]string{
+					metal3api.InspectAnnotationPrefix:   "disabled",
+					metal3api.HardwareDetailsAnnotation: hardwareDetails,
+				},
+			},
+			Spec: metal3api.BareMetalHostSpec{
```
lentzi90, line 56: Could you add automatedCleaningMode here so we make it explicit what we test?

test/e2e/automated_cleaning_test.go
```diff
@@ -0,0 +1,240 @@
...
+		By("Connecting via SSH to check disk state after cleaning")
+		client := EstablishSSHConnection(e2eConfig, bmc.IPAddress)
+		defer client.Close()
+
+		By("Checking that the first disk has been cleaned")
+		output, err := executeSSHCommand(client, "ls -la /test_file_vda.txt 2>/dev/null || echo 'file not found'")
+		Expect(err).NotTo(HaveOccurred())
+		Expect(output).To(ContainSubstring("file not found"), "Test file /test_file_vda.txt should have been cleaned")
+
+		By("Verifying second disk exists but filesystem is cleaned")
```
lentzi90: We would need to check this file before deprovisioning also, otherwise we cannot really trust that it was there to begin with.
Side note: If we could make this test work both ways it would be really awesome! I mean that we would test with cleaning enabled and see that the files disappear, and with cleaning disabled and see that they stay. It is probably better to leave that for another PR though 🙂

lentzi90 (commented):
Awesome work!
I would like to separate the disk attach logic from the test (so that it can work on real hardware) and I think we need to check that the files are there before we start deprovisioning. Otherwise I am really happy with this!

test/e2e/automated_cleaning_test.go
erjavaskivuori: Ow yeah this didn't cross my mind :sweat_smile:  But I can do that, it should be pretty straight forward.

test/e2e/automated_cleaning_test.go
erjavaskivuori: I will add the check before doing the deprovisioning.

I think it would be good idea to implement similar test in other way, too! But yeah maybe not in the scope of this PR 🙂

test/vbmctl/main.go
```diff
@@ -144,20 +143,12 @@ func CreateVolume(conn *libvirt.Connect, volumeName, poolName, poolPath string,
...
-	opts := make(map[string]any)
-	opts[qcow2.OPT_SIZE] = 3 * (1 << 30) //nolint: mnd // qcow2 file's size is 3g
-	opts[qcow2.OPT_FMT] = "qcow2"        // qcow2 format
-	opts[qcow2.OPT_SUBCLUSTER] = true    // enable sub-cluster
 
-	err := qcow2.Blk_Create("/tmp/"+name+".qcow2", opts)
-
-	if err != nil {
-		log.Println("Failed to create qcow2 file")
-		log.Printf("Error occurred: %v\n", err)
```
erjavaskivuori, line 156: When trying to add the second disk, I noticed that this doesn't really do anything. The disk's source file is expected to be at `{{ .PoolPath }}/{{ .Name }}.qcow2` which equals to `/tmp/pool_oo/<name>.qcow2` but here the file has been written to `/tmp/<name>.qcow2`. The file that is in `/tmp/pool_oo/<name>.qcow2`  is the one created in `CreateVolume` func.

lentzi90 (commented):
Really great work!
/approve
If you want to continue with the ginkgo labels stuff, we should at some point move from ginkgo focus/skip to use labels throughout. The tests should already be labeled, we just didn't switch the makefile/CI over to use them yet 🙂

erjavaskivuori:
> Really great work!
/approve
If you want to continue with the ginkgo labels stuff, we should at some point move from ginkgo focus/skip to use labels throughout. The tests should already be labeled, we just didn't switch the makefile/CI over to use them yet 🙂

Yes, I think it could be good to move to the labels as they seem bit more convenient to use! I'll write this down to get back to it later or create issues for it.

I had to make one more change to fix merge conflict and apply changes from #2539

kashifest:
/lgtm
Great work @erjavaskivuori and thanks for the thorough review @lentzi90  and @Rozzii

PR #2576, opened 2025-07-21
:sparkles: introduce preprovisioningKernelParams to BMH

Rozzii:
This PR:
- Implements support for additional kernel parameters during preprovisioning
- Adds PreprovisioningExtraKernelParams spec field to BareMetalHost CRD
- Implements kernel parameter combination logic between BMH and PreprovisioningImage CRs
- Adds comprehensive unit tests for parameter combination scenarios
- Renames getInstanceUpdateOpts to getProvisioningInstanceUpdateOptsForNode for consistency
- Updates BMC access utilities to support extra kernel parameters
- Adds PreprovisioningExtraKernelParams field to ManagementAccessData struct
  for provisioner interface
- Extends kernel extra preprov param handling to account for different preprov
  image scenarios (external initrd, external iso, self made PPI, unknown format, no PPI)
- Adds ownership detection logic to distinguish external vs BMO-managed PreprovisioningImages
- Adds BMO internal defaultPPImgFormat constant for internal BMO-managed PPI identification
- Extends default PPI provider and reconciler to handle extraKernelParams

Notes:
- The new spec field is optional and maintains backward compatibility
- ISO format PreprovisioningImages ignore BMH kernel params with warning
- In case of externally managed PreprovisioningImages with InitRD image format
  BMO will combine both BMH and PPI kernel params
- Kernel parameters cannot be changed during fast-track deployments

Other info:
- Related design PR: https://github.com/metal3-io/metal3-docs/pull/552
- E2E test will be done in separate PR
- I will open an issue related to a discussion about exploring status fields, events, and conditions related to the kernel parameters.


Rozzii:
@dtantsur this is the initial idea based on the proposal, I am testing it manually , so far it worked I think I will add the e2e in a separate PR, but  I think I need to add some unit testing still to this PR.

Rozzii:
I think the essential parts are there, I will push an e2e related PR  as a followup but this is long enough as it is.

Commits will be squashed if folks are okay with the implementation, but I will keep them separate for the time being.

Rozzii:
/cc @kashifest 
FYI

zaneb:
If I'm not mistaken, passing kargs to ironic only works when supplying a kernel+ramdisk (not a prebuilt ISO) for the IPA image.
Assuming that is the case, I would like to see the kargs also passed through to the PreprovisioningImage API so that they're available to anyone building their own ISO.

Rozzii:
> If I'm not mistaken, passing kargs to ironic only works when supplying a kernel+ramdisk (not a prebuilt ISO) for the IPA image. Assuming that is the case, I would like to see the kargs also passed through to the PreprovisioningImage API so that they're available to anyone building their own ISO.

It is available, this feature also integrates the kernel params passed via the PreprovisioningImage. 
Idea is that if preprov image is linked to the BMH then BMO  combines the kernel params from the preprov image and from the BMH both on a per node basis and it is checked when any of the LCM operations are initiated.

zaneb:
Note that the [documentation](https://github.com/metal3-io/baremetal-operator/blob/main/apis/metal3.io/v1alpha1/preprovisioningimage_types.go#L75-L76) for the `ExtraKernelParams` field in the PreprovisioningImage Status states that it "[o]nly makes sense for initrd images".

When booting an initrd+kernel image, ironic either has to put the kernel args into the iPXE config or (for virtualmedia) it builds a custom ISO containing the initrd+kernel+args. But when an ISO is supplied that is not built by ironic, there is no way for it to modify the kernel args. There's no standard place in an ISO where they are stored. So the only way to append to them is to pass them to the PreprovisioningImage controller to build into the ISO at the time it creates it.

Rozzii:
> Note that the [documentation](https://github.com/metal3-io/baremetal-operator/blob/main/apis/metal3.io/v1alpha1/preprovisioningimage_types.go#L75-L76) for the `ExtraKernelParams` field in the PreprovisioningImage Status states that it "[o]nly makes sense for initrd images".
> 
> When booting an initrd+kernel image, ironic either has to put the kernel args into the iPXE config or (for virtualmedia) it builds a custom ISO containing the initrd+kernel+args. But when an ISO is supplied that is not built by ironic, there is no way for it to modify the kernel args. There's no standard place in an ISO where they are stored. So the only way to append to them is to pass them to the PreprovisioningImage controller to build into the ISO at the time it creates it.

Yes that is correct and I agree with you, this feature takes into consideration what you have noted. The reason why I am concatenating the kernel params coming from BMH and PreprovImage is because that way folks who have customized ironics e.g. with custom deploy step that could in theory do whatever with the images would be supported also, so if someone would have a feature to repackage ISOs or would do some sort of magic in their own preprov image controller , they could also take advantage of the BMH preprovExtraKernelArgs param, for others it is just a noop.

zaneb:
AIUI this implementation correctly handles the case where the PreprovisioningImage's `status.format` is `ImageFormatInitRD`, but not the case where it is `ImageFormatISO`. It is up to the PreprovisioningImage controller to select the format from the list sent in `spec.acceptFormats` (for virtualmedia it contains both).

kashifest:
@zaneb @Rozzii planning to do .rc release this week and then minor release next week, do you think this feature can land in the upcoming release.

Rozzii:
> AIUI this implementation correctly handles the case where the PreprovisioningImage's `status.format` is `ImageFormatInitRD`, but not the case where it is `ImageFormatISO`. It is up to the PreprovisioningImage controller to select the format from the list sent in `spec.acceptFormats` (for virtualmedia it contains both).

Sorry for the late reply I was not working the last two weeks. 

I don't think I am touching any IPA format related field or at least not implementing new control logic. What you have mentioned should be operating logic with this feature too.

The only thing I aim to do related to PreprovisioningImage CR is to concatenate the the kernel params provided by  PreprovisioningImage CR and BMH CR. 

In case of the virtualmedia pre-built ISOs the new logic would mean a NO-OP from Ironic side because after BMO updates the preprov kernel params related Ironic node API endpoint with the concatenated params Ironic will disregard it. Ironic will disregard preprov kernel params coming from API or from config file or env variable in case a pre-built IPA ISO is used regardless if it was provided by a controller or "manually".

I might be misunderstanding the problem, in any case my intention is to not get in the way of PreprovisioningImage controller.

IMO this feature could land, I propose to put this behind a feature gate, mark it as alpha or beta, I have not used feature gates so far so I have to look and ask around, please wait with the RC until I add the gate.

Rozzii:
[v0.11-rc.0](https://github.com/metal3-io/baremetal-operator/pull/2648#issuecomment-3248881694) will go out without this anyhow so most likely this wont get merged to 0.11 but anyhow it would be nice to come to a consensus soon.

kashifest:
> [v0.11-rc.0](https://github.com/metal3-io/baremetal-operator/pull/2648#issuecomment-3248881694) will go out without this anyhow so most likely this wont get merged to 0.11 but anyhow it would be nice to come to a consensus soon.

I would say lets take it in even after the .rc if we have a consensus before the actual minor

zaneb:
To make this feature work universally we need to add an extraKernelParams field to the Spec of PreprovisioningImage.

When a PreprovisioningImage is used:

- BMH controller copies the preprovisioningExtraKernelParams from the BMH Spec to the PreprovisioningImage Spec
- PPI controller passes the extraKernelParams from the Spec to the image provider in the [ImageData struct](https://github.com/metal3-io/baremetal-operator/blob/main/pkg/imageprovider/imageprovider.go#L11-L16)
- The [default image provider](https://github.com/metal3-io/baremetal-operator/blob/main/pkg/imageprovider/default.go) copies the extraKernelParams input to the GeneratedImage output when the image type is initrd
- BMH controller passes the extraKernelParams field from the PreprovisioningImage status to ironic

And when a PreprovisioningImage is not used, the BMH controller passes the preprovisioningExtraKernelParams from the BMH directly to ironic, the same as in the PR currently.

I don't think this is super complicated to implement. The PreprovisioningImage controller side of things is +4 lines of code. The BMH controller side is +1/-6 lines of code. Let's go ahead and do it now rather than set ourselves up for inconsistent behaviour between image types.

pkg/provisioner/provisioner.go
```diff
@@ -100,29 +106,32 @@ type InspectData struct {
...
-	HardwareProfile profile.Profile
-	RootDeviceHints *metal3api.RootDeviceHints
-	CustomDeploy    *metal3api.CustomDeploy
+	Image                            metal3api.Image
+	HostConfig                       HostConfigData
+	BootMode                         metal3api.BootMode
+	HardwareProfile                  profile.Profile
+	RootDeviceHints                  *metal3api.RootDeviceHints
+	CustomDeploy                     *metal3api.CustomDeploy
+	PreprovisioningExtraKernelParams string
```
zaneb: Even from the naming alone here it doesn't make sense to me that ProvisionData would ever need to contain any data about PreProvisioning.

pkg/provisioner/ironic/ironic.go
```diff
@@ -461,9 +471,22 @@ func setDeployImage(config ironicConfig, accessDetails bmc.AccessDetails, hostIm
 				deployImageInfo[deployKernelKey] = config.deployKernelURL
 			}
 			deployImageInfo[deployRamdiskKey] = hostImage.ImageURL
+			// Extra params from the kernel have been already posted to
+			// Ironic node API in the Register function, so extra kernel args in driver info need
```
zaneb: er, this _is_ the Register function.

pkg/provisioner/ironic/ironic.go
```diff
@@ -694,6 +717,41 @@ func (p *ironicProvisioner) getInstanceUpdateOpts(ironicNode *nodes.Node, data p
 		p.setDirectDeployUpdateOptsForNode(ironicNode, &data.Image, updater)
 	}
 
+	driverOpts := clients.UpdateOptsData{
+		"kernel_append_params": fmtPreprovExtraKernParams(data.PreprovisioningExtraKernelParams),
+	}
+	updater.SetDriverInfoOpts(driverOpts, ironicNode)
```
zaneb: We don't need to set this again when it was already set in `Register()`. (And we definitely shouldn't set it 4 different times for 4 different provisioning states we might be in.) We can get rid of most of the changes to the provisioner API as well (only the `ManagementAccessData` needs to change).

Rozzii:
> To make this feature work universally we need to add an extraKernelParams field to the Spec of PreprovisioningImage.
> 
> When a PreprovisioningImage is used:
> 
>     * BMH controller copies the preprovisioningExtraKernelParams from the BMH Spec to the PreprovisioningImage Spec
> 
>     * PPI controller passes the extraKernelParams from the Spec to the image provider in the [ImageData struct](https://github.com/metal3-io/baremetal-operator/blob/main/pkg/imageprovider/imageprovider.go#L11-L16)
> 
>     * The [default image provider](https://github.com/metal3-io/baremetal-operator/blob/main/pkg/imageprovider/default.go) copies the extraKernelParams input to the GeneratedImage output when the image type is initrd
> 
>     * BMH controller passes the extraKernelParams field from the PreprovisioningImage status to ironic
> 
> 
> And when a PreprovisioningImage is not used, the BMH controller passes the preprovisioningExtraKernelParams from the BMH directly to ironic, the same as in the PR currently.
> 
> I don't think this is super complicated to implement. The PreprovisioningImage controller side of things is +4 lines of code. The BMH controller side is +1/-6 lines of code. Let's go ahead and do it now rather than set ourselves up for inconsistent behaviour between image types.

I understand your approach, I also don't think it is super complicated, I am not objecting to implementing this, if you are okay with BMH dictating extra kernel params to the PreprovisioningImage I am fine with it.

On the other hand I would like to clarify the intended use of the PreprovisioningImage controller because with the changes you are proposing I feel that we would provide a parallel way of specifying extraKernel params as there was such a way for PreprovImages already AFAIK. I am fine with that but I would like to make sure we are on the same page.

The reason why I have not yet implemented what you are asking for is that  I have seen the extraKernelParams status field here https://doc.crds.dev/github.com/metal3-io/baremetal-operator/metal3.io/PreprovisioningImage/v1alpha1@v0.10.2#status-extraKernelParams and checked https://github.com/openshift/image-customization-controller .

So my assumption was that the desired behavior from UX/developer perspective was that the "external co-controller/image provider" implementation would (and they do) facilitate a way to specify extraKernelParams used during preprov "image building" and then it is reported by the co-controller/image provider into the status field of the PreprovImage CR and then BMH would read that status field. So in the current way of using the PreprovImage CR , it is the PreprovImage CR's status field that dictates extraKernelParams to BMH and that field is populated by external means.  I have considered it redundant or potentially confusing for users to implement an extraKernelParams status field and a spec field the same time and to present the same information from to different "directions" that belongs to two different approaches of supplying this data. 

That said, we can leave it to the external controller/image provider to reconcile the extra kernel params if it would be coming from a new spec field of the PreprovisioningImage CR if you say we should extend this "interface" between the image providers and BMH. I am not an active user of the PreprovisioningImage CR and related controllers so I was just worried that I would confuse users and devs or introduce silent incompatibilities.  Many of the image providers are not even FOSS and they are used to implement custom preprov image handling soo it is hard for me to estimate the effect of the new spec field.

I also find the approach of BMH dictating extraKernelParams to the PreprovImage CR the more consistent and logical way same as you have explained, but I was deterred by the involvement of the external controllers and the fact that they are reporting external kernel params already.

pkg/hardwareutils/bmc/ilo4.go
```diff
@@ -63,11 +63,12 @@ func (a *iLOAccessDetails) DisableCertificateVerification() bool {
...
-func (a *iLOAccessDetails) DriverInfo(bmcCreds Credentials) map[string]interface{} {
+func (a *iLOAccessDetails) DriverInfo(bmcCreds Credentials, preProvExtraKernParams string) map[string]interface{} {
 	result := map[string]interface{}{
-		"ilo_username": bmcCreds.Username,
-		"ilo_password": bmcCreds.Password,
-		"ilo_address":  a.hostname,
+		"ilo_username":         bmcCreds.Username,
+		"ilo_password":         bmcCreds.Password,
+		"ilo_address":          a.hostname,
+		"kernel_append_params": preProvExtraKernParams,
```
zaneb: If this is not driver-specific (which certainly appears to be the case) then I don't feel like adding it to every driver is the right thing to do. You're already setting `kernel_append_params` later in the `Register()` call as part of `setDeployParams()`, there is no need to do it twice. So we can get rid of the change to the `DriverInfo()` API as well.

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -839,15 +866,16 @@ func (r *BareMetalHostReconciler) registerHost(prov provisioner.Provisioner, inf
...
-			HasCustomDeploy:            hasCustomDeploy(info.host),
-			DisablePowerOff:            info.host.Spec.DisablePowerOff,
-			CPUArchitecture:            getHostArchitecture(info.host),
+			BootMode:                         info.host.Status.Provisioning.BootMode,
+			AutomatedCleaningMode:            info.host.Spec.AutomatedCleaningMode,
+			State:                            info.host.Status.Provisioning.State,
+			CurrentImage:                     getCurrentImage(info.host),
+			PreprovisioningImage:             preprovImg,
+			PreprovisioningNetworkData:       preprovisioningNetworkData,
+			PreprovisioningExtraKernelParams: r.retrievePreprovisioningExtraKernelParamsSpec(info, prov),
```
zaneb: I believe this is actually the only function that we need to pass `PreprovisioningExtraKernelParams` to (see other comments), so once the others are removed we should pass `preprovImg` as a param here instead of calling `getPreprovImage()` a second time.

pkg/provisioner/ironic/ironic.go
```diff
@@ -314,14 +315,23 @@ func (p *ironicProvisioner) createPXEEnabledNodePort(uuid, macAddress string) er
...
+
 // configureNode configures Node properties that are not related to any specific provisioning phase.
 // It populates the AutomatedClean field, as well as capabilities and architecture in Properties.
 // It also calls setDeployImage to populate IPA parameters in DriverInfo and
 // checks if the required PreprovisioningImage is provided and ready.
 func (p *ironicProvisioner) configureNode(data provisioner.ManagementAccessData, ironicNode *nodes.Node, bmcAccess bmc.AccessDetails) (result provisioner.Result, err error) {
 	updater := clients.UpdateOptsBuilder(p.log)
 
-	deployImageInfo := setDeployImage(p.config, bmcAccess, data.PreprovisioningImage)
+	deployImageInfo := setDeployImage(p.config, bmcAccess, data.PreprovisioningImage, data)
```
zaneb: nit: either pass `data.PreprovisioningImage` and `data.PreprovisioningExtraKernelParams` or just `data`. Don't pass `data` and one subfield of it.

pkg/provisioner/ironic/ironic.go
```diff
@@ -461,9 +471,22 @@ func setDeployImage(config ironicConfig, accessDetails bmc.AccessDetails, hostIm
...
+				// Combine the default ironic node kernel params with the extra
+				// params from the BMH and the dynamically generated preprov image
+				comExtraKernelParams := defaultKernelParam
+				trimDataKParams := strings.TrimSpace(data.PreprovisioningExtraKernelParams)
+				trimPreprovImgKParams := strings.TrimSpace(hostImage.ExtraKernelParams)
+				if trimDataKParams != "" {
+					comExtraKernelParams += " " + data.PreprovisioningExtraKernelParams
+				}
+				if trimPreprovImgKParams != "" {
+					comExtraKernelParams += " " + hostImage.ExtraKernelParams
```
zaneb: Here you are adding the kargs supplied by the PreprovisioningImage to the ones passed in the BMH Spec.
But you already did that in the BMH controller (which is the correct place to do it - this shouldn't be provisioner-specific). So the ones from the PreprovisioningImage will be duplicated.

pkg/provisioner/provisioner.go
```diff
@@ -74,23 +74,29 @@ type PreprovisioningImage struct {
...
-	HasCustomDeploy            bool
-	DisablePowerOff            bool
-	CPUArchitecture            string
+	BootMode                         metal3api.BootMode
+	AutomatedCleaningMode            metal3api.AutomatedCleaningMode
+	State                            metal3api.ProvisioningState
+	CurrentImage                     *metal3api.Image
+	PreprovisioningImage             *PreprovisioningImage
+	PreprovisioningNetworkData       string
+	PreprovisioningExtraKernelParams string
```
zaneb, line 84: It's a bit unfortunate that we can't use the ExtraKernelParams field inside `PreprovisioningImage.GeneratedImage`, but when no image is set then `PreprovisioningImage` has to be `nil`, and I don't see a way around that.
I wonder if we should stop inheriting from the `imageprovider.GeneratedFormat` struct - which we were only really using for convenience - here though, so that we don't have two copies of the field with different values.

zaneb (commented):
I believe this patch could be far, far simpler if we stopped doing the same work 2-3 times per reconcile.

zaneb:
> I have considered it redundant or potentially confusing for users to implement an extraKernelParams status field and a spec field the same time and to present the same information from to different "directions" that belongs to two different approaches of supplying this data.

Unfortunately the domain we are modelling is inherently confusing: for an ISO image the user-supplied params must be an input to the PreprovisioningImage so they can be built into the ISO, and for an initrd image the image-supplied params must be an output from the PreprovisioningImage so that ironic can add them to the PXE command line.

We could try to give them different names to disambiguate (but unfortunately the one that really needs to change is the one that already exists and can't be changed without breaking APIs). But the only other real _choice_ we have to make is whether we add the user-supplied params to the image-supplied params inside the preprovisioningimage controller or inside the bmh controller. You could make an argument either way. I figured the API made more sense if the user-supplied params were always flowing through the PreprovisioningImage (and getting added to along the way). But if it were done in the BMH controller then no changes would be required to existing PreprovisioningImage controllers to get this to work, so there's a case for that too.

pkg/provisioner/ironic/ironic.go
Rozzii: I update it in so many places because I wanted to make it changeable not used only for initial deployment but I want to allow changing the kernel parameters before the LCM operations. Before cleaning, after registration, before deployment before servicing etc. 

And I know this is a limitation for pre-built IPA ISOs or LiveISO boots, but for dynamically built IPA ISO or PXE it is very useful for debugging.

Rozzii:
> Unfortunately the domain we are modelling is inherently confusing: for an ISO image the user-supplied params must be an input to the PreprovisioningImage so they can be built into the ISO, and for an initrd image the image-supplied params must be an output from the PreprovisioningImage so that ironic can add them to the PXE command line.

Okay, I see this is a solid idea for the target workflow, it makes sense for me. I was initally surprised that the extraKernelParams spec was missing from the PreprovisioningImage for the exact same reason. Now that I understand the intended way of directing the data I will modify it accordingly to your proposal (or at least according to my understanding of your proposal).

I will get back with the changes, and big thanks for the very useful review!

I will reply to the comments related to implementation details, reconciliation, API etc over time, I will need to check why I did some stuff the way I did as I have implemented this  some time ago.

pkg/provisioner/ironic/ironic.go
zaneb: `Register()` is already called in all of those places: https://github.com/metal3-io/baremetal-operator/blob/main/internal/controller/metal3.io/host_state_machine.go#L369

Rozzii:
This will be updated, but I am still working on something else
/remove-lifecycle stale

Rozzii:
Don't mind me, I am just slowly re-basing the related branch, will be green soon!

pkg/provisioner/ironic/ironic.go
Rozzii: Done

Rozzii:
/hold 
I am still working on the refactoring.

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -84,6 +84,49 @@ func (info *reconcileInfo) publishEvent(reason, message string) {
...
+		trimmedParams := strings.TrimSpace(kernelExtraPreprovParams)
+		if format == metal3api.ImageFormatISO {
+			if trimmedParams != "" {
+				// log warning
+				info.log.Info("Warning: PPI is in ISO fromat, BMH preprovisioningExtraKernelParams are ignored in favor of PPI extra kernel parameters!")
+			}
+			// Kernel params are already injected to the ISO
+			return ""
+		}
+		kernelExtraPreprovParams += " " + preprovImg.GeneratedImage.ExtraKernelParams
```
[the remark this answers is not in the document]
Rozzii: Not relevant it will be cleaned up on provisioner level.

internal/controller/metal3.io/preprovisioningimage_controller.go
```diff
@@ -193,23 +213,34 @@ func (r *PreprovisioningImageReconciler) update(ctx context.Context, img *metal3
...
+		ImageMetadata:     ppi.ObjectMeta.DeepCopy(),
 		Format:            format,
-		Architecture:      img.Spec.Architecture,
+		Architecture:      ppi.Spec.Architecture,
 		NetworkDataStatus: secretStatus,
 	}, networkDataContent, log)
 	if err != nil {
 		failure := imageprovider.ImageBuildInvalidError{}
 		if errors.As(err, &failure) {
 			log.Info("image build failed", "error", "err")
```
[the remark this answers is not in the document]
Rozzii: It is not even my code but I will fix it then.

pkg/provisioner/ironic/register.go
```diff
@@ -149,6 +150,9 @@ func (p *ironicProvisioner) Register(ctx context.Context, data provisioner.Manag
 		if credentialsChanged || bmcAddressChanged {
 			p.log.Info("Updating driver info because the credentials and/or the BMC address changed")
 			updater.SetTopLevelOpt("driver_info", driverInfo, ironicNode.DriverInfo)
+		} else if fmtPreprovExtraKernParams("") != finalPreprovKernParams {
+			p.log.Info("Updating driver info because extra preprovisioning kernel arguments are in use!")
+			updater.SetTopLevelOpt("driver_info", driverInfo, ironicNode.DriverInfo)
```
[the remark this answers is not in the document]
Rozzii: Updating driver info on each reconciliation iteration (if needed) is by design, "SetTopLevelOpt) will just add this update operation to the list of update operations that will be done at the end of the "registration" process. the kernel_append_params field of the driver info is ephemeral so the key value pairs don't accumulate up and in BMO this is the only place where the key+value pairs are set.

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -84,6 +84,49 @@ func (info *reconcileInfo) publishEvent(reason, message string) {
...
+	}
+	// If the execution reaches this point it means there is a PPI in the
+	// namespace with acceptable image fromat but it is managed by an external
+	// controller-manager.
+	// From this point onward there are 2 separate cases, if the image fromat
+	// is initrd then BMH and PPI kernel params are combined. In case
+	// the PPI is in ISO format then the BMH extra kernel params will be
+	// ignored and the spec field will be cleaned up and a warning will be
+	// logged to inform the user that in the specific use case only PPI
+	// kernel params matter.
```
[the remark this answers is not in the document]
Rozzii: fixed

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -84,6 +84,49 @@ func (info *reconcileInfo) publishEvent(reason, message string) {
...
 }
 
+// return the PreprovisioningExtraKernelParams from the reconciliation info.
+func (r *BareMetalHostReconciler) retrievePreprovisioningExtraKernelParamsSpec(ctx context.Context, info *reconcileInfo, prov provisioner.Provisioner) string {
+	if info == nil || info.host == nil {
+		return ""
+	}
+	kernelExtraPreprovParams := info.host.Spec.PreprovisioningExtraKernelParams
+
+	// This is a multy purpose check, this both checks whether BMO is running
```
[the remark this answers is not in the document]
Rozzii: fixed

config/base/crds/bases/metal3.io_baremetalhosts.yaml
```diff
@@ -344,6 +344,11 @@ spec:
                   state (e.g. provisioned), its power state will be forced to match
                   this value.
                 type: boolean
+              preprovisioningExtraKernelParams:
+                description: |-
+                  The value of the kernel commandline argument list that will be passed
+                  to the pre provisioning agent's kernel during boot.
```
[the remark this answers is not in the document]
Rozzii: fixed

config/render/capm3.yaml
```diff
@@ -344,6 +344,11 @@ spec:
                   state (e.g. provisioned), its power state will be forced to match
                   this value.
                 type: boolean
+              preprovisioningExtraKernelParams:
+                description: |-
+                  The value of the kernel commandline argument list that will be passed
+                  to the pre provisioning agent's kernel during boot.
```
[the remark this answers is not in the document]
Rozzii: fixed

pkg/provisioner/ironic/ironic.go
```diff
@@ -471,9 +482,12 @@ func setDeployImage(config ironicConfig, accessDetails bmc.AccessDetails, hostIm
 				deployImageInfo[deployKernelKey] = config.deployKernelURL
 			}
 			deployImageInfo[deployRamdiskKey] = hostImage.ImageURL
-			if hostImage.ExtraKernelParams != "" {
-				// Using %default% prevents overriding the config in ironic-image
-				deployImageInfo[kernelParamsKey] = "%default% " + hostImage.ExtraKernelParams
+			// In case the preprovImage format is initrd, the source of
+			// truth for extra kernel params is the BMH as the provisioner is
+			// in charge of constricting the pxe and grub config files
```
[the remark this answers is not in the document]
Rozzii: fixed

internal/controller/metal3.io/preprovisioningimage_controller.go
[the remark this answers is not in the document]
Rozzii: fixed

internal/controller/metal3.io/preprovisioningimage_controller.go
```diff
@@ -193,23 +213,34 @@ func (r *PreprovisioningImageReconciler) update(ctx context.Context, img *metal3
...
-			return setError(generation, &img.Status, reasonImageBuildInvalid, failure.Error()), nil
+			return setError(generation, &ppi.Status, reasonImageBuildInvalid, failure.Error()), nil
 		}
 		return false, err
 	}
+
+	if format == metal3api.ImageFormatInitRD {
+		// Potentially might be other PPI owners in the future when
+		// multy tenancy supporti is fully implemented
```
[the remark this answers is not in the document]
Rozzii: fixed

pkg/provisioner/ironic/ironic.go
```diff
@@ -316,14 +317,23 @@ func (p *ironicProvisioner) createPXEEnabledNodePort(ctx context.Context, uuid,
 	return nil
 }
 
+func fmtPreprovExtraKernParams(params string) string {
+	trimmedParams := strings.TrimSpace(params)
+	if trimmedParams == "" {
+		return defaultKernelParam
+	}
+	// spaceing matters for kernel params
```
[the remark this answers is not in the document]
Rozzii: fixed

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -495,6 +495,11 @@ type BareMetalHostSpec struct {
 	// without hardware profiles.
 	HardwareProfile string `json:"hardwareProfile,omitempty"`
 
+	// The value of the kernel commandline argument list that will be passed
+	// to the pre provisioning agent's kernel during boot.
```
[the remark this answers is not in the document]
Rozzii: fixed

internal/controller/metal3.io/preprovisioningimage_controller.go
```diff
@@ -193,23 +213,34 @@ func (r *PreprovisioningImageReconciler) update(ctx context.Context, img *metal3
...
+		bmh, err := r.findOwnerBMH(ctx, ppi)
+		if err != nil {
+			return false, fmt.Errorf("failed to find BareMetalHost: %w", err)
+		}
+		image.ExtraKernelParams = bmh.Spec.PreprovisioningExtraKernelParams
+		log.Info("extra kernel parameters added to PPI", "params", image.ExtraKernelParams)
```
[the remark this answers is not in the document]
Rozzii: This code belong soley to the built in PPI reconciler that is using the default PPI provider thus this type of PPI gets the kernel params directly from the BMH. Params won't be duplicated because when this default PPI reconciler+provider is used, the params of the PPI and the BMH won't be combined BMH will control what is present in the status of the PPI because BMO is creating the PPI in this scenario. When the default PPI provider and reconciler is turned off (with the related env variable) none of this code gets executed.

pkg/provisioner/ironic/ironic_test.go
```diff
@@ -137,3 +137,51 @@ func TestNewNoBMCDetails(t *testing.T) {
...
+		},
+		{
+			name:     "multiple params",
+			input:    "console=ttyS0,115200 debug=1",
+			expected: defaultKernelParam + " console=ttyS0,115200 debug=1",
+		},
+		{
+			name:     "params with leading/trailing spaces",
+			input:    "  console=ttyS0 debug=1  ",
+			expected: defaultKernelParam + "   console=ttyS0 debug=1  ",
```
lentzi90: Shouldn't we trim the extra spaces?

pkg/provisioner/ironic/ironic.go
```diff
@@ -316,14 +317,23 @@ func (p *ironicProvisioner) createPXEEnabledNodePort(ctx context.Context, uuid,
 	return nil
 }
 
+func fmtPreprovExtraKernParams(params string) string {
+	trimmedParams := strings.TrimSpace(params)
+	if trimmedParams == "" {
+		return defaultKernelParam
+	}
+	// spacing matters for kernel params
+	return defaultKernelParam + " " + params
```
lentzi90: Shouldn't we return the trimmed params instead of the original?
```suggestion
	return defaultKernelParam + " " + trimmedParams
```

pkg/provisioner/provisioner.go
```diff
@@ -111,14 +112,14 @@ type PrepareData struct {
 	TargetFirmwareComponents []metal3api.FirmwareUpdate
 }
 
+// HasFirmwareSpec True if any firmware spec exists (settings, components, or
+// legacy FirmwareConfig), used to distinguish "no updates" from "user cleared spec".
```
lentzi90: The comment makes more sense where it was I think

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -876,16 +941,17 @@ func (r *BareMetalHostReconciler) registerHost(ctx context.Context, prov provisi
...
-			DisablePowerOff:            info.host.Spec.DisablePowerOff,
-			CPUArchitecture:            getHostArchitecture(info.host),
+			BootMode:                         info.host.Status.Provisioning.BootMode,
+			AutomatedCleaningMode:            info.host.Spec.AutomatedCleaningMode,
+			State:                            info.host.Status.Provisioning.State,
+			OperationalStatus:                info.host.Status.OperationalStatus,
+			CurrentImage:                     getCurrentImage(info.host),
+			PreprovisioningImage:             preprovImg,
+			PreprovisioningNetworkData:       preprovisioningNetworkData,
+			PreprovisioningExtraKernelParams: r.retrievePreprovisioningExtraKernelParamsSpec(ctx, info, prov),
```
lentzi90: We have `preprovImg` up on line 923 that we could pass in here. That way we can skip the `getPreprovImage` call in `retrievePreprovisioningExtraKernelParamsSpec`.

pkg/provisioner/ironic/testbmc/testbmc.go
```diff
@@ -16,13 +16,15 @@ func NewTestBMCAccessDetails(parsedURL *url.URL, disableCertificateVerification
...
 		disableCertificateVerification: disableCertificateVerification,
+		driverInfo:                     "",
 	}, nil
 }
 
 type testAccessDetails struct {
 	bmcType                        string
 	hostname                       string
 	disableCertificateVerification bool
+	driverInfo                     string
```
lentzi90, line 27: Do we need this? I don't see it used anywhere.

pkg/hardwareutils/bmc/access.go
```diff
@@ -68,7 +68,7 @@ type AccessDetails interface {
 	// pre-populated with the access information, and the caller is
 	// expected to add any other information that might be needed
 	// (such as the kernel and ramdisk locations).
-	DriverInfo(bmcCreds Credentials) map[string]interface{}
+	DriverInfo(bmcCreds Credentials, preprovExtraKernelArgs string) map[string]interface{}
```
lentzi90, line 71: Is it necessary to change `DriverInfo`? I saw there was a comment before about this but I didn't get the conclusion.

pkg/provisioner/ironic/ironic_test.go
Rozzii: Kernel params are space separated, the number of spaces shouldn't matter. Current approach is that we do the absolute minimum modification to the params. (Basically only the concatenation) That said we can trim them I am not against it.

pkg/provisioner/ironic/ironic.go
Rozzii: Same is above, not mandatory but we can trim them I am not against that.

internal/controller/metal3.io/baremetalhost_controller.go
Rozzii: Yeah that makes sense, I will check how big of a change that would be .

pkg/provisioner/ironic/ironic.go
lentzi90: Ok, I think it would be nice to trim the "edges". This behavior of just trimming to see if it turns up empty is a bit surprising

pkg/provisioner/provisioner.go
Rozzii: I did this because the linter started complaining, I might be doing something wrong but I was unable to shut up the golangci linter any other way.

pkg/hardwareutils/bmc/access.go
Rozzii: It might be doable via instance info I am not sure about it,  but we are already using driver info and we are updating driver info at every reconciliation if needed already.

Relation to the drivers:
If you check ironic-image you can see that in ironic.conf we are setting driver specific defaults for kernel params so I felt if we are manipulating these values on a driver level in ironic I would just follow along. In the previous comment it was pointed out that I am setting kernel params her https://github.com/metal3-io/baremetal-operator/pull/2576/changes/BASE..7cdcfb5cda0ae8b32e75dc645c4da18bd7defd65#diff-f05fc4a0c9679815bc34bcef100c2a8d525688f2804a1128c5b9e1e575b76c29L326 but this also ends up setting the driver_info at the en off the call chain.

Granularity:
Might be that in the future there will be a driver that won't support kernel param injection on a per node basis or only with limitations, this injection is done differently for different drivers on the provisioner level. An other thing is that we are representing/mirroring the ironic dirver_info in BMO for supported drivers already so I thought why not just extend the BMO representation of the driver_info with what we have on the ironic side anyways.

pkg/provisioner/ironic/ironic_test.go
Rozzii: now it is trimming

pkg/provisioner/ironic/ironic.go
Rozzii: now it is trimming

internal/controller/metal3.io/baremetalhost_controller.go
Rozzii: It required new tests but is done.

pkg/provisioner/provisioner.go
lentzi90: Sounds very strange. I checked out the PR now and moved this down, then `make clean` and `make lint`. No issues.
Did you get those locally or also here when pushing?

Rozzii: Only checked those locally.

pkg/provisioner/ironic/testbmc/testbmc.go
Rozzii: It is used directly above in the NewTestBMCAccessDetails, need to return bmc.AccessDetails and that needs a driver info field.

pkg/provisioner/provisioner.go
lentzi90: Could you please try pushing it? I am quite happy with the PR otherwise but this is bothering me :sweat_smile:

Rozzii: Now it has worked for me too, I guess I had some issue with my env or with something else before.

lentzi90 (commented):
Should we mark this as ⚠️ instead of ✨ because of the `DriverInfo` change? I am not sure how "public" the go code is considered.
Anyway, this is lgtm. I am not familiar enough with this area to approve though.
/lgtm

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -495,6 +495,11 @@ type BareMetalHostSpec struct {
 	// without hardware profiles.
 	HardwareProfile string `json:"hardwareProfile,omitempty"`
 
+	// The value of the kernel commandline argument list that will be passed
+	// to the preprovisioning agent's kernel during boot.
+	// +optional
+	PreprovisioningExtraKernelParams string `json:"preprovisioningExtraKernelParams,omitempty"`
```
dtantsur, line 501: Maybe for v1beta1 we need to group all preprovisioning parameters

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -84,6 +85,49 @@ func (info *reconcileInfo) publishEvent(reason, message string) {
...
+	// ignored and a warning will be logged to inform the user that in the
+	// specific use case only PPI kernel params matter. In any other scenarios
+	// the preprovisioningExtraKernelParams are taken only from BMH spec.
+	if preprovImg != nil {
+		format := preprovImg.Format
+		trimmedParams := strings.TrimSpace(kernelExtraPreprovParams)
+		switch format {
+		case metal3api.ImageFormatISO:
+			if trimmedParams != "" {
+				info.log.Info("Warning: PPI is in ISO format, BMH preprovisioningExtraKernelParams are ignored in favor of PPI extra kernel parameters!")
```
dtantsur, line 113: This is confusing. PPI is not a user's input, it's something *we* create. What do we expect the operator to do on seeing this?

dtantsur: Ah, it's in PPI status, not spec. Nevermind, I'm forgetting how PPIs work :)

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -84,6 +85,49 @@ func (info *reconcileInfo) publishEvent(reason, message string) {
...
+			}
+			// Kernel params are already injected to the ISO.
+			kernelExtraPreprovParams = ""
+		case metal3api.ImageFormatInitRD:
+			trimmedImageParams := strings.TrimSpace(preprovImg.GeneratedImage.ExtraKernelParams)
+			if trimmedImageParams != "" {
+				kernelExtraPreprovParams += " " + trimmedImageParams
+			}
+			info.log.Info("BMH and external PPI (preprovisioning)ExtraKernelParams got combined!")
```
dtantsur, line 122: Is this going to be logged each time, even if either or both are empty? I suspect so. Let's log this line only if both are not empty?

dtantsur: By the way, let's log the full CRD name "PreprovisioningImage". I'm afraid that PPI is jargon. Applies everywhere.

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -776,6 +820,15 @@ func (r *BareMetalHostReconciler) getPreprovImage(ctx context.Context, info *rec
 		return nil, fmt.Errorf("failed to retrieve pre-provisioning image data: %w", err)
 	}
 
+	// Check who reconciles the given PPI, if it is not this BMO then it is
+	// and externally managed PPI.
+	isExternal := false
+	ctrlOwnerRef := metav1.GetControllerOf(&preprovImage)
+	if ctrlOwnerRef == nil || ctrlOwnerRef.Name != info.host.Name || ctrlOwnerRef.UID != info.host.UID {
+		info.log.Info("Using externally managed PreprovisioningImage")
```
dtantsur, line 828: `V(1)` please. This is not super important for operators.

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -58,6 +58,7 @@ const (
 	clarifySoftPoweroffFailure    = "Continuing with hard poweroff after soft poweroff fails. More details: "
 	hardwareDataFinalizer         = metal3api.BareMetalHostFinalizer + "/hardwareData"
 	NotReady                      = "Not ready"
+	defaultPPImgFormat            = "DEFAULT_BMH_PPI_IMG"
```
dtantsur, line 61: This constant is a part of the Provisioner API and should be moved somewhere there.

pkg/hardwareutils/bmc/access.go
dtantsur: Instance info won't work: it only applies during provisioning, not inspection or cleaning.

dtantsur: On the other hand, I don't quite understand why this is designed to be driver-specific. All drivers use kernel_append_params, it's a standard field.

pkg/provisioner/ironic/inspecthardware.go
```diff
@@ -39,6 +39,7 @@ func (p *ironicProvisioner) startInspection(ctx context.Context, data provisione
 			SetPropertiesOpts(opts, ironicNode),
 	)
 	if !started {
+		p.log.Info("node update not started", "node", ironicNode.UUID)
```
dtantsur, line 42: This is confusing even to me, to be honest. I'm not sure how operators are supposed to interpret this message. Expand or drop to `V(1)`?

pkg/provisioner/ironic/ironic.go
```diff
@@ -316,14 +317,23 @@ func (p *ironicProvisioner) createPXEEnabledNodePort(ctx context.Context, uuid,
 	return nil
 }
 
+func fmtPreprovExtraKernParams(params string) string {
+	trimmedParams := strings.TrimSpace(params)
+	if trimmedParams == "" {
+		return defaultKernelParam
```
dtantsur, line 323: nit: you can just as well return an empty string to use the defaults

pkg/provisioner/ironic/ironic.go
```diff
@@ -471,9 +482,12 @@ func setDeployImage(config ironicConfig, accessDetails bmc.AccessDetails, hostIm
...
 			}
 			deployImageInfo[deployRamdiskKey] = hostImage.ImageURL
-			if hostImage.ExtraKernelParams != "" {
-				// Using %default% prevents overriding the config in ironic-image
-				deployImageInfo[kernelParamsKey] = "%default% " + hostImage.ExtraKernelParams
+			// In case the preprovImage format is initrd, the source of
+			// truth for extra kernel params is the BMH as the provisioner is
+			// in charge of constructing the pxe and grub config files
+			// TODO Set the status on the preprov image CR
+			if data.PreprovisioningExtraKernelParams != "" {
```
dtantsur, line 489: nit: you check for an empty (trimmed) string in fmtPreprovExtraKernParams already

pkg/provisioner/ironic/register.go
```diff
@@ -149,6 +150,9 @@ func (p *ironicProvisioner) Register(ctx context.Context, data provisioner.Manag
 		if credentialsChanged || bmcAddressChanged {
 			p.log.Info("Updating driver info because the credentials and/or the BMC address changed")
 			updater.SetTopLevelOpt("driver_info", driverInfo, ironicNode.DriverInfo)
+		} else if fmtPreprovExtraKernParams("") != finalPreprovKernParams {
```
dtantsur, line 153: AI is right, this is a problem. You're going to update DriverInfo on every iteration, and it's never a no-op update because credentials are reported by Ironic as `******`. This is why line 151 is conditional on detecting an actual change in the BMH.

dtantsur:
/hold

I've left a whole bunch of comments but only two of them are serious enough to block this patch:
1. Updating DriverInfo on every reconciliation will loop infinitely
2. Seemingly unnecessary change to the Access interface (kernel_append_params is standard in Ironic)

pkg/provisioner/ironic/register.go
dtantsur: On the other hand, `kernel_append_params` is not a secret field, as far as I recall. You can directly compare the new value with the old one.

Rozzii: It might have changed in never Ironics, up to v32 it was so secret that I couldn't even find it in the database let alone on the API, it was totally ephemeral. It might have changed, I will check the newer versions but that  is the reason why I set it at every time as long as BMH has something other than the default. I might have missed something so will double check but I did this implementation exactly because I thought it was an ephemeral API endpoint.

pkg/provisioner/ironic/inspecthardware.go
Rozzii: Good catch. Tbh this might be an artifact of some re-base or some previous version (quite old PR) , I think this can be removed.

pkg/hardwareutils/bmc/access.go
Rozzii: I have nothing against making a "common" dirverInfoCommon function for the AccessDetails interface that assigns the values for the commonly available fields of the driver specific driver_info structs or implementing an AccessDetails implementations called "commonAccessDetails" that would be the "parent" for the driver specific ones. I thought restructuring the use of the AccessDetails interface and its DriverInfo function was/is a bit out of scope for this PR (this PR as it is already changing a lot). 

My implementation followed the existing pattern, fields like username and password and address fields are all common across different driver's driver_info API endpoints in Ironic yet each BMO BMC driver had/has a DriverInfo function that  populates these values  in the driver specific AccesDetails interface implementations. I thought this was a conscious choice in the past that is why I just appended the value assignment  in the driver specific DriverInfo functions as there is no "commonAccessDetails" implementation.

Rozzii: My proposal is that I create an issue related to this topic and refactor these interface implementations as part of that issue.

pkg/provisioner/ironic/register.go
dtantsur: The field has been around for a while already. Here is the oldest branch we support, corresponding to Ironic 26.1: https://github.com/openstack/ironic/blob/stable/2024.2/ironic/drivers/utils.py#L396-L404

apis/metal3.io/v1alpha1/baremetalhost_types.go
dtantsur, line 501: Oh, just realized. Should we maybe sanity-check the value on the webhook level, so that it does not contain newlines, for instance?

pkg/provisioner/ironic/register.go
Rozzii: I might be looking this issue from the wrong angle but now I have registered some BMHs with Ironic done inspection everything works fine. I used the feature of this PR to inject kernel params, that worked fine I have checked the live IPA, but If I curl the ironic endpoint I get this:
[ironicnode.json](https://github.com/user-attachments/files/27196665/ironicnode.json)
I can write the extra kernel param field of the driver info all right but when I query the node the field is not present, I checked the sqlite db and this field is not stored there persistently either, this is what I mean when I say secret, I can write it all right but can't query the same field. Ofc it might be that I should query it differently .

Rozzii: I did the above test with v32 Ironic I will do it with v35 soon. I was also not assuming any issue related to this field because it made sense to me that kernel params would be ephemeral as they could contain sensitive info e.g when the IPA ssh login feature is enable.

apis/metal3.io/v1alpha1/baremetalhost_types.go
Rozzii: I can add a check for  control chars and whitespaces.

pkg/provisioner/ironic/register.go
dtantsur: Can it be that this code overwrites something? I don't believe Ironic has any feature to remove kernel_append_params from driver_info.

dtantsur: Also, when Ironic sanitizes values, it looks like `ipmi_password` in your example: `******` instead of the actual value. We don't just hide the field entirely.

pkg/hardwareutils/bmc/access.go
dtantsur: User name and password fields are called differently for different drivers. The kernel_append_params field is standard. You don't really need to modify AccessDetails for it at all.

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -84,6 +85,49 @@ func (info *reconcileInfo) publishEvent(reason, message string) {
 	info.events = append(info.events, info.host.NewEvent(reason, message))
 }
 
+// return the PreprovisioningExtraKernelParams from the reconciliation info.
+func (r *BareMetalHostReconciler) retrievePreprovisioningExtraKernelParamsSpec(info *reconcileInfo, preprovImg *provisioner.PreprovisioningImage) string {
+	if info == nil || info.host == nil {
```
dtantsur, line 90: This looks like a dangerous bit to me. Why is this check here? If this code path is hit, this can explain why you don't see kernel_append_params in Ironic: it's changed to ""

pkg/provisioner/ironic/register.go
Rozzii: I checked the password with just curl 
`curl -u $(cat /opt/metal3-dev-env/ironic/auth/ironic-username):$(cat /opt/metal3-dev-env/ironic/auth/ironic-password) --cert /opt/metal3-dev-env/certs/ironic.crt --key /opt/metal3-dev-env/certs/ironic.key --cacert /opt/metal3-dev-env/certs/ironic-ca.pem -H "Content-Type: application/json" -X PATCH -d '[{"op": "replace", "path": "/driver_info/ipmi_username", "value": "admin"}]'  https://172.22.0.2:6385/v1/nodes/136d15c4-8870-49c8-8e05-45cbe3b43919`

Still got the **** after I checked the values, tested the same command on the username and it worked as expected the username change so I assume my command worked just fine. I have a feeling this https://github.com/openstack/ironic/blob/3a7f87e98ab8fd6dedc753413959c8b91fcb176b/ironic/objects/node.py#L196 has something to do with it not sure.

Anyways I have done test with the kernel_append_params field and it seems like that field is indeed getting removed somehow by the BMH reconciliation process even if I don't use my new feature rather patch in the kernel_append_params with curl. The params are there for a few seconds then they disappear completely, I have paused the reconciliation of the BMH object and then the phenomenon stopped. I will investigate it could be some bug that was always here but wasn't noticed.

dtantsur: > Anyways I have done test with the kernel_append_params field and it seems like that field is indeed getting removed somehow by the BMH reconciliation process even if I don't use my new feature rather patch in the kernel_append_params with curl.

As we've just discussed on the meeting, BMO is already managing this field even without your patch. `setDeployImage` initializes it to `nil` and only sets it to a value of PPI provides one.

pkg/provisioner/ironic/ironic.go
dtantsur, line 464: Highlight: this is where the value is always set to `nil`

dtantsur:
I run Claude on this commit, and here is what it said:

# AI-generated review

## Review of `introduce preprovisioningKernelParams to BMH`

### Critical: `defaultPPImgFormat` breaks deploy image configuration for BMO-managed PPIs

In `getPreprovImage` (baremetalhost_controller.go:860-867), when a PPI is BMO-managed (`!isExternal`), the format is overridden to the synthetic constant `"DEFAULT_BMH_PPI_IMG"`:

```go
if !isExternal {
    PPIFormat = defaultPPImgFormat  // "DEFAULT_BMH_PPI_IMG"
}
```

This format is then passed into `setDeployImage` (ironic.go:470), where it doesn't match either `ImageFormatISO` or `ImageFormatInitRD`. The switch falls through, and because `config.havePreprovImgBuilder` is true (we have a PPI builder), the function returns `nil` — meaning **no deploy image is configured at all** (no `deploy_kernel`, `deploy_ramdisk`, `deploy_iso`, or `kernel_append_params`).

Before this commit, the real format from `preprovImage.Status.Format` was used, so `setDeployImage` correctly matched InitRD or ISO. This is a regression.

Worse: later in `configureNode` (ironic.go:375-377), the nil check triggers `ErrNeedsPreprovisioningImage` for Deprovisioning/Inspecting/Preparing states — even though the PPI exists and has valid image URLs. This would cause a reconciliation loop.

**Fix:** Don't overwrite the `Format` field. Instead, pass an `isExternal` flag separately (e.g., as an additional field on `PreprovisioningImage`) so that `retrievePreprovisioningExtraKernelParamsSpec` can distinguish BMO-managed from external without corrupting the format that `setDeployImage` relies on.

---

### `kernel_append_params` set in `DriverInfo()` then immediately removed by `configureNode()`

In `Register()`, there are two sequential `tryUpdateNode` calls:

1. **First** (register.go:105-106, 150-156): `DriverInfo()` unconditionally sets `kernel_append_params` in the driver_info map (always at least `"%default%"`). This is written when the node is created, or when credentials/address/params change.

2. **Second** (register.go:251 → ironic.go:336-337): `configureNode()` calls `setDeployImage()`, which initializes `kernelParamsKey = nil` and only conditionally sets it. When `setDeployImage` returns a non-nil map with `kernelParamsKey = nil`, `SetDriverInfoOpts` compares nil against the current value and issues a **RemoveOp** (updateopts.go:117-123).

So within a single `Register()` call: step 1 sets `kernel_append_params`, then step 2 can remove it. This happens in the ISO paths and in any path where `PreprovisioningExtraKernelParams` is empty.

---

### Answer: Can `kernel_append_params` be unexpectedly unset?

**Yes**, in several scenarios:

1. **ISO preprovisioning image**: `setDeployImage` takes the ISO path (line 471-475), `kernelParamsKey` stays nil, `SetDriverInfoOpts` removes it. The user set `preprovisioningExtraKernelParams` on the BMH, the controller cleared it for ISO (with a log warning), but Register's DriverInfo() still set it — then configureNode removes it. The net result is removal, but the set-then-remove within one call is wasteful and confusing.

2. **Removal not detected by Register's update condition** (register.go:153): When a user *removes* `preprovisioningExtraKernelParams` from the BMH spec, `finalPreprovKernParams` becomes `"%default%"`, and the condition `fmtPreprovExtraKernParams("") != finalPreprovKernParams` is false. Register skips the driver_info update. But the node **still has the OLD value** (e.g., `"%default% console=ttyS0"`). Whether it gets cleaned up depends entirely on which `setDeployImage` path runs — and with BMO-managed PPIs (defaultPPImgFormat), `setDeployImage` returns nil, so the **stale value persists indefinitely**.

3. **No other provisioning phase manages it**: Only `Register()` → `configureNode()` touches `kernel_append_params`. The `Provision`, `Prepare`, `Deprovision`, `InspectHardware`, and `Service` methods use data types that don't carry `PreprovisioningExtraKernelParams`. If anything outside BMO modifies the Ironic node's driver_info and removes `kernel_append_params`, it won't be re-set until the next Register call where the update condition is true (credentials/address change, or params become non-default again).

---

### Other logical issues

**Leading space on combination** (baremetalhost_controller.go:96-99): When BMH params is empty and PPI params is `"debug=1"`, the result is `" debug=1"` (leading space). This survives into `ManagementAccessData.PreprovisioningExtraKernelParams`. It happens to work because `fmtPreprovExtraKernParams` trims it later, but the intermediate value is incorrect and the test enshrines this as expected behavior.

**Whitespace-only BMH params not normalized** (baremetalhost_controller.go:86-87): `trimmedParams` is computed but only used for the ISO warning check. The actual concatenation on line 98 uses the untrimmed `kernelExtraPreprovParams`. So `"   "` BMH params + `"debug=1"` PPI params → `"    debug=1"`. Again, `fmtPreprovExtraKernParams` trims it, but the intermediate value is wrong.

**`kernel_append_params` unconditionally added to all BMC drivers**: Every BMC driver (IPMI, Redfish, Redfish-VirtualMedia, iDRAC-VirtualMedia, Redfish-HTTPS) now always includes `kernel_append_params` in `DriverInfo()`. For virtual media drivers that boot from ISO, this field is meaningless and could potentially confuse certain Ironic driver implementations. Previously, kernel params were only set via `setDeployImage` where the format was already known. Now they're set blindly in all cases.

**Redundant nil check** (baremetalhost_controller.go:84): After `if preprovImg == nil { return ... }`, there's `if preprovImg != nil {` — this is always true at that point.

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -807,13 +860,20 @@ func (r *BareMetalHostReconciler) getPreprovImage(ctx context.Context, info *rec
 		return nil, err
 	}
 
+	PPIFormat := preprovImage.Status.Format
+	if !isExternal {
+		// PPI was generated by the default PPI provider thus it has to be
+		// handled separately from regular formats.
```
dtantsur, line 866: Why, by the way? See the top-level comment, AI has found a potential bug in its handling, but I'm also not sure why it's even needed?

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -776,6 +820,15 @@ func (r *BareMetalHostReconciler) getPreprovImage(ctx context.Context, info *rec
 		return nil, fmt.Errorf("failed to retrieve pre-provisioning image data: %w", err)
 	}
 
+	// Check who reconciles the given PPI, if it is not this BMO then it is
```
zaneb, line 823: This code checks nothing of the sort. We _always_ set the controller reference when the PPI is created (although we shouldn't).

internal/controller/metal3.io/baremetalhost_controller.go
zaneb: I don't understand what situation this is trying to handle (as noted above, the implementation does not match the one that is described in the comments), but we should never treat a k8s API as having different meanings depending on who is using it. That's a pretty fundamental violation of what a k8s API is.

internal/controller/metal3.io/preprovisioningimage_controller.go
```diff
@@ -140,49 +140,69 @@ func configChanged(img *metal3api.PreprovisioningImage, format metal3api.ImageFo
...
+}
+
 // update updates the PreprovisioningImage resource with the latest image information.
 // It checks if the image configuration has changed since the last update, and if so,
 // it discards the existing image, sets up the new image data, and triggers a new
 // image build. If the image configuration has not changed, it simply updates the
 // image status with the latest information.
-func (r *PreprovisioningImageReconciler) update(ctx context.Context, img *metal3api.PreprovisioningImage, log logr.Logger) (bool, error) {
-	generation := img.GetGeneration()
+func (r *PreprovisioningImageReconciler) update(ctx context.Context, ppi *metal3api.PreprovisioningImage, log logr.Logger) (bool, error) {
```
zaneb, line 168: If you're gonna rename variables can you at least do it in a separate commit so that we can review the substantive changes independently?

internal/controller/metal3.io/preprovisioningimage_controller.go
```diff
@@ -140,49 +140,69 @@ func configChanged(img *metal3api.PreprovisioningImage, format metal3api.ImageFo
 		img.Status.NetworkData == networkDataStatus)
 }
 
+// findOwnerBMH checks if the owner of a PreProvisioningImage is a BMH and
+// in case the owner is a BMH then the BMH will be returned.
+func (r *PreprovisioningImageReconciler) findOwnerBMH(ctx context.Context, ppi *metal3api.PreprovisioningImage) (*metal3api.BareMetalHost, error) {
+	ownerRef := metav1.GetControllerOf(ppi)
```
zaneb, line 146: Owner references are not how BMHs are linked to PreprovisioningImages. They're linked by having the same name and being in the same namespace.

Just in general, if we're doing anything at all with owner references other than using them to ensure that dangling objects are cleaned up, we should stop it.

internal/controller/metal3.io/preprovisioningimage_controller.go
```diff
@@ -193,23 +213,34 @@ func (r *PreprovisioningImageReconciler) update(ctx context.Context, img *metal3
...
+			log.Info("image build failed", "error", failure)
+			return setError(generation, &ppi.Status, reasonImageBuildInvalid, failure.Error()), nil
 		}
 		return false, err
 	}
+
+	if format == metal3api.ImageFormatInitRD {
+		// Potentially might be other PPI owners in the future when
+		// multi tenancy support is fully implemented
+		bmh, err := r.findOwnerBMH(ctx, ppi)
```
zaneb, line 233: I don't like this. We have gone to great lengths previously to ensure that the PreprovisioningImage has its own API and its controller does not need to load any other resources. Having to retrieve the BMH as well is inefficient, and the resulting API spread across two objects is poor.

It's also incorrect. We are not watching the BMH resource, so the controller will not respond to changes in the PreprovisioningExtraKernelParams. And even if it did (which would be even more inefficient), the BMH controller would not be able to tell whether the image is up to date because the Generation is not incremented.

We need to bite the bullet and add this field to the PreprovisioningImage spec as previously discussed so that the API is explicit.

pkg/hardwareutils/bmc/access.go
zaneb: The AccessDetails interface is used outside of BMO in code that vendors hardwareutils as a library (it has [its own go module](https://github.com/metal3-io/baremetal-operator/blob/main/pkg/hardwareutils/go.mod) and everything), so it is undesirable to modify it. And in this case also completely unnecessary.

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -84,6 +85,49 @@ func (info *reconcileInfo) publishEvent(reason, message string) {
...
+		return ""
+	}
+	kernelExtraPreprovParams := info.host.Spec.PreprovisioningExtraKernelParams
+
+	if preprovImg == nil {
+		return kernelExtraPreprovParams
+	}
+	// If the execution reaches this point it means there is a PPI in the
+	// namespace with acceptable image format but it is managed by an external
+	// controller-manager.
```
zaneb, line 100: It definitely doesn't mean that.

It appears to me that execution _always_ reaches this point, regardless of whether the ppimg is considered to be managed by an 'external' controller, which is not really a thing we are checking for.

internal/controller/metal3.io/baremetalhost_controller.go
zaneb: I don't see how we can hit this path. If there's no host, like, literally what are we even reconciling? So I think this is safe enough.

internal/controller/metal3.io/baremetalhost_controller.go
zaneb, line 113: I don't think this should be phrased as a warning. We've provided them to the PreprovisioningImage controller. It's up to it to decide what to do with them.

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -84,6 +85,49 @@ func (info *reconcileInfo) publishEvent(reason, message string) {
...
+			kernelExtraPreprovParams = ""
+		case metal3api.ImageFormatInitRD:
+			trimmedImageParams := strings.TrimSpace(preprovImg.GeneratedImage.ExtraKernelParams)
+			if trimmedImageParams != "" {
+				kernelExtraPreprovParams += " " + trimmedImageParams
+			}
+			info.log.Info("BMH and external PPI (preprovisioning)ExtraKernelParams got combined!")
+		default:
+			// This scenario covers when the image format is internally set to defaultPPIImgFormat.
+			info.log.Info("PPI is generated by BMO or has unknown image format, preprovisioningExtraKernelParams are taken only from BMH spec.")
```
zaneb, line 125: Even taking the code for deciding whether something is 'externally' managed, this makes no sense.

If we know that the ppimg is managed by the built-in controller then we know that it will have copied the params to the status if required. So there's no need for this whole `defaultPPIImgFormat` case.

internal/controller/metal3.io/baremetalhost_controller.go
Rozzii: Old comment, logic was referring to code that was removed, so I will remove this.

internal/controller/metal3.io/baremetalhost_controller.go
Rozzii: Okay to both of these, will change.

internal/controller/metal3.io/baremetalhost_controller.go
Rozzii: Yeah sure, if you all think this can never happen I will just get rid of this.

internal/controller/metal3.io/preprovisioningimage_controller.go
Rozzii: Okay will change it back.

Rozzii:
Okay I need to write a review summary because there are a lot of new design requirements that I have learned from your review @zaneb and @dtantsur .
- there are misc stuff like my temp comments left in code and some logging, reverting PPI renaming etc no big deal I will sort those out.
- kernel_param_apped is persistent so I will investigate and resolve the issue of that value being overwritten,will test it with older BMO versions because I think I have seen this behavior already when I started the development that is why I went with always overwriting the field. I know there is path in the new code that can do this, but it is happening all the time and not just being overwritten but completely removed. I will fix this. 
- kernel params have to stay away from access details because it is used as a lib : Okay I will store it in ManagementAccesData or somewhere else so it won't get mixed in with the accessDetails .
- the default PPI controller should not check any other CR so I can't look up owners - okay will rethink the approach
- you don't like the default PPI image format that signals it was made by the default ppi controller - okay I will rethink the approach

Thank you for your great reviews, I learned a lot, I will get back with the next attempt.

Rozzii:
> PR needs rebase.
> Details

noted, it is on the back burner at the moment but I will return to this PR as soon as I have time

s3rj1k:
Suggestion to move this further faster, split in 2 PRs, one is exclusively about extending API surface with no implementation and second is implementation itself.

PR #2577, opened 2025-07-25
:bug: Add IPv6 address check to run local

nuhakala:
After adding IPv6 support to dev env, it is possible that the used IP addresses are v6. In this case the address needs to be surrounded by brackets when using as a host address. This PR adds a check for IPv6 address and proper surrounding to IPv6 addresses in the `run_local_ironic.sh`.

Fixes #

### DIFF
--- a/tools/run_local_ironic.sh
+++ b/tools/run_local_ironic.sh
@@ -15,7 +15,14 @@ IRONIC_DATA_DIR=${IRONIC_DATA_DIR:-"/opt/metal3-dev-env/ironic"}
 CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-podman}"
 HTTP_PORT=${HTTP_PORT:-"6180"}
 PROVISIONING_IP="${PROVISIONING_IP:-"172.22.0.1"}"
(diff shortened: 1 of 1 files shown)


lentzi90 (commented):
/approve

tools/run_local_ironic.sh
```diff
@@ -16,6 +16,11 @@ CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-podman}"
 HTTP_PORT=${HTTP_PORT:-"6180"}
 PROVISIONING_IP="${PROVISIONING_IP:-"172.22.0.1"}"
 CLUSTER_PROVISIONING_IP="${CLUSTER_PROVISIONING_IP:-"172.22.0.2"}"
```
Rozzii, line 20: Add a comment here please and mention that this is needed for IPv6 addresses because if someone will see this at the first time they might nit realize what is happening.

Rozzii (commented):
Small nit.

tools/run_local_ironic.sh
nuhakala: Done

Rozzii (commented):
/lgtm

PR #2590, opened 2025-08-01
🌱 E2E: Set static IPs through user-data

lentzi90:

We have been "cheating" a bit in CI by configuring static IPs for the VMs through libvirt dhcp. This is not something we can rely on for real hardware. For that, we would either need to manually configure DHCP or make use of cloud-init to configure the IPs. This commit attempts to do the latter.

Fixes #

### DIFF
--- a/hack/e2e/net.xml
+++ b/hack/e2e/net.xml
@@ -8,7 +8,7 @@
   <bridge name='metal3'/>
   <ip address='192.168.222.1' netmask='255.255.255.0'>
     <dhcp>
-      <range start='192.168.222.2' end='192.168.222.199'/>
+      <range start='192.168.222.2' end='192.168.222.99'/>
       <bootp file='http://192.168.222.1:6180/boot.ipxe'/>
     </dhcp>
   </ip>
--- a/test/e2e/automated_cleaning_test.go
+++ b/test/e2e/automated_cleaning_test.go
@@ -146,7 +146,7 @@ var _ = Describe("Automated cleaning", Label("required", "automated-cleaning"), 
 		By("Patching the BMH again to trigger re-provisioning")
 		userDataSecretName = "user-data-ssh-setup"
 		// Create new userdata secret for only SSH setup
-		createSSHSetupUserdata(ctx, clusterProxy.GetClient(), namespace.Name, userDataSecretName, sshPubKeyPath)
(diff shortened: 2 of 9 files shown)


lentzi90:
Hmm you mean some invalid metadata or making ssh check to a free IP address to see that it fails?
We don't have anything like that at the moment. It would certainly be good to expand the tests with more metadata/userdata/networkdata checks. We don't have any test that checks what actually ends up on the disk compared to what was in the secret

tuminoid:
> Hmm you mean some invalid metadata or making ssh check to a free IP address to see that it fails? We don't have anything like that at the moment. It would certainly be good to expand the tests with more metadata/userdata/networkdata checks. We don't have any test that checks what actually ends up on the disk compared to what was in the secret

Something along those lines yes. Care to make an issue, since you probably have better grasp what we could add? We seem to have few new contributors looking at issues and adding tests right now, so having them documented in an issue would be lovely.

lentzi90:
Yes I'll create an issue!

lentzi90:
Issue created: https://github.com/metal3-io/baremetal-operator/issues/2626

Rozzii:
@lentzi90 what is the reason  that made you chose user-data instead of network-data?
Naively I would have thought network-data would be the place for setting static IPs.

lentzi90:
Good question! Perhaps we should actually use network-data now that I think about it...
Anyway, I have here just taken the easiest way forward and added it directly in the user-data. I think that we would need to switch to some other test image if we want to use network-data, because it seems like [cirros doesn't really support it](https://github.com/cirros-dev/cirros/issues/67). They only support this executable script format of user-data also.
It would be a good idea to investigate this a bit deeper actually. It would be very good to test network-data and perhaps also some more advanced user-data use-cases.

Rozzii:
> Good question! Perhaps we should actually use network-data now that I think about it... Anyway, I have here just taken the easiest way forward and added it directly in the user-data. I think that we would need to switch to some other test image if we want to use network-data, because it seems like [cirros doesn't really support it](https://github.com/cirros-dev/cirros/issues/67). They only support this executable script format of user-data also. It would be a good idea to investigate this a bit deeper actually. It would be very good to test network-data and perhaps also some more advanced user-data use-cases.

Thanks for the info, this explains it, I forgot cirros had no support for network data.

IMO in the long term it would be better to use some small ubuntu or cnetos image for these tests. Or at least something that has proper cloud-init openstack format support.

PR #2598, opened 2025-08-04
🌱 Bump osv-scanner to v2.1.0

kashifest:
- Bumps in the verify-release.sh
- Bumps in the osv-scanner workflow

### DIFF
--- a/hack/verify-release.sh
+++ b/hack/verify-release.sh
@@ -147,11 +147,11 @@ if [[ -n "${CONTAINER_RUNTIME}" ]]; then
         "${CONTAINER_RUNTIME}" run --rm
         -v "${PWD}":"/src:ro,z"
         -w /src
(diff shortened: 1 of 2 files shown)


hack/verify-release.sh
```diff
@@ -213,7 +213,7 @@ check_tools()
         case "${tool}" in
             osv-scanner)
                 version=$("${OSVSCANNER_CMD[@]}" -v | grep version | cut -f3 -d" ")
-                min_version="2.0.0"
+                min_version="2.1.0"
```
tuminoid, line 216: FYI: Min version doesn't have to be bumped unless old version is actually incompatible, ie. different flags etc. It was done for v1 vs v2 being different. But since this same bump is applied to other repos, let's take it in as is.

tuminoid (commented):
/lgtm
/cc @lentzi90

lentzi90 (commented):
/approve

PR #2605, opened 2025-08-08
🌱 Auto-generate API documentation from CRDs

### LINKED ISSUE #2509: Generate local API documentation from CRDs
We have some old `docs/api.md` that explains some things and points to external service for API documentation.

Consider taking in https://github.com/metal3-io/ironic-standalone-operator/pull/281 from IRSO for local API documentation and check what parts of the old documentation to keep and what to merge.

/kind documentation

MahnoorAsghar:
Auto-generate API documentation from CRDs

Assisted-by: Claude 
Fixes #2509

### DIFF
--- a/hack/gen-api-doc.sh
+++ b/hack/gen-api-doc.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+set -eux
+
+IS_CONTAINER="${IS_CONTAINER:-false}"
+CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-podman}"
+CRDOC_VERSION="0.6.2@sha256:355ef777a45021ee864e613b2234b4f2c6193762e3e0de94a26b66d06cec81c3"
+WORKDIR="${WORKDIR:-/workdir}"
+
+if [ "${IS_CONTAINER}" != "false" ]; then
+    # Running inside a container (e.g., from generate.sh)
+    # Skip API doc generation as it's not part of the validation checksums
+    # and would require nested containers which may not be supported
+    echo "Skipping API doc generation inside container (not part of validation)"
+    exit 0
+fi
+
+# Run crdoc in its own container
+"${CONTAINER_RUNTIME}" run --rm \
+    --volume "${PWD}:${WORKDIR}:ro,z" \
+    --workdir "${WORKDIR}" \
+    ghcr.io/fybrik/crdoc:"${CRDOC_VERSION}" \
+    --resources /workdir/config/base/crds/bases \
+    --output /dev/stdout \
+    > "${PWD}"/docs/api.md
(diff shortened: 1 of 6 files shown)


tuminoid:
This is copy from IRSO's api doc generator right?

MahnoorAsghar:
@tuminoid 

> This is copy from IRSO's api doc generator right?

Yes, it is

hack/ci-e2e.sh
```diff
@@ -54,7 +54,7 @@ export PATH="/usr/local/go/bin:${PATH}"
 "${REPO_ROOT}/hack/e2e/ensure_yq.sh"
 
 sudo apt-get update
-sudo apt-get install -y libvirt-dev pkg-config
+sudo apt-get install -y libvirt-dev pkg-config podman
```
tuminoid: This is just for e2e, not for generate.

hack/gen-api-doc.sh
```diff
@@ -0,0 +1,11 @@
...
+
+set -eu
+
+CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-podman}"
+CRDOC_VERSION="0.6.2@sha256:355ef777a45021ee864e613b2234b4f2c6193762e3e0de94a26b66d06cec81c3"
+
+"${CONTAINER_RUNTIME}" run --rm -v "${PWD}":/src:ro,z \
+    ghcr.io/fybrik/crdoc:"${CRDOC_VERSION}" \
+    --resources /src/config/base/crds/bases/ --output /dev/stdout \
+    > docs/api.md
```
tuminoid: missing eol

tuminoid (commented):
The difference between this BMO PR and [IRSO](https://github.com/metal3-io/ironic-standalone-operator/pull/281/files) here is that IRSO runs the "make manifests" as part of build Github workflow (in a VM), while BMO runs make generate as a Prow job (in a container).

The correct way for BMO would be to have it in a separate Prow job, which runs the crdoc container in Prow.

hack/ci-e2e.sh
MahnoorAsghar: Forgot to remove this, ack

MahnoorAsghar:
I dont remember how docs/api-overview.md came from, it must be output from the script, but executing it right now does not create it.

MahnoorAsghar:
For the check-links-pr job, https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.24/#objectmeta-v1-meta is the correct url, but the crdoc container is generating v1.20 links which are indeed broken

This could be because the latest crdoc container version is 3 years old. There are 2 tagged releases on the repo after that, but no associated container images. See https://github.com/fybrik/crdoc/issues/322

MahnoorAsghar:
I could add the exclude-path option to https://github.com/metal3-io/baremetal-operator/blob/main/.github/workflows/pr-link-check.yml and https://github.com/metal3-io/project-infra/blob/1f19a2b8b926ab9bec62b1b8099cecf492d2125d/.github/workflows/pr-link-check.yml#L64
to ignore the path failures :D

PR #2612, opened 2025-08-11
🌱 Add continue-on-error and improve Slack notifications for osv-scanner

lentzi90:

The osv-scanner workflow should continue on error when uploading SARIF files and send Slack notifications for both vulnerabilities and workflow failures.

This ensures we get notified of any issues with the workflow.

Fixes #


kashifest (commented):
/approve

PR #2614, opened 2025-08-13
🐛 Allow setting `force_persistent_boot_device: <value>` for deploy_interface: direct

### LINKED ISSUE #2613: [ironic] force_persistent_boot_device can only be configured from env, when mode `ramdisk` (deployIso) is used
I found the env variable `LIVE_ISO_FORCE_PERSISTENT_BOOT_DEVICE` which can configure the persistent boot device flag for all bmhs a bmo manages. Unfortunately the flag only works as indicated by the name for Live ISO or deploy_interface `ramdisk`.
As you could also run a Live ISO like system using the `direct` deploy_interface, the flag should also allow overwriting for this deploy_interface.

The variable can configure both deploy_interfaces and overwrite `driver_info: force_persistent_boot_device: "Default"` when configured to be `Never` or `Always`.

[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Baremetal Operator version: v0.9.1
- Environment (metal3-dev-env or other): production deloyment

/kind bug

Cellebyte:

It allows configuring the `force_persistent_boot_device: <value>` when `deploy_interface: direct` is used.
It uses the same variable as for `deploy_interface: ramdisk.`

I did not rename the current env var `LIVE_ISO_FORCE_PERSISTENT_BOOT_DEVICE` to be backwards compatible.
It introduces a new environment variable `DIRECT_DEPLOY_FORCE_PERSISTENT_BOOT_DEVICE` which can be toggled when using the direct interface. 
The default is the same behaviour as before, so people need to actively **opt-in**.

Fixes #2613

### DIFF
--- a/pkg/provisioner/ironic/factory.go
+++ b/pkg/provisioner/ironic/factory.go
@@ -101,6 +101,7 @@ func (f *ironicProvisionerFactory) init(havePreprovImgBuilder bool) error {
 		"deployRamdiskURL", f.config.deployRamdiskURL,
 		"deployISOURL", f.config.deployISOURL,
 		"liveISOForcePersistentBootDevice", f.config.liveISOForcePersistentBootDevice,
+		"directDeployForcePersistentBootDevice", f.config.directDeployForcePersistentBootDevice,
 		"CACertFile", tlsConf.TrustedCAFile,
 		"ClientCertFile", tlsConf.ClientCertificateFile,
 		"ClientPrivKeyFile", tlsConf.ClientPrivateKeyFile,
@@ -197,11 +198,17 @@ func loadConfigFromEnv(havePreprovImgBuilder bool) (ironicConfig, error) {
 		c.maxBusyHosts = value
 	}
 
-	if forcePersistentBootDevice := os.Getenv("LIVE_ISO_FORCE_PERSISTENT_BOOT_DEVICE"); forcePersistentBootDevice != "" {
-		if forcePersistentBootDevice != "Default" && forcePersistentBootDevice != "Always" && forcePersistentBootDevice != "Never" {
+	if liveISOForcePersistentBootDevice := os.Getenv("LIVE_ISO_FORCE_PERSISTENT_BOOT_DEVICE"); liveISOForcePersistentBootDevice != "" {
+		if liveISOForcePersistentBootDevice != "Default" && liveISOForcePersistentBootDevice != "Always" && liveISOForcePersistentBootDevice != "Never" {
 			return c, errors.New("invalid value for variable LIVE_ISO_FORCE_PERSISTENT_BOOT_DEVICE, must be one of Default, Always or Never")
 		}
-		c.liveISOForcePersistentBootDevice = forcePersistentBootDevice
+		c.liveISOForcePersistentBootDevice = liveISOForcePersistentBootDevice
+	}
+	if directDeployForcePersistentBootDevice := os.Getenv("DIRECT_DEPLOY_FORCE_PERSISTENT_BOOT_DEVICE"); directDeployForcePersistentBootDevice != "" {
+		if directDeployForcePersistentBootDevice != "Default" && directDeployForcePersistentBootDevice != "Always" && directDeployForcePersistentBootDevice != "Never" {
+			return c, errors.New("invalid value for variable DIRECT_DEPLOY_FORCE_PERSISTENT_BOOT_DEVICE, must be one of Default, Always or Never")
+		}
+		c.directDeployForcePersistentBootDevice = directDeployForcePersistentBootDevice
 	}
 
 	c.externalURL = os.Getenv("IRONIC_EXTERNAL_URL_V6")
--- a/pkg/provisioner/ironic/factory_test.go
+++ b/pkg/provisioner/ironic/factory_test.go
@@ -10,16 +10,17 @@ import (
 )
 
 type EnvFixture struct {
-	ironicEndpoint                   string
-	kernelURL                        string
-	ramdiskURL                       string
-	isoURL                           string
-	liveISOForcePersistentBootDevice string
-	ironicCACertFile                 string
-	ironicClientCertFile             string
-	ironicClientPrivateKeyFile       string
-	ironicInsecure                   string
-	ironicSkipClientSANVerify        string
+	ironicEndpoint                        string
+	kernelURL                             string
+	ramdiskURL                            string
+	isoURL                                string
+	liveISOForcePersistentBootDevice      string
+	directDeployForcePersistentBootDevice string
+	ironicCACertFile                      string
+	ironicClientCertFile                  string
+	ironicClientPrivateKeyFile            string
+	ironicInsecure                        string
+	ironicSkipClientSANVerify             string
 
 	origEnv map[string]string
 }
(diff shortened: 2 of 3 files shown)


Cellebyte:
@tuminoid is this error in the test related to my change?

tuminoid:
> @tuminoid is this error in the test related to my change?

Not sure, do not have time right now to dig into. So, let's rerun and if it fails to the same, then very likely yes :)  
/retest

Cellebyte:
@tuminoid thank you for rerunning, seems to look good :)

tuminoid:
> @tuminoid thank you for rerunning, seems to look good :)
> 
> 

The prow keywords work for anyone after initial ok-to-test btw. Just in case you need to make some revisions and want/need to run some tests.

Cellebyte:
Good to know :)  THX

tuminoid (commented):
/cc @lentzi90 @Rozzii @elfosardo @dtantsur 

PTAL

lentzi90:
I don't know this well enough to review properly, just noting that this may count as :sparkles: instead of :bug: maybe?

Cellebyte:
@lentzi90 I am not sure if it counts as ✨ .
I think I need to still add tests for ther kernel URL ramdisk URL part :)

Cellebyte:
So the tests are added. And everything is green :)

tuminoid:
> So the tests are added. And everything is green :)

Yay :) Can you squash the commit please.

Cellebyte:
I can :) Will do that today :)

Cellebyte:
Squashed the commits :)

Rozzii:
/hold
Keep it on hold until 0.11 goes out, I don't think we need to rush this into 0.11.

tuminoid:
/cc @zaneb @dtantsur @elfosardo 
Let's get more eyes on this meanwhile.

Cellebyte:
Can this be merged?

zaneb:
> If people rely on the behaviour, it could break hosts using deploy_interface: direct after this MR is implemented as it is currently drafted.

We do currently rely on this behaviour [in OpenShift](https://github.com/openshift/cluster-baremetal-operator/blob/423685e8c2734c2743619005ce4eb91f382f852a/provisioning/bmo_pod.go#L149-L152). So I assume that this change will break us and 100% of other people who are using the environment variable (if any).

In the bug you said that "you could also run a Live ISO like system using the direct deploy_interface" but _could_ means you also could _not_ do that, so setting it in an environment variable where it affects every BMH sounds like almost certainly the wrong way to configure it. For live ISO, users opt in to the configuration by selecting the `live-iso` image format. (TBH I have no idea why it is even configurable and not baked in to `live-iso` handling.)

Cellebyte:
I mean I could also add another environment variable which configures it for the direct interface. This would not break it for users which use both ways for their BMHs and only want to set it for live-iso.

Rozzii:
/hold cancel
Removing the hold because 0.11 has been released.

> I mean I could also add another environment variable which configures it for the direct interface. This would not break it for users which use both ways for their BMHs and only want to set it for live-iso.

This sounds reasonable IMO.

Rozzii (commented):
The code looks good but it will be reworked to provide separately configurable persistent boot device behavior.

Cellebyte:
@Rozzii I will update this PR to accomodate for it :)

Cellebyte:
PR is updated to accomodate for the feedback.

Cellebyte:
@Rozzi can this one be merged?

MahnoorAsghar:
I think the PR description needs to be updated about renaming the LIVE_ISO_FORCE_PERSISTENT_BOOT_DEVICE env variable, which is now backwards-compatible

pkg/provisioner/ironic/factory.go
```diff
@@ -134,11 +135,17 @@ func loadConfigFromEnv(havePreprovImgBuilder bool) (ironicConfig, error) {
...
-	if forcePersistentBootDevice := os.Getenv("LIVE_ISO_FORCE_PERSISTENT_BOOT_DEVICE"); forcePersistentBootDevice != "" {
-		if forcePersistentBootDevice != "Default" && forcePersistentBootDevice != "Always" && forcePersistentBootDevice != "Never" {
+	if liveISOForcePersistentBootDevice := os.Getenv("LIVE_ISO_FORCE_PERSISTENT_BOOT_DEVICE"); liveISOForcePersistentBootDevice != "" {
+		if liveISOForcePersistentBootDevice != "Default" && liveISOForcePersistentBootDevice != "Always" && liveISOForcePersistentBootDevice != "Never" {
 			return c, errors.New("invalid value for variable LIVE_ISO_FORCE_PERSISTENT_BOOT_DEVICE, must be one of Default, Always or Never")
 		}
-		c.liveISOForcePersistentBootDevice = forcePersistentBootDevice
+		c.liveISOForcePersistentBootDevice = liveISOForcePersistentBootDevice
+	}
+	if kernelAndRamdiskForcePersistentBootDevice := os.Getenv("KERNEL_AND_RAMDISK_FORCE_PERSISTENT_BOOT_DEVICE"); kernelAndRamdiskForcePersistentBootDevice != "" {
```
zaneb: What would you say if I told you that it's possible to provision an ISO without using `diskFormat: live-iso`? 😅 

I think this should be called `DIRECT_DEPLOY_FORCE_PERSISTENT_BOOT_DEVICE`.

Cellebyte: Now I am confused.
It got approved and the Environment Variable needs now a rename?
I am introducing the possibility to do it and I don't know if users understand what direct deploy even means.

Cellebyte: But I could rename it again.
This is taken a lot of effort for a small change ^^

Rozzii: Don't worry approve will stay if yo rename it. But I am impartial in the question of renaming.

Cellebyte: Okay, I mean I could just rename it to the proposed name. If it is the final name. ^^

zaneb: > It got approved and the Environment Variable needs now a rename?

Yes.

> I am introducing the possibility to do it and I don't know if users understand what direct deploy even means.

I think you're probably right, but I'm even more certain that they won't understand that the KERNEL_AND_RAMDISK part means that it applies to Raw, QCOW2, VDI, and VDMK deployments, but has nothing to do with kernels+ramdisks (which is not even a format that we support for provisioning), and that it is applies regardless of whether the pre-provisioning image is supplied as a kernel+ramdisk or an ISO.

> This is taken a lot of effort for a small change

Changing interfaces is hard because people rely on them.

> If it is the final name.

I'm open to other suggestions but, as you said in the description of this PR, this value applies "when `deploy_interface: direct` is used". so `DIRECT_DEPLOY_FORCE_PERSISTENT_BOOT_DEVICE` is, if nothing else, straightforwardly accurate about what it does.

Cellebyte: @zaneb Done :)

Cellebyte:
@zaneb is this now ok to merge? :)

zaneb:
This seems correctly-named and shouldn't break anybody now.

Cellebyte:
Can this be merged into the v0.12 + main?

Cellebyte:
Why was this not merged into 0.12, but the tag already exists?

Rozzii:
> Why was this not merged into 0.12, but the tag already exists?

My approval still stands, this PR still needs an "lgtm" label from one other reviewer. 0.12 was released without this because of no one gave "lgtm". I can't give you "lgtm" because I have already added the "approve" label.

This work can't be merged to 0.12 , I will change the miletone now.

@zaneb would you please lgtm this?

Rozzii:
@lentzi90 @dtantsur could you please take a look?

Cellebyte:
Ok I was just confused as it was part of v0.12 Milestone. Now with the changed Milestone I am fine. Thanks @Rozzii for the clarification 👍

PR #2615, opened 2025-08-14
🌱 Enable errname linter

### LINKED ISSUE #2353: Enable "errname" linter in golangci-lint
Enable "errname" linter in golangci-lint. Fix any warnings enabling it causes.

/area ci
/help
/triage accepted

nerok:
This enables the errname lint rule which just forces all Errors to be named something like "xxxError".

This is a naive attempt to enable errname. I just added Error to the end of the structs which it complained about, not sure if this is a desired solution. `ImageBuildInvalidError` should maybe just be `ImageBuildError`?

Fixes #2353


tuminoid (commented):
/lgtm

Thanks for the contribution!

tuminoid (commented):
/cc @kashifest @elfosardo

internal/controller/metal3.io/preprovisioningimage_controller.go
```diff
@@ -113,7 +113,7 @@ func (r *PreprovisioningImageReconciler) Reconcile(ctx context.Context, req ctrl
 		result.RequeueAfter = delay
 	}
 
-	notReady := imageprovider.ImageNotReady{}
```
Rozzii, line 116: This could be a breaking change, imageproviders and in practice preprov image controllers are implemented by the users e.g https://github.com/openshift/image-customization-controller/blob/main/pkg/imageprovider/rhcos.go

So external controllers rely on errors specified here e.g https://github.com/openshift/image-customization-controller/blob/main/pkg/imageprovider/rhcos.go#L97

I think we should discuss this more widely with the community.
I don't think a  formal proposal is needed.

zaneb: Yeah, we actually use this controller as a library (or did, until it got moved to `internal/`), and the imageprovider package is definitely designed to be used with it as a library.
So this is a breaking change to a library. That said, it's not a particularly difficult one to fix, so I am fairly relaxed about it. FWIW OpenShift's image provider only uses the `ImageBuildInvalid` type via the `BuildInvalidError()` function (I'd expect that's true for anyone) and doesn't use `ImageNotReady` at all because our images are always ready, so this change wouldn't break us. I don't know if there are any other users out there.

The way to prevent a breaking change for anybody is to add:

```
type ImageNotReady = ImageNotReadyError
```

But I don't know if that would pass the linter or not.

nerok: That does indeed fail the linter, but it is fine with a simple //nolint:errname, since they are already deprecated I don't think that is a problem. I'm not sure if it is enough to add the `type ImageNotReady = ImageNotReadyError` or if all the other functions should be kept duplicated (and of course deprecated) too. I'll push a fix without the functions duplicated now, just let me know if I should duplicate the functions too.

Rozzii (commented):
/approve

PR #2599, opened 2025-08-06
chore(ci): update github runners to oci gh arc runners

koksay:
**What this PR does / why we need it**:

CNCF has hosted ephemeral GitHub runners on Oracle that we want projects to use instead of the GitHub-hosted ones, which now incur a cost to use.

Please direct any questions to me, @jeefy, @krook, and @RobertKielty. You can also join `#cncf-ci-infra` channel on CNCF Slack.

PR #2621, opened 2025-08-19
Revert "🌱 CI: update github runners to oci gh arc runners"

lentzi90:
Reverts metal3-io/baremetal-operator#2599

/hold
This is to check if the new runners are what is causing issues


tuminoid:
OK, this proves the new runners are the culprit. New vm type in the new runners have been reverted, so this local revert may not be necessary. Let's keep the hold.

PR #2622, opened 2025-08-20
:seedling: Revert "update github runners to oci gh arc runners"

tuminoid:
This reverts commit dee17fb8c6af91e346368fc7b1a9b6cecc90c1ff / #2599 for now.

New Oracle runners are unstable, and CNCF is still looking for stable configuration. We have seen AMD instances that work, older AMD instances that lack kvm (nested virtualization) support etc.

To avoid too much going back and forth, let's revert the BMO main to unblock the development, and leave the BMO release branches as is, as most of the configurations do work with BMO well enough in release branch work.

I will also make new PR to revert this revert, so there we can re-run the tests and see when its stable again.


lentzi90 (commented):
/approve

PR #2623, opened 2025-08-20
:seedling: update github runners to oci gh arc runners

tuminoid:
Move to new CNCF Github runners, try nro 2. Previous try at PR #2599 and revertion in PR #2622. 

/hold
Until stable.

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -8,7 +8,7 @@ on:
         type: string
       runner:
         type: string
-        default: "ubuntu-latest-4-cores"
+        default: "oracle-vm-4cpu-16gb-x86-64"
       ginkgo-focus:
         type: string
         default: ""


tuminoid:
/retest
AMD E5 + faster storage enabled.

tuminoid:
```
          "message": "ReplicaSet \"ironic-645575d78c\" is progressing."
```

/retest

I'm thinking this is classic network speed issue, but it is not entirely runner's problem.

tuminoid:
OK, pass on 2nd run here.

tuminoid:
Rebased on top of latest.

tuminoid:
OK, BMO 0.9 and 0.10 branches have been using the new workers since the initial PR merged, and no failures. The one failure we had here was temporary due using VMs that did not support nested virtualization required by the e2e.

This reimplementation PR has not failed more than the current runner, this can be considered stable.

tuminoid:
/unhold
This has been stable and the only hiccup was the accidental change they made in the VM type. Runner itself has been stable for BMO all this time, as can be seen from the release-0.9 and release-0.10 periodics which have been consistently green all this time using the new runners.

Let's merge this as the BMO is our no1 most resource consuming repo thanks to higher PR number, and the fact that we have a patch for IRSO which makes that stable too. The network connection to opendev seems to be the IRSO instability problem, and we have patches for that too, and they're looking green in multiple runs in a row.

/cc @kashifest @Rozzii @dtantsur

Rozzii (commented):
/approve

PR #2625, opened 2025-08-21
🌱 Add validation in webhook for live-iso

nerok:
Add a validation for live-iso to require a bmc protocol (that is the part in the address before "://") which includes "virtualmedia", which seems to be what the requirement for live-isos are according to the docs: https://book.metal3.io/bmo/live-iso

This relates to #855


internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -386,3 +390,13 @@ func validatePowerStatus(host *metal3api.BareMetalHost) error {
 	}
 	return nil
 }
+
+func validateLiveISO(bmcAddress string, image *metal3api.Image) error {
+	if image != nil &&
+		image.DiskFormat != nil &&
+		*image.DiskFormat == "live-iso" &&
+		!strings.Contains(strings.Split(bmcAddress, "://")[0], "virtualmedia") {
+		return errors.New("Live-Iso can only be used with a virtualmedia bmc")
```
tuminoid:
```suggestion
		return errors.New("Live-ISO can only be used with a virtualmedia BMC")
```

internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -386,3 +390,13 @@ func validatePowerStatus(host *metal3api.BareMetalHost) error {
 	}
 	return nil
 }
+
+func validateLiveISO(bmcAddress string, image *metal3api.Image) error {
+	if image != nil &&
+		image.DiskFormat != nil &&
+		*image.DiskFormat == "live-iso" &&
+		!strings.Contains(strings.Split(bmcAddress, "://")[0], "virtualmedia") {
```
tuminoid: Can we use something like `hasPrefix("redfish-virtualmedia://")` , seems more robust?

tuminoid (commented):
Thanks for the contribution! Some minor comments left.

internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
nerok: Do we have a list somewhere over prefixes? because I found these in the docs:
- redfish-virtualmedia://
- idrac-virtualmedia://
- ilo5-virtualmedia://
- ilo4-virtualmedia://

and all of there can also have +http or +https.
I can probably use hasPrefix instead yeah

tuminoid: If you can make it work cleanly, sure, but if it gets complicated, then the current one is fine as well.

nerok: I think another look at this improved it a lot.
Still got the "contains", but I believe the right way there would be to add SupportsVirtualmedia() to the BMCs, and call those.
I could give that a go, but it is a bit outside of the scope of this PR, and I'm uncertain if we would actually need it anywhere else.

tuminoid: I'm good with this iteration.

internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -150,6 +150,10 @@ func validateBMCAccess(s metal3api.BareMetalHostSpec, bmcAccess bmc.AccessDetail
 		errs = append(errs, fmt.Errorf("BMC driver %s does not support secure boot", bmcAccess.Type()))
 	}
 
+	if s.Image != nil && s.Image.DiskFormat != nil && *s.Image.DiskFormat == "live-iso" && !strings.Contains(bmcAccess.Type(), "virtualmedia") {
```
zaneb: I'm not sure it's safe to make this assumption about the driver name.

I would expect that in practice `bmcAccess.SupportsISOPreprovisioningImage()` corresponds to exactly what we want to know. @dtantsur can you confirm?

nerok: That worked locally atleast, pushed it up in a new version

tuminoid: @zaneb @dtantsur can you check if you're happy with latest version?

tuminoid: /cc @zaneb @dtantsur

zaneb: /lgtm
The one where this gives a different answer is the UEFI HTTP driver, but I am not good enough at tracing the code through ironic to say one way or the other there whether live-iso should be allowed there. (If I had to guess though, I'd say yes and that this version is correct.)

internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -150,6 +150,10 @@ func validateBMCAccess(s metal3api.BareMetalHostSpec, bmcAccess bmc.AccessDetail
 		errs = append(errs, fmt.Errorf("BMC driver %s does not support secure boot", bmcAccess.Type()))
 	}
 
+	if s.Image != nil && s.Image.DiskFormat != nil && *s.Image.DiskFormat == "live-iso" && !bmcAccess.SupportsISOPreprovisioningImage() {
```
dtantsur, line 153: We tried this, and people complained. The thing is: technically you can boot an ISO over iPXE. It does not work with all ISO's, but apparently does work with some.

dtantsur (changes requested):

internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
nerok: Am I understanding you correct in that we shouldn't have this validation at all? Or is there a part of this validation that could be valuable? If not we should ensure it is documented, and I would think a test would be nice to ensure we don't break anything at a later point?

PR #2636, opened 2025-08-26
:seedling: Ensure BMH name is a valid DNS label

nerok:
We don't want anyone to create a BMH with a name which isn't a compliant DNS hostname, that will cause issues.

Fixes parts of #855

### DIFF
--- a/internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
+++ b/internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
@@ -20,7 +20,6 @@ import (
 	"fmt"
 	"net"
 	"net/url"
-	"regexp"
 	"slices"
 	"strings"
 
@@ -30,6 +29,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	k8svalidation "k8s.io/apimachinery/pkg/util/validation"
 )
 
 var (
@@ -61,7 +61,7 @@ func (webhook *BareMetalHost) validateHost(host *metal3api.BareMetalHost) []erro
 	errs = append(errs, validateBMCAccess(host.Spec, bmcAccess)...)
 
 	if err := validateBMHName(host.Name); err != nil {
-		errs = append(errs, err)
+		errs = append(errs, err...)
 	}
 
 	if err := validateDNSName(host.Spec.BMC.Address); err != nil {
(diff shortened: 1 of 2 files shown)


internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -184,9 +184,9 @@ func validateRAID(r *metal3api.RAIDConfig) []error {
 }
 
 func validateBMHName(bmhname string) error {
-	invalidname, _ := regexp.MatchString(`[^A-Za-z0-9\.\-\_]`, bmhname)
+	invalidname, _ := regexp.MatchString(`[^A-Za-z0-9\-]`, bmhname)
```
zaneb: There is more to validating a DNS name than this; e.g. it cannot start or end with a `-`, should be no more than 63 characters.

I suggest calling [this function](https://github.com/kubernetes/apimachinery/blob/f5dd29d6ada12819a4a6ddc97d5bdf812f8a1cad/pkg/util/validation/validation.go#L188-L203) to validate it.

nerok: Changed to that function now, some tests needed changes, and otherwise I added (maybe too many) tests to validate that we check what we think we check.

tuminoid (commented):
/lgtm
/cc @dtantsur @zaneb 

Btw, can we break existing installations by validating and failing this now, or did we have some other runtime check doing the same?

zaneb:
> Btw, can we break existing installations by validating and failing this now, or did we have some other runtime check doing the same?

CR names are actually already [validated](https://github.com/kubernetes/kubernetes/blob/d92a443ca7867e0d2c1047b0502a7da42f1ac5e7/staging/src/k8s.io/apiextensions-apiserver/pkg/registry/customresource/validator.go#L60) against `IsDNS1123Subdomain()`. And we already forbid `.`. So the only actual change is limiting the length to 63 I think?

Ideally we would only call this from  `ValidateCreate()` though, so it would only affects new BMHs, not existing ones.
`validateHost()` is called from `validateChanges()` as well.

nerok:
Should the length check be moved to ValidateCreate then, so it isn't ran on changes? Or should validateChanges pick that error out and rather log it? Those are the alternatives I could come up with

zaneb:
I think moving it to validateCreate makes sense

nerok:
It got a lot more complicated to ignore or reimplement the validations, so I think this solution was simple enough.

internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation_test.go
```diff
@@ -978,6 +1040,14 @@ func TestValidateUpdate(t *testing.T) {
 				TypeMeta: tm, ObjectMeta: om, Spec: metal3api.BareMetalHostSpec{ExternallyProvisioned: true}},
 			wantedErr: "externallyProvisioned can not be changed",
 		},
+		{
+			name: "tooLongNameShouldBeAllowedOnChange",
+			newBMH: &metal3api.BareMetalHost{
+				TypeMeta: tm, ObjectMeta: om2, Spec: metal3api.BareMetalHostSpec{}},
```
MahnoorAsghar, line 1046: I'm not sure what changed between oldBMH and newBMH, unless this was intended?

MahnoorAsghar: I think the point is just to run validateUpdate with a long name for newObj, don't mind me

MahnoorAsghar:
(LGTM)

PR #2642, opened 2025-09-01
:book: Update info on running e2e locally

sirpaleet:
Update the test/e2e/README.md to ease running e2e script locally.
Add info on GINKGO_FOCUS, GINKGO_SKIP and BMC_PROTOCOL, and how to clean up after running the ci-e2e.sh script.

Updates the README to better inform users on how to run the e2e locally, since at least one person appreciated new info  on these.

### DIFF
--- a/test/e2e/README.md
+++ b/test/e2e/README.md
@@ -35,6 +35,55 @@ export GINKGO_FOCUS="upgrade"
 ./hack/ci-e2e.sh
 ```
 
+`GINKGO_FOCUS` can be set manually to run specific tests. The options for these
+can be found as the first string value (formatting included) of the line with
+`Describe` or `It`. These can also be combined with other proceeding sections to
+match to even more specific test sections. The value `GINKGO_FOCUS` uses is a
+regexp that should match the description of the spec but not match the regexp
+specified in `GINKGO_SKIP`.
+
+Example:
+
+```go
+var _ = Describe("basic", ...
+  It("should control power cycle of BMH though annotations", ...
+...
+  )
+)
+```
+
+Could be used with:
+
+```bash
+export GINKGO_FOCUS="basic should control power"
+```
+
+Additionally, if you wish to run multiple different tests, just maually
+add another `--focus=` with string to the root Makefile's `test-e2e`
+target.
+
+Skipping tests works otherwise similiarly to adding focus, but in the Makefile
+`GINKGO_SKIP` is split separated by space. Thus, you can either use
+test-specific words with it or you can add another `--skip=` with a longer
+string to the `test-e2e` target.
+
+`BMC_PROTOCOL` can also be set manually. By default the [ci-e2e.sh](https://github.com/metal3-io/baremetal-operator/blob/main/hack/ci-e2e.sh)
+script runs it as `redfish`, but it can also be set to `redfish-virtualmedia`,
+`redfish`, or `ipmi`. Ipmi uses `vbmc` as the BMO e2e emulator, whereas the
+others use `sushy-tools`.
+
+After the tests are run, please ensure proper cleanup before running them again.
+The due process for ensuring all is clean for the next run is (in the
+root directory):
+
+```bash
+./hack/clean-e2e.sh
+make clean
+sudo rm -rf ./test/e2e/images
+```
+
+In addition, make sure related docker containers are removed as well.
+
 It is also possible to run the tests with the fixture provider instead of
 Ironic. Without any changes, the whole suite (including optional tests) will be
 run. Please note, however, that it is quite questionable to call this


sirpaleet:
I also do these steps for cleanup, should something on that be added as well? 
```
./hack/clean-e2e.sh
make clean
docker rm -f $(docker ps -aq)
sudo rm -rf ./test/e2e/images
```
Anything else?

tuminoid:
> I also do these steps for cleanup, should something on that be added as well?
> 
> ```
> docker rm -f $(docker ps -aq)
> ```
> 
> Anything else?

Please don't go nuking all running Docker containers on a machine. Someone might run this on a machine with other containers running, and this will promptly just kill them all.

test/e2e/README.md
```diff
@@ -35,6 +35,22 @@ export GINKGO_FOCUS="upgrade"
 ./hack/ci-e2e.sh
 ```
 
+GINKGO_FOCUS can be set manually to run specific tests. The options for these can be found as the first string value of the line with "Describe" or "It". These can also be combined with other proceeding sections to match to even more specific test sections. The value GINKGO_FOCUS uses is a regexp that should match the description of the spec but not match the regexp specified in GINKGO_SKIP. Example:
```
tuminoid:
```suggestion
`GINKGO_FOCUS` can be set manually to run specific tests. The options for these can be found as the first string value of the line with `Describe` or `It`. These can also be combined with other proceeding sections to match to even more specific test sections. The value `GINKGO_FOCUS` uses is a regexp that should match the description of the spec but not match the regexp specified in `GINKGO_SKIP`. Example:
```

Use backticks appropriately where applicable.

Also, limit the line length to 80, the checker will fail otherwise when you come out of draft.

test/e2e/README.md
```diff
@@ -35,6 +35,22 @@ export GINKGO_FOCUS="upgrade"
...
+
+```bash
```
tuminoid:
```suggestion
```go
```

test/e2e/README.md
```diff
@@ -35,6 +35,22 @@ export GINKGO_FOCUS="upgrade"
...
 
+GINKGO_FOCUS can be set manually to run specific tests. The options for these can be found as the first string value of the line with "Describe" or "It". These can also be combined with other proceeding sections to match to even more specific test sections. The value GINKGO_FOCUS uses is a regexp that should match the description of the spec but not match the regexp specified in GINKGO_SKIP. Example:
+
+```bash 
+var _ = Describe("basic", ...
+  It("should control power cycle of BMH though annotations", ...
+...
+  )
+)
+// Which could be used for example with:
```
tuminoid: This should be a bash block
```suggestion
# Which could be used for example with:
```

test/e2e/README.md
```diff
@@ -35,6 +35,22 @@ export GINKGO_FOCUS="upgrade"
...
+export GINKGO_FOCUS="basic should control power"
+```
+
+If you wish to run tests from different focuses, just maually add another "--focus=" to the root Makefile's "test-e2e". The same logic works for setting GINKGO_SKIP to skip specific test cases.
+
+BMC_PROTOCOL can also be set manually. By default the ci-e2e.sh script runs it as "redfish", but it can also be set to "redfish-virtualmedia", "redfish", or "ipmi". Ipmi uses "vbmc" as the BMO e2e emulator, whereas the others use "sushy-tools".
```
tuminoid: Backticks and line length.

adilGhaffarDev:
Should we add Makefile targets to prepare the environment for running e2e tests locally and for cleanup? Right now, users have to manually install dependencies before running make test-e2e.

sirpaleet:
> Should we add Makefile targets to prepare the environment for running e2e tests locally and for cleanup? Right now, users have to manually install dependencies before running make test-e2e.

I think that could be good! The makefile is easier to understand how to use for a newcomer :) Ideally someone would need to test what needs to be installed / changed before running the script on a clean machine (i did mine where some things were already edited and so cannot fully remember what I changed or what even affected anything).

test/e2e/README.md
sirpaleet: If I set this code to go this is incorrect syntax, so I'll keep "//" ?

tuminoid: The export part where this comment refers to is not Go, but Bash. This block needs to be split in two, one for Go part and one for Bash part.

sirpaleet: I see, Makes sense

test/e2e/README.md
```diff
@@ -35,6 +35,50 @@ export GINKGO_FOCUS="upgrade"
...
+Could be used with:
+
+```bash
+export GINKGO_FOCUS="basic should control power"
+```
+
+If you wish to run tests from different focuses, just maually add another
+"--focus=" to the root Makefile's "test-e2e" target. The same logic works for
+setting `GINKGO_SKIP` to skip specific test cases.
+
```
dtantsur, line 69: Could you split the text into sections? It's a bit hard to follow it now.

test/e2e/README.md
```diff
@@ -35,6 +35,51 @@ export GINKGO_FOCUS="upgrade"
...
+Could be used with:
+
+```bash
+export GINKGO_FOCUS="basic should control power"
+```
+
+The same logic works for setting `GINKGO_SKIP`.
+
+Additionally, if you wish to run or skip multiple different tests, just maually
+add another "--focus=" or "--skip=" to the root Makefile's "test-e2e" target.
```
sirpaleet: Now that I am looking at this makefile again, there seems to be a way to parse the GINKGO_SKIP flag for multiple tests actually... I'll add a comment on that as well (probably monday morning as I need to test it)

sirpaleet:
Should be ready now!

test/e2e/README.md
```diff
@@ -35,6 +35,55 @@ export GINKGO_FOCUS="upgrade"
...
+```
+
+Could be used with:
+
+```bash
+export GINKGO_FOCUS="basic should control power"
+```
+
+Additionally, if you wish to run multiple different tests, just maually
+add another `--focus=` with string to the root Makefile's "test-e2e"
```
tuminoid:
```suggestion
add another `--focus=` with string to the root Makefile's `test-e2e`
```

test/e2e/README.md
```diff
@@ -35,6 +35,55 @@ export GINKGO_FOCUS="upgrade"
...
+target.
+
+Skipping tests works otherwise similiarly to adding focus, but in the Makefile
+`GINKGO_SKIP` is split separated by space. Thus, you can either use
+test-specific words with it or you can add another `--skip=` with a longer
+string to the "test-e2e" target.
```
tuminoid:
```suggestion
string to the `test-e2e` target.
```

test/e2e/README.md
```diff
@@ -35,6 +35,55 @@ export GINKGO_FOCUS="upgrade"
...
+
+`BMC_PROTOCOL` can also be set manually. By default the ci-e2e.sh script runs it
```
tuminoid:
```suggestion
`BMC_PROTOCOL` can also be set manually. By default the [ci-e2e.sh](https://github.com/metal3-io/baremetal-operator/blob/main/hack/ci-e2e.sh) script runs it
```

test/e2e/README.md
```diff
@@ -35,6 +35,55 @@ export GINKGO_FOCUS="upgrade"
...
+as "redfish", but it can also be set to "redfish-virtualmedia", "redfish", or
+"ipmi". Ipmi uses "vbmc" as the BMO e2e emulator, whereas the others use
+"sushy-tools".
```
tuminoid:
```suggestion
as `redfish`, but it can also be set to `redfish-virtualmedia`, `redfish`, or
`ipmi`. Ipmi uses `vbmc` as the BMO e2e emulator, whereas the others use
`sushy-tools`.
```

sirpaleet: Since these are literal string values used are they also formatted like this?

sirpaleet: Ah, it is used that way above as well. I see, I'll use this then.

lentzi90 (commented):
/approve

sirpaleet:
> Should we add Makefile targets to prepare the environment for running e2e tests locally and for cleanup? Right now, users have to manually install dependencies before running make test-e2e.

Hi, I was thinking on starting on this in the meantime while working on the other tasks. It might be relatively simple as I have some notes on the installatioin already! I'll at least push some draft before I leave work :) Something like `make e2e-script-setup`, `make e2e-script-run` and  `make e2e-script-clean`? Or some better naming for this?

sirpaleet:
There is already a clean-e2e target in the Makefile too but that just runs the script. I could add some steps I mentioned in the README to this as well if the preferred method would be to just run the scripts via the targets.

lentzi90:
Go ahead! :slightly_smiling_face: 
I'm not sure what exactly you want to include in the make targets, but I would suggest something like `make e2e-prerequisites` for non-CI prep. Then there is already `make test-e2e` for running the tests, and as you mentioned `make clean-e2e`.

The thing is, everyone want different things and have different environments so it quickly becomes very hard to do one thing that works for everyone. Maybe the best thing to do is to add more parameters to the existing targets? For example. if the user wants to remove the images they could do `CLEAN_IMAGES=true make clean-e2e`.

sirpaleet:
> For example. if the user wants to remove the images they could do `CLEAN_IMAGES=true make clean-e2e`.

Hmm, that seems convenient too! Then I could just expand on the existing targets `make test-e2e` and `make clean-e2e` with these variables, plus adding the `make e2e-prerequisites`? At least that'd be clearer to use than the new targets then!

sirpaleet:
Would using snap for installations be okay or not, since the store is Linux-specific? It's just that go is either specific version or latest through snap and updating this installation seems tedious.. EDIT: or using PPA?

s3rj1k:
> Would using snap for installations be okay or not, since the store is Linux-specific? It's just that go is either specific version or latest through snap and updating this installation seems tedious.. EDIT: or using PPA?

Oh, Ubuntu will be very much happy with this

sirpaleet:
> > Would using snap for installations be okay or not, since the store is Linux-specific? It's just that go is either specific version or latest through snap and updating this installation seems tedious.. EDIT: or using PPA?
> 
> Oh, Ubuntu will be very much happy with this

:D Well that's what I was looking at but then... not? Is manual to a specific version better?

lentzi90:
> Would using snap for installations be okay or not, since the store is Linux-specific? It's just that go is either specific version or latest through snap and updating this installation seems tedious.. EDIT: or using PPA?

There is no way you will be able to implement something that works for everyone. If people run this on their own laptops they will want control over these things. The only thing you can do is provide an option that works in some cases and make it optional so that other people can bypass it and provide the necessary parts manually.

sirpaleet:
Hmm okay I'll make it working on my Ubuntu 24.04 first then and let it gather some critique :dancing_women:

sirpaleet:
Did the draft now!

PR #2649, opened 2025-09-03
:sparkle: Fix release note generator for pre-releases

kashifest:
Release note generator should add warning for pre-releases after the title otherwise markdown lint would error out.


tuminoid (commented):
/lgtm

hack/tools/release/notes.go
```diff
@@ -265,6 +256,15 @@ func run() int {
...
+			}
+			if isBeta(latestTag) {
+				fmt.Printf(warningTemplate, "BETA RELEASE")
+			}
+			if isRC(latestTag) {
+				fmt.Printf(warningTemplate, "RELEASE CANDIDATE")
+			}
 			if isAlpha(latestTag) || isBeta(latestTag) || isRC(latestTag) {
 				fmt.Printf("<details>\n")
 				fmt.Printf("<summary>More details about the release</summary>\n\n")
```
Rozzii, line 270:
```suggestion
```
I assume this is not needed anymore then.

kashifest: I think it is still needed, since that hides the detail for the pre-releases

Rozzii: Or not in this form but as a printf of a constant string in the already checked subconditions

kashifest: Yeah but then you repeat the same 3 times, IMO its just fine as is.

kashifest: Perhaps we can improve the coding style but then lets do it separately since we have similar code convention in other repos as well. Currently I am upto just fixing it and get go with rc release

Rozzii: Okay, in lue of the time constraints here I am okay with it. Please make a followup PR when you have time.

PR #2670, opened 2025-09-12
✨ Introduce InspectionMode field

dtantsur:
This change introduces the InspectionMode field from the BMH v1beta1 design proposal. It partly replaces the inspection annotation (namely, its "disabled" value) by something more user-friendly and consistent with AutomatedCleaningMode.

The "disabled" value of the inspection annotation is now deprecated. Other values are not affected by this change.

Assisted-By: Claude Code

### DIFF
--- a/apis/metal3.io/v1alpha1/baremetalhost_types.go
+++ b/apis/metal3.io/v1alpha1/baremetalhost_types.go
@@ -67,7 +67,20 @@ const (
 	// InspectAnnotationValueDisabled is a constant string="disabled"
 	// This is particularly useful to check if inspect annotation is disabled
 	// inspect.metal3.io=disabled.
+	//
+	// Deprecated: use InspectionMode instead.
 	InspectAnnotationValueDisabled = "disabled"
+)
+
+// InspectionMode represents the mode of host inspection.
+type InspectionMode string
+
+const (
+	// InspectionModeDisabled disables host inspection.
+	InspectionModeDisabled InspectionMode = "disabled"
+
+	// InspectionModeAgent runs standard agent-based inspection.
+	InspectionModeAgent InspectionMode = "agent"
 )
 
 // RootDeviceHints holds the hints for specifying the storage location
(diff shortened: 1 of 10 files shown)


apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -67,9 +67,22 @@ const (
...
 
+// InspectionMode represents the mode of host inspection.
+type InspectionMode string
+
+const (
+	// InspectionModeDisabled disables host inspection.
+	InspectionModeDisabled InspectionMode = "disabled"
+
+	// InspectionModeAgent runs standard agent-based inspection.
+	InspectionModeAgent InspectionMode = "agent"
```
MahnoorAsghar, line 83: Nit (and very debatable): due to the various meanings for agent, could we use 'ipa' instead?

apis/metal3.io/v1alpha1/baremetalhost_types_test.go
```diff
@@ -617,51 +617,56 @@ func TestSetHardwareProfile(t *testing.T) {
 }
 
 func TestHasBMCDetails(t *testing.T) {
+}
```
MahnoorAsghar, line 667: I'm wondering why you removed the TestBMCDetails function, perhaps it is no longer needed?

apis/metal3.io/v1alpha1/baremetalhost_types.go
dtantsur: I see where you're coming from, but the word "agent" is not ambiguous upstream

apis/metal3.io/v1alpha1/baremetalhost_types_test.go
dtantsur: I need to take a look, it could be a merge problem or claude getting creative.

dtantsur:
@MahnoorAsghar good catch, updated

lentzi90 (commented):
Nice!
I was about to say that maybe we should switch the e2e tests to use the new field, but this way (with the new e2e test I mean) we test both the annotation and field, so that is good. We can then switch the rest of the tests to use the new field later.
/approve

PR #2674, opened 2025-09-15
:seedling: Add e2e tests for release 0.11

Rozzii:
This commit:
 - Mostly adds overlays into /test/e2e, /config/overlays and in /ironic-deployment/overlays
 - Remove release-0.8 overlays


kashifest (commented):
/lgtm

hack/ci-e2e.sh
```diff
@@ -159,13 +159,16 @@ BMO_OVERLAYS=(
   "${REPO_ROOT}/config/overlays/e2e-release-0.8"
```
lentzi90: Should we remove release-0.8 here? I think it is no longer used after this merges.

lentzi90 (commented):
This all looks good to me. I just wonder if we should remove some of the older configurations at the same time? Release 0.8 doesn't seem to be used anymore after this merges, so maybe we can remove those manifests also?

hack/ci-e2e.sh
Rozzii: Yeah, you are right I will remove it.

Rozzii: Done

kashifest:
Does this need backporting?

Rozzii:
> We could probably also remove this whole folder: https://github.com/metal3-io/baremetal-operator/tree/main/config/overlays/e2e-release-0.8 Can also be done separately of course. /approve

I think it supposed to be removed now, jut the diff is showing it as a rename.

tuminoid:
> Does this need backporting?

AFAIK the whole thing runs on schedule from main, no backporting. @lentzi90 to confirm.

lentzi90:
No backport needed. The release branch overlays are used for upgrade tests and they all run from main.
However, we forgot to update the periodic workflows: https://github.com/metal3-io/baremetal-operator/blob/main/.github/workflows/e2e-test-periodic-release-0.10.yml
We should have one for each maintained release branch.

kashifest:
> No backport needed. The release branch overlays are used for upgrade tests and they all run from main. However, we forgot to update the periodic workflows: https://github.com/metal3-io/baremetal-operator/blob/main/.github/workflows/e2e-test-periodic-release-0.10.yml We should have one for each maintained release branch.

@Rozzii PTAL the periodic workflow as  @lentzi90 mentioned.

PR #2675, opened 2025-09-15
:seedling: Pin ironic release-31.0

Rozzii:
This commit:
 - Sets the default Ironic tag to release-31.0 in the base ironic-deployment overlay


tuminoid (commented):
/lgtm

PR #2677, opened 2025-09-15
:seedling: Fix OSV-Scanner

erjavaskivuori:
**What this PR does / why we need it**: Adopt reusable workflow from project-infra to run OSV scanner.

Renamed the workflow with `scheduled` prefix as it is a plan to add a workflow for all PRs too.


.github/workflows/osv-scanner-scheduled.yml
```diff
@@ -0,0 +1,73 @@
...
+
+permissions:
+  actions: read
+  contents: read
+  security-events: write # for uploading SARIF files
+
+jobs:
+  set-go-version:
+    name: Set up Go version
+    runs-on: ubuntu-latest
```
lentzi90: Let's skip setting up go if we are not going to run the scan anyway 😉 
```suggestion
    runs-on: ubuntu-latest
    if: ${{ github.repository == 'metal3-io/baremetal-operator' }}
```

erjavaskivuori: Oopsie, yes makes sense, I didn't think this thoroughly :D

.github/workflows/osv-scanner-scheduled.yml
```diff
@@ -0,0 +1,74 @@
...
+  scan-scheduled:
+    name: Run OSV Scanner
+    needs: set-go-version
+    if: ${{ github.repository == 'metal3-io/baremetal-operator' }}
+    uses: "google/osv-scanner-action/.github/workflows/osv-scanner-reusable.yml@v2.2.2"
+    with:
+      download-artifact: calculated-go-version
+      scan-args: |-
+        --config=./config.toml
+        --fail-on-vuln=false
```
lentzi90: Looks like there is some issue with this. Check the action logs https://github.com/metal3-io/baremetal-operator/actions/runs/17733329980/job/50389129066?pr=2677#step:5:12

.github/workflows/osv-scanner-scheduled.yml
```diff
@@ -0,0 +1,74 @@
...
+      uses: actions/upload-artifact@v4
+      with:
+        name: calculated-go-version
+        path: ./config.toml
+
+  scan-scheduled:
+    name: Run OSV Scanner
+    needs: set-go-version
+    if: ${{ github.repository == 'metal3-io/baremetal-operator' }}
+    uses: "google/osv-scanner-action/.github/workflows/osv-scanner-reusable.yml@v2.2.2"
```
lentzi90: We should use the hash here also instead of the v2.2.2 tag.

.github/workflows/osv-scanner-scheduled.yml
erjavaskivuori: Ah thanks for this, I think I have used the wrong way to pass the arg.

erjavaskivuori:
/hold

I need to fix the way config.toml is passed for the scanner. I will continue with this on Thursday.

.github/workflows/osv-scanner-scheduled.yml
```diff
@@ -0,0 +1,76 @@
...
+        name: calculated-go-version
+        path: ./config.toml
+
+  scan-scheduled:
+    name: Run OSV Scanner
+    needs: set-go-version
+    if: ${{ github.repository == 'metal3-io/baremetal-operator' }}
+    uses: "google/osv-scanner-action/.github/workflows/osv-scanner-reusable.yml@90b209d0ea55cea1da9fc0c4e65782cc6acb6e2e" # v2.2.2
+    with:
+      download-artifact: calculated-go-version
```
lentzi90: Do we need a separate job for this? It seems inefficient to me to upload the config file and then download in the next job. Can we instead just continue with the scan after creating the `config.toml`?

.github/workflows/osv-scanner-scheduled.yml
```diff
@@ -0,0 +1,76 @@
...
+        --recursive
+        ./
+
+  check-and-notify:
+    name: Check vulnerabilities and notify
+    needs: scan-scheduled
+    runs-on: ubuntu-latest
+    if: ${{ always() && github.repository == 'metal3-io/baremetal-operator' }}
+    steps:
+      - uses: actions/download-artifact@634f93cb2916e3fdff6788551b99b062d0335ce0 # v5.0.0
```
lentzi90: Here also I would prefer to skip downloading artifacts between different jobs. If possible, just keep it in one job so we don't have to do it.

lentzi90 (commented):
Looks like it is working! I am not so fond of uploading and downloading artifacts to pass between jobs though, see details below

.github/workflows/osv-scanner-scheduled.yml
erjavaskivuori: This is because the osv-scanner-reusable.yml has job that sets `runs-on: ubuntu-latest` (https://github.com/google/osv-scanner-action/blob/main/.github/workflows/osv-scanner-reusable.yml#L66) and it is not possible to have that multiple times in one job. So at least I didn't figure out any other way than implementing separate jobs so that I can do stuff before and after the scan.

erjavaskivuori: Oh and also if I try to add running the workflow in a step it says:
```
Unable to resolve action 
`google/osv-scanner-action/.github/workflows/osv-scanner-reusable.yml@90b209d0ea55cea1da9fc0c4e65782cc6acb6e2e`4,
repository or version not found
Selects an action to run as part of a step in your job. An action is a reusable unit of code.
You can use an action defined in the same repository as the workflow, a public repository,
or in a published Docker container image.
```

Because it isn't an action but complete workflow.

lentzi90: Oh I see :slightly_frowning_face:
I don't know any other way either then :shrug:

erjavaskivuori: Yea this is a bit tricky because it isn't recommended to use the bare actions and build own workflow from there.

erjavaskivuori:
Sorry at least one more change coming in. I noticed now that the Setup Go step isn't needed because it sets Go environment and Go in the PATH which isn't necessary in the current setup with separated jobs..

erjavaskivuori:
Sorry at least one more change coming in. I noticed now that the Setup Go step isn't needed because it sets Go environment and Go in the PATH which isn't necessary in the current setup with separated jobs..

tuminoid:
/hold
Still not working, let's keep all of these on hold, and iterate on one to get the right combination of flags. We'll do that iteration in IPAM PR.

tuminoid (commented):
/lgtm

Once the project-infra PR merges, we can test this. Until then, keeping the hold.

PR #2680, opened 2025-09-15
🌱 Add IPA_BASEURI to download ipa image from Nordix

smoshiur1237:
Tests are having issue when downloading the IPA image from upstream and hits timeout. So here we are passing IPA_BASEURI variable to ironic vars to download the IPA image from Nordix cache.


lentzi90:
Thanks @smoshiur1237 ! Let's check if we can run the optional tests to verify also.
/test ?

tuminoid (commented):
/lgtm
/hold
so the test can finish.

lentzi90:
We hit the timeout for IPMI. Redfish-virtualmedia was 7 minutes from timeout also. No errors though so I think we should take this in. Well spotted @smoshiur1237! @smoshiur1237  will update the timeout in project infra also
/hold cancel

PR #2681, opened 2025-09-16
:seedling: Replace ironic deployment with IrSO in e2e

nuhakala:
This PR replaces old Ironic deployment with Ironic standalone operator based installation. It only changes the basic tests, and not the upgrade tests. Upgrade tests can be changed in another PR. However, this PR already introduces overlays for the upgrade tests, but they are not used.

Fixes #


test/e2e/data/ironic-standalone-operator/ironic/base/ironic.yaml
```diff
@@ -0,0 +1,28 @@
+---
+apiVersion: ironic.metal3.io/v1alpha1
+kind: Ironic
+metadata:
+  name: ironic
+  namespace: "${IRONIC_NAMESPACE}"
```
adilGhaffarDev:
```suggestion
  namespace: baremetal-operator-system
```

test/e2e/data/ironic-standalone-operator/operator/kustomization.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: ${IRONIC_NAMESPACE}
```
adilGhaffarDev:
```suggestion
namespace: baremetal-operator-system
```

test/e2e/data/ironic-standalone-operator/operator/kustomization.yaml
```diff
@@ -0,0 +1,33 @@
...
+resources:
+  - https://github.com/metal3-io/ironic-standalone-operator/releases/download/v0.5.1/install.yaml
+
+generatorOptions:
+  disableNameSuffixHash: true
+
+configMapGenerator:
+  - name: ironic-operator-config
+    literals:
+      - FEATURE_GATES=HighAvailability=true
```
dtantsur: The HA gate is enabled by default nowadays, and you don't use the HA feature anyway.

dtantsur:
Shaping nicely so far, good job!

test/e2e/data/ironic-standalone-operator/operator/kustomization.yaml
nuhakala: Oh, well then I remove it.

test/e2e/config/ironic.yaml
```diff
@@ -49,6 +49,12 @@ variables:
   IRONIC_PASSWORD: "changeme"
   IRONIC_PROVISIONING_IP: "localhost"
   IRONIC_PROVISIONING_PORT: "6385"
+  IRSO_OPERATOR_LATEST: "data/ironic-standalone-operator/operator"
+  IRSO_IRONIC_26.0: "data/ironic-standalone-operator/ironic/overlays/e2e-release-26.0"
+  IRSO_IRONIC_27.0: "data/ironic-standalone-operator/ironic/overlays/e2e-release-27.0"
+  IRSO_IRONIC_28.0: "data/ironic-standalone-operator/ironic/overlays/e2e-release-28.0"
+  IRSO_IRONIC_29.0: "data/ironic-standalone-operator/ironic/overlays/e2e-release-29.0"
```
dtantsur: All these versions are out of support already or will be soon. I recommend you at least add 30 and 31. Testing 26 right now is especially questionable.

test/e2e/data/ironic-standalone-operator/ironic/base/ironic.yaml
```diff
@@ -0,0 +1,25 @@
+---
+apiVersion: ironic.metal3.io/v1alpha1
+kind: Ironic
+metadata:
+  name: ironic
+  namespace: ironic-standalone-operator-system
+spec:
+  images:
+    deployRamdiskBranch: "master"
```
dtantsur: This is the default, and we don't support any other branches for real now.

test/e2e/data/ironic-standalone-operator/ironic/base/ironic.yaml
```diff
@@ -0,0 +1,25 @@
...
+  networking:
+    dhcp:
+      rangeBegin: "192.168.222.100"
+      rangeEnd: "192.168.222.200"
+      networkCIDR: "192.168.222.0/24"
+    interface: "eth0"
+    ipAddress: "192.168.222.1"
+    ipAddressManager: keepalived
+  deployRamdisk:
+    sshKey: "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIBJk9vkbfFELINnsvubSMkk+8uPWYkZRx39Mp+KR55/w"
```
dtantsur: I'm curious why this is a constant. Are you planning on including the corresponding private key in the source tree?

test/e2e/data/ironic-standalone-operator/ironic/overlays/e2e/kustomization.yaml
```diff
@@ -0,0 +1,24 @@
...
+  - ../../../components/tls
+
+patches:
+- target:
+    kind: Ironic
+    name: ironic
+  patch: |-
+    - op: replace
+      path: /spec/images/ironic
+      value: "quay.io/metal3-io/ironic:main"
```
dtantsur: This is the default, why replace it?

test/e2e/config/ironic.yaml
nuhakala: I have not yet made my testing into upgrade tests, so I have mostly just ignored these. I added them in the beginning more like a placeholder, but added 30 and 31 now.

Also, I checked and 27 is still supported according to the documentation, so should it be dropped yet?

test/e2e/data/ironic-standalone-operator/ironic/base/ironic.yaml
nuhakala: Removed

test/e2e/data/ironic-standalone-operator/ironic/base/ironic.yaml
nuhakala: A mistake, there should be placeholder that gets replaced, but accidentally committed the key

test/e2e/data/ironic-standalone-operator/ironic/overlays/e2e/kustomization.yaml
nuhakala: Removed

test/e2e/data/ironic-standalone-operator/components/basic-auth/kustomization.yaml
```diff
@@ -0,0 +1,17 @@
...
+generatorOptions:
+  disableNameSuffixHash: true
+
+# NOTE: These credentials are generated automatically in hack/ci-e2e.sh
+secretGenerator:
+- name: ironic-auth
+  behavior: create
+  literals:
+  - username=bff18d63-ee0a-4947-bc03-6bdce43b4d62
+  - password=e7016bc0-cbca-4bc7-bf19-deb3883494ab
```
dtantsur: Maybe use a more obvious placeholder so that we don't start relying on these values?

test/e2e/data/ironic-standalone-operator/components/tls/certificate.yaml
```diff
@@ -0,0 +1,43 @@
...
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: ironic-cert
+  namespace: baremetal-operator-system
+spec:
+  commonName: ironic-cert
+  ipAddresses:
+  - 192.168.222.1
```
dtantsur: Should this be a placeholder too? I guess minikube is relatively predictable..

test/e2e/data/ironic-standalone-operator/ironic/base/ironic.yaml
```diff
@@ -0,0 +1,24 @@
...
+  networking:
+    dhcp:
+      rangeBegin: "192.168.222.100"
+      rangeEnd: "192.168.222.200"
+      networkCIDR: "192.168.222.0/24"
+    interface: "eth0"
+    ipAddress: "192.168.222.1"
+    ipAddressManager: keepalived
+  deployRamdisk:
+    sshKey: "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIOxoEJC7b9PllhdVt9tk9TJqSqNkbCyi3NGOqsN+03Rt"
```
dtantsur: Still hardcoded?

test/e2e/data/ironic-standalone-operator/operator/kustomization.yaml
```diff
@@ -0,0 +1,55 @@
...
+      spec:
+        template:
+          spec:
+            containers:
+              - name: manager
+                envFrom:
+                  - configMapRef:
+                      name: ironic-operator-config
+
+  # fix namespace in dnsNames
```
dtantsur: Is this something we need to fix upstream in IrSO?

test/e2e/data/ironic-standalone-operator/operator/kustomization.yaml
```diff
@@ -0,0 +1,55 @@
...
+      kind: Certificate
+      name: ironic-standalone-operator-serving-cert
+    patch: |-
+      - op: replace
+        path: /spec/dnsNames
+        value:
+          - ironic-standalone-operator-webhook-service.baremetal-operator-system.svc
+          - ironic-standalone-operator-webhook-service.baremetal-operator-system.svc.cluster.local
+
+  # Fix namespace in inject-ca-from annotation
```
dtantsur: Same question

test/e2e/common.go
```diff
@@ -667,3 +668,39 @@ func dumpIronicNodes(ctx context.Context, e2eConfig *Config, artifactFolder stri
...
+		err := input.Client.Get(ctx, client.ObjectKey{
+			Namespace: input.Namespace,
+			Name:      input.Name,
+		}, ironic)
+		GinkgoWriter.Printf("Ironic parsed with error: %v", err)
+		g.Expect(err).ToNot(HaveOccurred())
+
+		ready := false
+		for _, cond := range ironic.Status.Conditions {
+			if cond.Type == string(irsov1alpha1.IronicStatusReady) && cond.Status == metav1.ConditionTrue {
```
dtantsur: To be really-really sure, you can additionally check Status.InstalledVersion

test/e2e/data/ironic-standalone-operator/components/basic-auth/kustomization.yaml
nuhakala: There should now be better placeholders.

test/e2e/data/ironic-standalone-operator/ironic/base/ironic.yaml
nuhakala: Yes, it was overridden again and I staged changes in a batch and did not check it :( fixed again.

test/e2e/data/ironic-standalone-operator/operator/kustomization.yaml
nuhakala: I don't think so. The issue here is that we put irso under different namespace than is used in the install manifest. Kustomization namespace rule overrides the namespaces of the created objects, but it does not update the values inside the object such as these DNS names that depend on the namespace. I wrote this to the comment to avoid confusion in future.

test/e2e/data/ironic-standalone-operator/operator/kustomization.yaml
nuhakala: See above

test/e2e/config/bmcs-redfish-virtualmedia.yaml
```diff
@@ -1,14 +1,14 @@
 - user: admin
   password: password
-  address: "redfish-virtualmedia+http://192.168.222.1:8000/redfish/v1/Systems/bmo-e2e-0"
+  address: "redfish-virtualmedia+http://172.18.0.1:8000/redfish/v1/Systems/bmo-e2e-0"
```
nuhakala: `172.18.0.1` is the address that is automatically created by docker/kind when creating a kind cluster. It seems to be quite stable so maybe using this is okay? But maybe we should use something else nevertheless.

test/e2e/sushy-tools/sushy-emulator.conf
```diff
@@ -1,5 +1,5 @@
 # Listen on the local IP address 192.168.222.1
-SUSHY_EMULATOR_LISTEN_IP = u'192.168.222.1'
+SUSHY_EMULATOR_LISTEN_IP = u'0.0.0.0'
```
nuhakala: Ironic is not in the same network anymore, so sushy basically needs to listen localhost. Simplest to listen everything.

nuhakala (commented):
The basic tests passed with these changes. Putting ironic inside the kind node introduced a problem with IP addresses: sushy-tools container were listening the same IP-address as ironic, and it worked because previously both were in host network and using the same interface/IP. When using irso, ironic needs to call sushy-tools from inside the kind node. This caused problem because the same IP address was assigned both on host interface and inside the kind node. Hence the changes in IP addresses.

We might want to use something else than the automatically assigned docker/kind network IPs, but at least it is working now with them (see comments below). Any opinions how to proceed with these?

test/e2e/config/ironic.yaml
nuhakala: I decided to leave the upgrade tests out of this PR as they were not passing so easily and this PR was getting rather large, so I left these commented.

test/e2e/data/ironic-standalone-operator/components/tls/certificate.yaml
nuhakala: To have some benefit from that, it would require to replace the IPs with variables in the ironic configuration as well and in other places. I would say it is simpler not to do that, but can change this if it is wanted.

hack/ci-e2e.sh
```diff
@@ -59,8 +59,10 @@ sudo apt-get install -y libvirt-dev pkg-config
 # Build the container image with e2e tag (used in tests)
 IMG=quay.io/metal3-io/baremetal-operator:e2e make docker
 
-virsh -c qemu:///system net-define "${REPO_ROOT}/hack/e2e/net.xml"
-virsh -c qemu:///system net-start baremetal-e2e
+if ! sudo virsh net-list --all | grep baremetal-e2e; then
+    virsh -c qemu:///system net-define "${REPO_ROOT}/hack/e2e/net.xml"
+    virsh -c qemu:///system net-start baremetal-e2e
+fi
```
nuhakala, line 65: I added multiple if clauses in this file so that running the test locally is not such a pain. I can revert if you don't want these in.

There still is no cleanup for the iptable rules, VMs, virsh net, or docker containers. With these if clauses there is not really need to clean those up, other than the iptables rules as every iteration will add the same two rules again.

hack/ci-e2e.sh
```diff
@@ -75,13 +77,15 @@ IP_ADDRESS="192.168.222.1"
 export IRONIC_PROVISIONING_IP="${IP_ADDRESS}"
 
 # Build vbmctl
-make build-vbmctl
-# Create VMs to act as BMHs in the tests.
-./bin/vbmctl --yaml-source-file "${E2E_BMCS_CONF_FILE}"
+if [[ ! -f ./bin/vbmctl ]]; then
+    make build-vbmctl
+    # Create VMs to act as BMHs in the tests.
+    ./bin/vbmctl --yaml-source-file "${E2E_BMCS_CONF_FILE}"
```
lentzi90: I think this should be outside the if.
Possible improvement (but not required) could be to make the `build-vbmctl` target smarter instead of using this check here to to decide if we should build or not. This is after all something `make` is really good at.

hack/ci-e2e.sh
lentzi90: I think they are quite ok. I agree that this script has been a pain to use locally. What I usually do is to run the script only once and stopping before the `make test-e2e`. After that I just use `make test-e2e` to run the tests and never touch the script.

hack/ci-e2e.sh
```diff
@@ -189,6 +198,10 @@ for overlay in "${IRONIC_OVERLAYS[@]}"; do
   "${overlay}/ironic-auth-config"
 done
 
+sed -i "s|SSH_PUB_KEY_CONTENT|${pub_ssh_key}|" "${REPO_ROOT}"/test/e2e/data/ironic-standalone-operator/ironic/base/ironic.yaml
+sed -i "s/.*username.*/  - username=${IRONIC_USERNAME}/" "${REPO_ROOT}"/test/e2e/data/ironic-standalone-operator/components/basic-auth/kustomization.yaml
+sed -i "s/.*password.*/  - password=${IRONIC_PASSWORD}/" "${REPO_ROOT}"/test/e2e/data/ironic-standalone-operator/components/basic-auth/kustomization.yaml
```
lentzi90: This is not so nice, because it leaves the repo in dirty state.
There isn't much we can do about the public ssh key here, but the username and password would be better to create from files in my opinion. Those file should then be added to .gitignore.

test/e2e/data/ironic-standalone-operator/components/basic-auth/kustomization.yaml
```diff
@@ -0,0 +1,17 @@
...
+  disableNameSuffixHash: true
+
+# NOTE: These credentials are generated automatically in hack/ci-e2e.sh
+secretGenerator:
+- name: ironic-auth
+  behavior: create
+  literals:
+  - username=IRONIC_USERNAME
+  - password=IRONIC_PASSWORD
+  type: Opaque
```
lentzi90, line 17: Could we do it the same way as we do for BMO?
https://github.com/metal3-io/baremetal-operator/blob/afd8f3c76677c40d8552d32f4ade50083e388c93/config/overlays/e2e/kustomization.yaml#L33-L37

I.e. these should point to files that do not exist in git, but are generated by the ci-e2e.sh script. Then there is no risk that anyone will use test credentials by mistake

test/e2e/data/ironic-standalone-operator/ironic/base/ironic.yaml
```diff
@@ -0,0 +1,24 @@
...
+apiVersion: ironic.metal3.io/v1alpha1
+kind: Ironic
+metadata:
+  name: ironic
+  namespace: baremetal-operator-system
+spec:
+  images:
+    deployRamdiskDownloader: "quay.io/metal3-io/ironic-ipa-downloader:main"
+    ironic: "quay.io/metal3-io/ironic:main"
+    keepalived: "quay.io/metal3-io/keepalived"
```
lentzi90, line 11: Do we need keepalived? We have not used it in BMO e2e so far.

lentzi90 (commented):
Shaping up nicely!
I didn't quite have time to go through it all yet, but here are some comments from how far I got today.

hack/ci-e2e.sh
nuhakala: I will remove it. Luckily it does not slow down the test much, just throws some errors in the log when it fails to create the VMs.

test/e2e/data/ironic-standalone-operator/ironic/base/ironic.yaml
nuhakala: I think so. Tested without keepalived and ironic pod does not get up. Keepalived will create the configured ip address for the pod so that other containers like httpd can use it. I don't know how to make it work if we leave keepalived out and let irso detect the ip address from the interface. Maybe it could work but then the ip address could change in theory.

dtantsur: IrSO/ironic-image can detect an existing IP from the interface, but then you get whatever IP ends up. You need to assign it manually in this case.

test/e2e/data/ironic-standalone-operator/components/basic-auth/kustomization.yaml
dtantsur: Also note that, unlike deploy.sh, IrSO always uses authentication. You could, as an option, leave the secret empty, let IrSO generate it, and then read the generated secret from the tests.

test/e2e/config/bmcs-redfish-virtualmedia.yaml
lentzi90: I don't really understand this part. I didn't fully understand it before either, but it is interesting that it needed to be changed. So with 172.18.0.1, Ironic will go to the docker bridge and through that reach sushy-tools/vbmc that are running as containers outside of the kind cluster.
I guess that makes as much sense as going to the virsh network bridge (192.168.222.1).
Just commenting my thoughts in case it is helpful. I don't see an issue with this.

test/e2e/data/ironic-standalone-operator/components/basic-auth/kustomization.yaml
lentzi90: That is nice! Can we try it @nuhakala ?

test/e2e/data/ironic-standalone-operator/ironic/base/ironic.yaml
```diff
@@ -0,0 +1,24 @@
...
+  version: "latest"
+  networking:
+    dhcp:
+      rangeBegin: "192.168.222.100"
+      rangeEnd: "192.168.222.200"
+      networkCIDR: "192.168.222.0/24"
+    interface: "eth0"
+    ipAddress: "192.168.222.1"
```
lentzi90: This feels wrong to me. Maybe this is why the BMC addresses had to be changed?
This would be the IP of the virsh network gateway. I guess you used it because we have it [here](https://github.com/metal3-io/baremetal-operator/blob/6b4dbd9e3b37b10379f32f602bd414fa4f3b3756/ironic-deployment/overlays/e2e/kustomization.yaml#L26)? It is actually not the IP used directly by Ironic, but it is forwarded to Ironic. So we set the IP in the certificate for the clients.
We have kind running with the necessary ports forwarded so that Ironic can run in the "host network" (i.e. the kind node network) and still be reachable from the virsh network.

Could we try with for example 192.168.222.2 instead or an IP in the kind network range perhaps?

test/e2e/data/ironic-standalone-operator/components/basic-auth/kustomization.yaml
nuhakala: I will try that

test/e2e/data/ironic-standalone-operator/operator/kustomization.yaml
```diff
@@ -0,0 +1,57 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: baremetal-operator-system
```
lentzi90: Why use this namespace? I think it makes sense that IrSO has its own namespace and we get less headache from patching below 🙂

test/e2e/data/ironic-standalone-operator/ironic/overlays/e2e-release-27.0/kustomization.yaml
```diff
@@ -0,0 +1,18 @@
...
+
+patches:
+  - target:
+      kind: Ironic
+      name: ironic
+    patch: |-
+      - op: replace
+        path: /spec/images/ironic
+        value: "quay.io/metal3-io/ironic:release-27.0"
+        images:
```
lentzi90: What is the `images:` thing doing?

test/e2e/data/ironic-standalone-operator/operator/kustomization.yaml
nuhakala: I tried first, but that created its own problems. BMO was not able to read the same secrets as ironic, and we would need to duplicate the secrets (certs, auth secrets and such) to bmo namespace and irso namespace. Decided to go with one namespace as it felt simpler.

test/e2e/data/ironic-standalone-operator/ironic/base/ironic.yaml
nuhakala: Now I think of it, why do we need to assign it manually? Will BMO break (that is, not get connection to ironic) if we don't assign it here but let irso decide it? Because BMO is the only service I can think of which needs to connect to ironic.

test/e2e/data/ironic-standalone-operator/ironic/base/ironic.yaml
nuhakala: Yes, I used it because it was already used in the configs. I also tried using 192.168.222.2 as the bmc address, so the other way around: not this ironic address but the bmc addresses. That did not work, not sure why. What I managed to dig out, is that the default routing inside the kind node should go through the docker bridge back to host. However, the routing inside kind node did not work properly with these addresses, maybe it could be solved with more configuration, but I just took the easy route. I suspect that 192.168.222.0/24 subnet was routed back to the kind node itself, and hence the requests with slightly different address did not get out, but like I said I am not sure.

Thinking this more, if the problem was indeed the routing related to the subnet, using different enough IP could solve it, like one from kind network. I will check it out and do some tests.

test/e2e/data/ironic-standalone-operator/ironic/overlays/e2e-release-27.0/kustomization.yaml
nuhakala: I don't know, I shamelessly copied it from this PR https://github.com/metal3-io/cluster-api-provider-metal3/pull/2776

Maybe @adilGhaffarDev can answer that?

test/e2e/data/ironic-standalone-operator/ironic/base/ironic.yaml
dtantsur: Will BMO know the right IP address? IrSO will default to status.hostIP of the pod.

test/e2e/data/ironic-standalone-operator/operator/kustomization.yaml
dtantsur: Well, IrSO reads all namespaces by default, so you can keep IrSO itself in its own namespace and the Ironic object (and thus the Ironic pod) in baremetal-operator-system (like it is now).

test/e2e/data/ironic-standalone-operator/ironic/base/ironic.yaml
nuhakala: Yeah, I don't think it knows, as far as I know it cannot detect it automatically. ATM the ironic endpoint is given to BMO as env variable, so using the autoconfiguration would not work as the IP can change.

test/e2e/data/ironic-standalone-operator/ironic/overlays/e2e-release-27.0/kustomization.yaml
dtantsur: Looks invalid to me too.

test/e2e/data/ironic-standalone-operator/operator/kustomization.yaml
nuhakala: Oh, that is handy. I will try that.

test/e2e/data/ironic-standalone-operator/components/basic-auth/kustomization.yaml
nuhakala: I think this could have been done, however, irso adds a random generated suffix to the secret name and after checking the source this cannot be avoided. So it would require that we read the secret name from ironic pod and then pass it to BMO and this would become too complicated so decided to have the credentials in a file and create the secret manually.

test/e2e/data/ironic-standalone-operator/ironic/base/ironic.yaml
nuhakala: I changed the IP to 192.168.111.1 and now it seems to work. Is this IP okay or should I choose something else?

test/e2e/data/ironic-standalone-operator/operator/kustomization.yaml
nuhakala: This worked marvelously, thank you.

test/e2e/data/ironic-standalone-operator/ironic/overlays/e2e-release-27.0/kustomization.yaml
adilGhaffarDev: yes its wrong, remove `images:`
I will fix it on capm3 side too

test/e2e/common.go
nuhakala: Is this what you meant? No idea where I could compare the installed version information.

test/e2e/sushy-tools/sushy-emulator.conf
nuhakala: Outdated, this change seems not to be needed.

test/e2e/data/ironic-standalone-operator/ironic/overlays/e2e-release-27.0/kustomization.yaml
nuhakala: Removed.

test/e2e/e2e_suite_test.go
```diff
@@ -125,23 +128,40 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 	}
 
 	bmoIronicNamespace := "baremetal-operator-system"
+	irsoNamespace := "ironic-standalone-operator-system"
 
 	if e2eConfig.GetBoolVariable("DEPLOY_IRONIC") {
 		// Install Ironic
-		By("Installing Ironic")
+		By("Install IRSO")
```
lentzi90: nit: make the description so that it fits in with the function name, i.e. so that you can read the whole thing like a sentence: "By installing IRSO".
```suggestion
		By("Installing IRSO")
```

test/e2e/e2e_suite_test.go
```diff
@@ -125,23 +128,40 @@ var _ = SynchronizedBeforeSuite(func() []byte {
...
-				LogPath:             filepath.Join(artifactFolder, "logs", bmoIronicNamespace),
+				DeploymentName:      "ironic-standalone-operator-controller-manager",
+				DeploymentNamespace: irsoNamespace,
+				LogPath:             filepath.Join(artifactFolder, "logs", irsoNamespace),
 				WaitIntervals:       e2eConfig.GetIntervals("default", "wait-deployment"),
 			})
 		})
 		Expect(err).NotTo(HaveOccurred())
+
+		By("Install Ironic CR in the target cluster")
```
lentzi90:
```suggestion
		By("Installing Ironic CR in the target cluster")
```
(I would maybe also skip the "CR" here. The important part is that we install Ironic. That we do it by applying a CR is not so relevant.)

test/e2e/data/ironic-standalone-operator/ironic/base/ironic.yaml
lentzi90: I am lost. Are you sure it works? I see some tests are failing here now but maybe that is from other things?
If it works I think it is ok.

> I suspect that 192.168.222.0/24 subnet was routed back to the kind node itself, and hence the requests with slightly different address did not get out, but like I said I am not sure.

This sounds unlikely to me. That subnet is handled by virsh/libvirt so it would probably be routed to the VMs.

Putting some more context in case it helps clarifying. We have the following requirements:
1. Ironic needs to reach sushy-tools and IPA
2. BMO needs to reach Ironic
3. IPA needs to reach Ironic to post back inspection data

The current (hacky) solution is that Ironic runs in "host network" on the kind node, and the kind node has ports forwarded from the real host. The kind node has an address like 172.18.0.x. We create a virsh network for the VMs in 192.168.222.x. Ironic can luckily reach this. (I think it broke with a docker update recently, and we got it back working with [this](https://github.com/metal3-io/baremetal-operator/blob/cdc7eed61732b089acb46135adbf05329bbac888/hack/ci-e2e.sh#L65-L68).)

Then we still need some way for IPA to reach Ironic. It works by configuring 192.168.222.1 as the IP that [they use for accessing Ironic](https://github.com/metal3-io/baremetal-operator/blob/cdc7eed61732b089acb46135adbf05329bbac888/ironic-deployment/overlays/e2e/ironic_bmo_configmap.env#L5-L7) and making sure that is also in the certificate. So IPA reach the host through this IP and there we have [forwarded the relevant ports](https://github.com/metal3-io/baremetal-operator/blob/cdc7eed61732b089acb46135adbf05329bbac888/test/e2e/config/ironic.yaml#L73-L82) to the kind node where Ironic is listening and getting the traffic.

test/e2e/data/ironic-standalone-operator/components/basic-auth/kustomization.yaml
dtantsur: You can read it from the Ironic resource itself (I'm open to having an annotation that provides a template for the secret name).

nuhakala: Not sure if I understand correctly, but the name would still be dynamic and hence the problem is to inform BMO about the secret. Either we would need to patch the kustomize resources in the go code, or we would need to edit the deployment in go code. Or do you have a better idea that I have not thought about?

test/e2e/data/ironic-standalone-operator/ironic/base/ironic.yaml
nuhakala: Hmm, I messed things up a little with the IPs, but also forwarding between VMs and ironic was not working properly with the routing rules, and hence my previous testing failed.

Anyway, to clear things out

1. Ironic can reach sushy tools and IPA now because docker container default routing goes through host machine. Host machine will then correctly route the packages.
2. BMO can now reach ironic, as both are inside the cluster and the ironic endpoint is correctly configured for BMO.
3. IPA can now reach ironic, after I replaced the routing rules with a veth pair. Previously IPA was not able to reach ironic because it tried to access ironic endpoint, but it was hidden behind docker bridge IP. Connecting these two networks with a veth pair allows IPA to find ironic.

There still might be some problems but provisioning test passed locally now.

dtantsur:
Please collect Ironic logs, otherwise debugging the CI will be very hard.

hack/e2e/net.xml
```diff
@@ -8,8 +8,8 @@
   <bridge name='metal3'/>
   <ip address='192.168.222.1' netmask='255.255.255.0'>
     <dhcp>
-      <range start='192.168.222.2' end='192.168.222.99'/>
-      <bootp file='http://192.168.222.1:6180/boot.ipxe'/>
+      <range start='192.168.222.3' end='192.168.222.99'/>
```
lentzi90: I see you have DHCP configured in the Ironic manifest and I wonder if we should then remove this?

test/e2e/data/ironic-standalone-operator/ironic/base/ironic.yaml
nuhakala: Okay, turns out that we also need the routing rules in addition to the veth pair. With both of those the tests passed.

lentzi90:
Nice!
I wonder if the fixture test is actually broken/flaky on main as well... The upgrade tests are broken at the moment and that is where the fixture fails so it sounds quite suspicious

hack/e2e/net.xml
```diff
@@ -7,9 +7,5 @@
   </forward>
   <bridge name='metal3'/>
   <ip address='192.168.222.1' netmask='255.255.255.0'>
-    <dhcp>
-      <range start='192.168.222.2' end='192.168.222.99'/>
-      <bootp file='http://192.168.222.1:6180/boot.ipxe'/>
```
lentzi90, line 12: Seeing the IPMI fail and redfish-virtualmedia succeed makes me suspect this. I think we still need the `bootp` line to make network boot work.

nuhakala: Yeah, I came to the same conclusion. Adding this back.

hack/e2e/net.xml
lentzi90: Is it so that we have to specify a range? I was hoping we could remove that and keep the `bootp` only.
It doesn't really matter now that we no longer use reserved DHCP addresses for the VMs, but it might be very confusing at some point if we have two DHCP servers giving out different IPs.

nuhakala: At least local tests passed without that line. Lets see what happens in CI.

lentzi90: Ok looks like we still need that. That could indicate that we don't need/use the dnsmasq...
Anyway I think this can be left to clean up later.

lentzi90 (commented):
Thanks @nuhakala !
I think this is a great improvement, especially the veth part that came as a "bonus". We could probably clean up the DHCP config separately.
/approve

hack/e2e/net.xml
nuhakala: Yeah, maybe it can be removed once upgrade tests are also updated. As those still use the old system, it might be needed there.

test/e2e/config/ironic.yaml
```diff
@@ -17,6 +17,10 @@ images:
   loadBehavior: tryLoad
 - name: quay.io/jetstack/cert-manager-controller:v1.17.1
   loadBehavior: tryLoad
+- name: quay.io/metal3-io/ironic-standalone-operator:v0.5.1
```
dtantsur, line 20: Follow-up: 0.6.0 is available now

test/e2e/data/ironic-standalone-operator/components/basic-auth/kustomization.yaml
dtantsur: I think https://github.com/metal3-io/baremetal-operator/pull/2713, when finished, will make the entire problem go away

test/e2e/data/ironic-standalone-operator/ironic/overlays/e2e-release-27.0/kustomization.yaml
```diff
@@ -0,0 +1,17 @@
...
+  - ../../../components/tls
+  - ../../../components/basic-auth
+
+patches:
+  - target:
+      kind: Ironic
+      name: ironic
+    patch: |-
+      - op: replace
+        path: /spec/images/ironic
```
dtantsur, line 16: Late realization, sorry: you should not patch this. Instead, patch in `version: "27.0"` etc

test/e2e/data/ironic-standalone-operator/operator/kustomization.yaml
```diff
@@ -0,0 +1,32 @@
...
+      apiVersion: apps/v1
+      kind: Deployment
+      metadata:
+        name: ironic-standalone-operator
+      spec:
+        template:
+          spec:
+            containers:
+              - name: manager
+                envFrom:
```
dtantsur, line 30: We need to bring this change to IrSO upstream, it's convenient to have a configmap even for production deploys.

dtantsur:
@nuhakala could you please check the comment I left and potentially follow-up? Thanks!

/lgtm

PR #2685, opened 2025-09-16
:book: Add release-0.11 test triggers to README

Rozzii:
After v0.11 was released the ROOT README needs to be updated with the new triggers.


kashifest (commented):
/approve

PR #2696, opened 2025-09-22
🌱Increase optional periodic test time

smoshiur1237:
Optional periodic test is having timeout in 2 hours. We have decided to give it more time to complete the job.


tuminoid (commented):
/lgtm

PR #2697, opened 2025-09-23
🌱 E2E: Change how upgrade tests are cleaned up

lentzi90:

We used DeferCleanup in many places, which is nice in the way that it can clean up exactly the same thing we have applied. However, it becomes complicated to know what will be called and in what order. We have run into situations where BMO is cleaned up before the BMHs are deleted, causing a dead-lock since nothing can remove the finalizers anymore.

This commit moves the clean up to AfterEach instead. That way we have complete control over the order at the cost of not knowing exactly which kustomization was applied.
It also does the patching of the BMH using Eventually now to avoid errors if the webhook is a bit slow to start.

A future improvement (larger task) would be to make the upgrade test declarative instead of generated from the config file, similar to clusterctl upgrade tests for CAPI. That would allow us to clean up the correct kustomizations and make it easier to tell from the code what is being tested.

Fixes #

### DIFF
--- a/test/e2e/upgrade_test.go
+++ b/test/e2e/upgrade_test.go
@@ -214,11 +214,6 @@ func RunUpgradeTest(ctx context.Context, input *BMOIronicUpgradeInput, upgradeCl
 			})
 		})
 		Expect(err).NotTo(HaveOccurred())
-
-		DeferCleanup(func() {
-			By(fmt.Sprintf("Removing Ironic kustomization %s from the upgrade cluster", initIronicKustomization))
-			cleanupBaremetalOperatorSystem(ctx, upgradeClusterProxy, initIronicKustomization)
-		})
 	}
 	if input.DeployBMO {
 		// Install BMO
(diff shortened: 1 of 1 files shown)


lentzi90:
/hold
We should run the optional tests. They are the only ones touched here.
/test ?

lentzi90:
Optional tests are working
/hold cancel

lentzi90:
Going to rerun the optional tests a few times to see if we can catch some issues

lentzi90:
One BMH got stuck -> lots registration errors because of conflicting MAC. This is an existing issue also in periodic jobs, although rare.
But, the cleanup worked nicely in the way that it didn't get stuck and waiting until the whole test time out.

lentzi90:
/hold
Two failures in a row is not nice even though periodics are in similar shape at the moment.
Trying one more time

lentzi90:
Thanks!
The optional tests have passed twice and failed twice here. Not statistically significant, but it is better than the periodics, which has 3 failures in a row at the moment.
/unhold

kashifest (commented):
/approve

PR #2698, opened 2025-09-23
:warning: Change architecture default to match controller

honza:
Instead of hardcoding x86, we query the CPU architecture of the running Go binary.

Assisted-By: claude-4-sonnet

### DIFF
--- a/internal/controller/metal3.io/baremetalhost_controller.go
+++ b/internal/controller/metal3.io/baremetalhost_controller.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"runtime"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


zaneb:
/lgtm
This seems safe to me. To be broken by this you'd need to have:

- A management cluster on non-x86 installing x86 nodes
- Relying on the default hardware profile
- Not specifying the arch in the Spec
- On new BMHs that haven't been inspected yet

/hold for lazy consensus

lentzi90 (commented):
This seems sensible to me as well.
/lgtm

lentzi90:
The test failure is unrelated btw. The live-ISO test is failing also in periodic jobs.

dtantsur:
/retest

Should be fixed now

tuminoid (commented):
/lgtm
/unhold

PR #2699, opened 2025-09-25
:seedling: Add maketarget for ensuring docker is present for testing

sirpaleet:

Some people may find the installation process and running the hack e2e-scripts a bit difficult, so making a target for the script and fixing the installed packages and dependencies should be done to make it easier to run them on a VM, and documentation should be added as well.

There is some previous discussion in here: #2642 

Fixes: #2707

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -142,6 +142,11 @@ fuzz-run: ## Run all fuzz tests sequentially with fuzzing enabled (use FUZZ_TIME
 
 ARTIFACTS ?= ${ROOT_DIR}/test/e2e/_artifacts
 
+.PHONY: verify-e2e-prerequisites
+verify-e2e-prerequisites: ## Check that required tools exist for e2e tests
+	@echo "Ensure the local environment is ready for e2e tests..."
+	VERIFY_ONLY=1 ./hack/e2e/ensure_e2e_prerequisites.sh
+
 .PHONY: test-e2e
 test-e2e: $(GINKGO) ## Run the end-to-end tests
 	$(GINKGO) -v --trace -poll-progress-after=$(GINKGO_POLL_PROGRESS_AFTER) \
--- a/hack/ci-e2e.sh
+++ b/hack/ci-e2e.sh
@@ -35,6 +35,8 @@ echo "BMO_E2E_EMULATOR=${BMO_E2E_EMULATOR}"
 export E2E_CONF_FILE="${REPO_ROOT}/test/e2e/config/ironic.yaml"
 export E2E_BMCS_CONF_FILE="${REPO_ROOT}/test/e2e/config/bmcs-${BMC_PROTOCOL}.yaml"
 
+GINKGO_FOCUS="${GINKGO_FOCUS:-}"
+
 case "${GINKGO_FOCUS,,}" in
   *upgrade*)
     export DEPLOY_IRONIC="false"
--- a/hack/e2e/ensure_docker.sh
+++ b/hack/e2e/ensure_docker.sh
@@ -1,0 +1,136 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+# Verify mode turned off by default
+VERIFY_ONLY="${VERIFY_ONLY:-false}"
+
+# Source os-release to get ID variable
+# shellcheck source=/dev/null
+. /etc/os-release
+
+# Description:
+#   Prompt the user for a yes/no confirmation.
+#
+# Usage:
+#   confirm <default> <message>
+#   <default> must be "y" or "n" (case-insensitive).
+#
+# Returns:
+#   Exit code 0 (true) if the user confirms; 1 (false) otherwise.
+confirm() {
+    local default message prompt reply
+
+    # normalize to lowercase
+    default="${1,,}"
+    if [[ "${default}" == "y" ]]; then
+        prompt="[Y/n]"
+    else
+        prompt="[y/N]"
+    fi
+
+    message="${2}"
+    while true; do
+        read -r -p "${message} Do you want to continue? ${prompt} " reply
+        # use default if Enter pressed
+        reply="${reply:-${default}}"
+        case "${reply,,}" in
+            y) return 0 ;;
+            n) return 1 ;;
+            *) echo "Please enter y or n." ;;
+        esac
+    done
+}
+
+# Description:
+#   Install Docker on Ubuntu.
+#
+# Usage:
+#   install_docker_ubuntu download_url
+install_docker_ubuntu() {
+    local DOCKER_DOWNLOAD_URL
+
+    DOCKER_DOWNLOAD_URL="${1}"
+
+    sudo apt-get install -y ca-certificates curl
+    sudo install -m 0755 -d /etc/apt/keyrings
+    sudo curl -fsSL \
+        -o /etc/apt/keyrings/docker.asc \
+        "${DOCKER_DOWNLOAD_URL}/gpg"
+    sudo chmod a+r /etc/apt/keyrings/docker.asc
+    echo \
+        "deb [arch=$(dpkg --print-architecture)" \
+        "signed-by=/etc/apt/keyrings/docker.asc]" \
+        "${DOCKER_DOWNLOAD_URL}" \
+        "${UBUNTU_CODENAME:-$VERSION_CODENAME} stable" | \
+        sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
+    sudo apt-get update
+    sudo apt-get install -y \
+        docker-ce \
+        docker-ce-cli \
+        containerd.io \
+        docker-buildx-plugin \
+        docker-compose-plugin
+    sudo usermod -aG docker "${USER}"
+
+    echo "Docker installed successfully, please log"\
+         "out and log back in to apply the new group membership"
+}
+
+# Description:
+#   Prompt the user for confirmation and execute a function if confirmed.
+#
+# Usage:
+#   confirm_and_run <default> <message> <function> [args...]
+#
+# Returns:
+#   Exit code 0 (true) if confirmed and function ran; 1 (false) if declined.
+confirm_and_run() {
+    local default fn message
+
+    local default="${1}"
+    local message="${2}"
+    local fn="${3}"
+    shift 3
+
+   if confirm "${default}" "${message}"; then
+        set -x
+        "${fn}" "$@"
+        set +x
+    else
+        echo "Aborted."
+        return 1
+    fi
+}
+
+# Check if docker is installed and install it if not
+verify_docker() {
+    local DOCKER URL
+
+    DOCKER="$(command -v docker || true)"
+    if ! [[ -x "${DOCKER}" ]]; then
+        if [[ "${VERIFY_ONLY}" != "false" ]]; then
+            echo "docker is not in PATH"
+            return 0
+        fi
+        echo "docker not found, installing"
+        case "${ID}" in
+            ubuntu)
+                URL="https://download.docker.com/linux/ubuntu"
+                confirm_and_run \
+                    "n" \
+                    "This action will install the latest Docker from ${URL}." \
+                    install_docker_ubuntu \
+                    "${URL}"
+                ;;
+            *)
+                echo "ERROR: Distribution not supported: ${ID}"
+                return 2
+                ;;
+        esac
+    else
+        echo "$(${DOCKER} --version) is installed at ${DOCKER}"
+    fi
+}
+
+verify_docker
--- a/hack/e2e/ensure_e2e_prerequisites.sh
+++ b/hack/e2e/ensure_e2e_prerequisites.sh
@@ -1,0 +1,140 @@
+#!/usr/bin/env bash
+
+# -----------------------------------------------------------------------------
+# Description: This script sets up the environment for E2E tests. It can also
+#              be used in verify mode to check if the required tools are installed
+#              without making any changes to the system.
+# Usage:       To set up the environment: ./ensure_e2e_prerequisites.sh
+#              To verify prerequisites without installing:
+#              VERIFY_ONLY=1 ./ensure_e2e_prerequisites.sh
+# -----------------------------------------------------------------------------
+
+set -eu
+
+REPO_ROOT="$(realpath "$(dirname "${BASH_SOURCE[0]}")/../..")"
+VERIFY_ONLY="${VERIFY_ONLY:-false}"
+
+# Source os-release to get ID variable
+# shellcheck source=/dev/null
+. /etc/os-release
+
+declare -A PACKAGES=(
+  [build-essential]=""
+  [tar]=""
+  [libvirt-daemon-system]=""
+  [qemu-kvm]="qemu-system-x86"
+  [virt-manager]=""
+)
+
+declare -A TOOLS=(
+  [ubuntu]="install_pkg_ubuntu dpkg"
+)
+
+# Ensure required tools are available
+ensure_tools() {
+  local missing_tools cmd
+
+  # Ensure required low-level tools are present
+  missing_tools=()
+  for cmd in ${TOOLS[${ID}]}; do
+    if ! command -v "${cmd}" >/dev/null; then
+      missing_tools+=("${cmd}")
+    fi
+  done
+
+  if ((${#missing_tools[@]})); then
+    echo "ERROR: Missing required tools: ${missing_tools[*]}"
+    return 2
+  fi
+
+  return 0
+}
+
+# Install packages using apt-get on Ubuntu
+install_pkg_ubuntu() {
+  local packages
+
+  packages=("$@")
+
+  sudo apt-get update
+  sudo apt-get install -y "${packages[@]}"
+}
+
+# Ensure required packages are installed, or install them if not#
+ensure_packages() {
+  local missing installed packages_to_check alt check_tool install_tool
+  local version tools
+
+  read -ra tools <<< "${TOOLS[${ID}]}"
+
+  install_tool="${tools[0]}"
+  check_tool="${tools[1]}"
+
+  # Collect packages that are not satisfied by any alias
+  missing=()
+  for pkg in "${!PACKAGES[@]}"; do
+    # check any alias is installed
+    installed=""
+    packages_to_check=("${pkg}")
+    if [[ -n "${PACKAGES[${pkg}]}" ]]; then
+      read -ra aliases <<< "${PACKAGES[${pkg}]}"
+      packages_to_check+=("${aliases[@]}")
+    fi
+    for alt in "${packages_to_check[@]}"; do
+      version="$("${check_tool}" -s "${alt}" 2>/dev/null \
+                  | grep '^Version:' \
+                  || true)"
+      if [[ -n "${version}" ]]; then
+        installed="${alt} (${version})"
+        break
+      fi
+    done
+
+    if [[ -n "${installed}" ]]; then
+      echo "Package ${pkg} satisfied by installed package: ${installed}"
+      continue
+    fi
+
+    if [[ "${VERIFY_ONLY}" != "false" ]]; then
+      if [[ -n "${PACKAGES[${pkg}]}" ]]; then
+        echo "Package ${pkg} (aliases: ${PACKAGES[${pkg}]}) is not installed"
+      else
+        echo "Package ${pkg} is not installed"
+      fi
+      continue
+    fi
+
+    missing+=("${pkg}")
+  done
+
+  if [[ "${VERIFY_ONLY}" == "false" && ${#missing[@]} -gt 0 ]]; then
+    echo "Installing missing packages: ${missing[*]}"
+    set -x
+    "${install_tool}" "${missing[@]}"
+    set +x
+  fi
+}
+
+# Ensure all prerequisites are installed
+ensure_components() {
+  export PATH="${PATH}:/usr/local/go/bin"
+  VERIFY_ONLY="${VERIFY_ONLY}" "${REPO_ROOT}/hack/e2e/ensure_kubectl.sh"
+  VERIFY_ONLY="${VERIFY_ONLY}" "${REPO_ROOT}/hack/e2e/ensure_yq.sh"
+  VERIFY_ONLY="${VERIFY_ONLY}" "${REPO_ROOT}/hack/e2e/ensure_go.sh"
+  VERIFY_ONLY="${VERIFY_ONLY}" "${REPO_ROOT}/hack/e2e/ensure_htpasswd.sh"
+  VERIFY_ONLY="${VERIFY_ONLY}" "${REPO_ROOT}/hack/e2e/ensure_docker.sh"
+}
+
+# Main execution
+case "${ID}" in
+  ubuntu)
+    ;;
+  *)
+    echo "ERROR: Distribution not supported: ${ID}"
+    return 2
+    ;;
+esac
+
+ensure_tools
+ensure_packages
+ensure_components
--- a/hack/e2e/ensure_go.sh
+++ b/hack/e2e/ensure_go.sh
@@ -1,34 +1,55 @@
 #!/usr/bin/env bash
 
-set -eux
+set -eu
 
 MINIMUM_GO_VERSION=go1.25.7
 
+# Verify mode turned off by default
+VERIFY_ONLY="${VERIFY_ONLY:-false}"
+
 # Ensure the go tool exists and is a viable version, or installs it
-verify_go_version()
-{
+verify_go_version() {
     # If go is not available on the path, get it
-    if ! [ -x "$(command -v go)" ]; then
+    GO="$(command -v go || true)"
+    if ! [[ -x "${GO}" ]]; then
+        if [[ "${VERIFY_ONLY}" != "false" ]]; then
+            echo "go is not in PATH"
+            return 0
+        fi
         if [[ "${OSTYPE}" == "linux-gnu" ]]; then
-            echo 'go not found, installing'
-            curl -sLo "/tmp/${MINIMUM_GO_VERSION}.linux-amd64.tar.gz" "https://go.dev/dl/${MINIMUM_GO_VERSION}.linux-amd64.tar.gz"
-            sudo tar -C /usr/local -xzf "/tmp/${MINIMUM_GO_VERSION}.linux-amd64.tar.gz"
-            export PATH=/usr/local/go/bin:$PATH
+            echo "go not found, installing"
+            set -x
+            curl -sL \
+                -o "/tmp/${MINIMUM_GO_VERSION}.linux-amd64.tar.gz" \
+                "https://go.dev/dl/${MINIMUM_GO_VERSION}.linux-amd64.tar.gz"
+            sudo tar \
+                -C /usr/local \
+                -xzf "/tmp/${MINIMUM_GO_VERSION}.linux-amd64.tar.gz"
+            set +x
+            export PATH="${PATH}:/usr/local/go/bin"
+            GO="$(command -v go)"
         else
-            echo "Missing required binary in path: go"
+            echo "ERROR: Missing required binary in path: go"
             return 2
         fi
     fi
 
     local go_version
-    IFS=" " read -ra go_version <<< "$(go version)"
-    if [[ "${MINIMUM_GO_VERSION}" != $(echo -e "${MINIMUM_GO_VERSION}\n${go_version[2]}" | sort -s -t. -k 1,1 -k 2,2n -k 3,3n | head -n1) ]] && [[ "${go_version[2]}" != "devel" ]]; then
+    IFS=" " read -ra go_version <<< "$("${GO}" version)"
+    if [[ "${MINIMUM_GO_VERSION}" != $(
+        echo -e "${MINIMUM_GO_VERSION}\n${go_version[2]}" \
+        | sort -s -t. -k 1,1 -k 2,2n -k 3,3n \
+        | head -n1
+    ) ]] && \
+       [[ "${go_version[2]}" != "devel" ]]; then
         cat << EOF
 Detected go version: ${go_version[2]}.
 Requires ${MINIMUM_GO_VERSION} or greater.
 Please install ${MINIMUM_GO_VERSION} or later.
 EOF
         return 2
+    else
+        echo "${go_version[2]} is installed at ${GO}"
     fi
 }
 
--- a/hack/e2e/ensure_htpasswd.sh
+++ b/hack/e2e/ensure_htpasswd.sh
@@ -1,14 +1,24 @@
 #!/usr/bin/env bash
 
-set -eux
+set -eu
+
+# Verify mode turned off by default
+VERIFY_ONLY="${VERIFY_ONLY:-false}"
 
 # Check if htpasswd is installed and install it if not
-verify_htpasswd()
-{
-    if ! [ -x "$(command -v htpasswd)" ]; then
+verify_htpasswd() {
+    if ! [[ -x "$(command -v htpasswd)" ]]; then
+        if [[ "${VERIFY_ONLY}" != "false" ]]; then
+            echo "htpasswd is not in PATH"
+            return 0
+        fi
         echo "htpasswd could not be found, installing..."
+        set -x
         sudo apt-get update
         sudo apt-get install -y apache2-utils
+        set +x
+    else
+        echo "htpasswd is installed at $(command -v htpasswd)"
     fi
 }
 
--- a/hack/e2e/ensure_kubectl.sh
+++ b/hack/e2e/ensure_kubectl.sh
@@ -14,35 +14,55 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set -eux
+set -eu
 
 USR_LOCAL_BIN="/usr/local/bin"
 MINIMUM_KUBECTL_VERSION=v1.34.1
+KUBECTL_DOWNLOAD_URL="https://dl.k8s.io/release"
+
+# Verify mode turned off by default
+VERIFY_ONLY="${VERIFY_ONLY:-false}"
 
 # Ensure the kubectl tool exists and is a viable version, or installs it
-verify_kubectl_version()
-{
+verify_kubectl_version() {
     # If kubectl is not available on the path, get it
-    if ! [ -x "$(command -v kubectl)" ]; then
+    KUBECTL="$(command -v kubectl || true)"
+    if ! [[ -x "${KUBECTL}" ]]; then
+        if [[ "${VERIFY_ONLY}" != "false" ]]; then
+            echo "kubectl is not in PATH"
+            return 0
+        fi
         if [[ "${OSTYPE}" == "linux-gnu" ]]; then
             echo "kubectl not found, installing"
-            curl -LO "https://dl.k8s.io/release/${MINIMUM_KUBECTL_VERSION}/bin/linux/amd64/kubectl"
-            sudo install kubectl "${USR_LOCAL_BIN}/kubectl"
+            set -x
+            curl -LO \
+                --create-dirs \
+                --output-dir "/tmp" \
+                "${KUBECTL_DOWNLOAD_URL}/${MINIMUM_KUBECTL_VERSION}/bin/linux/amd64/kubectl"
+            sudo install "/tmp/kubectl" "${USR_LOCAL_BIN}/kubectl"
+            set +x
+            KUBECTL="$(command -v kubectl)"
         else
-            echo "Missing required binary in path: kubectl"
+            echo "ERROR: Missing required binary in path: kubectl"
             return 2
         fi
     fi
 
     local kubectl_version
-    IFS=" " read -ra kubectl_version <<< "$(kubectl version --client)"
-    if [[ "${MINIMUM_KUBECTL_VERSION}" != $(echo -e "${MINIMUM_KUBECTL_VERSION}\n${kubectl_version[2]}" | sort -s -t. -k 1,1 -k 2,2n -k 3,3n | head -n1) ]]; then
+    IFS=" " read -ra kubectl_version <<< "$("${KUBECTL}" version --client)"
+    if [[ "${MINIMUM_KUBECTL_VERSION}" != $(
+        echo -e "${MINIMUM_KUBECTL_VERSION}\n${kubectl_version[2]}" \
+        | sort -s -t. -k 1,1 -k 2,2n -k 3,3n \
+        | head -n1
+    ) ]]; then
         cat << EOF
 Detected kubectl version: ${kubectl_version[2]}.
 Requires ${MINIMUM_KUBECTL_VERSION} or greater.
 Please install ${MINIMUM_KUBECTL_VERSION} or later.
 EOF
         return 2
+    else
+        echo "kubectl version ${kubectl_version[2]} is installed at ${KUBECTL}"
     fi
 }
 
--- a/test/e2e/README.md
+++ b/test/e2e/README.md
@@ -15,6 +15,25 @@ as BMC. Ironic runs in the "host network" of the kind cluster in the test. The
 kind cluster is then configured to expose the relevant ports on the actual host
 so that they can be reached from the BareMetalHost VMs.
 
+However, before running the script ensure the environment is setup correctly:
+
+```bash
+make verify-e2e-prerequisites
+```
+
+This check uses a script that currently runs only on Ubuntu. If any required
+components are missing, run the same script to install them:
+
+```bash
+# After running the script, you may need to log out/in (or start a new shell)
+# before docker works without sudo.
+./hack/e2e/ensure_e2e_prerequisites.sh
+```
+
+Please note that the script installs Go in `/usr/local/go`. If you want to invoke
+the e2e Makefile target directly with `make test-e2e`, you may need to adjust your
+`PATH` accordingly.
+
 Currently there are two sets of tests, which cannot be ran together in the same
 cluster. One is the "optional" set, currently consists of only the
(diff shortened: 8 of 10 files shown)


sirpaleet:
In case no-one wants to pick this up I can just make this PR about this one target as well, and add some comments to documentation :)

Makefile
```diff
@@ -120,6 +123,58 @@ unit-verbose: ## Run unit tests with verbose output
...
+## NEXT:
+## - `test-e2e` is used in ci-e2e.sh script so make a new target
+## - expand on `clean-e2e` to have optional steps on cleanup with flags
+## - update info in docs/testing.md and test/e2e/README.md
+##
+## Installing required packages for running e2e scripts locally
+## These installations work for linux amd64, at least on Ubuntu 24.04
+## NOTE: you may need to log out and back in to apply all changes
+.PHONY: e2e-prerequisites
+e2e-prerequisites:
```
tuminoid: I don't think we want this much bash in Makefile, rather have a new setup script that the makefile runs.

Makefile
```diff
@@ -120,6 +123,58 @@ unit-verbose: ## Run unit tests with verbose output
...
+## then in other places the scripts would not know to find them
+##
+## NEXT:
+## - `test-e2e` is used in ci-e2e.sh script so make a new target
+## - expand on `clean-e2e` to have optional steps on cleanup with flags
+## - update info in docs/testing.md and test/e2e/README.md
+##
+## Installing required packages for running e2e scripts locally
+## These installations work for linux amd64, at least on Ubuntu 24.04
+## NOTE: you may need to log out and back in to apply all changes
```
nuhakala: I think this comment could be rewritten to be slightly more understandable. The first paragraph seems to have two points, maybe create a list or two paragraphs from it. The NEXT section seems to be like things to improve next, maybe it could be renamed as TODO or "things to improve next". Now I got the impression that I need to do those things next after running this target (though the context reveals that this is not the case).

Also "but then in other places the scripts would not know to find them" -> "the scripts would not know how to find them" ?

Makefile
nuhakala: Adding checks for existing installations will be also easier in a bash script.

Makefile
sirpaleet: Ah this was supposed to be notes for this PR but I'll take them off and make a separate comment on here then :D

hack/e2e-prerequisites.sh
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env bash
```
sirpaleet: Added this to script

sirpaleet:
Additional checks for if installations exist could be added.
The installations could be made to the project's bin, but then in other places the scripts would not know how to find them.

TODO:
- `test-e2e` is used in ci-e2e.sh script so make a new target
- expand on `clean-e2e` to have optional steps on cleanup with flags
- update info in docs/testing.md and test/e2e/README.md

Makefile
```diff
@@ -120,6 +120,13 @@ unit-verbose: ## Run unit tests with verbose output
 
 ARTIFACTS ?= ${ROOT_DIR}/test/e2e/_artifacts
 
+## Installing required packages for running e2e scripts locally
+## These installations work for linux amd64, at least on Ubuntu 24.04
+## NOTE: you may need to log out and back in to apply all changes
+.PHONY: e2e-prerequisites
```
tuminoid: Add another note that this should not be run on your laptop, but in a VM.

hack/e2e-prerequisites.sh
```diff
@@ -0,0 +1,55 @@
...
+#              GO_VERSION can be set separately as well.
+# -----------------------------------------------------------------------------
+
+USR_LOCAL_BIN="/usr/local/bin"
+MINIMUM_KUBECTL_VERSION=v1.28.1
+YQ_VERSION="v4.40.5"
+GO_VERSION="${GO_VERSION:-"1.24.7"}"
+
+echo "INSTALLING some core packages"
+sudo apt-get install build-essential tar
```
tuminoid:
```suggestion
sudo apt-get install -y build-essential tar
```

hack/e2e-prerequisites.sh
```diff
@@ -0,0 +1,55 @@
...
+echo "INSTALLING go, removing old versions"
+sudo rm -rf /usr/local/go
```
tuminoid: Hell no, don't go nuking that entirely. Use version specific directory, and don't reinstall if its there.

hack/e2e-prerequisites.sh
```diff
@@ -0,0 +1,55 @@
...
+wget -P $HOME https://go.dev/dl/go$GO_VERSION.linux-amd64.tar.gz
```
tuminoid:
```suggestion
wget -P "${HOME}" "https://go.dev/dl/go$GO_VERSION.linux-amd64.tar.gz"
```

hack/e2e-prerequisites.sh
```diff
@@ -0,0 +1,55 @@
...
+sudo tar -C /usr/local -xzf $HOME/go$GO_VERSION.linux-amd64.tar.gz
```
tuminoid: Fix all of the quotation and curly braces, too many to comment one by one.

hack/e2e-prerequisites.sh
```diff
@@ -0,0 +1,55 @@
...
+echo "EXPORTING go bin path, optionally add to .bashrc or .profile"
+export PATH="${PATH}:/usr/local/go/bin"
+go version
+type go
+sudo apt-get update
```
tuminoid: you want to run this before the "install core packages" step earlier.

hack/e2e-prerequisites.sh
```diff
@@ -0,0 +1,55 @@
...
+#              These installations work for linux amd64, at least on Ubuntu
+#              24.04.
+#              NOTE: you may need to log out and back in to apply all changes
+# Usage:       This script is usually ran from `make e2e-prerequisites`, but can
+#              be used separately with GO_VERSION=1.24.7 ./e2e-prerequisites.sh.
+#              GO_VERSION can be set separately as well.
+# -----------------------------------------------------------------------------
+
+USR_LOCAL_BIN="/usr/local/bin"
+MINIMUM_KUBECTL_VERSION=v1.28.1
```
tuminoid: kubectl is supporing version skew of 2 minors, so this is way too old.

hack/e2e-prerequisites.sh
```diff
@@ -0,0 +1,55 @@
...
+make build
+echo "INSTALLING kubectl, assuming no previous version exists and curl is \
+version 7.73.0 or later"
+curl --create-dirs -LO --output-dir $HOME "https://dl.k8s.io/release/${MINIMUM_KUBECTL_VERSION}/bin/linux/amd64/kubectl"
+sudo install $HOME/kubectl "${USR_LOCAL_BIN}/kubectl"
+echo "INSTALLING yq, assuming no previous version exists"
+go install github.com/mikefarah/yq/v4@$YQ_VERSION
+echo "INSTALLING docker and related, removing old versions of docker"
+echo "ADDING repository to apt sources, GPG key, and other changes"
+for pkg in docker.io docker-doc docker-compose docker-compose-v2 podman-docker containerd runc; do sudo apt-get remove $pkg; done
```
tuminoid: Why isn't a version of docker sufficing here? If docker is present, we should not go and nuke it.

hack/e2e-prerequisites.sh
```diff
@@ -0,0 +1,55 @@
...
+sudo apt-get update
```
tuminoid: we didn't yet add any repos, this update does nothing

hack/e2e-prerequisites.sh
```diff
@@ -0,0 +1,55 @@
...
+sudo apt-get install ca-certificates
+sudo install -m 0755 -d /etc/apt/keyrings
+sudo curl -fsSL https://download.docker.com/linux/ubuntu/gpg -o /etc/apt/keyrings/docker.asc
+sudo chmod a+r /etc/apt/keyrings/docker.asc
+echo \
+  "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.asc] https://download.docker.com/linux/ubuntu \
+  $(. /etc/os-release && echo "${UBUNTU_CODENAME:-$VERSION_CODENAME}") stable" | \
+  sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
+sudo apt-get update
+sudo apt-get install docker-ce docker-ce-cli containerd.io docker-buildx-plugin docker-compose-plugin
```
tuminoid:
```suggestion
sudo apt-get install -y docker-ce docker-ce-cli containerd.io docker-buildx-plugin docker-compose-plugin
```

hack/e2e-prerequisites.sh
```diff
@@ -0,0 +1,55 @@
...
+sudo chmod 777 /var/run/docker.sock
```
tuminoid: Please no. The `docker` group exists for a reason, and you're implementing that on following lines. 777 just allows anyone root basically.

hack/e2e-prerequisites.sh
sirpaleet: Ah this was from the official instructions, I'll take it off then

hack/e2e-prerequisites.sh
sirpaleet: Is this a general convention or somehow related to securing the read of the path etc?

hack/e2e-prerequisites.sh
sirpaleet: a-ha I got it from here: https://github.com/metal3-io/baremetal-operator/blob/main/hack/e2e/ensure_kubectl.sh just using what we install, I'll look at installing the latest stable then for this

hack/e2e-prerequisites.sh
tuminoid: Curlys are just readability/maintenance thing, prevents some unintended variable name faults.
The quotes are security thing, as without curlys you can have command injection or unintended side effects if the variable would contain a space.

hack/e2e-prerequisites.sh
tuminoid: For example, on my machine, it would nuke all of my local Go versions, not just one. Go doesn't care where you put it, as long as the go environment variables and $PATH is correct.

sirpaleet: Is there a reason why we are using this specific go version? Is it better to be the same then? I could just edit the https://github.com/metal3-io/baremetal-operator/blob/main/hack/e2e/ensure_go.sh file maybe then, and that does the check and installs only if no version exists and also complains about the version mismatch. This does however leave the upgrade to the user's responsibility but is that okay? (it could then easily be expanded on later meybe)

Maybe I should do a similar script for checking the docker, too? I just wouldn't complain about versions then?

tuminoid: We set certain version of Go as it is fixing vulnerabilities that affect the product. For e2e, it wouldn't matter too much, but also, why not use the same version?

Note that ensure-go.sh and others are strictly meant for e2e only and e2e only runs in disposable machines. Hence the earlier mention that it should not be run on anyone's laptop. If we establish that, then sure use the ensure-go.sh and install Go to the directory it expects.

hack/e2e-prerequisites.sh
tuminoid: Maybe make another PR where you bump the kubectl version in ensure_kubectl.sh to 1.33.x

hack/e2e-prerequisites.sh
sirpaleet: I added the comment on this being run on vm:s only, I'll use these then :)

sirpaleet:
I noticed that in here: https://github.com/metal3-io/baremetal-operator/blob/main/hack/ci-e2e.sh#L57 there is clear attempt at making the script self-sufficient and it should probably install all needed packages. But alas it does not. I'll expand on just this and remove the new target :D + edit the other scripts to fit this

sirpaleet:
This has now been confined to the scope of installations, and adding the script target. I'll open the PR but I will most likely not have the time to edit all changes if they are major. I'll make an issue on the "TODO" i added before. 

EDIT: #2707 has the cleanup issue, and #2708 has the kubectl version bump (which I unfortunately do not have time for)

> I noticed that in here: https://github.com/metal3-io/baremetal-operator/blob/main/hack/ci-e2e.sh#L57 there is clear attempt at making the script self-sufficient and it should probably install all needed packages. But alas it does not. I'll expand on just this and remove the new target :D + edit the other scripts to fit this

Additionally based on this finding, changed from a new target for installations to adding the necessary steps to the hack/ci-e2e.sh and making a new ensure_docker.sh for the docker installation.

hack/e2e-prerequisites.sh
sirpaleet: I added a new script for this, ensure_docker.sh

hack/e2e/ensure_docker.sh
```diff
@@ -0,0 +1,32 @@
...
+        if [[ "${OSTYPE}" == "linux-gnu" ]]; then
+            echo "docker not found, installing"
+            sudo apt-get install ca-certificates
+            sudo install -m 0755 -d /etc/apt/keyrings
+            sudo curl -fsSL https://download.docker.com/linux/ubuntu/gpg -o /etc/apt/keyrings/docker.asc
+            sudo chmod a+r /etc/apt/keyrings/docker.asc
+            # shellcheck source=/etc/os-release
+            echo \
+              "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.asc] https://download.docker.com/linux/ubuntu \
+              $(. /etc/os-release && echo "${UBUNTU_CODENAME:-$VERSION_CODENAME}") stable" | \
```
sirpaleet: Not quite sure how to deal with this shellcheck warning now, will look at in morning if I have time. This did not work here.

hack/e2e/ensure_docker.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+# shellcheck disable=SC1091
```
sirpaleet: Added this to fix the shellcheck issue, this code is from the official installation instructions of installing Docker Engine on Ubuntu.

hack/ci-e2e.sh
```diff
@@ -45,16 +45,20 @@ export E2E_BMCS_CONF_FILE="${REPO_ROOT}/test/e2e/config/bmcs-${BMC_PROTOCOL}.yam
...
 # Ensure requirements are installed
 export PATH="/usr/local/go/bin:${PATH}"
 "${REPO_ROOT}/hack/e2e/ensure_go.sh"
 "${REPO_ROOT}/hack/e2e/ensure_htpasswd.sh"
 # CAPI test framework uses kubectl in the background
 "${REPO_ROOT}/hack/e2e/ensure_kubectl.sh"
 "${REPO_ROOT}/hack/e2e/ensure_yq.sh"
+"${REPO_ROOT}/hack/e2e/ensure_docker.sh"
 
 sudo apt-get update
```
tuminoid:
```suggestion
sudo apt-get update
sudo apt-get install -y build-essential tar

# Ensure requirements are installed
export PATH="/usr/local/go/bin:${PATH}"
"${REPO_ROOT}/hack/e2e/ensure_go.sh"
"${REPO_ROOT}/hack/e2e/ensure_htpasswd.sh"
# CAPI test framework uses kubectl in the background
"${REPO_ROOT}/hack/e2e/ensure_kubectl.sh"
"${REPO_ROOT}/hack/e2e/ensure_yq.sh"
"${REPO_ROOT}/hack/e2e/ensure_docker.sh"

```

hack/e2e/ensure_docker.sh
```diff
@@ -0,0 +1,33 @@
...
+
+set -eux
+
+# Check if docker is installed and install it if not.
+# This is optimized to install on Ubuntu.
+# The `newgrp` will stop a non-interactive script run.
+verify_docker()
+{
+    if ! [ -x "$(command -v docker)" ]; then
```
tuminoid:
```suggestion
    if ! [[ -x "$(command -v docker)" ]]; then
```

hack/e2e/ensure_docker.sh
```diff
@@ -0,0 +1,33 @@
...
+        if [[ "${OSTYPE}" == "linux-gnu" ]] && [[ "$(. /etc/os-release && echo "$ID")" == "ubuntu" ]]; then
```
tuminoid:
```suggestion
        if [[ "${OSTYPE}" == "linux-gnu" ]] && [[ "$(. /etc/os-release && echo "${ID}")" == "ubuntu" ]]; then
```

hack/e2e/ensure_docker.sh
```diff
@@ -0,0 +1,33 @@
...
+            sudo curl -fsSL https://download.docker.com/linux/ubuntu/gpg -o /etc/apt/keyrings/docker.asc
+            sudo chmod a+r /etc/apt/keyrings/docker.asc
+            echo \
+              "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.asc] https://download.docker.com/linux/ubuntu \
+              $(. /etc/os-release && echo "${UBUNTU_CODENAME:-$VERSION_CODENAME}") stable" | \
+            sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
+            sudo apt-get update
+            sudo apt-get install -y docker-ce docker-ce-cli containerd.io docker-buildx-plugin docker-compose-plugin
+            sudo usermod -aG docker "${USER}"
+            newgrp docker
```
tuminoid: This does nothing

Makefile
```diff
@@ -120,6 +120,15 @@ unit-verbose: ## Run unit tests with verbose output
 
 ARTIFACTS ?= ${ROOT_DIR}/test/e2e/_artifacts
 
+## Checking that required tools exist and running e2e script on a VM
+##
+## To note: Run this on your VM and not your local machine. These installations
+## work for linux amd64, at least on Ubuntu 24.04. You may need to log out and
+## log in to apply all changes if the installations are done for the first time.
+.PHONY: run-e2e-script
```
tuminoid: this is not descriptive name for the target... setup-e2e or such would be nicer

hack/e2e/ensure_docker.sh
sirpaleet: I'm confused as to which part, since it works for me? The new docker group?

Makefile
sirpaleet: I agree the naming is a bit clunky, it's just that the "setup" would imply installations and anything more vague would imply it is used in actual e2e runs elsewhere no? Or maybe not?

sirpaleet:
Left a couple of comments, and fixed some things. This is as much as I have time today, in case someone has an opinion of the furture of this PR I'm all ears but contributing to this has to end on my part for now :( I'll try to answer if someone has a question though!

hack/ci-e2e.sh
```diff
@@ -45,16 +45,20 @@ export E2E_BMCS_CONF_FILE="${REPO_ROOT}/test/e2e/config/bmcs-${BMC_PROTOCOL}.yam
...
 "${REPO_ROOT}/hack/e2e/ensure_htpasswd.sh"
 # CAPI test framework uses kubectl in the background
 "${REPO_ROOT}/hack/e2e/ensure_kubectl.sh"
 "${REPO_ROOT}/hack/e2e/ensure_yq.sh"
+"${REPO_ROOT}/hack/e2e/ensure_docker.sh"
 
-sudo apt-get update
-sudo apt-get install -y libvirt-dev pkg-config
+sudo apt-get install -y libvirt-daemon-system qemu-kvm virt-manager libvirt-dev
+make build
```
sirpaleet: This might be repetition with L82, but this does some necessary steps also, just might slow down the script

terror96:
Addresses also: https://github.com/metal3-io/baremetal-operator/issues/2707

hack/e2e/ensure_docker.sh
```diff
@@ -0,0 +1,58 @@
...
+        case "${ID}" in
+            ubuntu)
+                install_docker_ubuntu
+                # Return an error code to interrupt the call sequence. This is
+                # necessary for the new group membership to take effect.
+                return 2
+                ;;
+            *)
+                echo "Distribution not supported: ${ID}"
+                return 2
```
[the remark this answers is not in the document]
terror96: Well, this was the whole point but converting to 0 value instead. Other scripts use '2', so I rather keep it.

hack/prerequisites-e2e.sh
```diff
@@ -0,0 +1,16 @@
...
+set -eux
+
+sudo apt-get update
+sudo apt-get install -y \
+    build-essential \
+    tar \
+    libvirt-daemon-system \
+    qemu-kvm \
+    virt-manager \
+    libvirt-dev
```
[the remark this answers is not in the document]
terror96: Oops.

test/e2e/README.md
```diff
@@ -15,6 +15,13 @@ as BMC. Ironic runs in the "host network" of the kind cluster in the test. The
 kind cluster is then configured to expose the relevant ports on the actual host
 so that they can be reached from the BareMetalHost VMs.
 
+There is a Makefile target for running this e2e script called `run-e2e-tests`.
+However, before running the target ensure the environment is setup correctly:
+
+```bash
+make ensure-e2e-environment
+```
+
```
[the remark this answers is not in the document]
terror96: Agree.

Makefile
```diff
@@ -142,6 +142,18 @@ fuzz-run: ## Run all fuzz tests sequentially with fuzzing enabled (use FUZZ_TIME
...
+## Checking that required tools exist on VM
+.PHONY: ensure-e2e-environment
+ensure-e2e-environment:
+	@echo "Ensuring environment is ready for e2e tests..."
+	./hack/prerequisites-e2e.sh
+	./hack/e2e/ensure_docker.sh
+
+## Run e2e tests on VM
+.PHONY: run-e2e-tests
+run-e2e-tests:
```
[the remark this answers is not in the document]
terror96: Agree.

hack/e2e/ensure_docker.sh
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env bash
+
+set -eux
```
elfosardo: -o pipefail too would be good

hack/ci-e2e.sh
```diff
@@ -55,9 +57,6 @@ export PATH="/usr/local/go/bin:${PATH}"
 "${REPO_ROOT}/hack/e2e/ensure_kubectl.sh"
 "${REPO_ROOT}/hack/e2e/ensure_yq.sh"
 
-sudo apt-get update
```
terror96, line 58: I realized that we should keep these here in order to prevent failures in case the script is called from somewhere else than via Makefile.

terror96:
Ping @lentzi90 @tuminoid suggested to include you in the process.

tuminoid:
/hold
for @lentzi90

Makefile
```diff
@@ -142,6 +142,16 @@ fuzz-run: ## Run all fuzz tests sequentially with fuzzing enabled (use FUZZ_TIME
...
 ARTIFACTS ?= ${ROOT_DIR}/test/e2e/_artifacts
 
+.PHONY: ensure-e2e-environment
+ensure-e2e-environment: ## Checking that required tools exist on VM
+	@echo "Ensuring environment is ready for e2e tests..."
+	./hack/prerequisites-e2e.sh
+	./hack/e2e/ensure_docker.sh
+
+.PHONY: run-e2e-tests
+run-e2e-tests: ## Run e2e tests on VM
```
lentzi90: This will be confusing. It is hard to tell what the difference is between `run-e2e-tests` and `test-e2e`. I would rather skip it. The one above (`ensure-e2e-environment`) is good though.

terror96: This is a fair point. Let's get rid of this target but maybe still keep `ensure-e2e-environment`?

terror96:
> I appreciate the intent of making it easier to run the e2e tests. However, I am also a bit hesitant to include so specific and invasive scripts to be called directly from the Makefile. I would rather see that anything used directly from the Makefile just checks that the environment has what is needed, not try to `sudo <install-command>` unconditionally. Not everyone installs `docker` this way and this could cause conflicts.
> 
> On the other hand, I am fine with adding hack scripts to use separate from the Makefile. The way I see it, the Makefile is a sort of "API" for developers and we should try to keep that "clean" from too much assumptions and invasive actions.

I might be a bit confused about the order of comments especially because GitHub shows the same comment twice and the first one shows your additional comment: "The one above (ensure-e2e-environment) is good though.".

I therefore assume that you would like to keep `ensure-e2e-environment` target but maybe change the scripts in a way that they do not install components without user's explicit acknowledgment. One possible way to meet this requirement would be to adjust `prerequisites-e2e.sh` not to have `-y`. Also, we could change `ensure_docker.sh` to display first what it is going to do and then ask user's confirmation before proceeding. Would that be better?

terror96:
> > I appreciate the intent of making it easier to run the e2e tests. However, I am also a bit hesitant to include so specific and invasive scripts to be called directly from the Makefile. I would rather see that anything used directly from the Makefile just checks that the environment has what is needed, not try to `sudo <install-command>` unconditionally. Not everyone installs `docker` this way and this could cause conflicts.
> > On the other hand, I am fine with adding hack scripts to use separate from the Makefile. The way I see it, the Makefile is a sort of "API" for developers and we should try to keep that "clean" from too much assumptions and invasive actions.
> 
> I might be a bit confused about the order of comments especially because GitHub shows the same comment twice and the first one shows your additional comment: "The one above (ensure-e2e-environment) is good though.".
> 
> I therefore assume that you would like to keep `ensure-e2e-environment` target but maybe change the scripts in a way that they do not install components without user's explicit acknowledgment. One possible way to meet this requirement would be to adjust `prerequisites-e2e.sh` not to have `-y`. Also, we could change `ensure_docker.sh` to display first what it is going to do and then ask user's confirmation before proceeding. Would that be better?

which could be something like this:

```
#!/usr/bin/env bash

set -euo pipefail

# Source os-release to get ID variable
# shellcheck source=/dev/null
source /etc/os-release

# Description:
#   Prompt the user for a yes/no confirmation.
#
# Usage:
#   confirm <default> <message>
#   <default> must be "y" or "n" (case-insensitive).
#
# Returns:
#   Exit code 0 (true) if the user confirms; 1 (false) otherwise.
confirm() {
    local default message prompt reply

    # normalize to lowercase
    default="${1,,}"
    if [[ "${default}" == "y" ]]; then
        prompt="[Y/n]"
    else
        prompt="[y/N]"
    fi

    message="${2}"
    while true; do
        read -r -p "${message} Do you want to continue? ${prompt} " reply
        # use default if Enter pressed
        reply="${reply:-$default}"
        case "${reply,,}" in
            y) return 0 ;;
            n) return 1 ;;
            *) echo "Please enter y or n." ;;
        esac
    done
}

# Description:
#   Install Docker on Ubuntu.
#
# Usage:
#   install_docker_ubuntu download_url
#
install_docker_ubuntu() {
    local DOCKER_DOWNLOAD_URL

    DOCKER_DOWNLOAD_URL="${1}"

    sudo apt-get install -y ca-certificates curl
    sudo install -m 0755 -d /etc/apt/keyrings
    sudo curl -fsSL \
        -o /etc/apt/keyrings/docker.asc \
        "${DOCKER_DOWNLOAD_URL}/gpg"
    sudo chmod a+r /etc/apt/keyrings/docker.asc
    echo \
        "deb [arch=$(dpkg --print-architecture)" \
        "signed-by=/etc/apt/keyrings/docker.asc]" \
        "${DOCKER_DOWNLOAD_URL}" \
        "${UBUNTU_CODENAME:-$VERSION_CODENAME} stable" | \
        sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
    sudo apt-get update
    sudo apt-get install -y \
        docker-ce \
        docker-ce-cli \
        containerd.io \
        docker-buildx-plugin \
        docker-compose-plugin
    sudo usermod -aG docker "${USER}"

    echo "Docker installed successfully, please log"\
         "out and log back in to apply the new group membership"
}

# Description:
#   Prompt the user for confirmation and execute a function if confirmed.
#
# Usage:
#   confirm_and_run <default> <message> <function> [args...]
#
# Returns:
#   Exit code 0 (true) if confirmed and function ran; 1 (false) if declined.
confirm_and_run() {
    local default fn message

    local default="${1}"
    local message="${2}"
    local fn="${3}"
    shift 3

   if confirm "${default}" "${message}"; then
        set -x
        "${fn}" "$@"
        set +x
    else
        echo "Aborted."
        return 1
    fi
}

# Check if docker is installed and install it if not
verify_docker()
{
    if ! [[ -x "$(command -v docker)" ]]; then
        echo "docker not found, installing"
        case "${ID}" in
            ubuntu)
                local URL="https://download.docker.com/linux/ubuntu"
                confirm_and_run \
                    "n" \
                    "This action will install the latest Docker from ${URL}." \
                    install_docker_ubuntu \
                    "${URL}"
                ;;
            *)
                echo "Distribution not supported: ${ID}"
                return 2
                ;;
        esac
    fi
}

verify_docker
```

lentzi90:
I had some issues with the network or github, which caused the double comments. Sorry about that.
What I would suggest is something like this:
1. Remove `run-e2e-tests` target. It will just confuse people.
2. Keep the `ensure_*.sh` scripts in the hack folder for people who know what they are doing, but don't call them from the Makefile.
3. Change the `ensure-e2e-environement` target to something like `verify-e2e-prerequisites` and use it just as a check that the tools are there. Check that docker, go, etc is installed. NOTE: Libvirt is NOT a requirement. It is only needed when using the `hack/ci-e2e.sh` for a virtualized baremetal lab.

This could then be used in two different ways:
1. Create your own configuration, run `make verify-e2e-prerequisites` as a sanity check, followed by `make test-e2e`. This is the generic scenario that allows running on real hardware and using specific configurations that we may not be testing in CI.
2. The "happy/easy path" that mimics CI on a ubuntu cloud VM: `hack/ensure_e2e_prerequisites.sh` followed by `hack/ci-e2e.sh`.

terror96:
> I had some issues with the network or github, which caused the double comments. Sorry about that. What I would suggest is something like this:
> 
>     1. Remove `run-e2e-tests` target. It will just confuse people.
> 
>     2. Keep the `ensure_*.sh` scripts in the hack folder for people who know what they are doing, but don't call them from the Makefile.
> 
>     3. Change the `ensure-e2e-environement` target to something like `verify-e2e-prerequisites` and use it just as a check that the tools are there. Check that docker, go, etc is installed. NOTE: Libvirt is NOT a requirement. It is only needed when using the `hack/ci-e2e.sh` for a virtualized baremetal lab.
> 
> 
> This could then be used in two different ways:
> 
>     1. Create your own configuration, run `make verify-e2e-prerequisites` as a sanity check, followed by `make test-e2e`. This is the generic scenario that allows running on real hardware and using specific configurations that we may not be testing in CI.
> 
>     2. The "happy/easy path" that mimics CI on a ubuntu cloud VM: `hack/ensure_e2e_prerequisites.sh` followed by `hack/ci-e2e.sh`.

Very good. I'll work a little bit on this, and throw in a new proposal.

Nod, I have not checked the hard requirements to run the tests, `libvirt` was in the original PR and thus I kept it there.

terror96:
Pushed a new version even though there might still be issues. Wanted to give you something to look at...

terror96:
> Pushed a new version even though there might still be issues. Wanted to give you something to look at...

I tested the procedure on a clean VM:

```
make verify-e2e-prerequisites
#...missing this and that
./hack/e2e/ensure_e2e_prerequisites.sh
#...yes, I want to install docker...
#...need to log out/in for docker to work without sudo
./hack/ci-e2e.sh
#...tests running
```

terror96:
> > Pushed a new version even though there might still be issues. Wanted to give you something to look at...
> 
> I tested the procedure on a clean VM:
> 
> ```
> make verify-e2e-prerequisites
> #...missing this and that
> ./hack/e2e/ensure_e2e_prerequisites.sh
> #...yes, I want to install docker...
> #...need to log out/in for docker to work without sudo
> ./hack/ci-e2e.sh
> #...tests running
> ```

Also, `PATH=${PATH}:/usr/local/go/bin make test-e2e` works.

test/e2e/README.md
```diff
@@ -15,6 +15,24 @@ as BMC. Ironic runs in the "host network" of the kind cluster in the test. The
 kind cluster is then configured to expose the relevant ports on the actual host
 so that they can be reached from the BareMetalHost VMs.
 
+However, before running the script ensure the environment is setup correctly:
+
+```bash
+make verify-e2e-prerequisites
+```
+
+If any required components are missing, you can install them by running:
```
lentzi90: I think we should mention also that this script is for ubuntu (and maybe debian?) only.

terror96: Agreed. I think we could add a check to `hack/e2e/ensure_e2e_prerequisites.sh` to confirm that we are running under Ubuntu, because as you pointed out, it will fail on other systems. Support for other systems can be added later, because that will require changes in most of the `hack/e2e/ensure*` scripts anyway.

`hack/e2e/ensure_docker.sh` has this kind of a check built-in, but others might struggle a little bit more...

terror96: It would also make sense then to prepare `hack/e2e/ensure_e2e_prerequisites.sh` to support other tools to check and install than just dpkg/apt-get i.e. I'll refactor it a bit more.

hack/e2e/ensure_go.sh
```diff
@@ -1,34 +1,55 @@
...
-            export PATH=/usr/local/go/bin:$PATH
+            set -x
+            curl -sL \
+                -o "/tmp/${MINIMUM_GO_VERSION}.linux-amd64.tar.gz" \
+                "https://go.dev/dl/${MINIMUM_GO_VERSION}.linux-amd64.tar.gz"
+            sudo tar \
+                -C /usr/local \
+                -xzf "/tmp/${MINIMUM_GO_VERSION}.linux-amd64.tar.gz"
+            export PATH="${PATH}:/usr/local/go/bin"
+            GO="$(command -v go)"
```
terror96, line 30: We are missing `set +x` here. I'll add it...

terror96: Fixed.

hack/e2e/ensure_docker.sh
```diff
@@ -0,0 +1,137 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+# Verify mode turned off by default
+VERIFY_ONLY="${VERIFY_ONLY:-false}"
+
+# Source os-release to get ID variable
+# shellcheck source=/dev/null
+source /etc/os-release
```
tuminoid:
```suggestion
. /etc/os-release
```

hack/e2e/ensure_docker.sh
```diff
@@ -0,0 +1,137 @@
...
+        set +x
+    else
+        echo "Aborted."
+        return 1
+    fi
+}
+
+# Check if docker is installed and install it if not
+verify_docker()
+{
```
tuminoid:
```suggestion
verify_docker() {
```
Consistency, either this or fix the others.

hack/e2e/ensure_docker.sh
```diff
@@ -0,0 +1,137 @@
...
+    local DOCKER URL
+
+    DOCKER="$(command -v docker || true)"
+    if ! [[ -x "${DOCKER}" ]]; then
+        if [[ "${VERIFY_ONLY}" != "false" ]]; then
+            echo "docker is not installed or not in PATH"
```
tuminoid:
```suggestion
            echo "ERROR: docker is not in PATH"
```

hack/e2e/ensure_e2e_prerequisites.sh
```diff
@@ -0,0 +1,140 @@
...
+# -----------------------------------------------------------------------------
+
+set -eu
+
+REPO_ROOT="$(realpath "$(dirname "${BASH_SOURCE[0]}")/../..")"
+VERIFY_ONLY="${VERIFY_ONLY:-false}"
+
+# Source os-release to get ID variable
+# shellcheck source=/dev/null
+source /etc/os-release
```
tuminoid:
```suggestion
. /etc/os-release
```

hack/e2e/ensure_go.sh
```diff
@@ -1,34 +1,56 @@
...
+# Verify mode turned off by default
+VERIFY_ONLY="${VERIFY_ONLY:-false}"
+
 # Ensure the go tool exists and is a viable version, or installs it
 verify_go_version()
 {
     # If go is not available on the path, get it
-    if ! [ -x "$(command -v go)" ]; then
+    GO="$(command -v go || true)"
+    if ! [ -x "${GO}" ]; then
```
tuminoid:
```suggestion
    if ! [[ -x "${GO}" ]]; then
```

hack/e2e/ensure_go.sh
```diff
@@ -1,34 +1,56 @@
...
+set -eu
 
 MINIMUM_GO_VERSION=go1.25.7
 
+# Verify mode turned off by default
+VERIFY_ONLY="${VERIFY_ONLY:-false}"
+
 # Ensure the go tool exists and is a viable version, or installs it
 verify_go_version()
 {
```
tuminoid:
```suggestion
verify_go_version() {
```
Might as well clean this up.

hack/e2e/ensure_htpasswd.sh
```diff
@@ -1,14 +1,25 @@
...
 
-set -eux
+set -eu
+
+# Verify mode turned off by default
+VERIFY_ONLY="${VERIFY_ONLY:-false}"
 
 # Check if htpasswd is installed and install it if not
 verify_htpasswd()
 {
```
tuminoid:
```suggestion
verify_htpasswd() {
```
Same

hack/e2e/ensure_htpasswd.sh
```diff
@@ -1,14 +1,25 @@
...
     if ! [ -x "$(command -v htpasswd)" ]; then
```
tuminoid:
```suggestion
    if ! [[ -x "$(command -v htpasswd)" ]]; then
```
Same

hack/e2e/ensure_kubectl.sh
```diff
@@ -14,35 +14,56 @@
...
 USR_LOCAL_BIN="/usr/local/bin"
 MINIMUM_KUBECTL_VERSION=v1.34.1
+KUBECTL_DOWNLOAD_URL="https://dl.k8s.io/release"
+
+# Verify mode turned off by default
+VERIFY_ONLY="${VERIFY_ONLY:-false}"
 
 # Ensure the kubectl tool exists and is a viable version, or installs it
 verify_kubectl_version()
 {
```
tuminoid:
```suggestion
verify_kubectl_version() {
```
Same

hack/e2e/ensure_kubectl.sh
```diff
@@ -14,35 +14,56 @@
...
     # If kubectl is not available on the path, get it
-    if ! [ -x "$(command -v kubectl)" ]; then
+    KUBECTL="$(command -v kubectl || true)"
+    if ! [ -x "${KUBECTL}" ]; then
```
tuminoid:
```suggestion
    if ! [[ -x "${KUBECTL}" ]]; then
```

hack/e2e/ensure_kubectl.sh
```diff
@@ -14,35 +14,56 @@
...
+            set -x
+            curl -LO \
+                --create-dirs \
+                --output-dir "/tmp" \
+                "${KUBECTL_DOWNLOAD_URL}/${MINIMUM_KUBECTL_VERSION}/bin/linux/amd64/kubectl"
+            sudo install "/tmp/kubectl" "${USR_LOCAL_BIN}/kubectl"
+            set +x
+            KUBECTL="$(command -v kubectl)"
         else
             echo "Missing required binary in path: kubectl"
```
tuminoid:
```suggestion
            echo "ERROR: Missing required binary in path: kubectl"
```

tuminoid (commented):
Some nits.

hack/e2e/ensure_docker.sh
Rozzii: /etc/os-release doesn't even have export statements could you please explain why are you suggesting "."  instead of source?

hack/e2e/ensure_e2e_prerequisites.sh
Rozzii: same question as about  about the use of source

hack/e2e/ensure_docker.sh
terror96: I guess nowadays "source" is preferred over ".", but I'm happy to do either way as long as we all agree.

tuminoid: `source` is not POSIX-compliant, `.` is. There is no functional difference.

hack/e2e/ensure_docker.sh
terror96: Wondering if this is an "ERROR" or a "WARNING" because we are not returning a non-zero value...

tuminoid: This is probably a leftover, I rolled back many of these when I noticed its actually return 0, and not return 1.

hack/e2e/ensure_docker.sh
terror96: Let's go with `.` then...

hack/e2e/ensure_docker.sh
terror96: Okay, reverting...

terror96: But adding now `ERROR` to all messages when returning a non-zero value.

terror96: and using a shorter message e.g. `docker is not in PATH`

lentzi90:
As far as I can see, the hold was for me to review. I have done that and given lgtm, so unholding
/hold cancel

PR #2710, opened 2025-10-01
🌱Update github-script version to v8.0.0

smoshiur1237:
Update  github-script version from v.7.1.0 to v8.0.0


smoshiur1237:
Lets depandabot to do the changes.

tuminoid:
> Shouldn't this be updated by dependabot? Why do we need to do it manually?

Dependabot will do it on its next run, which is Sunday.

PR #2713, opened 2025-10-02
[PoC] Integrate with IrSO for getting Ironic details

dtantsur:
DANGER: the current version is generated by Claude and has not been yet
reviewed or edited by me. Do not merge.

Generated-By: Claude Code (commercial license)


pkg/provisioner/ironic/factory.go
```diff
@@ -196,3 +261,153 @@ func loadTLSConfigFromEnv() clients.TLSConfig {
...
+		}
+	}
+	if !ready {
+		return "", clients.AuthConfig{}, clients.TLSConfig{}, fmt.Errorf("ironic CR %s/%s is not ready", f.ironicNamespace, f.ironicName)
+	}
+
+	// Construct endpoint URL
+	apiPort := ironicCR.Spec.Networking.APIPort
+	if apiPort == 0 {
+		apiPort = 6385 // default port
```
dtantsur: This is probably not correct: the service endpoint uses the default HTTP(s) port.

pkg/provisioner/ironic/factory.go
```diff
@@ -196,3 +261,153 @@ func loadTLSConfigFromEnv() clients.TLSConfig {
...
+
+	err = f.k8sClient.Get(ctx, key, ironicCR)
+	if err != nil {
+		return "", clients.AuthConfig{}, clients.TLSConfig{}, fmt.Errorf("failed to get Ironic CR %s/%s: %w", f.ironicNamespace, f.ironicName, err)
+	}
+
+	// Check if the Ironic CR is ready
+	ready := false
+	for _, condition := range ironicCR.Status.Conditions {
+		if condition.Type == "Ready" && condition.Status == metav1.ConditionTrue {
```
dtantsur: Use a constant for "Ready"

pkg/provisioner/ironic/factory.go
```diff
@@ -196,3 +261,153 @@ func loadTLSConfigFromEnv() clients.TLSConfig {
...
+	if err != nil {
+		return "", clients.AuthConfig{}, clients.TLSConfig{}, err
+	}
+
+	return endpoint, auth, tlsConfig, nil
+}
+
+func (f *ironicProvisionerFactory) loadAuthConfigFromIronicCR(ctx context.Context, ironicCR *ironicv1alpha1.Ironic) (clients.AuthConfig, error) {
+	// If APICredentialsName is set, we need to read the secret for HTTP basic auth
+	if ironicCR.Spec.APICredentialsName != "" {
```
dtantsur: Cannot be empty

pkg/provisioner/ironic/factory.go
```diff
@@ -196,3 +261,153 @@ func loadTLSConfigFromEnv() clients.TLSConfig {
...
+			Namespace: ironicCR.Namespace,
+		}
+
+		err := f.k8sClient.Get(ctx, key, secret)
+		if err != nil {
+			return clients.TLSConfig{}, fmt.Errorf("failed to get TLS secret %s/%s: %w", ironicCR.Namespace, ironicCR.Spec.TLS.CertificateName, err)
+		}
+
+		// Write certificate data to temporary files
+		caCert, hasCACert := secret.Data["ca.crt"]
```
dtantsur: TODO: check that this is a reasonable assumption

pkg/provisioner/ironic/factory.go
```diff
@@ -196,3 +261,153 @@ func loadTLSConfigFromEnv() clients.TLSConfig {
...
+		clientCert, hasClientCert := secret.Data["tls.crt"]
+		clientKey, hasClientKey := secret.Data["tls.key"]
```
dtantsur: Definitely no need for the private key.

pkg/provisioner/ironic/factory.go
```diff
@@ -196,3 +261,153 @@ func loadTLSConfigFromEnv() clients.TLSConfig {
...
+
+		if hasCACert {
+			caFile, err := writeTempFile("ironic-ca-", caCert)
+			if err != nil {
+				return clients.TLSConfig{}, fmt.Errorf("failed to write CA certificate: %w", err)
+			}
+			tlsConfig.TrustedCAFile = caFile
+		}
+
+		if hasClientCert && hasClientKey {
```
dtantsur: Complete nonsense: these are not client certificates.

pkg/provisioner/ironic/factory.go
```diff
@@ -196,3 +261,153 @@ func loadTLSConfigFromEnv() clients.TLSConfig {
...
+
+func (f *ironicProvisionerFactory) loadConfigFromIronicCR(ctx context.Context) (endpoint string, auth clients.AuthConfig, tlsConfig clients.TLSConfig, err error) {
+	// Get the Ironic CR
+	ironicCR := &ironicv1alpha1.Ironic{}
+	key := types.NamespacedName{
+		Name:      f.ironicName,
+		Namespace: f.ironicNamespace,
+	}
+
+	err = f.k8sClient.Get(ctx, key, ironicCR)
```
dtantsur, line 284: TODO: check that we don't need to set a watch for this resource (Claude keeps ignoring me).

main.go
```diff
@@ -288,7 +290,16 @@ func main() {
...
 		provLog := zap.New(zap.UseFlagOptions(&logOpts)).WithName("provisioner")
-		provisionerFactory = ironic.NewProvisionerFactory(provLog, preprovImgEnable)
+		// Check if we should use Ironic CR integration
+		ironicName := os.Getenv("IRONIC_NAME")
+		ironicNamespace := os.Getenv("IRONIC_NAMESPACE")
+		if ironicName != "" && ironicNamespace != "" {
+			ctrl.Log.Info("using Ironic CR integration", "ironicName", ironicName, "ironicNamespace", ironicNamespace)
+			provisionerFactory = ironic.NewProvisionerFactoryWithClient(provLog, preprovImgEnable, mgr.GetClient())
+		} else {
+			ctrl.Log.Info("using traditional environment variable configuration")
```
dtantsur: Redundant logging here

dtantsur:
[FIXED] The unresolved question is now: how to cache secrets from the Ironic object in a sane way?

lentzi90 (commented):
I guess you have tested this manually? Should we switch the e2e tests to use this?
We don't even have to switch them all over, we could e.g. do it for one of the upgrade tests. What do you think?

@tuminoid do you see any potential security implications from this? Anything wrong with getting the credentials or certificates in this way?

/cc @Rozzii @tuminoid

dtantsur:
> I guess you have tested this manually?

Yep, all manually so far. I haven't updated tests because the e2e work happened in parallel to this one. It may not be trivial though because of how much Kustomize and deploy-cli are hardcoded everwhere.

dtantsur:
Rebased to get rid of the parent PR.

lentzi90 (commented):
I would still want to get @tuminoid to check this for potential security issues, but from my side this is good to go.
/approve

lentzi90:
Cross referencing here that with https://github.com/metal3-io/baremetal-operator/pull/2736 merged we now have the basics for publishing release manifests. Once this lands we have everything needed so we can clean up that configmap and make the release manifests really useful!

tuminoid:
> I would still want to get @tuminoid to check this for potential security issues, but from my side this is good to go. /approve

I'll take a look shortly, back from PTO.

config/base/rbac/role.yaml
```diff
@@ -25,6 +25,14 @@ rules:
...
   - update
   - watch
+- apiGroups:
+  - ironic.metal3.io
+  resources:
+  - ironics
+  verbs:
+  - get
+  - list
+  - watch
```
tuminoid, line 35: This should also be added to `config/overlays/e2e/roles-rolebindings/roles-rolebindings.yaml` for namespace scoped BMO testing and preferably also `docs/namespace-scoped-setup.md` example updated with this.

tuminoid (commented):
I think the code is fine, Secrets access is namespace scoped to the Ironic CR, so you cannot create a CR that'd point to arbitrary secrets and use BMO to read them for you (like in previous CVEs we had). 

But, this does not make the similar changes to the namespace scoped mode, which we are (should be at least) using in e2e.

config/base/rbac/role.yaml
dtantsur: This file is autogenerated, whoever added roles-rolebindings.yaml should ensure it's autogenerated too, no?

tuminoid: @Sunnatillo PTAL

Sunnatillo: @dtantsur I understand your point. But in current implementation of namespaces scoped mode, we are using roles and role bindings for each namespaces, instead of clusterRoles and ClusterRolebindings. 
When you add kubebuilder flags they generated automatically in clusterRole. But we use separate template for namespace scoped roles. 
Could you please update it there as well? 
`config/overlays/e2e/roles-rolebindings/roles-rolebindings.yaml
`
Our kustomize implementation for namespace scoped mode needs improvement. If you have ideas please share. Current state was the best I came up with.

dtantsur: Mmm, I still don't understand why this file cannot be just synchronized with role.yaml (maybe with a simple python script). I'll update it manually for now.

Rozzii (commented):
/approve

PR #2729, opened 2025-10-13
🌱 E2E: Use IrSO also in upgrade tests

lentzi90:

We are already deploying Ironic using IrSO in the other e2e tests. Let's do the same for the upgrade tests so we have one unified way of doing things.

This should hopefully make the optional tests green.
Hold to ensure we run those tests.
/hold


ironic-deployment/overlays/e2e/kustomization.yaml
lentzi90, line 1: I have left the e2e overlay here for now, because it is used by the e2e-local overlay and I thought it might be useful for development. However, I think we should remove it eventually.

lentzi90:
/cc @nuhakala 
The tests were looking promising until they got to Ironic 26.0. I noticed that you did not include this version in your PR before. Was there some specific reason for this? Is this version unsupported by IrSO?

lentzi90:
> The tests were looking promising until they got to Ironic 26.0. I noticed that you did not include this version in your PR before. Was there some specific reason for this? Is this version unsupported by IrSO?

Never mind, I noticed it failed because I had wrong name for the deployment...

lentzi90:
Only Ironic 26.0 failed before timeout in the last test, so now I am testing without that. Let's discuss if we can drop that test or it is still needed.

lentzi90:
We won't be able to use IrSO for Ironic 26.0 at least. There is no IrSO version that supports it.
https://github.com/metal3-io/ironic-standalone-operator/blob/729893768a782b462657ddb7c050eb6fd49b6998/api/v1alpha1/ironic_types.go#L45-L51

nuhakala:
> We won't be able to use IrSO for Ironic 26.0 at least. There is no IrSO version that supports it. https://github.com/metal3-io/ironic-standalone-operator/blob/729893768a782b462657ddb7c050eb6fd49b6998/api/v1alpha1/ironic_types.go#L45-L51

Yeah, there was a comment in my PR that support for 26.0 is either ended already or ending really soon, so that is why I did not include it.

test/e2e/config/ironic.yaml
```diff
@@ -101,62 +96,55 @@ bmoIronicUpgradeSpecs:
...
-  upgradeEntityKustomization: "../../ironic-deployment/overlays/e2e"
-  upgradeEntityName: "ironic"
 # Upgrade Ironic 27.0 -> latest | BMO 0.9
 - deployIronic: true
   deployBMO: true
   initBMOKustomization: "../../config/overlays/e2e-release-0.9"
-  initIronicKustomization: "../../ironic-deployment/overlays/e2e-release-27.0"
-  upgradeEntityKustomization: "../../ironic-deployment/overlays/e2e"
+  initIronicKustomization: "data/ironic-standalone-operator/ironic/overlays/e2e-release-27.0"
+  upgradeEntityKustomization: "data/ironic-standalone-operator/ironic/overlays/e2e"
```
nuhakala, line 121: I checked the supported versions here https://book.metal3.io/irso/introduction and here https://book.metal3.io/version_support

The irso version 0.6.0 is out and we maybe want to bump to that soonish. But that would mean that we have to drop versions 27-29 from these tests.

test/e2e/config/fixture.yaml
```diff
@@ -22,8 +22,9 @@ variables:
   DEPLOY_BMO: "true"
   DEPLOY_CERT_MANAGER: "true"
   BMO_KUSTOMIZATION: "../../config/overlays/fixture"
-  # This setting is for the separate cluster that are used for upgrade tests
+  # These setting are for the separate cluster that are used for upgrade tests
```
nuhakala: Grammar police here >:D, I think this should be "These settingS are".

test/e2e/config/ironic.yaml
```diff
@@ -28,11 +28,13 @@ variables:
   DEPLOY_IRONIC: "true"
   DEPLOY_BMO: "true"
   DEPLOY_CERT_MANAGER: "true"
+  IRSO_KUSTOMIZATION: "data/ironic-standalone-operator/operator"
   BMO_KUSTOMIZATION: "../../config/overlays/e2e"
-  IRONIC_KUSTOMIZATION: "../../ironic-deployment/overlays/e2e"
+  IRONIC_KUSTOMIZATION: "data/ironic-standalone-operator/ironic/overlays/e2e"
 
-  # This setting is for the separate cluster that are used for upgrade tests
+  # These setting are for the separate cluster that are used for upgrade tests
```
nuhakala: same nit here

test/e2e/config/ironic.yaml
lentzi90: Yeah we need to think about how to do this. At the moment I think we test a bit too much, but we should definitely still test older releases of BMO and that requires some older Ironic releases as well.

lentzi90:
Ok one more time... Let's see if we can get them all green

test/e2e/config/ironic.yaml
elfosardo: please remember that old ironic-image containers depend on branches that are not supported in ironic anymore, meaning that they will probably have bugs and unexpected behaviors
even if we decide to test them we may consider adding the tests as optional for PR acceptance

lentzi90:
Looks like we could get it all green with some adjusted deadlines.
For the suite timeout I think we may need a PR in project-infra. I'll check.

**redfish-virtualmedia:**
```
Summarizing 2 Failures:
   [FAIL] Upgrade [It] Should upgrade bmo from ../../config/overlays/e2e-release-0.10 to latest version [optional, upgrade]
   /home/metal3ci/workspace/metal3-bmo-e2e-test-optional-pull/test/e2e/common.go:711

   [FAIL] Upgrade [It] Should upgrade ironic from data/ironic-standalone-operator/ironic/overlays/e2e-release-28.0 to latest version [optional, upgrade]
   /home/metal3ci/workspace/metal3-bmo-e2e-test-optional-pull/test/e2e/common.go:711

Ran 8 of 20 Specs in 6427.414 seconds  
FAIL! -- 6 Passed | 2 Failed | 0 Pending | 12 Skipped  
--- FAIL: TestE2e (6427.42s)  
FAIL
```

Ironic 28.0 timed out waiting to be ready after 600s.
For BMO 0.10 it was also Ironic (latest) that timed out waiting to be ready after 600s.

**IPMI:**

```
Summarizing 1 Failure:
   [TIMEDOUT] Upgrade [It] Should upgrade ironic from data/ironic-standalone-operator/ironic/overlays/e2e-release-29.0 to latest version [optional, upgrade]
   /home/metal3ci/workspace/metal3-bmo-e2e-test-optional-pull/test/e2e/upgrade_test.go:356

Ran 6 of 20 Specs in 7217.998 seconds
FAIL! - Suite Timeout Elapsed -- 5 Passed | 1 Failed | 0 Pending | 14 Skipped
--- FAIL: TestE2e (7218.00s)
FAIL
```

lentzi90:
/hold cancel
Optional tests are green! 🎉

test/e2e/config/ironic.yaml
```diff
@@ -28,11 +28,13 @@ variables:
   DEPLOY_IRONIC: "true"
   DEPLOY_BMO: "true"
   DEPLOY_CERT_MANAGER: "true"
+  IRSO_KUSTOMIZATION: "data/ironic-standalone-operator/operator"
```
kashifest, line 31: So do I understand correctly that  this will deploy `IRSO operator` and the other variable two lines down will bring in the ironic itself? Are we then handling the IRSO operator upgrade in these tests as well?

lentzi90: Yes, we install IrSO first and then use that for deploying Ironic.
We don't upgrade IrSO. We can add that to the tests if we want, but I would rather do IrSO upgrade tests in the IrSO repo.

kashifest: Yep that makes sense, but I am thinking from the point of view of upgrading BMO, whether we should also upgrade IrSO alongwith, for now if I understand correctly IrSO version will remain static, in IrSO upgrade if/when we test upgrade does it also test BMO deployments ?

kashifest:
Seems, theres some conflicts now, I can approve it once those are resolved, lgtm from my side for now

test/e2e/config/ironic.yaml
lentzi90: Well, IrSO does not really affect BMO. In IrSO upgrades we should check that Ironic works the same with old and new version. I don't think we have those tests yet but it is something we should add.
I think it would be enough to test it separate from BMO. It would not be nice to "pull up" too much of these things to BMO or CAPM3.

PR #2736, opened 2025-10-16
🌱 Add release manifests

### LINKED ISSUE #1722: Publish release artifacts with manifests

lentzi90:

Generate release manifests for BMO and include as release artifacts.
I have followed closely how it is done for CAPM3.

Some things to note:
- Split out image tag from `IMG` variable to keep things similar to CAPM3.
- Add registry to `IMG` so we get a proper URL.
- Switch `set-manifest-image-bmo` to use `sed`. Kustomize is doing all kinds of formatting changes which can be very confusing. Note that I left the rest of the `set-manfest-image-*` targets as is. I think we should get rid of them when metal3-dev-env switches to IrSO. Metal3-dev-env is the only place where they are used as far as I know.

Fixes #1722

### DIFF
--- a/.github/workflows/e2e-fixture-test.yml
+++ b/.github/workflows/e2e-fixture-test.yml
@@ -29,7 +29,8 @@ jobs:
 
     - name: Build BMO e2e Docker Image
       env:
-        IMG: quay.io/metal3-io/baremetal-operator:e2e
+        IMG: quay.io/metal3-io/baremetal-operator
+        IMG_TAG: e2e
       run: make docker
 
     - name: Set Up Environment and Run BMO e2e Tests
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -108,7 +108,7 @@ jobs:
         go-version: ${{ env.go_version }}
     - name: generate release artifacts
       run: |
-        make release-notes
+        make release
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     - name: get release notes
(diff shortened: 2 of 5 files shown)


lentzi90:
/hold
I must have missed something with how we set the tag for e2e tests...

lentzi90:
/hold cancel
Looks like it is working now

PR #2740, opened 2025-10-17
🐛 Make bootMACAddress webhook validation case-insensitive

sakhoury:

This PR makes the `bootMACAddress` field validation case-insensitive in the BareMetalHost webhook. Previously, the webhook would reject updates to a BareMetalHost if the `bootMACAddress` was changed only in case (e.g., from `00:00:5e:00:53:af` to `00:00:5E:00:53:AF`), even though these represent the same MAC address.

The fix uses `net.ParseMAC()` to normalize MAC addresses before comparing them. This allows case variations of the same MAC address while still enforcing the immutability constraint to prevent actual MAC address changes.

**Changes:**
- Parse both old and new MAC addresses using `net.ParseMAC()`
- Compare the normalized string representations
- Maintain proper error handling for invalid MAC addresses
- Preserve existing validation that prevents actual MAC address changes
- Add validation to prevent unsetting the MAC address once set

### DIFF
--- a/internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
+++ b/internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
@@ -109,7 +109,7 @@ func (webhook *BareMetalHost) validateChanges(oldObj *metal3api.BareMetalHost, n
 		errs = append(errs, errors.New("BMC address can not be changed if the BMH is not in the Registering state, or if the BMH is not detached"))
 	}
 
-	if oldObj.Spec.BootMACAddress != "" && newObj.Spec.BootMACAddress != oldObj.Spec.BootMACAddress {
+	if oldObj.Spec.BootMACAddress != "" && !strings.EqualFold(newObj.Spec.BootMACAddress, oldObj.Spec.BootMACAddress) {
 		errs = append(errs, errors.New("bootMACAddress can not be changed once it is set"))
 	}
 
--- a/internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation_test.go
+++ b/internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation_test.go
@@ -1055,10 +1055,39 @@ func TestValidateUpdate(t *testing.T) {
 		{
 			name: "updateBootMAC",
 			newBMH: &metal3api.BareMetalHost{
-				TypeMeta: tm, ObjectMeta: om, Spec: metal3api.BareMetalHostSpec{BootMACAddress: "test-mac-changed"}},
+				TypeMeta: tm, ObjectMeta: om, Spec: metal3api.BareMetalHostSpec{BootMACAddress: "00:11:22:33:44:66"}},
 			oldBMH: &metal3api.BareMetalHost{
-				TypeMeta: tm, ObjectMeta: om, Spec: metal3api.BareMetalHostSpec{BootMACAddress: "test-mac"}},
+				TypeMeta: tm, ObjectMeta: om, Spec: metal3api.BareMetalHostSpec{BootMACAddress: "00:11:22:33:44:55"}},
 			wantedErr: "bootMACAddress can not be changed once it is set",
+		},
+		{
+			name: "updateBootMACCaseOnly",
+			newBMH: &metal3api.BareMetalHost{
+				TypeMeta: tm, ObjectMeta: om, Spec: metal3api.BareMetalHostSpec{BootMACAddress: "AA:BB:CC:DD:EE:FF"}},
+			oldBMH: &metal3api.BareMetalHost{
+				TypeMeta: tm, ObjectMeta: om, Spec: metal3api.BareMetalHostSpec{BootMACAddress: "aa:bb:cc:dd:ee:ff"}},
+			wantedErr: "",
+		},
+		{
+			name: "updateBootMACCaseMixed",
+			newBMH: &metal3api.BareMetalHost{
+				TypeMeta: tm, ObjectMeta: om, Spec: metal3api.BareMetalHostSpec{BootMACAddress: "aA:Bb:cC:Dd:Ee:Ff"}},
+			oldBMH: &metal3api.BareMetalHost{
+				TypeMeta: tm, ObjectMeta: om, Spec: metal3api.BareMetalHostSpec{BootMACAddress: "AA:BB:CC:DD:EE:FF"}},
+			wantedErr: "",
+		},
+		{
+			name: "updateBootMACInvalidNew",
+			newBMH: &metal3api.BareMetalHost{
+				TypeMeta: tm, ObjectMeta: om, Spec: metal3api.BareMetalHostSpec{
+					BMC: metal3api.BMCDetails{
+						Address:         "redfish://127.0.0.1",
+						CredentialsName: "test1",
+					},
+					BootMACAddress: "invalid-mac"}},
+			oldBMH: &metal3api.BareMetalHost{
+				TypeMeta: tm, ObjectMeta: om, Spec: metal3api.BareMetalHostSpec{BootMACAddress: "00:11:22:33:44:55"}},
+			wantedErr: "address invalid-mac: invalid MAC address",
 		},
 		{
 			name: "updateExternallyProvisioned",


internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation_test.go
```diff
@@ -1055,11 +1055,69 @@ func TestValidateUpdate(t *testing.T) {
...
+			newBMH: &metal3api.BareMetalHost{
+				TypeMeta: tm, ObjectMeta: om, Spec: metal3api.BareMetalHostSpec{
+					BMC: metal3api.BMCDetails{
+						Address:         "redfish://127.0.0.1",
+						CredentialsName: "test1",
+					},
+					BootMACAddress: "invalid-new"}},
+			oldBMH: &metal3api.BareMetalHost{
+				TypeMeta: tm, ObjectMeta: om, Spec: metal3api.BareMetalHostSpec{BootMACAddress: "invalid-old"}},
+			wantedErr: "cannot validate bootMACAddress immutability: existing value \"invalid-old\" is not a valid MAC address",
```
iurygregory: If the initial mac was invalid, wouldn't we raise an error based on immutability?
You also have ` // BootMACAddress is immutable once set - cannot be changed or cleared`, so we are allowing people to create BMH with invalid mac format? 🤔

sakhoury: Great question! This looks contradictory at first glance, but it's actually an edge case for "legacy" data.
To clarify:
- New resources: invalid MACs are rejected at CREATE time by `validateBMCAccess` ([lines 146-151](https://github.com/metal3-io/baremetal-operator/blob/main/internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go#L146-L151)). We do not allow new BMHs with invalid MAC formats.
- Legacy resources: Invalid old MACs can only exist in resources created before validation was added (or through rare scenarios like direct etcd manipulation during debugging).

I can remove these edge test cases if it creates confusion rather than adding value.

iurygregory: > Legacy resources: Invalid old MACs can only exist in resources created before validation was added (or through rare scenarios like direct etcd manipulation during debugging).

I really don't see such cases tbh... let's see what others think about it

lentzi90: Probably not something we would really come across, but in theory there could be old deployments from before the webhook validation. Anyway, I am fine with the test case.

internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -109,8 +109,26 @@ func (webhook *BareMetalHost) validateChanges(oldObj *metal3api.BareMetalHost, n
 		errs = append(errs, errors.New("BMC address can not be changed if the BMH is not in the Registering state, or if the BMH is not detached"))
 	}
 
-	if oldObj.Spec.BootMACAddress != "" && newObj.Spec.BootMACAddress != oldObj.Spec.BootMACAddress {
-		errs = append(errs, errors.New("bootMACAddress can not be changed once it is set"))
+	if oldObj.Spec.BootMACAddress != "" {
+		// BootMACAddress is immutable once set - cannot be changed or cleared
+		if newObj.Spec.BootMACAddress == "" {
+			errs = append(errs, errors.New("bootMACAddress can not be unset once it is set"))
```
lentzi90: I'm not so sure how necessary it is to handle this special case. We would catch it anyway when validating it below. But I guess this gives a nicer message to the users, so from that perspective it is good.

Rozzii: I am honestly not convinced we need this, I think the old message was clear enough. This field is immutable, I don't think CAPM3 should care about the nature of mutation other than that the change should be blocked.
I would leave the old if condition and  add a new else branch for parsing the new mac only making sure it has the proper format.

In other words  I agree with the format checking and parsing at the time of resource modification/creation via the webhook. I don't agree with retroactively checking the format of the old value as part of admission control designed to test new changes. I also don't think we have to write more elaborate error messages rather we should provide 1 error when the immutability is enforced and 1 error when we have issue with the format.

internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -109,8 +109,26 @@ func (webhook *BareMetalHost) validateChanges(oldObj *metal3api.BareMetalHost, n
...
+			// The new MAC is validated by validateBMCAccess (called via validateHost above).
+			if oldErr != nil {
+				errs = append(errs, fmt.Errorf("cannot validate bootMACAddress immutability: existing value %q is not a valid MAC address", oldObj.Spec.BootMACAddress))
+			}
+
+			// If both MACs are valid and different, enforce immutability.
+			if oldErr == nil && newErr == nil && oldMAC.String() != newMAC.String() {
+				errs = append(errs, errors.New("bootMACAddress can not be changed once it is set"))
+			}
+		}
```
Rozzii:
```suggestion
    if oldObj.Spec.BootMACAddress != "" && newObj.Spec.BootMACAddress != oldObj.Spec.BootMACAddress {
		errs = append(errs, errors.New("bootMACAddress can not be changed once it is set"))
	} else {
	    newMAC, newErr := net.ParseMAC(newObj.Spec.BootMACAddress)
	    if newErr != nil  {
				errs = append(errs, errors.New("bootMACAddress format is incorrect!"))
		}
	}
```

Rozzii: Something like this.

internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
lentzi90: This suggestions sounds good to me also!

internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
sakhoury: Thank you for the feedback @Rozzii! Let me address the concerns:

1. Case-sensitivity issue
The core problem with the old condition, and the suggested condition (`newObj.Spec.BootMACAddress != oldObj.Spec.BootMACAddress`), is that it is case-sensitive. Since MAC addresses are hexadecimal and case-insensitive per RFC networking standards, `aa:bb:cc:dd:ee:ff` and `AA:BB:CC:DD:EE:FF` represent the same address. Using `net.ParseMAC` provides canonical case-insensitive comparison, which is the primary intent of this PR.
Example of the issue:
```
Old: "aa:bb:cc:dd:ee:ff"
New: "AA:BB:CC:DD:EE:FF"
String comparison: Different (blocked incorrectly)
net.ParseMAC comparison: Same (correctly allowed)
```

2. Issues with suggested code

The suggested code has two main concerns:
a) Redundant validation
The else clause validates the new bootMACAddress format:
```go
} else {
    newMAC, newErr := net.ParseMAC(newObj.Spec.BootMACAddress)
    if newErr != nil {
        errs = append(errs, errors.New("bootMACAddress format is incorrect!"))
    }
}
```
This is redundant because `validateBMCAccess` already validates the new bootMACAddress format. This would create duplicate error messages for invalid formats.
  b) Logic ambiguity
The `else` clause runs when:
- old bootMACAddress is empty (initial set)
- or old bootMACAddress equals new bootMACAddress (no change)
- or any other non-matching case
This means format validation would run even when no change is being made, which is unnecessary.

Furthermore, the explicit empty check improves the user experience: while string comparison would catch attempts to clear the bootMACAddress, the explicit empty check provides clearer, more intentional error messaging:
```go
if newObj.Spec.BootMACAddress == "" {
    errs = append(errs, errors.New("bootMACAddress can not be unset once it is set"))
}
```
Error message comparison:
- without explicit check: `"bootMACAddress can not be changed once it is set"` (generic)
- with explicit check: `"bootMACAddress can not be unset once it is set"` (specific, clearly indicates clearing is prohibited)
Since `net.ParseMAC("")` returns `"invalid bootMACAddress address"`, having an explicit empty check clarifies that the prohibition is due to immutability policy, not format validation failure.

I understand your concern about retroactive validation of old data. The `oldErr` check is defensive programming for legacy data, but I'm happy to remove it if you prefer:
- current approach: blocks updates when old bootMACAddress is invalid (defensive)
- simplified approach: skips immutability check when old bootMACAddress is invalid (permissive)

Since old bootMACAddresses would have been validated at CREATE time by `validateBMCAccess`, invalid old data should only exist in pre-validation legacy resources. The simplified approach would allow "fixing" such resources.

Would you prefer I remove the `oldErr` check? The case-insensitive comparison and empty-string check are essential for correctness and clear error messaging, and should be retained. I'm happy to simplify by removing the `oldErr` check if that addresses your concerns about retroactive validation.

Rozzii: I think I much better understand your case now, and I think a lot of complexity is introduced here with using net.ParseMac.

Your primary objective is to make sure that it is accepted when the the old mac is changed to a new  mac what is essentially the same but with a different capitalization/case.

My objective is to have the simplest admission control and not do things like handle errors related to the old mac, what if we do this?:
  ```suggestion
  if oldObj.Spec.BootMACAddress != "" && strings.ToLower(newObj.Spec.BootMACAddress) != strings.ToLower(oldObj.Spec.BootMACAddress) {
		errs = append(errs, errors.New("bootMACAddress can not be changed once it is set"))
  }
```

sakhoury: I actually looked into using `strings.ToLower()` before, but went with `net.ParseMAC()` for semantic accuracy and consistency with the existing code. That said, if simplicity is the priority, `strings.ToLower()` is a reasonable option - I’ll update the PR accordingly.

Rozzii (commented):
Thank you for you contribution and the willingness to compromise @sakhoury.
/lgtm
/hold cancel

PR #2744, opened 2025-10-20
🌱 E2E: Improve logs and artifacts in upgrade tests

### LINKED ISSUE #2743: E2E upgrade tests missing logs and resources
We are quite good at collecting logs and manifests from the normal e2e tests. However, the [upgrade tests](https://github.com/metal3-io/baremetal-operator/actions/workflows/e2e-test-optional-periodic.yml) are in a terrible state.

The issue is that every single BMH in all upgrade tests is named `upgrade`, so it is impossible to distinguish between them. The manifests are also overwriting each other so we do not even get all of them.
To make things worse, the namespace is also always either `upgrade-bmo` or `upgrade-ironic` for all these tests.

We also have issues with the logs from BMO and Ironic. Here the issue is that we get all logs duplicated. Each test has its own folder but they all have all the logs so you still have to check the name of the pod to know what logs are relevant for a specific test case.

The first thing we should do is to put the manifests in the test-case folder so they don't overwrite each other.

lentzi90:

This commit is an attempt to improve log and artifact collection in upgrade tests by storing them in separate folders for each test case. That way we avoid clobbering logs when they have the same file name and it is easier to find the relevant files for a specific case.

Fixes #2743

### DIFF
--- a/test/e2e/upgrade_test.go
+++ b/test/e2e/upgrade_test.go
@@ -6,7 +6,6 @@ package e2e
 import (
 	"context"
 	"fmt"
-	"path"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


lentzi90:
/hold
We need to run optional tests on this since that is the logs that are relevant here.
/test ?

lentzi90:
Logs look ok. There is more to fix but with this we at least get all the logs and manifests in separate folders.
More issues detected:
- We get ridiculous amounts of logs, especially from ironic and IPA (through Ironic). It can be several hundred MB for one test.
- This gets even worse because we duplicate the logs, as described in the issue. Luckily compression deals with this easily.
- The duplication has two causes: 1) We reuse the cluster between tests, 2) There is no way to cancel the log watch in CAPI at the moment (and ironic/bmo deployments in each test get picked up by the log watcher since they have the same labels).

I am planning to tackle these other issues separately.

PR #2745, opened 2025-10-22
✨ Accept per-host pull secrets for external OCI registries

mabulgu:

Adds per-host registry authentication for oci:// provisioning images.
New optional field on BareMetalHost:

```
spec:
  image:
    url: oci://<registry>/<repo>/<artifact>:<tag|digest>
    ociAuthSecretName: <k8s-secret-name>  # optional; same namespace as the BMH
```

When set, the controller validates the Kubernetes Docker-config secret (kubernetes.io/dockerconfigjson or kubernetes.io/dockercfg), selects the correct registry entry (supports exact host and host:port), and uses those credentials during provisioning so private OCI artefacts can be fetched on a per-host basis. Public images continue to work without credentials.
This removes a blocker for users who need different registries/accounts per machine.

Assisted-By: Claude-4.5-sonnet, Claude-opus-4.6


apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -881,6 +887,14 @@ type BareMetalHostStatus struct {
...
 	// +kubebuilder:default:=0
 	ErrorCount int `json:"errorCount"`
+
+	// Conditions provide observations of the operational state of a BareMetalHost.
+	// +optional
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	// +listType=map
+	// +listMapKey=type
+	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
```
dtantsur: While it's not a bad idea to have conditions for BMH, I suggest we don't conflate it with this patch. It will need a separate discussion.

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -2232,6 +2248,105 @@ func (r *BareMetalHostReconciler) getBMCSecretAndSetOwner(ctx context.Context, r
...
+	// If validation failed, return the error message
+	if !result.Valid {
+		return "", fmt.Errorf("auth secret validation failed: %s", result.Message)
+	}
+
+	// Return the extracted credentials
+	return result.Credentials, nil
+}
+
+// setImageAuthConditions sets the ImageAuthValid and ImageAuthInUse conditions on the host.
```
dtantsur: Conditions are used to communicate the state to 3rd parties, I don't quite understand why you need them within the same controller.

pkg/secretutils/validator.go
```diff
@@ -0,0 +1,143 @@
...
+type validator struct {
+	c        client.Client
+	recorder record.EventRecorder
+}
+
+func New(c client.Client, recorder record.EventRecorder) Validator {
+	return &validator{c: c, recorder: recorder}
+}
+
+func (v *validator) Validate(ctx context.Context, bmh *metal3api.BareMetalHost) (*Result, error) {
```
dtantsur: This one function does not really justify a new package, it can just as well leave under secretutils.

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -2317,14 +2432,38 @@ func (r *BareMetalHostReconciler) updateEventHandler(e event.UpdateEvent) bool {
 
 // SetupWithManager registers the reconciler to be run by the manager.
 func (r *BareMetalHostReconciler) SetupWithManager(mgr ctrl.Manager, preprovImgEnable bool, maxConcurrentReconcile int) error {
+	r.Recorder = mgr.GetEventRecorderFor("baremetalhost-controller")
+
+	// Add index for image auth secret name to enable efficient secret-to-BMH lookups
```
dtantsur: Please check secretutils for a prior art on handling secrets in BMO

pkg/secretutils/dockerconfig.go
```diff
@@ -0,0 +1,216 @@
...
+	Username string `json:"username,omitempty"`
+	Password string `json:"password,omitempty"`
+	Auth     string `json:"auth,omitempty"`
+	Email    string `json:"email,omitempty"`
+}
+
+// DockerConfig represents the structure of a kubernetes.io/dockercfg secret (legacy format).
+type DockerConfig map[string]DockerAuthConfig
+
+// ExtractRegistryCredentials extracts the registry credentials from a Kubernetes secret
```
dtantsur, line 15: Gemini told me there is a 3rd party library for this: https://github.com/cpuguy83/dockercfg

mabulgu: I saw that but I wasn't sure directly using that one as it doesnt seem to be a popular one. In the project, do we generally prefer less code and depending on the 3rd parties here, or more control and easier maintainability on our end? (genuinely asking, no sarcasm included:) )

mabulgu: Used by many projects tho. I only looked at the stars TBH. if we are ok to use this I can switch to this. no need to reinvent the tire. we should know that this comes with a cost tho. classic tradeoffs

MahnoorAsghar:
It might be too much code to be squashing into one commit...what do we think?

tuminoid:
> It might be too much code to be squashing into one commit...what do we think?

10 is definitely not right either.  One for implementation, one for tesrts, and one for docs? There is plenty of work left though, given the conditions need work per Dmitry's review.

mabulgu:
Thanks for your comments!

@MahnoorAsghar  > It might be too much code to be squashing into one commit...what do we think?

As soon as we itemize them and they are in the same context, I don't think so. I am going to remove everything related to conditions as they were like extra, but everything else share the same context and can be in the same squash commit IMO. 

@tuminoid  > One for implementation, one for tesrts, and one for docs

+1 for docs -1 for tests as tests are a part of the "implementation". Without tests, I would not count it as implemented. 

What I will do is: seperating the code commit (which will have less changes than the current changes because of the condition revert) and the commit for docs.

tuminoid:
> Thanks for your comments!
> @tuminoid > One for implementation, one for tesrts, and one for docs
> 
> +1 for docs -1 for tests as tests are a part of the "implementation". Without tests, I would not count it as implemented.

This is fine as well, but its quite common to implement tests in separate commit in same PR. Makes it maybe easier to manage, but like said, I'm 100% fine with code+tests in same commit.

apis/metal3.io/v1alpha1/baremetalhost_types.go
mabulgu: I will remove the conditions part to make this PR more clear and focused on the issue itself.

pkg/secretutils/dockerconfig.go
dtantsur: I cannot speak for the whole project, but I'd definitely avoid having more code to review and then maintain. Especially the code outside of my core expertise.

internal/controller/metal3.io/baremetalhost_controller.go
mabulgu: Done in secretutils/image_auth_validator.go

mabulgu:
@dtantsur I applied your suggestions. pls check when you have time. You will find the relevant commetns resolved but pls feel free to reopen them if you feel any of them are not implemented the way you suggested

mabulgu:
Not sure if the e2e test filure is related to my changes as it seems to be related to the BMC management credentials

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -2331,14 +2380,41 @@ func (r *BareMetalHostReconciler) updateEventHandler(e event.UpdateEvent) bool {
...
 	controller := ctrl.NewControllerManagedBy(mgr).
 		For(&metal3api.BareMetalHost{}).
 		WithEventFilter(
 			predicate.Funcs{
 				UpdateFunc: r.updateEventHandler,
 			}).
 		WithOptions(controller.Options{MaxConcurrentReconciles: maxConcurrentReconcile}).
-		Owns(&corev1.Secret{}, builder.MatchEveryOwner)
+		Owns(&corev1.Secret{}, builder.MatchEveryOwner).
+		// Watch Secrets to trigger reconciliation when auth secrets are updated (key rotation)
```
dtantsur: I don't think we need this. The auth secret is only used at a certain point during deployment. If you trigger a reconciliation at any other point, BMO won't be able to change anything. As long we watch these secrets, we're fine.

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -2349,6 +2425,38 @@ func (r *BareMetalHostReconciler) SetupWithManager(mgr ctrl.Manager, preprovImgE
 	return controller.Complete(r)
 }
 
+// findBMHsForAuthSecret returns a list of reconcile requests for BareMetalHosts
+// that reference the given Secret as their image auth secret.
+// This enables auto-reconciliation when auth secrets are updated (key rotation).
+func (r *BareMetalHostReconciler) findBMHsForAuthSecret(ctx context.Context, secret client.Object) []reconcile.Request {
```
dtantsur: See above

pkg/secretutils/dockerconfig.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// extractRegistryHost extracts the registry hostname from an OCI image URL.
+// For example, "oci://registry.example.com/repo/image:tag" returns "registry.example.com".
+func extractRegistryHost(imageURL string) (string, error) {
+	if !strings.HasPrefix(imageURL, "oci://") {
+		return "", fmt.Errorf("image URL does not have oci:// scheme: %s", imageURL)
+	}
+
+	// Remove the oci:// prefix
```
dtantsur: Maybe use `url.Parse` instead?

pkg/secretutils/image_auth_validator.go
```diff
@@ -0,0 +1,105 @@
...
+type Result struct {
+	Secret *corev1.Secret
+	Valid  bool
+	// Credentials contains the base64-encoded credentials in the format
+	// expected by Ironic (base64-encoded "username:password").
+	// This is only populated if Valid is true for OCI images.
+	Credentials string
+}
+
+type Validator interface {
```
dtantsur: I'm curious why you need an interface. Do you expect more implementations?

internal/controller/metal3.io/image_auth_validator.go
dtantsur, line 1: While this file is somewhat related to secrets, since its main purpose is to publish events, I'd keep it under internal/controller/metal3.io

pkg/secretutils/image_auth_validator.go
```diff
@@ -0,0 +1,105 @@
...
+		}
+		res.Credentials = credentials
+	}
+
+	res.Secret = sec
+	res.Valid = true
+	return res, nil
+}
+
+func isOCI(url string) bool {
```
dtantsur: If you make this a method on Image, you can probably reuse it in more places

pkg/secretutils/image_auth_validator.go
mabulgu: edit: oh yeah we dont need it right now as we already call it image auth validator. right!

internal/controller/metal3.io/image_auth_validator.go
mabulgu: makes sense

internal/controller/metal3.io/baremetalhost_controller.go
mabulgu: Yeah I should have removed it as I removed the conditions before.

MahnoorAsghar:
Other than the small CI linting failure, Dmitry's comments, and the many commits (I think this should be <5 commits), looks good to me

mabulgu:
> Other than the small CI linting failure, Dmitry's comments, and the many commits (I think this should be <5 commits), looks good to me

hey @MahnoorAsghar! yes as we discussed before this will be squashed before the merge for a better commit history. That will be the last thing to do as some update requests might occur and I don't want to restructure the commit structure each time. I will do it once I am sure everything is addressed. thanks for your comment!

internal/controller/metal3.io/image_auth_validator_test.go
```diff
@@ -0,0 +1,555 @@
...
+}
+
+func TestValidate_NoAuthSecret(t *testing.T) {
+	scheme := runtime.NewScheme()
+	_ = metal3api.AddToScheme(scheme)
+	_ = corev1.AddToScheme(scheme)
+
+	c := fake.NewClientBuilder().WithScheme(scheme).Build()
+	recorder := record.NewFakeRecorder(10)
+	secretManager := secretutils.NewSecretManager(t.Context(), testLogger(t), c, c)
```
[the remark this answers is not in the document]
tuminoid: LOL bot being in 2024.

mabulgu:
@tuminoid thanks for the feedback via copilot. 

`t.Context()` is valid in Go 1.24+. So no code changes are needed for this feedback. Could you check and approve accordingly?

tuminoid:
> @tuminoid thanks for the feedback via copilot.
> 
> `t.Context()` is valid in Go 1.24+. So no code changes are needed for this feedback. Could you check and approve accordingly?

Just noting that not everything Copilot complains about are valid, so use of human expertese should be used to consider if they're real or not. It is just nice way of quickly find inconsistencies and initial feedback. I'll run Copilot again for the latest changes.

I'll be taking another look when time permits, thanks for your patience.

internal/controller/metal3.io/image_auth_validator_test.go
```diff
@@ -0,0 +1,543 @@
...
```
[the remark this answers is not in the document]
tuminoid: This is invalid.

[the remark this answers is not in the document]
tuminoid: Invalid

internal/controller/metal3.io/image_auth_validator_test.go
```diff
@@ -0,0 +1,543 @@
...
+}
+
+func TestValidate_SecretNotFound(t *testing.T) {
+	scheme := runtime.NewScheme()
+	_ = metal3api.AddToScheme(scheme)
+	_ = corev1.AddToScheme(scheme)
+
+	c := fake.NewClientBuilder().WithScheme(scheme).Build()
+	recorder := record.NewFakeRecorder(10)
+	secretManager := secretutils.NewSecretManager(t.Context(), testLogger(t), c, c)
```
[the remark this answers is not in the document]
tuminoid: Invalid

internal/controller/metal3.io/image_auth_validator_test.go
```diff
@@ -0,0 +1,543 @@
...
+		Type: corev1.SecretTypeOpaque, // Wrong type
+		Data: map[string][]byte{
+			"username": []byte("user"),
+			"password": []byte("pass"),
+		},
+	}
+
+	c := fake.NewClientBuilder().WithScheme(scheme).WithObjects(secret).Build()
+	recorder := record.NewFakeRecorder(10)
+	secretManager := secretutils.NewSecretManager(t.Context(), testLogger(t), c, c)
```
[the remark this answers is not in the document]
tuminoid: Invalid

internal/controller/metal3.io/image_auth_validator_test.go
```diff
@@ -0,0 +1,543 @@
...
+}
+
+func TestValidate_NilImage(t *testing.T) {
+	scheme := runtime.NewScheme()
+	_ = metal3api.AddToScheme(scheme)
+	_ = corev1.AddToScheme(scheme)
+
+	c := fake.NewClientBuilder().WithScheme(scheme).Build()
+	recorder := record.NewFakeRecorder(10)
+	secretManager := secretutils.NewSecretManager(t.Context(), testLogger(t), c, c)
```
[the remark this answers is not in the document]
tuminoid: invalid

internal/controller/metal3.io/image_auth_validator_test.go
```diff
@@ -0,0 +1,543 @@
...
+		t,
+		corev1.SecretTypeDockerConfigJson,
+		map[string][]byte{
+			corev1.DockerConfigJsonKey: dockerConfigJSON,
+		},
+		"oci://quay.io/metal3-io/ironic:latest",
+	)
+
+	recorder := record.NewFakeRecorder(10)
+	secretManager := secretutils.NewSecretManager(t.Context(), testLogger(t), c, c)
```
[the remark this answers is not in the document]
tuminoid: invalid

tuminoid (commented):
OK, took a quick look for any major blockers, don't see any.

I've invalidated the context whining from Copilot, they're invaild in go 1.24 as said.

For testing part, I think we have minor gap:
- Consider unit tests for kubernetes.io/dockercfg (legacy) secret format in pkg/secretutils/dockerconfig_test.go.
- Consider test case in internal/controller/metal3.io/baremetalhost_controller_test.go ensuring getImageAuthSecret is called correctly.

I'm also fine to be proven wrong about the test coverage gaps, if I've misunderstood.

When done with final changes, just make sure the fixup commits are properly squashed into the 3 main ones you have.

Ping others as well for more eyes and opinions:
/cc @elfosardo @dtantsur @lentzi90 @Rozzii

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -634,14 +634,24 @@ type Image struct {
 	// be live-booted and not deployed to disk.
 	// +kubebuilder:validation:Enum=raw;qcow2;vdi;vmdk;live-iso
 	DiskFormat *string `json:"format,omitempty"`
+
+	// AuthSecretName optionally names a Docker-config secret containing
+	// registry credentials for oci:// images. Must be in the same namespace
+	// as the BareMetalHost. Allowed types: kubernetes.io/dockerconfigjson|dockercfg.
+	// Only used when Image.URL has the oci:// scheme.
+	AuthSecretName *string `json:"authSecretName,omitempty"`
```
Rozzii: Can you please name this some other way so that it would be self explanatory that this is used for OCI registry authentication, maybe OCIAuthSecretName ?

internal/controller/metal3.io/image_auth_validator.go
```diff
@@ -0,0 +1,102 @@
...
+func NewImageAuthValidator(recorder record.EventRecorder) *ImageAuthValidator {
+	return &ImageAuthValidator{recorder: recorder}
+}
+
+// Validate validates the image authentication secret for the given BMH.
+func (v *ImageAuthValidator) Validate(_ context.Context, bmh *metal3api.BareMetalHost, secretMgr secretutils.SecretManager) (*ImageAuthResult, error) {
+	res := &ImageAuthResult{Valid: false}
+
+	img := bmh.Spec.Image
+	if img == nil || img.URL == "" {
```
Rozzii: isn't this img == nil redundant? For me it looks like this is already checked above at https://github.com/metal3-io/baremetal-operator/pull/2745/changes#diff-b8822d3cb639bc85513ca4e296888679cf7c95dc1425427da4659481a50bda15R2379

internal/controller/metal3.io/image_auth_validator.go
```diff
@@ -0,0 +1,102 @@
...
+		return res, nil
+	}
+
+	ociRelevant := img.IsOCI()
+
+	// No per-host secret referenced → not required, valid for public images
+	if img.AuthSecretName == nil || *img.AuthSecretName == "" {
```
Rozzii: Similar question as above wasn't this tested in getImageAuthSecret ?

Rozzii (commented):
In general great improvement and very well tested.
I mainly paid attention to the new logic, just skimmed over the tests.
I have a few nits but in general looks very nice.

I know I have a naming related comment and naming things is hard but I think in this case the name really could be a bit more specific.

Also please rebase the PR. 
I have also triggered Co-Pilote reviews too because this change is quite large 1400+ lines.

docs/api.md
```diff
@@ -23,6 +23,43 @@ Several conditions must be met in order to initiate provisioning.
 
 To initiate deprovisioning, clear the image URL from the host spec.
 
+## OCI Image Authentication
```
dtantsur: This file is not really for user-visible docs any more, https://book.metal3.io/bmo/introduction.html is (source https://github.com/metal3-io/metal3-docs/tree/main/docs/user-guide/src/bmo)

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -2367,6 +2377,40 @@ func (r *BareMetalHostReconciler) getBMCSecretAndSetOwner(ctx context.Context, r
...
+		return "", nil
+	}
+
+	// Check for per-host auth secret
+	if image.OCIAuthSecretName == nil || *image.OCIAuthSecretName == "" {
+		return "", nil
+	}
+
+	// Use SecretManager following the BMC credentials pattern
+	reqLogger := r.Log.WithValues("baremetalhost", request.NamespacedName)
```
dtantsur: I think `r.Log` already has the BMH context

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -2367,6 +2377,40 @@ func (r *BareMetalHostReconciler) getBMCSecretAndSetOwner(ctx context.Context, r
...
+	// Validate and extract credentials
+	validator := NewImageAuthValidator(r.Recorder)
+	result, err := validator.Validate(ctx, host, secretManager)
+	if err != nil {
+		return "", fmt.Errorf("failed to validate auth secret: %w", err)
+	}
+
+	// If validation failed, return empty (validator already emitted events)
+	if !result.Valid {
+		return "", nil
```
dtantsur: In this case, the reconciler will continue with the host, ignoring the user-provided secret. I don't think it's the right thing to do, most likely it will cause a confusing failure during deployment.

internal/controller/metal3.io/image_auth_validator.go
```diff
@@ -0,0 +1,95 @@
...
+func (v *ImageAuthValidator) Validate(ctx context.Context, bmh *metal3api.BareMetalHost, secretMgr secretutils.SecretManager) (*ImageAuthResult, error) {
+	res := &ImageAuthResult{Valid: false}
+
+	img := bmh.Spec.Image
+	ociRelevant := img.IsOCI()
+	secretName := *img.OCIAuthSecretName
+
+	// Warn if secret is set for non-OCI image (future-proof, do not fail)
+	if !ociRelevant && v.recorder != nil {
+		v.recorder.Eventf(bmh, corev1.EventTypeWarning, EventAuthSecretIrrelevant,
```
dtantsur: Side note: we already have a different pattern to publish events in the controller, we need to reconcile them eventually..

dtantsur:
@mabulgu the linter failure looks real from a quick glance + a couple of minor comments

internal/controller/metal3.io/image_auth_validator.go
mabulgu: Agreed, that's worth a follow-up to unify the event publishing patterns across the controller.

mabulgu:
> @mabulgu the linter failure looks real from a quick glance + a couple of minor comments

Thanks @dtantsur. I addressed the feedback + commented for one. Pls check when available.

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -1315,13 +1317,21 @@ func (r *BareMetalHostReconciler) actionProvisioning(ctx context.Context, prov p
 		image = *info.host.Spec.Image.DeepCopy()
 	}
 
+	// Extract OCI auth secret credentials if needed
+	authSecret, err := r.getImageAuthSecret(ctx, info.host, &image)
+	if err != nil {
+		return recordActionFailure(info, metal3api.ProvisioningError,
+			"failed to get image auth secret: "+err.Error())
```
dtantsur: nit: the error messages that getImageAuthSecret produces already contain enough context, you don't need to add more here

internal/controller/metal3.io/image_auth_validator.go
```diff
@@ -0,0 +1,95 @@
...
+			if v.recorder != nil {
+				v.recorder.Eventf(bmh, corev1.EventTypeWarning, EventAuthParseError,
+					"Failed to extract credentials from secret %q: %v", secretName, err)
+			}
+			return res, nil
+		}
+		res.Credentials = credentials
+	}
+
+	res.Secret = sec
```
dtantsur: It looks like Secret is actually unused? In this case, you can simplify this function to return `(credentials, error)`. After all, that's what getImageAuthSecret returns anyway.

internal/controller/metal3.io/image_auth_validator.go
```diff
@@ -0,0 +1,95 @@
...
+	res.Valid = true
+	return res, nil
+}
+
+func isAllowedDockerConfigType(t corev1.SecretType) bool {
```
dtantsur: nit: inline this

tuminoid:
This needs rebase due linter updates.
/hold

mabulgu:
@tuminoid @dtantsur I will need an approval/lgtm for merging this along.

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/metal3-io/baremetal-operator
 go 1.25.0
 
 require (
+	github.com/cpuguy83/dockercfg v0.3.2
```
tuminoid, line 6: This has not been maintained for 2 years, and it is basically not being used by anybody per godoc.  Not exactly fond of adding such dependencies.

tuminoid (commented):
Please check Copilot nits.

I don't have other blockers, except the importing of basically unmaintained and not widely used dockercfg.

go.mod
elfosardo: agree on this

dtantsur: I asked for this because I'm not fond of reviewing a copy of that code.

Also, why is it a concern? We still use https://github.com/pkg/errors that hasn't updated for 6 years. Are we going to maintain a copy of that code any better?

tuminoid: I've actually been rewriting pkg/errors away ([capm3](https://github.com/metal3-io/cluster-api-provider-metal3/pull/3059), [ipam](https://github.com/metal3-io/ip-address-manager/pull/1258)) as it provides little value these days over stdlib.

CRA is not directly a Metal3 issue but since you work at RH, you've probably heard enough about effects of CRA, and if not maybe you, but it has made me quite wary of adding "random" deps, especially when they appear unmaintained. From consumer pov, there is also a difference of using something that is used by 360000 packages versus 15. One five. It does not exactly sound like the dep is either very useful or that there might be better alternatives. 

That said, I did review the code and it seems to be fine, so unless a maintained replacement is available, I don't block using it versus copying the code.

dtantsur: If we do decide to import it in-tree, let's import it more or less cleanly in a separate pkg/ subdirectory. Even with 15 users, I'd be somewhat more confident than in a from-scratch rewrite.

tuminoid: We can leave it as is for now, and follow up with optimizations later if we feel like its necessary.

mabulgu: ok leaving it as is and keeping the lib usage.

my 2 cents: Somehow nowadays it makes more sense to use a non-lib approach rather than a lib but also in terms of maintenance it looks like the same affordwise as if we need a change/update per a bug we will either go with our own custom release (if not maintained) or our custom change on our own code.

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -2407,6 +2416,23 @@ func (r *BareMetalHostReconciler) getBMCSecretAndSetOwner(ctx context.Context, r
 	return bmcCredsSecret, nil
 }
 
+// getImageAuthSecret validates and extracts the OCI registry credentials for the image.
+// It returns the base64-encoded credentials in the format expected by Ironic, or an empty
+// string if no auth secret is configured.
+func (r *BareMetalHostReconciler) getImageAuthSecret(ctx context.Context, host *metal3api.BareMetalHost, image *metal3api.Image) (string, error) {
+	if image == nil || !image.IsOCI() {
```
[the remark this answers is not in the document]
mabulgu: This is by design. When the image URL is not OCI, the ociAuthSecretName field has no effect on provisioning, so there is no point in fetching or validating the secret. The early return in getImageAuthSecret avoids unnecessary API calls and potential errors from missing or misconfigured secrets that are irrelevant anyway. The warning event in Validate serves as a safety net if the code path ever changes, but currently getImageAuthSecret is the correct gate.

mabulgu:
@honza can I get a re- `/lgtm` when you can, as I had to rebase this + got some copilot feedback via @tuminoid

internal/controller/metal3.io/image_auth_validator.go
```diff
@@ -0,0 +1,77 @@
...
+// Validate validates the image authentication secret for the given BMH and
+// returns the base64-encoded credentials in the format expected by Ironic.
+// Callers must ensure bmh.Spec.Image and OCIAuthSecretName are non-nil
+// (getImageAuthSecret already performs these checks).
+func (v *ImageAuthValidator) Validate(ctx context.Context, bmh *metal3api.BareMetalHost, secretMgr secretutils.SecretManager) (string, error) {
+	img := bmh.Spec.Image
+	ociRelevant := img.IsOCI()
+	secretName := *img.OCIAuthSecretName
+
+	if !ociRelevant {
```
tuminoid: In the way its implemented, this can never fail, since getImageAuthSecret returns early if its not true there.

internal/controller/metal3.io/image_auth_validator.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+// Validate validates the image authentication secret for the given BMH and
+// returns the base64-encoded credentials in the format expected by Ironic.
+// Callers must ensure bmh.Spec.Image and OCIAuthSecretName are non-nil
+// (getImageAuthSecret already performs these checks).
+func (v *ImageAuthValidator) Validate(ctx context.Context, bmh *metal3api.BareMetalHost, secretMgr secretutils.SecretManager) (string, error) {
+	img := bmh.Spec.Image
+	ociRelevant := img.IsOCI()
+	secretName := *img.OCIAuthSecretName
```
tuminoid: Regardless of the comment above, we should guard against nil pointers.

pkg/provisioner/ironic/ironic.go
```diff
@@ -616,6 +616,13 @@ func (p *ironicProvisioner) setDirectDeployUpdateOptsForNode(ironicNode *nodes.N
 		"image_disk_format": imageData.DiskFormat,
 	}
 
+	// Add image pull secret for OCI images
+	if imagePullSecret != "" {
+		optValues["image_pull_secret"] = imagePullSecret
+	} else {
+		optValues["image_pull_secret"] = nil
```
tuminoid, line 623: Maybe a comment why this is needed?

pkg/provisioner/ironic/ironic.go
```diff
@@ -672,6 +679,12 @@ func (p *ironicProvisioner) setCustomDeployUpdateOptsForNode(ironicNode *nodes.N
 				"image_disk_format":   imageData.DiskFormat,
 			}
 		}
+		// Add image pull secret for OCI images
+		if imagePullSecret != "" {
+			optValues["image_pull_secret"] = imagePullSecret
+		} else {
+			optValues["image_pull_secret"] = nil
```
tuminoid, line 686: Same here

pkg/provisioner/ironic/provision_test.go
```diff
@@ -1607,6 +1607,99 @@ func TestGetUpdateOptsForNodeSecureBoot(t *testing.T) {
...
+			if update.Path != e.Path {
+				t.Errorf("did not find %q in updates", e.Path)
+				return
+			}
+			assert.Equal(t, e.Value, update.Value, "%s does not match", e.Path)
+		})
+	}
+}
+
+func TestGetUpdateOptsForNodeOCIWithoutPullSecret(t *testing.T) {
```
tuminoid, line 1668: Missing case for node had image_pull_secret and it is cleared?

mabulgu: added. thanks for the headsup!

mabulgu:
@tuminoid thanks for the good eye and comments (again). Pls check and let me know.

tuminoid:
> @tuminoid thanks for the good eye and comments (again). Pls check and let me know.

Fix the copilot's latest findings, and lets then lgtm and unhold. Thanks for the patience!

mabulgu:
Hi @tuminoid 

> Fix the copilot's latest findings

Please use at least "please" or similar, as we are not AI agents (yet) right? :)

I will check out copilot's comments and apply anything worth applying/fixing.

> Thanks for the patience!

NP! As Captain America says: "I can do this all day" but other stakeholders won't let me as this should be merged one way or another:)

Thanks!

tuminoid:
> Please use at least "please" or similar, as we are not AI agents (yet) right? :)

My bad, sorry! Not intention to be rude, I'm just Finnish :) 

> NP! As Captain America says: "I can do this all day" but other stakeholders won't let me as this should be merged one way or another:)

I just held BMO 0.13 RC creation so we can get this (and maybe some others) feature in before it branches.

mabulgu:
> > Please use at least "please" or similar, as we are not AI agents (yet) right? :)
> 
> My bad, sorry! Not intention to be rude, I'm just Finnish :)
> 

Lol. Well, I have many Finnish friends who can say "please"; I verified with them:) j/k

> > NP! As Captain America says: "I can do this all day" but other stakeholders won't let me as this should be merged one way or another:)
> 
> I just held BMO 0.13 RC creation so we can get this (and maybe some others) feature in before it branches.

ai ai captain! on its way!

tuminoid (commented):
Let's get this merged to fit into v0.13 release. 

/lgtm
/unhold

Thank you for the contribution!

mabulgu:
thanks for your reviews and help @tuminoid !

Issue #1731, opened 2024-05-19
Add LLDP collection in bmh


shibaPuppy:
**User Story**
since ironic allows you to configure LLDP collection, you should similarly be able to enter LLDP values ​​in bmh status.
see your network topology, understand the physical connections between devices, and troubleshoot more efficiently.

**Detailed Description**
since LLDP information is collected at the ironic inspection stage, I think it should be collected at the same timing as HardwareData.

**Anything else you would like to add:**
https://github.com/metal3-io/baremetal-operator/issues/1227

If a new CR (ex. network discovery) is created, will a new controller be needed?
when I looked at the code again recently, I thought that going similar to [PR](https://github.com/metal3-io/baremetal-operator/pull/1167) would be the easiest direction without any changes.

/kind feature

dtantsur:
I'm generally in favour of this idea but I'd like a design proposal on it since it adds API. I'd rather not see API that simply exposes everything that Ironic collects as raw information.

Rozzii:
/triage accepted
I will close https://github.com/metal3-io/baremetal-operator/issues/1227 in favor of this issue, create a design proposal first please in the metal3-docs repo.

shibaPuppy:
@dtantsur @Rozzii 
I think it is reasonable to add LLDP content to the existing hardwareData.
do you think we need to add a new CR?

Rozzii:
I am fine with putting it to the hardwareData, what about you @dtantsur  ?

dtantsur:
It depends on how much you want to add, really. If it's just a couple of fields, I'm good with that.

PR #1167, opened 2022-09-19
Add LLDPProcessed information from ironic-inspector data

shibaPuppy:
Collect node LLDP information.

ex) if there is no LLDP information
```
apiVersion: metal3.io/v1alpha1
kind: BareMetalHost
metadata:
  name: node-3
  namespace: metal3
spec:
  automatedCleaningMode: metadata
  bmc:
    address: ipmi://192.168.111.1:6233
    credentialsName: node-3-bmc-secret
  bootMACAddress: 00:cc:0f:73:1b:fb
  bootMode: legacy
  online: true
status:
.....
    nics:
    - ip: 192.168.111.49
      lldp: {}
      mac: 00:cc:0f:73:1b:fd
      model: 0x1af4 0x0001
      name: enp2s0
    - ip: fe80::2cc:fff:fe73:1bfd%enp2s0
      lldp: {}
      mac: 00:cc:0f:73:1b:fd
      model: 0x1af4 0x0001
      name: enp2s0
    - ip: 172.22.0.83
      lldp: {}
      mac: 00:cc:0f:73:1b:fb
      model: 0x1af4 0x0001
      name: enp1s0
      pxe: true
    - ip: fe80::2cc:fff:fe73:1bfb%enp1s0
      lldp: {}
      mac: 00:cc:0f:73:1b:fb
      model: 0x1af4 0x0001
      name: enp1s0
      pxe: true
    ramMebibytes: 8192
....
```

PR #2752, opened 2025-10-28
✨ Capture and store LLDP data in HardwareData

alegacy:
Add support for capturing Link Layer Discovery Protocol (LLDP) data during hardware inspection and storing it in the HardwareData CRD. This provides network topology information that can be used for port identification and network configuration.

Changes:
- Add LLDP struct to store switch topology data (chassis ID, port ID, system name)
- Add LLDP field to NIC struct in BareMetalHost API types
- Implement getLLDPData() to extract LLDP data from Ironic introspection
- Populate LLDP data for each network interface during hardware inspection
- Update CRD manifests for BareMetalHost and HardwareData

The LLDP data includes:
- switch_chassis_id: Switch chassis identifier
- switch_port_id: Switch port identifier
- switch_system_name: Switch system name

Additional fields could be added in the future if necessary.

Fixes: #1731
Assisted-by: Claude Code/sonnet-4.5


honza:
Previous discussion #1167 for reference.

tuminoid:
/cc @dtantsur @Rozzii @lentzi90 
PTAL

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -678,6 +678,21 @@ type VLAN struct {
...
+	// +optional
+	SwitchID string `json:"switchID,omitempty"`
+
+	// The switch port ID from LLDP
+	// +optional
+	PortID string `json:"portID,omitempty"`
+
+	// The switch system name from LLDP
+	// +optional
+	SwitchInfo string `json:"switchInfo,omitempty"`
```
dtantsur: Why not call this switchSystemName then?

config/base/crds/bases/metal3.io_baremetalhosts.yaml
```diff
@@ -716,6 +716,19 @@ spec:
                             if one is present.  If both IPv4 and IPv6 addresses are present in a
                             dual-stack environment, two nics will be output, one with each IP.
                           type: string
+                        lldp:
```
dtantsur, line 719: Commenting here because github won't let me comment on the exact place: please make sure that the new lldp structure is not set for the hardware data on BMH, only in the new HardwareData resource. Otherwise, BMH will become even less readable and manageable than they are now.

pkg/provisioner/ironic/hardwaredetails/hardwaredetails_test.go
```diff
@@ -129,6 +129,130 @@ func TestGetVLANsMalformed(t *testing.T) {
...
+		SwitchID: "aa:bb:cc:dd:ee:ff",
+		PortID:   "Ethernet1/1",
+	}
+	if !reflect.DeepEqual(lldpData, expected) {
+		t.Errorf("Expected LLDP data %+v, got %+v", expected, lldpData)
+	}
+
+	// Test nil input
+	lldpData = getLLDPData(nil)
+	if lldpData != nil {
```
dtantsur: Please use things like assert.NotNil, assert.Equal, etc instead of direct comparisons.

config/base/crds/bases/metal3.io_baremetalhosts.yaml
dtantsur: This is probably the place where you can clone and hack the details: https://github.com/metal3-io/baremetal-operator/blob/main/internal/controller/metal3.io/baremetalhost_controller.go#L1049

apis/metal3.io/v1alpha1/baremetalhost_types.go
alegacy: I guess I was subconsciously lining this up with Ironic's local_link_connection (i.e., switch_info, switch_id, port_id) since I am planning on using this in BMO to interact with Ironic... eventually, hopefully, etc...

I can certainly rename this here to SwitchSystemName to better align with the LLDP field name.

config/base/crds/bases/metal3.io_baremetalhosts.yaml
alegacy: Ok, just so that I'm clear... you want the LLDP fields of the NICs explicitly cleared from the hardware details that are stored in bmh.spec.status.hardwaredetails, but kepts for hardwaredata.spec.hardwaredetails

dtantsur: Correct.

BMH v1beta1 will not have the former, only the separate CR.

apis/metal3.io/v1alpha1/baremetalhost_types.go
alegacy: updated

config/base/crds/bases/metal3.io_baremetalhosts.yaml
alegacy: updated

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -1046,7 +1058,9 @@ func (r *BareMetalHostReconciler) actionInspecting(prov provisioner.Provisioner,
 	}
 
 	clearError(info.host)
-	info.host.Status.HardwareDetails = details
+	// To avoid further cluttering the BMH status with more information we'll remove the LLDP
+	// info before updating the status, but retain it for the HardwareData contents.
+	info.host.Status.HardwareDetails = removeLLDPInfo(details)
```
Rozzii: Thanks for the contribution. In general the PR makes sense, I have one suggestion.

This is a bit subjective for me, I would suggest not doing this operation. I understand that LLDP info will be duplicated but for me that is a smaller "issue" compared to doing extra CR modifications just to remove data that we have queried before. For me this looks like it is a extra work for not much functional benefit.

In order to try to keep the reconciliation as "simple and lightweight" I would not add this cleanup unless there is an actual functional problem caused by this few extra lines of inspection data on the BMH.

Rozzii: I think this issue will be eventually solved by displaying inspection data only on the HardwareDetails CR.

dtantsur: I asked to do this because BMH will be unmanageable with LLDP data. I mean, it's already unmanageable... I don't feel strongly about this one but if we keep piling information into BMH, users will need to develop custom tools for handling BMH :D

alegacy:
something went wrong in the merge conflict resolution.   Will address it shortly!

internal/controller/metal3.io/baremetalhost_controller.go
Rozzii: Yeah these UX stuff is a bit subjective, I like to only check 1 place instead of 2 and also used to logs and manifests of MB size so I might be not the best to determine what is the most user friendly.  

For me this is more of a balancing act of do I want to get rid of some duplicate data because of etcd and CR size concerns or do I want to avoid "complicated" propagation of data  during reconciliation like deep copying data between CRs and automatically deleting data from one CR .

Would it be possible to check what is the size difference of the two CRs? Or paste examples here?

dtantsur: I guess if it's only 3 fields, it will not matter for most users... I'm fine if @alegacy drops removeLLDPInfo for the sake of speeding up merging this work.

PR #2754, opened 2025-10-29
:sparkles: Allow recovering hardware details from existing HardwareData

dtantsur:
With HardwareData CR, we don't actually need to use annotations to
populate the hardware details. And with new InspectionMode, we don't
need the disabled annotation either.

The goal here is to eventually un-teach people to rely on the annotation
as part of the transition to dedicated HardwareData resources only.


dtantsur:
An initially test-only change ended up being a feature

kashifest:
/lgtm
Indeed this is nice. I wonder how far are we now on not relying on BMH status anymore for source of truth. It would be really nice not to have any persistent data there.

dtantsur:
> I wonder how far are we now on not relying on BMH status

I think the biggest outlier is the logic for default CPU architecture, which is not aware of HardwareData.

PR #2761, opened 2025-11-04
[DNM] Test ironic-image change for master_iso_images

hroyrh:
Test e2e tests for the ironic-image pr https://github.com/metal3-io/ironic-image/pull/781


hroyrh:
The ironic-image pr has merged, so this is no longer needed.
/close

PR #2773, opened 2025-11-12
:book: Add explanation to run verify-release.sh script in releasing.md

Sunnatillo:
This PR adds explanation to run verify-release.sh script to releasing.md. Also, changes tag mentions from 1.x.x to 0.x.x to avoid accidental mistakes

**Checklist:**

### DIFF
--- a/docs/releasing.md
+++ b/docs/releasing.md
@@ -21,6 +21,19 @@ Things you should check before making a release:
    - Verify any other direct or indirect dependency is uplifted to close any public
      vulnerabilities
 
+  Run the script against the release branch you are releasing from, and pass the
+  version tag as an argument, for example:
+
+```shell
+    git checkout release-0.x
+    ./hack/verify-release.sh 0.x.x
+```
+
+Also, after creating the tag and preparing the draft release, run
+`./hack/verify-release.sh` again right before publishing. This script verifies
+that all required release artifacts exist, the release tags are correct, and
+that the container images were built and pushed successfully.
+
 ## Permissions
 
 Creating a release requires repository `write` permissions for:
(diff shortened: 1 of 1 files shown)


docs/releasing.md
```diff
@@ -21,6 +21,14 @@ Things you should check before making a release:
    - Verify any other direct or indirect dependency is uplifted to close any public
      vulnerabilities
 
+  Run the script against the release branch you are releasing from, and pass the
+  version tag as an argument, for example:
```
tuminoid, line 25: Do we want to also document the behavior difference when running it before release (tag) is done versus when its not (pre-release?)

tuminoid:
/retitle :book: Add explanation to run verify-release.sh script in releasing.md

docs/releasing.md
Sunnatillo: I can add the part explaining running after tag is created before release is published. But how is it different in pre-release?

tuminoid: Pre-release there is no tag, so it cannot check if images are built, release notes in place etc. So it will do only the checking if the repo itself is fine for cutting a release.

Sunnatillo: We have tag for pre releases. 
https://github.com/metal3-io/baremetal-operator/releases/tag/v0.11.0-rc.0
https://github.com/metal3-io/baremetal-operator/releases/tag/v0.12.0-beta.0

tuminoid: Sorry, without -, pre release ie. before release.

Sunnatillo: Sorry, I do not understand. We can discuss this in person

tuminoid: You can run the script before a release, and in case it tells you if something should be done before cutting a release. If you run it after the release, then it also checks if the images were build and pushed correctly, the tags are properly created etc. It thus "verifies" releases in two different ways: before and after.

MahnoorAsghar:
lgtm

kashifest (commented):
/approve

PR #2774, opened 2025-11-12
✨ Allow enabling ExternallyProvisioned for available hosts

sakhoury:

This PR relaxes the webhook validation for the `ExternallyProvisioned` field to allow it to be changed from `false` to `true` when the host is in an `Available` state, while still blocking changes in any other state.

Previously, the webhook rejected any change to `ExternallyProvisioned`. This prevented a critical workflow for O-RAN deployments using Red Hat's Image-based Installer (IBI).

IBI is used for deploying single-node OpenShift clusters in edge/O-RAN environments. The required workflow is:
1. Create BMH with `ExternallyProvisioned: false` to enable Ironic inspection
   - When `ExternallyProvisioned: true`, the [`NeedsHardwareInspection()`](https://github.com/metal3-io/baremetal-operator/blob/main/apis/metal3.io/v1alpha1/baremetalhost_types.go#L1031) function returns `false` (preventing inspection)
   - The controller state machine bypasses the Inspecting state entirely when the field is `true`
2. Ironic performs hardware inspection to gather CPU, memory, disk, and NIC details
3. Hardware data is used for O-RAN resource selection
4. Update BMH to `ExternallyProvisioned: true` to enable IBI provisioning
   - IBI requires `ExternallyProvisioned: true` as it handles provisioning externally
   - This signals to BMO that provisioning is managed outside the operator

**Checklist:**

### DIFF
--- a/test/e2e/externally_provisioned_test.go
+++ b/test/e2e/externally_provisioned_test.go
@@ -16,6 +16,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/cluster-api/test/framework"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 var _ = Describe("Create as externally provisioned, deprovision", Label("required", "provision", "deprovision"),
@@ -37,25 +38,26 @@ var _ = Describe("Create as externally provisioned, deprovision", Label("require
 			})
 		})
 
-		It("provisions a BMH as externally provisioned, then deprovisions", func() {
+		It("provisions a BMH as externally provisioned, validates state immutability, then deprovisions", func() {
+			testSecretName := secretName + "-external"
 			By("Creating a secret with BMH credentials")
 			bmcCredentialsData := map[string]string{
 				"username": bmc.User,
 				"password": bmc.Password,
 			}
-			CreateSecret(ctx, clusterProxy.GetClient(), namespace.Name, secretName, bmcCredentialsData)
+			CreateSecret(ctx, clusterProxy.GetClient(), namespace.Name, testSecretName, bmcCredentialsData)
 
 			By("Creating a BMH as externally provisioned")
 			bmh := metal3api.BareMetalHost{
 				ObjectMeta: metav1.ObjectMeta{
-					Name:      specName,
+					Name:      specName + "-external",
 					Namespace: namespace.Name,
 				},
 				Spec: metal3api.BareMetalHostSpec{
 					Online: true,
 					BMC: metal3api.BMCDetails{
 						Address:                        bmc.Address,
-						CredentialsName:                "bmc-credentials",
+						CredentialsName:                testSecretName,
 						DisableCertificateVerification: bmc.DisableCertificateVerification,
 					},
 					BootMACAddress:        bmc.BootMacAddress,
@@ -83,6 +85,28 @@ var _ = Describe("Create as externally provisioned, deprovision", Label("require
 			Expect(bmh.Status.OperationHistory.Inspect.Start.IsZero()).To(BeTrue())
 			Expect(bmh.Status.OperationHistory.Provision.Start.IsZero()).To(BeTrue())
 
+			By("Attempting to set ExternallyProvisioned to false (should be blocked by webhook)")
+			err = clusterProxy.GetClient().Get(ctx, types.NamespacedName{
+				Name:      bmh.Name,
+				Namespace: bmh.Namespace,
+			}, &bmh)
+			Expect(err).NotTo(HaveOccurred())
+
+			patch := client.MergeFrom(bmh.DeepCopy())
+			bmh.Spec.ExternallyProvisioned = false
+			err = clusterProxy.GetClient().Patch(ctx, &bmh, patch)
+			Expect(err).To(HaveOccurred(), "Webhook should reject transition from true to false")
+			Expect(err.Error()).To(ContainSubstring("externallyProvisioned can not be changed from true to false"))
+
+			By("Verifying BMH remains in ExternallyProvisioned state")
+			err = clusterProxy.GetClient().Get(ctx, types.NamespacedName{
+				Name:      bmh.Name,
+				Namespace: bmh.Namespace,
+			}, &bmh)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(bmh.Spec.ExternallyProvisioned).To(BeTrue(), "ExternallyProvisioned should still be true")
+			Expect(bmh.Status.Provisioning.State).To(Equal(metal3api.StateExternallyProvisioned))
+
 			By("Deleting the BMH")
 			// Wait for 2 seconds to allow time to confirm annotation is set
 			// TODO: fix this so we do not need the sleep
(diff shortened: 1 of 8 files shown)


dtantsur:
/ok-to-test

Have you checked what actually happens when you set `externallyProvisioned` to `true`? Have you checked it works correctly in all possible initial states?

sakhoury:
@dtantsur I have refined the validation to only allow enabling `externallyProvisioned` from the `Available` state.

test/e2e/externally_provisioned_test.go
```diff
@@ -103,18 +104,211 @@ var _ = Describe("Create as externally provisioned, deprovision", Label("require
...
+			Expect(err).NotTo(HaveOccurred())
+
+			WaitForBmhDeleted(ctx, WaitForBmhDeletedInput{
+				Client:    clusterProxy.GetClient(),
+				BmhName:   bmh.Name,
+				Namespace: bmh.Namespace,
+			}, e2eConfig.GetIntervals(specName, "wait-bmh-deleted")...)
+		})
+
+		It("blocks enabling ExternallyProvisioned when not in Available state", func() {
```
dtantsur: E2e tests are pretty expensive. This one is covered well enough by the webhook unit tests, so I suggest not adding it.

test/e2e/externally_provisioned_test.go
```diff
@@ -103,18 +104,211 @@ var _ = Describe("Create as externally provisioned, deprovision", Label("require
...
+			Expect(err).NotTo(HaveOccurred())
+
+			WaitForBmhDeleted(ctx, WaitForBmhDeletedInput{
+				Client:    clusterProxy.GetClient(),
+				BmhName:   bmh.Name,
+				Namespace: bmh.Namespace,
+			}, e2eConfig.GetIntervals(specName, "wait-bmh-deleted")...)
+		})
+
+		It("blocks transition from ExternallyProvisioned true to false", func() {
```
dtantsur: Instead of a new test case, let's add this verification to the existing "provisions a BMH as externally provisioned, then deprovisions"

test/e2e/externally_provisioned_test.go
```diff
@@ -35,16 +36,16 @@ var _ = Describe("Create as externally provisioned, deprovision", Label("require
...
 			})
-		})
 
-		It("provisions a BMH as externally provisioned, then deprovisions", func() {
 			By("Creating a secret with BMH credentials")
 			bmcCredentialsData := map[string]string{
 				"username": bmc.User,
 				"password": bmc.Password,
 			}
 			CreateSecret(ctx, clusterProxy.GetClient(), namespace.Name, secretName, bmcCredentialsData)
```
dtantsur: This part fails in the tests because of a duplicate

dtantsur:
/retitle ✨ Allow enabling ExternallyProvisioned for available hosts

test/e2e/externally_provisioned_test.go
sakhoury: Removed!

test/e2e/externally_provisioned_test.go
sakhoury: Done!

dtantsur:
/approve

Thanks!

sakhoury:
@Rozzii When you have a moment, could you please review this PR? Thank you!

Rozzii:
> @Rozzii When you have a moment, could you please review this PR? Thank you!

I am checking

Rozzii:
> > @Rozzii When you have a moment, could you please review this PR? Thank you!
> 
> I am checking

I need a little bit of time to 100% understand this workflow but I am on it :D . The use-case is clear but it is a bit unexpected that something is "internally inspected" but "externally provisioned" after inspection but makes sense, I am just wondering how would this translate to CAPI/CAPM3 machine povisioning specifically.

Rozzii:
/cc @kashifest @lentzi90 
FYI

Rozzii:
/cc @adilGhaffarDev 
Could you please take a look from CAPM3 perspective, what is your opinion could this have some unwanted side effect for CAPM3?
AFAIK the BMO side looks solid.

Rozzii:
@kashifest had a good point , based on this https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/baremetal/metal3machine_manager.go#L818-L835 it looks like that in CAPM3 is skipping the externally provisioned BMHs when it chooses a host.

The problem is that CAPM3 first  checks if the BMH is available or not and if it is then the BMH will be picked, but so far there was no situation wheren the Available BMH could become ExternallyProvisioned in the future and with this change we would have so this BMO change would induce an unexpected state of the Metal3Machine. 

We might need to agree on an annotation that would signal to CAPM3 that this BMH might turn into an externally provisioned BMH in the future so skip it even if it is available.

MahnoorAsghar:
@Rozzii would it not be possible for CAPM3 to ignore the BMH as soon as it becomes ExternallyProvisioned: true?

dtantsur:
> We might need to agree on an annotation that would signal to CAPM3 that this BMH might turn into an externally provisioned BMH in the future so skip it even if it is available.

@kashifest @Rozzii could it be just a question of documentation, sort of "if you use CAPM3, don't do this"?

dtantsur:
I'm also wondering how it's different from CAPM3 racing with something else for the same available hosts?

kashifest:
> @Rozzii would it not be possible for CAPM3 to ignore the BMH as soon as it becomes ExternallyProvisioned: true?

My concern is would it go to `Available` after inspection for a moment while` ExternallyProvisioned` is still `false`. if yes then theres a chance for race.

 
> @kashifest @Rozzii could it be just a question of documentation, sort of "if you use CAPM3, don't do this"?

Currently the high level workflow in CAPM3 while choosing a BMH for Metal3 machine is
1. If it already has a consumerRef (meaning an m3m is associated) ignore this host
2. If it already has a `nodereuse` label ignore this host
3. If the host has `paused` or `unhealthy` annotation ignore this host
4. Switch-case: 
  a. the host is `ready` or `available` choose this host 
  b. the host is `externallyprovisioned` state or `provisioned`  state or etc (few other states) ignore this host.
  
Hence my previous concern. So whats your opinion on this @dtantsur @MahnoorAsghar . cc @Rozzii

dtantsur:
> Currently the high level workflow in CAPM3 while choosing a BMH for Metal3 machine is

@kashifest thanks! So how are races with other potential consumers handled? If they aren't, I don't think this patch adds match to the problem.

kashifest:
> @kashifest thanks! So how are races with other potential consumers handled? If they aren't, I don't think this patch adds match to the problem.

What other races you mean? I didnt get your point.

kashifest:
I am probably misunderstanding the feature. Let me phrase this way "is it fine that CAPM3 chooses an `available` host for provisioning, while you later mark it as `externallyprovisioned`? " Wouldn't it create a conflicting situation? or am I understanding it wrongly and it will never happen like this

dtantsur:
> What other races you mean?

This whole discussion assumes co-existence of CAPM3 with its regular flow and *something else* that can mark an available host as externally provisioned. These two entities can race for the same hosts, so we may end up with a situation where *something else* marks a host externally provisioned after this host was chosen by CAPM3. Right?

Note, by the way, that it's not the case downstream: the hosts in questions won't be handled by CAPM3.

So, my question is, how is this sort of situation handled right now without externally provisioned in the picture? I.e. if CAPM3 and *something else* try to provision (normally, without externally provisioned) the same hosts?

kashifest:
> This whole discussion assumes co-existence of CAPM3 with its regular flow and _something else_ that can mark an available host as externally provisioned. These two entities can race for the same hosts, so we may end up with a situation where _something else_ marks a host externally provisioned after this host was chosen by CAPM3. Right?

Yeah thats true, But from CAPM3 perspective prior to this change, we won't have this issue right? you won't have an `available`  bmh which you won't want to CAPM3 to `provision` ? Ofcourse, we have a case like `nodereuse` as I explained for upgrades where we use labels to make sure the host is not chosen even if available and thats what @Rozzii also meant perhaps. May be add a label/annotation to make sure CAPM3 ignores it. 

> Note, by the way, that it's not the case downstream: the hosts in questions won't be handled by CAPM3.

Ok. fair enough. I am more thinking of other consumers of CAPM3. I had a chat with one of the consumers of Metal3 during KubeCon NA'25. (they are not public adopters yet). They are using CAPM3 and they use `externallyprovisioned` hosts also. 

I am very much in favour of this feature. My only point is lets make it also working when CAPM3 is in the picture.

dtantsur:
> But from CAPM3 perspective prior to this change, we won't have this issue right? you won't have an available bmh which you won't want to CAPM3 to provision ?

That's my question, basically. Imagine we have two independent CAPM3 instances (can be anything else) working with the same hosts. How do we prevent them trying to use the same available hosts at (nearly exactly) the same time?

dtantsur:
E.g. if we rely on both consumers setting `consumerRef` before doing anything else (and rerunning the scheduler on conflict), then we can just document that `consumerRef` must be set together with `externallyProvisioned` on available hosts?

Rozzii:
> > But from CAPM3 perspective prior to this change, we won't have this issue right? you won't have an available bmh which you won't want to CAPM3 to provision ?
> 
> That's my question, basically. Imagine we have two independent CAPM3 instances (can be anything else) working with the same hosts. How do we prevent them trying to use the same available hosts at (nearly exactly) the same time?

This is actually a very good example. The situation that you have painted should be handled by the host selector of the Metal3MachineTemplate. HostSelector works based on label, the selector is matched against label and picks the BMHs that are available, has no owner reference and and has the selected label. 

Technically one could use the HostSelector to block picking a machine that might transition from externallyProvisioned: false to true . The issue is that host selection can't really make a mistake in the current implementation. In case the user picks an incorrect host with an incorrect label the CAPI Machine provisioning might get stuck because the machine would not boot but functionally no feature of CAPI, CAPM3 are blocked. Even if stuck the Machine and M3M are in a known stable or transitional state with  other word it is still "withing the bounds" of expected operational behavior, the same is also true for the theoretical situation of 2 CAMP3 fighting over one BMH, one will win and pick the host the other won't, even the hostSelector label overlap and both will end up "within the bounds". 

The problem with this feature is that it opens the possibility that host selection goes wrong silently and potentially induce a side effects e.g CAPM3 starts generating secrets for the machine, it might even attempt to initiate a provisioning... who knows what would this scenario induce .

For me this feature wold look OK if it would require a label or annotation (label is fine) to be set in order to allow the ExternallyProvisioned to be set to true if the machine is in available state. So there would be 1 additional condition, and then on CAPM3 side we can look for this label to make sure this machine would never be picked. I don't think it is a big modification.


So I am fine allowing setting externallyProvisioned to true but let's have +1 condition that checks whether the label is there .

adilGhaffarDev:
Suggestion: Instead of modifying the existing validations, why not introduce a new flag that explicitly allows inspection when ExternallyProvisioned is set to true? For example, something like AllowInspectionExternallyProvisioned. I’m not sure how far-fetched this idea is, but it seems like a less intrusive alternative.

Label and annotation suggested by Adam and Kashif are also good way to go.

dtantsur:
> Suggestion: Instead of modifying the existing validations, why not introduce a new flag that explicitly allows inspection when ExternallyProvisioned is set to true? For example, something like AllowInspectionExternallyProvisioned. 

Firmly against it, this changes the meaning of ExternallyProvisioned. I also don't want to create the 3rd flag if people ever want to do preparation.

> The situation that you have painted should be handled by the host selector of the Metal3MachineTemplate.

Yeah, but nothing guarantees that. A user can make a mistake, as you rightfully note. Also, what about a provisioner that does not use Metal3 machines? How will that conflict be handled before this patch?

> So I am fine allowing setting externallyProvisioned to true but let's have +1 condition that checks whether the label is there

Why new label when we already have a consumerRef? Let's enforce having a non-empty consumerRef? @kashifest @Rozzii?

dtantsur:
I think the focus on ExternallyProvisioned specifically is wrong here. Let's look at the bigger picture: we have *another provisioner* that is working with the same hosts as CAPM3. How do we prevent them from clashes?

Metal3MachineTemplate is not an answer: *another provisioner* may not even have a host selector to begin with. I assume, `consumerRef` is essentially a lock that is relying on Kubernetes rejecting conflicting PUT operations. So, if the two provisioners can agree on checking `consumerRef` and setting it as part of the very first provisioning step, the conflict is resolved. One of the provisioners will need to retry, the other one will win.

If we think we need a new label, it has to be added regardless of this PR, because this PR does not really change anything.

Rozzii:
After discussions, I am fine with this approach but please document that when used with CAPM3 make sure the BMH is labelled so that the host Selector of CAPM3 can avoid such hosts it if needed.

dtantsur:
@sakhoury please rebase the PR and update the documentation per Adam's suggestion.

PR #2775, opened 2025-11-13
🌱 E2E: External inspection name collision fix

lentzi90:

We have two external inspection tests (one for annotation and one for hardwaredata). Both were using the same name both for BMH and BMC secret and in the same namespace. Since we run multiple tests simultaneously, this could lead to failures. This commit fixes the issue by giving each BMH and BMC secret its own name.

Fixes #

**Checklist:**


lentzi90:
Our periodic e2e tests have been quite unstable since 4th November. I think the reason is the name collision that came with https://github.com/metal3-io/baremetal-operator/pull/2754. This should fix it.

tuminoid (commented):
/lgtm

PR #2791, opened 2025-11-17
🌱 E2E: Bump kubectl minimum version to v1.34.1

### LINKED ISSUE #2708: Bump ensured kubectl minimum version in e2e
As per previous discussion here: #2699 

Bump the kubectl version in [ensure_kubectl.sh](https://github.com/metal3-io/baremetal-operator/blob/main/hack/e2e/ensure_kubectl.sh#L20) to 1.33.x. so that it is not too old for kubectl, which is supporting version skew of 2 minors.

lentzi90:

Keep kubectl up to date. Kind is already on v1.34 so let's have kubectl on the same minor version.

Fixes #2708

**Checklist:**

### DIFF
--- a/hack/e2e/ensure_kubectl.sh
+++ b/hack/e2e/ensure_kubectl.sh
@@ -17,7 +17,7 @@
 set -eux
 
 USR_LOCAL_BIN="/usr/local/bin"
-MINIMUM_KUBECTL_VERSION=v1.28.1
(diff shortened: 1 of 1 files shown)


lentzi90:
Ah we have v1.34.1 installed, and the script only installs if it is not installed at all. Switching the version to v1.34.1.

tuminoid (commented):
/lgtm

Good catch, version skew has grown too large.

PR #2793, opened 2025-11-17
:sparkles: Make bootMACAddress optional for virtual media with inspection enabled

donpenney:
Virtual media BMC drivers (redfish-virtualmedia, idrac-virtualmedia,
ilo5-virtualmedia) can boot from virtual media and discover the MAC
address during hardware inspection. Therefore, bootMACAddress is optional
when inspection is enabled but still required when inspection is disabled
(since there's no other way to discover the MAC address).

This change updates the virtual media driver implementations to return
false from their NeedsMAC() methods, and adds validation logic to require
bootMACAddress when virtual media is used with inspection disabled.

Drivers that require PXE boot (like libvirt, ipmi) continue to require
bootMACAddress in all cases.

Changes:
- pkg/hardwareutils/bmc/redfish_virtualmedia.go: NeedsMAC() returns false
- pkg/hardwareutils/bmc/idrac_virtualmedia.go: NeedsMAC() returns false
- internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go:
  Add logic to require bootMACAddress for virtual media when inspection
  is disabled (via InspectionMode field or inspect.metal3.io annotation)
- pkg/provisioner/ironic/ironic.go: Skip MAC-based port queries when
  bootMACAddress is empty to prevent false MAC conflicts

Test Coverage:
- internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation_test.go:
  * bootMACAddress not required for virtual media with inspection enabled
  * bootMACAddress required for virtual media with inspection disabled
  * bootMACAddress valid when provided for virtual media with inspection disabled
- pkg/hardwareutils/bmc/access_test.go:
  Update virtual media test expectations (needsMac: false)

Assisted-By: Claude <noreply@anthropic.com>

**Checklist:**

### DIFF
--- a/internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
+++ b/internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
@@ -141,11 +141,8 @@ func validateBMCAccess(host *metal3api.BareMetalHost, bmcAccess bmc.AccessDetail
 	}
 
 	// Check if bootMACAddress is required
-	requiresMAC := bmcAccess.NeedsMAC()
 	// Virtual media drivers (NeedsMAC() returns false) still require MAC when inspection is disabled
-	if !requiresMAC && host.InspectionDisabled() {
-		requiresMAC = true
-	}
+	requiresMAC := bmcAccess.NeedsMAC() || host.InspectionDisabled()
 
 	if requiresMAC && s.BootMACAddress == "" {
 		errs = append(errs, fmt.Errorf("BMC driver %s requires a BootMACAddress value", bmcAccess.Type()))


dtantsur:
I'm not sure I understand the link between PreprovisioningNetworkDataName and BootMacAddress. Honestly, I suggest just dropping the requirement on BootMacAddress for virtual media based drivers.

donpenney:
> I'm not sure I understand the link between PreprovisioningNetworkDataName and BootMacAddress. Honestly, I suggest just dropping the requirement on BootMacAddress for virtual media based drivers.

Thanks @dtantsur . I was aiming to be a bit surgical here to minimize behavoural changes. My thinking was that if the network config has been provided in an nmstate secret as part of the initial configuration, then the BootMacAddress was unnecessary for the inspection stage. I assumed it would always be the case for inspection with virtual media, but hadn't thought about using that as the determining factor, so I'll go that route instead.

dtantsur:
For background: since quite some time, the Ironic inspection implementation for Redfish is actually hybrid: it starts with getting MAC addresses from Redfish, then reboots into the agent ramdisk.

Now that I said that, I think we should still require BootMacAddress if inspection is disabled on the BMH @donpenney.

internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -109,7 +109,7 @@ func (webhook *BareMetalHost) validateChanges(oldObj *metal3api.BareMetalHost, n
 		errs = append(errs, errors.New("BMC address can not be changed if the BMH is not in the Registering state, or if the BMH is not detached"))
 	}
 
-	if oldObj.Spec.BootMACAddress != "" && !strings.EqualFold(newObj.Spec.BootMACAddress, oldObj.Spec.BootMACAddress) {
+	if oldObj.Spec.BootMACAddress != "" && newObj.Spec.BootMACAddress != oldObj.Spec.BootMACAddress {
```
dtantsur: What's the reason you're reverting the EqualFold part? I think it was here on purpose so that we're not case sensitive.

internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -120,8 +120,22 @@ func (webhook *BareMetalHost) validateChanges(oldObj *metal3api.BareMetalHost, n
 	return errs
 }
 
-func validateBMCAccess(s metal3api.BareMetalHostSpec, bmcAccess bmc.AccessDetails) []error {
+// isInspectionDisabled checks if inspection is disabled via annotation or spec field.
+func isInspectionDisabled(host *metal3api.BareMetalHost) bool {
```
dtantsur: There is already `host.InspectionDisabled`

donpenney: Thanks! I'll get these fixed up and retested in my lab, then update the PR

internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
donpenney: Fixed

internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
donpenney: Fixed

donpenney:
Rebased my branch and removed a change to a test case that had been failing in my env due to running with golang 1.24.5. The format of an url.Parse error message changed in 1.24.9. Because my golang was older, the test was failing in my environment. I have now removed this unrelated change from my update

internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -139,7 +140,14 @@ func validateBMCAccess(s metal3api.BareMetalHostSpec, bmcAccess bmc.AccessDetail
 		}
 	}
 
-	if bmcAccess.NeedsMAC() && s.BootMACAddress == "" {
+	// Check if bootMACAddress is required
+	requiresMAC := bmcAccess.NeedsMAC()
+	// Virtual media drivers (NeedsMAC() returns false) still require MAC when inspection is disabled
+	if !requiresMAC && host.InspectionDisabled() {
+		requiresMAC = true
+	}
```
dtantsur: nit:

```go
requiresMAC := bmcAccess.NeedsMAC() || host.InspectionDisabled()
```

donpenney: Fixed

Issue #2478, opened 2025-05-22
BMO cannot abort inspection when deletion is requested


dtantsur:
If you try to delete a host in the `inspecting` state, it will go to the `powering off before delete` state, but the Ironic node will stay in inspection:

```
NAME                   STATE                        CONSUMER                ONLINE   ERROR   AGE
ostest-extraworker-0   powering off before delete                           true             3m29s
ostest-master-0        provisioned                  ostest-6r4cc-master-0   true             56m
ostest-master-1        provisioned                  ostest-6r4cc-master-1   true             56m
ostest-master-2        provisioned                  ostest-6r4cc-master-2   true             56m
+--------------------------------------------+--------------------+
| Name                                       | Provisioning State |
+--------------------------------------------+--------------------+
| openshift-machine-api~ostest-master-2      | active             |
| openshift-machine-api~ostest-master-0      | active             |
| openshift-machine-api~ostest-master-1      | active             |
| openshift-machine-api~ostest-extraworker-0 | inspect wait       |
+--------------------------------------------+--------------------+
```

The logs show

```
{"level":"info","ts":1747930123.4329689,"logger":"provisioner.ironic","msg":"host in state that does not allow power change, try again after delay","host":"openshift-machine-api~ostest-extra
worker-0","state":"inspect wait","target state":"manageable"}
```

It is going to work eventually, but it's a bad user experience. Inspection and cleaning can be aborted in Ironic.

/kind bug
/help

PR #2796, opened 2025-11-20
:bug: Abort inspection/cleaning before powering off during deletion

elfosardo:
When deleting a host in inspecting or cleaning state, it would get stuck because Ironic rejects power changes while a target provision state is active. This fix detects these states and sends an abort command first, allowing deletion to proceed immediately instead of waiting for inspection/cleaning to complete.

Assisted-By: Claude Sonnet 4.5 (commercial license)

Closes: https://github.com/metal3-io/baremetal-operator/issues/2478

### DIFF
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -1628,14 +1628,73 @@ func (p *ironicProvisioner) PowerOn(force bool) (result provisioner.Result, err 
 	return result, nil
 }
 
+// abortInspectionOrCleaning aborts inspection or cleaning if the node is in one of those states.
+// This is necessary to allow power changes when the node is being deleted.
+// Abort only works in *Wait states; for *ing states we return Dirty to retry later.
+// For cleaning, we abort manual cleaning or automated cleaning if it's been disabled.
+func (p *ironicProvisioner) abortInspectionOrCleaning(ironicNode *nodes.Node, automatedCleaningMode metal3api.AutomatedCleaningMode) (result provisioner.Result, err error) {
+	provState := nodes.ProvisionState(ironicNode.ProvisionState)
+
+	switch provState {
+	case nodes.InspectWait:
+		p.log.Info("aborting inspection to allow power off during deletion")
+		return p.changeNodeProvisionState(
+			ironicNode,
+			nodes.ProvisionStateOpts{Target: nodes.TargetAbort},
+		)
+	case nodes.Inspecting:
+		p.log.Info("inspection in progress, waiting for it to reach inspect wait state")
+		return operationContinuing(provisionRequeueDelay)
+	case nodes.CleanWait:
+		// Abort manual cleaning or automated cleaning if disabled.
+		// Use automatedCleaningMode from BMH spec rather than ironicNode.AutomatedClean
+		// which may be stale if BMO couldn't update it during cleaning states.
+		isManualCleaning := ironicNode.TargetProvisionState == string(nodes.TargetClean)
+		if isManualCleaning || automatedCleaningMode == metal3api.CleaningModeDisabled {
+			if isManualCleaning {
+				p.log.Info("aborting manual cleaning to allow power off during deletion")
+			} else {
+				p.log.Info("aborting automated cleaning (disabled) to allow power off during deletion")
+			}
+			return p.changeNodeProvisionState(
+				ironicNode,
+				nodes.ProvisionStateOpts{Target: nodes.TargetAbort},
+			)
+		}
+		// Automated cleaning is enabled - let it finish
+		p.log.Info("automated cleaning in progress, waiting for it to complete")
+		return operationContinuing(provisionRequeueDelay)
+	case nodes.Cleaning:
+		// Check if it's manual cleaning or automated cleaning that's been disabled
+		if ironicNode.TargetProvisionState == string(nodes.TargetClean) ||
+			automatedCleaningMode == metal3api.CleaningModeDisabled {
+			p.log.Info("cleaning in progress, waiting for it to reach clean wait state before aborting")
+		} else {
+			p.log.Info("automated cleaning in progress, waiting for it to complete")
+		}
+		return operationContinuing(provisionRequeueDelay)
+	default:
+		// Node is not in a state that needs aborting
+		return operationComplete()
+	}
+}
+
 // PowerOff ensures the server is powered off independently of any image
 // provisioning operation.
-func (p *ironicProvisioner) PowerOff(rebootMode metal3api.RebootMode, force bool) (result provisioner.Result, err error) {
+func (p *ironicProvisioner) PowerOff(rebootMode metal3api.RebootMode, force bool, automatedCleaningMode metal3api.AutomatedCleaningMode) (result provisioner.Result, err error) {
 	p.log.Info(fmt.Sprintf("ensuring host is powered off (mode: %s)", rebootMode))
 
 	ironicNode, err := p.getNode()
 	if err != nil {
 		return transientError(err)
+	}
+
+	// If the node is in inspection or cleaning, we need to handle it before we can power off.
+	// For *Wait states, we can abort. For *ing states, we wait for them to transition to *Wait.
+	// This is especially important during deletion to avoid getting stuck waiting for inspection/cleaning to complete.
+	result, err = p.abortInspectionOrCleaning(ironicNode, automatedCleaningMode)
+	if err != nil || result.Dirty {
+		return result, err
 	}
 
 	if ironicNode.PowerState != powerOff {
@@ -1670,7 +1729,7 @@ func (p *ironicProvisioner) PowerOff(rebootMode metal3api.RebootMode, force bool
 		return p.changePower(ironicNode, powerTarget)
 	}
 
-	return operationComplete()
+	return result, nil
 }
 
 func ironicNodeName(objMeta metav1.ObjectMeta) string {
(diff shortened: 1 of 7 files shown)


pkg/provisioner/ironic/ironic.go
```diff
@@ -1631,6 +1628,30 @@ func (p *ironicProvisioner) PowerOn(force bool) (result provisioner.Result, err
 	return result, nil
 }
 
+// abortInspectionOrCleaning aborts inspection or cleaning if the node is in one of those states.
+// This is necessary to allow power changes when the node is being deleted.
+func (p *ironicProvisioner) abortInspectionOrCleaning(ironicNode *nodes.Node) (result provisioner.Result, err error) {
+	provState := nodes.ProvisionState(ironicNode.ProvisionState)
+
+	switch provState {
+	case nodes.InspectWait, nodes.Inspecting:
```
dtantsur: You cannot abort anything in *ing states, only in *Wait. For *ing states, we can only wait a bit.

pkg/provisioner/ironic/ironic.go
```diff
@@ -1641,6 +1662,22 @@ func (p *ironicProvisioner) PowerOff(rebootMode metal3api.RebootMode, force bool
 		return transientError(err)
 	}
 
+	// If the node is in inspection or cleaning, we need to abort it first before we can power off.
+	// This is especially important during deletion to avoid getting stuck waiting for inspection/cleaning to complete.
+	if ironicNode.TargetProvisionState != "" {
```
dtantsur: This is not necessary, the check for provision state suffices

pkg/provisioner/ironic/ironic.go
```diff
@@ -1631,6 +1628,30 @@ func (p *ironicProvisioner) PowerOn(force bool) (result provisioner.Result, err
...
+	provState := nodes.ProvisionState(ironicNode.ProvisionState)
+
+	switch provState {
+	case nodes.InspectWait, nodes.Inspecting:
+		p.log.Info("aborting inspection to allow power off during deletion")
+		return p.changeNodeProvisionState(
+			ironicNode,
+			nodes.ProvisionStateOpts{Target: nodes.TargetAbort},
+		)
+	case nodes.CleanWait, nodes.Cleaning:
```
dtantsur: Only abort manual cleaning (here you can actually check TargetProvisionState) or automated cleaning if the user has disabled it in the meantime (I'm not sure if you can check Node.AutomatedClean, BMO probably won't be able to update it during cleaning; you may want to get this information all the way from the controller).

pkg/provisioner/ironic/ironic.go
```diff
@@ -1641,6 +1662,22 @@ func (p *ironicProvisioner) PowerOff(rebootMode metal3api.RebootMode, force bool
 		return transientError(err)
 	}
 
+	// If the node is in inspection or cleaning, we need to abort it first before we can power off.
+	// This is especially important during deletion to avoid getting stuck waiting for inspection/cleaning to complete.
+	if ironicNode.TargetProvisionState != "" {
+		provState := nodes.ProvisionState(ironicNode.ProvisionState)
+		if provState == nodes.InspectWait || provState == nodes.Inspecting ||
+			provState == nodes.CleanWait || provState == nodes.Cleaning {
```
dtantsur: Both the state check and the log message are duplicated here and in abortInspectionOrCleaning.

pkg/provisioner/ironic/ironic.go
elfosardo: right....
I guess we just wait until we're in *Wait state for inspection and cleaning before aborting then

pkg/provisioner/ironic/ironic.go
elfosardo: yeah, makes sense
maybe check ironicNode AutomatedClean?
I'll propose a change, see if that could work

dtantsur: ironicNode.AutomatedClean may be outdated if BMO is unable to set its value (I don't remember if it's updateable during *WAIT states).

pkg/provisioner/ironic/ironic.go
```diff
@@ -1631,6 +1628,61 @@ func (p *ironicProvisioner) PowerOn(force bool) (result provisioner.Result, err
...
+	case nodes.Cleaning:
+		// Check if it's manual or automated cleaning
+		isManualOrDisabled := ironicNode.TargetProvisionState == string(nodes.TargetClean) ||
+			(ironicNode.AutomatedClean != nil && !*ironicNode.AutomatedClean)
+		if isManualOrDisabled {
+			p.log.Info("cleaning in progress, waiting for it to reach clean wait state before aborting")
+			return operationContinuing(provisionRequeueDelay)
+		}
+		// Automated cleaning is enabled, let it finish
+		p.log.Info("automated cleaning in progress, waiting for it to complete")
```
lentzi90: These two seem very similar to me. Is it worth it to separate them, or could we just skip the check for manual or disabled? I mean we would always just log "cleaning is in progress..." and return. No need to check anything else if we are in `Cleaning` state.

elfosardo: we need to separate automated or manual cleaning, we actually don't want to abort automated cleaning, that's why we're doing this distinction
check also @dtantsur comment in the discussion

pkg/provisioner/ironic/ironic.go
elfosardo: mmmmmm what if we abort only manual cleaning and let automated cleaning finish ?

dtantsur: I'm fine with limiting the scope for now, but we really need to allow aborting automated cleaning by disabling it through BMH.

elfosardo: before reducing the scope, I'll propose another implementation to include the "automated cleaning disabled" scenario
the only way I can think about is passing AutomatedCleaningMode from BMH spec, changing the Provisioner interface
it should not have any external impact

elfosardo:
/retest
weird failure, this change should not influence the failing part

pkg/provisioner/ironic/ironic.go
lentzi90: I could be missing something, but it looks to me like we just put a different log message now. We return the exact same thing for these two cases. Or perhaps we are talking about different things? I mean this:

```go
	case nodes.Cleaning:
		// Check if it's manual cleaning or automated cleaning that's been disabled
		if ironicNode.TargetProvisionState == string(nodes.TargetClean) ||
			automatedCleaningMode == metal3api.CleaningModeDisabled {
			p.log.Info("cleaning in progress, waiting for it to reach clean wait state before aborting")
            // NOTE: This is the same return as outside the if, so it is not necessary.
			return operationContinuing(provisionRequeueDelay)
		}
		// Automated cleaning is enabled - let it finish
		p.log.Info("automated cleaning in progress, waiting for it to complete")
		return operationContinuing(provisionRequeueDelay)
```
If I rewrite a bit to make my point clearer:

```go
		// Check if it's manual cleaning or automated cleaning that's been disabled
		if ironicNode.TargetProvisionState == string(nodes.TargetClean) ||
			automatedCleaningMode == metal3api.CleaningModeDisabled {
			p.log.Info("special manual cleaning message")
		} else {
		    p.log.Info("automated cleaning message")
        }
		return operationContinuing(provisionRequeueDelay)
```

Maybe we could just have a generic message to simplify this?

elfosardo: I'd rather keep two messages to be very clear if that's automated cleaning or not, but definitely going to switch to the if/else statement

elfosardo: @lentzi90 I've refactored the code a little to simplify it

lentzi90 (commented):
/lgtm
Adding hold in case I missed somethings still being discussed. Feel free to drop it if all was resolved already.
/hold

pkg/provisioner/ironic/ironic.go
```diff
@@ -1628,16 +1628,75 @@ func (p *ironicProvisioner) PowerOn(force bool) (result provisioner.Result, err
...
 	}
 
+	// If the node is in inspection or cleaning, we need to handle it before we can power off.
+	// For *Wait states, we can abort. For *ing states, we wait for them to transition to *Wait.
+	// This is especially important during deletion to avoid getting stuck waiting for inspection/cleaning to complete.
+	result, err = p.abortInspectionOrCleaning(ironicNode, automatedCleaningMode)
+	if err != nil || result.Dirty {
+		return result, err
+	}
+
```
MahnoorAsghar, line 1699: Apologies in advance if I am mistaken here, but once `p.abortInspectionOrCleaning` returns successfully, if shouldnt go into the next if block, correct? L1700-L1730?

elfosardo: no need to apologize!
that's correct, only when the node is NOT in inspection/cleaning state, we will go through the next if statement (the actual power off logic)
we return early in case of abort command sent or if we're waiting for inspection or cleaning

MahnoorAsghar: Do we need to return result at L1732?

elfosardo: you mean instead of return operationComplete() ?
we could, they both work, and are basically identical
it's just a style improvement, probably returning result is a bit more clean

elfosardo: I changed it, it's still an improvement :)

PR #2812, opened 2025-11-28
:seedling: add unit tests to SecretManager

tuminoid:
Add unit tests to SecretManager, similar to IRSO's tests in https://github.com/metal3-io/ironic-standalone-operator/pull/441 . Take some, give some!

**Checklist:**

### DIFF
--- a/pkg/secretutils/label_test.go
+++ b/pkg/secretutils/label_test.go
@@ -1,0 +1,23 @@
+package secretutils
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func TestAddSecretSelector_NilInput(t *testing.T) {
+	result := AddSecretSelector(nil)
+	assert.Len(t, result, 1)
+}
+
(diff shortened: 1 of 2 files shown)


tuminoid:
Fixed similar issues as in source PR, and rebased.

kashifest:
Its been long I have used these assertions. When something breaks in the code, would this output the reason of the failure as logs from the code or just the failure since I don't see any log outputs here. I mean the test itself is not generating any log but would it atleast propagate the failure log from the code itself it is testing ?

tuminoid:
> Its been long I have used these assertions. When something breaks in the code, would this output the reason of the failure as logs from the code or just the failure since I don't see any log outputs here. I mean the test itself is not generating any log but would it atleast propagate the failure log from the code itself it is testing ?

Unit tests should never log by design. They assert, which tell the failure reason. None of the unit tests in BMO or in our other repos print logs.

kashifest:
> Unit tests should never log by design. They assert, which tell the failure reason. None of the unit tests in BMO or in our other repos print logs.

Ok so it will propagate the failure reason. I was not meaning add logs here. My question was how assert handles the failure reason. 

/hold cancel

MahnoorAsghar:
lgtm

PR #2813, opened 2025-11-28
:sparkles: add image SBOM attestation to image builds

tuminoid:
Use reusable container build workflow from project-infra to enable image SBOM attestations to be attached to the image digest when any IPAM image is built.

For this, we need id-token permission for keyless signing and enabling the SBOM generation via flag.


tuminoid:
/hold
until next week

PR #2815, opened 2025-11-28
🌱 Apply gopls modernize suggestions

fmuyassarov:

Refactor code to use modern Go idioms and patterns.

### DIFF
--- a/main.go
+++ b/main.go
@@ -476,11 +476,8 @@ func getMaxConcurrentReconciles(controllerConcurrency int) (int, error) {
 
 	// controller-concurrency value is 0 i.e. no values passed via the flag
 	// maxConcurrentReconcile value would be set based on env var or number of CPUs.
-	maxConcurrentReconciles := runtime.NumCPU()
-	if maxConcurrentReconciles > 8 { //nolint:mnd
-		maxConcurrentReconciles = 8
-	}
-	if maxConcurrentReconciles < 2 { //nolint:mnd
+	maxConcurrentReconciles := min(runtime.NumCPU(), 8) //nolint:mnd
+	if maxConcurrentReconciles < 2 {                    //nolint:mnd
 		maxConcurrentReconciles = 2
 	}
 	if mcrEnv, ok := os.LookupEnv("BMO_CONCURRENCY"); ok {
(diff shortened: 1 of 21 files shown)


pkg/utils/stringlist.go
```diff
@@ -1,14 +1,11 @@
...
 // StringInList returns a boolean indicating whether strToSearch is a
 // member of the string slice passed as the first argument.
 func StringInList(list []string, strToSearch string) bool {
-	for _, item := range list {
-		if item == strToSearch {
-			return true
-		}
-	}
-	return false
+	return slices.Contains(list, strToSearch)
```
lentzi90: We should `sed s/StringInList/slices.Contains/g` throughout the code base and drop this function IMO. Let's do that separately though. I also see it is exported, so we may have to put that as a breaking change...

main.go
```diff
@@ -476,10 +476,9 @@ func getMaxConcurrentReconciles(controllerConcurrency int) (int, error) {
 
 	// controller-concurrency value is 0 i.e. no values passed via the flag
 	// maxConcurrentReconcile value would be set based on env var or number of CPUs.
-	maxConcurrentReconciles := runtime.NumCPU()
-	if maxConcurrentReconciles > 8 { //nolint:mnd
-		maxConcurrentReconciles = 8
-	}
+	maxConcurrentReconciles := min(runtime.NumCPU(),
+		//nolint:mnd
+		8)
```
lentzi90: Is it possible to put this all on one line? Does that break the linter?

fmuyassarov: I think that won't break the linter, but the line break was part of gopls.

fmuyassarov: Fixed now.

pkg/utils/stringlist.go
lentzi90: Created issue for this: https://github.com/metal3-io/baremetal-operator/issues/2818

fmuyassarov: How about this https://github.com/metal3-io/baremetal-operator/pull/2819

pkg/utils/stringlist.go
```diff
@@ -1,14 +1,11 @@
 package utils
 
+import "slices"
+
 // StringInList returns a boolean indicating whether strToSearch is a
 // member of the string slice passed as the first argument.
 func StringInList(list []string, strToSearch string) bool {
```
dtantsur: Let's remove this function and maybe this module entirely?

fmuyassarov: it is done here separately https://github.com/metal3-io/baremetal-operator/pull/2819

fmuyassarov:
Had to rebase as the other patch landed before this.

Issue #2818, opened 2025-12-01
Refactor away unecessary StringInList function


lentzi90:
See: https://github.com/metal3-io/baremetal-operator/blob/f3a2ec6dd1a5d6a708647f7a17b3b5ca8e67f212/pkg/utils/stringlist.go#L5

This function can be replaced with strings.Contains now so there is no need for it.
Replace it throughout the code base. Then remove the function completely.

fmuyassarov:
Fix: https://github.com/metal3-io/baremetal-operator/pull/2819

PR #2819, opened 2025-12-01
🌱 Replace custom string utility with standard library

fmuyassarov:
Replace `utils.StringInList` with `slices.Contains`. Since Go 1.21, the slices package provides these functions in the standard library, so there is no need for custom wrapper function.

Fixes: https://github.com/metal3-io/baremetal-operator/issues/2818

### DIFF
--- a/internal/controller/metal3.io/baremetalhost_controller.go
+++ b/internal/controller/metal3.io/baremetalhost_controller.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"reflect"
 	"runtime"
+	"slices"
 	"strings"
 	"time"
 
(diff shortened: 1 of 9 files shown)


fmuyassarov:
> /approve Now we just need to agree if this is a breaking change or not. I don't think we have a clear agreement on if the go modules are part of the API or if it is just the YAML that is the API. If go modules are counted, then this is a breaking change since the function was exported (capitalized). It should be trivial for any users to migrate though so I have no concerns removing it. It is more a question if we should put ⚠️ in the PR title. /cc @kashifest @Rozzii

I don’t have a strong opinion on this, but I tend to think it’s fine to treat it as a minor PR since it doesn’t introduce any user-facing changes, only internal adjustments that shouldn’t affect user interactions. Just my two cents.

tuminoid:
> /approve Now we just need to agree if this is a breaking change or not. I don't think we have a clear agreement on if the go modules are part of the API or if it is just the YAML that is the API. If go modules are counted, then this is a breaking change since the function was exported (capitalized). It should be trivial for any users to migrate though so I have no concerns removing it. It is more a question if we should put ⚠️ in the PR title. /cc @kashifest @Rozzii

`pkg/util` is not part of any go module that can be reused externally, it is internal, so not a breaking change despite capital letter.

PR #2827, opened 2025-12-05
:seedling: E2E: Fix race condition in upgrade tests

tuminoid:
After BMO deployment rollout, there's a window where the deployment is marked 'available' but the new pod hasn't acquired the leader lease or started its watches yet. If the test patches the BMH during this window, the controller may miss the update event, causing the BMH to remain stuck in 'available' state instead of transitioning to 'provisioning'.

This fix adds WaitForBmhReconciled() which:
1. Adds a test annotation to the BMH to trigger a reconcile
2. Waits for status.lastUpdated to change, proving the controller processed the update

This ensures the controller is actually processing events before the test makes the critical provisioning patch.

Fixes flakiness in optional upgrade tests where BMH gets stuck in 'available' state, causing cascade failures (MAC conflicts) in subsequent tests.

### DIFF
--- a/test/e2e/common.go
+++ b/test/e2e/common.go
@@ -142,6 +142,40 @@ func PatchBMHForProvisioning(ctx context.Context, input PatchBMHForProvisioningI
 	return helper.Patch(ctx, input.bmh)
 }
 
+// WaitForBmhReconciled waits for the BMO controller to process a BMH update.
+// This is used after BMO deployment rollout to ensure the controller is actually
+// processing events before making further changes. It works by adding/updating
+// an annotation and waiting for the status.lastUpdated timestamp to change,
+// which proves the controller reconciled the BMH.
+func WaitForBmhReconciled(ctx context.Context, c client.Client, bmh metal3api.BareMetalHost, intervals ...interface{}) {
+	// Get the current BMH state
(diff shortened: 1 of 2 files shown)


tuminoid:
docker pull fail... 

/retest

tuminoid:
> docker pull fail... 
> 
> 
> 
> /retest

I think this prow optional is bust or it suffers from CloudFlare outage or something. Need to investigate next week.

lentzi90:
Oh this makes so much sense! Good find!
That docker error is weird though. I see we also may have to teach the fixture provisioner about status.lastUpdated (or use some other method to check that the controller is operational).

tuminoid:
Created issue about the failing prow bmo optional: https://github.com/metal3-io/baremetal-operator/issues/2832

tuminoid:
I'm checking in https://github.com/metal3-io/baremetal-operator/pull/2833 if we can get optional-pull sorted out, some new Kind issue faced in CAPD as well.

tuminoid:
> I'm checking in #2833 if we can get optional-pull sorted out, some new Kind issue faced in CAPD as well.

OK, there is deeper issue with optional pull issue, filed #2834 . Clearly GH workflow based optional is not the same as the Prow pull one. 

/override metal3-bmo-e2e-test-optional-pull
/cc @Rozzii @kashifest

Issue #2832, opened 2025-12-08
bmo-optional-pull fails consistently at docker issues


tuminoid:
### Which jobs are failing?

BMO optional pull.  Note that BMO optional workflow is not, so it looks like setup issue in Prow version.

### Which tests are failing?

```
[2025-12-08T07:38:16.856Z]   [FAILED] Unexpected error:

[2025-12-08T07:38:16.856Z]       <*errors.withStack | 0xc0008280a8>: 

[2025-12-08T07:38:16.856Z]       Failed to load image "quay.io/metal3-io/baremetal-operator:release-0.9" into the kind cluster "bmo-e2e": failed to load image: command "docker exec --privileged -i bmo-e2e-control-plane ctr --namespace=k8s.io images import --all-platforms --digests --snapshotter=overlayfs -" failed with error: exit status 1
```

### Since when has it been failing?

Since last week at minimum, probably long but it hasn't been used really.

### Jenkins link

https://jenkins.nordix.org/view/Metal3/job/metal3-bmo-e2e-test-optional-pull/58/

### Label(s) to be applied

/kind failing-test
One or more /area label. See https://github.com/metal3-io/baremetal-operator/labels for the list of labels.

tuminoid:
/triage accepted

Bot analysis outcome (to be human verified):

```
   === BUILD DEBUG: metal3-bmo-e2e-test-optional-pull #58 ===

   Context:

     - Health: 93.8% failure rate, 15 consecutive failures
     - Started failing: build #54 (Dec 5, 2025)
     - Last success: build #38 (Nov 4, 2025)
     - Build #53 (Nov 12) used CAPI v1.12.0-beta.0 and image loading worked

   Root Cause: CAPI v1.12.0-rc.0 + containerd image import incompatibility

   The failure occurs when loading images into kind cluster during test setup. The error:

     Failed to load image "quay.io/metal3-io/baremetal-operator:release-0.9" into the kind cluster "bmo-e2e": 
     failed to load image: command "docker exec --privileged -i bmo-e2e-control-plane ctr --namespace=k8s.io 
     images import --all-platforms --digests --snapshotter=overlayfs -" failed with error: exit status 1

   Technical Details:

     - PR #2816 merged on Dec 1, 2025 upgraded CAPI from v1.12.0-beta.0 to v1.12.0-rc.0
     - Kind uses ctr images import --all-platforms which expects all platform layers to be present
[...]

PR #2833, opened 2025-12-08
:seedling: e2e: pre-pull images to work around kind ctr import issue

tuminoid:
The clusterctl.ContainerImage struct from CAPI has no yaml struct tags, so when gopkg.in/yaml.v2 unmarshals the e2e config, the LoadBehavior field remains empty. The Defaults() function then sets LoadBehavior to MustLoadImage, causing all image load failures to be fatal instead of the intended tryLoad (warning only) behavior.
     
This adds a local ContainerImage struct with proper yaml tags and a conversion method to use with CAPI's bootstrap framework.

Fixes: #2832

### DIFF
--- a/test/e2e/e2e_config.go
+++ b/test/e2e/e2e_config.go
@@ -33,6 +33,18 @@ const (
 	ironicString string = "ironic"
 )
 
+// ContainerImage describes an image to load into a cluster.
+// This is a copy of clusterctl.ContainerImage with proper yaml tags for YAML unmarshaling.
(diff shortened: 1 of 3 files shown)


tuminoid:
OK, it worked, but failed to other random issue.

tuminoid:
/override metal3-bmo-e2e-test-optional-pull

Prow based BMO optional pull is broken - I think need work to make namespaced secrets work. Log is full of RBAC issues. @Sunnatillo @lentzi90 

This will fix the docker issue though, so this should be merged.

PR #2835, opened 2025-12-08
🌱 lint: allow long lines in tables and code fences

tuminoid:
Allow lines longer than 80 chars, if they're in tables or code fences. In those, long lines cannot often be avoided, and ignoring them one by one is painful and not pretty.

The same change is applied to all repos with markdownlint, but only on main. Documentation is 99% of the time maintained only in main, so it doesn't matter in release branches.


Rozzii (commented):
/lgtm

PR #2836, opened 2025-12-09
:seedling: e2e: retry reconcile loop after patching

tuminoid:
Do another round of fixing into BMO patching for e2e optional.


tuminoid:
BMO optional pull passed, so BMO optional e2e should pass too when merged.

/cc @lentzi90 @kashifest

lentzi90 (commented):
Looks good! I see one potential corner case that we don't handle: webhook is ready but somehow the controller is not. Then it will miss the annotation and get stuck I guess?
Should we rather remove the annotation and patch again on each try make sure?

tuminoid:
> Looks good! I see one potential corner case that we don't handle: webhook is ready but somehow the controller is not. Then it will miss the annotation and get stuck I guess? Should we rather remove the annotation and patch again on each try make sure?

That is practically impossible, but possible in theory, but removing the annotation in e2e  is harmless, possibly just triggering second reconcile in an edge case, so I'm going to implement the suggestion.

tuminoid:
Tests passed.

/unhold

PR #2856, opened 2025-12-15
✨ HostClaim custom resources

pierrecregut:

This is the first PR of a serie implementing HostClaims in the baremetal-operator.

It adds declarations for HostClaims and HostDeployPolicy custom resources and a placeholder for a HostClaim controller. HostDeployPolicies are mostly declarations guiding the selection process of the HostClaim controller and have no associated controller.

related design: https://github.com/metal3-io/metal3-docs/pull/408


apis/metal3.io/v1alpha1/hostdeploypolicy_types.go
```diff
@@ -0,0 +1,74 @@
...
+*/
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// HostDeployPolicySpec defines the desired state of HostDeployPolicy.
+type HostDeployPolicySpec struct {
```
diconico07, line 24: The proposal/design doc has an additional `metadataCopy` field, was this field dropped or will it come in a further iteration ?

apis/metal3.io/v1alpha1/hostclaim_types.go
```diff
@@ -0,0 +1,215 @@
...
+)
+
+const (
+	// Finalizer for HostClaims that are handled by the hostclaim_controller.
+	HostFinalizer = "metal3.io/hostclaim"
+
+	// AssociateCondition documents the status of the association of HostClaim with a BareMetalHost.
+	AssociationCondition = "Association"
+
+	// BareMetalHostAssociatedReason is the reason used when the Metal3Machine is successfully associated with a BareMetalHost.
```
diconico07:
```suggestion
	// BareMetalHostAssociatedReason is the reason used when the HostClaim is successfully associated with a BareMetalHost.
```

diconico07 (commented):
Doing a first pass while going through the design docs and related discussions, so tell me if I misunderstood something.

apis/metal3.io/v1alpha1/hostdeploypolicy_types.go
pierrecregut: This field was dropped because unfortunately an implementation would be too expensive (filtering all annotations and labels on each baremetalhost change with several regular expressions from several HostDeployPolicy). The fact that the filters on the transfered metadata and the filter on the authorized hostclaims may lead the user to believe that they are linked (the transfered metadata from BMH to HardwareData are only the one in the policy that validate the associated hostclaim) which would not be true.

apis/metal3.io/v1alpha1/hostclaim_types.go
pierrecregut: Fixed.

I also removed the definition of the Provisioned condition. The reason is not that it is not needed, but because of github.com/metal3-io/baremetal-operator/issues/2888, it should be defined in the baremetalhost_type file with the available condition (also useful for the capm3 provider).

apis/metal3.io/v1alpha1/hostclaim_types.go
```diff
@@ -0,0 +1,211 @@
...
+	// +optional
+	HostSelector HostSelector `json:"hostSelector,omitempty"`
+
+	// ConsumerRef can be used to store information about something
+	// that is using a host. When it is not empty, the host is
+	// considered "in use". The common use case is a link to a Machine
+	// resource when the host is used by Cluster API.
+	ConsumerRef *corev1.ObjectReference `json:"consumerRef,omitempty"`
+
+	// FailureDomain is the failure domain unique identifier this HostClaim should be attached to.
```
dtantsur: This needs more explanation in the BMO context

apis/metal3.io/v1alpha1/hostclaim_types.go
```diff
@@ -0,0 +1,211 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/selection"
+)
+
+const (
+	// Finalizer for HostClaims that are handled by the hostclaim_controller.
+	HostFinalizer = "metal3.io/hostclaim"
+
+	// AssociateCondition documents the status of the association of HostClaim with a BareMetalHost.
+	AssociationCondition = "Association"
```
dtantsur: Conditions tend to be adjective (is it the right word?), so something like "Associated".

apis/metal3.io/v1alpha1/hostclaim_types.go
```diff
@@ -0,0 +1,211 @@
...
+	// HostPausedReason is a reason used when Host or Cluster is paused.
+	HostPausedReason = "HostPaused"
+	// HostClaimDeletingReason is used while the hostclaim is being deleted.
+	HostClaimDeletingReason = "HostClaimDeleting"
+	// HostClaimDeletionFailedReason is used when the deletion of hostclaim encountered an unexpected error.
+	HostClaimDeletionFailedReason = "HostClaimDeletionFailed"
+
+	// SynchronizationCondition documents the status of the transfer of information from the
+	// HostClaim to the BareMetalHost.
+	SynchronizationCondition = "Synchronization"
```
dtantsur: Similarly, Synchronized or something like that

apis/metal3.io/v1alpha1/hostclaim_types.go
```diff
@@ -0,0 +1,211 @@
...
+	// BareMetalHostNotSynchronizedReason is the reason used when the synchronization of BareMetalHost state
+	// is not successful.
+	BareMetalHostNotSynchronizedReason = "BareMetalHostNotSynchronized"
+)
+
+// HostClaimSpec defines the desired state of HostClaim.
+type HostClaimSpec struct {
+	// Should the compute resource be powered on? Changing this value will trigger
+	// a change in power state of the targeted host.
+	Online bool `json:"online"`
```
dtantsur: In https://github.com/metal3-io/metal3-docs/blob/main/design/baremetal-operator/bmh-v1beta1.md#design-details, we agreed to rename `online` to `poweredOn` on BMH, so let's start with the new name right away.

apis/metal3.io/v1alpha1/hostdeploypolicy_types.go
```diff
@@ -0,0 +1,74 @@
...
+	HostclaimNamespaces *HostclaimNamespaces `json:"hostclaimNamespaces,omitempty"`
+}
+
+type HostclaimNamespaces struct {
```
dtantsur: HostClaim (case)

apis/metal3.io/v1alpha1/hostdeploypolicy_types.go
```diff
@@ -0,0 +1,74 @@
...
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// HostDeployPolicySpec defines the desired state of HostDeployPolicy.
+type HostDeployPolicySpec struct {
+	HostclaimNamespaces *HostclaimNamespaces `json:"hostclaimNamespaces,omitempty"`
```
dtantsur: HostClaim

apis/metal3.io/v1alpha1/hostclaim_types.go
```diff
@@ -0,0 +1,211 @@
...
+	// ramdisk. Most users will want to use "image" instead. Setting this
+	// field triggers provisioning.
+	// +optional
+	CustomDeploy *CustomDeploy `json:"customDeploy,omitempty"`
+
+	// When set to disabled, automated cleaning of host disks will be skipped
+	// during provisioning and deprovisioning.
+	// +kubebuilder:validation:Enum:=metadata;disabled
+	// +optional
+	AutomatedCleaningMode *AutomatedCleaningMode `json:"automatedCleaningMode,omitempty"`
```
dtantsur: Right now, cleaning is not a security feature in Metal3 because Metal3 does not expose Ironic's secure disk removal. Now that we're looking into multi-tenancy, it will become important. Let us add some knob to allow/disallow changing this in the HostClaim. Maybe a new field in the Policy?

apis/metal3.io/v1alpha1/hostclaim_types.go
```diff
@@ -0,0 +1,211 @@
...
+	// +optional
+	LastUpdated *metav1.Time `json:"lastUpdated,omitempty"`
+	// Conditions defines current service state of the HostClaim.
+	// +optional
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+	// HardwareData is a pointer to the name of the bound HardwareData
+	// structure.
+	// +optional
+	HardwareData *HardwareReference `json:"hardwareData,omitempty"`
+	// HostUID is the UID of the underlying compute. It is used
```
dtantsur: s/compute/host/ ?

apis/metal3.io/v1alpha1/hostclaim_types.go
```diff
@@ -0,0 +1,211 @@
...
+	// as ID by the capm3 controller.
```
dtantsur: Is the capm3 reference correct here?

apis/metal3.io/v1alpha1/hostclaim_types.go
pierrecregut: I have added some text. I am not sure it is the right place to go in more details.

apis/metal3.io/v1alpha1/hostclaim_types.go
pierrecregut: done

apis/metal3.io/v1alpha1/hostclaim_types.go
pierrecregut: done

apis/metal3.io/v1alpha1/hostclaim_types.go
pierrecregut: done

apis/metal3.io/v1alpha1/hostdeploypolicy_types.go
pierrecregut: done

apis/metal3.io/v1alpha1/hostdeploypolicy_types.go
pierrecregut: done

apis/metal3.io/v1alpha1/hostclaim_types.go
pierrecregut: This is more tricky. I have added a field. But this is not enough because the HostDeployPolicy is only used when the binding is done. A user could later change the field. To prevent that I have also added a ValidatingAdmissionPolicy that 
prevents setting the field if it was not set or clearing it if it was set.

apis/metal3.io/v1alpha1/hostclaim_types.go
pierrecregut: done

apis/metal3.io/v1alpha1/hostclaim_types.go
pierrecregut: It is really a capm3 thing. I have explicitly said it is the metal3machine controller.

dtantsur: Even so, maybe explain how the HostClaim controller will set it rather than what will use it?

apis/metal3.io/v1alpha1/hostclaim_types.go
dtantsur: Mmmm, fair point. Remind me please: why do we need this field here at all? Is it because of the node reuse?

dtantsur: Where I'm going with this is: if we only need this field for node reuse, maybe we make it explicit: add a boolean `nodeReuse` field that can only be set to `true` initially and is unset by the controller automatically when the time is right? And it will imply disabled cleaning (among potentially other things)?

pierrecregut: Node reuse now is the fact that the hostclaim is kept, the baremetalhost deprovisioned and reprovisioned. We could
disable cleaning at that level. and force cleaning when the hostclaim is really deleted. In that case, we could keep automatedCleaningMode but it would only apply when we deprovision the hostClaim. When we delete it we would always force metadata (fortunately there are only two values now). In that case I remove the field in HostDeployPolicy and the ValidatingAdmissionPolicy. But we have to agree on the implementation of node-reuse (see [discussion](https://github.com/orgs/metal3-io/discussions/2795#discussioncomment-15475232)).

apis/metal3.io/v1alpha1/hostclaim_types.go
pierrecregut: Modified. I hope there is now enough explanation.

apis/metal3.io/v1alpha1/hostclaim_types.go
pierrecregut: So I reverted back the field in HostDeployPolicy and the validatingpolicy. I have added a few words explaining that the cleaning mode is only valid while the HostClaim exists.

dtantsur: > When we delete it we would always force metadata (fortunately there are only two values now).

This can work for the initial implementation, but ideally we need to account for different values on the BMH, including adding a secure one in the future. Hmm. I'm out of good ideas for now, will think about it tomorrow.

pierrecregut: An annotation baremetal.metal3.io/default-automated-cleaning-mode on the BMH with the value that the HostClaim controller would use if it exists. ?

config/base/crds/bases/metal3.io_hostclaims.yaml
```diff
@@ -0,0 +1,358 @@
...
+            description: HostClaimSpec defines the desired state of HostClaim.
+            properties:
+              automatedCleaningMode:
+                allOf:
+                - enum:
+                  - metadata
+                  - disabled
+                - enum:
+                  - metadata
+                  - disabled
```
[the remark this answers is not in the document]
pierrecregut: Removed superfluous kubebuilder constraint on automatedCleaningMode (already set because of the enum type).

apis/metal3.io/v1alpha1/hostclaim_types.go
dtantsur: Let's maybe drop this field for now to unblock this PR? It's easier to add than to remove later.

apis/metal3.io/v1alpha1/hostclaim_types.go
```diff
@@ -0,0 +1,215 @@
...
+	// +optional
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+	// HardwareData is a pointer to the name of the bound HardwareData
+	// structure.
+	// +optional
+	HardwareData *HardwareReference `json:"hardwareData,omitempty"`
+	// HostUID is the UID of the bound BareMetalHost. It is only set
+	// when the BareMetalHost is fully provisioned.
+	// +optional
+	HostUID string `json:"hostUID,omitempty"`
```
dtantsur: Curious: why UID and not a name or everything? UID is probably helpful for CAPM3 but not for regular users? Maybe something like:

```yaml
status:
  host:
    uid: ...
    name: ...
    namespace: ...
```
?

internal/controller/metal3.io/hostclaim_controller.go
```diff
@@ -0,0 +1,63 @@
...
+
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
+// TODO(user): Modify the Reconcile function to compare the state specified by
+// the HostClaim object against the actual cluster state, and then
+// perform operations to make the cluster state reflect the state specified by
+// the user.
+//
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.14.4/pkg/reconcile
```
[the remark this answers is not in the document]
dtantsur: In fact, we should remove the TODO and this comment completely

dtantsur:
I think this is really close, just two comments left from my side.

apis/metal3.io/v1alpha1/hostclaim_types.go
pierrecregut: Clearly this is only useful for CAPM3 and I would rather make it disappear rather than adding other fields on the BMH which should be hidden (but in fact the user has the name through HardwareData).

I think there is a way to do that. The reason for uid is that the node label metal3.io/uuid is set to ds.meta_data.uuid and its value is hardwired in the baremetalhost controller and as hostclaim user including metal3machine controller do not see the BMH, they cannot guess it.  If we accept that the node label content  is different for hostclaim and set to ds.meta_data.metal3_uuid, then we do not need this HostUID field anymore and we can synthesize it in the capm3 provider. We can modify the renderMetadata function in capm3 to add this metal3_uuid field that would typically expand to the hostclaim UID and everybody would be happy. The bonus point is that we can also define it for the non hostclaim scenario and set it to bmh.UID, so that we keep the full compatibility with current behavior and users can also modify their legacy  kubeadmconfig to have something that works with or without hostclaims. 

The logic behind node identification and providerID is still hard to follow for me. Did I miss something (@lentzi90, @honza) ? Is the change acceptable ?

apis/metal3.io/v1alpha1/hostclaim_types.go
pierrecregut: I think there are two distinct issues : 
* do we need to expose an override of the cleaning mode on hostclaims
  * If not we force the cleaning mode to be disabled as long as the hostclaim exists even when we deprovision it
  * we could limit it to a boolean, but the question is why ? what does it change to not authorize a different more efficient cleaning mode for a use case ?
  * we keep the override to be the full enumeration
 * how do we implement it ? Here I agree with you  we need more eyes to take a decision. But this has no impact on the HostClaim API. If fields or labels must be added, it is on the BMH.

diconico07: To me having the cleaning mode to disabled as long as the HostClaim exists makes sense if we take it from a similar sense of PV/PVC. There to trigger a cleaning phase, one has to delete and re-create the HostClaim.

pierrecregut: Ok. Let us remove the field. The end-user or the m3 machine will not have control over the cleaning. It will be disabled by default in the hostclaim operator and restored (to metadata unless we add something on the bmh)  when the hostclaim is deleted.

apis/metal3.io/v1alpha1/hostclaim_types.go
pierrecregut: From my experiments the metal3_uuid approach is working (in fact a question is can we even make it simpler by using the m3machine uid ?). So let us remove this field too.

internal/controller/metal3.io/hostclaim_controller.go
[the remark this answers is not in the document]
pierrecregut: done

apis/metal3.io/v1alpha1/hostclaim_types.go
pierrecregut: But this means that we will ignore AutomatedCleaningMode specified in the metal3 machine with hostclaims. Which probably makes sense unless node-reuse is specified.

dtantsur: > AutomatedCleaningMode specified in the metal3 machine

We have this? Ugh. I have the same arguments against having it on Metal3Machine as on HostClaims. I really wonder if the folks will accept deprecating it in favour of a field with a clearer semantics (e.g. `reuseNode`).

dtantsur: I guess we should just continue in https://github.com/orgs/metal3-io/discussions/2795#discussioncomment-15531386

dtantsur:
/lgtm

I think this is a safe start, and we can build more on top of it later. Thanks!

lentzi90:
Just noting that this is on my radar and I hope to review it soon. It is just hard to find the time :slightly_frowning_face:

apis/metal3.io/v1alpha1/hostclaim_types.go
```diff
@@ -0,0 +1,208 @@
...
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/selection"
+)
+
+const (
+	// Finalizer for HostClaims that are handled by the hostclaim_controller.
+	HostFinalizer = "metal3.io/hostclaim"
```
elfosardo: nit: HostClaimFinalizer ?

pierrecregut: done

pierrecregut:
> I guess it's too early to add tests and docs, right?

So far there is only a data type so there are no tests. For the user documentation, there is a proposal here: https://github.com/metal3-io/metal3-docs/pull/641

apis/metal3.io/v1alpha1/hostclaim_types.go
```diff
@@ -0,0 +1,208 @@
...
+	// This is used to limit the set of BareMetalHost objects considered for
+	// claiming for a HostClaim.
+	// +optional
+	HostSelector HostSelector `json:"hostSelector,omitempty"`
+
+	// ConsumerRef can be used to store information about something
+	// that is using a host. When it is not empty, the host is
+	// considered "in use". The common use case is a link to a Machine
+	// resource when the host is used by Cluster API.
+	ConsumerRef *corev1.ObjectReference `json:"consumerRef,omitempty"`
```
lentzi90, line 112: Do we need the `ConsumerRef` also on the `HostClaim`? I was thinking that we could use owner references for that.

apis/metal3.io/v1alpha1/hostclaim_types.go
```diff
@@ -0,0 +1,208 @@
...
+
+// SetConditions sets conditions for an API object.
+func (h *HostClaim) SetConditions(conditions []metav1.Condition) {
+	h.Status.Conditions = conditions
+}
+
+type HardwareReference struct {
+	// `namespace` is the namespace of the HardwareDetail bound
+	Namespace string `json:"namespace"`
+	// `name` is the name of the HardwareDetail bound
```
lentzi90: nit:

```suggestion
	// `namespace` is the namespace of the HardwareData bound
	Namespace string `json:"namespace"`
	// `name` is the name of the HardwareData bound
```

apis/metal3.io/v1alpha1/hostclaim_types.go
pierrecregut: When we use the identityRef to reference a remote hostClaim, this claim cannot use an ownership reference. In fact the consumerRef is not perfect, we need that the remote namespace is only used by one client cluster because there is no indication of which remote cluster is valid.

apis/metal3.io/v1alpha1/hostclaim_types.go
pierrecregut: done

apis/metal3.io/v1alpha1/hostclaim_types.go
pierrecregut: On the [prototype hostclaim branch of capm3](https://github.com/pierrecregut/cluster-api-provider-metal3/tree/hostclaim), see the change in the [metal3machine controller](https://github.com/pierrecregut/cluster-api-provider-metal3/blob/hostclaim/controllers/metal3machine_controller.go#L470-L474) and the [remote context cache](https://github.com/pierrecregut/cluster-api-provider-metal3/blob/hostclaim/baremetal/remote_context.go) for the planned use.

PR #2859, opened 2025-12-16
:bug: Require PreprovisioningImage for deprovisioning when cleaning is enabled

elfosardo:
When deprovisioning a host with automated cleaning enabled, BMO now checks if a PreprovisioningImage is available before proceeding. Previously, the StateDeprovisioning case was not handled, causing Ironic to fall back to building its own ISO and fail with a misleading "uefi_esp.img not found" error when the PPI was unavailable.

Assisted-By: Claude Sonnet 4.5 (commercial license)

### DIFF
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -360,7 +360,8 @@ func (p *ironicProvisioner) configureNode(data provisioner.ManagementAccessData,
 	}
 
 	switch data.State {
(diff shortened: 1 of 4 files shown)


pkg/provisioner/ironic/ironic.go
dtantsur, line 365: Just add the new state here

lentzi90 (commented):
/lgtm

PR #2889, opened 2025-12-22
🌱 Refactor: move HardwareDetails type to HardwareData

dtantsur:
Both to prepare for v1beta1 and to reduce the size of the source file.


lentzi90:
Probably some temporary hickup. Periodics are green, and I don't see how installing go could have been broken by this PR.
/retest

kashifest:
lgtm, thinking if this would be 🌱 or ✨ since we are touching apis.

dtantsur:
> thinking if this would be 🌱 or ✨ since we are touching apis.

Moving structure between files in the same modules should not affect how they're used from outside, I think.

lentzi90 (commented):
/approve

PR #2890, opened 2025-12-30
:bug: Fix errors.As usage for NoDataInSecretError

elfosardo:
The errors.As calls were using new(*NoDataInSecretError) which creates a **NoDataInSecretError, but the error is a value type. This prevented the error from being matched, causing PreprovisioningNetworkData to return an error instead of empty data when the networkData key is missing from a secret.


lentzi90 (commented):
/lgtm

kashifest (commented):
/approve

PR #2901, opened 2026-01-12
🌱 Support building multi-arch container images

### LINKED ISSUE #1715: Support building for multiple architectures
**User Story**

As a developer/user/operator I would like to use pre-built BMO images on multiple architectures because it is convenient (compared to building everything yourself).

**Detailed Description**
We currently only publish amd64 container images, but there are other popular architectures. Go has good support for building for other architectures also, so there is no need for special hardware for testing this.
I suggest we look at how CAPI/CAPM3 has structured the Makefile with a `docker-build-all` target for building all supported architectures. This should be simple to implement for us as well.

Once we can build these images, it should be fairly trivial to also make CI build and publish them together with a multi-arch manifest for easy consumption.

/kind feature

vishalanarase:

This PR adds support for building container images for multiple architectures (amd64 and arm64)

##### Dockerfile
- Updated to accept `ARCH` build argument for cross-compilation
- Sets `GOARCH` and `GOOS` appropriately for each target architecture

##### Makefile
- Added `docker-build-all` target to build images for all supported architectures sequentially
- Updated `docker-build` to tag amd64 images with base image name for backward compatibility
- Supports architectures: `amd64`, `arm64`

Fixes #1715

**Checklist:**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -92,7 +92,7 @@ IMG_NAME ?= baremetal-operator
 IMG_TAG ?= latest
 IMG ?= $(REGISTRY)/$(IMG_NAME)
 ARCH ?= $(shell go env GOARCH)
-ALL_ARCH = amd64 arm arm64 ppc64le s390x
+ALL_ARCH = amd64 arm64
 
 # Which configuration to use when deploying (from the config directory)
 DEPLOY_CONFIG ?= default
--- a/Makefile
+++ b/Makefile
@@ -91,6 +91,8 @@ REGISTRY ?= quay.io/metal3-io
 IMG_NAME ?= baremetal-operator
 IMG_TAG ?= latest
 IMG ?= $(REGISTRY)/$(IMG_NAME)
+ARCH ?= $(shell go env GOARCH)
+ALL_ARCH = amd64 arm arm64 ppc64le s390x
 
 # Which configuration to use when deploying (from the config directory)
 DEPLOY_CONFIG ?= default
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -269,22 +271,55 @@ generate: $(CONTROLLER_GEN) ## Generate code
...
+## Docker — All ARCH
+## --------------------------------------
+
+.PHONY: docker-build-all ## Build all the architecture docker images
+docker-build-all: $(addprefix docker-build-,$(ALL_ARCH))
+
+docker-build-%:
+	$(MAKE) ARCH=$* docker-build
+
+.PHONY: docker-push-all ## Push all the architecture docker images
```
tuminoid: I don't think we use docker push targets for anything, so they can be left out.

tuminoid (commented):
/ok-to-test

Makefile
vishalanarase: Removed the push targets

Makefile
```diff
@@ -91,6 +91,8 @@ REGISTRY ?= quay.io/metal3-io
 IMG_NAME ?= baremetal-operator
 IMG_TAG ?= latest
 IMG ?= $(REGISTRY)/$(IMG_NAME)
+ARCH ?= $(shell go env GOARCH)
+ALL_ARCH = amd64 arm arm64 ppc64le s390x
```
dtantsur: We don't currently support anything except for amd64 and arm64, let's not pretend that we do.

vishalanarase: Done, Set ALL_ARCH to amd64/arm64

tuminoid:
We probably need container build test for all architectures now. BMO e2e builds only the one we're testing for so arm builds are not tested. Similarly, our release builds wouldn't built arm now.
/hold
until we do. The build support should probably be added in another PR for amd64 and in this one extended for arm.

Can you also update the PR description since its no longer about building and pushing, and arch list is limited? Thanks.

vishalanarase:
> The build support should probably be added in another PR for amd64

Created new PR https://github.com/metal3-io/baremetal-operator/pull/2908

vishalanarase:
Closing in favour of https://github.com/metal3-io/baremetal-operator/pull/2908
/close

PR #2902, opened 2026-01-13
🌱 Log when cannot detach immediately

dtantsur:
Otherwise, the users are left wondering why detachment does not work.

Honestly, maybe we should also fire an event in this case? Comments welcome.


dtantsur:
Since I'll be quite busy in the near future, I've moved the events topic to https://github.com/metal3-io/baremetal-operator/issues/2907

/hold cancel

PR #2908, opened 2026-01-14
🌱 Support building multi-arch container images for amd64

### LINKED ISSUE #1715: Support building for multiple architectures

vishalanarase:

This PR adds support for building container images with architecture-specific
tags, starting with amd64 support as the foundation for multi-arch builds.

Changes Made:
- Dockerfile: Accepts ARCH build argument for cross-compilation
- Makefile: Adds docker-build-all target and ARCH variable
- Maintains backward compatibility: base image name points to amd64
- Updates docker-push to also push base image tag for backward compatibility

Fixes #1715

**Checklist:**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -273,7 +273,7 @@ generate: $(CONTROLLER_GEN) ## Generate code
 .PHONY: docker
 docker: docker-build ## Alias for docker-build (for backwards compatibility)
 docker-build: generate manifests ## Build the docker image for controller-manager
-	docker build \
+	docker build --platform=linux/$(ARCH) \
 	--build-arg ARCH=$(ARCH) \
 	--build-arg http_proxy=$(http_proxy) \
 	--build-arg https_proxy=$(https_proxy) \
@@ -285,7 +285,7 @@ docker-build: generate manifests ## Build the docker image for controller-manage
 
 .PHONY: docker-debug
 docker-debug: generate manifests ## Build the docker image with debug info
-	docker build \
+	docker build --platform=linux/$(ARCH) \
 	--build-arg ARCH=$(ARCH) \
 	--build-arg http_proxy=$(http_proxy) \
 	--build-arg https_proxy=$(https_proxy) \
@@ -295,14 +295,18 @@ docker-debug: generate manifests ## Build the docker image with debug info
 # Push the docker image
 .PHONY: docker-push
 docker-push:
-	docker push ${IMG}:${IMG_TAG}
+	docker push ${IMG}-$(ARCH):${IMG_TAG}
+	@# Push base image tag for backward compatibility (amd64 is the default)
+	@if [ "$(ARCH)" = "amd64" ]; then \
+		docker push ${IMG}:${IMG_TAG}; \
+	fi
 
 ## --------------------------------------
 ## Docker — All ARCH
 ## --------------------------------------
 
-.PHONY: docker-build-all ## Build all the architecture docker images
-docker-build-all: $(addprefix docker-build-,$(ALL_ARCH))
+.PHONY: docker-build-all
+docker-build-all: $(addprefix docker-build-,$(ALL_ARCH)) ## Build all the architecture docker images
 
 docker-build-%:
 	$(MAKE) ARCH=$* docker-build


Makefile
```diff
@@ -269,22 +271,45 @@ generate: $(CONTROLLER_GEN) ## Generate code
...
 	--build-arg https_proxy=$(https_proxy) \
-	--build-arg LDFLAGS="-extldflags=-static"
+	--build-arg LDFLAGS="-extldflags=-static" \
+	. -t ${IMG}-$(ARCH):${IMG_TAG}
 
 # Push the docker image
 .PHONY: docker-push
 docker-push:
-	docker push ${IMG}:${IMG_TAG}
+	docker push ${IMG}-$(ARCH):${IMG_TAG}
```
tuminoid: Docker Push isn't in any repo, so you may drop these parts.

vishalanarase: Done

tuminoid (commented):
Since we're building foundations here to adding more archs, I think we also need to add the workflow to build amd64, similar to https://github.com/metal3-io/ironic-image/blob/main/.github/workflows/ironic-image-build-pr.yml 
which we can then extend with other archs later.

Currently amd64  build is tested due it being built for e2e testing, but arm64 doesn't have e2e (at least yet), so explicit build check would be nice to make sure it doens't break.

vishalanarase:
> Since we're building foundations here to adding more archs, I think we also need to add the workflow to build amd64, similar to https://github.com/metal3-io/ironic-image/blob/main/.github/workflows/ironic-image-build-pr.yml which we can then extend with other archs later.
> 
> Currently amd64 build is tested due it being built for e2e testing, but arm64 doesn't have e2e (at least yet), so explicit build check would be nice to make sure it doens't break.

@tuminoid Updated PR

.github/workflows/build-pr.yml
```diff
@@ -0,0 +1,48 @@
...
+      id: vars
+      run: echo "go_version=$(make go-version)" >> $GITHUB_OUTPUT
+
+    - name: Set up Go
+      uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5.6.0
+      with:
+        go-version: ${{ steps.vars.outputs.go_version }}
+
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v3
```
tuminoid: This needs to be pinned.

.github/workflows/build-pr.yml
```diff
@@ -0,0 +1,48 @@
...
+    paths:
+    - 'Dockerfile'
+    - 'Makefile'
+    - '.github/workflows/build-pr.yml'
+
+permissions:
+  contents: read
+
+jobs:
+  build-amd64:
```
tuminoid: I'd love this to be defined as matrix, with amd64 being the only option. Easier to expand later.

Makefile
```diff
@@ -269,23 +271,42 @@ generate: $(CONTROLLER_GEN) ## Generate code
 ## --------------------------------------
 
 .PHONY: docker
-docker: generate manifests ## Build the docker image
-	docker build . -t ${IMG}:${IMG_TAG} \
+docker: docker-build ## Alias for docker-build (for backwards compatibility)
+docker-build: generate manifests ## Build the docker image for controller-manager
+	docker build --network=host --pull \
```
tuminoid: Do we really need `--network=host` and `--pull` ?

.github/workflows/build-pr.yml
tuminoid: Also, we don't use buildx?

tuminoid (commented):
Some comments and suggestions.

.github/workflows/build-pr.yml
```diff
@@ -0,0 +1,51 @@
...
+on:
+  pull_request:
+    types: [opened, edited, reopened, synchronize, ready_for_review]
+    branches:
+    - 'main'
+    - 'release-*'
+    paths:
+    - 'Dockerfile'
+    - 'Makefile'
+    - '.github/workflows/build-pr.yml'
```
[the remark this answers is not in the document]
tuminoid: Go home Copilot, you're drunk.

.github/workflows/build-pr.yml
```diff
@@ -0,0 +1,51 @@
...
+        IMG_NAME: baremetal-operator
+        IMG_TAG: pr-test
+
+    - name: Verify image was built
+      run: |
+        docker images | grep baremetal-operator
+        docker image inspect quay.io/metal3-io/baremetal-operator-${{ matrix.arch }}:pr-test >/dev/null
+        if [ "${{ matrix.arch }}" = "amd64" ]; then
+          docker image inspect quay.io/metal3-io/baremetal-operator:pr-test >/dev/null
+        fi
```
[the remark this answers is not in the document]
tuminoid: Sigh.

Makefile
```diff
@@ -269,23 +271,42 @@ generate: $(CONTROLLER_GEN) ## Generate code
...
 .PHONY: docker-push
 docker-push:
 	docker push ${IMG}:${IMG_TAG}
 
+## --------------------------------------
+## Docker — All ARCH
+## --------------------------------------
+
+.PHONY: docker-build-all ## Build all the architecture docker images
+docker-build-all: $(addprefix docker-build-,$(ALL_ARCH))
```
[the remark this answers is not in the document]
tuminoid: This is valid.

Rozzii:
Thanks for your contribution @vishalanarase

tuminoid (commented):
/lgtm

Thanks a lot!

PR #2928, opened 2026-01-22
:bug: Fix deletion hang when aborting inspection during power-off

elfosardo:
When deleting a BMH during inspection, power-off could hang because:

1. After abort, node is in InspectFail with TargetProvisionState still set, blocking power changes
2. Ironic does not auto-clear TargetProvisionState in InspectFail state
3. PowerOff() misinterpreted the abort message in LastError as a power-off failure

Fix by:
- Explicitly transitioning from InspectFail to manageable to clear TargetProvisionState and allow power changes
- Ignoring "aborted" errors in LastError during power-off


elfosardo:
this has been tested with virtual and real bare metal deployment, and verified in downstream CI https://github.com/openshift/baremetal-operator/pull/447

Issue #2888, opened 2025-12-22
[RFE] Add Conditions to BareMetalHost


dtantsur:
This is a continuation of #25 as well as recent community discussion.

# User story

We have a very sophisticated Status for hosts, which has its up- and downsides. The downside is that understanding various status fields requires careful reading of the (sometimes scarce) documentation, and that standard tools like `kubectl wait` are tricky to use with BareMetalHosts.

This RFE proposes a simple set of conditions that provides operators with an overview of what is going on with the host.

# Proposed conditions

## Manageable

Rule: `Status.ProvisioningState not in ["", "registering", "unmanaged"] and IronicNode.Fault != "power failure"`.

Reasons for True: `Manageable`

Reasons for False: `Registering`, `MissingCredentials`,  `RegistrationFailed` (covers any cases not covered by the previous two), `PowerFailure` (covers loosing BMC access for already verified nodes).

## Available

Rule: `Condition:Manageable and Status.ProvisioningState == "available"`. Justification: can be used to track registration and inspection together.

Reasons for True: `Available`

Reasons for False: `NotManageable`, `Inspecting`, `NotAvailable` (covers all remaining cases).

## Provisioned

Rule: `Status.ProvisioningState == "provisioned"`. Reasoning: can be used to track the successful provisioning process.

Reasons for True: `Provisioned`

Reasons for False: `Provisioning` (for in-progress state), `NotProvisioned` (any other case).

## Ready

Rule: `Condition:Provisioned && Status.OperationalStatus in ["OK", "detached"]`. Justification: can be used to track servicing and the overall host readiness. Seems to be pretty standard in Kubernetes.

Reasons for True: `Provisioned`.

Reasons for False: `Error`, `Servicing` (based on `OperationalStatus`), `Provisioning` (for in-progress state except for servicing), `NotProvisioned` (any other case).

## Progressing

Rule: `Status.ProvisioningState in ["registering", "inspecting", "preparing", "provisioning", "deprovisioning", "powering off before delete", "deleting"]`. Justification: tracks running operations, seems to be pretty standard in Kubernetes.

Reasons for True: a CamelCase version of the corresponding `ProvisioningState` (e.g. `Preparing`, `PoweringOffBeforeDelete`).

Reasons for False: `NotProgressing`.

## Healthy

Credit for this idea goes to @rhjanders and his health rollup work. This is one way to implement it in the BMH. The Ironic side is WIP, so this is a draft.

[...]

PR #2945, opened 2026-01-28
:sparkles: Conditions implementation for BareMetalHosts

pierrecregut:
Five conditions are defined and handled: Manageable, Available, Provisioned, Ready and Progressing. They provide a synthetic view of the state of the BareMetalHost.

Partially implements #2888.


apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -710,6 +780,10 @@ type BareMetalHostStatus struct {
 	// ErrorCount records how many times the host has encoutered an error since the last successful operation
 	// +kubebuilder:default:=0
 	ErrorCount int `json:"errorCount"`
+
+	// Conditions defines current service state of the BareMetalHost.
+	// +optional
```
dtantsur, line 786: I sometimes see tags like https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/api/v1beta2/metal3machine_types.go#L264-L266 on conditions, maybe we need them here too

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -764,6 +838,16 @@ type BareMetalHost struct {
 	Status BareMetalHostStatus `json:"status,omitempty"`
 }
 
+// GetConditions returns the set of conditions for this object.
+func (host *BareMetalHost) GetConditions() []metav1.Condition {
```
dtantsur, line 846: These methods look trivial, are they here to satisfy some interface?

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -2164,6 +2166,81 @@ func (r *BareMetalHostReconciler) getHostFirmwareComponents(info *reconcileInfo)
...
+func setConditionTrue(host *metal3api.BareMetalHost, typ, reason string) {
+	conditions.Set(host, metav1.Condition{Type: typ, Status: metav1.ConditionTrue, Reason: reason})
+}
+
+func setConditionFalse(host *metal3api.BareMetalHost, typ, reason string) {
+	conditions.Set(host, metav1.Condition{Type: typ, Status: metav1.ConditionFalse, Reason: reason})
+}
+
+func setConditionsProgressing(host *metal3api.BareMetalHost, progressingReason string) {
+	setConditionTrue(host, metal3api.ManageableCondition, metal3api.ManageableReason)
```
dtantsur, line 2178: It's technically possible for a host to be progressing and experience a power fault.

dtantsur: Ah, I see that you're handling it separately in the end.

dtantsur:
/approve

Thanks! Adding more people because of the API impact:

/cc @Rozzii @lentzi90 @kashifest @elfosardo

apis/metal3.io/v1alpha1/baremetalhost_types.go
pierrecregut: Yes cluster-api conditions Getter and Setter.

apis/metal3.io/v1alpha1/baremetalhost_types.go
pierrecregut: Yes indeed you are right to get strategic merge patch to work as expected.

pierrecregut: done

Rozzii (commented):
/hold
/lgtm
Letting others take a look too @Sunnatillo @adilGhaffarDev @lentzi90 @kashifest

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -2164,6 +2166,81 @@ func (r *BareMetalHostReconciler) getHostFirmwareComponents(info *reconcileInfo)
...
+		setConditionFalse(host, metal3api.ProvisionedCondition, metal3api.NotProvisionedReason)
+		setConditionFalse(host, metal3api.ReadyCondition, metal3api.NotProvisionedReason)
+		setConditionFalse(host, metal3api.ProgressingCondition, metal3api.NotProgressingReason)
+	case metal3api.StateProvisioning:
+		setConditionsProgressing(host, metal3api.ProvisioningReason)
+	case metal3api.StateProvisioned, metal3api.StateExternallyProvisioned:
+		setConditionTrue(host, metal3api.ManageableCondition, metal3api.ManageableReason)
+		setConditionFalse(host, metal3api.AvailableCondition, metal3api.NotAvailableReason)
+		setConditionTrue(host, metal3api.ProvisionedCondition, metal3api.ProvisionedReason)
+		setConditionTrue(host, metal3api.ReadyCondition, metal3api.ProvisionedReason)
```
adilGhaffarDev, line 2217: I would suggest we follow k8s api conventions, and for ready/true and ready/false we show the summary of conditions instead of setting it explicitly.

adilGhaffarDev (commented):
Its a good addition, thank you @pierrecregut 
We should bubble up these in capm3, I have opened an issue and tagged you there. Perhaps in the next release cycle we can work on it.

internal/controller/metal3.io/baremetalhost_controller.go
pierrecregut: The problem is that Ready as defined in #2888 is just Provisioned + operationalStatus ok and we did not introduce another condition just for this last predicate (We could but this would be verbose and it is not clear that this new condition would be helpful per se). In most cluster-api cases, Ready is the conjunction of a long list of independent conditions whereas here conditions are almost exclusive because they correspond to distinct underlying states.  @dtantsur what do you think about this ?

dtantsur: I'm sorry, I don't quite understand what @adilGhaffarDev is requesting. Could you link to the exact convention you're referring to or quote it please?

adilGhaffarDev: `
When designing Conditions for a resource, it's helpful to have a common top-level condition which summarizes more detailed conditions. Simple consumers may simply query the top-level condition. Although they are not a consistent standard, the Ready and Succeeded condition types may be used by API designers for long-running and bounded-execution objects, respectively.
`
ref: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties

It does not have to summary of all conditions, it can be of few. Why dont we add another condition for operationalStatus? I think it is a good candidate for condition. 
And Ready condition should be summary of operationalStatusCondition and ProvisionedCondition. 
You can check how we use SetSummaryCondition() function in capm3 and capi for this in the defer of reconciles.

Also in future we might add more conditions that will effect readiness I think, for example I see `Healthy` condition that can be added in future https://github.com/metal3-io/baremetal-operator/issues/2888 .

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -139,6 +139,76 @@ const (
...
+	// operational status.
+	ReadyCondition = "Ready"
+	// ErrorCondition is the reason used when the operational state is in error.
+	ErrorReason = "Error"
+	// DetachedReason  the reason used when the host is detached.
+	DetachedReason = "Detached"
+	// ServicingReason is the reason used when the host is servicing.
+	ServicingReason = "Servicing"
+	// AvailableCondition documents the availability of the BareMetalHost toward the Available state.
+	AvailableCondition = "Available"
```
adilGhaffarDev: I would highly suggest that we change this to something else. Maybe as following
```suggestion
	AvailableForProvisioningCondition = "AvailableForProvisioning"
```
In k8s usually available means resource is ready for some amount of time. I would suggest avoid using just "Available".

apis/metal3.io/v1alpha1/baremetalhost_types.go
pierrecregut: In fact is is probably also missing:
// +patchStrategy=merge
// +patchMergeKey=type

internal/controller/metal3.io/baremetalhost_controller.go
pierrecregut: This will be a cluttered copy of OperationalStatus field with reason being the capitalized version of the OperationalStatus value. OperationalStatus is already some kind of mix with Manageable. Conditions are great when there are not too many of them and they are simple to understand. In fact, I am not very convinced by the Ready condition in the BMH context. May be we should remove it until its meaning is clear..

apis/metal3.io/v1alpha1/baremetalhost_types.go
pierrecregut: In fact they are not:: although those fields are mentioned in k8s documentation, they have no impact on what controller-gen generates.

internal/controller/metal3.io/baremetalhost_controller.go
dtantsur: Ready is handy as a one-stop-shop for waiting for BMH to stabilize. I don't see a point in a condition for OperationalStatus, especially since you can already wait for any values of OperationalStatus.

Also, "it's helpful" cannot be called a convention, it's just a tip. I think the important part of it is "you should have a simple condition for people to wait on" (which is what we do here) rather than "any compound condition must be a summary of other conditions".

dtantsur: I'll accept dropping the Ready condition if we cannot agree on what it means, even though it will be quite a loss of value for users.

adilGhaffarDev: Its fine to keep how it is if does not make sense in BMO workflow.

apis/metal3.io/v1alpha1/baremetalhost_types.go
pierrecregut: Done... even if I am not really convinced (minReadiness for deployment is just  a "hack" to capture crash looping pods that could be taken for available).

internal/controller/metal3.io/baremetalhost_controller.go
pierrecregut: Ready kept as defined.

adilGhaffarDev:
LGTM

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -139,6 +139,77 @@ const (
...
+	ManageableReason = "Manageable"
+	// NotManagedReason is the reason used when the BareMetalHost is not managed.
+	NotManagedReason = "NotManaged"
+	// RegisteringReason is the reason used for the ProgressingCondition when
+	// the BareMetalHost is registering.
+	RegisteringReason = "Registering"
+	// RegistrationFailedReason is the reason used when the BareMetalHost is not
+	// registered.
+	RegistrationFailedReason = "RegistrationFailed"
+	// PowerFailure is the reason used when the BareMetalHost is experiencing a
```
kashifest:
```suggestion
	// PowerFailureReason is the reason used when the BareMetalHost is experiencing a
```

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -139,6 +139,77 @@ const (
...
+	// power failure.
+	PowerFailureReason = "PowerFailure"
+
+	// ProvisionedCondition documents the provisioning state of the BareMetalHost
+	// toward the Provisioned goal.
+	ProvisionedCondition = "Provisioned"
+	// ProvisionedReason is the reason used by the Provisioned and the Ready conditions
+	// when the BareMetalHost is provisioned.
+	ProvisionedReason = "Provisioned"
+	// ProvisionedReason is the reason used when the BareMetalHost is not provisioned and not provisioning.
```
kashifest:
```suggestion
	// NotProvisionedReason is the reason used when the BareMetalHost is not provisioned and not provisioning.
```

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -139,6 +139,77 @@ const (
...
+	NotProvisionedReason = "NotProvisioned"
+	// ProvisionedReason is the reason used when the BareMetalHost is provisioning.
```
kashifest:
```suggestion
	// ProvisioningReason is the reason used when the BareMetalHost is provisioning.
```

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -139,6 +139,77 @@ const (
...
+	ProvisioningReason = "Provisioning"
+
+	// ReadyCondition documents the fact that the BareMetalHost is provisioned and in a good
+	// operational status.
+	ReadyCondition = "Ready"
+	// ErrorCondition is the reason used when the operational state is in error.
```
kashifest:
```suggestion
	// ErrorReason is the reason used when the operational state is in error.
```

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -139,6 +139,77 @@ const (
...
+	// the BareMetalHost is registering.
+	PreparingReason = "Preparing"
+	// DeprovisioningReason is the reason used for the ProgressingCondition when
+	// the BareMetalHost is registering.
+	DeprovisioningReason = "Deprovisioning"
+	// PoweringOffBeforeDeleteReason is the reason used for the ProgressingCondition when
+	// the BareMetalHost is registering.
+	PoweringOffBeforeDeleteReason = "PoweringOffBeforeDelete"
+	// DeletingReason is the reason used for the ProgressingCondition when
+	// the BareMetalHost is registering.
```
kashifest: Is this statement correct? `DeletingReason` when BMH is registering?

pierrecregut: Fixed this copy/paste with two others just above in the last version. The other typos have been fixed too.

kashifest:
Ow my bad, forgot that prow doesnt wait for optional test . Anyways, hopefully this wont break any upgrade.

Issue #2927, opened 2026-01-22
E2E: Logs are duplicated for containers


lentzi90:
As mentioned in https://github.com/metal3-io/baremetal-operator/pull/2910#issuecomment-3759959945, the container logs are duplicated in our E2E tests. I think there are three issues actually:

1. The framework we use will "restart" collection if there are any disruptions (e.g. a network issue). On each restart, it collects the logs from the beginning. This means that each container log file can have the messages duplicated multiple times, starting again from the beginning for each such disruption.
2. We sometimes retry deployments multiple times and add a log watcher each time. There is no built-in deduplication so each watcher will add more logs.
3. In upgrade tests, we use the same cluster, the same namespace and the same deployment name for each upgrade. This means that the log watcher we create in the first test will match on the pods both before and after the upgrade, it will even match on the next test. For each time we deploy or upgrade we add one more watcher.

lentzi90:
One more discovery. The code for dumping resources is full of `Expect`. As soon as there is an issue, the AfterEach fails and skips the rest of the collection. We need to fix this. Start from https://github.com/metal3-io/baremetal-operator/blob/ffa8ad1d3767691ae2966a80b96da6484b0386a7/test/e2e/common.go#L773

lentzi90:
Not really related, but discovered during the work in #2947. We do not need the "extra port mappings" anymore. Now that we have proper veth for communication between the VMs and kind, they do not fill any function but we still have them in the config for example here: https://github.com/metal3-io/baremetal-operator/blob/main/config/overlays/e2e-release-0.11/ironic.env.
This is something I intend to clean up.

lentzi90:
I think we are now as far as we can get here without implementing our own log collection. #3024 takes care of the last duplication that is caused by "us".

Unfortunately there is still a *lot* of duplication and I think I now have an explanation. It is point 1 in the description (CAPI framework restarting collection from beginning all the time), but not due to network issues or container restarts. It just doesn't have the code to handle init containers properly! CAPI has no init containers, and didn't even collect them until I added that in https://github.com/kubernetes-sigs/cluster-api/pull/12208, so they have had no reason to improve on it.

The thing is, the log watchers will keep trying to get logs *as long as the pod is there*. I think this causes init container logs to be duplicated every 2 seconds as it tries again and again, because the pod is still there. We also get a few duplications from when the pods terminate. This can take some time for Ironic and during that time the log watcher may end up trying a few times.

[...]

PR #2947, opened 2026-01-29
🌱 E2E: Use per-test cluster for upgrades

lentzi90:

Avoid cleanup issues by using a fresh cluster for each upgrade test.
We create a separate kind cluster for each upgrade test. These "upgrade
clusters" use a separate provisioning IP than the normal kind cluster in
order to be able to co-exist without issues. However, all the upgrade
tests use the same IP, so they must run one at a time.
    
Put all logs directly in the logs folder instead of under init-bmo,
init-ironic or similar. This is because the log watchers still pick up
the "other" deployments so we get confusing duplicates. Better then to
have all in the same folder so it is obvious one needs to check which is
which. We could improve this later by canceling log watches explicitly
when they are no longer needed.

This is part of #2927. The upgrade tests use the same namespace and name for ironic and BMO, which means that we easily confuse the logs when we run them all in the same cluster. Using separate clusters also helps us avoid issues where cleanup fails and then affects the rest of the tests.

**Checklist:**


lentzi90:
/hold
This breaks the fixture test, because there we do run the upgrade tests together with other tests. I'll try to change the ports for upgrade tests so that they can co-exist.

lentzi90:
Currently blocked by https://github.com/metal3-io/ironic-standalone-operator/issues/524.
I want to use custom ports so that the upgrade tests could have their own kind clusters and run parallel with the other tests.

Initially I just deleted the "normal" kind cluster when the upgrade tests started. This works for our proper e2e tests because the upgrade tests run in a separate job. However, the fixture tests run all tests together, so we cannot do this for them. We could of course split them up also, but I would rather fix the custom ports and get a proper solution in that way.

lentzi90:
Trying IRSO main that has the fix for the ports. I suspect this will still fail as we test Ironic versions that are not supported by IRSO main. But I want to confirm if the ports works

lentzi90:
/cc @adilGhaffarDev @kashifest 
/hold cancel
This is ready now

test/e2e/upgrade_test.go
```diff
@@ -297,10 +303,25 @@ var _ = Describe("Upgrade", Label("optional", "upgrade"), func() {
 	)
 
 	AfterEach(func() {
-		DumpResources(ctx, e2eConfig, upgradeClusterProxy, testArtifactFolder)
+		upgradeIronicIP := ""
+		if e2eConfig.HasVariable("UPGRADE_IRONIC_PROVISIONING_IP") {
```
adilGhaffarDev, line 307: why do we need `HasVariable`, isn't just `GetVariable` enough?

lentzi90: `GetVariable` fails the test if the variable doesn't exist. (It uses `Expect`.)
We could change it so that it returns an error instead. Then it would be possible to check for that instead. I don't have a preference which way we go. The current way avoids error checks, which is quite nice when all variables are expected to be defined. Returning an error would mean we don't need `HasVariable`.

PR #2949, opened 2026-01-29
✨: Expose infiniband clientId

### LINKED ISSUE #2948: [RFE] Expose infiniband client_id
**User Story**

As an operator I would like to expose client_id for infniband interfaces for having ability to use this data later when working with infiniband networks.

**Detailed Description**

Ironic discovers infiniband client_id for a [long time](https://specs.openstack.org/openstack/ironic-specs/specs/6.2/add-infiniband-support.html). We need to expose this field for a NIC into baremetalhost status and hardwaremeta.

/kind feature

jumpojoy:
Ironic collects infiniband client_id this patch makes sure its propagated into hardwaredata.

fixes #2948


s3rj1k (approved):
Looks good, but some one else should also look at this

Issue #2687, opened 2025-09-17
Run cleaning before moving to available, not afterwards


dtantsur:
When enrolling a host, we leave it in the `available` state with the Ironic Node in the `manageable` state. Not only this mismatch may be confusing, it also means that the first deployment will take somewhat longer because cleaning must run on the path to the Ironic's `available` state.

Do we have any reasons not to run cleaning right away?

/kind bug

dtantsur:
/cc @zaneb 

Zane, do you remember why it was done this way?

dtantsur:
I think I see a problem: any preparing action will then trigger a new automated cleaning because we'd need to move back to `manageable`. This may be a good reason to reject this proposal. I'll leave it open for further comments.

zaneb:
I have a vague recollection that I did this and it got reverted.

zaneb:
I believe that #882 (and specifically 9a55ce5bc63a2f24916f1f604c736570836fdaea) implemented this.
That was reverted in #929 to work around a bug in RAID on Fujitsu.
Discussion in #959 suggests that the issues that caused us to revert have now been solved, such that we could restore the desired behaviour. I don't remember if I followed up to confirm that was the case, but there was a PR to reinstate it, #958, which I did eventually [review](https://github.com/metal3-io/baremetal-operator/pull/958#discussion_r1071717881). Reading that thread, it sounds like the owner found a neat solution to the problem, but by that point I had no time to re-review and it died on the vine.

So the answer _may_ be as simple as resurrecting #958.

> I think I see a problem: any preparing action will then trigger a new automated cleaning because we'd need to move back to manageable.

I don't actually see a problem with that (and [neither did you](https://groups.google.com/g/metal3-dev/c/N-EpqQn94bw/m/2KHWVA5PAAAJ) when we discussed it on the mailing list 🙂). Yes, it's more cleanings. But latency is almost always reduced.

PR #2954, opened 2026-01-30
WIP: :sparkles: Consumer Override for BareMetalHost Cleaning

### LINKED ISSUE #2946: [RFE] Temporary cleaning mode override for the consumer of a BareMetalHost
**User Story**

The setting is a multi-tenant deployment with two kind of actors: consumers of BareMetalHost who hold a reference on a BareMetalHost, typically through a hostclaim, an infrastructure administrator responsible for the state of the BareMetalHosts given to consumers.
As a consumer, I want to control the level of cleaning applied when the BareMetalHost is deprovisionned but I still have a reference on it. For example I may want to disable cleaning (required for the implementation of capm3 node reuse with hostclaims).
As the administrator, I want to enforce a minimum level of cleaning as soon as the baremetalhost is given back to the pool (the consumerRef is removed).

**Detailed Description**

There is only one field (automatedCleaningMode). If the consumer overwrite the value, the default intended by the administrator is lost.

Review of solutions:
1. Annotation for single override of the cleaning mode.
  The user can set an annotation (baremetal.metal3.io/cleaning). The value is the cleaning level for the next cleaning operation. If the node is available, the cleaning operation is performed immediately. If the node is provisionned, the cleaning (or lack of) will be performed during deprovisioning. The consumer (typically the hostclaim controller) is responsible for pushing the annotation at the right time. More details in #2922.

2. Cleaning mode override
  We introduce a new field in the spec of the BareMetalHost to specify how cleaning is performed while the consumerRef exists. To enforce the fact that this field is cleared when the consumerRef is cleared, we either 
    a. put it inside the consumerRef or 
    b. in a consumerOverride field with an admission constraint requesting that the field is nil unless a consumerRef exists.
  We also need a field in the status that records the last level of cleaning performed with the override. The reason is that we may clear the consumerRef of an unprovisionned BareMetalHost. If we rely only on automated cleaning and the override specified no cleaning, then the host will not be cleaned before it is handled to another user. As soon as the override disapear, if the bareMetalHost is available but the laast cleaning performed (recorded in status) is weaker than themode specified in automatedCleaning, a manual cleaning is performed immediately. 

3. Fully declarative cleaning
   We use the same fields as in the previous proposal. The main change is that we disable automatedCleaning and   always relie on the comparison between the cleaning status and the current mode. The current mode is either the value in the override if it exists, or the value of .spec.automatedCleaningMode. When the BareMetalHost is available and a current mode is set to a stronger value than the one recorded in status, a cleaning operation is performed.

Solution 1. is very imperative in spirit and without an additional cleaningStatus, the end user cannot know the level of cleaning performed on an available host.
Solution 2. distinguishes a declarative behavior while the overrides is defined (cleaning status follows strictly the value specified in the override) from the legacy behaviour used when the overrides is nil. 
The main drawback of 3. is that it changes the behaviour of automatedCleaningMode. As soon as the field is changed, on an available BareMetalHost, a cleaning operation may be performed if the previous cleaning status was not good enough. If the bareMetalHost is provisionned (provisioning), there is no change. The setting will only apply on the next deprovisioning operation. But the semantics of 3 is much cleaner and simpler to understand.

Solution 2.b (3.b) seem better than 2.a (3.a) because it does not modify the type of consumerRef (a corev1 reference).

/kind feature

pierrecregut:
Introduce a ConsumerOverride field to override some fields of the BareMetalHost without touching its definition. This is only used when a ConsumerRef exists and a CEL invariant enforces it.

Introduces a OverridenCleaningPerformed recording what is the last cleaning step applied on the provisionned OS when ConsumerOverride is defined. If the OS is provisionned or no Consumer override is defined, CleaningStatus is nil.

When an available machine loose its consumerRef and the new automated cleaning mode is stronger than the last cleaning performed, the baremetal host is pushed to PreparingState as for a firmware setting or raid reconfiguration and a manual cleaning step is performed.

So it looks as if the last deprovisioning occured at the time the consumerRef was removed and the standard cleaning mode was enforced.

Implements proposal 2b of #2946 

Fixes #2946

**Checklist:**


pierrecregut:
/cc @dtantsur

Clearly not ready for merge yet. Missing tests and strange behaviour on e2e test.

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -944,6 +961,10 @@ func (host *BareMetalHost) UpdateTriedCredentials(currentSecret corev1.Secret) {
 // to be published to the kubernetes API.
 func (host *BareMetalHost) NewEvent(reason, message string) corev1.Event {
 	t := metav1.Now()
+	var ref *corev1.ObjectReference
+	if host.Spec.ConsumerRef != nil {
+		ref = host.Spec.ConsumerRef
+	}
```
dtantsur: This change does not seem to do anything

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -869,7 +870,7 @@ func (r *BareMetalHostReconciler) registerHost(prov provisioner.Provisioner, inf
 	provResult, provID, err := prov.Register(
 		provisioner.ManagementAccessData{
 			BootMode:                   info.host.Status.Provisioning.BootMode,
-			AutomatedCleaningMode:      info.host.Spec.AutomatedCleaningMode,
+			AutomatedCleaningMode:      metal3api.CleaningModeDisabled,
```
dtantsur: Shouldn't this be currentCleaningMode?

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -1192,6 +1193,12 @@ func (r *BareMetalHostReconciler) actionPreparing(prov provisioner.Provisioner,
 		RootDeviceHints:  newStatus.Provisioning.RootDeviceHints.DeepCopy(),
 		FirmwareConfig:   newStatus.Provisioning.Firmware.DeepCopy(),
 	}
+	// Setup cleaning
+	var manualCleaning = requiresCleaning(info.host)
+	if manualCleaning {
+		r.Log.Info("Manual cleaning is required")
```
dtantsur: "manual cleaning" may be confusing here, let's do something more detailed like

```go
r.Log.Info("re-running cleaning because the current mode is stricter than the last applied", "currentMode", ..., "lastAppliedMode", ...)
```

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -1357,6 +1370,8 @@ func (r *BareMetalHostReconciler) actionProvisioning(prov provisioner.Provisione
 		info.host.Status.Provisioning.CustomDeploy = info.host.Spec.CustomDeploy.DeepCopy()
 	}
 
+	// it is not clean
```
dtantsur: s/not/now/ ?

test/e2e/data/ironic-standalone-operator/ironic/base/ironic.yaml
```diff
@@ -14,5 +14,5 @@ spec:
     ipAddress: "192.168.222.2"
     ipAddressManager: "keepalived"
   deployRamdisk:
-    sshKey: "SSH_PUB_KEY_CONTENT"
```
dtantsur, line 17: You probably did not want to commit this

pkg/provisioner/ironic/ironic.go
```diff
@@ -1038,6 +1038,22 @@ func (p *ironicProvisioner) buildManualCleaningSteps(bmcAccess bmc.AccessDetails
 		)
 	}
 
+	if data.RequiredCleaning != nil {
```
dtantsur: There is a completely different path you could take here. Automated cleaning runs when going from Ironic's `manageable` state to `available` (both map to `available` in BMH). So you could update `Prepare` to make sure that if cleaning is required, the node is left in the `available` state.

The benefit of this approach is that you won't run cleaning twice: first explicitly here, second on the next provisioning when going to Ironic's `available` state.

See https://github.com/metal3-io/baremetal-operator/issues/2687 for a discussion of this whole situation.

apis/metal3.io/v1alpha1/baremetalhost_types.go
pierrecregut: Good catch. Bad reverse from the use of extended consumerRef where it was necessary to extract ConsumerRef.ObjectReference

internal/controller/metal3.io/baremetalhost_controller.go
pierrecregut: Yes wrong initialization. Again mix with the "declarative" version.

internal/controller/metal3.io/baremetalhost_controller.go
pierrecregut: We just provisionned an OS. So clearly it is not clean. We must forget the last cleaning status.

test/e2e/data/ironic-standalone-operator/ironic/base/ironic.yaml
pierrecregut: Yes.... This would be nice if it was in the .gitignore...

pkg/provisioner/ironic/ironic.go
pierrecregut: Sorry for the delay, it took me a while to fully understand the point. So, because the manual cleaning step ended in the manageable state, a new automatic cleaning would be triggered at the next provisioning.... or not, because if the next consumer is a  hostclaim, when the bmh is  provisionned, the automatic cleaning has probably been disabled until it is given back to the infrastructure.
So the right solution seems to go to manageable, perform nothing but going back to available state immediately using the 'provide' verb. This is what this solution is doing. In prepare data, I only keep a boolean to request a "manually triggered automatic cleaning" and if it is set I just do a provide without looking at anything else.

PR #2955, opened 2026-01-30
✨ Allow forced detachment of a host from Ironic

### LINKED ISSUE #2923: [RFE] Forced detachment of a host from Ironic
**User Story**

As an operator of a misbehaving host, I want to be able to set it aside as soon as possible for further investigation, even if the currently running process cannot finish correctly.

**Detailed Description**

The real world is messy, and I see an increasing number of situation where a host gets wedged in a certain state, and an external intervention is required. It's not really helpful if BMO and Ironic are trying to enforce the desired state at the same time, potentially rebooting and reconfiguring the host.

The current primitives don't fit the bill:
- The regular detachment only works for available and provisioned hosts.
- The pause annotation does not prevent Ironic from e.g. retrying failed boot or reconfiguring the Node on timeout.
- Now that deletion can abort cleaning and inspection, it can be used as a substitute at the cost of losing the host record. We'd still need to be able to abort deployment.

Here is a rough idea. The detach annotation already accepts a JSON payload for another obscure feature. We can expand it with a flag `force: true`. Unlike with a normal detachment, such a detachment will be propagated to the provisioner (aka Ironic) immediately instead of waiting for a suitable state.

On the provisioner side, we'll abort any current process whenever possible (i.e. in DEPLOYWAIT, CLEANWAIT, SERVICEWAIT or INSPECTWAIT states). Once a stable state is reached (e.g. CLEANFAIL), the Node will be removed from Ironic (with maintenance on if needed). In case of provisioning, https://review.opendev.org/c/openstack/ironic/+/973279 will be a soft dependency for this feature. Without it, the Node will need to go to cleaning before we'll be able to abort the process. The logic will be adjusted based on the available API version.

Removing the detachment annotation will cause the Node to be re-created. After that, the controller will drive it through the necessary steps to reach the expected status again.

/kind feature

CrystalChun:
Introduces the `force` argument for detaching. If set to true, it would bypass whichever status the host is currently in, abort the action in Ironic, and force the host's detachment.

Note: there is a soft dependency on the [new feature in Ironic](https://review.opendev.org/c/openstack/ironic/+/973279) that allows the `abort` action on deploying nodes. Without it, deprovisioning will be triggered first, which won't preserve the exact host's state.

Finally, I have prepared e2e tests for this change, but they ended up being timing-dependent. To fix this dependency, I need direct access to Ironic Node record, which is going to greatly increase the already large size of this change. The tests will be proposed separately.

Fixes #2923 

**Checklist:**

### DIFF
--- a/apis/metal3.io/v1alpha1/baremetalhost_types.go
+++ b/apis/metal3.io/v1alpha1/baremetalhost_types.go
@@ -699,6 +699,10 @@ const (
 type DetachedAnnotationArguments struct {
 	// DeleteAction indicates the desired delete logic when the detached annotation is present
 	DeleteAction DetachedDeleteAction `json:"deleteAction,omitempty"`
+
+	// Force indicates if detaching should be forced regardless of the host's state
+	// +optional
+	Force bool `json:"force,omitempty"`
 }
 
 // Match compares the saved status information with the name and
--- a/internal/controller/metal3.io/host_state_machine.go
+++ b/internal/controller/metal3.io/host_state_machine.go
@@ -2,7 +2,6 @@ package controllers
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 
 	"github.com/go-logr/logr"
@@ -266,17 +265,6 @@ func (hsm *hostStateMachine) checkInitiateDelete(log logr.Logger) bool {
 	return true
 }
 
-// hasDetachedAnnotation checks for existence of baremetalhost.metal3.io/detached.
-func hasDetachedAnnotation(host *metal3api.BareMetalHost) bool {
-	annotations := host.GetAnnotations()
-	if annotations != nil {
-		if _, ok := annotations[metal3api.DetachedAnnotation]; ok {
-			return true
-		}
-	}
-	return false
-}
-
 // skipReconcileSubresource checks if the applied annotations and status
 // prevent a subresource (HFS/HFC) from being reconciled.
 func skipReconcileSubresource(host *metal3api.BareMetalHost, logger logr.Logger) bool {
@@ -300,27 +288,6 @@ func skipReconcileSubresource(host *metal3api.BareMetalHost, logger logr.Logger)
 	return false
 }
 
-func delayDeleteForDetachedHost(host *metal3api.BareMetalHost) bool {
-	annotations := host.GetAnnotations()
-	args := metal3api.DetachedAnnotationArguments{}
-	val, present := annotations[metal3api.DetachedAnnotation]
-
-	// if the host is detached, but missing the annotation, also delay delete
-	// to allow for the host to be re-attached
-	if !present && host.OperationalStatus() == metal3api.OperationalStatusDetached {
-		return true
-	}
-
-	if present {
-		if err := json.Unmarshal([]byte(val), &args); err != nil {
-			// default behavior if these are missing or not json is to not delay
-			return false
-		}
-	}
-
-	return args.DeleteAction == metal3api.DetachedDeleteActionDelay
-}
-
 func (hsm *hostStateMachine) checkDetachedHost(ctx context.Context, info *reconcileInfo) (result actionResult) {
 	// If the detached annotation is set we remove any host from the
 	// provisioner and take no further action
@@ -333,7 +300,14 @@ func (hsm *hostStateMachine) checkDetachedHost(ctx context.Context, info *reconc
 		case metal3api.StateDeleting:
 			// No point in detaching a host that is being deleted already
 		default:
-			if hasForceDetachAnnotation(hsm.Host) {
+			annotation, err := getDetachedAnnotation(hsm.Host)
+			if err != nil {
+				// FIXME(dtantsur): ignoring errors is not great and can cause a lot of confusion.
+				// However, we even document (in https://book.metal3.io/bmo/detached_annotation.html)
+				// that the value can be anything, so starting failing is a breaking change.
+				info.log.Error(err, "ignoring detached annotation value")
+			}
+			if annotation != nil && annotation.Force {
 				info.log.Info("forcing detach of host", "provisioningState", info.host.Status.Provisioning.State)
 				return hsm.Reconciler.detachHost(ctx, hsm.Provisioner, info, true)
 			}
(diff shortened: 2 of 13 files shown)


dtantsur:
@CrystalChun while I'm testing something locally, could you squash your commits and remove references to RH Jira from the messages?

CrystalChun:
> @CrystalChun while I'm testing something locally, could you squash your commits and remove references to RH Jira from the messages?

Yes, squashed my commits and removed the RH Jira mentions from the commit message and the description of this PR!

dtantsur:
@CrystalChun I've provided a draft implementation, needs testing now (also together with soon-to-be-merged Ironic change https://review.opendev.org/c/openstack/ironic/+/973279)

dtantsur:
/retest

Nothing seems obviously wrong, the hosts just don't get reconciled.

dtantsur:
/retest

At least some failures seem unrelated

dtantsur:
/retest

Timeout

dtantsur:
/hold

Despite 2 seconds sleep, one of the tests still did not reach "deploying" before forced detachment was applied.

dtantsur:
/hold cancel

I had to drop the e2e tests for now to avoid blocking the entire feature. There are signs that they would pass, but they're timing-dependent, which is pretty bad. I'll propose them separately.

Rozzii:
I will take a look soon, for some reason this PR have not received the 2reviewer automatically.
/cc @Rozzii

internal/controller/metal3.io/host_state_machine.go
```diff
@@ -314,11 +303,17 @@ func (hsm *hostStateMachine) checkDetachedHost(ctx context.Context, info *reconc
 	// provisioner and take no further action
 	// Note this doesn't change the current state, only the OperationalStatus
 	if hasDetachedAnnotation(hsm.Host) {
-		// Only allow detaching hosts in Provisioned/ExternallyProvisioned/Ready/Available states
+		// Only allow detaching hosts in Provisioned/ExternallyProvisioned/Ready/Available states unless forced
 		switch info.host.Status.Provisioning.State {
 		case metal3api.StateProvisioned, metal3api.StateExternallyProvisioned, metal3api.StateReady, metal3api.StateAvailable:
-			return hsm.Reconciler.detachHost(ctx, hsm.Provisioner, info)
+			return hsm.Reconciler.detachHost(ctx, hsm.Provisioner, info, false)
```
MahnoorAsghar, line 299: IIUC there is no way to force detach in these states?

dtantsur: It's not required: detachment always works for them.

Rozzii:
In my review queue now, I would like to merge this before 0.13 release @Sunnatillo @smoshiur1237 please keep an eye on this PR.

pkg/provisioner/ironic/delete_test.go
```diff
@@ -24,6 +24,118 @@ func TestDetach(t *testing.T) {
...
+			ironic: testserver.NewIronic(t).Node(nodes.Node{
+				UUID:           nodeUUID,
+				ProvisionState: string(nodes.InspectWait),
+			}).WithNodeStatesProvisionUpdate(nodeUUID),
+			expectedDirty:     true,
+			expectedRequeue:   provisionRequeueDelay,
+			expectedProvState: nodes.TargetAbort,
+		},
+		{
+			name: "cleanwait-abort",
```
elfosardo, line 71: this duplicates L61

pkg/provisioner/ironic/ironic.go
```diff
@@ -1592,10 +1627,26 @@ func (p *ironicProvisioner) Delete(ctx context.Context) (result provisioner.Resu
 // for the target system.  It may be called multiple times,
 // and should return true for its dirty  flag until the
 // deletion operation is completed.
-func (p *ironicProvisioner) Detach(ctx context.Context) (result provisioner.Result, err error) {
+func (p *ironicProvisioner) Detach(ctx context.Context, force bool) (result provisioner.Result, err error) {
 	// Currently the same behavior as Delete()
```
elfosardo: mmm this is no longer accurate?

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -699,6 +699,11 @@ const (
 type DetachedAnnotationArguments struct {
 	// DeleteAction indicates the desired delete logic when the detached annotation is present
 	DeleteAction DetachedDeleteAction `json:"deleteAction,omitempty"`
+
+	// Force indicates if detaching should be forced regardless of the host's state
+	// +optional
+	// +kubebuilder:default:=false
```
elfosardo: nit: this is redundant, zero value for bool is already false

pkg/provisioner/ironic/delete_test.go
```diff
@@ -24,6 +24,118 @@ func TestDetach(t *testing.T) {
...
+				UUID:           nodeUUID,
+				ProvisionState: string(nodes.DeployWait),
+			}).WithNodeStatesProvisionUpdate(nodeUUID),
+			maxVersion:        95,
+			expectedDirty:     true,
+			expectedRequeue:   provisionRequeueDelay,
+			expectedProvState: nodes.TargetDeleted,
+		},
+		{
+			name: "cleanwait-abort",
```
elfosardo: maybe change this to "inspectwait-abort" ?

internal/controller/metal3.io/host_state_machine.go
```diff
@@ -311,37 +288,22 @@ func skipReconcileSubresource(host *metal3api.BareMetalHost, logger logr.Logger)
...
-}
-
 func (hsm *hostStateMachine) checkDetachedHost(ctx context.Context, info *reconcileInfo) (result actionResult) {
 	// If the detached annotation is set we remove any host from the
 	// provisioner and take no further action
 	// Note this doesn't change the current state, only the OperationalStatus
 	if hasDetachedAnnotation(hsm.Host) {
-		// Only allow detaching hosts in Provisioned/ExternallyProvisioned/Ready/Available states
+		// Only allow detaching hosts in Provisioned/ExternallyProvisioned/Ready/Available states unless forced
 		switch info.host.Status.Provisioning.State {
```
elfosardo, line 297: I wonder if we could handle a failed parse annotation here instead of relying on hasForceDetachAnnotation, which ignores errors
I know we're ignoring it, so maybe in a follow up?

dtantsur: Very unfortunately, we even document in https://book.metal3.io/bmo/detached_annotation.html that the value is ignored. I'll update with a log message.

dtantsur:
@Rozzii @tuminoid looking for a final lgtm here please

Rozzii:
Look at it right now.

Rozzii (commented):
/lgtm

PR #2963, opened 2026-02-03
🐛 Do not log firmware components on each BMH reconciliation

dtantsur:
GetFirmwareComponents is called each time, which creates a lot of
logging. Only log the components when the list changes.


adilGhaffarDev (commented):
/lgtm

PR #2964, opened 2026-02-03
🐛 Correctly handle resource generations for HFC and HFS

dtantsur:
Currently, the controller does not save the status if only the
generation changes, leading to an observed generation mismatch on the
ChangeDetected condition.

Meanwhile, the servicing code assumes that there are no updates when
generations mismatch, even if the condition is true.

This patch fixes the controller and makes the servicing code retry
if generations do not match.

In the 2nd commit, I'm generalizing the fix to cover HFS too.


iurygregory:
@dtantsur can you also fix for HFS? According to Don's testing the same issue happens there.

dtantsur:
> can you also fix for HFS? According to Don's testing the same issue happens there.

I was going to take a look, although the HFS logic is somewhat different.

PR #2966, opened 2026-02-03
:sparkles:  Add multi-architecture PXE boot support

honza:
Enable aarch64 workers to be provisioned from an x86_64 control plane.

Architecture-aware default image provider:

The built-in PreprovisioningImage provider (pkg/imageprovider/default.go)
now selects architecture-specific IPA images. It supports multiple
environment variable formats, following the ironic-image conventions:

  1. Per-arch variable (e.g. `DEPLOY_KERNEL_URL_AARCH64`)
  2. BY_ARCH variable  (e.g. `DEPLOY_KERNEL_BY_ARCH=aarch64:http://...`)
  3. Base variable     (e.g. `DEPLOY_KERNEL_URL`)

Per-arch variable names use uppercase architecture suffixes to match the
ironic-image convention (e.g. `DEPLOY_KERNEL_URL_X86_64`,
`DEPLOY_RAMDISK_URL_AARCH64`). The BY_ARCH format uses comma-separated
`arch:url` pairs, allowing a single variable that both ironic-image and
BMO can consume. Architecture matching in the BY_ARCH value is
case-insensitive. The `BY_ARCH` variable name is derived by replacing
the `_URL` suffix with `_BY_ARCH` (e.g. `DEPLOY_KERNEL_URL ->
DEPLOY_KERNEL_BY_ARCH`).

For InitRD format, the provider now returns both KernelURL and ImageURL,
so the PreprovisioningImage status carries the correct
architecture-specific URLs.

Setting cpu_arch on Ironic nodes:

The cpu_arch property is set on Ironic nodes during both registration and
inspection, based on the host's architecture (from Spec.Architecture,
HardwareDetails, or the controller's own architecture as fallback).
This allows Ironic to select architecture-specific IPA images via its
deploy_kernel_by_arch and deploy_ramdisk_by_arch conductor configuration
when PreprovisioningImage is not in use.

By handling architecture-specific image selection in the image provider
layer rather than the ironic provisioner, we avoid mixed per-node and
by_arch configurations that Ironic considers invalid. Both kernel and
ramdisk URLs are set per-node from the PreprovisioningImage, so there is
no conflict with Ironic's conductor-level by_arch settings.

Depends on https://github.com/metal3-io/ironic-image/pull/898

### DIFF
--- a/pkg/imageprovider/default.go
+++ b/pkg/imageprovider/default.go
@@ -3,6 +3,7 @@ package imageprovider
 import (
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/go-logr/logr"
 	metal3api "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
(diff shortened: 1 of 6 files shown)


s3rj1k:
Oh, I was exactly thinking about this for Preprovision image, I would however expect this to support similar `by-arch` environs as Ironic-Image now has, so we would have unified experience

honza:
> Oh, I was exactly thinking about this for Preprovision image, I would however expect this to support similar by-arch environs as Ironic-Image now has, so we would have unified experience

I explored that idea at length but ultimately it didn't work. Ironic [considers mixed configurations a bug](https://github.com/openstack/ironic/commit/a78697462).

s3rj1k:
> > Oh, I was exactly thinking about this for Preprovision image, I would however expect this to support similar by-arch environs as Ironic-Image now has, so we would have unified experience
> 
> I explored that idea at length but ultimately it didn't work. Ironic [considers mixed configurations a bug](https://github.com/openstack/ironic/commit/a78697462).

This is strange, I use Ironic `by_arch` feature without PreprovisionImage and it works fine.

pkg/imageprovider/default.go
```diff
@@ -20,6 +20,18 @@ func NewDefaultImageProvider() ImageProvider {
 	}
 }
 
+// envWithArchFallback returns the value of an architecture-specific environment
```
s3rj1k: Please use same environ keys as in Ironic-Image

https://github.com/metal3-io/ironic-image/blob/main/README.md?plain=1#L106-L125

DEPLOY_KERNEL_BY_ARCH
DEPLOY_RAMDISK_BY_ARCH

s3rj1k (approved):
@honza Thanks for handling that by-arch environ things, much appreciated

pkg/provisioner/ironic/ironic.go
```diff
@@ -472,7 +472,6 @@ func setDeployImage(config ironicConfig, accessDetails bmc.AccessDetails, hostIm
 			}
 			deployImageInfo[deployRamdiskKey] = hostImage.ImageURL
 			if hostImage.ExtraKernelParams != "" {
-				// Using %default% prevents overriding the config in ironic-image
```
dtantsur, line 475: Hmm, why is this removed?

honza: Fixed.

dtantsur:
/lgtm

Thanks!

PR #2972, opened 2026-02-06
🌱 Avoid meaningless logging when a secret is owned by another controller

dtantsur:
Currently, BMO logs that it's going to update the secret, even though
it's not the case when the owner is not BMH.

Add missing unit tests.

### DIFF
--- a/pkg/secretutils/secret_manager.go
+++ b/pkg/secretutils/secret_manager.go
@@ -83,11 +83,13 @@ func (sm *SecretManager) claimSecret(secret *corev1.Secret, owner client.Object,
 		for _, ref := range secret.GetOwnerReferences() {
 			// We used to add controller references to BMC
 			// secrets. This was wrong, update.
-			if ref.UID == ownerUID && ref.Controller == nil {
-				alreadyOwned = true
+			if ref.UID == ownerUID {
+				if ref.Controller == nil {
+					alreadyOwned = true
+				} else if *ref.Controller {
+					ownerLog.Info("updating secret to no longer have an owner of type controller")
+				}
 				break
-			} else if ref.Controller != nil && *ref.Controller {
-				ownerLog.Info("updating secret to no longer have an owner of type controller")
 			}
 		}
 		if !alreadyOwned {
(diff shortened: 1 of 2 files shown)


pkg/secretutils/secret_manager.go
```diff
@@ -83,11 +83,13 @@ func (sm *SecretManager) claimSecret(secret *corev1.Secret, owner client.Object,
 		for _, ref := range secret.GetOwnerReferences() {
 			// We used to add controller references to BMC
 			// secrets. This was wrong, update.
-			if ref.UID == ownerUID && ref.Controller == nil {
-				alreadyOwned = true
+			if ref.UID == ownerUID {
+				if ref.Controller == nil {
+					alreadyOwned = true
+				} else if *ref.Controller {
+					ownerLog.Info("updating secret to no longer have an owner of type controller")
```
kashifest, line 90: /lgtm
/hold
Although I agree with the change, but I couldn't get my head around this log line, is it only me or does it actually makes sense?

tuminoid: @dtantsur

dtantsur: I decided against changing it but can do if you wish. I'm not sure how to put it better though. The idea is to remove the `Controller` flag, so "no longer have an owner of type controller" kind of is true.

dtantsur:
@kashifest hey, what would you prefer to happen to unblock this PR?

pkg/secretutils/secret_manager.go
kashifest: The issue is not in this patch the log message itself which was there before is hard to interpret. Does it mean updating secret to remove any owner of type controller ?

kashifest:
Can be taken separately as well,  feel free to unhold, 

> @kashifest hey, what would you prefer to happen to unblock this PR?

pkg/secretutils/secret_manager.go
dtantsur: Not necessarily remove the owner. If the owner exists, change it to Controller=false.

PR #2979, opened 2026-02-11
:seedling: enforce ol-prefix 'one' style for ordered lists

tuminoid:
We have it in metal3-io wide use except couple lists in BMO, and some old design docs, so enforcing it in config in all repos for consistency.


tuminoid:
> /approve

Please LGTM, we have approve already.
/cc @Rozzii

PR #2980, opened 2026-02-11
✨ Add annotation to override Ironic vendor property

### LINKED ISSUE #2978: Allow to set Ironic node `properties`.
**User Story**

As a user I would like to be able to set custom `properties` in Ironic because some hardware vendor rebrand firmwares, making it impossible for Ironic to correctly guess the vendor without setting `properties.vendor`.

**Detailed Description**

I would like to be able to set Ironic's `properties` either through an annotation or a new field on the BMH CRD.

Annotation:
```yaml
kind: BareMetalHost
metadata:
  annotations:
    properties.metal3.io: '{"vendor":"ami"}'
  name: my-server
spec: {}
```

CRD :
```go
// BareMetalHostSpec defines the desired state of BareMetalHost.
type BareMetalHostSpec struct {
    Properties map[string]string `json:"properties,omitempty"`
}
```

An example use case it to be able to set the vendor to "ami" on GBx00 Nvidia GPU servers, as the firmware is AMI, but the vendor is reported as Dell, which makes Ironic send the wrong request to set the boot device. Setting the `properties.vendor` seems more straightforward than patching Ironic to have a more complex vendor detection logic (if it is even possible).

/kind feature

synthe102:

~~This PR adds a new `hostProvisionerProperties` to the `BareMetalHost` CRD as well as the implementation for the Ironic provisioner.~~

~~As discussed during the community meeting, list of possible keys is managed by the provisioner itself to avoid leaking Ironic specific concepts.~~

~~Ironic currently supports setting the following properties:~~
~~- `vendor`~~

Since I'm not an Ironic expert, I'll gladly take feedback on any sane default for the properties allow list.

⚠️ This PR now uses an annotation `ironic.provisioners.metal3.io/vendor` to set the vendor rather for adding an arbitrary metadata map to the BMH specs.

Fixes #2978 

**Checklist:**

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -115,3 +115,26 @@ See [PreprovisioningImage
 CR](https://doc.crds.dev/github.com/metal3-io/baremetal-operator/metal3.io/PreprovisioningImage/v1alpha1)
 or check the source code at `apis/metal3.io/v1alpha1/preprovisioningimage_types.go`
 for a detailed API description.
+
+## Vendor Annotation
+
+The **BareMetalHost** resource supports overriding the vendor property on the
+Ironic node via the `ironic.provisioners.metal3.io/vendor` annotation. This
+is useful when the vendor detected by Ironic does not match the expected value.
+
+### Example Usage
+
+```yaml
+apiVersion: metal3.io/v1alpha1
+kind: BareMetalHost
+metadata:
+  name: gpu-server-01
+  annotations:
+    ironic.provisioners.metal3.io/vendor: "ami"
+spec:
+  online: true
+  bootMACAddress: "00:11:22:33:44:55"
+  bmc:
+    address: redfish-virtualmedia://192.168.1.100/redfish/v1/Systems/1
+    credentialsName: bmc-secret
+```
--- a/pkg/provisioner/ironic/ironic.go
+++ b/pkg/provisioner/ironic/ironic.go
@@ -336,6 +336,9 @@ func (p *ironicProvisioner) configureNode(ctx context.Context, data provisioner.
 	if data.CPUArchitecture != "" {
 		opts["cpu_arch"] = data.CPUArchitecture
 	}
+	if data.Vendor != "" {
+		opts["vendor"] = data.Vendor
+	}
 	updater.SetPropertiesOpts(opts, ironicNode)
 
 	_, success, result, err := p.tryUpdateNode(ctx, ironicNode, updater)
--- a/pkg/provisioner/ironic/register.go
+++ b/pkg/provisioner/ironic/register.go
@@ -241,6 +241,14 @@ func (p *ironicProvisioner) Register(ctx context.Context, data provisioner.Manag
 }
 
 func (p *ironicProvisioner) enrollNode(ctx context.Context, data provisioner.ManagementAccessData, bmcAccess bmc.AccessDetails, driverInfo map[string]any) (ironicNode *nodes.Node, retry bool, err error) {
+	properties := map[string]any{
+		"capabilities": buildCapabilitiesValue(nil, data.BootMode),
+		"cpu_arch":     data.CPUArchitecture,
+	}
+	if data.Vendor != "" {
+		properties["vendor"] = data.Vendor
+	}
+
 	nodeCreateOpts := nodes.CreateOpts{
 		Driver:              bmcAccess.Driver(),
 		BIOSInterface:       bmcAccess.BIOSInterface(),
@@ -255,10 +263,7 @@ func (p *ironicProvisioner) enrollNode(ctx context.Context, data provisioner.Man
 		RAIDInterface:       bmcAccess.RAIDInterface(),
 		VendorInterface:     bmcAccess.VendorInterface(),
 		DisablePowerOff:     &data.DisablePowerOff,
-		Properties: map[string]any{
-			"capabilities": buildCapabilitiesValue(nil, data.BootMode),
-			"cpu_arch":     data.CPUArchitecture,
-		},
+		Properties:          properties,
 	}
 
 	ironicNode, err = nodes.Create(ctx, p.client, nodeCreateOpts).Extract()
--- a/apis/metal3.io/v1alpha1/baremetalhost_types.go
+++ b/apis/metal3.io/v1alpha1/baremetalhost_types.go
@@ -64,6 +64,11 @@ const (
 	// is disabled.
 	HardwareDetailsAnnotation = InspectAnnotationPrefix + "/hardwaredetails"
 
+	// VendorAnnotation is the annotation that allows users to override the
+	// vendor property on the provisioner's host representation (e.g., Ironic
+	// node properties).
+	VendorAnnotation = "ironic.provisioners.metal3.io/vendor"
+
 	// InspectAnnotationValueDisabled is a constant string="disabled"
 	// This is particularly useful to check if inspect annotation is disabled
 	// inspect.metal3.io=disabled.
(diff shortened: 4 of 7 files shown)


pkg/provisioner/ironic/ironic.go
```diff
@@ -47,6 +47,24 @@ const (
...
+}
+
+// applyHostProvisionerProperties applies allowed host provisioner properties to the
+// given properties map, logging a warning for any unsupported properties.
+func (p *ironicProvisioner) applyHostProvisionerProperties(properties map[string]any, hostProps map[string]string) {
+	for key, value := range hostProps {
+		if AllowedHostProvisionerProperties[key] {
+			properties[key] = value
+		} else {
+			p.log.Info("WARNING: ignoring unsupported host provisioner property", "key", key, "value", value)
```
diconico07: This `Info`/`WARNING` doesn't feel right

config/base/crds/bases/metal3.io_baremetalhosts.yaml
```diff
@@ -257,6 +257,15 @@ spec:
...
                   without hardware profiles.
                 type: string
+              hostProvisionerProperties:
+                additionalProperties:
+                  type: string
+                description: |-
+                  HostProvisionerProperties allows users to set specific properties
+                  on the provisioner's host representation (e.g., Ironic node properties).
+                  Only properties in the provisioner's allowlist can be set.
+                  Currently allowed: vendor.
```
diconico07: I'd rather point to the docs here for the full list, as it's a generic field we won't have to update the CRD every time we add a new allowed value.

pkg/provisioner/ironic/ironic.go
synthe102: `logr` doesn't have a warning level. I can remove the `WARNING` and keep it as info.

config/base/crds/bases/metal3.io_baremetalhosts.yaml
synthe102: You're right, will fix.

pkg/provisioner/ironic/register.go
```diff
@@ -239,6 +239,12 @@ func (p *ironicProvisioner) Register(data provisioner.ManagementAccessData, cred
 }
 
 func (p *ironicProvisioner) enrollNode(data provisioner.ManagementAccessData, bmcAccess bmc.AccessDetails, driverInfo map[string]any) (ironicNode *nodes.Node, retry bool, err error) {
+	properties := map[string]any{
+		"capabilities": buildCapabilitiesValue(nil, data.BootMode),
```
elfosardo, line 245: applyHostProvisionerProperties is called after capabilities and cpu_arch are set in both configureNode and enrollNode. If capabilities or cpu_arch were ever added to the allowlist, a user could overwrite BMO-managed properties.  We should consider adding a "denied" list of BMO-internal keys that can never be overridden, or checking for this explicitly.

pkg/provisioner/ironic/ironic.go
```diff
@@ -47,6 +47,24 @@ const (
 	kernelParamsKey  = "kernel_append_params"
 )
 
+// AllowedHostProvisionerProperties defines which properties users can override
+// via the BareMetalHost spec.hostProvisionerProperties field.
+var AllowedHostProvisionerProperties = map[string]bool{
```
elfosardo: AllowedHostProvisionerProperties is an exported var. Any code could modify it at runtime. Maybe either make it unexported or use a function that returns a copy?

docs/api.md
```diff
@@ -115,3 +115,32 @@ See [PreprovisioningImage
...
+
+The **BareMetalHost** resource supports setting custom properties on the
+provisioner's host representation (e.g., Ironic node properties) via the
+`spec.hostProvisionerProperties` field.
+
+Each provisioner maintains its own allowlist of properties that can be set.
+Properties not in the allowlist are silently ignored. For the Ironic
+provisioner, the currently allowed properties are:
+
+- `vendor` - Override the vendor identifier for the host
```
elfosardo: Unsupported properties are logged at Info level and silently ignored. Users might not realize their property isn't being applied. We should probbaly consider a way to make that visible, for example in the BMH status?

elfosardo (commented):
looks good, just some comments for discussion, and you will probably need a rebase

pkg/provisioner/ironic/register.go
synthe102: Since this implementation already has an allow list I'm not sure of the value added by having a deny list. Unless someone adds an internal key to the allow list, we can't override those properties.

pkg/provisioner/ironic/ironic.go
synthe102: Agree, no value exporting this. Will unexport.

docs/api.md
synthe102: Makes sense ! Updated to PR to add lists of applied/ignored properties keys to the provisioning status.

elfosardo:
@synthe102 you need to rebase this :)

synthe102:
@elfosardo, done 👌

dtantsur (dismissed):
/hold

We have repeatedly rejected calls for providing arbitrary Ironic-level properties, so I'd prefer a proper design discussion to happen for this addition.

I'm not sure I understand why override the vendor and why a better detection cannot be done on the  Ironic level, but this feature seems to be an overkill for that.

dtantsur:
I cannot find more details on the motivation, I assume that some logic in Ironic does not kick in because the vendor does not literally match? I think it does need to be fixed in Ironic. We cannot expect the operators to know which magic strings Ironic detects, especially as they evolve over time.

My second favourite option (honestly, far behind fixing Ironic) is to use annotations somewhat similarly to what you suggested initially:

```yaml
annotations:
  ironic.provisioners.metal3.io/vendor: ami
```

In the end, annotations are very commonly used for this specific use case: customize behaviors that are not a part of the core API contract.

My least favourite option would be an explicit `vendorOverride` in the spec. I don't like it because it adds a whole API feature for one use case that may be gone sooner or later (because Ironic is fixed or because the vendor changes their behavior).

synthe102:
@dtantsur the main motivation for this vendor override is that some vendors (eg: Dell) will override the Redfish vendor endpoint on AMI BMCs to return Dell. Without explicitly setting the vendor property, Ironic can't properly talk to those BMCs.
As far as I know, properly handling those edge cases in Ironic would require some important code changes.

I'm fine going with the annotation instead of the BMO spec changes.
I completely agree that fixing Ironic is the best option, the annotation can be used temporarily and deprecated once Ironic improves its vendor detection.

dtantsur:
> some vendors (eg: Dell) will override the Redfish vendor endpoint on AMI BMCs to return Dell

Sigh. Why cannot we have nice things... Do you think it's something we could detect through a different field (maybe something on Manager)? I guess it's not perfect that in Ironic we equate BMC vendor with the machine vendor, maybe we need to split these fields?

synthe102:
I reported this issue to Dell, without success. I will try to find out more info regarding the detection process.

dtantsur:
@synthe102 could you file a bug describing the original problem and your findings to https://bugs.launchpad.net/ironic so that we can have a discussion on paths forward? (and feel welcome to join the Ironic IRC for that)

I don't mind having an annotation to override the vendor for now. The proposed solution seems to be an overkill though. My only concern with any override approach: Ironic updates the vendor property during cleaning/servicing, it can get overridden at an unfortunate point.

dtantsur:
@synthe102 with https://review.opendev.org/c/openstack/ironic/+/978439 merging, do you still need this PR?

Rozzii:
> @synthe102 with https://review.opendev.org/c/openstack/ironic/+/978439 merging, do you still need this PR?

I assume not, I have closed the issue too, I will close the PR too, @synthe102 also looks a bit inactive, please feel free to reopen the PR if you need it and also please ping me.

PR #2996, opened 2026-02-16
:sparkles: Adds Associate logic for HostClaim

pierrecregut:

Implementation of the choice logic of HostClaims
* Respect the HostDeployPolicy for choosing the BareMetalHost
* Respect annotations set by standard capm3 on BareMetalHosts

The controller implementation for HostClaim will follow the template of Metal3Machine controller with a manager implementing the main actions : Associate, Update, Delete. 
Implementation of Associate : Uses the choice process and modify consumerRef of the BMH and annotate the HostClaim to bind the two resources.

A separate commit provides the tooling for unit tests. It will be shared with the controller.


internal/testing/common.go
```diff
@@ -0,0 +1,343 @@
...
+			},
+		},
+	}
+	for _, option := range options {
+		option.EnhanceBmh(bmh)
+	}
+	return bmh
+}
+
+func (bl WithLabels) EnhanceBmh(bmh *metal3api.BareMetalHost) {
```
dtantsur: A more common pattern for these sort of things (which we also have in our testing code) is the builder pattern:

```go
bmh := testing.BMHBuilder().WithLables(...).WithSomethingElse(...).build()
```

pkg/hostclaim/requeue_error.go
```diff
@@ -0,0 +1,62 @@
...
+limitations under the License.
+*/
+
+package hostclaim
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/pkg/errors"
```
dtantsur: Please avoid the deprecated `pkg/errors` in new code

pkg/hostclaim/hostclaim_manager_test.go
```diff
@@ -0,0 +1,358 @@
...
+
+import (
+	"context"
+	"testing"
+
+	metal3api "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
+	. "github.com/metal3-io/baremetal-operator/internal/testing"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/pkg/errors"
```
dtantsur: Please avoid the deprecated `pkg/errors` in new code

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,542 @@
...
+	"math/big"
+	"regexp"
+	"slices"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/go-logr/logr"
+	metal3api "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
+	"github.com/pkg/errors"
```
dtantsur: Please avoid the deprecated `pkg/errors` in new code

internal/testing/common.go
```diff
@@ -0,0 +1,343 @@
+//go:build unit
```
dtantsur: This approach (which we already use elsewhere unfortunately) throws off tools working with the file, e.g. vim plugins cannot work with the file and fail with `No packages found for open file /home/dtantsur/Projects/baremetal-operator/internal/testing/common.go` on the line `package testing`.

internal/testutil/common.go
```diff
@@ -0,0 +1,343 @@
+//go:build unit
+
+/*
+Copyright 2025 The Metal3 Authors.
```
dtantsur, line 2: nit: I don't think it's a valid copyright notice. If your employer does not require a specific copyright notice, better just omit it and leave only the license.

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,542 @@
...
+	client      client.Client
+	HostClaim   *metal3api.HostClaim
+	Log         logr.Logger
+	PatchHelper *patch.Helper
+	APIReader   client.Reader
+}
+
+const (
+	// PausedAnnotationKey is an annotation to be used for pausing a hostclaim.
+	PausedAnnotationKey = "metal3.io/hostmgr"
```
dtantsur: The paused annotation on BMH looks like `baremetalhost.metal3.io/paused`, I suggest we use something consistent with it (`hostclaim.metal3.io/paused`, I guess)

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,542 @@
...
+	// BareMetalHostAnnotation is the key for an annotation that should go on a Host to
+	// reference what BareMetalHost it corresponds to.
+	BareMetalHostAnnotation = "metal3.io/BareMetalHost"
+	// UnhealthyAnnotation is the annotation used by the Metal3Health
+	// that sets unhealthy status of BMH.
+	UnhealthyAnnotation = "capi.metal3.io/unhealthy"
+	// nodeReuseLabelName is the label set on BMH when node reuse feature is enabled.
+	// and the label set on HostClaim as target for reuse.
+	nodeReuseLabelName = "infrastructure.cluster.x-k8s.io/node-reuse"
```
dtantsur: These three seem to be CAPI-specific, shouldn't CAPM3 handle them instead? I thought the whole cleaning discussion was to have a way to express node reuse without actually adding the node reuse feature to BMO?

pkg/hostclaim/hostclaim_manager_test.go
```diff
@@ -0,0 +1,358 @@
+//go:build unit
```
dtantsur: Please remove this (see common.go for why)

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,542 @@
...
+		return false
+	}
+	if bmh.Labels == nil {
+		return false
+	}
+	_, ok := bmh.Labels[nodeReuseLabelName]
+	return ok
+}
+
+type Set[T comparable] = map[T]struct{}
```
dtantsur, line 438: You can use https://github.com/deckarep/golang-set if you need generic code for sets, but the code here seems trivial enough to be just inlined.

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,542 @@
...
+		return err
+	}
+	if bmh == nil {
+		m.Log.Info("No available host found. Requeuing.")
+		m.SetConditionHostToFalse(
+			metal3api.AssociatedCondition, metal3api.NoBareMetalHostReason,
+			"No available host found: requeuing.")
+		return &RequeueAfterError{RequeueAfter: HostClaimRequeueDelay}
+	}
+	m.Log.Info("Associating machine with host", "BareMetalHost", bmh.Name)
```
dtantsur: s/machine/hostClaim/

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,542 @@
...
+// ensureAnnotation makes sure the machine has an annotation that references the
+// host and uses the API to update the machine if necessary.
+func (m *HostManager) ensureAnnotation(ctx context.Context, bmh *metal3api.BareMetalHost) error {
+	annotations := m.HostClaim.Annotations
+	if annotations == nil {
+		annotations = map[string]string{}
+		m.HostClaim.Annotations = annotations
+	}
+	bmhKey := cache.MetaObjectToName(bmh).String()
+	annotations[BareMetalHostAnnotation] = bmhKey
```
dtantsur: If we need to track the BMH in the HostClaim, please add a new field for that instead of an annotation.

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,542 @@
...
+		for i, bmh := range bmhs.Items {
+			m.Log.Info("Testing host", "name", bmh.Name, "namespace", bmh.Namespace)
+			if bmh.Spec.ConsumerRef != nil && consumerRefMatches(bmh.Spec.ConsumerRef, m.HostClaim) {
+				m.Log.Info("Found host with existing ConsumerRef", "host", bmh.Name)
+				helper, err2 := patch.NewHelper(&bmhs.Items[i], m.client)
+				return &bmhs.Items[i], helper, err2
+			}
+
+			// nodeReuse is not handled at the level of HostClaims but we still
+			// honor nodeReuse labels of others.
```
dtantsur: We end up checking a CAPM3/CAPI specific annotation in BMO, which we have never (to my best knowledge) respected before. Cannot CAPM3 configure its HostClaims to exclude the node reuse annotation?

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,542 @@
...
+				continue
+			}
+
+			// continue if BaremetalHost is paused or marked with UnhealthyAnnotation.
+			annotations := bmh.GetAnnotations()
+			if annotations != nil {
+				if _, ok := annotations[metal3api.PausedAnnotation]; ok {
+					continue
+				}
+				if _, ok := annotations[UnhealthyAnnotation]; ok {
```
dtantsur: Same comment for the unhealthy annotation.

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,542 @@
...
+	}
+	m.Log.Info("Acceptable namespaces", "namespaces", namespaces)
+	return namespaces, nil
+}
+
+func hostLabelSelectorForHostClaim(hostClaim *metal3api.HostClaim, log logr.Logger) (labels.Selector, error) {
+	labelSelector := labels.NewSelector()
+
+	for labelKey, labelVal := range hostClaim.Spec.HostSelector.MatchLabels {
+		log.Info("Adding requirement to match label",
```
dtantsur: This is too verbose, please use `V(1)` if you need this logging

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,542 @@
...
+		r, err := labels.NewRequirement(labelKey, selection.Equals, []string{labelVal})
+		if err != nil {
+			log.Error(err, "Failed to create MatchLabel requirement, not choosing host")
+			return nil, err
+		}
+		labelSelector = labelSelector.Add(*r)
+	}
+
+	for _, req := range hostClaim.Spec.HostSelector.MatchExpressions {
+		log.Info("Adding requirement to match label",
```
dtantsur: same here re logging

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,542 @@
...
+
+func hostLabelSelectorForHostClaim(hostClaim *metal3api.HostClaim, log logr.Logger) (labels.Selector, error) {
+	labelSelector := labels.NewSelector()
+
+	for labelKey, labelVal := range hostClaim.Spec.HostSelector.MatchLabels {
+		log.Info("Adding requirement to match label",
+			"label key", labelKey,
+			"label value", labelVal)
+		r, err := labels.NewRequirement(labelKey, selection.Equals, []string{labelVal})
+		if err != nil {
```
dtantsur, line 339: I wonder if we can check the validness of HostSelector on the webhook level, so that the errors are not caught only during reconciliation.

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,542 @@
...
+}
+
+// For a given HostClaim in a given namespace, computes all the namespaces that can contain BMH
+// that can be bound to the HostClaim.
+//
+// The function lists the HostDeployPolicy and keeps the namespaces of the ones that
+// match the namespace of the claim. overridesCleaningPolicy specifies whether the hostclaim
+// defines the automatedCleaningPolicy field or not.
+func (m *HostManager) acceptableNamespaces(ctx context.Context, namespace string) (Set[string], error) {
+	m.Log.Info("Searching for suitable namespaces")
```
dtantsur: `Log.V(1)`

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,542 @@
...
+	err = m.client.Get(ctx, client.ObjectKey{Name: hostNs}, hostNsResource)
+	if err != nil {
+		m.Log.Error(err, "cannot access the namespace of the claim")
+		return nil, err
+	}
+	nsLabels := hostNsResource.Labels
+	if nsLabels == nil {
+		nsLabels = map[string]string{}
+	}
+	m.Log.Info("labels of the NS of hostclaim", "labels", nsLabels)
```
dtantsur: same (overly verbose logging)

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,542 @@
...
+LOOP_POLICY:
+	for _, hostDeployPolicy := range hostdeploypolicies.Items {
+		log := m.Log.WithValues("policyNamespace", hostDeployPolicy.Namespace, "policyName", hostDeployPolicy.Name)
+		if SetContains(namespaces, hostDeployPolicy.Namespace) {
+			// namespace already added, no reason to check this hdp.
+			continue
+		}
+		constraints := hostDeployPolicy.Spec.HostClaimNamespaces
+		if constraints == nil {
+			log.Info("Rejecting hostdeploypolicy without constraint")
```
dtantsur: nit: inconsistent spelling in messages (let's choose between `hostDeployPolicy`, `HostDeployPolicy` or `host deploy policy`).

dtantsur: A smaller nit: a policy is not "rejected", it can be "ignored", for example (same below)

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,542 @@
...
+			}
+			if !b {
+				log.Info("Rejecting hostdeploypolicy because claim namespace does not match regex")
+				continue
+			}
+		}
+		if constraints.HasLabels != nil {
+			// Behaves as a 'forall' on labels
+			for _, pair := range constraints.HasLabels {
+				m.Log.Info("checking label", "name", pair.Name, "value", pair.Value)
```
dtantsur: `Log.V(1)`

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,542 @@
...
+				if v, ok := nsLabels[pair.Name]; ok {
+					if pair.Value != "" && v != pair.Value {
+						log.Info("Rejecting hostdeploypolicy because claim namespace labels does not have correct value", "label", pair.Name)
```
dtantsur: And here it's actually useful to log the expected and the actual values

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,542 @@
...
+					err, "Error during regexp matching on hostclaim namespace (bad regexp)",
+					"regexp", constraints.NameMatches)
+				return nil, err
+			}
+			if !b {
+				log.Info("Rejecting hostdeploypolicy because claim namespace does not match regex")
+				continue
+			}
+		}
+		if constraints.HasLabels != nil {
```
dtantsur: nit: redundant (you can iterate over `nil`)

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,542 @@
...
+			}
+		}
+		log.Info("Accepting namespace because of HostDeployPolicy")
+		AddSet(namespaces, hostDeployPolicy.Namespace)
+	}
+	m.Log.Info("Acceptable namespaces", "namespaces", namespaces)
+	return namespaces, nil
+}
+
+func hostLabelSelectorForHostClaim(hostClaim *metal3api.HostClaim, log logr.Logger) (labels.Selector, error) {
```
dtantsur: nit: slight inconsistency: instead of passing `m.Log` as `log`, you could make this function a method on HostManager just like `acceptableNamespaces`

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,542 @@
...
+						log.Info("Rejecting hostdeploypolicy because claim namespace labels does not have correct value", "label", pair.Name)
+						continue LOOP_POLICY
+					}
+				} else {
+					log.Info("Rejecting hostdeploypolicy because claim namespace labels does not have a required label", "label", pair.Name)
+					continue LOOP_POLICY
+				}
+			}
+		}
+		log.Info("Accepting namespace because of HostDeployPolicy")
```
dtantsur: It's not clear here which namespace is accepted (and it's probably redundant because of logging on line 400)

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,542 @@
...
+	if err != nil || len(namespaces) == 0 {
+		return nil, nil, err
+	}
+	// get list of BMH.
+	labelSelector, err := hostLabelSelectorForHostClaim(m.HostClaim, m.Log)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// Different from M3M: We do not restrict to a namespace
```
dtantsur: nit: This is not clear for someone who is not familiar enough with CAPM3: we definitely restrict namespaces here

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,542 @@
...
+
+	for namespace := range namespaces {
+		bmhs := metal3api.BareMetalHostList{}
+		m.Log.Info("Testing in namespace", "namespace", namespace)
+		err = m.client.List(ctx, &bmhs, client.MatchingLabelsSelector{Selector: labelSelector}, client.InNamespace(namespace))
+		if err != nil {
+			return nil, nil, err
+		}
+		for i, bmh := range bmhs.Items {
+			m.Log.Info("Testing host", "name", bmh.Name, "namespace", bmh.Namespace)
```
dtantsur: Hmm, this is a useful message, but there may be thousands of hosts... I'm not sure, but it may be a case for `V(1)` too

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,542 @@
...
+			if bmh.Spec.ConsumerRef != nil && consumerRefMatches(bmh.Spec.ConsumerRef, m.HostClaim) {
+				m.Log.Info("Found host with existing ConsumerRef", "host", bmh.Name)
+				helper, err2 := patch.NewHelper(&bmhs.Items[i], m.client)
+				return &bmhs.Items[i], helper, err2
```
dtantsur: nit: `bmhs.Items[i]` is `bmh`

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,542 @@
...
+			// nodeReuse is not handled at the level of HostClaims but we still
+			// honor nodeReuse labels of others.
+			if bmh.Spec.ConsumerRef != nil || nodeReuseLabelExists(&bmh) {
+				continue
+			}
+			if bmh.GetDeletionTimestamp() != nil {
+				continue
+			}
+			if bmh.Status.ErrorMessage != "" {
+				continue
```
dtantsur: On the other hand, this may justify a log message since errors are not expected to be a norm for available hosts

dtantsur: In this case, please move it to under line 495, so that we don't log anything for hosts that e.g. fail inspection

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,542 @@
...
+func (m *HostManager) pickHost(availableHosts []*metal3api.BareMetalHost) (*metal3api.BareMetalHost, error) {
+	var chosenHost *metal3api.BareMetalHost
+	var availableHostsInFailureDomain []*metal3api.BareMetalHost
+
+	// When failureDomain is set, create a list from available hosts in failureDomain
+	if m.HostClaim.Spec.FailureDomain != "" {
+		labelSelector := labels.NewSelector()
+		var reqs labels.Requirements
+		var r *labels.Requirement
+		r, err := labels.NewRequirement(FailureDomainLabelName, selection.Equals, []string{m.HostClaim.Spec.FailureDomain})
```
dtantsur, line 219: Why is this not a part of the initial listing? It seems more efficient to do this on the server side.

dtantsur: ... ah, because we have a fallback to all available hosts. Okay, got it.

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,542 @@
...
+	// From here the hostClaim is definitely associated
+	m.SetConditionHostToTrue(metal3api.AssociatedCondition, metal3api.BareMetalHostAssociatedReason)
+
+	return nil
+}
+
+// PatchHost patch the HostClaim and ensures that the conditions are initialized.
+// Can be used several times without creating a conflict.
+func (m *HostManager) PatchHost(ctx context.Context, options ...patch.Option) error {
+	if m.PatchHelper == nil {
```
dtantsur: I don't quite get what you're trying to achieve with PatchHelper

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,542 @@
...
+func (m *HostManager) SetConditionHostToTrue(
+	t string,
+	reason string,
+) {
+	conditions.Set(m.HostClaim, metav1.Condition{Type: t, Status: metav1.ConditionTrue, Reason: reason, Message: ""})
+}
+
+func (m *HostManager) Associate(ctx context.Context) error {
+	// Parallel attempts to associate is problematic since the same BMH
+	// could be selected for multiple M3Ms. Therefore we use a mutex lock here.
```
dtantsur: nit: s/M3M/host claims/

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,542 @@
...
+	associateBMHMutex.Lock()
+	defer associateBMHMutex.Unlock()
```
dtantsur: This mutex-based approach is probably fine for the first iteration (especially since CAPM3 does it, apparently).

Still, a lot of code is running while holding this lock. We'll need to investigate ways of reducing its scope, especially since it anyway does not cover all cases (something else competing for consumerRef, for instance).

One rough idea is to let the BMH list be assembled without locking, then sort it in a way that depends on the exact HostClaim (I don't know if Go has any primitives for that, maybe using a random number generator seeded from the HostClaim's UID), thus reducing the changes of conflicts when the pool of hosts is large.

I suspect, however, that small deployments (where the number of host claims == number of available hosts) may suffer from retries in this case. Maybe it needs to be an option *somewhere* that the deployer will set based on the proportion of host claims to available hosts.

cc @lentzi90 @kashifest @zaneb for more opinions

pkg/hostclaim/hostclaim_manager.go
pierrecregut: Unfortunately we have to take care of node reuse for capm3 targetting BMH not hostclaims... There is no consumer ref but the BMH is "booked" by the metal3machine.

pkg/hostclaim/hostclaim_manager.go
pierrecregut: As said above, it is not hostClaims node reuse but standard node reuse with metal3machine targetting bmh. Capm3 legacy (ie m3machine targetting bmh, we need a word for it) could put an arbitrary consumerRef on those nodes instead of just an annotation (eg the cluster object) but we should probably handle the transition with care.

pkg/hostclaim/hostclaim_manager.go
pierrecregut: I think it is another topic and a complex one. This lock is not here to resolve somebody else trying to lock the same BMH (conflict resolution will do that)  but it ensures that for this HostClaim (or Metal3Machine) there is not two concurrent reconcile selecting two different BMH with one not managing to update the annotation and not managing to undo its choice. Here, it will have to run after the first one and find that a BMH is already selected (that's also why we have to look at all the potential BMH...) and it will try to set the same value for the annotation.

pkg/hostclaim/hostclaim_manager.go
pierrecregut: This is a direct copy of what Metal3Machine do. My guess is that moving the spec of hostclaim/m3m inside hostclaim/m3m controller is not considered as a good option and status is not permanent enough. The bad news is that we cannot trust this annotation and we need to check the back pointer (consumerRef) in getHost to avoid hostclaims stealing bmh.

pkg/hostclaim/hostclaim_manager.go
zaneb: I believe `bmh` is a _copy of_ `bmhs.Items[i]` because it's a list of structs, not pointers or interfaces - not that there's any way to tell without checking the definition (I hate golang sometimes).

Anyway, I suspect this was done as a defensive coding practice predating https://go.dev/blog/loopvar-preview (although since we are returning here and not continuing with the loop, I don't think it ever would have mattered).

There's a chance that this is the cause of [this issue](https://github.com/metal3-io/cluster-api-provider-metal3/issues/1004) as noted in the above discussion about the mutex though.

pkg/hostclaim/hostclaim_manager.go
zaneb: Distributed systems either work correctly without local mutexes or they are wrong. The mutex doesn't make things less wrong, it just makes the problems harder to reproduce and fix.

Looking at the [original commit](https://github.com/metal3-io/cluster-api-provider-metal3/pull/1012) and [associated issue](https://github.com/metal3-io/cluster-api-provider-metal3/issues/1004), it describes the problem the mutex was solving as multiple M3Ms locking the same BMH. As you note, this is already handled by the write conflict, and the error logs reported in the issue are to be expected in a distributed system. So the real problem - that we apparently don't recover to provision a host if the first one we picked turned out to be taken because the M3M is annotating itself with the name of a BMH that it has not in fact been able to secure - has been covered up by the mutex. Since HostClaim and M3M don't share the same mutex but are competing for the same BMHs, I guess we can expect to see more of those errors soon.

(BTW since the code seems mostly ok - it's clearly not safe to call `setBMCSecretLabel` before `Patch`, but that isn't the problem here - best guess is that we are returning a pointer to the cache data directly, so that when we make changes to the BMH object we are in fact modifying the cache.)

pkg/hostclaim/hostclaim_manager.go
zaneb: In the case of Metal3Machine I suspect it is a holdover from the early versions of CAPI where you just had a single Machine CRD for every platform type, and therefore any platform-specific data you needed to store had to be an annotation.

I can't think of any reason this couldn't be part of the Status.

internal/testing/common.go
pierrecregut: I removed my comment. Ok... options are also pervasive in k8s code. I'll change it if you think that confuse people.

pkg/hostclaim/hostclaim_manager.go
pierrecregut: I do not think the mutex is about the competition between different consumers. Because this is the first thing done: setting the consumerRef and only one wins. But you may concurrently reconcile on the same element because an event fired after you began your first reconcile. The two reconcile will enter the choose and select two different BMHs for a single hostclaim and the recovery may be hard. 
To avoid this without a mutex, you would need to commit on the annotation first... but now you must be ready to roll back if you fail to get the commit on consumerRef and it also implies that you can never trust the annotation (which is already the case but because you do not trust the hostClaim owner) and that getHost must act as a barrier (you thought you were at the update step but you must perform a choose again). 
Because the risk is on a given reconciled object, I do not think there is a problem with different mutex for m3m and hostclaims and you can also have different hostclaim controllers as long as they do not reconcile the same sets of hostclaims (eg. reconcile on a namespace).
I hope we are not returning pointers to the cache. Unless DefaultUnsafeDisableDeepCopy is set, we should be safe. Otherwise explicit deep copies are clearly missing.

pkg/hostclaim/hostclaim_manager.go
pierrecregut: You are probably right because even if we loose status, we can still recover the value during the choose because choose will find the consumerRef.

pkg/hostclaim/hostclaim_manager.go
dtantsur: What I'm objecting to is parts of CAPI API (the annotation literally has CAPI in it) in HostClaims. What you did here can be emulated by HostClaims created with a exclusion rule targeting this annotation.

pkg/hostclaim/hostclaim_manager.go
dtantsur: If that's the case, let's make it more obvious, e.g. by  creating a new variable and commenting on why it's done. Otherwise, someone will remove it eventually.

pkg/hostclaim/hostclaim_manager.go
pierrecregut: The name is a misnomer. I will rename it PausedAnnotationValue. With Metal3Machine (where it is called a key) it is "metal3.io/capm3", I can use "metal3.io/hostclaim".

pkg/hostclaim/hostclaim_manager.go
pierrecregut: Yes I did not want to add an external dependency.

pkg/hostclaim/hostclaim_manager.go
pierrecregut: So I will add a BareMetalHost field next to the HardwareData field. As it is today, their content is the same. But I think we should distinguish them. I will open a point in the hostClaim discussion and probably an issue. There should be at least a mode (if it is not the standard one) where HardwareData are copied from the BMH namespace to the HostClaim namespace and the hostClaim owner has only visibility on his hardwareData (and relevant metadata).

pkg/hostclaim/hostclaim_manager.go
pierrecregut: a. HostClaim have labelSelector, not annotationSelectors.
b. Users should be able to use HostClaims without knowing CAPM3 quirks. 
We can try to change the CAPM3 behaviour so that when it wants to book/disable a host, it always use the consumerRef. But, that will take time to deprecate the old behavior. As long as it is not done, HostClaim implementation must honor capm3 quirks pretty or not.

pkg/hostclaim/hostclaim_manager.go
pierrecregut: I agree  with @zaneb: as we exit, there is no chance to have two conflicting addresses on the same variable. So simplified.

pkg/hostclaim/hostclaim_manager.go
pierrecregut: The major problem of this comment copied from M3M controller is that it is completely misleading.

pierrecregut:
@zaneb, unfortunately because of HostDeployPolicies, the amount of code that can be shared decreases and it will probably decrease more in the near future because we may add additional labelSelectors per HostDeployPolicy.

internal/testing/common.go
pierrecregut: Done. Implementation may be shorter in common but less readable from my point of view during use.

pkg/hostclaim/hostclaim_manager.go
pierrecregut: done (all three)

pkg/hostclaim/hostclaim_manager_test.go
pierrecregut: done

pkg/hostclaim/hostclaim_manager.go
pierrecregut: done with the following comments on the choose function. Some were removed. Otherwise pushed to verbosity 1

pkg/hostclaim/hostclaim_manager.go
pierrecregut: done

pkg/hostclaim/hostclaim_manager.go
pierrecregut: improved message

pkg/hostclaim/hostclaim_manager.go
pierrecregut: Agreed removed

pkg/hostclaim/hostclaim_manager.go
pierrecregut: PatchHost can be used as a defered function at the end of the controller or to ensure the HostClaim is clearly associated to a BMH. If we have two calls to the helper but the first fails the second will be done without the revision set and so without conflict management...

pkg/hostclaim/hostclaim_manager.go
dtantsur: Mmm, okay, I'd be happier with something like `IgnoredAnnotations` on host claims than hardcoding stuff specific to one consumer of HostClaims.

pkg/hostclaim/hostclaim_manager.go
pierrecregut: done (misread it the first time so it was not corrected in the first iteration).

pkg/hostclaim/hostclaim_manager.go
zaneb: It seems to me that CAPM3 is doing the Wrong Thing here by claiming to own the BMH while unsetting the consumerRef. The annotation can only protect the BMH from other Metal3Machines claiming it (unless, as we're seeing here, every other controller that interacts with BMHs is encoded with knowledge of CAPM3's quirks). Our design principles [state](https://github.com/metal3-io/metal3-docs/blob/main/design/api-design-principles.md#dont-assume-machine-api) that we must not assume that CAPM3 is the only potential consumer of a host, but that's what it is doing here.

There's no validation on the consumerRef that it matches a real object, so I don't see any reason to remove it just because e.g. the M3M has been deleted.

> that will take time to deprecate the old behavior

Users will have to update baremetal-operator to get the HostClaim implementation. Do we not have any restrictions on which versions of CAPM3 and BMO can work together?

pkg/hostclaim/hostclaim_manager.go
zaneb: > I do not think the mutex is about the competition between different consumers.

The issue that this code was introduced to fix is extremely clear that it _is_ about that, and includes logs showing two different consumers claiming to own the same BMH:

> ```
> $ kubectl get m3m -n metal3 -o jsonpath="{.items[*].metadata.annotations.metal3\.io/BareMetalHost}"
> metal3/node-1 metal3/node-2 metal3/node-3 metal3/node-3
> ```

so I don't know what to tell you.

> But you may concurrently reconcile on the same element because an event fired after you began your first reconcile.

That's [not a thing](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/internal/controller/controller.go#L312).

pkg/hostclaim/hostclaim_manager.go
pierrecregut: The metal3machine operator choosing directly BMH **is not** a hostclaim consumer. It is just the major user of of BMH and ignoring its quirks and pushing their handling to hostclaim users is not the way to go. We can fill a bug to make the contract cleaner in the cluster-api provider side (clearly for the node-reuse case, there should be a consumerRef). 

For the unhealthy annotation, I am uneasy with the concept : a node with a given OS  and config cannot be remediated, why should it be considered as unhealthy at hardware level ? On the other hand if it "fails" with different clusters/customers, yes it is probably better to mark it as unusable but how to count and reset this phenomenon ? It it is possible, there should be a clear procedure to handle such BMH independently of the consumer.

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,527 @@
...
+			"No available host found: requeuing.")
+		return &RequeueAfterError{RequeueAfter: HostClaimRequeueDelay}
+	}
+	m.Log.Info("Associating hostClaim with host", "BareMetalHost", bmh.Name)
+
+	// First we record the association in the BMH. If we fail, we must redo the
+	// whole selection process and remove the annotation.
+	m.setBmhConsumerRef(bmh)
+
+	if err = helper.Patch(ctx, bmh); err != nil {
```
zaneb: @dtantsur pointed out to me (correctly) that doing `Patch()` here instead of `Update()` is wrong because it ignores any changes made by other controllers since we read the BMH and just stomps over any existing consumerRef. This is why a mutex is required to get it to behave. Presumably in the CAPM3 bug the owner references got stomped on too, and then the first M3M later (in `Update()`) attempted to re-add itself as an owner (after seeing the other M3M as an owner), resulting in the error we see in the logs - after both M3Ms were annotated as owning the same host.

pkg/hostclaim/hostclaim_manager.go
dtantsur: See https://github.com/metal3-io/cluster-api-provider-metal3/issues/1004#issuecomment-3965973079 and the discussion below: PatchHelper is probably a wrong solution to an incorrectly understood problem. We may not need to have it at all if we fix the issues as Zane explained in his comments.

pierrecregut: I am sincerely sorry for my incorrect understanding, you are right. I thought, as there was a lot of code on capturing retryOnConflict errors, that patchHelper was just a more compact Update and sent the resource version in the patch (MergeFromWithOptimisticLock).

So yes the solution is to use update as soon as we get the chosen BMH and then use patch(Helper) if we want to as long as the object is owned. This modification will need to be propagated to capm3, otherwise we may have problems on deployments where hostclaims coexist with metal3machines directly booking bmh.

And yes then, the code can live without locks.

zaneb: On reflection, I suspect the reason that Patch() was used is because we were unconditionally doing an Update() of the BMH when nothing had changed. If the patch is empty then Patch() won't make an API call. So rather than an event filter, we should just check if we actually did anything before we call Update(). Then we can get rid of Patch() everywhere, since it's quite unsafe.

zaneb: I opened metal3-io/cluster-api-provider-baremetal#3189 to request a fix to CAPM3.

pkg/hostclaim/hostclaim_manager.go
dtantsur: > pushing their handling to hostclaim users

In this case, CAPM3 is *the* HostClaim user. It's responsible for configuring HostClaims correctly.

> Do we not have any restrictions on which versions of CAPM3 and BMO can work together?

IIRC, by default we tightly couple versions of all components with CAPM3 version. It, of course, does not prevent users from installing and upgrading components using a different process. Yet, it's not unrealistic to prescribe something like "If you intend to use the HostClaims API in the same cluster as CAPM3, you must update CAPM3 to $this_version first, otherwise conflicts are possible."

> It it is possible, there should be a clear procedure to handle such BMH independently of the consumer.

I agree. Maybe we could add something like "claiming taints" (exact name tbd): a list of namespaced things that prevent a host from getting claimed, very similar to finalizers:

```
spec:
  taints:
    - capi.metal3.io/unhealthy
```

(if anyone is aware of prior art in other controllers, please speak up)

pkg/hostclaim/hostclaim_manager.go
dtantsur: If we solve the patch issue, we most likely don't need the mutex. At least, I'd not start with adding a Mutex - this will it will stuck forever, and nobody will dare removing it later.

pkg/hostclaim/hostclaim_manager.go
dtantsur: For the sake of making progress while the CAPM3 discussion is ongoing, I'd suggest dropping CAPI bits from this PR (except for the already agreed failure domain) and revisit them later.

It would be good to discuss the issue on the next meeting. Will you be there?

pierrecregut: Probably. I have opened https://github.com/metal3-io/cluster-api-provider-metal3/issues/3190 that may be triaged. I have not done anything for the unhealthy part yet.

pkg/hostclaim/hostclaim_manager.go
pierrecregut: New version uploaded uses update and no mutex.

pkg/hostclaim/hostclaim_manager.go
pierrecregut: Removed references to capm3 labels and annotations

pierrecregut:
Note an additional commit to solve new linter issues with go-1.25 (rebase).

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,502 @@
...
+	"crypto/rand"
+	"errors"
+	"math/big"
+	"regexp"
+	"slices"
+	"strings"
+	"time"
+
+	"github.com/go-logr/logr"
+	metal3api "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
```
dtantsur, line 30: nit: we tend to group local imports separately in the end of the list

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,502 @@
...
+	// Small delay on conflict error.
+	ConflictRequeueDelay = time.Millisecond * 100
+	// FailureDomainLabelName is a label name for FailureDomains.
+	FailureDomainLabelName = "infrastructure.cluster.x-k8s.io/failure-domain"
+	// HostClaimKind is the name of the kind.
+	HostClaimKind = "HostClaim"
+)
+
+// An error used when there is no BMH satisfying the constraints.
+var ErrNoAvailableBMH = errors.New("no available BMH")
```
dtantsur: nit: if this is user-visible, let's spell out BareMetalHost

pkg/hostclaim/requeue_error.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"errors"
+	"fmt"
+	"time"
+)
+
+// HasRequeueAfterError represents that an actuator managed object should
+// be requeued for further processing after the given RequeueAfter time has
+// passed.
+type HasRequeueAfterError interface {
```
dtantsur: Why having an interface with only one implementation?

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,502 @@
...
+				metal3api.AssociatedCondition, metal3api.NoBareMetalHostReason,
+				"No available host found: requeuing.")
+			return &RequeueAfterError{RequeueAfter: HostClaimRequeueDelay}
+		}
+		return err
+	}
+	m.Log.Info("Associating hostClaim with host", "BareMetalHost", bmh.Name)
+
+	// First we record the association in the BMH. If we fail, we must redo the
+	// whole selection process and remove the annotation.
```
dtantsur: s/remove the annotation/remove the BMH link from status/

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,502 @@
...
+	)
+	err := m.PatchHelper.Patch(ctx, m.HostClaim, options...)
+	if err != nil {
+		// Deactivate pathHelper so that it cannot be reused
+		m.PatchHelper = nil
+	}
+	return err
+}
+
+func (m *HostManager) setBmhConsumerRef(bmh *metal3api.BareMetalHost) {
```
dtantsur: This function is trivial, please inline it

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,502 @@
...
+
+// SetConditionHostToFalse sets Host condition status to False.
+func (m *HostManager) SetConditionHostToTrue(
+	t string,
+	reason string,
+) {
+	conditions.Set(m.HostClaim, metav1.Condition{Type: t, Status: metav1.ConditionTrue, Reason: reason, Message: ""})
+}
+
+func (m *HostManager) Associate(ctx context.Context) error {
```
dtantsur, line 103: Since this function is an internal API, let's add a docstring for it, including the behavior on errors (incl. the RequeueAfterError aspect)

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,502 @@
...
+	}
+	m.Log.Info("Associating hostClaim with host", "BareMetalHost", bmh.Name)
+
+	// First we record the association in the BMH. If we fail, we must redo the
+	// whole selection process and remove the annotation.
+	m.setBmhConsumerRef(bmh)
+
+	if err = m.client.Update(ctx, bmh); err != nil {
+		m.Log.Error(err, "Error while updating the consumerRef on BMH")
+		m.HostClaim.Status.BareMetalHost = nil
```
dtantsur: nit: this value is not set yet

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,502 @@
...
+			metal3api.AssociatedCondition, metal3api.BareMetalHostNotSynchronizedReason,
+			"Failed to set consumer Reference on BareMetalHost")
+		return hideConflictError(err)
+	}
+
+	// Then we record the commitment to this given BMH.
+	err = m.ensureCommitment(ctx, bmh)
+	if err != nil {
+		m.SetConditionHostToFalse(
+			metal3api.AssociatedCondition, metal3api.HostClaimAnnotationNotSetReason,
```
dtantsur: We need to rename the reason to not mention the annotation. Also, if you use BareMetalHostNotSynchronizedReason on line 137, it makes sense to use it here as well

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,502 @@
...
+		Name:       m.HostClaim.Name,
+		Namespace:  m.HostClaim.Namespace,
+		APIVersion: metal3api.GroupVersion.Identifier(),
+	}
+}
+
+func hideConflictError(err error) error {
+	var aggr kerrors.Aggregate
+	if ok := errors.As(err, &aggr); ok {
+		for _, kerr := range aggr.Errors() {
```
dtantsur: Not sure why it does not happen in the CI, but vim shows "loop can be simplified using slices.ContainsFunc" here

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,502 @@
...
+	"sigs.k8s.io/cluster-api/util/conditions"
+	"sigs.k8s.io/cluster-api/util/patch"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type HostManager struct {
+	client      client.Client
+	HostClaim   *metal3api.HostClaim
+	Log         logr.Logger
+	PatchHelper *patch.Helper
```
dtantsur: Please drop this per previous discussion

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,502 @@
...
+			metal3api.SynchronizedCondition,
+			metal3api.ProvisionedCondition,
+			metal3api.AvailableForProvisioningCondition,
+		}},
+		patch.WithStatusObservedGeneration{},
+	)
+	err := m.PatchHelper.Patch(ctx, m.HostClaim, options...)
+	if err != nil {
+		// Deactivate pathHelper so that it cannot be reused
+		m.PatchHelper = nil
```
dtantsur: This all is avoided by not having a global PatchHelper

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,502 @@
...
+
+	// From here the hostClaim is definitely associated
+	m.SetConditionHostToTrue(metal3api.AssociatedCondition, metal3api.BareMetalHostAssociatedReason)
+
+	return nil
+}
+
+// PatchHost patch the HostClaim and ensures that the conditions are initialized.
+// Can be used several times without creating a conflict.
+func (m *HostManager) PatchHost(ctx context.Context, options ...patch.Option) error {
```
dtantsur: nit: PatchHostClaim

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,502 @@
...
+	// FailureDomainLabelName is a label name for FailureDomains.
+	FailureDomainLabelName = "infrastructure.cluster.x-k8s.io/failure-domain"
+	// HostClaimKind is the name of the kind.
+	HostClaimKind = "HostClaim"
+)
+
+// An error used when there is no BMH satisfying the constraints.
+var ErrNoAvailableBMH = errors.New("no available BMH")
+
+func NewHostManager(client client.Client, log logr.Logger, host *metal3api.HostClaim, apireader client.Reader) (*HostManager, error) {
```
zaneb:
```suggestion
func NewHostManager(client client.Client, log logr.Logger, claim *metal3api.HostClaim, apireader client.Reader) (*HostManager, error) {
```

pkg/hostclaim/hostclaim_manager.go
zaneb: Let's just not use `Patch()` ever.

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,502 @@
...
+	m.HostClaim.Status.BareMetalHost = &metal3api.ObjectReference{
+		Namespace: bmh.Namespace,
+		Name:      bmh.Name,
+	}
+	return m.PatchHost(ctx)
+}
+
+// consumerRefMatches returns a boolean based on whether the consumer
+// reference and bareMetalHost metadata match.
+func consumerRefMatches(consumer *corev1.ObjectReference, host *metal3api.HostClaim) bool {
```
zaneb:
```suggestion
func consumerRefMatches(consumer *corev1.ObjectReference, claim *metal3api.HostClaim) bool {
```

pkg/hostclaim/requeue_error.go
pierrecregut: It is taken from an old version of capm3. My guess is to satisfy errors.As but there is also an unneeded pointer to a single field struct. I simplified it.

pkg/hostclaim/hostclaim_manager.go
pierrecregut: done

pkg/hostclaim/hostclaim_manager.go
pierrecregut: it is not but changed it anyway (circumvent return nil, nil disallowed).

pkg/hostclaim/hostclaim_manager.go
pierrecregut: I removed the patch action. It does not bring anything as there is no other reconciliation that could conflict with this one. With it I removed PatchHost which has no other purpose and brought back the exact code that was in metal3machine controller at the controller level (not this commit). So I inlined the two remaining lines of ensureCommitment and there is no more error to treat.

pkg/hostclaim/hostclaim_manager.go
pierrecregut: done

pkg/hostclaim/hostclaim_manager.go
pierrecregut: done

pkg/hostclaim/requeue_error.go
pierrecregut: Note that if T is an error, so is *T. But the As.errors will not work if the recipient is a T and the error an *T. If we use an interface for the type of the recipient, the problem does not occur.

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,447 @@
...
+			m.SetConditionHostToFalse(
+				metal3api.AssociatedCondition, metal3api.NoBareMetalHostReason,
+				"Failed to pick a BaremetalHost for the Host")
+		}
+		if errors.Is(err, ErrNoAvailableBMH) {
+			m.Log.Info("No available host found. Requeuing.")
+			m.SetConditionHostToFalse(
+				metal3api.AssociatedCondition, metal3api.NoBareMetalHostReason,
+				"No available host found: requeuing.")
+			return RequeueAfterError{RequeueAfter: HostClaimRequeueDelay}
```
dtantsur, line 125: It recently came to my attention that we don't use jitters (e.g. https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#Jitter) in our requeue delays. Having one may be quite important for HostClaims: imagine a large number of HCs created for a relatively small number of hosts (especially when some of the hosts are still cleaning): there may be a lot of conflicts, and then the next attempt will happen for all HCs at the same time.

Not a blocking concern, but let's plan to add at least a small (in this context may be 1 second) jitter.

pkg/hostclaim/hostclaim_manager.go
dtantsur: +1

pkg/hostclaim/hostclaim_manager.go
dtantsur: Let's do this, even if in a follow-up?

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,447 @@
...
+		return nil, err
+	}
+
+	// Different from M3M: We do not restrict to a single namespace (namespace of Metal3Machine)
+
+	availableHosts := []*metal3api.BareMetalHost{}
+
+	for namespace := range namespaces {
+		bmhs := metal3api.BareMetalHostList{}
+		m.Log.Info("Testing in namespace", "namespace", namespace)
```
dtantsur: `V(1)` here too

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,447 @@
...
+		err = m.client.List(ctx, &bmhs, client.MatchingLabelsSelector{Selector: labelSelector}, client.InNamespace(namespace))
+		if err != nil {
+			return nil, err
+		}
+		for i, bmh := range bmhs.Items {
+			if bmh.Spec.ConsumerRef != nil && consumerRefMatches(bmh.Spec.ConsumerRef, m.HostClaim) {
+				m.Log.Info("Found host with existing ConsumerRef", "host", bmh.Name)
```
dtantsur: nit: missing namespace in logging

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,447 @@
...
+			}
+
+			if bmh.Status.ErrorMessage != "" {
+				m.Log.Info("Found an available host with an error message (should not occur)",
+					"bmh", bmh.Name, "bmhNamespace", bmh.Namespace)
+				continue
+			}
+
+			m.Log.Info("Host matched hostSelector for Host, adding it to availableHosts list",
+				"host", bmh.Name)
```
dtantsur: nit: missing namespace in logging

pkg/hostclaim/hostclaim_manager.go
pierrecregut: This means checking https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set on keys and values. It could be expressed as CEL rules directly in the custom resource.... but I am afraid that on a list, we quickly exhaust our runtime budget especially with regex. I do not think validating admission policy would add anything either as runtime budget or expressiveness. So the last solution is standard webhook... I would prefer to do it in a completely separate commit. May be we should open an issue to keep track of it ?

pkg/hostclaim/hostclaim_manager.go
pierrecregut: done

pkg/hostclaim/hostclaim_manager.go
pierrecregut: oups mismatched it with inlined function. done.

pkg/hostclaim/hostclaim_manager.go
pierrecregut: I have added some jitter on the hideConflictError. Rationale:
* when there is no host available, it is not really a conflict, everybody can retry at some point, the real question is whether there are new resources or not and it is a slow process.
* when there is a conflict, one wins the other must retry. The hideConflictError avoids the exponential backoff but as you said is too deterministic. The retry is now between 100 ms and 1s.

pkg/hostclaim/hostclaim_manager.go
pierrecregut: done

pkg/hostclaim/hostclaim_manager.go
pierrecregut: done

pkg/hostclaim/hostclaim_manager.go
dtantsur: I'm fine with a follow-up.

pkg/hostclaim/hostclaim_manager.go
dtantsur: > when there is no host available, it is not really a conflict, everybody can retry at some point

Dunno about this. I'm worried about a scenario like this:

You have 50 hosts and want to create 50 host claims. The problem is, 20 hosts are still going through cleaning, so 20 claims retry. After 4*30=120 seconds, 2 hosts are free, so all 20 claims try to reserve them simultaneously, causing a lot of conflicts on the k8s API level. After 30 more seconds, 10 more hosts are free, and again, all 18 remaining claims race for them. In this example, we get 18+8=26 useless API requests, most of which are preventable.

If you add a small jitter (say, a factor of 0.1), you'll make sure that most of the claims will already see that the hosts are reserved and will retry without hitting the Kubernetes API.

dtantsur:
@pierrecregut I think the PR is ready to get merged, and we can continue discussions after that. Could you please squash some commits for a nicer history? It does not have to be only one final commit, but fix-ups should be squashed into the commits they're fixing.

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,450 @@
...
+	}
+
+	return chosenHost, nil
+}
+
+// For a given HostClaim in a given namespace, computes all the namespaces that can contain BMH
+// that can be bound to the HostClaim.
+//
+// The function lists the HostDeployPolicy and keeps the namespaces of the ones that
+// match the namespace of the claim. overridesCleaningPolicy specifies whether the hostclaim
```
dtantsur: nit: there is no overridesCleaningPolicy so far

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,450 @@
...
+	err = m.client.Get(ctx, client.ObjectKey{Name: hostNs}, hostNsResource)
+	if err != nil {
+		m.Log.Error(err, "cannot access the namespace of the claim")
+		return nil, err
+	}
+	nsLabels := hostNsResource.Labels
+	if nsLabels == nil {
+		nsLabels = map[string]string{}
+	}
+	namespaces := NewSet[string]()
```
dtantsur, line 281: Reflecting my question on https://github.com/orgs/metal3-io/discussions/2795#discussioncomment-16181003, shouldn't we always include the HostClaim's own namespace?

pierrecregut: It would be simpler in some cases.... But there is not a lot of reason to use hostclaims if you have BMH in your namespace and  there is also some kind of contract: "no HostDeployPolicy, no risk of use by a hostclaim". I have also not played enough with the multiple namespaces mode of the controller except for e2e test. I would not like to have an added rbac requirement to allow listing BMH on a namespace that is only intended for hostclaims. Right now all the namespaces have the same rbac rules.

pierrecregut:
I have kept three commits : one on the test utilities that have a broader scope, one on the field added to the status of hostclaim resource and finally the core of the feature. @dtantsur, I can squash it further. For example not sure if isolating the change to the custom resource improves the readability.

go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/metal3-io/baremetal-operator/apis v0.5.1
 	github.com/metal3-io/baremetal-operator/pkg/hardwareutils v0.5.1
 	github.com/metal3-io/ironic-standalone-operator/api v0.8.0
+	github.com/onsi/ginkgo/v2 v2.28.0
```
dtantsur, line 13: This is a constant source of merge conflicts...

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,449 @@
...
+
+	return chosenHost, nil
+}
+
+// For a given HostClaim in a given namespace, computes all the namespaces that can contain BMH
+// that can be bound to the HostClaim.
+//
+// The function lists the HostDeployPolicy and keeps the namespaces of the ones that
+// match the namespace of the claim.
+func (m *HostManager) acceptableNamespaces(ctx context.Context, namespace string) (Set[string], error) {
```
lentzi90, line 258: I am missing an explanation for the `namespace` argument. From the usage, I conclude that it is an "external requirement", as in the `HostClaim.Spec.HostSelector.InNamespace`. So `acceptableNamespaces` looks for HostDeployPolicies in all namespaces if empty, otherwise in this specific namespace.
Is this right? Please add to the doc string an explanation of it

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -0,0 +1,449 @@
...
+		return nil, err
+	}
+
+	// Different from M3M: We do not restrict to a single namespace (namespace of Metal3Machine)
+
+	availableHosts := []*metal3api.BareMetalHost{}
+
+	for namespace := range namespaces {
+		bmhs := metal3api.BareMetalHostList{}
+		m.Log.V(1).Info("Testing in namespace", "namespace", namespace)
```
lentzi90:
```suggestion
		m.Log.V(1).Info("Looking for BMHs in namespace", "namespace", namespace)
```

PR #2997, opened 2026-02-18
:bug: Make BMH detach handling more robust

hroyrh:
Check and update the detach status before initiating host delete so that we have the latest OperationalStatus to work with when deleting the host. This will prevent race conditions and unexpected behaviors due to stale status.


elfosardo (commented):
should we update unit tests?

Rozzii (commented):
Approved from my side but indeed please make or modify the unit tests and please rebase.

Rozzii:
@hroyrh  this PR went stale, are you planning to continue the work on this?

hroyrh:
/remove-lifecycle stale
I will look into it next week and update here.

PR #2998, opened 2026-02-18
:seedling: Bump golang to v1.25

kashifest:
bumps golang to v1.25 since v1.24 is EOL
As a side effect also bumps the golangci-lint to v2.10.1 and fixes the linting issues.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Support FROM override
-ARG BUILD_IMAGE=docker.io/golang:1.25.7@sha256:sha256:85c0ab0b73087fda36bf8692efe2cf67c54a06d7ca3b49c489bbff98c9954d64
+ARG BUILD_IMAGE=docker.io/golang:1.25.7@sha256:85c0ab0b73087fda36bf8692efe2cf67c54a06d7ca3b49c489bbff98c9954d64
 ARG BASE_IMAGE=gcr.io/distroless/static:nonroot@sha256:9ecc53c269509f63c69a266168e4a687c7eb8c0cfd753bd8bfcaa4f58a90876f
 
 # Build the manager binary
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -36,5 +36,5 @@ jobs:
     - name: golangci-lint-${{matrix.working-directory}}
       uses: golangci/golangci-lint-action@1e7e51e771db61008b38414a730f564565cf7c20 # v9.2.0
       with:
-        version: v2.1.6
+        version: v2.10.1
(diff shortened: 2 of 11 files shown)


kashifest:
/hold
We need to take this PR https://github.com/metal3-io/project-infra/pull/1254  in first for the tests to pass here.

kashifest:
> Do we need to bump in the go.mod files or is that just a precaution to avoid having to bump on the release branch later?

Yes precisely. And now that 1.24 is EOL, new deps would also bring in go 1.25 as a requirement like for example cert manager is asking https://github.com/metal3-io/baremetal-operator/pull/2973/, so IMO main branch go mod can and should bump to the supported go version

Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Support FROM override
-ARG BUILD_IMAGE=docker.io/golang:1.24.13@sha256:c29cdf32d47053ab0d914852d9c2ed2da12b3cf13079aaef1704ef21335e68a3
+ARG BUILD_IMAGE=docker.io/golang:1.25.7@sha256:sha256:85c0ab0b73087fda36bf8692efe2cf67c54a06d7ca3b49c489bbff98c9954d64
```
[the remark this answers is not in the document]
kashifest: Fixed

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/metal3-io/baremetal-operator
 
-go 1.24.0
+go 1.25.0
```
dtantsur, line 3: Note: normally we shouldn't bump versions in go.mod. It will happen anyway because of Gophercloud, but maybe wait for it?

lentzi90: I asked about it above also and we discussed it briefly offline. The thing is, we don't want to make these bumps on release branches since it is kind of breaking for consumers (as you know). We then thought that it is much better to do the bump now on main, before we cut the next release. Otherwise we could end up with an old (unsupported) go version on the next release branch while all dependencies start bumping their required version. :slightly_frowning_face: That would be painful.

MahnoorAsghar:
@kashifest We probably need to update .github/workflows/golangci-lint.yaml L39, to fix the golangci-lint errors
For the generate, gomod, and test jobs, merging https://github.com/metal3-io/project-infra/pull/1254 will help, AFAIU.

kashifest:
> @kashifest We probably need to update .github/workflows/golangci-lint.yaml L39, to fix the golangci-lint errors For the generate, gomod, and test jobs, merging [metal3-io/project-infra#1254](https://github.com/metal3-io/project-infra/pull/1254) will help, AFAIU.

Oh thanks good catch, I fixed the local lint test and didn't notice that the workflow needed update as well. Fixed now.

PR #3002, opened 2026-02-20
 🌱Add first fuzzing test for bmc address parsing

smoshiur1237:
This PR will add the following fuzzing test :
`FuzzGetParsedURL` - Tests BMC address parsing with various formats (ipmi, redfish, libvirt, IPv4/IPv6)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -122,6 +122,21 @@ unit-cover: ## Run unit tests with code coverage
 unit-verbose: ## Run unit tests with verbose output
 	TEST_FLAGS=-v make unit
 
+FUZZ_TIME ?= 30s
+
+.PHONY: fuzz
+fuzz: ## Run fuzz tests with seed corpus (no fuzzing, regression test only)
+	cd test/fuzz && go test --race -v ./...
+
+.PHONY: fuzz-run
+fuzz-run: ## Run all fuzz tests sequentially with fuzzing enabled (use FUZZ_TIME=duration)
+	@echo "Discovering fuzz tests..."
+	@cd test/fuzz && go test -list='Fuzz.*' | grep '^Fuzz' | while read -r fuzz_test; do \
+		echo "Running $$fuzz_test for $(FUZZ_TIME)..."; \
+		go test -fuzz=$$fuzz_test -fuzztime=$(FUZZ_TIME) || exit 1; \
+	done
+	@echo "All fuzz tests completed successfully!"
+
 ARTIFACTS ?= ${ROOT_DIR}/test/e2e/_artifacts
 
 .PHONY: test-e2e
--- a/test/fuzz/README.md
+++ b/test/fuzz/README.md
@@ -1,0 +1,40 @@
+# Fuzzing
+
+## Running Fuzz Tests
+
+### Quick Start with Makefile
+
+The easiest way to run fuzz tests is using the Makefile targets:
+
+```bash
+# Run fuzz tests as regression tests (using seed corpus only, fast)
+make fuzz
+
+# Run all fuzz tests sequentially with fuzzing enabled (default: 30 seconds each)
+make fuzz-run
+
+# Run all fuzz tests for custom duration (e.g., 5 minutes each)
+make fuzz-run FUZZ_TIME=5m
+```
+
+The `fuzz-run` target automatically discovers and runs all fuzz tests by
+iteration, dedicating the specified time to each test.
+
+### Crash Corpus and Regression Testing
+
+When fuzzing discovers a crash, Go automatically saves the failing input to
+`testdata/fuzz/<FuzzTestName>/` in the test directory. These crash files should
+be committed to the repository:
+
+```bash
+git add test/fuzz/testdata/
+git commit -m "Add fuzz crash corpus"
+```
+
+Once committed, these crashes are automatically replayed as regression tests
+when running `make fuzz` (or `go test` without `-fuzz`).
+
+## Resources
+
+- [Go Fuzzing Documentation](https://go.dev/doc/fuzz/)
+- [Go Fuzzing Tutorial](https://go.dev/security/fuzz/)
--- a/test/fuzz/bmc_fuzz_test.go
+++ b/test/fuzz/bmc_fuzz_test.go
@@ -1,0 +1,73 @@
+package fuzz
+
+import (
+	"testing"
+
+	"github.com/metal3-io/baremetal-operator/pkg/hardwareutils/bmc"
+)
+
+func FuzzGetParsedURL(f *testing.F) {
+	// Add seed corpus
+	f.Add("ipmi://192.168.122.1:6233")
+	f.Add("redfish://192.168.122.1")
+	f.Add("idrac-virtualmedia://192.168.122.1")
+	f.Add("idrac-virtualmedia://192.168.122.1:443")
+	f.Add("ilo5-redfish://192.168.122.1")
+	f.Add("ilo5-redfish://ilo.example.com")
+	f.Add("idrac-redfish://192.168.122.1")
+	f.Add("idrac-redfish://idrac.example.com:443")
+	f.Add("redfish+http://192.168.122.1")
+	f.Add("idrac-virtualmedia+http://192.168.122.1")
+	f.Add("idrac-virtualmedia+http://192.168.122.1:443")
+	f.Add("ilo5-redfish+http://192.168.122.1")
+	f.Add("ilo5-redfish+http://ilo.example.com")
+	f.Add("idrac-redfish+http://192.168.122.1")
+	f.Add("idrac-redfish+http://idrac.example.com:443")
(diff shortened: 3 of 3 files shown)


tuminoid:
Thanks @smoshiur1237 , I'll take a look a look later this week.

tuminoid:
> Thanks @smoshiur1237 , I'll take a look a look later this week.

/hold

tuminoid:
> > Thanks @smoshiur1237 , I'll take a look a look later this week.
> 
> /hold

Looking at it this week, pinky promise.

test/fuzz/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+### Manual Execution
+
+You can also run fuzz tests directly with `go test`:
+
+```bash
+cd test/fuzz
+
+# Run all fuzz tests for 30 seconds each
+go test -fuzz=. -fuzztime=30s
```
tuminoid: This only works as there is exactly one fuzz function right now, it will fail immediately when there is more.

Makefile
```diff
@@ -122,6 +122,18 @@ unit-cover: ## Run unit tests with code coverage
 unit-verbose: ## Run unit tests with verbose output
 	TEST_FLAGS=-v make unit
 
+FUZZ_TIME ?= 30s
+FUZZ_TARGET ?= .
```
tuminoid: Same, the dot cannot be the default. The same issue slipped through to ip-address-manager repo.

We need a target that fuzzes all the targets by iteration.

test/fuzz/bmc_fuzz_test.go
```diff
@@ -0,0 +1,41 @@
...
+			t.Fatalf("GetParsedURL(%q) returned nil URL without error", address)
+		}
+
+		// Attempt round-trip parsing for well-formed URLs
+		// Skip if hostname or scheme is empty
+		if parsedURL.Hostname() != "" && parsedURL.Scheme != "" {
+			_, err := bmc.GetParsedURL(parsedURL.String())
+			// Round-trip errors are acceptable for some edge cases
+			// We just verify it doesn't panic
+			_ = err
```
tuminoid: Fuzzer is designed to catch panics, it doesn't need second round. Either make it meaningful (compare parsedURL.Hostname() equals reparsed.Hostname(), assert scheme consistency, etc. — like IPAM's invariant checks), or remove the block since it's dead weight that gives a false impression of testing.

test/fuzz/bmc_fuzz_test.go
```diff
@@ -0,0 +1,41 @@
...
+func FuzzGetParsedURL(f *testing.F) {
+	// Add seed corpus
+	f.Add("")
+	f.Add("192.168.122.1")
+	f.Add("192.168.122.1:6233")
+	f.Add("ipmi://192.168.122.1:6233")
+	f.Add("redfish://192.168.122.1")
+	f.Add("libvirt://192.168.122.1:6233/?abc=def")
+	f.Add("[fe80::fc33:62ff:fe83:8a76]:6233")
+	f.Add("my.favoritebmc.com")
```
tuminoid: GetParsedURL dispatches to factories registered by scheme, and BMO registers many more:  idrac-virtualmedia, irmc, ilo5-redfish, idrac-redfish, redfish+https, redfish-virtualmedia, ilo5-virtualmedia, redfish-uefihttp. Without seeds containing those schemes, the fuzzer starts from a narrow input space and has to "discover" those scheme strings by random mutation — which is extremely unlikely for multi-character strings like idrac-virtualmedia.

Give it more combinations to start with.

test/fuzz/bmc_fuzz_test.go
```diff
@@ -0,0 +1,41 @@
...
+		parsedURL, err := bmc.GetParsedURL(address)
+		if err != nil {
+			// Error is valid, skip
+			return
+		}
+
+		// If no error, URL must be non-nil
+		if parsedURL == nil {
+			t.Fatalf("GetParsedURL(%q) returned nil URL without error", address)
+		}
```
tuminoid, line 62:
```suggestion
		if parsedURL == nil {
			t.Fatalf("GetParsedURL(%q) returned nil URL without error", address)
		}
       if parsedURL.Scheme == "" {
           t.Error("successful parse should always have a scheme")
       }
       if parsedURL.Hostname() == "" {
           t.Error("successful parse should always have a hostname")
       }
```

Add some more tests that the returned stuff is actually valid.

Makefile
```diff
@@ -122,6 +122,18 @@ unit-cover: ## Run unit tests with code coverage
...
+
+.PHONY: fuzz
+fuzz: ## Run fuzz tests with seed corpus (no fuzzing, regression test only)
+	cd test/fuzz && go test -v ./...
```
tuminoid:
```suggestion
	cd test/fuzz && go test  -race -v ./...
```
Future-proofing the fuzzing for race conditions. The current fuzz tests isn't benefitting, but there is no real downside either.

test/fuzz/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Running Fuzz Tests
+
+### Quick Start with Makefile
+
+The easiest way to run fuzz tests is using the Makefile targets:
+
+```bash
+# Run fuzz tests as regression tests (using seed corpus only, fast)
+make fuzz
```
tuminoid, line 11: This should explain how the found crashes should be commited to repo to be replayed as part of the regression suite.

```
Commit any files generated in testdata/fuzz/ to preserve regression cases.

File: test/fuzz/bmc_fuzz_test.go (hunk 5 above)
Comment (smoshiur1237) [reply test/fuzz/bmc_fuzz_test.go:62]:
```
parsedURL.Scheme == ""
parsedURL.Hostname() == ""
```
The conditions are casuing error in fuzz test. it is not passing at all.

File: test/fuzz/bmc_fuzz_test.go [hunk 8]
```diff
@@ -0,0 +1,70 @@
+package fuzz
+
+import (
+	"testing"
+
+	"github.com/metal3-io/baremetal-operator/pkg/hardwareutils/bmc"
+)
+
+func FuzzGetParsedURL(f *testing.F) {
+	// Add seed corpus
```
tuminoid, line 10: Were still missing seeds with `+http` and `+https`, which the fuzzer is unlikely to guess.

test/fuzz/bmc_fuzz_test.go
```diff
@@ -0,0 +1,70 @@
...
+			validSchemes := map[string]bool{
+				"ipmi": true, "redfish": true, "redfish+https": true,
+				"idrac-virtualmedia": true, "ilo5-virtualmedia": true, "redfish-virtualmedia": true,
+				"idrac-redfish": true, "ilo5-redfish": true,
+				"irmc": true, "libvirt": true, "redfish-uefihttp": true,
+			}
+			if !validSchemes[parsedURL.Scheme] {
+				t.Logf("Parsed URL with unrecognized scheme: %q from input %q", parsedURL.Scheme, address)
+			}
+		}
```
tuminoid: We can remove this part. This is more like a unit test instead of fuzzing which catches crashes etc. We already test for nil and accessing which might crash.

test/fuzz/README.md
```diff
@@ -0,0 +1,41 @@
...
+iteration, dedicating the specified time to each test.
+
+### Crash Corpus and Regression Testing
+
+When fuzzing discovers a crash, Go automatically saves the failing input to
+`testdata/fuzz/<FuzzTestName>/` in the test directory. These crash files should
+be committed to the repository:
+
+```bash
+# After a crash is found, commit the new testdata files
```
tuminoid: This is same comment as the text paragraph above.

test/fuzz/bmc_fuzz_test.go
smoshiur1237: Had offline discussion with @Rozzii  and updated the sid accordingly

test/fuzz/bmc_fuzz_test.go
```diff
@@ -0,0 +1,79 @@
...
+	f.Add("irmc://192.168.122.1")
+	f.Add("irmc://192.168.122.1:8080")
+	f.Add("ilo5-redfish://192.168.122.1")
+	f.Add("ilo5-redfish://ilo.example.com")
+	f.Add("idrac-redfish://192.168.122.1")
+	f.Add("idrac-redfish://idrac.example.com:443")
+	f.Add("redfish+http://192.168.122.1")
+	f.Add("idrac-virtualmedia+http://192.168.122.1")
+	f.Add("idrac-virtualmedia+http://192.168.122.1:443")
+	f.Add("irmc+http://192.168.122.1")
```
Rozzii: IRMC is deprecated, if you want to future proof it it might be a better idea to not include it.

smoshiur1237: Done.

PR #3012, opened 2026-02-23
:seedling: add smoshiur1237 as reviewer

Rozzii:
Moshiur has 20+ merged commits and working with Metal3 for many years, I think we should add him as reviewer.

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -15,5 +15,6 @@ aliases:
   - iurygregory
   - peppi-lotta
   - s3rj1k
+  - smoshiur1237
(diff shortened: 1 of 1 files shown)


OWNERS_ALIASES
```diff
@@ -17,3 +17,4 @@ aliases:
   - s3rj1k
   - tuminoid
   - zhouhao3
+  - smoshiur1237
```
kashifest: this is maintained alphabatically, so can you instead add Moshiur before Tuomo

Rozzii: Done.

PR #3024, opened 2026-02-26
🌱 Prevent duplicate log watchers in E2E tests

lentzi90:
We sometimes use BuildAndApplyKustomization multiple times, e.g. inside
FlakeAttempt. This is an issue for the log watchers, since we create a
new watcher each time. Each watcher writes logs to the files so we end
up with the logs duplicated. To fix this, keep track of which watchers
we have and skip setting up duplicates. Additionally, clean up the
watchers after the test is done, to avoid leaking and spamming.

I think we use FlakeAttempts only during upgrade tests, so don't expect to see a difference in the normal tests.

Part of #2927

**Checklist:**

### DIFF
--- a/test/e2e/common.go
+++ b/test/e2e/common.go
@@ -129,6 +129,54 @@ func CopyIncrementalVMLog(vmName, destFolder string) error {
 	return nil
 }
 
+// logWatcherManager manages deployment log watchers, allowing per-watcher
+// cancellation and preventing duplicates. Each watcher gets its own child
+// context so it can be individually stopped without affecting others.
+type logWatcherManager struct {
+	mu       sync.Mutex
+	watchers map[string]context.CancelFunc
+}
+
+// logWatchers is the package-level instance used to track all deployment log watchers.
+var logWatchers = &logWatcherManager{
+	watchers: make(map[string]context.CancelFunc),
+}
+
+// Start creates a log watcher for the given key if one doesn't already exist.
+// It creates a child context from the provided ctx so the watcher can be
+// individually cancelled without affecting others.
+func (m *logWatcherManager) Start(ctx context.Context, key string, startFunc func(ctx context.Context)) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	if _, exists := m.watchers[key]; exists {
+		Logf("Log watcher already active for %s, skipping", key)
+		return
+	}
+	watcherCtx, cancel := context.WithCancel(ctx)
+	m.watchers[key] = cancel
+	startFunc(watcherCtx)
+}
+
+// Stop cancels and removes the watcher for the given key.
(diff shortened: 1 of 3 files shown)


lentzi90:
/hold
BMO and Ironic logs are missing here. I must have missed something.

lentzi90:
Now the logs look correct
/hold cancel

lentzi90:
I forgot to run upgrade tests...
/hold

lentzi90:
Log duplication fixes are working. There are more duplications still that we cannot fix here though. See https://github.com/metal3-io/baremetal-operator/issues/2927#issuecomment-3997668549.
The test failure is unrelated. The issue behind that is fixed in #3033.

lentzi90:
Still expected to fail often, until #3033 is merged.

test/e2e/common.go
```diff
@@ -621,13 +669,15 @@ func BuildAndApplyKustomization(ctx context.Context, input *BuildAndApplyKustomi
 	}
```
adilGhaffarDev, line 669: I think this can be make simpler if we just return if WaitForDeploymentsAvailable fails,
and call WatchDeploymentLogsByName only when Deployment is successful. 
What do you think?

lentzi90: Hmm not sure I understand. I think WaitForDeploymentAvailable already fails the test if it fails, so in that case we do not continue.
The thing I am trying to fix here is when we use `FlakeAttempts`, here: https://github.com/metal3-io/baremetal-operator/blob/45602b775b4e68cddd685ce451842a2c0215a6ba/test/e2e/upgrade_test.go#L88

adilGhaffarDev: Does FlakeAttempt always run twice even if first run is successful? 
I think it would be better to use Eventually here. In Eventually as soon as there is success it will not try again.

lentzi90: No it should only run 2 times if the first fails. Hmm maybe I could remove this actually. I think we added it when IPA downloader was very flaky. It may not be needed anymore.
/hold
Let me check this

lentzi90:
Ok it looks like we still need the FlakeAttempts. From what I understand we need it at least when applying BMO, because we create namespaces together with resources in the namespaces, so things fail the first time because the namespaces are missing.

lentzi90:
/ok-to-test
Comon... run the tests

test/e2e/common.go
lentzi90: Looks like we do need FlakeAttempts still. I think it is not really flaky, rather the first fails every time. Maybe we could clean that up in some other way, but for now I think this log watcher fix is ok.
/hold cancel

lentzi90:
/hold
Discussed with @adilGhaffarDev . We will hold for now so we can get the CAPI fix in first. Then it will be easier to say what duplication is coming from where.

PR #3032, opened 2026-02-27
:seedling: Bump golangci-lint to v2.5.0 and fix linter findings

kashifest:
v2.1.6–v2.4.0 fail to install via go install because the upstream dependency github.com/tdakkota/asciicheck was deleted from GitHub. v2.5.0 is the minimum version that resolves the install failure.

The bumped linter also surfaced few linter violations, fixed those as well.

Fixes #

**Checklist:**


smoshiur1237 (commented):
/lgtm

PR #3033, opened 2026-02-27
🌱 E2E: Remove libvirt DHCP reservation from VMs

lentzi90:

We used to set DHCP reservations through libvirt in order to get predictable IPs to use
in the tests (ssh check after provisioning). However, we have since a
while already moved to use user-data for this. Now these DHCP
reservations are redundant and potentially harmful, since they may
enable libvirt DHCP during pre-provisioning, which then conflicts with
the Ironic dnsmasq instance.

This DHCP reservation was still used by the automated cleaning test.
This commit also changes the test to use user-data to set the fixed IP,
in the same way as the other tests already did.

There is still one test that needed the DHCP reservation: Live-ISO.
We cannot use user-data there, so instead the reservation is now made
through the Ironic dnsmasq instance.

Fixes #

**Checklist:**

### DIFF
--- a/test/e2e/common.go
+++ b/test/e2e/common.go
@@ -494,13 +494,13 @@ echo "%s" >> /root/.ssh/authorized_keys`, staticIP, sshPubKeyData)
 }
 
 // createDiskTestUserdata creates a Kubernetes secret with cloud-init userdata for disk operations.
-// This userdata sets up SSH authorized keys and then formats /dev/vdb, mounts it, and creates a test file.
-// Intended for testing automated cleaning of disks.
-func createDiskTestUserdata(ctx context.Context, client client.Client, namespace string, secretName string, sshPubKeyPath string) {
+// This userdata configures a static IP, sets up SSH authorized keys, formats /dev/vdb, mounts it,
+// and creates test files on both disks. Intended for testing automated cleaning of disks.
+func createDiskTestUserdata(ctx context.Context, client client.Client, namespace string, secretName string, sshPubKeyPath string, staticIP string) {
 	sshPubKeyData, err := os.ReadFile(sshPubKeyPath) // #nosec G304
 	Expect(err).NotTo(HaveOccurred(), "Failed to read SSH public key file")
 	userDataContent := fmt.Sprintf(`#!/bin/sh
-# Create the .ssh directory and authorized_keys file
+ip a add %s dev eth0
 mkdir /root/.ssh
 chmod 700 /root/.ssh
 echo "%s" >> /root/.ssh/authorized_keys
@@ -512,7 +512,7 @@ mount /dev/vdb /mnt/data
 
 # Create test files on both disks
 touch /mnt/data/test_file_vdb.txt
-touch /test_file_vda.txt`, sshPubKeyData)
+touch /test_file_vda.txt`, staticIP, sshPubKeyData)
 
 	CreateSecret(ctx, client, namespace, secretName, map[string]string{"userData": userDataContent})
 }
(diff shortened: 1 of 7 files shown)


lentzi90:
/hold
We need to test the optional (upgrade) tests. Setting hold to make sure we don't accidentally merge.

lentzi90:
One pass. The new code did not catch any case when dnsmasq was not ready. Maybe just "happy" coincidence.
Trying again

lentzi90:
Making progress. This currently fixes the flaky upgrade test (based on 2-3 runs) by removing the reserved DHCP leases that we are mostly not using. They are not needed for upgrade tests and they seem to cause problems, possibly because they enable libvirts dnsamasq instance, which then conflicts with ironics.

Removing these DHCP reservations makes automated cleaning and live-iso tests fail though, because they were relying on the reservations. The automated cleaning test is simple enough to fix, but the live-iso is tricky. I need to think about what to do about it. We cannot use user-data for live-iso so some other method is needed.

lentzi90:
All green! I will clean this up now. I think the dnsmasq health check was really a red herring and can be removed. The real issue was the conflict with libvirts dnsmasq that had DHCP enabled.

test/e2e/live_iso_test.go
```diff
@@ -83,15 +83,6 @@ var _ = Describe("Live-ISO", Label("required", "live-iso"), func() {
...
 		}
-		if e2eConfig.GetVariable("SSH_CHECK_PROVISIONED") == "true" {
-			userDataSecretName := "user-data"
-			sshPubKeyPath := e2eConfig.GetVariable("SSH_PUB_KEY")
-			createSSHSetupUserdata(ctx, clusterProxy.GetClient(), namespace.Name, userDataSecretName, sshPubKeyPath, bmc.IPAddress)
-			bmh.Spec.UserData = &corev1.SecretReference{
-				Name:      userDataSecretName,
-				Namespace: namespace.Name,
-			}
-		}
```
lentzi90, line 94: This creates the user-data, which is worthless for Live-ISO.
We rely completely on DHCP reservation to get a predictable IP here.

lentzi90:
I think I have about 5 successful runs now. This is far better than the average across other PRs so it seems this does work.
/cc @Sunnatillo

lentzi90:
Rebased to keep up with latest changes on main.

lentzi90:
/hold cancel
We are good to go!

test/e2e/common.go
```diff
@@ -494,13 +494,13 @@ echo "%s" >> /root/.ssh/authorized_keys`, staticIP, sshPubKeyData)
 }
 
 // createDiskTestUserdata creates a Kubernetes secret with cloud-init userdata for disk operations.
-// This userdata sets up SSH authorized keys and then formats /dev/vdb, mounts it, and creates a test file.
-// Intended for testing automated cleaning of disks.
-func createDiskTestUserdata(ctx context.Context, client client.Client, namespace string, secretName string, sshPubKeyPath string) {
+// This userdata configures a static IP, sets up SSH authorized keys, formats /dev/vdb, mounts it,
```
Rozzii, line 497: Could you please explain why are you  using user data, why not network data or metadata?
I know the user can configure anything with uer data but wouldn't it be more beneficial to test the dedicated network configuration features?

Rozzii: It is not a blocker I a just curious .

lentzi90: It is just because cirros doesn't support anything else. We should switch to some "proper" OS at some point. Then we could test this much better.

PR #3034, opened 2026-02-27
:warning: Remove the iRMC driver, deprecate BMH.Spec.Firmware

dtantsur:
See https://groups.google.com/g/metal3-dev/c/8-rbzNZagqg


s3rj1k:
I wonder if we want to do something about this field with webhooks also.

MahnoorAsghar:
lgtm

PR #3046, opened 2026-03-03
:sparkles: Add BareMetalSwitch CRD and controller for switch config generation

alegacy:
Introduces the BareMetalSwitch (BMS) CRD for defining Top-of-Rack switches managed by the Ironic Networking Service. The controller watches BMS resources and credential secrets, then generates two output secrets:

1. Switch configs secret (IRONIC_SWITCH_CONFIGS_SECRET): a single INI-format file with per-switch sections including address, credentials, driver/device type, and optional fields.

2. Switch credentials secret (IRONIC_SWITCH_CREDENTIALS_SECRET): SSH private key files (keyed by <mac-address>.key) for switches using publickey authentication. Password-authenticated switches store credentials inline in the INI config instead.

The controller is only registered when IRONIC_NETWORKING_ENABLED=true and requires IRONIC_SWITCH_CONFIGS_SECRET, IRONIC_SWITCH_CREDENTIALS_SECRET, and IRONIC_SWITCH_CREDENTIALS_PATH to be set at startup.

Credential secret changes (e.g. password rotation, key replacement) trigger config regeneration via a secret watch with a mapper that filters to only secrets referenced by BMS resources.

Partially implements: https://github.com/metal3-io/metal3-docs/pull/586

**Checklist:**

      note: api docs updated here, but metal3-docs repo will be updated once feature is complete


s3rj1k:
I am starting to wonder if all of this networking should be moved to a dedicated controller (like we have for IPAM) instead of gluing it into BMO.

apis/metal3.io/v1alpha1/baremetalswitch_types.go
```diff
@@ -0,0 +1,120 @@
...
+// +kubebuilder:printcolumn:name="Device Type",type="string",JSONPath=".spec.deviceType",description="Switch device type"
+// +kubebuilder:printcolumn:name="Address",type="string",JSONPath=".spec.address",description="Switch address"
+// +kubebuilder:printcolumn:name="Credential Type",type="string",JSONPath=".spec.credentials.type",description="Credential Type"
+
+// BareMetalSwitch represents a Top-of-Rack switch managed by Ironic Networking.
+type BareMetalSwitch struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec BareMetalSwitchSpec `json:"spec,omitempty"`
```
MahnoorAsghar: no status?

dtantsur: @alegacy it's not in the spec, but it would be nice to have a Condition that is set to True (with appropriate ObservedGeneration) whenever the switch is successfully reconciled into the secret.

config/overlays/e2e/ironic.env
```diff
@@ -1 +1,5 @@
 IRONIC_ENDPOINT=https://192.168.222.2:6385/v1/
+IRONIC_NETWORKING_ENABLED=true
```
dtantsur, line 2: Follow-up request: BMO is capable of reading an IrSO Ironic object to populate its configuration. Ideally, we should support the switch configuration in this way as well (once the IrSO change lands, of course).

internal/controller/metal3.io/baremetalswitch_config.go
```diff
@@ -0,0 +1,256 @@
...
+		configEntries: make(map[string][]byte),
+		keyFiles:      make(map[string][]byte),
+	}
+
+	// Generate config for each switch
+	for i := range switchList.Items {
+		if err := writeSwitchEntry(ctx, c, &switchList.Items[i], credentialsPath, result.configEntries, result.keyFiles); err != nil {
+			var credErr *credentialConfigError
+			if errors.As(err, &credErr) {
+				logger.Info("skipping switch due to credential error",
```
dtantsur, line 93: This seems to be another use case for a Condition (we have prior art in form of `Valid`), otherwise a user cannot detect this situation.

internal/controller/metal3.io/baremetalswitch_config.go
```diff
@@ -0,0 +1,256 @@
...
+					"switch", switchList.Items[i].Name, "error", err)
+				continue
+			}
+			return nil, fmt.Errorf("failed to generate config for switch %s: %w", switchList.Items[i].Name, err)
```
dtantsur, line 98: Same here

internal/controller/metal3.io/baremetalswitch_config.go
```diff
@@ -0,0 +1,256 @@
...
+
+// writeSwitchEntry generates a single switch's INI config entry and adds it
+// to configEntries (keyed by switch name). For publickey-authenticated switches,
+// it also adds the SSH private key to keyFiles (keyed by "<mac-address>.key")
+// and emits a key_file= directive in the config instead of password=.
+func writeSwitchEntry(ctx context.Context, c client.Client, sw *metal3api.BareMetalSwitch, credentialsPath string, configEntries map[string][]byte, keyFiles map[string][]byte) error {
+	var buf bytes.Buffer
+
+	// Section header: [switch:name]
+	fmt.Fprintf(&buf, "[switch:%s]\n", sw.Name)
```
dtantsur: nit: I'd prefer us to use https://pkg.go.dev/text/template to separate the template from data

internal/controller/metal3.io/baremetalswitch_config.go
```diff
@@ -0,0 +1,256 @@
...
+			strconv.FormatBool(*sw.Spec.DisableCertificateVerification))
+	}
+
+	// Fetch credentials from secret
+	secret := &corev1.Secret{}
+	secretKey := types.NamespacedName{
+		Namespace: sw.Namespace,
+		Name:      sw.Spec.Credentials.SecretName,
+	}
+	if err := c.Get(ctx, secretKey, secret); err != nil {
```
dtantsur: We have some hacks to make sure we can fetch secrets without listening for all events on them: see existing usages of secretutils (AcquireSecret)

internal/controller/metal3.io/baremetalswitch_config.go
dtantsur: (or maybe not if this line covers transient errors)

internal/controller/metal3.io/baremetalswitch_controller.go
```diff
@@ -0,0 +1,118 @@
...
+				return ctrl.Result{}, fmt.Errorf("failed to update switch config after deletion: %w", updateErr)
+			}
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, fmt.Errorf("could not load BareMetalSwitch: %w", err)
+	}
+
+	// Regenerate switch config from all BareMetalSwitch resources in the namespace
+	if err := updateSwitchConfigSecret(ctx, r.Client, bmSwitch.Namespace, r.SwitchConfigsSecretName, r.SwitchCredentialSecretName, r.SwitchCredentialPath, logger); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to update switch config: %w", err)
```
dtantsur, line 90: This covers both transient errors (cannot fetch something) and user errors (malformed secret or switch). Returning an error from the controller will cause exponential retries which are only useful in the former case. For user errors it's better to set a condition and return without retrying.

apis/metal3.io/v1alpha1/baremetalswitch_types.go
alegacy: Good point.  Will add it.

internal/controller/metal3.io/baremetalswitch_config.go
alegacy: Good point.  Will change.

internal/controller/metal3.io/baremetalswitch_config.go
alegacy: Ok.  Will look into it.

internal/controller/metal3.io/baremetalswitch_controller.go
alegacy: Good point.  Will change.

internal/controller/metal3.io/baremetalswitch_config.go
alegacy: yes, this should only be API errors.

internal/controller/metal3.io/baremetalswitch_controller.go
alegacy: fixed

internal/controller/metal3.io/baremetalswitch_config.go
alegacy: fixed

internal/controller/metal3.io/baremetalswitch_config.go
alegacy: fixed

internal/controller/metal3.io/baremetalswitch_config.go
alegacy: fixed

apis/metal3.io/v1alpha1/baremetalswitch_types.go
alegacy: fixed

internal/controller/metal3.io/baremetalswitch_config.go
```diff
@@ -0,0 +1,283 @@
...
+	}
+	secret, err := sm.AcquireSecret(ctx, secretKey, sw, false)
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			return &credentialConfigError{msg: fmt.Sprintf("credentials secret %s/%s not found",
+				sw.Namespace, sw.Spec.Credentials.SecretName)}
+		}
+		return fmt.Errorf("failed to get credentials secret %s/%s: %w",
+			sw.Namespace, sw.Spec.Credentials.SecretName, err)
+	}
```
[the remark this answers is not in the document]
alegacy: Unless I've misunderstood something I don't agree with this assessment.  I believe I've set this up similar to how the BMH's BMC secret is setup.   This should work even if the secret is shared across multiple switches as there would be multiple owner references preventing a secret from being deleted until all owners have been deleted.

internal/controller/metal3.io/baremetalswitch_controller.go
```diff
@@ -0,0 +1,134 @@
...
+		meta.SetStatusCondition(&bmSwitch.Status.Conditions, metav1.Condition{
+			Type:               string(metal3api.SwitchConditionReconciled),
+			Status:             metav1.ConditionFalse,
+			ObservedGeneration: bmSwitch.Generation,
+			Reason:             "CredentialError",
+			Message:            credErr,
+		})
+		if statusErr := r.Status().Update(ctx, bmSwitch); statusErr != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to update BareMetalSwitch status: %w", statusErr)
+		}
```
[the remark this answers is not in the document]
alegacy: fixed

internal/controller/metal3.io/baremetalswitch_controller.go
```diff
@@ -0,0 +1,134 @@
...
+	meta.SetStatusCondition(&bmSwitch.Status.Conditions, metav1.Condition{
+		Type:               string(metal3api.SwitchConditionReconciled),
+		Status:             metav1.ConditionTrue,
+		ObservedGeneration: bmSwitch.Generation,
+		Reason:             "Reconciled",
+		Message:            "Switch configuration has been successfully reconciled into the config secret",
+	})
+	if err := r.Status().Update(ctx, bmSwitch); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to update BareMetalSwitch status: %w", err)
+	}
```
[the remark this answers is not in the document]
alegacy: fixed

apis/metal3.io/v1alpha1/baremetalswitch_types.go
```diff
@@ -0,0 +1,152 @@
...
+	// This is currently limited to "password", but will be expanded to others.
+	// +kubebuilder:validation:Enum=password;publickey
+	// +kubebuilder:default=password
+	Type SwitchCredentialType `json:"type"`
+
+	// The name of the secret containing the switch credentials.
+	// For password authentication, requires keys "username" and "password"
+	// For SSH key authentication, requires keys "username" and "ssh-privatekey".
+	// In both cases, an optional "enable-secret" key can be provided if needed
+	// to enable privileged mode.
```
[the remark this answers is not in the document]
alegacy: Good catch copilot, but the mistake was that enable-secret wasn't properly handled in the ini file generation.  Fixed.

main.go
```diff
@@ -411,6 +412,39 @@ func main() {
 		os.Exit(1)
 	}
 
+	networkingEnabled, _ := strconv.ParseBool(os.Getenv("IRONIC_NETWORKING_ENABLED"))
```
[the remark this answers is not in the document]
alegacy: Fixed

internal/controller/metal3.io/baremetalswitch_config.go
```diff
@@ -0,0 +1,291 @@
...
+
+	// Update the switch configs secret
+	if err := updateSecretData(ctx, c, sm, namespace, configSecretName, map[string][]byte{
+		switchConfigKey: configBuf.Bytes(),
+	}); err != nil {
+		return nil, fmt.Errorf("failed to update switch configs secret: %w", err)
+	}
+
+	// Update the switch credentials secret
+	if err := updateSecretData(ctx, c, sm, namespace, credentialSecretName, result.keyFiles); err != nil {
```
tuminoid, line 246: keyFiles only contains keys that were successfully reconciled this cycle. So if there is some (transient) error in switch A, then those keys will be dropped until next reconcile and access is lost meanwhile. That intentional?

internal/controller/metal3.io/baremetalswitch_controller.go
```diff
@@ -0,0 +1,136 @@
...
+				return ctrl.Result{}, fmt.Errorf("failed to update BareMetalSwitch status: %w", statusErr)
+			}
+		}
+		// User error: do not return an error to avoid exponential retries.
+		// Requeue after a delay so that if the credential secret is created
+		// later we will eventually pick it up (the Owns() watch cannot detect
+		// creation of a secret that has no owner reference yet and no label).
+		logger.Info("BareMetalSwitch has credential error, requeueing", "error", credErr)
+		return ctrl.Result{RequeueAfter: credentialErrorRequeueDelay}, nil
+	}
```
tuminoid, line 111: It has a comment, but 10s requeue means for user misconfugration we run reconcile 8640 times a day (360 times an hour). Do we REALLY need 10s requeues?

docs/api.md
```diff
@@ -115,3 +115,21 @@ See [PreprovisioningImage
...
+that is written to a Kubernetes secret. This secret is expected to be mounted
+into the Ironic networking service, which in turn reads it to know how
+to connect to each switch and configure port VLANs during bare metal
+provisioning. Each BareMetalSwitch references a credentials secret containing
+the authentication details needed to access the switch. If a switch's
+credential secret is missing or misconfigured, the controller skips that switch
+and continues generating config for the remaining healthy switches.
+
+See [BareMetalSwitch
+CR](../apis/metal3.io/v1alpha1/baremetalswitch_types.go)
```
tuminoid, line 134: This should point to doc.crds.dev link others, but chicken-egg issue: PR link checker fails as it doesn't exist until this is merged.

For consistency, consider follow-up.

tuminoid (commented):
Generally LGTM, asking some clarifying questions.

internal/controller/metal3.io/baremetalswitch_config.go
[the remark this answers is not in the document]
tuminoid: You're correct IMO. I've seen Copilot trip on the GC mechanics before.

docs/api.md
alegacy: ya, I noticed the chicken-n-egg issue here... I'll circle back to update this later.

internal/controller/metal3.io/baremetalswitch_config.go
alegacy: Yes, that was my intent.  I figured it would be better to have some data written to the config secret rather than no data to the config secret (i.e., 1 bad switch config shouldn't block all of the good ones).   I can change that if there is consensus, but I do think this is the best approach.

internal/controller/metal3.io/baremetalswitch_controller.go
alegacy: @tuminoid I modeled this after the BMH handling of BMC secrets.  It also requeues every 10 secs (`hostErrorRetryDelay`) when the BMC secret cannot be found.   The difference being that I treated all errors the same... I'll change this so that a missing secret is requeued while some sort of user error on processing the secret is not.  

Given that the BMH requeue timer is also 10 seconds should I leave it as-is or do you want me to back it off?  I was aiming for consistency, but do agree that 10 seconds is too fast.   60secs?  5mins?

tuminoid: @dtantsur any opinions ?

internal/controller/metal3.io/baremetalswitch_config.go
tuminoid: I have no strong preference. @Rozzii @dtantsur maybe more hands-on with this feature?

config/overlays/e2e/ironic.env
alegacy, line 2: @dtantsur @tuminoid  basic question about the e2e tests... i started adding other tests for my other PR that will come in the future.  In doing so I've been wondering if the way I did it in this PR is the right approach.  

I've basically set `IRONIC_NETWORKING_ENABLED=true` always so the e2e tests always run this way regardless of whether the test is networking related or not.  Should I split this out into its own overlay and not disturb the existing e2e tests with this variable addition?  

On the other hand, it would be good for the networking code to be exercised in some of those other scenarios too.   

Not sure.

alegacy: Sorry, gave  this some more thought today... I guess there wouldn't be much different in the code execution if there were no `BareMetalSwitch` CRs, and in the future no `HostNetworkAttachment` CRs or references to them in BMH.spec.networkInterfaces.  

...so maybe I'm overthinking this a bit and always having `IRONIC_NETWORKING_ENABLED=true` isn't problematic.

alegacy: @dtantsur I wonder if it wouldn't be a good idea to add a couple of fields to the `status` of the `Ironic` CR so that the current secret namespaces/names being used for the switch config & credentials secrets are advertised publicly.  ...because the names in the `spec` part are optional and the operator will create some using hardcoded names if a user hasn't provided them.  

I can add that as a follow-up both here and in IrSO rather than keep pilling on.

dtantsur: Yeah, I also don't see it as immediately problematic.

But that makes me think: why would anyone disable this feature? If, as you say, there is no difference in the code without the corresponding CRs?

dtantsur:
/retest

Network failure in the check-links-pr job

apis/metal3.io/v1alpha1/baremetalswitch_types.go
```diff
@@ -0,0 +1,152 @@
...
+	// +kubebuilder:validation:Enum=generic-switch
+	// +kubebuilder:default=generic-switch
+	// +optional
+	Driver string `json:"driver,omitempty"`
+
+	// DeviceType specifies the device type for the generic-switch driver.
+	// Must be one of the device types supported by networking-generic-switch.
+	// Examples: netmiko_cisco_ios, netmiko_dell_force10, netmiko_dell_os10,
+	//   netmiko_juniper_junos, netmiko_arista_eos
+	// See https://github.com/openstack/networking-generic-switch/blob/master/setup.cfg
```
dtantsur: Dunno, maybe better suggest looking for device_type in https://docs.openstack.org/networking-generic-switch/latest/configuration.html? setup.cfg is not very user-friendly

apis/metal3.io/v1alpha1/baremetalswitch_types.go
```diff
@@ -0,0 +1,152 @@
...
+	// +kubebuilder:default=password
+	Type SwitchCredentialType `json:"type"`
+
+	// The name of the secret containing the switch credentials.
+	// For password authentication, requires keys "username" and "password"
+	// For SSH key authentication, requires keys "username" and "ssh-privatekey".
+	// In both cases, an optional "enable-secret" key can be provided if needed
+	// to enable privileged mode.
+	// +kubebuilder:validation:MinLength=1
+	SecretName string `json:"secretName"`
```
dtantsur: Other places in BMO use a `SecretReference`, which uses `name` instead of `secretName`. Same for IrSO's ResourceReference. Maybe use just Name for consistency?

apis/metal3.io/v1alpha1/baremetalswitch_types.go
```diff
@@ -0,0 +1,152 @@
...
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=65535
+	// +optional
+	Port *int32 `json:"port,omitempty"`
+
+	// DisableCertificateVerification disables TLS certificate verification when using
+	// HTTPS to connect to the switch. This is required for self-signed certificates,
+	// but is insecure as it allows man-in-the-middle attacks.
+	// +optional
+	DisableCertificateVerification *bool `json:"disableCertificateVerification,omitempty"`
```
dtantsur: nit: probably don't need a pointer, bool already defaults to false nicely

config/overlays/e2e/ironic.env
dtantsur: The pattern we use for API credentials is to automatically set the field in spec if the user does not do it. Maybe do the same? You're using a label to distinguish managed credentials anyway.

internal/controller/metal3.io/baremetalswitch_config.go
```diff
@@ -0,0 +1,301 @@
...
+	return result, nil
+}
+
+// switchConfigData holds the template data for generating a switch's INI config entry.
+type switchConfigData struct {
+	Name         string
+	Address      string
+	MACAddress   string
+	Port         *int32
+	DriverType   string
```
dtantsur: nit: just `Driver` for consistency with the API

apis/metal3.io/v1alpha1/baremetalswitch_types.go
```diff
@@ -0,0 +1,152 @@
...
+	// Type is the type of switch credentials.
+	// This is currently limited to "password", but will be expanded to others.
+	// +kubebuilder:validation:Enum=password;publickey
+	// +kubebuilder:default=password
+	Type SwitchCredentialType `json:"type"`
+
+	// The name of the secret containing the switch credentials.
+	// For password authentication, requires keys "username" and "password"
+	// For SSH key authentication, requires keys "username" and "ssh-privatekey".
+	// In both cases, an optional "enable-secret" key can be provided if needed
```
dtantsur: The enable-secret part is pretty confusing (I personally cannot remember what it means). Does it have to be a part of the secret or can we move it to public API, e.g. here in SwitchCredentials?

internal/controller/metal3.io/baremetalswitch_config.go
dtantsur: Transient errors short-circuit generateSwitchConfig, so we're fine on this front.

I think I'm with Allain on "1 bad switch config shouldn't block all of the good ones". Since BareMetalSwitches are independent resources, failing here would create an unexpected and hard-to-debug dependency between them.

apis/metal3.io/v1alpha1/baremetalswitch_types.go
```diff
@@ -0,0 +1,152 @@
...
+	DisableCertificateVerification *bool `json:"disableCertificateVerification,omitempty"`
+}
+
+// SwitchConditionType defines the condition types for BareMetalSwitch.
+type SwitchConditionType string
+
+const (
+	// SwitchConditionReconciled indicates whether the switch has been
+	// successfully reconciled into the switch config secret.
+	SwitchConditionReconciled SwitchConditionType = "Reconciled"
```
dtantsur: I'd go with "Valid" which has prior art on several BMO resources already. "Reconciled" can make an impression that it's changed also for transient errors, which is not the case.

test/e2e/config/ironic.yaml
```diff
@@ -82,6 +82,7 @@ intervals:
   default/wait-secret-deletion: [ "1m", "1s" ]
   default/wait-connect-ssh: [ "2m", "10s" ]
   default/wait-externally-provisioned: [ "1m", "10ms" ]
+  baremetalswitch/wait-secret-updated: [ "30s", "500ms" ]
```
dtantsur, line 92: nit: The new test doesn't actually require Ironic, does it? Maybe skip it when running Ironic-based tests and only run in the fixture job? Just to save some resources.

apis/metal3.io/v1alpha1/baremetalswitch_types.go
alegacy: It's an additional password to be used on the switch to enter admin privilege level for configuration changes.   I do believe that it is best in a secret.

apis/metal3.io/v1alpha1/baremetalswitch_types.go
alegacy: ok, I can change that.  Figured `setup.cfg` would avoid cases of stale docs.

apis/metal3.io/v1alpha1/baremetalswitch_types.go
alegacy: That would make the API a bit hard to read if just looking at the yaml.  A user may not immediately recognize that `name` is meant to point to a Secret.

alegacy: happy to change it if that's what is preferred though!

test/e2e/config/ironic.yaml
alegacy: it doesn't today, but I was sort of hoping in the future when the IrSO changes are merged we could theoretically ensure that the switch config file was written properly via the mount and via the processing/translating that ironic-networking does to it.... and then even further into the future maybe we'd have an actual virtual switch to configure, but that's a bit over past the horizon.

I could do something like this in `hack/ci-e2e.sh` to skip them in non-fixture cases.
```
  -export GINKGO_SKIP_LABELS="${GINKGO_SKIP_LABELS:-}"                                                                                                                                                                                                                                                            
  +export GINKGO_SKIP_LABELS="${GINKGO_SKIP_LABELS:-baremetalswitch}" 
```

apis/metal3.io/v1alpha1/baremetalswitch_types.go
alegacy: changed

internal/controller/metal3.io/baremetalswitch_config.go
alegacy: changed

apis/metal3.io/v1alpha1/baremetalswitch_types.go
alegacy: changed

apis/metal3.io/v1alpha1/baremetalswitch_types.go
alegacy: changed

apis/metal3.io/v1alpha1/baremetalswitch_types.go
```diff
@@ -0,0 +1,151 @@
+/*
+Copyright 2025 The Metal3 Authors.
```
dtantsur: 2026 already :)

apis/metal3.io/v1alpha1/baremetalswitch_types.go
dtantsur: Can we call it something like `adminPassword` then? Or even have a separate secret iIt feels like an anti-pattern to have two passwords in one secret)?

apis/metal3.io/v1alpha1/baremetalswitch_types.go
dtantsur: The downside is that now users need to remember what field uses name and what - secretName. I'd rather change still.

config/overlays/e2e/namespaced-manager-patch.yaml
```diff
@@ -9,5 +9,5 @@ spec:
       containers:
       - env:
         - name: WATCH_NAMESPACE
-          value: basic-ops,external-inspection,externally-provisioned,firmware-settings,inspection,live-iso-ops,provisioning-ops,re-inspection,automated-cleaning,upgrade-bmo,upgrade-ironic
+          value: basic-ops,external-inspection,externally-provisioned,firmware-settings,inspection,live-iso-ops,provisioning-ops,re-inspection,automated-cleaning,upgrade-bmo,upgrade-ironic,baremetalswitch
```
dtantsur: Another merge conflict here (sorry!)

alegacy: no worries, I'll rebase

apis/metal3.io/v1alpha1/baremetalswitch_types.go
alegacy: :) time flies!

apis/metal3.io/v1alpha1/baremetalswitch_types.go
alegacy: Sure, I don't mind renaming and/or splitting it out if you prefer, but I just want to point out that anyone familiar with switches will immediately know what "enable password" means, "admin password" might be a little confusing to them but if I add a detailed description/comment to the field that should help clarify.

apis/metal3.io/v1alpha1/baremetalswitch_types.go
alegacy: ok, will rename

apis/metal3.io/v1alpha1/baremetalswitch_types.go
dtantsur: Okay, we can start with a detailed explanation and see what the others think

apis/metal3.io/v1alpha1/baremetalswitch_types.go
alegacy: I had a chat with my good friend Claude and we settled on switching this over to a `SecretReference` and renaming the expected key to `admin-password`.

apis/metal3.io/v1alpha1/baremetalswitch_types.go
alegacy: As per the other comment above ... I had a chat with my good friend Claude and we settled on switching this over to a SecretReference and renaming the expected key to admin-password.

apis/metal3.io/v1alpha1/baremetalswitch_types.go
alegacy: fixed

alegacy:
@dtantsur any further comments?

test/e2e/baremetalswitch_test.go
```diff
@@ -0,0 +1,613 @@
...
+			g.Expect(config).To(ContainSubstring("device_type=netmiko_cisco_ios"))
+			g.Expect(config).To(ContainSubstring("username=admin"))
+			g.Expect(config).To(ContainSubstring("password=secret123"))
+		}, e2eConfig.GetIntervals(specName, "wait-secret-updated")...).Should(Succeed())
+
+		By("verifying the Reconciled condition is True on the first switch")
+		Eventually(func(g Gomega) {
+			sw := &metal3api.BareMetalSwitch{}
+			g.Expect(c.Get(ctx, client.ObjectKeyFromObject(switch1), sw)).To(Succeed())
+			cond := meta.FindStatusCondition(sw.Status.Conditions, string(metal3api.SwitchConditionValid))
```
dtantsur: nit: I've recently added a ContainCondition matcher that you can use here.

test/e2e/baremetalswitch_test.go
```diff
@@ -0,0 +1,613 @@
...
+				}
+			}
+		}, e2eConfig.GetIntervals(specName, "wait-secret-updated")...).Should(Succeed())
+	})
+
+	AfterEach(func() {
+		DumpResources(ctx, e2eConfig, clusterProxy, path.Join(artifactFolder, specName))
+		if !skipCleanup {
+			isNamespaced := e2eConfig.GetBoolVariable("NAMESPACE_SCOPED")
+			Cleanup(ctx, clusterProxy, namespace, cancelWatches, isNamespaced, e2eConfig.GetIntervals("default", "wait-namespace-deleted")...)
```
dtantsur: When you rebase, could you check the new pattern of cleaning up resources? This function will have a different signature.

test/e2e/baremetalswitch_test.go
```diff
@@ -0,0 +1,613 @@
...
+		By("deleting switch-a and switch-c")
+		Expect(c.Delete(ctx, switches[0])).To(Succeed())
+		Expect(c.Delete(ctx, switches[2])).To(Succeed())
+
+		By("verifying config has only the header comment")
+		Eventually(func(g Gomega) {
+			config := getSwitchConfig(ctx, c, namespace.Name)
+			g.Expect(config).To(Equal(switchConfigHeader))
+		}, e2eConfig.GetIntervals(specName, "wait-secret-updated")...).Should(Succeed())
+
```
dtantsur, line 503: maybe that's the point where you could split the test into two smaller ones?

config/overlays/e2e/ironic.env
alegacy: agreed, leaving it on is probably best.

dtantsur:
/retest

Early timeout

internal/controller/metal3.io/baremetalswitch_config.go
```diff
@@ -0,0 +1,308 @@
...
+	secret, err := sm.ObtainSecret(ctx, types.NamespacedName{
+		Name:      secretName,
+		Namespace: namespace,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to get secret %s: %w", secretName, err)
+	}
+
+	// Only update if the data has changed
+	if secretDataEqual(secret.Data, newData) {
```
dtantsur, line 284: nit: `reflect.DeepEqual`?

test/e2e/baremetalswitch_test.go
```diff
@@ -0,0 +1,603 @@
...
+					g.Expect(section).To(ContainSubstring("insecure=true"))
+				}
+			}
+		}, e2eConfig.GetIntervals(specName, "wait-secret-updated")...).Should(Succeed())
+	})
+
+	AfterEach(func() {
+		DumpResources(ctx, e2eConfig, clusterProxy, path.Join(artifactFolder, specName))
+		if !skipCleanup {
+			Cleanup(ctx, clusterProxy, namespace, cancelWatches, e2eConfig, nil)
```
dtantsur, line 600: nit: The last argument is not `nil`, it's a list of objects you create (particularly, BareMetalSwitches)

dtantsur:
/lgtm

Thanks! Please check the nits I posted for a potential follow-up.

PR #3048, opened 2026-03-04
:bug: Fix race condition where DataImage node-busy blocks power-on

elfosardo:
During parallel deployments with high concurrency (1o+), hosts can geti stuck powered off after virtual media is attached. The root cause is in handleDataImageActions: when GetDataImageStatus returns ErrNodeIsBusy (Ironic node has a reservation), the function always returns a requeue, preventing manageHostPower from ever reaching PowerOn.

Under concurrent load, Ironic nodes frequently hold reservations from overlapping operations (virtual media attach, power state sync, manage). This causes repeated 30s+ requeue delays that effectively starve the power-on step, leaving part of the hosts unpowered.

When the node is busy but the DataImage is already in the desired state (Spec.URL matches Status.AttachedImage.URL), skip the status check and let power-on proceed. The attachment will be verified against Ironic on a future reconciliation when the node is no longer busy.

Assisted-By: Claude 4.6 Opus High (Commercial License)

### DIFF
--- a/internal/controller/metal3.io/baremetalhost_controller.go
+++ b/internal/controller/metal3.io/baremetalhost_controller.go
@@ -1733,9 +1733,25 @@ func (r *BareMetalHostReconciler) handleDataImageActions(ctx context.Context, pr
 	// a longer wait ?
 	isImageAttached, getVmediaError := prov.GetDataImageStatus(ctx)
 	if getVmediaError != nil {
-		info.log.Error(getVmediaError, "Error fetching Virtual Media details")
-
-		if !errors.Is(getVmediaError, provisioner.ErrNodeIsBusy) {
+		if errors.Is(getVmediaError, provisioner.ErrNodeIsBusy) {
+			// When the node is busy (e.g. reservation held by another
+			// operation), we cannot query virtual media status. If the
+			// DataImage is already in the desired state according to our
+			// status, don't block power-on -- let it proceed and verify
+			// attachment on a future reconciliation.
+			deleteDataImage := !dataImage.DeletionTimestamp.IsZero()
+			if !deleteDataImage && dataImage.Spec.URL == dataImage.Status.AttachedImage.URL {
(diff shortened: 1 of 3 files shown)


tuminoid:
Btw, it would be really helpful if you left a quick comment on the Copilot review notes if you changed something or they were false. Most of the time its not trivial to see from the following diff if author reacted to them or not.

elfosardo:
> Btw, it would be really helpful if you left a quick comment on the Copilot review notes if you changed something or they were false. Most of the time its not trivial to see from the following diff if author reacted to them or not.

@tuminoid of course, will do

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -1733,9 +1733,25 @@ func (r *BareMetalHostReconciler) handleDataImageActions(ctx context.Context, pr
 	// a longer wait ?
 	isImageAttached, getVmediaError := prov.GetDataImageStatus(ctx)
 	if getVmediaError != nil {
-		info.log.Error(getVmediaError, "Error fetching Virtual Media details")
-
-		if !errors.Is(getVmediaError, provisioner.ErrNodeIsBusy) {
+		if errors.Is(getVmediaError, provisioner.ErrNodeIsBusy) {
+			// When the node is busy (e.g. reservation held by another
+			// operation), we cannot query virtual media status. If the
```
dtantsur, line 1738: Shouldn't we fix it in Ironic instead? GET API should not require an exclusive lock..

elfosardo: you're right
checking the logs and tracking the GET method I found an inconsistency
proposed a fix in ironic https://review.opendev.org/c/openstack/ironic/+/980089

elfosardo:
I'm going to keep this on hold until we verify the patch in ironic fixes the issue

Rozzii (commented):
/remove-lifecycle stale
Any info whether the fix on Ironic side worked @elfosardo @dtantsur ?

PR #3053, opened 2026-03-06
:seedling: add zizmor scanner

tuminoid:
Zizmor scans workflow files to keep them hardened. It runs on push to produce a report in security tab, and it runs on PRs and blocks them if they is any errors.

    - Harden dependabot workflow: disable persist-credentials on checkout,
      pass token explicitly to add-and-commit action
    - Suppress artipacked on release workflow checkout that needs
      credentials for pushing branches and tags
    - Suppress superfluous-actions on softprops/action-gh-release pending
      replacement with gh CLI


tuminoid:
/hold
This is expected to fail, I will adjust.

PR #3068, opened 2026-03-09
🌱 e2e: check that all namespaces are added to WATCH_NAMESPACE

dtantsur:
It's a huge footgun that newly added tests get their hosts ignored by
default. Add a validation that will at least warn a developer.

### DIFF
--- a/hack/check-e2e.sh
+++ b/hack/check-e2e.sh
@@ -1,0 +1,25 @@
+#!/usr/bin/env bash
+
+set -eu -o pipefail
+
+REPO_ROOT=$(realpath "$(dirname "${BASH_SOURCE[0]}")/..")
+cd "${REPO_ROOT}" || exit 1
+
+WATCH_NS="$(grep -A1 WATCH_NAMESPACE config/overlays/e2e/namespaced-manager-patch.yaml | grep 'value:' | awk '{ print $2; }')"
+EXCEPTIONS=upgrade
+
+EXITCODE=0
+for spec in $(grep -E --no-filename ' *specName +:?=' test/e2e/*_test.go | grep -o '".*"' | tr -d '"'); do
+    # shellcheck disable=SC2076
+    if ! [[ ",${WATCH_NS},${EXCEPTIONS}," =~ ",${spec}," ]]; then
+        echo "ERROR: namespace ${spec} is declared in e2e test but not added to config/overlays/e2e/namespaced-manager-patch.yaml or EXCEPTIONS"
+        EXITCODE=1
+    fi
+
+    # shellcheck disable=SC2076
+    if ! [[ ",${EXCEPTIONS}," =~ ",${spec}," ]] && ! [[ -f "config/overlays/e2e/${spec}/namespace.yaml" ]]; then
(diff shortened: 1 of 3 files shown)


hack/check-e2e.sh
```diff
@@ -0,0 +1,19 @@
...
+REPO_ROOT=$(realpath "$(dirname "${BASH_SOURCE[0]}")/..")
+cd "${REPO_ROOT}" || exit 1
+
+WATCH_NS="$(grep 'value:' config/overlays/e2e/namespaced-manager-patch.yaml | awk '{ print $2; }')"
+EXCEPTIONS=upgrade
+
+EXITCODE=0
+for spec in $(grep -E --no-filename ' *specName +:?=' test/e2e/*_test.go | grep -o '".*"' | tr -d '""'); do
+    # shellcheck disable=SC2076
+    if ! [[ ",$WATCH_NS,$EXCEPTIONS," =~ ",$spec," ]]; then
```
tuminoid:
```suggestion
    if ! [[ ",${WATCH_NS},${EXCEPTIONS}," =~ ",${spec}," ]]; then
```

hack/check-e2e.sh
```diff
@@ -0,0 +1,19 @@
...
+        echo "ERROR: namespace $spec is declared in e2e test but not added to config/overlays/e2e/namespaced-manager-patch.yaml or EXCEPTIONS"
```
tuminoid:
```suggestion
        echo "ERROR: namespace ${spec} is declared in e2e test but not added to config/overlays/e2e/namespaced-manager-patch.yaml or EXCEPTIONS"
```

hack/check-e2e.sh
```diff
@@ -0,0 +1,19 @@
...
+        EXITCODE=1
+    fi
+done
+exit $EXITCODE
```
tuminoid:
```suggestion
exit ${EXITCODE}
```

hack/check-e2e.sh
tuminoid: Had the edit the suggestion, highlighting perfectly why the curlys are needed :)

hack/check-e2e.sh
```diff
@@ -0,0 +1,19 @@
...
+set -eu -o pipefail
+
+REPO_ROOT=$(realpath "$(dirname "${BASH_SOURCE[0]}")/..")
+cd "${REPO_ROOT}" || exit 1
+
+WATCH_NS="$(grep 'value:' config/overlays/e2e/namespaced-manager-patch.yaml | awk '{ print $2; }')"
+EXCEPTIONS=upgrade
+
+EXITCODE=0
+for spec in $(grep -E --no-filename ' *specName +:?=' test/e2e/*_test.go | grep -o '".*"' | tr -d '""'); do
```
tuminoid:
```suggestion
for spec in $(grep -E --no-filename ' *specName +:?=' test/e2e/*_test.go | grep -o '".*"' | tr -d '"'); do
```
I don't think double double quote does nothing. You're deleting all of them anyways.

hack/check-e2e.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
tuminoid: :+1:

Issue #3067, opened 2026-03-09
Excessive error logging in the conditions code


dtantsur:
During host registration or right after its deletion, the conditions handling code logs messages like this:

```
{"level":"error","ts":1773069845.8253586,"logger":"provisioner.ironic","msg":"Ignored error while checking for Power Failure","host":"live-iso-ops~live-iso-ops","error":"host not registered","stacktrace":"github.com/metal3-io/baremetal-operator/pkg/provisioner/ironic.(*ironicProvisioner).HasPowerFailure\n\t/workspace/pkg/provisioner/ironic/ironic.go:2039\ngithub.com/metal3-io/baremetal-operator/internal/controller/metal3%2eio.computeConditions\n\t/workspace/internal/controller/metal3.io/baremetalhost_controller.go:2280\ngithub.com/metal3-io/baremetal-operator/internal/controller/metal3%2eio.(*BareMetalHostReconciler).Reconcile\n\t/workspace/internal/controller/metal3.io/baremetalhost_controller.go:252\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Reconcile\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.22.5/pkg/internal/controller/controller.go:216\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.22.5/pkg/internal/controller/controller.go:461\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.22.5/pkg/internal/controller/controller.go:421\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func1.1\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.22.5/pkg/internal/controller/controller.go:296"}
```

They pollute the logs and will probably cause false bug reports in the future. Let's try to remove them. Maybe don't even bother with checking for a power failure for hosts in `registering` and `deleting` states, as well as detached hosts?

/cc @pierrecregut 
/kind bug

PR #3069, opened 2026-03-09
:bug: Do not check power failure when not applicable

pierrecregut:
During conditions computation, power failure were checked in all states which lead to errors in registering/deleting states and for detached hosts. This commit removes the check in these conditions.

Fixes: #3067

### DIFF
--- a/internal/controller/metal3.io/baremetalhost_controller.go
+++ b/internal/controller/metal3.io/baremetalhost_controller.go
@@ -2224,6 +2224,7 @@ func setConditionsProgressing(host *metal3api.BareMetalHost, progressingReason s
 }
 
 func computeConditions(ctx context.Context, host *metal3api.BareMetalHost, prov provisioner.Provisioner) {
+	var powerFailureCheck = true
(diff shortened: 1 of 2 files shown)


internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -2231,6 +2224,7 @@ func setConditionsProgressing(host *metal3api.BareMetalHost, progressingReason s
 }
 
 func computeConditions(ctx context.Context, host *metal3api.BareMetalHost, prov provisioner.Provisioner) {
+	var powerFailureCheck = true
 	switch host.Status.Provisioning.State {
 	case metal3api.StateNone, metal3api.StateUnmanaged:
```
dtantsur, line 2229: Let's also skip in these states: we're guaranteed to have no ironic node

pierrecregut: done.

dtantsur:
/lgtm

Thanks!

PR #3074, opened 2026-03-13
🐛 Add handling of paused annotation to the HFS controller

dtantsur:
Currently, it ignores this annotation, causing unnecessary reconciles.
Create a common function to avoid these situations in the future.

While looking into this, I've realized that HFC/HFS controllers are not notified
on BMH changes, so fixed that too.


dtantsur:
/hold

You know what, let's actually fix triggering on BMH changes

PR #3113, opened 2026-03-24
:bug: Remove unused RBAC permissions from controller ClusterRole

tuminoid:
Remove unused verbs from kubebuilder RBAC markers to enforce least privilege:
- secrets: remove delete (we're consuming secrets, not removing them)
- events: remove get/list/watch (not setup via runtime)
- namespaces: remove list/watch (we have explicit list, if we care about them)

### DIFF
--- a/config/render/capm3.yaml
+++ b/config/render/capm3.yaml
@@ -2908,25 +2908,19 @@ rules:
   - events
   verbs:
   - create
-  - get
-  - list
   - patch
   - update
-  - watch
 - apiGroups:
   - ""
   resources:
   - namespaces
   verbs:
   - get
-  - list
-  - watch
 - apiGroups:
   - ""
   resources:
   - secrets
   verbs:
-  - delete
   - get
   - list
   - update
--- a/internal/controller/metal3.io/baremetalhost_controller.go
+++ b/internal/controller/metal3.io/baremetalhost_controller.go
@@ -90,8 +90,8 @@ func (info *reconcileInfo) publishEvent(reason, message string) {
 // +kubebuilder:rbac:groups=metal3.io,resources=preprovisioningimages,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=metal3.io,resources=hardwaredata,verbs=get;list;watch;create;delete;patch;update
 // +kubebuilder:rbac:groups=metal3.io,resources=hardware/finalizers,verbs=update
-// +kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;update;delete
-// +kubebuilder:rbac:groups="",resources=events,verbs=get;list;watch;create;update;patch
+// +kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;update
+// +kubebuilder:rbac:groups="",resources=events,verbs=create;update;patch
 
 // Allow for managing hostfirmwaresettings, firmwareschema, bmceventsubscriptions and hostfirmwarecomponents
 // +kubebuilder:rbac:groups=metal3.io,resources=hostfirmwaresettings,verbs=get;list;watch;create;update;patch
--- a/config/base/rbac/role.yaml
+++ b/config/base/rbac/role.yaml
@@ -10,25 +10,19 @@ rules:
   - events
   verbs:
   - create
-  - get
-  - list
   - patch
   - update
-  - watch
 - apiGroups:
   - ""
   resources:
   - namespaces
   verbs:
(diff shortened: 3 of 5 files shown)


tuminoid:
/hold
Need to mark it "ready" to get GHA tests running. Not ready.

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -104,10 +104,10 @@ func (info *reconcileInfo) publishEvent(reason, message string) {
 // +kubebuilder:rbac:groups=metal3.io,resources=dataimages/status,verbs=get;update;patch
 
 // Allow for updating hostupdatepolicies
-// +kubebuilder:rbac:groups=metal3.io,resources=hostupdatepolicies,verbs=get;list;watch;update
+// +kubebuilder:rbac:groups=metal3.io,resources=hostupdatepolicies,verbs=get;list;watch
```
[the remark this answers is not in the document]
tuminoid: fixed

tuminoid:
We have serious lack of testing coverage. Some of the RBAC as mentioned by Copilot are absolutely needed, but nothing in bmo e2e or e2e integration or feature tests fail if they're removed.

tuminoid:
Restored some permissions as they're needed in non-default codepaths or by controller-runtime in the background for cache mechanisms.

tuminoid:
/test metal3-centos-e2e-feature-test-main-features
/test metal3-centos-e2e-integration-test-main
/test metal3-ubuntu-e2e-integration-test-main
/test metal3-ubuntu-e2e-feature-test-main-pivoting
/test metal3-ubuntu-e2e-feature-test-main-remediation

These do not really test anything, but I'll run them anyways.

tuminoid:
/cc @lentzi90 
Would you agree we probably don't use the verb left in this PR?

lentzi90:
Argh... the "lack of test coverage" is because we use namespaced mode for the normal tests. You will need to update the RBAC here manually: https://github.com/metal3-io/baremetal-operator/blob/main/config/overlays/e2e/roles-rolebindings/roles-rolebindings.yaml
Sigh. I wonder if we could find a better way to keep these in sync with what is generated for the cluster scope...

tuminoid:
Oh right. You mean BMO e2e or e2e integration/feature suites, or both?

Part of the issue is probably also half-baked IRSO testing?

lentzi90:
This should be only used by BMO e2e. All other tests use cluster scope as far as I know.
I think we test the IRSO integration enough to tell if the RBAC works but I am not completely sure

tuminoid:
Unrelated failures. Will run a new set if the namespaced RBAC I now fixed passes.

tuminoid:
> This should be only used by BMO e2e. All other tests use cluster scope as far as I know. I think we test the IRSO integration enough to tell if the RBAC works but I am not completely sure

There are optional codepaths with IRSO, depending on what environment variables are and are not set, and some of the RBAC changes affect non-default paths as we don't set some of the environment variables. I think IRONIC_NAME was one, which led to some listener not being activated, which would've failed with reduced RBAC.

tuminoid:
/cc @lentzi90 
Take another look when time,

lentzi90 (commented):
This makes sense to me but I cannot say I have deep knowledge of all the paths where the controller might use the RBAC.
/lgtm

adilGhaffarDev (commented):
LGTM

kashifest (commented):
/approve

config/render/capm3.yaml
```diff
@@ -2908,25 +2908,19 @@ rules:
...
   verbs:
   - get
-  - list
-  - watch
 - apiGroups:
   - ""
   resources:
   - secrets
   verbs:
-  - delete
```
rabi, line 2929: This change removes delete permission on secrets, but the BMO BareMetalHost controller still updates the BMC credentials Secret to add an [ownerReference](https://github.com/metal3-io/baremetal-operator/blame/main/pkg/secretutils/secret_manager.go#L95). Kubernetes rejects that update without delete on the Secret, and we’re seeing:

_an unhandled failure occurred with the BMC secret: failed to update secret edpm-compute-baremetal-00-bmc-secret in namespace openstack: secrets "edpm-compute-baremetal-00-bmc-secret" is forbidden: cannot set an ownerRef on a resource you can't delete_

So this looks like a regression.

tuminoid: Thanks for spotting it. Opened #3304 for it. Can you there provide more information on your environment?

PR #3120, opened 2026-03-26
🌱 Remove legacy vbmctl entrypoint and update E2E VM config

### LINKED ISSUE #3060: Migrate E2E tests to use new vbmctl and clean up
With https://github.com/metal3-io/baremetal-operator/pull/2953 merged, we can now switch to the new CLI and remove all traces of the legacy CLI.

First, ci-e2e.sh must be updated to use the new CLI: https://github.com/metal3-io/baremetal-operator/blob/c70ba879ced150ef5d01978945169b493488508b/hack/ci-e2e.sh#L111-L114
This means also updating the BMC config files to the new format.
It is likely that we will uncover cases where the old config is entangled with the e2e tests. The goal is to clean these up. There will be some duplicated information between e2e and vbmctl, but they should not be tied together. Separate configuration files are better to avoid accidentally making things worse.
My suggestion would be to make e2e use **only** the e2e config. Add any needed data in there and keep the vbmctl config separate, with only the data it needs.

When ci-e2e.sh is properly migrated, remove the [legacy CLI](https://github.com/metal3-io/baremetal-operator/blob/main/test/vbmctl/main.go) and the corresponding make target: https://github.com/metal3-io/baremetal-operator/blob/c70ba879ced150ef5d01978945169b493488508b/Makefile#L223-L225

Part of https://github.com/metal3-io/baremetal-operator/issues/2751

lentzi90:

- Delete test/vbmctl/main.go and build-legacy-vbmctl target
- Switch E2E to use new vbmctl CLI with YAML config
- Add test/e2e/config/vbmctl.yaml for VM definitions

Fixes #3060

**Checklist:**

/hold


lentzi90:
Great, tests are passing! I'll also remove the Network struct from the BMC config now. It was only used by the legacy vbmctl and is no longer needed.

PR #3124, opened 2026-03-27
:bug: Fix HFC controller error loop on deleted resource

honza:
When a HostFirmwareComponents resource is deleted while its BMH still exists, the HFC controller was returning a NotFound error, causing controller-runtime to requeue indefinitely with exponential backoff.

Return nil instead of the error since there is nothing to reconcile.


tuminoid:
Hmm mobile copilot review request removes blunderbuss reviews. Interesting... 🧐 

/cc tuminoid sunnatillo lentzi90

tuminoid:
> Hmm mobile copilot review request removes blunderbuss reviews. Interesting... 🧐
> 
> c

Prow was down for maintenacne. Adding back:
/cc tuminoid sunnatillo zhouhao3

Also need some approvers:
/cc @elfosardo @dtantsur

dtantsur:
/approve

The HFS implementation is also odd (it requeues for some reason), may be worth looking into as well.

tuminoid:
/cherry-pick release-0.12
/cherry-pick release-0.11

Lets see if they apply.

PR #3144, opened 2026-04-08
🐛 Fix HostFirmwareSettings blocking servicing abort due to stale ObservedGeneration

### LINKED ISSUE #3143: HostFirmwareSettings controller fails to update ObservedGeneration, blocking servicing abort

When using the servicing-abort feature (#3000) with HostFirmwareSettings (HFS), clearing the HFS spec to abort a failed servicing operation leaves the host stuck in ServicingError. The BMH controller repeatedly fails with:

`hostFirmwareSettings not ready yet: generation 9 != observed generation 7`

The HFS controller's setCondition function only compares the condition Status field (True/False/Unknown) when deciding whether to persist a status update. When only ObservedGeneration changes but the condition status stays the same, the update is skipped, leaving ObservedGeneration permanently stale and blocking the BMH controller's generation-based consistency checks.

A secondary issue compounds this: even after fixing the HFS controller, the BMH controller's servicing error recovery path calls generation-sensitive functions (getHostFirmwareSettings) before the HFS sub-controller has reconciled the spec change, causing a transient generation mismatch that blocks the abort with exponential backoff.

Clearing HFS spec should promptly abort the failed servicing operation and return the host to a provisioned/operational state, consistent with how HostFirmwareComponents (HFC) already works. The HFC controller's equivalent setUpdatesCondition correctly uses the return value of meta.SetStatusCondition, so it does not have this bug.

Proposed fix incoming

**Environment:**

- openshift dev-scripts, BMO close to latest (includes PR3000)
- Environment: VM lab with sushy-tools BMC emulator

/kind bug

jacob-anders:

The HostFirmwareSettings (HFS) controller has a bug in its setCondition function that prevents ObservedGeneration from being updated when only the generation changes but the condition status (True/False/Unknown) stays the same. This causes the BMH controller's generation-based consistency check in getHostFirmwareSettings to fail permanently with:

`hostFirmwareSettings not ready yet: generation 9 != observed generation 7`

This blocks the servicing-abort recovery path (#3000): when a user clears the HFS spec to abort a failed servicing operation, the host gets stuck in ServicingError because the BMH controller cannot get past the stale generation check.

Root cause: The HFS setCondition compared only the condition Status field against the old info.hfs.Status.Conditions to decide whether a status update was needed. When ObservedGeneration changed but Status did not, the update was silently skipped. The equivalent function in HostFirmwareComponents (setUpdatesCondition) does not have this bug -- it correctly returns the result of meta.SetStatusCondition, which compares all fields including ObservedGeneration.

Fix (two parts):

HFS controller: Replace the manual dirty-check logic in setCondition with the return value of meta.SetStatusCondition, matching the HFC controller's correct pattern. Remove the now-unused info parameter.

BMH controller: Add a fast-path in the servicing error recovery flow that checks HFS/HFC specs directly (via lightweight r.Get() calls that don't validate generation) before calling the generation-sensitive getHostFirmwareSettings/getHostFirmwareComponents. If specs are already cleared, abort immediately without waiting for the sub-controllers to reconcile. Also add watches on HFS/HFC generation changes so the BMH controller reconciles promptly on spec changes rather than waiting for error-state exponential backoff.

Fixes #3143

**Checklist:**

### DIFF
--- a/test/e2e/firmware_settings_test.go
+++ b/test/e2e/firmware_settings_test.go
@@ -573,6 +573,157 @@ var _ = Describe("Host Firmware Settings", Label("required", "firmware"), func()
 		Expect(hfs.Status.Conditions).To(ContainCondition(string(metal3api.FirmwareSettingsChangeDetected), metav1.ConditionFalse))
 	})
 
+	It("should abort servicing when HostFirmwareSettings spec is cleared during servicing", func() {
+		bmhName := specName + "-servicing-abort"
+		secretName := bmhName + "-bmc"
+
+		By("Creating a secret with BMH credentials")
+		bmcCredentialsData := map[string]string{
+			"username": bmc.User,
+			"password": bmc.Password,
+		}
+		secret := CreateSecret(ctx, clusterProxy.GetClient(), namespace.Name, secretName, bmcCredentialsData)
+		toCleanup = append(toCleanup, secret)
+
+		By("Creating a BMH with inspection and cleaning disabled")
+		bmh := metal3api.BareMetalHost{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      bmhName,
+				Namespace: namespace.Name,
+			},
+			Spec: metal3api.BareMetalHostSpec{
+				Online: true,
+				BMC: metal3api.BMCDetails{
+					Address:                        bmc.Address,
+					CredentialsName:                secretName,
+					DisableCertificateVerification: bmc.DisableCertificateVerification,
+				},
+				BootMode:              metal3api.BootMode(e2eConfig.GetVariable("BOOT_MODE")),
+				BootMACAddress:        bmc.BootMacAddress,
+				AutomatedCleaningMode: metal3api.CleaningModeDisabled,
+				InspectionMode:        metal3api.InspectionModeDisabled,
+			},
+		}
+		Expect(clusterProxy.GetClient().Create(ctx, &bmh)).To(Succeed())
+		toCleanup = append(toCleanup, &bmh)
+
+		By("Waiting for the BMH to become available")
+		WaitForBmhInProvisioningState(ctx, WaitForBmhInProvisioningStateInput{
+			Client: clusterProxy.GetClient(),
+			Bmh:    bmh,
+			State:  metal3api.StateAvailable,
+		}, e2eConfig.GetIntervals(specName, "wait-available")...)
+
+		By("Provisioning the BMH")
+		Expect(PatchBMHForProvisioning(ctx, PatchBMHForProvisioningInput{
+			client:    clusterProxy.GetClient(),
+			bmh:       &bmh,
+			bmc:       bmc,
+			e2eConfig: e2eConfig,
+			namespace: namespace.Name,
+		})).To(Succeed())
+
+		By("Waiting for the BMH to be provisioned")
+		WaitForBmhInProvisioningState(ctx, WaitForBmhInProvisioningStateInput{
+			Client: clusterProxy.GetClient(),
+			Bmh:    bmh,
+			State:  metal3api.StateProvisioned,
+		}, e2eConfig.GetIntervals(specName, "wait-provisioned")...)
+
+		By("Creating a HostUpdatePolicy to allow firmware changes on reboot")
+		hup := &metal3api.HostUpdatePolicy{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      bmhName,
+				Namespace: namespace.Name,
+			},
+			Spec: metal3api.HostUpdatePolicySpec{
+				FirmwareSettings: metal3api.HostUpdatePolicyOnReboot,
+			},
+		}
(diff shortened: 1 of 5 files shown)


dtantsur:
Would you mind if we hold this patch until https://github.com/metal3-io/baremetal-operator/pull/3246 is merged, so that you could add a functional test for the whole service-abort logic?

jacob-anders:
> Would you mind if we hold this patch until #3246 is merged, so that you could add a functional test for the whole service-abort logic?

of course, no problem

dtantsur:
#3246 is in. Could you rebase this PR and add a functional test for aborting servicing via HFS?

jacob-anders:
need to retest this in the lab post-rebase and post-copilot review
/hold

jacob-anders:
lab tests consistently successful
/unhold

test/e2e/firmware_settings_test.go
```diff
@@ -433,11 +423,149 @@ var _ = Describe("Host Firmware Settings", Label("required", "firmware"), func()
...
+		}, e2eConfig.GetIntervals(specName, "wait-firmware-settings")...)
+
+		By("Verifying the host is not stuck in ServicingError")
+		Expect(clusterProxy.GetClient().Get(ctx, types.NamespacedName{
+			Name:      bmhName,
+			Namespace: namespace.Name,
+		}, &bmh)).To(Succeed())
+		Expect(bmh.Status.ErrorType).To(BeEmpty())
+
+		By("Deleting the BMH")
```
dtantsur: If you rebase your chain on top of main, the BMH deletion part will be handled differently (see toDelete variable)

test/e2e/firmware_settings_test.go
```diff
@@ -433,11 +423,149 @@ var _ = Describe("Host Firmware Settings", Label("required", "firmware"), func()
...
+		Expect(helper.Patch(ctx, hfs)).To(Succeed())
+
+		By("Triggering a reboot via annotation")
+		Expect(clusterProxy.GetClient().Get(ctx, types.NamespacedName{
+			Name:      bmhName,
+			Namespace: namespace.Name,
+		}, &bmh)).To(Succeed())
+		AnnotateBmh(ctx, clusterProxy.GetClient(), bmh, metal3api.RebootAnnotationPrefix, ptr.To(""))
+
+		By("Waiting for the BMH to enter servicing")
```
dtantsur, line 677: I'm somewhat worried that servicing on sushy-tools happens too fast and that your test may end up testing nothing.

What I'm thinking is, maybe we add fault injection to sushy-tools? E.g. a fake value that never succeeds to be updated?

test/e2e/firmware_settings_test.go
```diff
@@ -433,11 +423,149 @@ var _ = Describe("Host Firmware Settings", Label("required", "firmware"), func()
...
+		hfs := &metal3api.HostFirmwareSettings{}
+		Expect(clusterProxy.GetClient().Get(ctx, types.NamespacedName{
+			Name:      bmhName,
+			Namespace: namespace.Name,
+		}, hfs)).To(Succeed())
+
+		helper, err := patch.NewHelper(hfs, clusterProxy.GetClient())
+		Expect(err).NotTo(HaveOccurred())
+		hfs.Spec.Settings = metal3api.DesiredSettingsMap{
+			hfsTestKey2: intstr.FromString(hfsTestNewValue2),
```
dtantsur, line 655: You need to rebase on top of freshly merged https://github.com/metal3-io/baremetal-operator/pull/3300, otherwise reusing keys may cause a conflict between tests.

dtantsur:
Doesn't seem to work:

>  BMH is in an unexpected state: servicing

test/e2e/firmware_settings_test.go
```diff
@@ -574,6 +574,147 @@ var _ = Describe("Host Firmware Settings", Label("required", "firmware"), func()
...
+
+		By("Waiting for servicing to abort and the BMH to return to OK")
+		WaitForBmhInOperationalStatus(ctx, WaitForBmhInOperationalStatusInput{
+			Client: clusterProxy.GetClient(),
+			Bmh:    bmh,
+			State:  metal3api.OperationalStatusOK,
+			UndesiredStates: []metal3api.OperationalStatus{
+				metal3api.OperationalStatusError,
+				metal3api.OperationalStatusServicing,
+			},
```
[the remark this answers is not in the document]
adilGhaffarDev: this is valid, test is also failing here.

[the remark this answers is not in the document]
jacob-anders: Agreed and fixed, thank you.

test/e2e/firmware_settings_test.go
jacob-anders: Good idea. I will add this in a follow up PR, I have a draft commit ready, now looking at sushy-tools side.

Rozzii:
I don't think this will land in 0.14. removing the target from related issue.

PR #3149, opened 2026-04-09
:bug: Skip cleaning when host is being deleted

honza:
When a BMH is deprovisioning in a terminating namespace, the controller cannot create the `PreprovisioningImage` required for automated cleaning. This causes the BMH to get stuck in deprovisioning indefinitely.

We fix this by skipping automated cleaning when the host has a deletion timestamp, since we know the PPI cannot be created. This is applied in three places: `registerHost` (PPI format selection and Register call) and `actionDeprovisioning` (Deprovision call). A check in `getPreprovImage` also handles the Forbidden error gracefully.

### DIFF
--- a/internal/controller/metal3.io/baremetalhost_controller.go
+++ b/internal/controller/metal3.io/baremetalhost_controller.go
@@ -765,6 +765,10 @@ func (r *BareMetalHostReconciler) getPreprovImage(ctx context.Context, info *rec
 		}
 
 		err = r.Create(ctx, &preprovImage)
+		if k8serrors.IsForbidden(err) && info.host.Status.Provisioning.State == metal3api.StateDeprovisioning {
+			info.log.Info("skipping PreprovisioningImage creation during deprovisioning, namespace may be terminating")
+			return nil, nil //nolint:nilnil
+		}
 		return nil, err
 	}
 	if err != nil {
@@ -844,7 +848,9 @@ func (r *BareMetalHostReconciler) registerHost(ctx context.Context, prov provisi
 		}
 	case metal3api.StateDeprovisioning:
 		// PreprovisioningImage is not required for deprovisioning when cleaning is disabled
-		if info.host.Spec.AutomatedCleaningMode == metal3api.CleaningModeDisabled {
+		// or when the host is being deleted (e.g. namespace is terminating)
+		if info.host.Spec.AutomatedCleaningMode == metal3api.CleaningModeDisabled ||
+			!info.host.DeletionTimestamp.IsZero() {
 			preprovImgFormats = nil
 		}
 	default:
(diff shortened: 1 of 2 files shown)


lentzi90 (commented):
I am not sure this is a good idea.
It could be seen as a way to tamper with the cleaning process, or as a workaround for a "bad workflow" that, as a side effect, skips cleaning.

I do acknowledge the issue though! We hit it in the e2e tests and are now making sure we delete BMHs separately before the namespace is removed.
Does anyone have any other ideas for how to handle this in a nicer way?

tuminoid:
> I am not sure this is a good idea. It could be seen as a way to tamper with the cleaning process, or as a workaround for a "bad workflow" that, as a side effect, skips cleaning.
> 
> I do acknowledge the issue though! We hit it in the e2e tests and are now making sure we delete BMHs separately before the namespace is removed. Does anyone have any other ideas for how to handle this in a nicer way?

I don't see tampering here as an issue, you can disable the automatedcleaningmode with the same permissions you can delete a BMH. 

I do see issues in patch though. This lacks checks if the namespace is terminating so the same skipping happens on just regular BMH delete, and also errors get swallowed on any deprovisioning, not only host deletion.

Rozzii:
IMO this is not in 100% an upstream Metal3 issue, the scenario described in the PR is relevant when an "external" PPI  provider is in use. In cases when an external PPI provider is in use in such a deprovisioning scenario that is described here the PPI generation is  indeed problematic. The default PPI provider can be turned off with an environment variable so I would suggest making the proposed logic conditional based on 2 variables
1. the https://github.com/metal3-io/baremetal-operator/blob/main/main.go#L142 flag 
2. whether there is an existing PPI ready to be used in the namespace that is being terminated I guess if the BMH was in use before and there was PPI generated in the past that could be re-used for cleaning or no?

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -844,7 +848,9 @@ func (r *BareMetalHostReconciler) registerHost(ctx context.Context, prov provisi
 		}
 	case metal3api.StateDeprovisioning:
 		// PreprovisioningImage is not required for deprovisioning when cleaning is disabled
-		if info.host.Spec.AutomatedCleaningMode == metal3api.CleaningModeDisabled {
+		// or when the host is being deleted (e.g. namespace is terminating)
+		if info.host.Spec.AutomatedCleaningMode == metal3api.CleaningModeDisabled ||
```
dtantsur, line 852: We cannot do this. Setting AutomatedCleaningMode to anything but CleaningModeDisabled *requires* us to run cleaning before deletion.

We could introduce a new flag along the lines of AutomatedCleaningPhase that can be used to decide of cleaning is done on any deprovisioning or only between provisionings (I think we've had a similar request in the past), but this change violates the existing API semantics of cleaning on deletion.

honza:
Superseded by #3214

PR #3152, opened 2026-04-10
:sparkles: Update logic for HostClaims

pierrecregut:

Update ensures the synchronization between HostClaim and BareMetalHost.
* Copy HostClaim secrets to BareMetalHost namespace
* Ensure propagation of reboot annotations
* Synchronize hostclaim status (mirrors BareMetalHost conditions)

**Checklist:**

### DIFF
--- a/apis/metal3.io/v1alpha1/baremetalhost_types.go
+++ b/apis/metal3.io/v1alpha1/baremetalhost_types.go
@@ -713,6 +713,12 @@ type DetachedAnnotationArguments struct {
 	// Force indicates if detaching should be forced regardless of the host's state
 	// +optional
 	Force bool `json:"force,omitempty"`
+
+	// SetByHostClaim is an informal flag that tells that the annotation was set by the
+	// HostClaim controller. HostClaim controller will never delete an annotation without
+	// SetByHostClaim set to true
+	// +optional
+	SetByHostClaim bool `json:"setByHostClaim,omitempty"`
 }
 
 // Match compares the saved status information with the name and
--- a/apis/metal3.io/v1alpha1/hostdeploypolicy_types.go
+++ b/apis/metal3.io/v1alpha1/hostdeploypolicy_types.go
@@ -25,6 +25,9 @@ type HostDeployPolicySpec struct {
 	// HostClaimNamespaces constrains the namespaces of the HostClaims allowed
 	// to bind the BareMetalHosts in the same namespace as the HostDeployPolicy
 	HostClaimNamespaces *HostClaimNamespaces `json:"hostClaimNamespaces,omitempty"`
+	// AllowsDetaching is a boolean flag specifying if the hostClaim can set the
+	// detached annotation of the BareMetalHost.
+	AllowsDetaching bool `json:"allowsDetaching,omitempty"`
 }
 
 type HostClaimNamespaces struct {
--- a/internal/controller/metal3.io/baremetalhost_controller.go
+++ b/internal/controller/metal3.io/baremetalhost_controller.go
@@ -30,6 +30,7 @@ import (
 	metal3api "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
 	"github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1/profile"
 	"github.com/metal3-io/baremetal-operator/pkg/hardwareutils/bmc"
+	"github.com/metal3-io/baremetal-operator/pkg/hostclaim"
 	"github.com/metal3-io/baremetal-operator/pkg/imageprovider"
 	"github.com/metal3-io/baremetal-operator/pkg/provisioner"
 	"github.com/metal3-io/baremetal-operator/pkg/secretutils"
@@ -39,8 +40,10 @@ import (
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/util/workqueue"
 	"sigs.k8s.io/cluster-api/util/conditions"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
@@ -48,7 +51,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
 const (
@@ -76,6 +81,7 @@ type BareMetalHostReconciler struct {
 type reconcileInfo struct {
 	log                              logr.Logger
 	host                             *metal3api.BareMetalHost
+	hostclaim                        *metal3api.HostClaim
 	hardwareData                     *metal3api.HardwareData
 	request                          ctrl.Request
 	bmcCredsSecret                   *corev1.Secret
@@ -227,9 +233,14 @@ func (r *BareMetalHostReconciler) Reconcile(ctx context.Context, request ctrl.Re
 	}
 
 	initialState := host.Status.Provisioning.State
+	hostclaim, err := r.getHostClaimIfExists(ctx, host)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("temporary failure while accessing hostclaim: %w", err)
+	}
 	info := &reconcileInfo{
 		log:                              reqLogger.WithValues("provisioningState", initialState),
 		host:                             host,
+		hostclaim:                        hostclaim,
 		hardwareData:                     hardwareData,
 		request:                          request,
 		bmcCredsSecret:                   bmcCredsSecret,
@@ -889,6 +900,7 @@ func (r *BareMetalHostReconciler) registerHost(ctx context.Context, prov provisi
 		host:          info.host,
 		log:           info.log.WithName("host_config_data"),
 		secretManager: r.secretManager(ctx, info.log),
+		hostClaim:     info.hostclaim,
 	}
 	preprovisioningNetworkData, err := hostConf.PreprovisioningNetworkData(ctx)
 	if err != nil {
@@ -1331,6 +1343,7 @@ func (r *BareMetalHostReconciler) actionProvisioning(ctx context.Context, prov p
 		host:          info.host,
 		log:           info.log.WithName("host_config_data"),
 		secretManager: r.secretManager(ctx, info.log),
+		hostClaim:     info.hostclaim,
 	}
 	info.log.Info("provisioning")
 
@@ -2565,6 +2578,43 @@ func (r *BareMetalHostReconciler) updateEventHandler(e event.UpdateEvent) bool {
 	return true
 }
 
+func queueBareMetalHostFromHostClaim(obj client.Object, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	hostClaim, ok := obj.(*metal3api.HostClaim)
+	if !ok {
+		return
+	}
+	bmh := hostClaim.Status.BareMetalHost
+	if bmh == nil {
+		return
+	}
+	q.Add(reconcile.Request{NamespacedName: types.NamespacedName{Namespace: bmh.Namespace, Name: bmh.Name}})
+}
+
+var hostClaimEventHandler = handler.TypedFuncs[client.Object, reconcile.Request]{
+	CreateFunc: func(_ context.Context, e event.TypedCreateEvent[client.Object], q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+		queueBareMetalHostFromHostClaim(e.Object, q)
+	},
+	UpdateFunc: func(_ context.Context, e event.TypedUpdateEvent[client.Object], q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+		objectNew, ok1 := e.ObjectNew.(*metal3api.HostClaim)
+		objectOld, ok2 := e.ObjectOld.(*metal3api.HostClaim)
+		if !ok1 || !ok2 {
+			return
+		}
+		if reflect.DeepEqual(objectNew.Spec.UserData, objectOld.Spec.UserData) &&
+			reflect.DeepEqual(objectNew.Spec.NetworkData, objectOld.Spec.NetworkData) &&
+			reflect.DeepEqual(objectNew.Spec.MetaData, objectOld.Spec.MetaData) {
+			return
+		}
+		queueBareMetalHostFromHostClaim(e.ObjectNew, q)
+	},
+	DeleteFunc: func(_ context.Context, e event.TypedDeleteEvent[client.Object], q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+		queueBareMetalHostFromHostClaim(e.Object, q)
+	},
+	GenericFunc: func(_ context.Context, e event.TypedGenericEvent[client.Object], q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+		queueBareMetalHostFromHostClaim(e.Object, q)
+	},
+}
+
 // SetupWithManager registers the reconciler to be run by the manager.
 func (r *BareMetalHostReconciler) SetupWithManager(mgr ctrl.Manager, preprovImgEnable bool, maxConcurrentReconcile int) error {
 	r.Recorder = mgr.GetEventRecorderFor("baremetalhost-controller")
@@ -2576,6 +2626,7 @@ func (r *BareMetalHostReconciler) SetupWithManager(mgr ctrl.Manager, preprovImgE
 				UpdateFunc: r.updateEventHandler,
 			}).
 		WithOptions(controller.Options{MaxConcurrentReconciles: maxConcurrentReconcile}).
+		Watches(&metal3api.HostClaim{}, hostClaimEventHandler).
 		Owns(&corev1.Secret{}, builder.MatchEveryOwner)
 
 	if preprovImgEnable {
@@ -2664,4 +2715,35 @@ func (r *BareMetalHostReconciler) reconcileHostData(ctx context.Context, host *m
 		return ctrl.Result{Requeue: true}, hardwareData, nil
 	}
 	return ctrl.Result{}, hardwareData, nil
+}
+
+// getHostClaimIfExists gives back a pointer to a HostClaim if the BareMetalHost is associated to it and HostClaim has correct backpointer in status.
+// Any error encountered during the check of the consumerRef is ignored.
+func (r *BareMetalHostReconciler) getHostClaimIfExists(ctx context.Context, host *metal3api.BareMetalHost) (hc *metal3api.HostClaim, err error) {
+	consumer := host.Spec.ConsumerRef
+	if consumer == nil {
+		return
+	}
+	gv, err := schema.ParseGroupVersion(consumer.APIVersion)
+	if err != nil || gv.Group != metal3api.GroupVersion.Group {
+		return
+	}
+	if consumer.Kind != hostclaim.HostClaimKind {
+		return
+	}
+	hostclaim := &metal3api.HostClaim{}
+	key := client.ObjectKey{Namespace: consumer.Namespace, Name: consumer.Name}
+	if err = r.Get(ctx, key, hostclaim); err != nil {
+		return
+	}
+	if hostclaim.Status.BareMetalHost == nil {
+		err = errors.New("no back link on hostclaim - no hostclaim for initialization data")
+		return
+	}
+	if hostclaim.Status.BareMetalHost.Name != host.Name || hostclaim.Status.BareMetalHost.Namespace != host.Namespace {
+		err = errors.New("bad back link to baremetalhost - no hostclaim for initialization data")
+		return
+	}
+	hc = hostclaim
+	return
 }
--- a/internal/controller/metal3.io/host_config_data.go
+++ b/internal/controller/metal3.io/host_config_data.go
@@ -3,12 +3,10 @@ package controllers
 import (
 	"context"
 	"errors"
-	"fmt"
 
 	"github.com/go-logr/logr"
 	metal3api "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
 	"github.com/metal3-io/baremetal-operator/pkg/secretutils"
-	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 )
@@ -33,16 +31,13 @@ type hostConfigData struct {
 	host          *metal3api.BareMetalHost
 	log           logr.Logger
 	secretManager secretutils.SecretManager
+	hostClaim     *metal3api.HostClaim
 }
 
 // Generic method for data extraction from a Secret. Function uses dataKey
 // parameter to detirmine which data to return in case secret contins multiple
 // keys.
 func (hcd *hostConfigData) getSecretData(ctx context.Context, name, namespace, dataKey string, addFinalizer bool) (string, error) {
-	if namespace != hcd.host.Namespace {
-		return "", fmt.Errorf("%s secret must be in same namespace as host %s/%s", dataKey, hcd.host.Namespace, hcd.host.Name)
-	}
-
 	key := types.NamespacedName{
 		Name:      name,
 		Namespace: namespace,
@@ -69,17 +64,27 @@ func (hcd *hostConfigData) getSecretData(ctx context.Context, name, namespace, d
 
 // UserData get Operating System configuration data.
 func (hcd *hostConfigData) UserData(ctx context.Context) (string, error) {
-	if hcd.host.Spec.UserData == nil {
-		hcd.log.Info("UserData is not set return empty string")
-		return "", nil
-	}
-	namespace := hcd.host.Spec.UserData.Namespace
-	if namespace == "" {
+	var name, namespace string
+	if hcd.hostClaim != nil {
+		userData := hcd.hostClaim.Spec.UserData
+		if userData == nil {
+			hcd.log.Info("UserData is not set return empty string")
+			return "", nil
+		}
+		name = userData.Name
+		namespace = hcd.hostClaim.Namespace
+	} else {
+		userData := hcd.host.Spec.UserData
+		if userData == nil {
+			hcd.log.Info("UserData is not set return empty string")
+			return "", nil
+		}
+		name = userData.Name
 		namespace = hcd.host.Namespace
 	}
 	return hcd.getSecretData(
 		ctx,
-		hcd.host.Spec.UserData.Name,
+		name,
 		namespace,
 		"userData",
 		false,
@@ -88,23 +93,29 @@ func (hcd *hostConfigData) UserData(ctx context.Context) (string, error) {
 
 // NetworkData get network configuration.
 func (hcd *hostConfigData) NetworkData(ctx context.Context) (string, error) {
-	networkData := hcd.host.Spec.NetworkData
-	if networkData == nil && hcd.host.Spec.PreprovisioningNetworkDataName != "" {
-		networkData = &corev1.SecretReference{
-			Name: hcd.host.Spec.PreprovisioningNetworkDataName,
+	var name, namespace string
+	if hcd.hostClaim != nil {
+		networkData := hcd.hostClaim.Spec.NetworkData
+		if networkData == nil {
+			return "", nil
 		}
-	}
-	if networkData == nil {
-		hcd.log.Info("NetworkData is not set, returning empty data")
-		return "", nil
-	}
-	namespace := networkData.Namespace
-	if namespace == "" {
+		namespace = hcd.hostClaim.Namespace
+		name = networkData.Name
+	} else {
+		networkData := hcd.host.Spec.NetworkData
 		namespace = hcd.host.Namespace
+		if networkData == nil {
+			if hcd.host.Spec.PreprovisioningNetworkDataName == "" {
+				return "", nil
+			}
+			name = hcd.host.Spec.PreprovisioningNetworkDataName
+		} else {
+			name = networkData.Name
+		}
 	}
 	networkDataRaw, err := hcd.getSecretData(
 		ctx,
-		networkData.Name,
+		name,
 		namespace,
 		"networkData",
 		false,
@@ -148,17 +159,25 @@ func (hcd *hostConfigData) PreprovisioningNetworkData(ctx context.Context) (stri
 
 // MetaData get host metatdata.
 func (hcd *hostConfigData) MetaData(ctx context.Context) (string, error) {
-	if hcd.host.Spec.MetaData == nil {
-		hcd.log.Info("MetaData is not set returning empty(nil) data")
-		return "", nil
-	}
-	namespace := hcd.host.Spec.MetaData.Namespace
-	if namespace == "" {
+	var name, namespace string
+	if hcd.hostClaim != nil {
+		if hcd.hostClaim.Spec.MetaData == nil {
+			hcd.log.Info("MetaData is not set returning empty(nil) data")
+			return "", nil
+		}
+		name = hcd.hostClaim.Spec.MetaData.Name
+		namespace = hcd.hostClaim.Namespace
+	} else {
+		if hcd.host.Spec.MetaData == nil {
+			hcd.log.Info("MetaData is not set returning empty(nil) data")
+			return "", nil
+		}
 		namespace = hcd.host.Namespace
+		name = hcd.host.Spec.MetaData.Name
 	}
 	return hcd.getSecretData(
 		ctx,
-		hcd.host.Spec.MetaData.Name,
+		name,
 		namespace,
 		"metaData",
 		false,
--- a/pkg/hostclaim/hostclaim_manager.go
+++ b/pkg/hostclaim/hostclaim_manager.go
@@ -28,15 +28,19 @@ import (
 
 	"github.com/go-logr/logr"
 	metal3api "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
+	"github.com/metal3-io/baremetal-operator/pkg/secretutils"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
+	"k8s.io/apimachinery/pkg/types"
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"sigs.k8s.io/cluster-api/util/conditions"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
 type HostManager struct {
@@ -50,6 +54,8 @@ const (
 	// PausedAnnotationValue is the value used to mark a BareMetalHost as paused by
 	// a HostClaim.
 	PausedAnnotationValue = "metal3.io/hostclaim"
+	// rebootDomain is the domain of metal3 reboot annotations.
+	rebootDomain = "reboot.metal3.io"
 	// Requeueing after 0 is in fact not requeuing.
 	TerminalReueueDelay time.Duration = 0
 	// Standard delay when waiting for other to settle.
@@ -64,8 +70,14 @@ const (
 	HostClaimKind = "HostClaim"
 )
 
-// An error used when there is no BMH satisfying the constraints.
-var ErrNoAvailableBMH = errors.New("no available BareMetalHost")
+var (
+	// An error used when there is no BMH satisfying the constraints.
+	ErrNoAvailableBMH = errors.New("no available BareMetalHost")
+	// An error raised when the secret to synchronize does not exists.
+	ErrNoSecret = errors.New("no Secret found")
+	// ErrNoBMH raised when the BareMetalHost is not found.
+	ErrNoBMH = errors.New("no BareMetalHost")
+)
 
 func NewHostManager(client client.Client, log logr.Logger, claim *metal3api.HostClaim, apireader client.Reader) (*HostManager, error) {
 	return &HostManager{
@@ -85,7 +97,7 @@ func (m *HostManager) SetConditionHostToFalse(
 	conditions.Set(m.HostClaim, metav1.Condition{Type: t, Status: metav1.ConditionFalse, Reason: reason, Message: message})
 }
 
-// SetConditionHostToFalse sets Host condition status to False.
+// SetConditionHostToTrue sets Host condition status to True.
 func (m *HostManager) SetConditionHostToTrue(
 	t string,
 	reason string,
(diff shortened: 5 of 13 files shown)


pkg/hostclaim/hostclaim_manager.go
```diff
@@ -157,6 +169,87 @@ func (m *HostManager) Associate(ctx context.Context) error {
 	return nil
 }
 
+// Update updates a machine and is invoked by the Machine Controller.
```
dtantsur: Outdated docstrings

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -157,6 +169,87 @@ func (m *HostManager) Associate(ctx context.Context) error {
...
+func (m *HostManager) Update(ctx context.Context) error {
+	m.Log.Info("Updating machine")
```
dtantsur: Please find and replace all cases of "machine" here

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -157,6 +169,87 @@ func (m *HostManager) Associate(ctx context.Context) error {
...
+
+	if err != nil {
+		m.SetConditionHostToFalse(
+			metal3api.SynchronizedCondition, metal3api.BareMetalHostNotSynchronizedReason,
+			"Failed to update BareMetalHost")
+		m.Log.Error(err, "Error while patching the BareMetalHost")
+		return hideConflictError(err)
+	}
+
+	// transient rebootAnnotation was successfully transmitted. We can delete it on HostClaim.
```
dtantsur, line 219: This is a bit unfortunate: this way it's hard to track when reboot is actually executed. But I don't see a good way out.

pkg/hostclaim/hostclaim_manager.go
dtantsur: And this should be `V(1)` too unless we detect that actual changes are needed

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -157,6 +169,87 @@ func (m *HostManager) Associate(ctx context.Context) error {
...
+		m.SetConditionHostToFalse(
+			metal3api.AssociatedCondition, metal3api.MissingBareMetalHostReason,
+			"Failed to get a BareMetalHost for the Host")
+		return err
+	}
+	if bmh == nil {
+		m.SetConditionHostToFalse(
+			metal3api.AssociatedCondition, metal3api.MissingBareMetalHostReason,
+			"BareMetalHost associated to the claim not found")
+		return errors.New("BareMetalHost not found")
```
dtantsur, line 187: I"m curious what the expected remediation is. Is a user supposed to delete and re-create the claim? Or is the controller supposed to provide another BMH?

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -157,6 +169,87 @@ func (m *HostManager) Associate(ctx context.Context) error {
...
+	}
+
+	bmh := metal3api.BareMetalHost{}
+	key := types.NamespacedName{
+		Name:      bmhRef.Name,
+		Namespace: bmhRef.Namespace,
+	}
+	err := m.client.Get(ctx, key, &bmh)
+	if k8serrors.IsNotFound(err) {
+		m.Log.Info("Annotated host not found", "bmh", bmhRef.Name, "bmhNamespace", bmhRef.Namespace)
```
dtantsur: s/annotated/linked/ ?

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -157,6 +169,87 @@ func (m *HostManager) Associate(ctx context.Context) error {
...
+		return nil, ErrNoBMH
+	} else if err != nil {
+		return nil, err
+	}
+	if !consumerRefMatches(bmh.Spec.ConsumerRef, hostClaim) {
+		m.Log.Info("The consumer ref does not point to the hostClaim", "consumerRef", bmh.Spec.ConsumerRef)
+		hostClaim.Status.BareMetalHost = nil
+		return nil, ErrNoBMH
```
dtantsur, line 253: Similar question here: what's the expected action in this case? Especially since we only provide a generic error message (which is understandable)?

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -167,6 +260,161 @@ func hideConflictError(err error) error {
...
+
+func (m *HostManager) synchronizeDataSecret(
+	ctx context.Context,
+	secretManager secretutils.SecretManager,
+	bmh *metal3api.BareMetalHost,
+	role string,
+	sourceRef *corev1.SecretReference,
+	namespace string,
+	hostName string,
+) (*corev1.SecretReference, error) {
```
dtantsur, line 364: This call should exit early if the target namespace is the same as the source namespace (a non-multi-tenant case).

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -435,6 +683,54 @@ func (m *HostManager) chooseBMH(ctx context.Context) (*metal3api.BareMetalHost,
...
+	m.HostClaim.Status.PoweredOn = bmh.Status.PoweredOn
+	m.HostClaim.Status.HardwareData = &metal3api.ObjectReference{
+		Namespace: bmh.Namespace,
+		Name:      bmh.Name,
+	}
+	conditions.SetMirrorCondition(bmh, m.HostClaim, metal3api.AvailableForProvisioningCondition)
+	conditions.SetMirrorCondition(bmh, m.HostClaim, metal3api.ProvisionedCondition)
+	m.SetConditionHostToTrue(metal3api.AssociatedCondition, metal3api.BareMetalHostAssociatedReason)
+
+	if !equality.Semantic.DeepEqual(m.HostClaim.Status, hostOld) {
```
dtantsur, line 708: And this is where logging something would actually be useful

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -157,6 +169,87 @@ func (m *HostManager) Associate(ctx context.Context) error {
...
+	updated, err := m.setBmhSpec(ctx, bmh)
+	if err != nil {
+		return err
+	}
+
+	if bmh.Annotations == nil {
+		bmh.Annotations = map[string]string{}
+	}
+
+	if syncReboot(m.HostClaim.Annotations, bmh.Annotations) {
```
dtantsur, line 200: We should also sync the detachment annotation (also maybe without the "force" flag being introduced in https://github.com/metal3-io/baremetal-operator/pull/2955 - i.e. sanitize the value). Detaching is commonly used in my world (e.g. by ACM) to prevent Ironic from further touching the deployed instance.

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -157,6 +169,87 @@ func (m *HostManager) Associate(ctx context.Context) error {
...
+		updated = true
+	}
+	if updated {
+		err = m.client.Update(ctx, bmh)
+	}
+
+	if err != nil {
+		m.SetConditionHostToFalse(
```
dtantsur: Maybe don't update the condition on conflict?

pkg/hostclaim/hostclaim_manager.go
pierrecregut: done

pkg/hostclaim/hostclaim_manager.go
pierrecregut: Yes and if the HostClaim patch fails we may even reboot twice in rare cases.

pkg/hostclaim/hostclaim_manager.go
pierrecregut: In the consumerRef case status.BaremetalHost is reset to nil, so the association logic is performed again (because in the controller introduced later the defer patch will be done as in m3m controller). This was not done here and I have added it.

pkg/hostclaim/hostclaim_manager.go
pierrecregut: see above.

pkg/hostclaim/hostclaim_manager.go
pierrecregut: Done

pkg/hostclaim/hostclaim_manager.go
pierrecregut: @dtantsur I have not modified this part yet. I want to be sure that we want to give control of the Detach annotation to the customer.... An analogy of hostclaim is taking a taxi rather than renting/owning a car: you do not have the keys. With detach, you ask the taxi driver to throw away the keys.... but you still do not have them. As it is done, I do not see a way of sharing the right to detach by both the customer and the infra owner (no consumerOverride).

pkg/hostclaim/hostclaim_manager.go
pierrecregut: done.

pkg/hostclaim/hostclaim_manager.go
dtantsur: We could have a knob to enable/disable it per HDP. But it's quite important for users to be able isolate their instance from Ironic's power management. It also allows reducing the load on Ironic.

I guess we'd need to re-attach on HostClaim deletion.

pierrecregut: I have added a way to authorize Detach in the HostDeployPolicy. (At least one matching policy must have the allowsDetachedMode flag set). I have also added a way to let admin detach the BMH independently of the HostClaim with a manager flag. 
I still do not understand how it can really be used in practice: how is the user managing the power setting without access to the BMC ? He can reboot or halt through the OS, but the first should not interfere with Ironic and if he uses the second, he will need Ironic to power on again. 
There is some kind of sanitization and the way it is done, it should "transparently" leaves off force flag when it is merged. 

I would keep it as a separate revertible commit for all the reasons above.

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -435,6 +700,92 @@ func (m *HostManager) chooseBMH(ctx context.Context) (*metal3api.BareMetalHost,
...
+			compatiblePolicies,
+			func(hdp *metal3api.HostDeployPolicy) bool {
+				return hdp.Spec.AllowsDetachedMode
+			},
+		) {
+			m.Log.Info("HostClaim not allowed to detach underlying BareMetalHost")
+			return false, err
+		}
+		inputArgs := metal3api.DetachedAnnotationArguments{}
+		_ = json.Unmarshal([]byte(annotValue), &inputArgs)
```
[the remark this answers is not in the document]
pierrecregut: Intentional. At least on BMH arbitrary values are authorized.

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -50,6 +55,8 @@ const (
 	// PausedAnnotationValue is the value used to mark a BareMetalHost as paused by
 	// a HostClaim.
 	PausedAnnotationValue = "metal3.io/hostclaim"
+	// rebootDomain is the domain of metal3 reboot annotations.
+	rebootDomain = "reboot.metal3.io"
```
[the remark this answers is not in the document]
pierrecregut: done

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -279,56 +541,59 @@ func (m *HostManager) acceptableNamespaces(ctx context.Context, namespace string
...
+	}
+	m.Log.Info("Acceptable namespaces", "namespaces", namespaces)
+	return namespaces, nil
+}
+
+func (m *HostManager) checkPolicy(hostDeployPolicy *metal3api.HostDeployPolicy, hostNs string, hostNsLabels map[string]string) (bool, error) {
+	log := m.Log.WithValues("policyNamespace", hostDeployPolicy.Namespace, "policyName", hostDeployPolicy.Name)
+	constraints := hostDeployPolicy.Spec.HostClaimNamespaces
+	if constraints == nil {
+		m.Log.V(1).Info("Ignoring HostDeployPolicy without constraint")
```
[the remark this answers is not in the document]
pierrecregut: done

pkg/hostclaim/hostclaim_manager_test.go
```diff
@@ -358,6 +368,292 @@ var _ = Describe("HostClaim manager", func() {
...
+			MetaData:    NewSecret("s1", HostclaimNamespace).SetData(map[string][]byte{"f": []byte("mdt")}).Build(),
+			NetworkData: NewSecret("s2", HostclaimNamespace).SetData(map[string][]byte{"f": []byte("nwdt")}).Build(),
+		}),
+		Entry("set meta-data/network-data (overide)", testCaseSetBMHSpec{
+			MetaData:       NewSecret("s1", HostclaimNamespace).SetData(map[string][]byte{"f": []byte("mdt")}).Build(),
+			NetworkData:    NewSecret("s2", HostclaimNamespace).SetData(map[string][]byte{"f": []byte("nwdt")}).Build(),
+			BMHMetaData:    NewSecret("bmh-metadata", "ns").SetData(map[string][]byte{"f": []byte("other")}).Build(),
+			BMHNetworkData: NewSecret("bmh-networkdata", "ns").SetData(map[string][]byte{"f": []byte("other")}).Build(),
+		}),
+		Entry("reset meta-data/network-data (overide)", testCaseSetBMHSpec{
```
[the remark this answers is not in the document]
pierrecregut: done

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -167,6 +271,164 @@ func hideConflictError(err error) error {
...
+func (m *HostManager) synchronizeDataSecret(
+	ctx context.Context,
+	secretManager secretutils.SecretManager,
+	bmh *metal3api.BareMetalHost,
+	role string,
+	sourceRef *corev1.SecretReference,
+	namespace string,
+	hostName string,
+) (*corev1.SecretReference, error) {
+	if namespace == bmh.Namespace {
```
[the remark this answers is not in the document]
pierrecregut: done (with err because of linter).

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -449,3 +800,33 @@ func SetContains[T comparable](m Set[T], s T) bool {
...
+	if nsLabels == nil {
+		nsLabels = map[string]string{}
+	}
+
+	var compatiblePolicies []*metal3api.HostDeployPolicy
+	for _, hdp := range policiesInNs.Items {
+		if accept, err := m.checkPolicy(&hdp, hostNs, nsLabels); err != nil {
+			return nil, err
+		} else if accept {
+			compatiblePolicies = append(compatiblePolicies, &hdp)
```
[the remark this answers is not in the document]
pierrecregut: done. Good catch.

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -712,7 +712,13 @@ type DetachedAnnotationArguments struct {
 
 	// Force indicates if detaching should be forced regardless of the host's state
 	// +optional
+
```
dtantsur: nit: this empty line should come after `Force` instead

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -712,7 +712,13 @@ type DetachedAnnotationArguments struct {
...
 	Force bool `json:"force,omitempty"`
+	// Manager is an informal flag that tells that the annotation was set by an administrator.
```
dtantsur: I don't think `Manager` is clear enough. Maybe be very-very explicit and call it something like `SetByHostClaim`?

apis/metal3.io/v1alpha1/hostdeploypolicy_types.go
```diff
@@ -25,6 +25,9 @@ type HostDeployPolicySpec struct {
 	// HostClaimNamespaces constrains the namespaces of the HostClaims allowed
 	// to bind the BareMetalHosts in the same namespace as the HostDeployPolicy
 	HostClaimNamespaces *HostClaimNamespaces `json:"hostClaimNamespaces,omitempty"`
+	// AllowsDetachedMode is a boolean flag specifying if the hostClaim can set the
+	// detached annotation of the BareMetalHost.
+	AllowsDetachedMode bool `json:"allowsDetachedMode,omitempty"`
```
dtantsur: nit: I don't think we call it a "mode", so something like `allowDetaching` would feel more consistent with our documentation to me

pkg/hostclaim/hostclaim_manager.go
dtantsur: I'm now curious about the difference between this case and the case on line 179. Line 179 is transient, line 185 presumably isn't. Are we treating them the same? I'd rather not dissociate host claims on a transient networking situation.

pkg/hostclaim/hostclaim_manager.go
dtantsur: > He can reboot or halt through the OS, but the first should not interfere with Ironic and if he uses the second, he will need Ironic to power on again.

I don't remember the details of what our folks are doing, unfortunately.

There is a parallel concern: detaching hosts significantly reduces the load on Ironic, helping with scaling.

To be clear, both cases are for trusted tenants.

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -167,6 +269,162 @@ func hideConflictError(err error) error {
...
+
+	m.SetConditionHostToTrue(metal3api.SynchronizedCondition, metal3api.ConfigurationSyncedReason)
+	return updated, nil
+}
+
+func referencesDiffer(ref1, ref2 *corev1.SecretReference) bool {
+	if ref1 == nil {
+		return ref2 != nil
+	}
+	return ref2 == nil || ref1.Name != ref2.Name
```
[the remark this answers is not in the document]
dtantsur: I agree, namespaces will mismatch in practice

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -167,6 +269,162 @@ func hideConflictError(err error) error {
...
+	secretManager secretutils.SecretManager,
+	bmh *metal3api.BareMetalHost,
+	role string,
+	sourceRef *corev1.SecretReference,
+	namespace string,
+	hostName string,
+) (*corev1.SecretReference, error) {
+	if namespace == bmh.Namespace {
+		if sourceRef == nil {
+			return nil, ErrNoSecret
```
dtantsur: Why is this an error? It seems odd to me to return an error just to ignore it at the caller side.

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -167,6 +269,162 @@ func hideConflictError(err error) error {
...
+		}
+		return sourceRef.DeepCopy(), nil
+	}
+	log := m.Log.WithValues("hostclaimName", hostName, "hostclaimNamespace", namespace, "secret-type", role)
+	secretName := bmh.Name + "-" + role
```
dtantsur: This logic established a name-based relationship that we haven't had so far. Are you trying to preserve an admin-created secret here? The problem is that these secrets are not always unique per host (OpenShift uses one user data secret per node role, for example).

I wonder if this aspect complicates the code unnecessarily. We don't provide the default image or the default custom deploy, why should we allow the default user or network data? Both are image-specific and won't just work with arbitrary images.

pkg/hostclaim/hostclaim_manager.go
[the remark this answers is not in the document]
pierrecregut: I thought that it was definitely agreed that bmh only referenced secrets in its own namespace and that the value of Namespace field was ignored to avoid creating security issues. That is the reason for copying all the secrets from hostClaim namespace to BMH namespace, otherwise a solution would be to reference the original one. But this is not acceptable in practice as a BMH owner could reference arbitrary secrets.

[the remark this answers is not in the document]
dtantsur: Ah, so it's all in the target namespace? I'd still validate namespace as a sort of defence-in-depth approach. Also then you may be able to use `reflect.DeepEqual` instead of the custom code?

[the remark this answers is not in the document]
pierrecregut: I think a more drastic change is first not to use corev1.SecretReference in hostclaims (see question on slack regarding v1beta2 of Metal3Machine). There is no retro compatibility to ensure, it could even be a string. I am not sure what is in-depth defense in that case. If namespace="" means namespace=&lt;resource namespace&gt;, then using deepequal may be not the right approach.

pkg/hostclaim/hostclaim_manager.go
pierrecregut: I am not sure, I follow the point. Images are URL, we can just copy the value. Here we have secrets in the hostclaim namespace which is not the baremetalhost namespace. We must copy them, so we need a name for the copy. We could either generate an arbitrary name but then we need to sync the names before generating the content because an update conflict will lead to a new name being used and secrets would not be reclaimed. Or we could use a fixed one. This is what is done here because it was simpler and there is no obvious drawback. We must only ensure that the secret is created/updated before the reference is set.

pkg/hostclaim/hostclaim_manager.go
pierrecregut: It is odd but the nilerr linter is unhappy with it. In fact most Err introduced in this file are NotFound errors and I really prefer testing for nil. But if we choose that path, we should do it everywhere and fight the linter with a lot of nolint comment.

pkg/hostclaim/hostclaim_manager.go
pierrecregut: The important point is do we keep the associated condition set to true ? If not, then the association has been broken (permanently) and we will run Associate on the error recovery Reconcile.

pkg/hostclaim/hostclaim_manager.go
dtantsur: I hate the nilerr linter and personally just disable it for the given line

pkg/hostclaim/hostclaim_manager.go
dtantsur: > In the consumerRef case status.BaremetalHost is reset to nil

I'm referring to this sentence. Are you suggested that every time `Update` returns an error, the controller unsets consumerRef? If so, it will work for permanent errors but it's an overkill for transient situations. Or do I misunderstand?

pkg/hostclaim/hostclaim_manager.go
dtantsur: Okay, I misread the following code a bit, but I think my argument holds. You cannot just assume that any secret that is called `bmh.Name + "-" + role` can be safely deleted. Nor can we assume that we'll never change this naming. I think there are two parts to this:

1. You need to make sure you're deleting the correct secret. You need to navigate to what BMH.Spec.{UserData/MetaData/NetworkData} points at and delete or replace that.
2. However! We need to ensure we don't accidentally override admin-provided secret. Which leads me to a broad question: what do we do if the selected BMH already has a secret attached to it (by mistake or on purpose)?

I think we need to label the copies somehow so that we can distinguish them from admin-created secret. If we find a labeled secret, we can safely delete or replace it. If we find something else... probably fail and let another host to be picked?

pierrecregut: It is some kind of convention over configuration in the same way that HardwareData have the BMH name. If it is set by Hostclaim the name is bmh-role (we could make it slightly more complex and especially mention hostclaim if you are afraid of confusion).  If it was set before by somebody else, the secret is not deleted unless it followed the convention. We could even strengthen to avoid the risk of deleting a secret by mistake (why is the user following the convention ?) by adding an additional label and checking it exists (if the secret manually set follows also that convention, the deletion is intentional). 

On the other hand it is clear that, with this approach,  we are not going to share secrets between baremetalhost as OpenShift is doing.

pkg/hostclaim/hostclaim_manager.go
pierrecregut: When status.BareMetalHost is set to nil, it means that either the BMH no longer exists (NotFound not a transient error) or that the consumerRef is pointing elsewheree (Bad !), these are not transient errors. On the other hand, I agree that on transient error, we should keep the Associated condition as is because we will not redo Association (as status.BareMetalHost is kept as said previously) and so we will not restore the value of the Associated condition.

pkg/hostclaim/hostclaim_manager.go
dtantsur: > It is some kind of convention over configuration in the same way that HardwareData have the BMH name. 

We already have a convention for linking secrets to BMH, and for better or worse it's not name-based.

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -713,6 +713,12 @@ type DetachedAnnotationArguments struct {
 	// Force indicates if detaching should be forced regardless of the host's state
 	// +optional
 	Force bool `json:"force,omitempty"`
+
+	// SetByHostClaim is an informal flag that tells that the annotation was set by the
```
dtantsur, line 717: nit: drop "informal", it's a part of the API in any case

apis/metal3.io/v1alpha1/hostclaim_types.go
```diff
@@ -119,6 +119,13 @@ type HostClaimSpec struct {
 	FailureDomain string `json:"failureDomain,omitempty"`
 }
 
+// SecretReference represents a Secret Reference in the same namespace as the resource.
+type SecretReference struct {
+	// name is unique within a namespace to reference a secret resource.
+	// +optional
```
dtantsur: Hmm, it's probably not optional in our case?

apis/metal3.io/v1alpha1/hostclaim_types.go
```diff
@@ -119,6 +119,13 @@ type HostClaimSpec struct {
...
+	Name string `json:"name,omitempty" protobuf:"bytes,1,opt,name=name"`
```
dtantsur: nit: I suspect we don't need to copy the "protobug" stuff

pkg/hostclaim/hostclaim_manager.go
dtantsur: Okay, let's circle back to the initial question: what happens if `Update` returns an error? How is a user notified and what is the remediation? Right now, it looks like you're returning the same thing for transient and fatal errors.

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -157,6 +169,94 @@ func (m *HostManager) Associate(ctx context.Context) error {
...
+	}
+
+	if bmh.Annotations == nil {
+		bmh.Annotations = map[string]string{}
+	}
+
+	if syncReboot(m.HostClaim.Annotations, bmh.Annotations) {
+		updated = true
+	}
+	if upd, err2 := m.syncDetached(ctx, bmh.Namespace, m.HostClaim.Annotations, bmh.Annotations); err2 != nil {
```
dtantsur: nit: s/upd/detachUpdated/

pkg/hostclaim/hostclaim_manager.go
pierrecregut: I guess I misunderstood what you wanted:  the capability to have defaults set by the infrastructure manager only optionally overridden by the tenant with a clean way back to the default when the hostclaim is deleted. The current implementation does not provide this feature but changing the handling of names will not solve it and the only clean way to get it is again a ``consumerOverride`` or ``consumerSettings`` field (the later name is better I think). Then I still think that my way of handling the naming is correct and simple. Here is a [draft of design proposal](https://gist.github.com/pierrecregut/8d424ef0401b1aef09a114bd2cd7aafb ) for the new field. Shall I propose it in metal3-docs ? As soon as it is implemented, changing the
setting of ``xxxData`` to ``consumerSettings.xxxData`` should provide the behavior you expect.

pierrecregut: I pushed a new version that:

* introduces a consumerSettings field, we can wait until there is a consensus on the design document pushed in parallel,
* keep the principle of a naming convention but adds an -hc- in the middle and put a label so that we are really sure we do not either overwrite or delete a secret that was not set by the hostclaim controller.

I am not really convinced the later is needed. If an infra manager uses a reserved name for a secret he should only expect to see his secret overwritten, not protected by the system.

pierrecregut: I reverted the consumerSettings field following the disucssion on the design document. Kept the defense to avoid overwrtting an existing secret with the wrong name provided by somebody else.

pierrecregut:
@dtantsur Here is an implementation of the alternative design. The new version adds two commits (to be squashed in a final version, at least for the one modifying host_manager) that implements the no-copy version of setBmh where the baremetal-operator is aware of HostClaims and follow the references. Because it uses the same logic as the previous version to validate the bound between HostClaim and BMH, it does not change the behavior for the provisionner but because the copies of secrets do not exist in the BMH namespace, nobody can read them there.

internal/controller/metal3.io/host_config_data.go
```diff
@@ -18,16 +21,14 @@ type hostConfigData struct {
 	host          *metal3api.BareMetalHost
 	log           logr.Logger
 	secretManager secretutils.SecretManager
+	hostClaim     *metal3api.HostClaim
```
dtantsur, line 34: Maybe I'm missing it, but you don't seem to be passing this value from the controller.

internal/controller/metal3.io/host_config_data.go
```diff
@@ -52,19 +53,58 @@ func (hcd *hostConfigData) getSecretData(ctx context.Context, name, namespace, d
...
+		return errors.New("nil consumerRef: no hostclaim for initialization data")
+	}
+	gv, err := schema.ParseGroupVersion(consumer.APIVersion)
+	if err != nil || gv.Group != metal3api.GroupVersion.Group {
+		return errors.New("bad group - no hostclaim for initialization data")
+	}
+	if consumer.Kind != hostclaim.HostClaimKind {
+		return errors.New("bad consumerRef kind: no hostclaim for initialization data")
+	}
+	hc := metal3api.HostClaim{}
```
dtantsur: I don't quite understand why you're loading a HostClaim here if you already pass the object to HostConfigData

internal/controller/metal3.io/host_config_data.go
```diff
@@ -52,19 +53,58 @@ func (hcd *hostConfigData) getSecretData(ctx context.Context, name, namespace, d
...
 	if hcd.host.Spec.UserData == nil {
 		hcd.log.Info("UserData is not set return empty string")
 		return "", nil
 	}
-	namespace := hcd.host.Spec.UserData.Namespace
-	if namespace == "" {
-		namespace = hcd.host.Namespace
+	name := hcd.host.Spec.UserData.Name
+	namespace := hcd.host.Namespace
+	if name == hostclaim.HostClaimReference {
```
dtantsur: Why do you need this name-based logic? I thought it would work as follows: if there is a consumerRef om BMH, and it is a link to a HostClaim, then (1) load the HostClaim, (2) use secrets that are linked to it.

We could, for the sake of visibility, extend the SecretReference to have a new field like `setByHostClaim` (similar to what you're adding for detachment) and omit name entirely. But in this case something will need to clean-up the references, which may not be worth the effort.

dtantsur:
Although... reading comments on https://github.com/metal3-io/metal3-docs/pull/691 and remembering more context, wasn't the goal first and foremost to access the cleaning mode override, not to avoid copying userData and company?

pkg/hostclaim/hostclaim_manager.go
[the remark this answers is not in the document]
zaneb: From a security perspective, I think it would be acceptable for the BMH to read a Secret from a remote namespace _provided_ that:

- The remote namespace is listed in the HostDeployPolicy
- A HostClaim in that namespace is the consumer of the BMH and references the named Secret

That confirms that a user in that namespace with permission to edit the HostClaim intended to give access to that Secret to the BMO. From a security perspective it's equivalent to copying the Secret.

That said, it might be simpler to just copy the Secret, given the amount of blurring the boundaries between the HostClaim and BMH controllers that would be involved.

For HostClaim itself, you never want to allow referencing a Secret from another namespace. It was a mistake to use SecretReference in BMH and this mistake has unfortunately propagated into HostClaim. If you still have the chance to change it, I would just use a string.

internal/controller/metal3.io/host_config_data.go
pierrecregut: It is not intended to be filled by the caller but only when needed (lazy evaluation).

internal/controller/metal3.io/host_config_data.go
pierrecregut: see above.

internal/controller/metal3.io/host_config_data.go
pierrecregut: It is part of the lazy evaluation approach. Unless the hostclaim tells there is a secret path to follow, the BMH does not even try to find the hostclaim. 
On the other hand, if we intend to use the hostclaim directly as source of information for most fields instead of copying its fields in ConsumerSettings., it is probably better to access the hostclaim early in the baremetal_host reconciliation and give the resource to subtask. (for example cleaningmode).

internal/controller/metal3.io/host_config_data.go
dtantsur: It may not be worth the effort given that the HostClaim is in the cache anyway.

internal/controller/metal3.io/host_config_data.go
dtantsur: The controller will cache the HostClaim anyway (especially when the HostClaims controller runs in the same process). Unless we see clear evidence that loading HostClaims is a bottleneck, I'd keep the code simple.

pierrecregut:
The current version no more use the lazy approach. hostclaim is fetched earlier and stored in the info structure. It will be available for defining the current value of automatedCleaningMode. CoPilot comments have been addressed.

pierrecregut:
> Although... reading comments on [metal3-io/metal3-docs#691](https://github.com/metal3-io/metal3-docs/pull/691) and remembering more context, wasn't the goal first and foremost to access the cleaning mode override, not to avoid copying userData and company?

With current approach, we do not need ConsumerSettings. That is why I also think the design PR can be closed... if we agree this is the right way to do it.

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -2565,6 +2575,43 @@ func (r *BareMetalHostReconciler) updateEventHandler(e event.UpdateEvent) bool {
...
+	},
+	UpdateFunc: func(_ context.Context, e event.TypedUpdateEvent[client.Object], q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+		objectNew, ok1 := e.ObjectNew.(*metal3api.HostClaim)
+		objectOld, ok2 := e.ObjectOld.(*metal3api.HostClaim)
+		if !ok1 || !ok2 {
+			return
+		}
+		if reflect.DeepEqual(objectNew.Spec.UserData, objectOld.Spec.UserData) &&
+			reflect.DeepEqual(objectNew.Spec.NetworkData, objectOld.Spec.NetworkData) &&
+			reflect.DeepEqual(objectNew.Spec.NetworkData, objectOld.Spec.MetaData) {
```
tuminoid: As noted by Copilot as well, copy paste'ish issue here on the types.

internal/controller/metal3.io/host_config_data.go
```diff
@@ -33,16 +31,13 @@ type hostConfigData struct {
...
 	log           logr.Logger
 	secretManager secretutils.SecretManager
+	hostClaim     *metal3api.HostClaim
 }
 
 // Generic method for data extraction from a Secret. Function uses dataKey
 // parameter to detirmine which data to return in case secret contins multiple
 // keys.
 func (hcd *hostConfigData) getSecretData(ctx context.Context, name, namespace, dataKey string, addFinalizer bool) (string, error) {
-	if namespace != hcd.host.Namespace {
```
tuminoid, line 42: This removes protection of reading Secrets outside your namespace, which I guess is needed for the feature. The getHostClaimIfExists validates the pointer, but the pointer is not immutable, so technically you can set it to point to secrets outside your namespace and get it to read it and pass to Ironic etc?

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -227,9 +233,11 @@ func (r *BareMetalHostReconciler) Reconcile(ctx context.Context, request ctrl.Re
 	}
 
 	initialState := host.Status.Provisioning.State
+	hostclaim, err := r.getHostClaimIfExists(ctx, host)
```
tuminoid, line 236: Err is never checked?

apis/metal3.io/v1alpha1/hostclaim_types.go
```diff
@@ -81,16 +81,16 @@ type HostClaimSpec struct {
 	// which is passed to the Config Drive and interpreted by the
 	// first-boot software such as cloud-init. The format of user data is
 	// specific to the first-boot software.
-	UserData *corev1.SecretReference `json:"userData,omitempty"`
```
tuminoid, line 84: This is an API change, so ⚠️ instead ✨  ?

internal/controller/metal3.io/baremetalhost_controller.go
pierrecregut: my bad. corrected.

internal/controller/metal3.io/host_config_data.go
pierrecregut: I am not sure I understand the point. There are two potential namespaces : hostclaim or bmh namespace. getSecretData argument is always one of them (never something passed by a end-user). Yes it only validates the reference at one point of time and the secret is accessed at another but this is always true in any setting. There are potentially two actors but if ironic access another namespace, at some point the hostclaim owner agreed on that and wrote a hostclaim manifest with the name of the secret in it.

internal/controller/metal3.io/baremetalhost_controller.go
pierrecregut: Oups ! Strange that the compiler did not complain on this never used variable.

apis/metal3.io/v1alpha1/hostclaim_types.go
pierrecregut: Is it not a little excessive :

* There is no controller for hostclaim  so no user.
* The api change just removes a useless field (note that I have not set it to a direct string for consistency and to keep the possibility of other values in the future).

internal/controller/metal3.io/host_config_data.go
tuminoid: I'm not sure I have a point yet beyond just noting that it removes the namespace check which typically is the protection for cross-namespace access which is generally the source of vulnerabilities in all controllers having cluster wide secrets access. I'll take another look with fresh eyes some other day.

apis/metal3.io/v1alpha1/hostclaim_types.go
tuminoid: OTOH its just emoji on PR title too, so is it excessive? :) No hard opinion really, just noting the API change. We're alpha v1 so its allowed to change no problem.

internal/controller/metal3.io/host_config_data.go
pierrecregut: Note that the code was slightly refactored. The original code was checking the values put by the user and then using those values. This code ignores user values because there is no choice and hardcode either host.namespace or hostclaim.namespace depending on the context. Checking the value provided by the user is left to the webhook.

apis/metal3.io/v1alpha1/hostclaim_types.go
pierrecregut: Well the core of the PR is adding code for a new feature. Removing the field is just a small cleanup with no real impact. See also my comment on the change of approach for namespace field in bmh.

zaneb: Removing an existing field without bumping the API to v2 is absolutely forbidden and will break all existing deployments of Metal³ that include the HostClaim CRD.

/hold for API breakage
Feel free to remove the hold when this is reverted.

pierrecregut: It does not break anything because there is no controller for hostclaims

elfosardo: @pierrecregut in the Kubernetes ecosystem, zaneb's argument is standard best practice.
Even if we're in alpha, abruptly removing fields from an existing schema version without deprecation notice or a version bump (e.g. v1alpha1 to v1alpha2) is generally forbidden because it breaks the declarative nature of Kubernetes manifests.
The field removal needs to be reverted or we need to introduce a proper API migration strategy to get this merged.

pierrecregut: We are in less than alpha: there is no controller and you cannot use hostclaims so making a migration strategy is ridiculous. I will revert back to the standard secretRef but this is a mistake. It is clear this cannot be done on bmh or m3m (there was a missed chance to do it with v1beta2).

elfosardo: @pierrecregut appreciate the flexibility to keep the schemas aligned.
I totally get the frustration since it's not fully wired up yet, but keeping the schema strictly additive prevents a lot of headaches down the line for anyone testing the raw CRDs.

zaneb: If this CRD hasn't been in a release yet then I'd say it is not too late to remove fields.
But it is not the case that you can remove fields from an API as long as it is in alpha.

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -2701,3 +2716,34 @@ func (r *BareMetalHostReconciler) reconcileHostData(ctx context.Context, host *m
...
+	if consumer.Kind != hostclaim.HostClaimKind {
+		return
+	}
+	hostclaim := &metal3api.HostClaim{}
+	key := client.ObjectKey{Namespace: consumer.Namespace, Name: consumer.Name}
+	if err = r.Get(ctx, key, hostclaim); err != nil {
+		return
+	}
+	if hostclaim.Status.BareMetalHost == nil {
+		err = errors.New("no back link on hostclaim - no hostclaim for initialization data")
```
dtantsur, line 2740: Making this condition an error allows a malformed HostClaim to block the BMH reconciliation. This is a potential DoS vulnerability. I suggest logging a warning and moving on.

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -2701,3 +2716,34 @@ func (r *BareMetalHostReconciler) reconcileHostData(ctx context.Context, host *m
...
+		return
+	}
+	if hostclaim.Status.BareMetalHost.Name != host.Name || hostclaim.Status.BareMetalHost.Namespace != host.Namespace {
+		err = errors.New("bad back link to baremetalhost - no hostclaim for initialization data")
```
dtantsur, line 2744: same

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -2701,3 +2716,34 @@ func (r *BareMetalHostReconciler) reconcileHostData(ctx context.Context, host *m
...
+	if err != nil || gv.Group != metal3api.GroupVersion.Group {
+		return
+	}
+	if consumer.Kind != hostclaim.HostClaimKind {
+		return
+	}
+	hostclaim := &metal3api.HostClaim{}
+	key := client.ObjectKey{Namespace: consumer.Namespace, Name: consumer.Name}
+	if err = r.Get(ctx, key, hostclaim); err != nil {
+		return
```
dtantsur, line 2737: Hmm. Maybe filter out NotFound? It's a possible situation when the claim is force-removed.

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -713,6 +713,12 @@ type DetachedAnnotationArguments struct {
...
+	// HostClaim controller. HostClaim controller will never delete an annotation without
+	// SetByHostClaim set to true
+	// +optional
+	SetByHostClaim bool `json:"setByHostClaim,omitempty"`
```
dtantsur, line 721: By the way, now that the BMH controller has access to HostClaims, do we need this? It looks like we can make the BMH controller check both places when detecting detachment (and take into account HDP, obviously).

internal/controller/metal3.io/host_config_data.go
dtantsur: As part of defence in depth, I'd rather not rely on webhooks. But it looks like the logic is safe here. Only the error message is removed.

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -157,6 +166,91 @@ func (m *HostManager) Associate(ctx context.Context) error {
 	return nil
 }
 
+// Update updates a hostclaim and is invoked by the HostClaim Controller.
+func (m *HostManager) Update(ctx context.Context) error {
+	m.Log.V(1).Info("Updating HostClaim")
+
```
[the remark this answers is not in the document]
dtantsur: @pierrecregut do you think it will be hard to wire `Update` in the reconciler? Doing so would massively simplify testing. We could even start adding functional tests as the next step.

pkg/hostclaim/hostclaim_manager.go
dtantsur: As an aside, I want to rethink rebooting, servicing and other annotation-based logic. We can leave reboot as it is for now.

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -167,6 +261,48 @@ func hideConflictError(err error) error {
...
+			updated = true
+			bmh.Spec.CustomDeploy.Method = m.HostClaim.Spec.CustomDeploy.Method
+		}
+	}
+
+	if bmh.Spec.Online != m.HostClaim.Spec.PoweredOn {
+		updated = true
+		bmh.Spec.Online = m.HostClaim.Spec.PoweredOn
+	}
+
```
dtantsur, line 301: Thinking aloud: can we/should we set *something* in UserData/MetaData/NetworkData to indicate that it's present, just not as a local secret?

We cannot set the real value because the webhook will reject the namespace. Maybe we copy the core SecretReference as a local type (bonus: deprecate the namespace field) and add something like a boolean flag `setByHostClaim`?

dtantsur:
Added a few new comments, also some old ones are still outstanding. Also needs a rebase.

I'd suggest, for the sake of making progress with this work, to split the PR into sequential logical parts:

1. Create a fake Update and wire it all the way into the reconciler (otherwise, we cannot even test the feature, even though the API is already present).
2. Implement copying of Image, CustomDeploy, and Online (the simplest bit that makes something work)
3. Add logic for *Data secrets
4. Handle reboots
5. Handle detachments

I know that it's more work, but I'm starting being worried about how long it takes with this PR, and how it keeps changing its shape.

pierrecregut:
To follow @dtantsur  request, that there is a controller first. This PR is paused until #3406 is merged.

pierrecregut:
Following the merge of end to end tests for hostclaims, this PR is amended and put out of draft mode:

* its scope has been limited to the propagation of standard fields (image, power). As discussed above, BMH can access the initialization data directly on the hostclaim. But this will be done in a separate commit.
* handling detached is complex and subject to discussion. It will be handled in another PR
* reboot annotations are handled. This could be removed as there are plans for more declarative actions. 
* Delete is implemented.
* End to end test have been amended. But as there is still no user-data, there is no ssh check.
* As pointed out in #3407, getBmh is a security barrier and this is checked in the e2e test.

PR #3162, opened 2026-04-14
🐛Add webhook readiness check before starting manager

MahnoorAsghar:
**What this PR does / why we need it**:
Add a readiness check for webhooks to prevent the pod starting before the webhooks are ready to serve requests

Assisted-by: Claude

**Checklist:**

- [ ] Documentation has been updated, if necessary.
- [ ] Unit tests have been added, if necessary.
- [ ] E2E tests have been added, if necessary.
- [ ] Integration tests have been added, if necessary.

PR #3163, opened 2026-04-15
🐛 controller: cap BareMetalHost reconcile backoff at 30 seconds

MahnoorAsghar:
The default controller-runtime exponential rate limiter has a maximum delay of 1000 seconds (~16 minutes).  During BMO startup there is a brief race between the reconcile loop becoming active and BMO's own validating-webhook Service endpoint being propagated: any BareMetalHost reconciliation that triggers a webhook call in that window gets a "no endpoints available" error.  A burst of such errors is enough to drive the per-item exponential backoff to its ceiling, after which BMO silently waits up to 16 minutes before attempting the next reconcile even though the webhook has long since become reachable.

Replace the default rate limiter with one that is otherwise identical but caps the per-item exponential delay at 30 seconds.  This bounds the recovery window to at most one retry interval after the endpoint is propagated, matching the behaviour of other Metal3 controllers.

Generated-by: Cursor, claude-4.6-sonnet-medium model

**Checklist:**

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,6 +16,7 @@ require (
 	github.com/stretchr/testify v1.11.1
 	go.etcd.io/etcd/client/pkg/v3 v3.6.10
 	go.uber.org/zap v1.27.1
+	golang.org/x/time v0.9.0
 	k8s.io/api v0.34.6
 	k8s.io/apimachinery v0.34.6
 	k8s.io/client-go v0.34.6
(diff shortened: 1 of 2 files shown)


dtantsur:
Is https://github.com/metal3-io/baremetal-operator/pull/3162 solving the same problem? Do we need both?

MahnoorAsghar:
@dtantsur They're both needed, AFAIU

Sunnatillo (commented):
Sorry, But I am not convinced this change is needed. 
IMHO, the time between webhook ready and endoint available is not huge. Lets says 10s(probably way less). The exponential backoff never reaches 16 minutes that we are saying here.

MahnoorAsghar:
The downstream failures may have a tangle of other causes....Investigating further.

PR #3164, opened 2026-04-15
✨ Add Starlark-based BYO provisioner

s3rj1k:
Add [Starlark](https://starlark-lang.org/)-based BYO provisioner

Adds a provisioner backend whose Provisioner interface methods are
implemented by functions in a user-supplied Starlark script, selected at
startup with `--starlark-provisioner-script`. This lets operators adapt
Ironic provisioning behavior without forking BMO or rebuilding the
controller image or using custom provisioner (no Ironic).

The Starlark runtime, builtins (HTTP, JSON/YAML, logging, events,
getenv/read_file), and Go↔Starlark conversion are provided by the new
`pkg/provisioner/starlark/` subpackages.

A reference implementation, `efi-vmedia.star`, targets Ironic's
redfish-virtualmedia boot interface with UEFI and supports both
custom-deploy and direct (qcow2) image deployments.

**[ToDo] Checklist:**


dtantsur:
Starlark is a pretty exotic thing (I haven't heard about it before), I wonder it's better for a wider audience to just support shell scripts (and pass all arguments as JSON in stdin, for example).

s3rj1k:
> Starlark is a pretty exotic thing (I haven't heard about it before), I wonder it's better for a wider audience to just support shell scripts (and pass all arguments as JSON in stdin, for example).

We can discuss this on next community call, but there is reason why this is using Starlark and not JS or Lua for example, and it is effectively only one, Starlark is more lightweight and supported vs other possible embeddable runtimes (and I did consider JS, Lua, WASM and few others).

We can replace Starlark with other runtimes if needed, I am after acceptance on general approach, allow actual BYO provisioner usage during runtime instead of compile-time. 

> just support shell scripts

I believe that a fork penalty here will be high, however this could work also.

dtantsur:
We'll see what other people think but I'd rather not pick any runtime at all. Because yes, we pick this, and someone will ask for JS, and then somebody wants Lua...

Also what happened with the idea of turning Provisioner interface into actual plugins?

s3rj1k:
> Also what happened with the idea of turning Provisioner interface into actual plugins?

Still on the table if we are referencing go plugins, only issue that I see with it is CGO, BMO would need to have it enabled and provide "SDK" to compile against.

By SDK I mean, just stable dockerfile that was used to compile BMO binary releases.

--

I do think that go plugins are way better for this kind of thing but they do require to cross that CGO mental block.

--

Also go-plugins would require a lot of Release and CI discipline, basically each release would need to be tested against some stub plugin implementation, this is doable, just highlighting.

s3rj1k:
Status: Rebased on top of `OutOfTree` plugin system that will be released in v0.14.0

s3rj1k:
For `demo` purposes this has been migrated to dedicated repo https://github.com/s3rj1k/metal3-baremetal-operator-starlark

PR #3165, opened 2026-04-15
🐛 Add missing `context` to `PreprovisioningImageFormats` func

s3rj1k:
Adds missing `context` function argument to `PreprovisioningImageFormats`

This is needed for cases when custom provisioner gets implemented, against existing Provisioner interface.


dtantsur:
I'm now curious why you need a context in this particular function, but I see nothing against it, so

/lgtm

Rozzii (commented):
I am not against this either. But if you are working on custom providers that will utilize PPI then please  follow the development of this PR https://github.com/metal3-io/baremetal-operator/pull/2576
/approve

PR #3168, opened 2026-04-17
:seedling: fix help target alignment and numeric target names

terror96:
The awk pattern in the help target used [a-zA-Z_-]+ which excludes digits, causing targets like `verify-e2e-prerequisites` and `test-e2e` to be omitted from the help output.

Add 0-9 to the character class and increase the column width from 15 to 30 characters so all target descriptions align consistently.

Makes info for all `makefile` targets visible.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,7 +79,7 @@ endif
 
 .PHONY: help
 help:  ## Display this help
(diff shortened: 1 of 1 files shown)


lentzi90 (commented):
/approve

Sunnatillo (commented):
/lgtm

PR #3181, opened 2026-04-21
:sparkles: Allow exiting the externally provisioned state

### LINKED ISSUE #2465: Confusing behavior on removing `externallyProvisioned: true`

dtantsur:
This change unblocks setting externallyProvisioned to false to allow
a host to follow the regular lifecycle.

During the transition, image and customDeploy are copied over to the
status to make sure the host does not get unconditionally deprovisioned.
Hosts without image and customDeploy will still be deprovisioned.

This is an independently written alternative to #2566 since that PR
got abandoned.

Fixes: #2465 
Generated-By: Claude Code (commercial license)

### DIFF
--- a/internal/controller/metal3.io/host_state_machine.go
+++ b/internal/controller/metal3.io/host_state_machine.go
@@ -433,11 +433,17 @@ func (hsm *hostStateMachine) handleExternallyProvisioned(ctx context.Context, in
 		return hsm.Reconciler.actionManageSteadyState(ctx, hsm.Provisioner, info)
 	}
 
-	if hsm.Host.NeedsHardwareInspection() {
-		hsm.NextState = metal3api.StateInspecting
-	} else {
-		hsm.NextState = metal3api.StatePreparing
+	// The host is exiting externally provisioned at this point.
+	// Set image and customDeploy in status to prevent unconditional deprovisioning.
+	if hsm.Host.Spec.Image != nil {
+		hsm.Host.Status.Provisioning.Image = *hsm.Host.Spec.Image
 	}
+	if hsm.Host.Spec.CustomDeploy != nil {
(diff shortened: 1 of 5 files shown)


internal/controller/metal3.io/host_state_machine.go
```diff
@@ -464,11 +464,17 @@ func (hsm *hostStateMachine) handleExternallyProvisioned(ctx context.Context, in
...
+	// Set image and customDeploy in status to prevent unconditional deprovisioning.
+	if hsm.Host.Spec.Image != nil {
+		hsm.Host.Status.Provisioning.Image = *hsm.Host.Spec.Image
 	}
+	if hsm.Host.Spec.CustomDeploy != nil {
+		hsm.Host.Status.Provisioning.CustomDeploy = hsm.Host.Spec.CustomDeploy.DeepCopy()
+	}
+
+	// Move to Provisioned. If image and customDeploy are not set, deprovisioning will start on the next reconciliation.
+	hsm.NextState = metal3api.StateProvisioned
```
tuminoid, line 446: Should we reset error count here?

dtantsur: I would expect `actionManageSteadyState` to happen on the next reconciliation, doing the needful (just like on line 463)

lentzi90 (commented):
/approve

tuminoid (commented):
/lgtm

tuminoid (commented):
/lgtm

PR #3186, opened 2026-04-24
🐛 Reject non-empty checksum for OCI images

mabulgu:
## Summary
- `GetChecksum()` made checksums optional for OCI images but never rejected a non-empty checksum
- This causes the webhook to accept invalid configurations where both `oci://` scheme and a checksum are provided, leading to confusing provisioning failures
- Now returns an error when checksum is set for OCI images

## Test plan
- [x] Unit tests updated (`TestGetImageChecksum` "OCI image with checksum" case)
- [x] Verify `make test` passes
- [ ] Verify BareMetalHost with `oci://` image URL and non-empty checksum is rejected
- [ ] Verify BareMetalHost with `oci://` image URL and empty checksum is accepted

Assisted by: Claude Opus 4.6

### DIFF
--- a/apis/metal3.io/v1alpha1/baremetalhost_types.go
+++ b/apis/metal3.io/v1alpha1/baremetalhost_types.go
@@ -1103,8 +1103,10 @@ func (image *Image) GetChecksum() (checksum, checksumType string, err error) {
 		return "", "", nil
 	}
 
-	// Checksum is not required for OCI images as they have embedded checksums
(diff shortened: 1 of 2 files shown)


smoshiur1237 (commented):
/lgtm

tuminoid (commented):
/lgtm

kashifest (commented):
/approve

PR #3196, opened 2026-04-30
:sparkles: Validation webhook for hostclaims

pierrecregut:

This PR adds a validation webhook for hostclaims.
* Checks that hostSelector requirements are valid (both keys and values constraints for matchLabels and matcExprs, also operation coherence for matchExpr).
* Checks image coherence.

**Checklist:**

### DIFF
--- a/config/base/webhook/manifests.yaml
+++ b/config/base/webhook/manifests.yaml
@@ -46,3 +46,24 @@ webhooks:
     resources:
     - bmceventsubscriptions
   sideEffects: None
+- admissionReviewVersions:
+  - v1
+  - v1beta1
+  clientConfig:
+    service:
+      name: webhook-service
+      namespace: system
+      path: /validate-metal3-io-v1alpha1-hostclaim
+  failurePolicy: Fail
+  name: hostclaim.metal3.io
+  rules:
+  - apiGroups:
+    - metal3.io
+    apiVersions:
+    - v1alpha1
+    operations:
+    - CREATE
+    - UPDATE
+    resources:
+    - hostclaims
+  sideEffects: None
--- a/internal/webhooks/metal3.io/v1alpha1/hostclaim_validation.go
+++ b/internal/webhooks/metal3.io/v1alpha1/hostclaim_validation.go
@@ -1,0 +1,104 @@
+/*
+Copyright 2026 The Metal3 Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package webhooks
+
+import (
+	"fmt"
+	"strings"
+
+	metal3api "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
+	"k8s.io/apimachinery/pkg/selection"
+	"k8s.io/apimachinery/pkg/util/validation"
+)
+
+// validateHostClaim validates a HostClaim resource.
+func (webhook *HostClaimWebhook) validateHostClaim(hostclaim *metal3api.HostClaim) []error {
+	var errs []error
+	for lblKey, lblValue := range hostclaim.Spec.HostSelector.MatchLabels {
+		for _, err := range validation.IsQualifiedName(lblKey) {
+			errs = append(errs, fmt.Errorf("%s=%s: %s", lblKey, lblValue, err))
+		}
+		for _, err := range validation.IsValidLabelValue(lblValue) {
+			errs = append(errs, fmt.Errorf("%s=%s: %s", lblKey, lblValue, err))
+		}
+	}
+	for i, hsr := range hostclaim.Spec.HostSelector.MatchExpressions {
+		for _, err := range validation.IsQualifiedName(hsr.Key) {
+			errs = append(errs, fmt.Errorf("matchExpr %d: %s", i+1, err))
+		}
+		switch hsr.Operator {
+		case selection.Equals, selection.DoubleEquals, selection.NotEquals:
+			if len(hsr.Values) != 1 {
+				errs = append(errs, fmt.Errorf(
+					"matchExpr %d: exactly one value for operator %s in match expression on label key %s",
+					i+1, hsr.Operator, hsr.Key))
+			} else {
+				for _, err := range validation.IsValidLabelValue(hsr.Values[0]) {
+					errs = append(errs, fmt.Errorf("matchExpr %d (value %q): %s", i+1, hsr.Values[0], err))
+				}
+			}
+		case selection.In, selection.NotIn:
+			if len(hsr.Values) == 0 {
+				errs = append(errs, fmt.Errorf(
+					"matchExpr %d: At least one value in list for operator %s in match expression on label key %s",
+					i+1, hsr.Operator, hsr.Key))
+			}
+			for _, v := range hsr.Values {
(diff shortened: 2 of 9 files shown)


pierrecregut:
This should address @dtantsur comment: https://github.com/metal3-io/baremetal-operator/pull/2996#discussion_r2846930268
This is implemented as a regular webhook, not through CEL expressions because the budget required could be arbitrarily high.

internal/webhooks/metal3.io/v1alpha1/hostclaim_validation.go
```diff
@@ -0,0 +1,92 @@
...
+				errStrings = append(errStrings, fmt.Sprintf(
+					"matchExpr %d: values not authorized for operator %s in match expression on label key %s",
+					i+1, hsr.Operator, hsr.Key))
+			}
+		default:
+			errStrings = append(errStrings, fmt.Sprintf(
+				"matchExpr %d: invalid operation %s in match expression on label key %s", i+1, hsr.Operator, hsr.Key))
+		}
+	}
+	var errs = make([]error, len(errStrings))
```
dtantsur: nit: why not make `[]error` right away and populate it by appending `fmt.Errorf`?

internal/webhooks/metal3.io/v1alpha1/hostclaim_validation.go
```diff
@@ -0,0 +1,92 @@
...
+
+func validateHostclaimAnnotations(hostclaim *metal3api.HostClaim) []error {
+	var errs []error
+	var err error
+
+	for annotation, value := range hostclaim.Annotations {
+		switch {
+		case strings.HasPrefix(annotation, metal3api.RebootAnnotationPrefix+"/") || annotation == metal3api.RebootAnnotationPrefix:
+			err = validateRebootAnnotation(value)
+		// TODO: should we check Detached annotation.
```
dtantsur, line 94: Probably yes, even if we don't allow it by default right now

internal/webhooks/metal3.io/v1alpha1/hostclaim_validation.go
pierrecregut: Indeed. First version was concatenating directly the output of validation.IsValid.... functions but there was not enough context, so I had to iterate over the result to add the info. There is no more reason to keep this intermediate array.

internal/webhooks/metal3.io/v1alpha1/hostclaim_validation.go
pierrecregut: What do you mean by "we don't allow it"  ?  (https://github.com/metal3-io/baremetal-operator/pull/3152/changes/59b2f9dfb1028413df03f7456bbab08165515357).
We could constrain the values to be valid json. with both boolean fields set to false.

internal/webhooks/metal3.io/v1alpha1/hostclaim_validation.go
```diff
@@ -0,0 +1,106 @@
...
+import (
+	"fmt"
+	"strings"
+
+	metal3api "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
+	"k8s.io/apimachinery/pkg/selection"
+	"k8s.io/apimachinery/pkg/util/validation"
+)
+
+// validateHost validates HostClaim resource for creation.
```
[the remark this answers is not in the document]
lentzi90: Valid comment

config/base/webhook/manifests.yaml
```diff
@@ -46,3 +46,24 @@ webhooks:
     resources:
     - bmceventsubscriptions
   sideEffects: None
+- admissionReviewVersions:
+  - v1
+  - v1beta
+  clientConfig:
```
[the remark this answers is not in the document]
lentzi90: We have the same for bmceventsubscription above, so I would say disregard for now. We should clean these all up though. I don't think there is a reason to keep v1beta in there.

internal/webhooks/metal3.io/v1alpha1/hostclaim_validation.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	if err := validateHostclaimAnnotations(hostclaim); err != nil {
+		errs = append(errs, err...)
+	}
+
+	if hostclaim.Spec.Image != nil {
+		if err := validateImage(hostclaim.Spec.Image); err != nil {
+			errs = append(errs, err...)
+		}
```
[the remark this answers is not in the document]
lentzi90: Looks valid also to me

internal/webhooks/metal3.io/v1alpha1/hostclaim_webhook.go
```diff
@@ -0,0 +1,59 @@
...
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type.
+func (webhook *HostClaimWebhook) ValidateCreate(_ context.Context, hc *metal3api.HostClaim) (admission.Warnings, error) {
+	hostclaimlog.Info("validate create", "name", hc.Name, "namespace", hc.Namespace)
+	return nil, kerrors.NewAggregate(webhook.validateHostClaim(hc))
+}
+
+// ValidateUpdate implements webhook.Validator so a webhook will be registered for the type.
+func (webhook *HostClaimWebhook) ValidateUpdate(_ context.Context, _, newHc *metal3api.HostClaim) (admission.Warnings, error) {
+	hostclaimlog.Info("validate update", "name", newHc.Name, "namespace", newHc.Namespace)
```
lentzi90, line 62: Could you add the same `nil` checks as here: https://github.com/metal3-io/baremetal-operator/blob/d41c7b2bfbb8cda7c9dddf43baffdbaa1057add0/internal/webhooks/metal3.io/v1alpha1/baremetalhost_webhook.go#L56-L64

internal/webhooks/metal3.io/v1alpha1/hostclaim_webhook.go
```diff
@@ -0,0 +1,59 @@
...
+// hostclaimlog is for logging in this webhook.
+var hostclaimlog = logf.Log.WithName("webhooks").WithName("HostClaim")
+
+func (webhook *HostClaimWebhook) SetupWebhookWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewWebhookManagedBy(mgr, &metal3api.HostClaim{}).
+		WithValidator(webhook).
+		Complete()
+}
+
+//+kubebuilder:webhook:verbs=create;update,path=/validate-metal3-io-v1alpha1-hostclaim,mutating=false,failurePolicy=fail,sideEffects=none,admissionReviewVersions=v1;v1beta1,groups=metal3.io,resources=hostclaims,versions=v1alpha1,name=hostclaim.metal3.io
```
lentzi90, line 39: I see `v1beta1` here but `v1beta` (without the `1`) in the manifests. Maybe missing `make generate` (although CI should have caught that)?

In any case, I think `v1beta1` is the correct and we just have a typo in the other webhooks. I also think we can remove v1beta1.

I'll create an issue for cleaning these up.

lentzi90 (commented):
Thanks for this!
There are a couple of `nil` checks I think we should add (see below). Otherwise this is LGTM

internal/webhooks/metal3.io/v1alpha1/hostclaim_webhook.go
pierrecregut: Because it is not regenerated... The Makefile is wrong. It will be fixed in next push.

pierrecregut: @lentzi90 I can add the commit fixing #3266 on top of this merge request. There is also a nil check missing in ValidateCreate of the baremetalhost webhook.

internal/webhooks/metal3.io/v1alpha1/hostclaim_webhook.go
pierrecregut: done with other suggested changes from copilot.

lentzi90 (commented):
/lgtm
Thanks for fixing that Makefile issue!

PR #3210, opened 2026-05-04
🐛 Create ports if they are already inspected and machine is re-registering

### LINKED ISSUE #3211: Ports are not recreated if Ironic database has been dropped.
1. Have hosts in provisioned stage.
2. Delete Ironic instance.
3. Let BMO sync to Ironic
4. Ironic doesn't have ports for hosts
5. As side effect, any IPA interaction would fail.

Ports should be recreated after Ironic database drop, from already inspected data exists in `HardwareData`

**Environment:**

- Baremetal Operator version: v0.12.4

/kind bug

1nv0k32:
Fixes #3211

**Checklist:**


internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -888,6 +888,7 @@ func (r *BareMetalHostReconciler) registerHost(ctx context.Context, prov provisi
 			HasCustomDeploy:            hasCustomDeploy(info.host),
 			DisablePowerOff:            info.host.Spec.DisablePowerOff,
 			CPUArchitecture:            getHostArchitecture(info.host),
+			HardwareData:               info.host.Status.HardwareDetails,
```
dtantsur: Mostly an aside: the HardwareDetails field is deprecated, we'll move on to using HardwareData CRs

pkg/provisioner/ironic/register.go
```diff
@@ -309,6 +309,21 @@ func (p *ironicProvisioner) ensurePort(ctx context.Context, ironicNode *nodes.No
 			if err != nil {
 				return err
 			}
+
+			if len(nics) > 0 {
```
dtantsur: nit: len check is not needed (range works without it)

pkg/provisioner/ironic/register.go
```diff
@@ -309,6 +309,21 @@ func (p *ironicProvisioner) ensurePort(ctx context.Context, ironicNode *nodes.No
...
+				macs := map[string]string{}
+
+				for _, nic := range nics {
+					macs[nic.MAC] = nic.Name
+				}
+
+				for mac, _ := range macs {
+					err = p.createPXEEnabledNodePort(ctx, ironicNode.UUID, mac)
```
dtantsur: You cannot assume that all ports for PXE enabled ones. Check the `nic.pxe` field instead.

pkg/provisioner/ironic/register.go
```diff
@@ -292,7 +292,7 @@ func (p *ironicProvisioner) enrollNode(ctx context.Context, data provisioner.Man
 	return ironicNode, false, nil
 }
 
-func (p *ironicProvisioner) ensurePort(ctx context.Context, ironicNode *nodes.Node) error {
+func (p *ironicProvisioner) ensurePort(ctx context.Context, ironicNode *nodes.Node, nics []metal3api.NIC) error {
```
dtantsur: ensurePort is only called when bootMACAddress is set, it's probably not what you want to modify.

tuminoid (commented):
/ok-to-test

pkg/provisioner/ironic/register.go
1nv0k32: But do we need to check this field? Shouldn't we add the ports anyways? 🤔

internal/controller/metal3.io/baremetalhost_controller.go
1nv0k32: Now using `*metal3api.HardwareData`

pkg/provisioner/ironic/register.go
1nv0k32: Moved out of here, after creating node if it doesn't exist.

pkg/provisioner/ironic/ironic.go
```diff
@@ -296,18 +296,26 @@ func (p *ironicProvisioner) findExistingHost(ctx context.Context, bootMACAddress
...
+	p.log.Info("creating ironic port for node", "NodeUUID", uuid, "MAC", macAddress, "PXE status", pxe)
 
-	enable := true
+	portsList, errPortList := p.listAllPorts(ctx, macAddress)
+	if errPortList != nil {
+		p.log.Info("failed to look for existing ports in Ironic", "MAC", macAddress)
+		return errPortList
+	}
+	if len(portsList) > 0 {
+		p.log.Info("port already exists in Ironic", "MAC", macAddress)
```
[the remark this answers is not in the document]
1nv0k32: Not valid, cause the port should be created in any case.
The problematic case is when we import a node without inspection and we push hardwareData to Ironic.

1nv0k32:
@dtantsur @tuminoid 
Can you please take a look at this one?
Thanks in advance.

pkg/provisioner/ironic/register.go
```diff
@@ -140,6 +147,13 @@ func (p *ironicProvisioner) Register(ctx context.Context, data provisioner.Manag
 			}
 		}
 
+		// Try to create ports from HardwareData when re-registering
+		err = p.createPortsFromHardwareData(ctx, ironicNode, data.HardwareData)
```
tuminoid: This runs on every reconcile, not just reregister, which ends up listing all ports on node, wasting up to NIC amount of calls to Ironic. Use `nodeHasAssignedPort` to guard?

1nv0k32: Well technically, `nodeHasAssignedPort` does the same, it calls Ironic to list all the ports to check if the node has ports :thinking:
 
And also, I was wondering if we need to have the registered ports being part of the BMH's status data? (Currently, they are only part of hardware data, and they don't reflect if they are registered in Ironic or not. Personally, everytime I want to check if the node has registered ports, I need to go to Ironic shell.)

pkg/provisioner/ironic/ironic.go
```diff
@@ -185,48 +163,28 @@ func (p *ironicProvisioner) getNode(ctx context.Context) (*nodes.Node, error) {
 	return nil, fmt.Errorf("failed to find node by ID %s: %w", p.nodeID, err)
 }
 
-// Verifies that node has port assigned by Ironic.
-func (p *ironicProvisioner) nodeHasAssignedPort(ctx context.Context, ironicNode *nodes.Node) (bool, error) {
+// get ports in Ironic with address or node uuid filter.
+func (p *ironicProvisioner) getPorts(ctx context.Context, nodeUUID string, macAdress string) ([]ports.Port, error) {
```
tuminoid:
```suggestion
func (p *ironicProvisioner) getPorts(ctx context.Context, nodeUUID string, macAddress string) ([]ports.Port, error) {
```

pkg/provisioner/ironic/ironic.go
```diff
@@ -185,48 +163,28 @@ func (p *ironicProvisioner) getNode(ctx context.Context) (*nodes.Node, error) {
...
-		return false, fmt.Errorf("failed to page over list of ports: %w", err)
+	if nodeUUID != "" {
+		opts.NodeUUID = nodeUUID
 	}
-
-	empty, err := allPages.IsEmpty()
-	if err != nil {
-		return false, fmt.Errorf("failed to check port list status: %w", err)
+	if macAdress != "" {
+		opts.Address = macAdress
```
tuminoid:
```suggestion
	if macAddress != "" {
		opts.Address = macAddress
```

test/e2e/inspection_test.go
```diff
@@ -168,6 +169,40 @@ var _ = Describe("Inspection", Label("required", "inspection"), func() {
...
+			})
+
+			By("Waiting for BMH to be reconciled")
+			WaitForBmhReconciled(ctx, clusterProxy.GetClient(), bmh,
+				e2eConfig.GetIntervals("default", "wait-deployment")...)
+
+			By("Get ports in Ironic after dropping the database and reconciling and check if they are the same")
+			Eventually(func(g Gomega) {
+				portsAfter, errPotrsAfter := getIronicPorts(ctx, e2eConfig)
+				g.Expect(errPotrsAfter).NotTo(HaveOccurred())
```
tuminoid:
```suggestion
				portsAfter, errPortsAfter := getIronicPorts(ctx, e2eConfig)
				g.Expect(errPortsAfter).NotTo(HaveOccurred())
```

tuminoid (commented):
Just couple typos, LGTM otherwise to me. All the previous comments are addresses AFAICS.

tuminoid:
If it fails, its still about the arm cache CI failure which is being investigated.

test/e2e/common.go
```diff
@@ -1046,6 +1052,49 @@ func dumpIronicNodes(ctx context.Context, e2eConfig *Config, artifactFolder stri
 	}
 }
 
+func getIronicClient(e2eConfig *Config) (*gophercloud.ServiceClient, error) {
```
lentzi90: Since https://github.com/metal3-io/baremetal-operator/pull/3072/changes#diff-9e3e0045b2433cf2f909c70750b8c101357568967023ccad15e1983073d425cd, we have similar code to this in https://github.com/metal3-io/baremetal-operator/blob/f6014df64ac9fadd4099b7675cf287362445570a/test/e2e/ironic_helpers.go#L23
Please rebase and use that.

lentzi90 (commented):
Is it necessary to scale down/up ironic for the e2e test, or could we get away with a detach instead? If we can avoid the scaling I would prefer that. Scaling adds time and makes it harder to understand the logs, since we have logs from two separate pods then.

test/e2e/common.go
1nv0k32: Good catch, this change was after my PR being made, on it :D

1nv0k32: > Is it necessary to scale down/up ironic for the e2e test, or could we get away with a detach instead? If we can avoid the scaling I would prefer that. Scaling adds time and makes it harder to understand the logs, since we have logs from two separate pods then.

@lentzi90 By scaling up/down the Ironic deployment, I wanted to simulate real behaviour of Ironic database being dropped, by my limited knowledge, not sure if detaching would do the same :thinking: 
If you know it's better idea, I would apply

lentzi90: Detach basically means that the node is removed from Ironic. I would expect that we get the same situation as when scaling, but in a "safer" way. It is documented here: https://book.metal3.io/bmo/detached_annotation.html. We [have an e2e test](https://github.com/metal3-io/baremetal-operator/blob/f6014df64ac9fadd4099b7675cf287362445570a/test/e2e/provisioning_and_annotation_test.go#L131-L148) for it also where we could possibly add this check. It detaches a host in provisioned state though. Do you need it to be available?

Anyway, I realized there is an even worse issue with the scale down. We run tests in parallel! When we scale down ironic in one test, that can affect other tests running at the same time. So we really need to do it differently, or isolate the test completely.

lentzi90 (commented):
Thanks!
/lgtm

I think we may want to split the e2e test out into its own thing at some point, but this is fine for now.

1nv0k32:
> Thanks!
> /lgtm
> 
> I think we may want to split the e2e test out into its own thing at some point, but this is fine for now.

I was thinking to separate but it would add more time to tests :)

lentzi90:
> I was thinking to separate but it would add more time to tests :)

Yeah I was more thinking in the direction of making some tests optional. We have an optional job already that has all the upgrade tests for now. They run nightly, so we will quickly notice if things break, even when they are not triggered on PRs. We could split and move this test there, perhaps also some other tests.

I'll think about it a bit more and see if I can come up with some reasonable plan. We may want to restructure a bit more while at it.

lentzi90 (commented):
/lgtm

pkg/provisioner/ironic/ironic.go
```diff
@@ -296,18 +254,53 @@ func (p *ironicProvisioner) findExistingHost(ctx context.Context, bootMACAddress
...
+	portsList, errPortList := p.getPorts(ctx, "", macAddress)
+	if errPortList != nil {
+		p.log.Info("failed to look for existing ports in Ironic", "MAC", macAddress)
+		return errPortList
+	}
+	for _, port := range portsList {
+		if port.NodeUUID != uuid {
+			// try to check if the node exists or the port is orphaned
+			_, err := nodes.Get(ctx, p.client, port.NodeUUID).Extract()
+			if gophercloud.ResponseCodeIs(err, http.StatusNotFound) {
```
dtantsur: This cannot happen, Ironic delete ports when a node is deleted

pkg/provisioner/ironic/ironic.go
```diff
@@ -296,18 +254,53 @@ func (p *ironicProvisioner) findExistingHost(ctx context.Context, bootMACAddress
...
+				p.log.Info(
+					"the port is orphaned, deleting and recreating",
+					"MAC", macAddress,
+					"old NodeUUID", port.NodeUUID,
+					"current NodeUUID", uuid,
+				)
+				err = ports.Delete(
```
dtantsur: This can be a bad idea. Sometimes, users make a mistake and create a Host with an existing bootMACAddress. Your change will force the original port to be deleted, which will cause a competition between two hosts.

pkg/provisioner/ironic/register.go
```diff
@@ -280,35 +279,51 @@ func (p *ironicProvisioner) enrollNode(ctx context.Context, data provisioner.Man
...
+	nodeAssignedPorts, err := p.getPorts(ctx, ironicNode.UUID, "")
 	if err != nil {
 		return err
 	}
 
-	if !nodeHasAssignedPort {
-		addressIsAllocatedToPort, err := p.isAddressAllocatedToPort(ctx, p.bootMACAddress)
-		if err != nil {
-			return err
+	isInAssignedPorts := func(address string) bool {
```
dtantsur: nit: an alternative would be to create a `map[string]Port` and use it later

test/e2e/common.go
```diff
@@ -1115,6 +1118,30 @@ func dumpIronicNodes(ctx context.Context, e2eConfig *Config, artifactFolder stri
 	}
 }
 
+func getIronicNodePorts(ctx context.Context, e2eConfig *Config, nodeName string) (ports []ironicPort.Port, err error) {
```
dtantsur: nit: move to ironic_helpers

test/e2e/inspection_test.go
```diff
@@ -151,6 +154,78 @@ var _ = Describe("Inspection", Label("required", "inspection"), func() {
...
+			}, e2eConfig.GetIntervals(specName, "wait-detached")...)
+
+			By("Retrieving the latest BMH object")
+			err = clusterProxy.GetClient().Get(ctx, types.NamespacedName{
+				Name:      bmh.Name,
+				Namespace: bmh.Namespace,
+			}, &bmh)
+			Expect(err).NotTo(HaveOccurred())
+
+			By("Get ports in Ironic after detachment and check if all the ports are removed")
```
dtantsur: You're actually checking that the node has been removed (line 1127)

test/e2e/inspection_test.go
```diff
@@ -151,6 +154,78 @@ var _ = Describe("Inspection", Label("required", "inspection"), func() {
...
+
+			By("Retrieving the latest BMH object")
+			err = clusterProxy.GetClient().Get(ctx, types.NamespacedName{
+				Name:      bmh.Name,
+				Namespace: bmh.Namespace,
+			}, &bmh)
+			Expect(err).NotTo(HaveOccurred())
+
+			By("Get ports in Ironic after re-attachment and check if they are the same")
+			Eventually(func(g Gomega) {
```
dtantsur: I wonder if you actually need Eventually here. We're not supposed to report OK until the node is fully re-created in Ironic. If it does not happen, it's a bug.

pkg/provisioner/ironic/ironic.go
1nv0k32: I also seen that behaviour, but placed it to prevent a small chance where the port is orphaned for whatever reason.
But if you say the chance is 0 (I can imagine that the port has FK to Node object), I will remove :D

pkg/provisioner/ironic/ironic.go
1nv0k32: That's why I'm checking if the node exists for each port in the list in line 269.
Only when the /nodes query returns 404, I'm proceeding to delete the port. (same topic of the above comment)

pkg/provisioner/ironic/register.go
dtantsur: What I mean here: instead of calling isInAssignedPorts each time, you can compose a map of addresses to Port objects and use it later.

1nv0k32: That wasn't a map to be used later. That was searching in the already existing ports in Ironic to not to create existing port.
But now, I made a map out of it and using go's native way to check if it exists or not, lesser for loops :D

1nv0k32:
@dtantsur @lentzi90 I guess it's ready for another review, thanks in advance :D

lentzi90 (commented):
/lgtm

PR #3214, opened 2026-05-05
:sparkles: Add AutomatedCleaningPhase API field

honza:
Introduce a new `AutomatedCleaningPhase` field on `BareMetalHostSpec` that controls when automated cleaning is performed during deprovisioning.

The field has two values:
- "deprovisioning" (default): cleaning runs on every deprovisioning, including deletion. This preserves existing behavior.
- "reprovisioning": cleaning only runs between provisionings and is skipped when the host is being deleted (e.g. namespace terminating).

This allows users to explicitly opt out of cleaning on deletion without violating the existing API semantics of `AutomatedCleaningMode`.

### DIFF
--- a/apis/metal3.io/v1alpha1/baremetalhost_types.go
+++ b/apis/metal3.io/v1alpha1/baremetalhost_types.go
@@ -575,6 +575,18 @@ type BareMetalHostSpec struct {
 	// +kubebuilder:validation:Optional
 	AutomatedCleaningMode AutomatedCleaningMode `json:"automatedCleaningMode,omitempty"`
 
+	// Defines the phase of deprovisioning where automated cleaning is
+	// performed. When set to "deprovisioning" (the default), cleaning
+	// runs on every deprovisioning including deletion. When set to
+	// "reprovisioning", cleaning is skipped when the host is being
+	// deleted, and only runs between provisionings.
+	// This field is only relevant when automatedCleaningMode is not
+	// set to "disabled".
+	// +optional
+	// +kubebuilder:default:=deprovisioning
+	// +kubebuilder:validation:Optional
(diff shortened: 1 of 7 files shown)


apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -610,6 +622,21 @@ const (
...
 
+// AutomatedCleaningPhase indicates when automated cleaning should be performed.
+// +kubebuilder:validation:Enum:=deprovisioning;reprovisioning
+type AutomatedCleaningPhase string
+
+// Allowed automated cleaning phases.
+const (
+	// CleaningPhaseDeprovisioning means cleaning runs on every
+	// deprovisioning, including when the host is being deleted.
+	CleaningPhaseDeprovisioning AutomatedCleaningPhase = "deprovisioning"
```
dtantsur, line 633: I'm not sure "deprovisioning" conveys well that it also happens on deletion, but I don't have better ideas.

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -610,6 +622,21 @@ const (
...
+
+	// CleaningPhaseReprovisioning means cleaning runs only between
+	// provisionings. Cleaning is skipped when the host is being deleted.
+	CleaningPhaseReprovisioning AutomatedCleaningPhase = "reprovisioning"
```
dtantsur, line 637: This is also not perfect: a user can request a deprovisioning without a new provisioning by deleting the image/customDeploy.

dtantsur:
The idea is good, the bikeshedding will be brutal :)

honza:
Closing in favour of #3354

Issue #3195, opened 2026-04-30
E2E tests with Ironic do not clean-up on failures


dtantsur:
Currently, clean-up (e.g. removal of the BMH and its secrets) happens in one of two cases:

1. Manually in the end of a successful test run
2. In AfterEach but only NAMESPACE_SCOPED is false, and it's true for Ironic-based tests

As a result, if a test fails or if its author forgets the manual deletion, the BMH will remain for the duration of the run, interfering with other tests and causing their cascading failures (usually with a registration timeout).

Example: https://github.com/metal3-io/baremetal-operator/actions/runs/25124220454?pr=3072

/kind ci

PR #3243, opened 2026-05-11
🌱 E2E: Verify IPA boot was managed by Ironic

### LINKED ISSUE #3242: E2E: Verify used iPXE script
We are not currently verifying which iPXE script is used when booting IPA in the E2E tests.
The point has been to use the custom script created by Ironic, but in fact we sometimes end up with the fallback inspector.ipxe. This happened for example when a bug was introduced in ironic-image, that broke the iPXE script (see https://github.com/metal3-io/ironic-image/issues/971).

We should come up with a way to verify which script is used in E2E to avoid surprises like this.
Normally, our IPMI or Redfish tests should fail if we end up falling back to inspector.ipxe. It may be worth creating a specific test for verifying the fallback though.

lentzi90:

The point of this is to make sure we notice if a BMH boots using the fallback inspector.ipxe script instead of the ironic "controlled" way (either per-node ipxe script or virtualmedia).

Fixes #3242

**Checklist:**

### DIFF
--- a/test/e2e/inspection_test.go
+++ b/test/e2e/inspection_test.go
@@ -161,6 +161,13 @@ var _ = Describe("Inspection", Label("required", "inspection"), func() {
 			State:  metal3api.StateAvailable,
 		}, e2eConfig.GetIntervals(specName, "wait-available")...)
 
+		if e2eConfig.GetVariable("SSH_CHECK_PROVISIONED") == "true" {
+			By("verifying IPA booted from the expected Ironic-managed boot source")
+			VerifyIronicManagedBoot(e2eConfig, bmc.Address, bmc.IPAddress)
+		} else {
+			Logf("WARNING: Skipping boot source verification since SSH_CHECK_PROVISIONED != true")
+		}
+
 		By("Delete BMH")
 		err = clusterProxy.GetClient().Delete(ctx, &bmh)
 		Expect(err).NotTo(HaveOccurred())
--- a/test/e2e/common.go
+++ b/test/e2e/common.go
@@ -537,6 +537,114 @@ func PerformSSHBootCheck(e2eConfig *Config, expectedBootMode string, ipAddress s
 	Expect(isExpectedBootMode).To(BeTrue(), fmt.Sprintf("Expected booting from %s, but found different mode", expectedBootMode))
 }
 
+// UsesIPXE returns true if the given BMC address uses iPXE boot
+// (as opposed to virtual media boot). Virtual media addresses contain
+// "virtualmedia" in the protocol prefix.
+func UsesIPXE(bmcAddress string) bool {
+	return !strings.Contains(bmcAddress, "virtualmedia")
+}
+
+// VerifyIronicManagedBoot SSHes into IPA during inspection and checks
+// /proc/cmdline to verify that IPA was booted by an Ironic-managed process
+// (per-node iPXE config or virtual media ISO), not the inspector.ipxe fallback.
+//
+// For iPXE BMCs (IPMI, Redfish), this verifies that the per-node iPXE config
+// (rendered from ipxe_config.template) was used instead of the fallback. The
+// per-node template hardcodes "selinux=0" in the kernel command line, which
+// inspector.ipxe does not include.
+//
+// For virtual media BMCs (Redfish-VirtualMedia), this verifies that IPA
+// booted from the Ironic-managed virtual media ISO by checking for
+// "boot_method=vmedia" in the kernel command line. This parameter is set by
+// Ironic's redfish boot module and is the same signal IPA uses internally
+// to determine it was booted from virtual media.
+//
+// The detection relies on kernel parameters from Ironic's kernel_append_params
+// configuration (e.g. "nofb") which are included in both per-node iPXE configs
+// and virtual media ISOs, but NOT in inspector.ipxe. Additionally, "selinux=0"
+// is hardcoded in the iPXE per-node template but not in kernel_append_params,
+// so its presence confirms iPXE boot and its absence (combined with "nofb")
+// confirms virtual media boot.
+func VerifyIronicManagedBoot(e2eConfig *Config, bmcAddress, ipAddress string) {
+	Logf("Verifying IPA boot source by SSHing into IPA at %s", ipAddress)
+
+	client := EstablishSSHConnection(e2eConfig, ipAddress)
+	defer func() {
+		if client != nil {
+			client.Close()
+		}
+	}()
+
+	output, err := executeSSHCommand(client, "cat /proc/cmdline")
+	Expect(err).NotTo(HaveOccurred(), "Failed to read /proc/cmdline from IPA via SSH")
+
+	cmdline := strings.TrimSpace(output)
+	Logf("IPA kernel command line: %s", cmdline)
+
+	// Sanity check: verify that the custom marker we set in
+	// Ironic.spec.deployRamdisk.extraKernelParams actually made it into
+	// the kernel command line. This confirms that Ironic's kernel parameter
+	// pipeline (extraKernelParams -> IRONIC_KERNEL_PARAMS -> kernel cmdline)
+	// is working correctly. Note that this marker appears in ALL boot paths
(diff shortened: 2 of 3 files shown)


lentzi90:
/hold
We need to verify that it works.

I am not completely happy with the solution so far. We are parsing random kernel arguments that happen to be different between the fallback, per-node script and virtualmedia.
Perhaps we should investigate other ways? One thing I was thinking about is the serial logs. We collect them from the VMs, but then that only works for VMs the way it is currently implemented. I would want a solution that is also feasible when pointing the e2e suit to a real bare metal lab.

lentzi90:
/cc @nuhakala
you have recently worked on the e2e networking, what do you think about this?

lentzi90:
```
[90](https://github.com/metal3-io/baremetal-operator/actions/runs/25669187293/job/75350134062?pr=3243#step:5:691)
  STEP: verifying IPA booted from the expected Ironic-managed boot source @ 05/11/26 12:20:44.252
  INFO: Verifying IPA boot source by SSHing into IPA at 192.168.222.123
  INFO: IPA kernel command line: BOOT_IMAGE=/vmlinuz nofb nomodeset vga=normal ipa-insecure=1 "sshkey=ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIOPeKE3dG2YlNHZ4BTBE7nRku1RJqZjtW7B5vfV+u83c" console=ttyS0 bmo-e2e-kernel-params-ok systemd.journald.forward_to_console=yes ir_pub_id=51c26a79-170e-4ca6-bb45-3bef57acedc2 ipa-inspection-collectors=default,logs ipa-enable-vlan-interfaces=all ipa-inspection-dhcp-all-interfaces=1 ipa-collect-lldp=1 ipa-inspection-callback-url=https://192.168.222.2:6385/v1/continue_inspection ipa-api-url=https://192.168.222.2:6385/ ipa-agent-token=s-GdJ2uw0E_Crt9-dr7fMSAX2tMS5jSTjZNEHfS4rLM ipa-debug=1 boot_method=vmedia --
  INFO: Confirmed: IPA booted from the Ironic virtual media ISO
```

Interesting! I was not aware of `boot_method=vmedia`. Checking if we can use this...

dtantsur:
> Interesting! I was not aware of boot_method=vmedia. Checking if we can use this...

This is a part of the internal contract between Ironic and IPA. It comes with no warranty, but it has been like that for years, so I don't think it will just disappear overnight.

test/e2e/inspection_test.go
```diff
@@ -154,6 +154,13 @@ var _ = Describe("Inspection", Label("required", "inspection"), func() {
 			State:  metal3api.StateInspecting,
 		}, e2eConfig.GetIntervals(specName, "wait-inspecting")...)
 
+		if e2eConfig.GetVariable("SSH_CHECK_PROVISIONED") == "true" {
+			By("verifying IPA booted from the expected Ironic-managed boot source")
+			VerifyIronicManagedBoot(e2eConfig, bmc.Address, bmc.IPAddress)
```
dtantsur: Potentially risky. State `inspecting` does not mean that the ramdisk is already booted.

nuhakala (commented):
I can't think of any other way than SSH'ing into the nodes. There was discussion about making some kind of agent that would run on the target machine: https://github.com/metal3-io/metal3-dev-env/issues/955 but I think checking the boot source from the proper OS is as ambiguous as this way. And it would also be slower as one would need to wait for the OS to boot.

You could add dedicated params to the kernel command line in each boot method which indicate the boot source. That way we would not need to rely on any "random" parameters.

test/e2e/inspection_test.go
lentzi90: Ah, good point! I can move it down to after inspection finishes.

lentzi90:
Looks like we are at least catching the current situation!
```
[FAILED] Kernel command line does not contain 'nofb'. IPA appears to have booted from the inspector.ipxe fallback instead of an Ironic-managed boot (per-node iPXE or virtual media). Full cmdline: ironic-python-agent.kernel ipa-insecure=1 ipa-inspection-collectors=default,logs systemd.journald.forward_to_console=yes BOOTIF=00:60:2f:31:81:02 ipa-debug=1 ipa-enable-vlan-interfaces= ipa-inspection-dhcp-all-interfaces=1 ipa-collect-lldp=1 ipa-inspection-callback-url=https://192.168.222.2:6385/v1/continue_inspection ipa-api-url=https://192.168.222.2:6385/ initrd=ironic-python-agent.initramfs sshkey="ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIKwJWaCcl13WsoBJhMjite1/h1UgqNXX7cMimYesz5pX" console=ttyS0 bmo-e2e-kernel-params-ok
```
I am not sure why some other tests later failed though. Investigating...

lentzi90:
Quite happy with this now! The test failure is simply showing that we detect the ironic-image bug causing IPA to boot from the fallback script.
/hold cancel

lentzi90:
Forgot to mention the secondary failures. Those were because of #3195. Working on a fix for that in #3245.

Rozzii:
@lentzi90 checking the cmldine is very useful, but this PR might be  assuming a bit too much. SSHing into the IPA could fail for different reasons, I am not against this check I think the current functionality is useful but first IMO the pxe logs have to be checked in /var/log/libvirt/qemy/node<number>-serial.log . If "inspector.ipxe" is mentioned than we know explicitly that the fallback has been initiated. I think if we check both the pxe logs and the ssh then we can nicely distinguish between different early boo failures.

lentzi90:
I was looking for ways to check those logs also, but they all seem to assume we run on VMs. We do collect the logs in CI, but I don't want to build the tests so that they only work there. Ideally we should be able to run this on real hardware and then we don't have the logs there.
If you know a way to get them, e.g. through the ironic-log-collector container then I would be very interested! So far I have not had much luck getting any logs from there though.

Rozzii:
> I was looking for ways to check those logs also, but they all seem to assume we run on VMs. We do collect the logs in CI, but I don't want to build the tests so that they only work there. Ideally we should be able to run this on real hardware and then we don't have the logs there. If you know a way to get them, e.g. through the ironic-log-collector container then I would be very interested! So far I have not had much luck getting any logs from there though.

I have an other alternative suggestion, we can get the the serverside logs of the "GET" requests from the httpd container in the ironic pod. 

There is stuff like this` [12/May/2026:18:05:08 +0000] "GET /v1/ports?fields=node_uuid&node_uuid=66d9e577-7fac-4937-afc9-e1d4f033c07c HTTP/1.1" 200 537 "-" "gophercloud/v2.12.0"` in the logs and we can compare the uuid with the provisioner uuid from the  BMH status data.

I don't want to block your PR so if you don't want to implement this now I can make a separate Issue and someone can do it later.

Rozzii (commented):
As for review, as said this approach isokay but it is not  only testing the managed boot, but won't block this either after the tests will be passing.

lentzi90:
Great stuff! The tests are passing, confirming the fix worked!
I would suggest we take this in as a first step, so we have at least something in place to detect regressions.

Let's work on improvements and additions then as a separate issue. Getting the serial logs in some way that works for bare metal would be great. Parsing the Ironic logs would also be nice, but I am not sure how easy it will be to distinguish between the tests, when there are multiple running in parallel.

Rozzii:
> Great stuff! The tests are passing, confirming the fix worked! I would suggest we take this in as a first step, so we have at least something in place to detect regressions.
> 
> Let's work on improvements and additions then as a separate issue. Getting the serial logs in some way that works for bare metal would be great. Parsing the Ironic logs would also be nice, but I am not sure how easy it will be to distinguish between the tests, when there are multiple running in parallel.

We are not supposed to fall back to using inspector.ipxe AFAIK in any BMO scenarios or at least in the ones we are testing now so I would still suggest  checking the logs of the httpd eventually.

Rozzii (commented):
/approve

tuminoid (commented):
/lgtm

PR #3246, opened 2026-05-12
:bug: Fix reconciling HFS created before BMH and add e2e tests

dtantsur:
Currently, the BMH controller sets an OwnerReference on HFC objects
regardless of what created them. For HFS, on the other hand, only
controller-created objects get an OwnerReference. Not only is it
inconsistent, but it also opens a possibility of a bug: if the HFS is
reconciled before the corresponding BMH is in the controller cache,
the HFS will skip reconciliation and will never be reconciled again.

Adjust the event handler to allow reconciliations when only owner
references are changed.

Also fix the obviously wrong error message. And update log messages.

Also add e2e tests that verify HFS (with servicing too).

### DIFF
--- a/config/overlays/e2e/firmware-settings/kustomization.yaml
+++ b/config/overlays/e2e/firmware-settings/kustomization.yaml
@@ -1,0 +1,7 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- ../roles-rolebindings
+- namespace.yaml
+
+namespace: firmware-settings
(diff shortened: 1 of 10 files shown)


dtantsur:
/retest

Mirror problems

dtantsur:
This test caught its first bug before even getting merged: https://review.opendev.org/c/openstack/ironic/+/988275

iurygregory:
looks good to me, going to read a bit more about the e2e since I'm not 100% familiar with it.

Rozzii (commented):
/approve

PR #3249, opened 2026-05-15
:seedling: Bump golang.org/x/net to v0.53.0

kashifest:
This PR has 2 commits

- Bump golang.org/x/net to v0.53.0 

Note that this requires a bump of go lang in go mod files to go 1.25.0. 

Reason for bump:

Fixes critical HTTP/2 infinite loop DoS (GO-2026-4918) where a
malicious SETTINGS_MAX_FRAME_SIZE=0 frame could hang all connections.
Also fixes GO-2026-4441 infinite parsing loop.

- fix golangci-lint findings (deprecated comments, prealloc, and targeted gosec/revive suppressions)

This comes as a response to the OSV scanner failures in the CI


Sunnatillo (commented):
/lgtm

lentzi90 (commented):
This is ok for me, but let's make sure we have consensus in the community.
/hold
/approve

I think we have three options for how to handle this:
1. (this PR) fix the issue by bumping the versions and explain in the release notes that the breaking change was needed for security
2. argue that the issue is not critical for us and does not need to be patched (e.g. BMO only communicates with trusted servers (Kubernetes API and Ironic), so an adversary server is not part of the attack surface we care about)
3. (semver absolutist) the fix is breaking, thus we have no fix for this minor. Users that want a fix must upgrade to a newer minor version.

Rozzii (commented):
I would go with option 2 honestly, I don't see the go bump justified in our oldest supported release branch.

tuminoid:
> This is ok for me, but let's make sure we have consensus in the community. /hold /approve
> 
> I think we have three options for how to handle this:
> 
> 1. (this PR) fix the issue by bumping the versions and explain in the release notes that the breaking change was needed for security
> 2. argue that the issue is not critical for us and does not need to be patched (e.g. BMO only communicates with trusted servers (Kubernetes API and Ironic), so an adversary server is not part of the attack surface we care about)
> 3. (semver absolutist) the fix is breaking, thus we have no fix for this minor. Users that want a fix must upgrade to a newer minor version.

I reviewed the issues fixed in 1.25.10 and x/net:

- 3 are build time only, mitigated by trusted GOPROXY, not runtime issues anyways
- 7 are in packages we don't use (net/mail etc)
- 1 (the http2) hits, but we are exposed to it in only 2 scenarios:
    1. as k8s API client, that would require k8s API server be malicious, then this vuln is least of our worries
    2. ironic API connection, which doesn't use http2, but 1.1, no problem there

IMO there is zero incentive for us to bump go.mod to 1.25 and break consumers.

lentzi90:
Thanks @tuminoid for the analysis! I was hoping that this might be the case.
Let's then close this and instead figure out how to silence the scanners
/approve cancel

tuminoid:
> Thanks @tuminoid for the analysis! I was hoping that this might be the case. Let's then close this and instead figure out how to silence the scanners /approve cancel

Lets close then as Kashif is out for a while.
/close

PR #3260, opened 2026-05-19
🐛 Missing RootVolume attribute for RAID volumes

### LINKED ISSUE #3270: Explicit RootVolume option for RAID
In case you don't specify `rootDeviceHints` and there is raid configuration, BMO supposed to consider that raid volume as root device.
But somewhere else in the code, if the `rootDeviceHints` is empty, it will fill it from hardware data which will defeats the purpose of the logic above.

1. If there is raid config and no `rootDeviceHints`, consider that as root.
2. Don't populate `rootDeviceHints` if raid is configured.
3. Have an explicit option for a raid volume to be considered as root volume.

**Environment:**

- Baremetal Operator version: v0.13.0

/kind bug

1nv0k32:

Adding missing feature from Ironic for choosing raid volume as root device.
It should be validated to only be set on one volume.
It should also prevent rootDeviceHint to set itself automatically from hardwareData in case there is one raid root volume.

Fixes #3270

**Checklist:**


1nv0k32:
/cc @dtantsur @lentzi90 
Thanks in advance :D

pkg/provisioner/ironic/raid.go
```diff
@@ -170,9 +170,10 @@ func buildTargetSoftwareRAIDCfg(volumes []metal3api.SoftwareRAIDVolume) (logical
 	for _, volume := range volumes {
 		// Build logicalDisk
 		logicalDisk = nodes.LogicalDisk{
-			SizeGB:     volume.SizeGibibytes,
-			RAIDLevel:  nodes.RAIDLevel(volume.Level),
-			Controller: "software",
+			SizeGB:       volume.SizeGibibytes,
+			RAIDLevel:    nodes.RAIDLevel(volume.Level),
+			Controller:   "software",
+			IsRootVolume: volume.RootVolume,
```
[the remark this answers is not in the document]
1nv0k32: There is another bug that I just discovered.
`setTargetRAIDCfg` this function is testing if `RootDeviceHints == nil`
But if you don't set it, BMO sets it as default which causes `is_root_volume` to never set on any raid volume.
Gonna fix that :D

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -368,6 +368,10 @@ type BMCDetails struct {
 
 // HardwareRAIDVolume defines the desired configuration of volume in hardware RAID.
 type HardwareRAIDVolume struct {
+	// Setting this to true, causes installer to consider this volume as root device for installation.
+	// This can only be set for one of the volumes.
+	RootVolume *bool `json:"rootVolume,omitempty"`
```
dtantsur: I don't think any of the hardware RAID backends actually support `is_root_volume`. Unless you're sure, let's not add it to HardwareRAIDVolume.

1nv0k32: But seems like it's openstack option for all the raid types :thinking: 
https://docs.openstack.org/ironic/latest/admin/raid.html

1nv0k32: Isn't it just hinting openstack which volume to use as root? Does it actually need support from hardware RAID backend? :thinking:

kubajj: It does say:

> **For hardware RAID**, this value currently only affects the create_root_volume and create_nonroot_volumes parameters to the create_configuration step, but out-of-tree RAID interfaces may also use it to populate root device hints. Otherwise, you need to populate the hints yourself.

> **For software RAID** (since the Gazpacho release - through the change [962670](https://review.opendev.org/c/openstack/ironic/+/962670/)), setting the parameter to false, prevents it from being picked up as the root volume. If both root device hint, and the is_root_volume (set to true) parameters are set, the is_root_volume parameter takes priority, and the volume should be picked up for the root device.

1nv0k32: > ... out-of-tree RAID interfaces may also use it to populate root device hints

I assume all the interfaces inside Ironic should already support it, and in case they are out-of-tree interfaces and they don't support this field, operators themselves should take care of it, even in metal3 :thinking:

dtantsur: It's the other way around: no in-tree hardware interfaces support it. And Metal3 does not support out-of-tree RAID interfaces out-of-box.

1nv0k32: I see, good to know :D
Removed the field from hardware raid.

lentzi90:
Sorry, I am not much familiar with this topic and I am swamped with reviews. Hopefully someone else can review this.

pkg/provisioner/ironic/raid.go
dtantsur: nit: this message needs an update now

internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -166,8 +166,9 @@ func validateBMCAccess(host *metal3api.BareMetalHost, bmcAccess bmc.AccessDetail
 	return errs
 }
 
-func validateRAID(r *metal3api.RAIDConfig) []error {
+func validateRAID(host *metal3api.BareMetalHost) []error {
```
tuminoid, line 169: Per bmh types.go, enforcing RAID-1 for reducing risk of non-booting is a thing, while this allows yout to set any raid device, including RAID-0, that violates that policy.

1. should we enforce it here
2. should we not care, and relax the wording in types.go?

pkg/provisioner/ironic/raid.go
```diff
@@ -36,12 +36,13 @@ func setTargetRAIDCfg(ctx context.Context, p *ironicProvisioner, raidInterface s
 		return provisioner.Result{}, nil
 	}
 
-	// set root volume
-	if data.RootDeviceHints == nil {
+	// set root volume in case there is no rootDeviceHint or RAID root volume
+	rootCount := data.TargetRAIDConfig.GetRootVolumeCount()
+	if data.RootDeviceHints == nil && rootCount == 0 {
```
tuminoid, line 42: I think we're missing unit test path for hints == null and rootcount > 0

pkg/provisioner/ironic/raid.go
```diff
@@ -171,9 +172,10 @@ func buildTargetSoftwareRAIDCfg(volumes []metal3api.SoftwareRAIDVolume) (logical
 	for _, volume := range volumes {
 		// Build logicalDisk
 		logicalDisk = nodes.LogicalDisk{
-			SizeGB:     volume.SizeGibibytes,
-			RAIDLevel:  nodes.RAIDLevel(volume.Level),
-			Controller: "software",
+			SizeGB:       volume.SizeGibibytes,
+			RAIDLevel:    nodes.RAIDLevel(volume.Level),
+			Controller:   "software",
+			IsRootVolume: volume.RootVolume,
```
tuminoid: Here we might send now false pointer which is different from previously leaving it undefined  (nil). false will be sent to Ironic while nil would leave field omitted.

Not sure it matters for Ironic, but just possible nuance diff.

tuminoid (commented):
LGTM, just few notes on glancing over it.

/Cc @Rozzii 
for actual LGTM

internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
1nv0k32: I also feel we should enforce it, I added validation to check that.

pkg/provisioner/ironic/raid.go
1nv0k32: Now sending nil in case of false case.

pkg/provisioner/ironic/raid.go
```diff
@@ -161,19 +163,24 @@ func buildTargetSoftwareRAIDCfg(volumes []metal3api.SoftwareRAIDVolume) (logical
...
-		return
+		return logicalDisks, nil
 	}
 
 	if nodes.RAIDLevel(volumes[0].Level) != nodes.RAID1 {
 		return nil, errors.New("the level in first volume of software raid must be RAID1")
 	}
 
 	for _, volume := range volumes {
+		var isRootVolume *bool
```
Rozzii: I am not really found of handling a boolean in a way that it would have a "shadow value", and also complicating things with the pointer conversions. In other words I don't think it make sense to allow "None" as a value.

IMO "isRootVolume" should be set to "false" by default for all volumes, and the one volume the user decides to be root has to be explicitly set to true.

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -404,6 +404,10 @@ type HardwareRAIDVolume struct {
 
 // SoftwareRAIDVolume defines the desired configuration of volume in software RAID.
 type SoftwareRAIDVolume struct {
+	// Setting this to true, causes installer to consider this volume as root device for installation.
+	// This can only be set for one of the volumes.
+	RootVolume *bool `json:"rootVolume,omitempty"`
```
Rozzii: I would like this to default to false for all volumes and could be set to true for 1 volume.

I would like to avoid handling this as a pointer, it is not an object, list, or integer, "None" as a potential value makes no sense for me here, kind of defeats the purpose of a boolean.

Rozzii (commented):
Thanks for the contribution.
The feature makes sense and it was long time missing, but I would like you to please handle the new spec value as a true boolean and not as a pointer.

Rozzii (commented):
/lgtm

Issue #3261, opened 2026-05-19
HFS reports all values as strings regardless of their type


dtantsur:
In GopherCloud, `BIOSSetting.Value` is a string, so numbers get stringified and booleans get converted to strings "1" or "0". We haven't noticed it so far because most values are Enumerations (i.e. strings) and because sushy-tools have been converting everything to strings historically.

I don't know how to fix it. GopherCloud alone will require preserving backward compatibility, but our public HostFirmwareSettings API also assumes strings.

/kind bug

pierrecregut:
May be we can use HostFirmwareSchema at the level of BMO to hide the fact that gophercloud view everything as strings. 
* When we read a setting, the type helps to interpret a '1' as 'true' for example or even true if we change the HFSettings custom resource. That's a hack but at least we do not to "break" gophercloud behaviour.
* In the other direction, we could update gophercloud (or sushy) to understand typed value (eg weird string syntax 'bool#settingName' for the field name  or 'bool#true'). This should not break anything as it would be a new feature.

PR #3262, opened 2026-05-19
🌱 e2e: avoid non-string values in the HFS tests

dtantsur:
Boolean values are particularly broken (see #3261), and QuietBoot is
a Boolean. It only happened to work because sushy-tools used to turn
every value into a string.

Use a string value until we can figure out what to do.


iurygregory (commented):
/lgtm

PR #3263, opened 2026-05-19
🌱 e2e: make sushy-tools image customizable

dtantsur:
Although we don't update sushy-tools very often, it will be good to have
CI coverage for these updates.

Here on the BMO side this requires an ability to use a custom image.


tuminoid (commented):
/lgtm

Rozzii (commented):
/approve

PR #3265, opened 2026-05-20
🐛 Fix preprovisioning network Secret lifecycle during BMH deletion

MahnoorAsghar:
Apply the same finalizer on the Secret referenced by preprovisioningNetworkDataName, as the finalizer used for BMC credentials, and remove it when the host finishes deleting. Treat a missing Secret as non-fatal during host deletion to avoid RegistrationError noise when namespaces are torn down concurrently.

**Checklist:**

### DIFF
--- a/internal/controller/metal3.io/baremetalhost_controller.go
+++ b/internal/controller/metal3.io/baremetalhost_controller.go
@@ -39,6 +39,7 @@ import (
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
 	"sigs.k8s.io/cluster-api/util/conditions"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -73,12 +74,13 @@ type BareMetalHostReconciler struct {
 // Instead of passing a zillion arguments to the action of a phase,
 // hold them in a struct.
 type reconcileInfo struct {
-	log               logr.Logger
-	host              *metal3api.BareMetalHost
-	request           ctrl.Request
-	bmcCredsSecret    *corev1.Secret
-	events            []corev1.Event
-	postSaveCallbacks []func()
+	log                              logr.Logger
+	host                             *metal3api.BareMetalHost
+	request                          ctrl.Request
+	bmcCredsSecret                   *corev1.Secret
+	preprovisioningNetworkDataSecret *corev1.Secret
+	events                           []corev1.Event
+	postSaveCallbacks                []func()
 }
 
 // match the provisioner.EventPublisher interface.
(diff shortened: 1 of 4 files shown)


internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -207,12 +209,27 @@ func (r *BareMetalHostReconciler) Reconcile(ctx context.Context, request ctrl.Re
 		}
 	}
 
+	var preprovisioningNetworkDataSecret *corev1.Secret
+	if host.Spec.PreprovisioningNetworkDataName != "" &&
+		host.Status.Provisioning.State != metal3api.StateNone &&
+		host.Status.Provisioning.State != metal3api.StateUnmanaged {
+		preprovisioningNetworkDataSecret, err = r.acquirePreprovisioningNetworkDataSecret(ctx, host)
+		if err != nil {
+			if !host.DeletionTimestamp.IsZero() {
```
MahnoorAsghar: Should I use host_config_data.go's hostInDeletionFlow() here?
Should I not attempt to acquire the secret if the host is in the deletion flow?

internal/controller/metal3.io/host_config_data.go
```diff
@@ -107,18 +124,24 @@ func (hcd *hostConfigData) PreprovisioningNetworkData(ctx context.Context) (stri
 	if hcd.host.Spec.PreprovisioningNetworkDataName == "" {
 		return "", nil
 	}
+	addFinalizer := hcd.host.Status.Provisioning.State != metal3api.StateDeleting
```
MahnoorAsghar: Same question here: should I use hostInDeletionFlow() here?

lentzi90:
/cc
Just so I don't forget about this. I'll try to review next week

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -207,12 +209,27 @@ func (r *BareMetalHostReconciler) Reconcile(ctx context.Context, request ctrl.Re
...
+				preprovisioningNetworkDataSecret = &corev1.Secret{}
+			} else {
+				reqLogger.Info("failed to acquire preprovisioning network data secret", "error", err)
+			}
+		}
```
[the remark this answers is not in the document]
MahnoorAsghar: This looks good to me, revising commit

internal/controller/metal3.io/host_config_data.go
```diff
@@ -23,7 +38,7 @@ type hostConfigData struct {
 // Generic method for data extraction from a Secret. Function uses dataKey
 // parameter to detirmine which data to return in case secret contins multiple
 // keys.
-func (hcd *hostConfigData) getSecretData(ctx context.Context, name, namespace, dataKey string) (string, error) {
+func (hcd *hostConfigData) getSecretData(ctx context.Context, name, namespace, dataKey string, addFinalizer bool) (string, error) {
```
lentzi90, line 41: The semantics here are very surprising to me. Based on the name and description I would never have guessed that this method modifies the secret in any way. However, I see that it already did modification before this PR, so you are not changing it in that sense. For this reason I am ok with it. I still want to highlight that this is something that we should refactor to clarify the behavior.

lentzi90 (commented):
Looks good. I think we should treat the BMC credentials the same way. It would make sense to have one common way of doing this for all the secrets we reference. But that needs to be discussed separately...

To answer your questions, yes I think it would make sense to use hostInDeletionFlow in both places. This avoids small differences in how we determine the state and makes updates to the behavior easier since it applies to all places automatically.

internal/controller/metal3.io/host_config_data.go
MahnoorAsghar: Done

internal/controller/metal3.io/host_config_data.go
MahnoorAsghar: I agree, I was surprised by this as well

MahnoorAsghar:
I've updated the PR to use the hostInDeletionFlow() function, and re-tested it locally. Ready for an lgtm now :)

lentzi90 (commented):
/lgtm

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -2425,6 +2451,22 @@ func (r *BareMetalHostReconciler) getBMCSecretAndSetOwner(ctx context.Context, r
...
+func (r *BareMetalHostReconciler) acquirePreprovisioningNetworkDataSecret(ctx context.Context, host *metal3api.BareMetalHost) (*corev1.Secret, error) {
+	secretManager := r.secretManager(ctx, r.Log.WithValues(
+		"baremetalhost", types.NamespacedName{Namespace: host.Namespace, Name: host.Name},
+	))
+	key := types.NamespacedName{
+		Name:      host.Spec.PreprovisioningNetworkDataName,
+		Namespace: host.Namespace,
+	}
+
+	return secretManager.ObtainSecretWithFinalizer(ctx, key, host.Status.Provisioning.State != metal3api.StateDeleting)
```
MahnoorAsghar, line 2467: I just noticed that I didnt use the host_config_data.go's hostInDeletionFlow() function here..

PR #3267, opened 2026-05-21
:bug: Skip webhook for metadata-only BMH updates via matchConditions

elfosardo:
During cluster bootstrap, BMO's validating webhook endpoint may not be ready when the controller tries to add finalizers to BareMetalHost resources. This causes repeated "no endpoints available" or "context deadline exceeded" errors for up to several minutes.

Since the webhook only validates .spec and .metadata.annotations, it fires unnecessarily on metadata-only updates like finalizer additions. Add a CEL matchConditions filter to the ValidatingWebhookConfiguration so the API server skips the webhook call entirely when neither spec nor annotations changed. This is implemented as a kustomize patch to survive controller-gen regeneration of manifests.yaml.

### DIFF
--- a/config/base/webhook/kustomization.yaml
+++ b/config/base/webhook/kustomization.yaml
@@ -4,5 +4,8 @@ resources:
 - manifests.yaml
 - service_patch.yaml
 
+patches:
+- path: matchconditions_patch.yaml
+
 configurations:
 - kustomizeconfig.yaml
--- a/config/base/webhook/matchconditions_patch.yaml
+++ b/config/base/webhook/matchconditions_patch.yaml
@@ -1,0 +1,12 @@
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
+metadata:
+  name: validating-webhook-configuration
+webhooks:
+- name: baremetalhost.metal3.io
+  matchConditions:
+  - name: "spec-or-annotations-changed"
+    expression: >-
+      request.operation == 'CREATE' ||
+      object.spec != oldObject.spec ||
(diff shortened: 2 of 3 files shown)


config/base/webhook/kustomization.yaml
```diff
@@ -4,5 +4,8 @@ resources:
 - manifests.yaml
 - service_patch.yaml
 
+patches:
```
dtantsur, line 7: Patches will only benefit people who use upstream Kustomize manifests. Is it possible to add conditions to the webhook definition in Go code?

elfosardo: right
I'm not sure that's possible, at least to my knowledge.
We could add an early return in ValidateUpdate when nothing relevant changes maybe ?
That would not prevent the calls to the webhook, but the handler will return faster.

config/base/webhook/matchconditions_patch.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
+metadata:
+  name: validating-webhook-configuration
+webhooks:
+- name: baremetalhost.metal3.io
+  matchConditions:
+  - name: "spec-or-annotations-changed"
+    expression: >-
+      request.operation == 'CREATE' ||
```
dtantsur, line 10: It may be an odd OpenShift-specific case that creation happens successfully but finalizers are blocked by a webhook. On a normal deployment, it is creation that will fail.

Maybe we need  to solve it in openshift-installer by pre-adding the finalizer?

PR #3279, opened 2026-05-26
:bug: Correct webhook admissionReview version

### LINKED ISSUE #3266: Invalid API version for AdmissionReview in webhooks
In our webhooks, we specify the AdmissionReview API versions that are supported by the webhooks. Currently we list `v1` and `v1beta` there. See for example https://github.com/metal3-io/baremetal-operator/blob/d41c7b2bfbb8cda7c9dddf43baffdbaa1057add0/internal/webhooks/metal3.io/v1alpha1/baremetalhost_webhook.go#L40
This seems like a typo. There is no `v1beta`, it should be `v1beta1`.

Considering that we have had the typo there for quite some time without anyone noticing, I suggest we simply drop `v1beta`. The alternative is to fix the typo so we have `v1` and `v1beta1`.

Ref docs: https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#webhook-request-and-response
The API versions are mentioned in Kubernetes v1.22 deprecations: https://v1-34.docs.kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-22. I have not been able to find any later notes about them so I think both are still supported. There is probably not much point in keeping `v1beta1` around though.

/kind bug

pierrecregut:
Limit to v1 as v1beta1 is deprecated and not served since v1.22. Do not support v1beta which does not exists.
Adds a nil check in baremetalhost webhook (ValidateCreate).

Fixes: #3266

### DIFF
--- a/config/base/webhook/manifests.yaml
+++ b/config/base/webhook/manifests.yaml
@@ -6,7 +6,6 @@ metadata:
 webhooks:
 - admissionReviewVersions:
   - v1
-  - v1beta
   clientConfig:
     service:
       name: webhook-service
@@ -27,7 +26,6 @@ webhooks:
   sideEffects: None
 - admissionReviewVersions:
   - v1
-  - v1beta
   clientConfig:
     service:
       name: webhook-service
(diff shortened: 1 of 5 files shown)


pierrecregut:
@lentzi90 this is a followup to #3196

lentzi90 (commented):
Thank you!
/approve

PR #3282, opened 2026-05-27
:seedling: Drive e2e setup via vbmctl config template

terror96:
Replace individual vbmctl CLI calls for image server and BMC emulator setup with envsubst-rendered vbmctl.yaml.tmpl. Export variables needed for substitution and move VM/emulator creation to after the config is generated.

Also fix ApplyDefaults to always override StoragePool and LibvirtURI in BMCEmulator config with values from the top-level spec.

We can simplify `hack/ci-e2e.sh` to use a single `vbmctl` command to setup the test environment.

### DIFF
--- a/hack/ci-e2e.sh
+++ b/hack/ci-e2e.sh
@@ -21,9 +21,9 @@ cd "${REPO_ROOT}" || exit 1
 
 BMC_PROTOCOL="${BMC_PROTOCOL:-"redfish-virtualmedia"}"
 if [[ "${BMC_PROTOCOL}" == "redfish" ]] || [[ "${BMC_PROTOCOL}" == "redfish-virtualmedia" ]]; then
-  BMO_E2E_EMULATOR="sushy-tools"
+  export BMO_E2E_EMULATOR="sushy-tools"
 elif [[ "${BMC_PROTOCOL}" == "ipmi" ]]; then
-  BMO_E2E_EMULATOR="vbmc"
+  export BMO_E2E_EMULATOR="vbmc"
 else
   echo "FATAL: Invalid BMC protocol specified: ${BMC_PROTOCOL}"
   exit 1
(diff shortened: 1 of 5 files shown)


hack/ci-e2e.sh
```diff
@@ -130,8 +120,25 @@ if [[ ! -f "${IMAGE_DIR}/${IPA_FILE}" ]]; then
...
+# Need to do some extra setup for the vbmc emulator
+if [[ "${BMO_E2E_EMULATOR}" == "vbmc" ]]; then
+  readarray -t BMCS < <(yq e -o=j -I=0 '.[]' "${E2E_BMCS_CONF_FILE}")
+  for bmc in "${BMCS[@]}"; do
+    address=$(echo "${bmc}" | jq -r '.address')
+    hostName=$(echo "${bmc}" | jq -r '.name')
+    vbmc_port="${address##*:}"
+    "${REPO_ROOT}/tools/bmh_test/vm2vbmc.sh" "${hostName}" "${vbmc_port}" "${IP_ADDRESS}"
+  done
+fi
```
lentzi90, line 141: This we should build into vbmctl! (Can be done separately)

test/e2e/config/vbmctl.yaml.tmpl
```diff
@@ -45,3 +45,12 @@ spec:
...
     ipv6: true
+  bmcEmulator:
+    type: "${BMO_E2E_EMULATOR}"
+    image: "${BMO_E2E_IMAGE}"
+    listenAddress: "${IP_ADDRESS}"
+    listenPort: ${BMO_E2E_LISTEN_PORT}
+  imageServer:
+    port: 80
+    dataDir: "${IMAGE_DIR}"
+    containerName: "vbmctl-image-server-e2e"
```
lentzi90, line 56: What do you think about hard-coding these? We would need separate files for vbmc and sushy-tools then, so there is definitely some duplication. On the other hand, it becomes dead simple to know what the config is.
I am thinking that most of these are in practice never changed. We always use the same IP, port and image dir. Is it then worth it to do the templating dance just for the emulator? I am doubtful, but I am sure there are other opinions 😄

terror96: We can surely eliminate the template circus in case there is no real need for it.

hack/ci-e2e.sh
terror96: Let's make a separate PR for that in order to keep this change cleaner.

test/e2e/config/vbmctl.yaml.tmpl
terror96: I forgot this: commit 350bcce7b0407c79cd93632a66d3ca2d20237578 ... maybe we should just keep the template approach. At least it will support further adjusting...

terror96: /cc @dtantsur

lentzi90: Right, let's go with the templating for now. Maybe at some point we can make it so that the flag can be set as override for the config. Then we can have the config hard coded and still use the flag for overriding in the script.

lentzi90 (commented):
/approve

test/e2e/config/vbmctl.yaml.tmpl
nuhakala: Looks like both bmc emulators use the same IP address ATM, so maybe that can be hardcoded into the config file instead of substitution? Would make the config tiny bit simpler.

nuhakala (commented):
LGTM

test/e2e/config/vbmctl.yaml.tmpl
terror96: True, the reason not to use hard-cording is that `hack/ci-e2e.sh` defines it, and the value could easily be changed with a single edit.

nuhakala: Oh yeah, also, if the IP address is used also elsewhere than vbmctl, then we should keep it in `hack/ci-e2e.sh`. But if only vbmctl uses it, then it can also be easily changed in the config.

terror96: Nod, defining and using the IP address in `hack/ci-e2e.sh` is part of the current codebase.

terror96:
> LGTM
>
>Looks like both bmc emulators use the same IP address ATM, so maybe that can be hardcoded into the config file instead of substitution? Would make the config tiny bit simpler.

This question has been answered above.

test/e2e/config/vbmctl.yaml.tmpl
dtantsur: A small note on hardcoded stuff: whatever we hardcode, we better not change it arbitrary in the future because things will need to be similarly hardcoded in Ironic job definitions that use this functionality.

lentzi90 (commented):
/lgtm

PR #3285, opened 2026-05-27
🌱 e2e tests - bios firmware upgrade

iurygregory:
This commit adds e2e test to cover bios firmware update scenarios

Scenario1 - Cleaning
- HFC is created before the BMH and the firmware update happens
during cleaning.

Scenario2 - Cleaning
- HFC is automatically created and we patch the CR so BMO triggers
cleaning again because the BMH was in available.

Scenario3 - Servicing
- We create a HUP for a Provisioned BMH and we update the HFC and
BMH to trigger a firmware update.

Scenario4 - Servicing
- Testing updating firmware on an externally provisioned BMH,
the HFC should be created automaticaly and the HUP manualy created.

### DIFF
--- a/test/e2e/firmware_components_test.go
+++ b/test/e2e/firmware_components_test.go
@@ -1,0 +1,566 @@
+//go:build e2e
+// +build e2e
+
+package e2e
+
+import (
+	"context"
+	"path"
+
+	metal3api "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/cluster-api/test/framework"
+	"sigs.k8s.io/cluster-api/util/patch"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	// The firmware update URL does not need to point to a real binary since
+	// sushy-tools emulates the upgrade by incrementing the BIOS version
+	// without downloading anything.
+	biosFirmwareUpdateURL = "http://192.168.222.1/bios-update.bin"
+)
+
+func waitForBIOSFirmwareUpdate(ctx context.Context, cl client.Client, hfc *metal3api.HostFirmwareComponents, name, namespace string, intervals ...interface{}) {
+	Eventually(func(g Gomega) {
+		g.Expect(cl.Get(ctx, types.NamespacedName{
+			Name:      name,
+			Namespace: namespace,
+		}, hfc)).To(Succeed())
+		var biosComponent *metal3api.FirmwareComponentStatus
+		for i := range hfc.Status.Components {
+			if hfc.Status.Components[i].Component == "bios" {
+				biosComponent = &hfc.Status.Components[i]
+				break
+			}
+		}
+		g.Expect(biosComponent).NotTo(BeNil(), "bios component should exist in status.components")
+		g.Expect(biosComponent.InitialVersion).NotTo(BeEmpty())
+		g.Expect(biosComponent.CurrentVersion).NotTo(BeEmpty())
+		g.Expect(biosComponent.CurrentVersion).NotTo(Equal(biosComponent.InitialVersion),
+			"CurrentVersion should differ from InitialVersion after firmware update")
+	}, intervals...).Should(Succeed())
+}
+
+var _ = Describe("Host Firmware Components", Label("required", "firmware-components"), func() {
+	var (
+		specName      = "firmware-components"
+		namespace     *corev1.Namespace
+		cancelWatches context.CancelFunc
+		toCleanup     []client.Object
+	)
+
+	BeforeEach(func() {
+		if !e2eConfig.GetBoolVariable("DEPLOY_IRONIC") || bmc.AccessDetails.Driver() != "redfish" {
+			Skip("Firmware component tests require a real Ironic and a host with Redfish")
+		}
+
+		toCleanup = nil
+		namespace, cancelWatches = framework.CreateNamespaceAndWatchEvents(ctx, framework.CreateNamespaceAndWatchEventsInput{
+			Creator:             clusterProxy.GetClient(),
+			ClientSet:           clusterProxy.GetClientSet(),
+			Name:                specName,
+			LogFolder:           artifactFolder,
+			IgnoreAlreadyExists: true,
+		})
+	})
+
+	It("should apply BIOS firmware update during cleaning on a BMH", func() {
+		bmhName := specName + "-before-bmh"
+		secretName := bmhName + "-bmc"
+
+		By("Creating a secret with BMH credentials")
+		bmcCredentialsData := map[string]string{
+			"username": bmc.User,
+			"password": bmc.Password,
+		}
+		secret := CreateSecret(ctx, clusterProxy.GetClient(), namespace.Name, secretName, bmcCredentialsData)
+		toCleanup = append(toCleanup, secret)
+
+		By("Creating a HostFirmwareComponents with a BIOS update before creating the BMH")
+		hfc := &metal3api.HostFirmwareComponents{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      bmhName,
+				Namespace: namespace.Name,
+			},
+			Spec: metal3api.HostFirmwareComponentsSpec{
+				Updates: []metal3api.FirmwareUpdate{
+					{
+						Component: "bios",
+						URL:       biosFirmwareUpdateURL,
+					},
+				},
+			},
+		}
+		Expect(clusterProxy.GetClient().Create(ctx, hfc)).To(Succeed())
+
+		By("Creating a BMH with inspection and cleaning disabled")
+		bmh := metal3api.BareMetalHost{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      bmhName,
+				Namespace: namespace.Name,
+			},
+			Spec: metal3api.BareMetalHostSpec{
+				Online: true,
+				BMC: metal3api.BMCDetails{
+					Address:                        bmc.Address,
+					CredentialsName:                secretName,
+					DisableCertificateVerification: bmc.DisableCertificateVerification,
+				},
+				BootMode:              metal3api.BootMode(e2eConfig.GetVariable("BOOT_MODE")),
+				BootMACAddress:        bmc.BootMacAddress,
+				AutomatedCleaningMode: metal3api.CleaningModeDisabled,
+				InspectionMode:        metal3api.InspectionModeDisabled,
+			},
+		}
+		Expect(clusterProxy.GetClient().Create(ctx, &bmh)).To(Succeed())
+		toCleanup = append(toCleanup, &bmh)
+
+		By("Waiting for the BMH to start preparing")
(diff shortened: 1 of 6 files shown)


test/e2e/firmware_components_test.go
```diff
@@ -0,0 +1,282 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/cluster-api/test/framework"
+	"sigs.k8s.io/cluster-api/util/patch"
+)
+
+const (
+	// The firmware update URL does not need to point to a real binary since
+	// sushy-tools emulates the upgrade by incrementing the BIOS version
+	// without downloading anything.
```
dtantsur, line 26: Are you sure Ironic does not cache it?

test/e2e/firmware_components_test.go
```diff
@@ -0,0 +1,282 @@
...
+			Namespace: namespace.Name,
+		}, e2eConfig.GetIntervals(specName, "wait-bmh-deleted")...)
+	})
+
+	AfterEach(func() {
+		CollectSerialLogs(bmc.Name, path.Join(artifactFolder, specName))
+		DumpResources(ctx, e2eConfig, clusterProxy, path.Join(artifactFolder, specName))
+		if !skipCleanup {
+			isNamespaced := e2eConfig.GetBoolVariable("NAMESPACE_SCOPED")
+			Cleanup(ctx, clusterProxy, namespace, cancelWatches, isNamespaced, e2eConfig.GetIntervals("default", "wait-namespace-deleted")...)
```
dtantsur: I hope https://github.com/metal3-io/baremetal-operator/pull/3245 merges soon, and we can update this with a saner logic.

test/e2e/firmware_components_test.go
```diff
@@ -0,0 +1,282 @@
...
+		namespace, cancelWatches = framework.CreateNamespaceAndWatchEvents(ctx, framework.CreateNamespaceAndWatchEventsInput{
+			Creator:             clusterProxy.GetClient(),
+			ClientSet:           clusterProxy.GetClientSet(),
+			Name:                specName,
+			LogFolder:           artifactFolder,
+			IgnoreAlreadyExists: true,
+		})
+	})
+
+	It("should upgrade BIOS firmware via cleaning when HostFirmwareComponents is updated on an available host", func() {
```
dtantsur: Check my HFS test: upgrading via cleaning can be done together with the other test. I'd rather see the us cover servicing as well.

test/e2e/firmware_components_test.go
iurygregory: will re-run some tests, my previous tests was creating a random txt file in a a webserver and use that URL...
Maybe we can do something similar, accessing the httpd creating a txt file and giving this URL...

test/e2e/firmware_components_test.go
iurygregory: I can totally wait for the other PR

test/e2e/firmware_components_test.go
iurygregory: Yeah, I'm planning on cover servicing, but I thought it should be in a separate PR, if people are ok I can add in the same.

test/e2e/firmware_components_test.go
dtantsur: It has merged, so please rebase and update

iurygregory: Tks!

test/e2e/firmware_components_test.go
iurygregory: We can re-use any url for this, I've tested with `http://127.0.0.1`  for example and worked multiple times with bifrost

test/e2e/firmware_components_test.go
iurygregory: I've tested servicing in bifrost and for some weird reason the virtual baremetal didn't power off, so I'will update this to cover the cleaning scenario and investigate the servicing issue I've found

dtantsur: The cleaning test can be merged into the second test (with pre-created HFC). I don't quite like having inconsistent HFS and HFC tests given how similar the features are.

I'd probably even add the servicing test and see how it fails here. Maybe even merge with a workaround, if one is possible, and file a bug to fix it.

iurygregory: Ok, will do that!

iurygregory:
The error on CI doesn't seem realted .-.
```
  [FAIL] Inspection [It] should inspect a newly created BMH [required, inspection]
  /home/ubuntu/_work/baremetal-operator/baremetal-operator/test/e2e/common.go:649
 ```
 @dtantsur this should be ready to review, I'm wondering if it's a CI issue, since the firmware tests passed based on the e2e logs.

File: test/e2e/firmware_components_test.go [hunk 4]
```diff
@@ -0,0 +1,669 @@
...
+			for i := range hfc.Status.Components {
+				if hfc.Status.Components[i].Component == "bios" {
+					biosComponent = &hfc.Status.Components[i]
+					break
+				}
+			}
+			g.Expect(biosComponent).NotTo(BeNil(), "bios component should exist in status.components")
+			g.Expect(biosComponent.InitialVersion).NotTo(BeEmpty())
+			g.Expect(biosComponent.CurrentVersion).NotTo(BeEmpty())
+			Logf("BIOS InitialVersion: %s, CurrentVersion: %s, LastVersionFlashed: %s",
```
dtantsur: These messages are pretty verbose, I'd rather not log them

test/e2e/firmware_components_test.go
```diff
@@ -0,0 +1,669 @@
...
+		By("Making sure HFC was deleted too")
+		Eventually(func() bool {
+			return k8serrors.IsNotFound(clusterProxy.GetClient().Get(ctx, types.NamespacedName{
+				Name:      bmhName,
+				Namespace: namespace.Name,
+			}, hfc))
+		}, e2eConfig.GetIntervals(specName, "wait-bmh-deleted")...).Should(BeTrue())
+
+		By("Creating a secret with BMH credentials again")
+		CreateSecret(ctx, clusterProxy.GetClient(), namespace.Name, secretName, bmcCredentialsData)
```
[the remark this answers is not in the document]
iurygregory: @dtantsur thoughts on this one? I was following the same approach we have in https://github.com/metal3-io/baremetal-operator/blob/main/test/e2e/firmware_settings_test.go#L162 , but I do think it's make sense

test/e2e/firmware_components_test.go
iurygregory: done, removed, will push on the next update

test/e2e/firmware_components_test.go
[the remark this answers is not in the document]
dtantsur: If all secrets are added to `toCleanup`, we don't need to assume that they're present.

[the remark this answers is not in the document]
iurygregory: ok, updated

test/e2e/firmware_components_test.go
```diff
@@ -0,0 +1,662 @@
...
+		if e2eConfig.GetVariable("SSH_CHECK_PROVISIONED") == "true" {
+			userDataSecretName := "user-data"
+			sshPubKeyPath := e2eConfig.GetVariable("SSH_PUB_KEY")
+			createSSHSetupUserdata(ctx, clusterProxy.GetClient(), namespace.Name, userDataSecretName, sshPubKeyPath, bmc.IPAddress)
+			userDataSecret = &corev1.SecretReference{
+				Name:      userDataSecretName,
+				Namespace: namespace.Name,
+			}
+		}
+		Expect(PatchBMHForProvisioning(ctx, PatchBMHForProvisioningInput{
```
MahnoorAsghar, line 280: Why are we patching the BMH with the userDataSecret, instead of adding its ref on BMH creation?

test/e2e/firmware_components_test.go
```diff
@@ -0,0 +1,662 @@
...
+		namespace, cancelWatches = framework.CreateNamespaceAndWatchEvents(ctx, framework.CreateNamespaceAndWatchEventsInput{
+			Creator:             clusterProxy.GetClient(),
+			ClientSet:           clusterProxy.GetClientSet(),
+			Name:                specName,
+			LogFolder:           artifactFolder,
+			IgnoreAlreadyExists: true,
+		})
+	})
+
+	It("should apply BIOS firmware update during cleaning on a BMH", func() {
```
MahnoorAsghar, line 74: Sorry for a potentially noob question, but cleaning is disabled on the BMH, no?

MahnoorAsghar:
Looks alright to me, other than the 2 questions :)

test/e2e/firmware_components_test.go
dtantsur: I thought about commenting on this: we use "cleaning" to mean two different things here. Automated cleaning is disabled, HFS uses manual cleaning.

MahnoorAsghar: I see

test/e2e/firmware_components_test.go
iurygregory: In case we have `SSH_CHECK_PROVISIONED` configure we need to update the configuration of the secret, this is also the pattern we have in other e2e tests.

iurygregory:
@dtantsur anything else you want me to update here?

test/e2e/firmware_components_test.go
```diff
@@ -0,0 +1,662 @@
...
+				}
+			}
+			g.Expect(biosComponent).NotTo(BeNil(), "bios component should exist in status.components")
+			g.Expect(biosComponent.InitialVersion).NotTo(BeEmpty())
+			g.Expect(biosComponent.CurrentVersion).NotTo(BeEmpty())
+			g.Expect(biosComponent.CurrentVersion).NotTo(Equal(biosComponent.InitialVersion),
+				"CurrentVersion should differ from InitialVersion after firmware update")
+		}, e2eConfig.GetIntervals(specName, "wait-firmware-components")...).Should(Succeed())
+
+		By("Deleting the BMH to get rid of cached values")
```
dtantsur: I'm not sure I understand the purpose of deleting the BMH. It's going to make the already long test somewhat longer.

test/e2e/firmware_components_test.go
```diff
@@ -0,0 +1,662 @@
...
+		}, hfc)).To(Succeed())
+		Expect(hfc.Status.Updates).To(HaveLen(1))
+		Expect(hfc.Status.Updates[0].Component).To(Equal("bios"))
+		Expect(hfc.Status.Updates[0].URL).To(Equal(biosFirmwareUpdateURL))
+		Expect(hfc.Status.Conditions).To(ContainCondition(string(metal3api.HostFirmwareComponentsChangeDetected), metav1.ConditionFalse))
+		Expect(hfc.Status.Conditions).To(ContainCondition(string(metal3api.HostFirmwareComponentsValid), metav1.ConditionTrue))
+
+		By("Waiting for the HFC controller to update BIOS component status from Ironic")
+		Eventually(func(g Gomega) {
+			g.Expect(clusterProxy.GetClient().Get(ctx, types.NamespacedName{
```
dtantsur: This pattern is repeated 3 times, let's extract it into a function

test/e2e/firmware_components_test.go
iurygregory: So, my idea is that we make sure things associated with the BMH gets deleted (HFC, HUP...)
I can remove this without problems if you don't think it's worth it

test/e2e/firmware_components_test.go
iurygregory: ok, will work on it.

test/e2e/firmware_components_test.go
dtantsur: You can check it at the end of the test too

iurygregory: Updated, will mark as resolved

iurygregory:
Conflicts solved

PR #3288, opened 2026-05-28
🌱 e2e: servicing of externally provisioned hosts

dtantsur:


dtantsur:
/retest

See https://github.com/metal3-io/baremetal-operator/pull/3289 for the cause

PR #3310, opened 2026-06-04
:bug: Restore delete permission on secrets

### LINKED ISSUE #3304: Regression: The removal of  DELETE RBAC combined with K8s --enable-admission-plugins=NodeRestriction breaks BMC secret deletion
This change removes delete permission on secrets, but the BMO BareMetalHost controller still updates the BMC credentials Secret to add an [ownerReference](https://github.com/metal3-io/baremetal-operator/blame/main/pkg/secretutils/secret_manager.go#L95). Kubernetes rejects that update without delete on the Secret, and we’re seeing:

_an unhandled failure occurred with the BMC secret: failed to update secret edpm-compute-baremetal-00-bmc-secret in namespace openstack: secrets "edpm-compute-baremetal-00-bmc-secret" is forbidden: cannot set an ownerRef on a resource you can't delete_

So this looks like a regression.

_Originally posted by @rabi in https://github.com/metal3-io/baremetal-operator/pull/3113#discussion_r3347147878_

rabi:
Restore delete permission on secrets because owner reference updates are rejected by the API server without it.

Fixes: #3304

### DIFF
--- a/config/base/rbac/role.yaml
+++ b/config/base/rbac/role.yaml
@@ -23,6 +23,7 @@ rules:
   resources:
   - secrets
   verbs:
+  - delete
   - get
   - list
   - update
--- a/config/render/capm3.yaml
+++ b/config/render/capm3.yaml
@@ -2954,6 +2954,7 @@ rules:
   resources:
   - secrets
   verbs:
+  - delete
   - get
   - list
   - update
--- a/internal/controller/metal3.io/baremetalhost_controller.go
+++ b/internal/controller/metal3.io/baremetalhost_controller.go
@@ -92,7 +92,7 @@ func (info *reconcileInfo) publishEvent(reason, message string) {
 // +kubebuilder:rbac:groups=metal3.io,resources=preprovisioningimages,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=metal3.io,resources=hardwaredata,verbs=get;list;watch;create;delete;patch;update
 // +kubebuilder:rbac:groups=metal3.io,resources=hardware/finalizers,verbs=update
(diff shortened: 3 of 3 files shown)


lentzi90 (commented):
This is missing the changes to the RBAC used in e2e tests from the original PR.
I would also still want a minimal reproducer of the issue. Something is very wrong here. Either our tests or config or something else we are completely missing.

/ok-to-test
/hold

lentzi90:
Commenting here also to spread the word. See the issue for my full comment.
TL;DR: The reason we do not see the issue in CI is that it requires an admission plugin to be enabled on the Kubernetes API server. It is not enabled by default. We don't enabled it in CI and this is why we don't see the issue.

Rozzii:
/hold cancel
Based on the related discussion on the related issue, even though  allowing secret deletion in the current fashion is a reliability/security issue in multi tenant environmetns, the fact remains,  removing the problematic RBAC roles without the "correct secret handling (no secret handling)" logic does indeed induces a regression. 

Maintainers seem to be mostly aligned on the proper "fix" but we have to provide a proper (2 release) deprecation window for users to transition to the new system, we can't break compatibility with the upstream --enable-admission-plugins=NodeRestriction in favor of removing the RBAC role IMO.

/hold cancel
/approve

adilGhaffarDev:
> Maintainers seem to be mostly aligned on the proper "fix" but we have to provide a proper (2 release) deprecation window for users to transition to the new system, we can't break compatibility with the upstream --enable-admission-plugins=NodeRestriction in favor of removing the RBAC role IMO.

I agree.
/lgtm
I would suggest that we either keep this open https://github.com/metal3-io/baremetal-operator/issues/3304 or create new issue to track future work.

PR #3311, opened 2026-06-04
:bug: Fix webhook state guard status checks

nuhakala:

The webhook allows changing the BMC address only when the BMH is in detached or registering state. However, it reads that state from the incoming/updated BMH. So the user can simply set the status to the correct value to bypass the check. This PR changes the webhook so that it checks the status of the old object and prevents bypassing the check.

**Checklist:**

### DIFF
--- a/internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
+++ b/internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
@@ -104,7 +104,9 @@ func (webhook *BareMetalHost) validateChanges(oldObj *metal3api.BareMetalHost, n
 
 	if oldObj.Spec.BMC.Address != "" &&
 		newObj.Spec.BMC.Address != oldObj.Spec.BMC.Address &&
+		oldObj.Status.OperationalStatus != metal3api.OperationalStatusDetached &&
 		newObj.Status.OperationalStatus != metal3api.OperationalStatusDetached &&
+		oldObj.Status.Provisioning.State != metal3api.StateRegistering &&
 		newObj.Status.Provisioning.State != metal3api.StateRegistering {
 		errs = append(errs, errors.New("BMC address can not be changed if the BMH is not in the Registering state, or if the BMH is not detached"))
 	}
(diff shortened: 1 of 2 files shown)


lentzi90 (commented):
I was first thinking that should we remove the "duplicate" checks for the `newObj`, but perhaps it makes sense that you don't change the status at the same time as the BMC address anyway.
/lgtm

nuhakala:
> I was first thinking that should we remove the "duplicate" checks for the `newObj`, but perhaps it makes sense that you don't change the status at the same time as the BMC address anyway. /lgtm

I actually first did it like that, but then added checks for the new object as well. I don't know what would be the most logical, but we had the checks for the new object already so this only adds more checks instead of modifying them.

lentzi90:
Yeah I was thinking that the check on the `newObj` was just a mistake, that it should always have been the `oldObj`. But as said, perhaps it is best to just check both.

smoshiur1237 (commented):
/lgtm

lentzi90 (commented):
Switching my lgtm to approval. Adding hold to make sure we have a few more eyes on this.
/approve
/hold

internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -115,10 +117,11 @@ func (webhook *BareMetalHost) validateChanges(oldObj *metal3api.BareMetalHost, n
 
 	// Only allow enabling externallyProvisioned from Available state.
 	if !oldObj.Spec.ExternallyProvisioned && newObj.Spec.ExternallyProvisioned &&
+		oldObj.Status.Provisioning.State != metal3api.StateAvailable &&
```
MahnoorAsghar, line 120: Shouldn't L121 be removed? Because oldObj should have Status.Provisioning.State as Available, and we don't care about the newObj's state?)

nuhakala: I would imagine that there is some reason why it has been like this. It would make sense that the new object is not available if we set that it is externally provisioned. It would not be provisioned if it was available.

Perhaps @tuminoid or @lentzi90 have better idea.

lentzi90: I can see reasons for both. It makes sense to me that we do not care about newObj, and it was probably just a mistake that it was there from the beginning. On the other hand, I think it would be strange to change the state from Available to something else at the same time as ExternallyProvisioned. So I am not sure what to think. At least it seems to work fine in the e2e tests this way.

nuhakala:
This has been on hold for some time now. Could perhaps @kashifest or @Rozzii check this also?

lentzi90:
I'd say if we don't get more reviews today/tomorrow, let's remove the hold. We will find out if there are issues after merging then.

kashifest (commented):
lgtm
/hold cancel

PR #3312, opened 2026-06-04
:bug: Fix network creation error check in vbmctl

nuhakala:

Fixes the error check in vbmctl when creating a network. Previously, the function would not detect the existing network correctly and it would return empty ID without any error even though it should have returned an error indicating that the network existed.

**Checklist:**


smoshiur1237 (commented):
/lgtm

PR #3327, opened 2026-06-08
⚠️ Provisioner: drop the separate TryInit call

dtantsur:
All of our production code calls TryInit right after NewProvisioner,
so there is little point for them to be separate. A new ErrNotReady
error indicates what returning ready=false used to indicate.

This simplification is needed for my follow-up changes to improve
caching of Ironic parameters between reconciliations.

Using ⚠️ to highlight the change in the Provisioner interface.

Generated-By: Claude Code (commercial)


dtantsur:
/retest

Race in the automated cleaning test?

internal/controller/metal3.io/baremetalhost_controller.go
tuminoid, line 1: L61 NotReady string is not used anymore.

tuminoid (commented):
One nit.

/lgtm

PR #3328, opened 2026-06-08
🌱 e2e: wait a bit for /mnt/data to appear

dtantsur:
First boot scripts are not instant. It's possible to catch the system
at the stage where SSH already works but the script hasn't finished yet.
I suspect I've seen one such case in the CI.


Sunnatillo (commented):
/lgtm

PR #3329, opened 2026-06-08
🐛 Avoid calling getNode in GetFirmwareSettings

dtantsur:
It's pointless since we only use node UUID, which is already known.

Fix the unit test to check what it actually claims to check.

### DIFF
--- a/pkg/provisioner/ironic/bios_test.go
+++ b/pkg/provisioner/ironic/bios_test.go
@@ -22,6 +22,7 @@ func TestGetFirmwareSettings(t *testing.T) {
 
 	cases := []struct {
 		name                string
+		nodeUUID            string
 		expectedSettingsMap metal3api.SettingsMap
 		expectedSchemaMap   map[string]metal3api.SettingSchema
 		includeSchema       bool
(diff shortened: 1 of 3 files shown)


jacob-anders:
Thank you for the PR, Dmitry. It's a good change, LTGM. One minor thing that came up during Claude-assisted code review - sharing for your consideration:

Nit: since GetFirmwareSettings no longer calls getNode(), the AddDefaultResponseJSON for the node endpoint in the new NoBIOS (line ~470) is never hit — it's dead code from the start. The same applies to the existing node responses in BIOSSettings (line 460) and BIOSDetailSettings (same pattern) which also become dead after this change. Not a blocker, but you could either drop those lines or leave a short comment noting they're there for other potential callers of the mock helper.

dtantsur:
> The same applies to the existing node responses in BIOSSettings (line 460) and BIOSDetailSettings (same pattern) which also become dead after this change. Not a blocker, but you could either drop those lines or leave a short comment noting they're there for other potential callers of the mock helper.

It's a sanity thing. I don't want anyone to reuse these helpers in the future and get very unhelpful error messages (like "EOF")

iurygregory (commented):
/lgtm

PR #3330, opened 2026-06-08
🐛 Avoid calling getNode in GetFirmwareComponents

dtantsur:
The only actual purpose of the call there is to see if firmware
components are supported by the Node. Which, in turn, is needed
to determine if the HFC object needs to be created by the BMH
controller.

This commit changes the behavior slightly: now HFC is not created when
the list of components is empty. This happens both when firmware
settings are not supported and during early stages of enrollment.

Most users will only see that the HFC resource may get created slightly
later than before.

### DIFF
--- a/internal/controller/metal3.io/baremetalhost_controller.go
+++ b/internal/controller/metal3.io/baremetalhost_controller.go
@@ -939,21 +939,24 @@ func (r *BareMetalHostReconciler) registerHost(ctx context.Context, prov provisi
 		return actionUpdate{}
 	}
 
-	// Check if the host can support firmware components before creating the resource
-	_, errGetFirmwareComponents := prov.GetFirmwareComponents(ctx)
-	supportsFirmwareComponents := !errors.Is(errGetFirmwareComponents, provisioner.ErrFirmwareUpdateUnsupported)
-
 	// Create the hostFirmwareSettings resource with same host name/namespace if it doesn't exist
 	// Create the hostFirmwareComponents resource with same host name/namespace if it doesn't exist
 	if info.host.Name != "" {
 		if !info.host.DeletionTimestamp.IsZero() {
 			info.log.Info("will not attempt to create new hostFirmwareSettings and hostFirmwareComponents in " + info.host.Namespace)
 		} else {
+			// Check if the host can support firmware components before creating the resource
+			firmwareComponents, errGetFirmwareComponents := prov.GetFirmwareComponents(ctx)
+			if errGetFirmwareComponents != nil {
+				info.log.V(1).Error(errGetFirmwareComponents, "failed to retrieve firmware components; deferring HostFirmwareComponents creation")
+				firmwareComponents = nil
+			}
+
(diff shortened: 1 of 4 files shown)


internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -939,21 +939,20 @@ func (r *BareMetalHostReconciler) registerHost(ctx context.Context, prov provisi
...
+			// Check if the host can support firmware components before creating the resource
+			firmwareComponents, _ := prov.GetFirmwareComponents(ctx)
+
 			if err = r.createHostFirmwareSettings(ctx, info); err != nil {
 				info.log.Info("failed creating hostfirmwaresettings")
 				return actionError{fmt.Errorf("failed to create or update hostFirmwareSettings: %w", err)}
 			}
-			if supportsFirmwareComponents {
+			if len(firmwareComponents) > 0 {
 				if err = r.createHostFirmwareComponents(ctx, info); err != nil {
```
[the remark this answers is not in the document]
jacob-anders: nit: PR description still has HFS (it's fixed in the latest code) - worth adjusting

jacob-anders:
Looks good, thank you Dmitry. 

nit: I left the comment ( https://github.com/metal3-io/baremetal-operator/pull/3330/changes#r3377314687 ) under the resolved Copilot comment about HFS/HFC swap in commit message - latest commit is correct but PR description has stale  text.

internal/controller/metal3.io/baremetalhost_controller.go
```diff
@@ -939,21 +939,24 @@ func (r *BareMetalHostReconciler) registerHost(ctx context.Context, prov provisi
...
 	// Create the hostFirmwareSettings resource with same host name/namespace if it doesn't exist
 	// Create the hostFirmwareComponents resource with same host name/namespace if it doesn't exist
 	if info.host.Name != "" {
 		if !info.host.DeletionTimestamp.IsZero() {
 			info.log.Info("will not attempt to create new hostFirmwareSettings and hostFirmwareComponents in " + info.host.Namespace)
 		} else {
+			// Check if the host can support firmware components before creating the resource
+			firmwareComponents, errGetFirmwareComponents := prov.GetFirmwareComponents(ctx)
+			if errGetFirmwareComponents != nil {
+				info.log.V(1).Error(errGetFirmwareComponents, "failed to retrieve firmware components; deferring HostFirmwareComponents creation")
```
jacob-anders, line 951: idea: would it be worth logging this at a V0 (Info) for easier debugability or would that be too noisy?

dtantsur: Previously we would ignore this error entirely, so I thought nobody cared all too much.

iurygregory: This info we can check from the ironic logs so this would be ok

iurygregory (commented):
/lgtm

PR #3332, opened 2026-06-09
🌱 E2E: Verify Ironic TLS and basic-auth

lentzi90:

Make sure that basic-auth is enabled and using the expected credentials. Check that TLS is used. This does not yet verify the expected certificate is used.

Fixes #

**Checklist:**

### DIFF
--- a/test/e2e/common.go
+++ b/test/e2e/common.go
@@ -1142,6 +1142,10 @@ func WaitForIronicReady(ctx context.Context, input WaitForIronicInput) {
 	}, input.Intervals...).Should(Succeed())
 
 	Logf("Ironic %q is Ready", input.Name)
+
+	if input.SecurityConfig != nil {
+		VerifyIronicSecurityConfig(ctx, *input.SecurityConfig)
+	}
 }
 
 // WaitForIronicInput bundles the parameters for WaitForIronicReady.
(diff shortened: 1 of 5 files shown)


lentzi90:
/hold
Just checking that we don't break the upgrade tests.

dtantsur:
Does this verification really belong in BMO or rather in IrSO?

lentzi90:
> Does this verification really belong in BMO or rather in IrSO?

Hmm I did not consider it. I started from the kustomizations which didn't use IRSO so it was sort of outside my train of thought. But now that you mention it, perhaps we should have this in IRSO. Or both?
I am open to move it there if you think it would be better.

lentzi90:
@dtantsur I looked into adding a test in IRSO, but it seems we already check this there. See https://github.com/metal3-io/ironic-standalone-operator/blob/651d6a677937e5ce0fdb6a9d6863f95f6355c189/test/suite_test.go#L426-L452.
I would still prefer to have a check here to make sure we are testing with the configuration we think we do. It is more for verifying that we configure IRSO properly than verifying that IRSO does what it should.

lentzi90:
Optional tests have passed
/hold cancel

smoshiur1237 (commented):
/lgtm

PR #3333, opened 2026-06-09
:sparkles: Add webhooks to validate URLs

nuhakala:
- Adds new webhooks for DataImages and HostFirmwareComponents
- Validates URLs in DataImages, HostFirmwareComponents and BMHs
- Adds common function to validate URL in webhooks

Currently the URLs are not validated at all or very little. Adding validation reduces intentional and accidental mistakes.

**Checklist:**

### DIFF
--- a/internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation_test.go
+++ b/internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation_test.go
@@ -580,6 +580,19 @@ func TestValidateCreate(t *testing.T) {
 			oldBMH: nil,
 		},
 		{
+			name: "validImageURLOCI",
+			newBMH: &metal3api.BareMetalHost{
+				TypeMeta:   tm,
+				ObjectMeta: om,
+				Spec: metal3api.BareMetalHostSpec{
+					Image: &metal3api.Image{
+						URL: "oci://example.com/image:latest",
+					},
+				},
+			},
+			oldBMH: nil,
+		},
+		{
 			name: "validImageLiveISO",
 			newBMH: &metal3api.BareMetalHost{
 				TypeMeta:   tm,
@@ -607,6 +620,21 @@ func TestValidateCreate(t *testing.T) {
 			},
 			oldBMH:    nil,
 			wantedErr: "image URL test1 is invalid: parse \"test1\": invalid URI for request",
+		},
+		{
+			name: "invalidImageURLScheme",
+			newBMH: &metal3api.BareMetalHost{
+				TypeMeta:   tm,
+				ObjectMeta: om,
+				Spec: metal3api.BareMetalHostSpec{
+					Image: &metal3api.Image{
+						URL:      "unix://example.com",
+						Checksum: "be254ebfd73e66ca91f6d91f5050aa2ee1ec4813ee65ba472f608ed340cbff09",
+					},
+				},
+			},
+			oldBMH:    nil,
+			wantedErr: "image URL unix://example.com is invalid: invalid scheme in URL, \"unix\" not allowed",
 		},
 		{
 			name: "emptyImageURL",
--- a/internal/webhooks/metal3.io/v1alpha1/common_validation.go
+++ b/internal/webhooks/metal3.io/v1alpha1/common_validation.go
@@ -1,0 +1,38 @@
+/*
+Copyright 2025 The Metal3 Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package webhooks
+
+import (
+	"fmt"
+	"net/url"
+	"slices"
+)
+
+// validateURL validates the given URL. The URL is assumed to come from HTTP
+// request. On success, no error is returned. Empty string also returns an error.
+func validateURL(input string) error {
+	urlObj, err := url.ParseRequestURI(input)
+	if err != nil {
+		return err
+	}
+
+	// Check the URL scheme
+	allowed := []string{"http", "https", "ftp", "gopher", "oci"}
+	if !slices.Contains(allowed, urlObj.Scheme) {
+		return fmt.Errorf("invalid scheme in URL, \"%s\" not allowed", urlObj.Scheme)
+	}
+
+	return nil
+}
--- a/internal/webhooks/metal3.io/v1alpha1/dataimage_validation.go
+++ b/internal/webhooks/metal3.io/v1alpha1/dataimage_validation.go
@@ -1,0 +1,32 @@
+/*
+Copyright 2025 The Metal3 Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package webhooks
+
+import (
+	"fmt"
+
+	metal3api "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
+)
+
+// validateDataImage validates DataImage resource for creation and updating.
+func (webhook *DataImage) validateDataImage(img *metal3api.DataImage) []error {
+	var errs []error
+
+	if err := validateURL(img.Spec.URL); err != nil {
+		errs = append(errs, fmt.Errorf("URL \"%s\" is invalid: %w", img.Spec.URL, err))
+	}
+
+	return errs
+}
--- a/internal/webhooks/metal3.io/v1alpha1/dataimage_webhook.go
+++ b/internal/webhooks/metal3.io/v1alpha1/dataimage_webhook.go
@@ -1,0 +1,83 @@
+/*
+Copyright 2025 The Metal3 Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package webhooks
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	metal3api "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
+	ctrl "sigs.k8s.io/controller-runtime"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+// dataimagelog is for logging in this webhook.
+var dataimagelog = logf.Log.WithName("webhooks").WithName("DataImage")
+
+// SetupWebhookWithManager registers the DataImage validation and defaulting webhooks with the manager.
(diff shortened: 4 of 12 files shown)


internal/webhooks/metal3.io/v1alpha1/common_validation.go
```diff
@@ -0,0 +1,38 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Check the URL scheme
+	allowed := []string{"http", "https", "ftp", "gopher" }
+	if !slices.Contains(allowed, url_obj.Scheme) {
+		return fmt.Errorf("invalid scheme in URL, \"%s\" not allowed", url_obj.Scheme)
+	}
+
```
nuhakala, line 36: Addition to the old BMH url validation, scheme check. Do we want to have this and if so, what schemes we want to include?

nuhakala: Uhh, messed up the comment creation, here is the full snippet:
```go
allowed := []string{"http", "https", "ftp", "gopher" }
if !slices.Contains(allowed, url_obj.Scheme) {
    return fmt.Errorf("invalid scheme in URL, \"%s\" not allowed", url_obj.Scheme)
}
```

lentzi90: Good question. I have no answer, but I am here to learn. Maybe we could find some kind of supported list for IPA and put the same protocols here?

internal/webhooks/metal3.io/v1alpha1/dataimage_webhook_test.go
```diff
@@ -0,0 +1,121 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
lentzi90: Where is this coming from? I am thinking that even if you copied originally from a file where Kubernetes authors had copyright, everything meaningful has now been changed in the file so this could be removed.

internal/webhooks/metal3.io/v1alpha1/hostfirmwarecomponents_webhook_test.go
```diff
@@ -0,0 +1,158 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
lentzi90: Here the same

lentzi90 (commented):
Two nits, otherwise this looks good to me. I don't have much knowledge about what kind of protocols should be accepted though, so let's here what others have to say.

internal/webhooks/metal3.io/v1alpha1/dataimage_webhook_test.go
nuhakala: I copied bmceventsubscription files as template and kept the copyright as is.

lentzi90: I see. Maybe best to keep it then.

lentzi90 (commented):
/lgtm

dtantsur:
Judging by https://github.com/metal3-io/ironic-standalone-operator/pull/679/changes, URL validation can be done with Kubebuilder only without new webhooks.

nuhakala:
> Judging by https://github.com/metal3-io/ironic-standalone-operator/pull/679/changes, URL validation can be done with Kubebuilder only without new webhooks.

That would not allow validating the schema, though not sure if we want to do it anyway. Otherwise it would be quite convenient.

And @lentzi90 said to me that we most likely want to add some other validation than the URI later, so I guess the webhooks could be added anyway (even if the URI validation is done by kubebuilder).

lentzi90:
Hmm that is an interesting option. If we can get away without setting up a separate webhook then that would be very nice. We should be able to get pretty far with kubebuilder and CEL.

@dtantsur do you see any reason/possibility to restrict the protocols? If not, then I am all for going with kubebilder and skipping the webhook.

nuhakala:
No other opinions so I removed the webhooks and did this using the kubebuilder annotations instead.
/cc @Rozzii

Rozzii (commented):
Looks nice, please check the webhook logic (if any) in the corresponding controllers, this validation may make some checks redundant in controllers. If you have already checked then that is great! 

Also please check whether the fuzzer is aligned with this format validation logic, as I understand now the "error" in case of incorrect format will be expressed by the api server so I assume our fuzzer should not expect any error directly from the controllers when in case of a uri format validation.

Let me know if all these areas are clear and I will approve.

nuhakala:
> Looks nice, please check the webhook logic (if any) in the corresponding controllers, this validation may make some checks redundant in controllers. If you have already checked then that is great!

I removed the annotation for the BMH url, because there is already validation in the webhook, so I think it is easier to keep the existing webhook validation and not mix it with the kubebuilder annotation. Even though probably there is no harm having both?

> Also please check whether the fuzzer is aligned with this format validation logic, as I understand now the "error" in case of incorrect format will be expressed by the api server so I assume our fuzzer should not expect any error directly from the controllers when in case of a uri format validation.

Checked the fuzzer, it only checks BMC `GetParsedURL` function and nothing else. That function is not related to the webhooks, because it takes the address as parameter and checks that given address only.

Rozzii:
> > Looks nice, please check the webhook logic (if any) in the corresponding controllers, this validation may make some checks redundant in controllers. If you have already checked then that is great!
> 
> I removed the annotation for the BMH url, because there is already validation in the webhook, so I think it is easier to keep the existing webhook validation and not mix it with the kubebuilder annotation. Even though probably there is no harm having both?
> 
> > Also please check whether the fuzzer is aligned with this format validation logic, as I understand now the "error" in case of incorrect format will be expressed by the api server so I assume our fuzzer should not expect any error directly from the controllers when in case of a uri format validation.
> 
> Checked the fuzzer, it only checks BMC `GetParsedURL` function and nothing else. That function is not related to the webhooks, because it takes the address as parameter and checks that given address only.

So based on this https://github.com/metal3-io/baremetal-operator/blob/ee186a05135ec2b4108b1fde54375efd5825f3de/internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go#L287 The url validation is basically the same either way but for an image BMO also provides a relatively complex checksum validation. I prefer for sake of consistency to remove the url validation from https://github.com/metal3-io/baremetal-operator/blob/ee186a05135ec2b4108b1fde54375efd5825f3de/internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go#L287 , leave only the checksum validation and do the same way wherevere you can find url validation in BMO. Let's offload url validation everywhere in BMO to the api server in a consistent manner. FYI @dtantsur

internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation_test.go
```diff
@@ -593,35 +593,6 @@ func TestValidateCreate(t *testing.T) {
...
-				ObjectMeta: om,
-				Spec: metal3api.BareMetalHostSpec{
-					Image: &metal3api.Image{
-						Checksum: "be254ebfd73e66ca91f6d91f5050aa2ee1ec4813ee65ba472f608ed340cbff09",
-					},
-				},
-			},
-			oldBMH:    nil,
-			wantedErr: "image URL  is invalid: parse \"\": empty url",
-		},
```
Rozzii, line 624: Please check if `// +kubebuilder:validation:Format=uri `  handles this case , if not then keep our own webhook to block the empty value, I suspect if we would allow BMO to accept an empty image url that that could trigger provisioning without an image and that would lead to all kinds of messy situations.

nuhakala:
I deployed BMO to test how the annotation handles empty url. I found out that it actually did not do any validation, I was able to edit BMH resource to contain arbitrary string in the image.url field.

Kubebuilder does the validation using OpenAPI schema 3 validation ([link](https://book.kubebuilder.io/reference/markers/crd-validation.html)):

> Custom resources are validated using the generated OpenAPI v3 schema and must comply with Kubernetes structural schema rules.

Checking OpenAPI v3 specs ([link](https://github.com/OAI/OpenAPI-Specification/blob/3.0.0/versions/3.0.0.md#dataTypeFormat)):

> Primitives have an optional modifier property: format. OAS uses several known formats to define in fine detail the data type being used. However, to support documentation needs, the format property is an open string-valued property, and can have any value. Formats such as "email", "uuid", and so on, MAY be used even though undefined by this specification. Types that are not accompanied by a format property follow the type definition in the JSON Schema. Tools that do not recognize a specific format MAY default back to the type alone, as if the format is not specified.

And the format specifiers listed in the spec did not include uri. So looks like the format=uri annotation is more just documentation that actual validation enforcement. @dtantsur I guess you were not aware of this, otherwise you would not have recommended it, or did I miss something?

Anyway, I am going to revert this back to the implementation using webhooks. @Rozzii

Rozzii (commented):
/approve

internal/webhooks/metal3.io/v1alpha1/common_validation.go
```diff
@@ -0,0 +1,38 @@
...
+	urlObj, err := url.ParseRequestURI(input)
+	if err != nil {
+		return err
+	}
+
+	// Check the URL scheme
+	allowed := []string{"http", "https", "ftp", "gopher"}
+	if !slices.Contains(allowed, urlObj.Scheme) {
+		return fmt.Errorf("invalid scheme in URL, \"%s\" not allowed", urlObj.Scheme)
+	}
```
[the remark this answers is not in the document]
nuhakala: fixed

internal/webhooks/metal3.io/v1alpha1/dataimage_webhook.go
```diff
@@ -0,0 +1,83 @@
...
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
+	ctrl "sigs.k8s.io/controller-runtime"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+// dataimagelog is for logging in this webhook.
+var dataimagelog = logf.Log.WithName("webhooks").WithName("DataImage")
+
+// SetupWebhookWithManager registers the BMCEventSubscription validation and defaulting webhooks with the manager.
```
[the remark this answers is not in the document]
nuhakala: fixed

internal/webhooks/metal3.io/v1alpha1/hostfirmwarecomponents_validation_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+	metal3api "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestHostFirmwareComponentsValidateCreate(t *testing.T) {
+	tm := metav1.TypeMeta{
+		Kind:       "BMCEventSubscription",
+		APIVersion: "metal3.io/v1alpha1",
+	}
```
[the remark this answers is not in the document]
nuhakala: fixed

internal/webhooks/metal3.io/v1alpha1/dataimage_validation.go
```diff
@@ -0,0 +1,32 @@
...
+package webhooks
+
+import (
+	"fmt"
+
+	metal3api "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
+)
+
+// validateDataimage validates DataImage resource for creation and updating.
+func (webhook *DataImage) validateDataimage(img *metal3api.DataImage) []error {
```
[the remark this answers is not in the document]
nuhakala: fixed

internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation_test.go
```diff
@@ -608,6 +608,21 @@ func TestValidateCreate(t *testing.T) {
...
+				Spec: metal3api.BareMetalHostSpec{
+					Image: &metal3api.Image{
+						URL:      "unix://example.com",
+						Checksum: "be254ebfd73e66ca91f6d91f5050aa2ee1ec4813ee65ba472f608ed340cbff09",
+					},
+				},
+			},
+			oldBMH:    nil,
+			wantedErr: "image URL unix://example.com is invalid: invalid scheme in URL, \"unix\" not allowed",
+		},
```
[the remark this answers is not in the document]
nuhakala: Hmm, not sure if we need to valid URL test cases, but I added the test regardless.

lentzi90 (commented):
/lgtm

internal/webhooks/metal3.io/v1alpha1/common_validation.go
```diff
@@ -0,0 +1,38 @@
...
+// validateURL validates the given URL. The URL is assumed to come from HTTP
+// request. On success, no error is returned. Empty string also returns an error.
+func validateURL(input string) error {
+	urlObj, err := url.ParseRequestURI(input)
+	if err != nil {
+		return err
+	}
+
+	// Check the URL scheme
+	allowed := []string{"http", "https", "ftp", "gopher", "oci"}
```
dtantsur, line 32: How did we end up adding ftp and gopher here? Could you follow-up to remove them?

PR #3334, opened 2026-06-09
:seedling: Add support to skip specific parts during setup

terror96:
When running the `hack/ci-e2e.sh` script multiple times, it is not necessary to repeat certain parts of it.

Makes it possible to skip unnecessary parts when running the script again.

Fixes #

**Checklist:**

### DIFF
--- a/hack/ci-e2e.sh
+++ b/hack/ci-e2e.sh
@@ -56,16 +56,39 @@ case "${GINKGO_FOCUS,,}" in
     ;;
 esac
 
+# Set CI_E2E_SKIP_INSTALLATION/CI_E2E_SKIP_BUILDING/CI_E2E_SKIP_SETUP to "true"
+# (case-insensitive) to skip that stage when re-running this script.
+CI_E2E_SKIP_INSTALLATION="${CI_E2E_SKIP_INSTALLATION:-false}"
+CI_E2E_SKIP_BUILDING="${CI_E2E_SKIP_BUILDING:-false}"
+CI_E2E_SKIP_SETUP="${CI_E2E_SKIP_SETUP:-false}"
+
+# Ensure common requirements are installed before proceeding with the setup and tests
+ensure_requirements() {
+  "${REPO_ROOT}/hack/e2e/ensure_go.sh"
+  "${REPO_ROOT}/hack/e2e/ensure_htpasswd.sh"
+  # CAPI test framework uses kubectl in the background
+  "${REPO_ROOT}/hack/e2e/ensure_kubectl.sh"
+  "${REPO_ROOT}/hack/e2e/ensure_yq.sh"
+
+  sudo apt-get update
+  sudo apt-get install -y libvirt-dev pkg-config gettext-base curl
+}
+
+# Build binaries required for the tests
+build_binaries() {
+  # Build the container image with e2e tag (used in tests)
+  IMG=quay.io/metal3-io/baremetal-operator IMG_TAG=e2e make docker
+
+  # Build vbmctl
+  make build-vbmctl
+  sudo setcap cap_net_admin+epi ./bin/vbmctl
+}
+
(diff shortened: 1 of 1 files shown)


hack/ci-e2e.sh
```diff
@@ -56,29 +56,49 @@ case "${GINKGO_FOCUS,,}" in
     ;;
 esac
 
+# These variables are used to control the script behavior
```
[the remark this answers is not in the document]
terror96: Looks sensible.

hack/ci-e2e.sh
```diff
@@ -56,29 +56,49 @@ case "${GINKGO_FOCUS,,}" in
...
-# Build the container image with e2e tag (used in tests)
-IMG=quay.io/metal3-io/baremetal-operator IMG_TAG=e2e make docker
-
-# Build vbmctl
-make build-vbmctl
-sudo setcap cap_net_admin+epi ./bin/vbmctl
+# Build binaries required for the tests
+if [[ "${CI_E2E_SKIP_BUILDING,,}" != "true" ]]; then
+  build_binaries
+fi
```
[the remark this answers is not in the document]
terror96: Feels unnecessary to log such a message, because no phase is skipped by default. It expected that if the user skips e.g. `building` phase, the user realizes the possible effects of it.

hack/ci-e2e.sh
```diff
@@ -142,21 +165,21 @@ fi
...
 
 # Build an ISO image with baked ssh key
 # See https://www.system-rescue.org/scripts/sysrescue-customize/
 # We use the systemrescue ISO and their script for customizing it.
 if [[ ! -f "${IMAGE_DIR}/sysrescue-out.iso" ]];then
-    pushd "${IMAGE_DIR}"
-    wget -O sysrescue-customize https://gitlab.com/systemrescue/systemrescue-sources/-/raw/main/airootfs/usr/share/sysrescue/bin/sysrescue-customize?inline=false
-    chmod +x sysrescue-customize
+  pushd "${IMAGE_DIR}"
+  wget -O sysrescue-customize https://gitlab.com/systemrescue/systemrescue-sources/-/raw/main/airootfs/usr/share/sysrescue/bin/sysrescue-customize?inline=false
```
tuminoid:
```suggestion
  wget -O sysrescue-customize "https://gitlab.com/systemrescue/systemrescue-sources/-/raw/main/airootfs/usr/share/sysrescue/bin/sysrescue-customize?inline=false"
```

tuminoid (commented):
It is named ci-e2e.sh so that gives us the primary usecase, is this something humans could use as well, and hence it would make sense to switch the flags other way around, ie. harmless by default, and CI knows to set the flag?

terror96:
> It is named ci-e2e.sh so that gives us the primary usecase, is this something humans could use as well, and hence it would make sense to switch the flags other way around, ie. harmless by default, and CI knows to set the flag?

Yes, the script can be invoked manually and I have, e.g., been using it for running e2e tests manually. You have a valid point, I just did not want to change the current behavior i.e. we are not skipping anything by default.

hack/ci-e2e.sh
terror96: Sure, I'll fix this one and the other `wget` invocations at 130 - 131. I guess these popped up because I touched the indentation.

tuminoid (commented):
/lgtm

lentzi90 (commented):
/approve

hack/ci-e2e.sh
terror96, line 225: Note: I decided not to skip `wait_for_sushy_tools` even if `CI_E2E_SKIP_SETUP == true` in order to confirm that it is still operational which is not a time consuming operation and causes no state change if run again on a working setup.

nuhakala (commented):
/lgtm

PR #3337, opened 2026-06-10
🌱 e2e: fix breakage caused by hardcoded MAC address in hardwareDetails

dtantsur:
After f20dfc776bd3efa3fd782c50a13b8bc767f3045b, ports are created for
all NICs in HardwareData. However, hardwareDetails used in tests
hardcodes the MAC address of one of the testing nodes, causing tests
that hit this code path to fail with 50% chance.

While here, remove usage of hardware details from tests where it's not
relevant. Inspection can simply be skipped instead.

### DIFF
--- a/test/e2e/automated_cleaning_test.go
+++ b/test/e2e/automated_cleaning_test.go
@@ -50,10 +50,6 @@ var _ = Describe("Automated cleaning", Label("required", "automated-cleaning"), 
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      specName,
 				Namespace: namespace.Name,
-				Annotations: map[string]string{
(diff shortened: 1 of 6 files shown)


tuminoid:
I've pinged the Nordix team for the ARM failure.

1nv0k32:
/lgtm
Just my blessings, I don't have power here :D
@dtantsur Thanks for helping and finding it out.

kashifest (commented):
/approve

tuminoid:
> I've pinged the Nordix team for the ARM failure.

There was some upgrade being done and a hiccup with database, should be good now as seen by passing retest.

tuminoid (commented):
/lgtm

PR #3338, opened 2026-06-11
✨  Add BMH and HNA types and cross resource validation

alegacy:
Introduce the HostNetworkAttachment custom resource for defining switch port configurations (mode, VLANs, MTU) that can be referenced by BareMetalHost network interfaces.

Includes:
- HostNetworkAttachment and HostNetworkAttachmentRef API types
- CRD schema with kubebuilder validation markers
- Webhook for create/update/delete validation
- Switchport mode cross-field validation (e.g., no allowedVLANs in access mode)
- NetworkInterface and SwitchPort types on BMH spec
- BMH webhook validation for NetworkInterface fields
- Immutability enforcement: NetworkInterfaces locked after available state
- HNA update/delete protection when referenced by BMH interfaces
- Field indexer for efficient BMH->HNA reference lookups
- Unit tests for all validation logic

Assisted-by: Claude <noreply@anthropic.com>

Partially implements: https://github.com/metal3-io/metal3-docs/pull/586

**Checklist:**

- [X] Unit tests have been added, if necessary.


alegacy:
No sure why `unit-cover` is failing here... works here locally.

dtantsur:
Let's get the first AI pass, I'll try to get to this next week.

/copilot-review

apis/metal3.io/v1alpha1/hostnetworkattachment_types.go
```diff
@@ -0,0 +1,103 @@
...
+	SwitchportModeHybrid SwitchPortMode = "hybrid"
+)
+
+// HostNetworkAttachmentSpec defines the desired switchport configuration.
+type HostNetworkAttachmentSpec struct {
+	// Mode specifies the network attachment mode.
+	// +kubebuilder:validation:Enum=access;hybrid;trunk
+	// +kubebuilder:required
+	// +kubebuilder:default=access
+	Mode SwitchPortMode `json:"mode"`
```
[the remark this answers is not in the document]
alegacy: fixed

internal/webhooks/metal3.io/v1alpha1/hostnetworkattachment_validation_test.go
```diff
@@ -0,0 +1,765 @@
...
+			expectedError: true,
+			errorContains: "immutable while referenced",
+		},
+		{
+			name:          "invalid-new-spec",
+			oldAttachment: oldAttachment,
+			newAttachment: newAttachmentInvalid,
+			bmhs:          []metal3api.BareMetalHost{},
+			expectedError: true,
+			errorContains: "allowedVlans cannot be specified for access mode",
```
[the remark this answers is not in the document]
alegacy: fixed

internal/webhooks/metal3.io/v1alpha1/hostnetworkattachment_validation_test.go
```diff
@@ -0,0 +1,765 @@
...
+			name: "invalid-access-mode-with-allowed-vlans",
+			attachment: &metal3api.HostNetworkAttachment{
+				Spec: metal3api.HostNetworkAttachmentSpec{
+					Mode:         metal3api.SwitchportModeAccess,
+					NativeVLAN:   100,
+					AllowedVLANs: []int{200},
+				},
+			},
+			expectedError: true,
+			errorContains: "allowedVlans cannot be specified for access mode",
```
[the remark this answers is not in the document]
alegacy: Fixed

internal/webhooks/metal3.io/v1alpha1/hostnetworkattachment_validation.go
```diff
@@ -0,0 +1,151 @@
...
+
+// validateSwitchportConfiguration validates mode-specific switchport constraints
+// that cannot be expressed as CRD schema markers (cross-field validation).
+// VLAN range and mode enum validation are handled by CRD markers on the type.
+func validateSwitchportConfiguration(attachment *metal3api.HostNetworkAttachment) error {
+	if attachment.Spec.Mode == metal3api.SwitchportModeAccess && len(attachment.Spec.AllowedVLANs) > 0 {
+		return errors.New("allowedVlans cannot be specified for access mode")
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
alegacy: Fixed

internal/webhooks/metal3.io/v1alpha1/hostnetworkattachment_validation.go
```diff
@@ -0,0 +1,151 @@
...
+	var references []string
+	for _, bmh := range bmhList.Items {
+		for _, netIf := range bmh.Spec.NetworkInterfaces {
+			refNS := netIf.HostNetworkAttachment.Namespace
+			if refNS == "" {
+				refNS = bmh.Namespace
+			}
+			if netIf.HostNetworkAttachment.Name == attachment.Name && refNS == attachment.Namespace {
+				references = append(references, fmt.Sprintf("%s/%s[%s]", bmh.Namespace, bmh.Name, netIf.Name))
+			}
```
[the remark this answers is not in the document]
alegacy: Fixed

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -598,6 +638,16 @@ type BareMetalHostSpec struct {
...
 	// +kubebuilder:validation:Enum=disabled;agent
 	InspectionMode InspectionMode `json:"inspectionMode,omitempty"`
+
+	// NetworkInterfaces defines the network configuration for each interface.
+	// This will be used to configure switch ports for the host.  Interface
+	// names must correspond to actual NICs discovered in hardware details by
+	// the hardware inspection process.  They are referenced by either the name
+	// or MAC address of the NIC.
+	// This field is immutable unless the host is in an early lifecycle states
+	// (registering, inspecting, preparing, available).
```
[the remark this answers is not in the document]
alegacy: Fixed

internal/webhooks/metal3.io/v1alpha1/hostnetworkattachment_webhook.go
```diff
@@ -0,0 +1,98 @@
...
+
+// log is for logging in this package.
+var hostnetworkattachmentlog = logf.Log.WithName("webhooks").WithName("HostNetworkAttachment")
+
+// bmhNetworkAttachmentIndexField is the field index name for BMH -> HostNetworkAttachment references.
+const bmhNetworkAttachmentIndexField = ".spec.networkInterfaces.hostNetworkAttachment.name"
+
+func (webhook *HostNetworkAttachment) SetupWebhookWithManager(ctx context.Context, mgr ctrl.Manager) error {
+	webhook.Client = mgr.GetClient()
+
```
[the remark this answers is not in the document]
alegacy: The field index selector (bmhNetworkAttachmentIndexField) used in findBMHReferences is a cache-only feature — it relies on the informer cache's field indexer set up via mgr.GetFieldIndexer().IndexField(). Using mgr.GetAPIReader() would bypass the cache entirely and wouldn't support field selectors, so we'd have to list all BareMetalHost resources across all namespaces and filter in-memory. For clusters with many BMH objects this would be a significant performance regression on every HNA create/update/delete.

The staleness window of the informer cache is quite small (not impossible, but small), and the race condition — a BMH referencing an HNA being created and that HNA being deleted before the cache syncs — is unlikely in practice. If it did occur, the consequence is allowing a delete that would have been blocked, which is recoverable by recreating the HNA.

main.go
```diff
@@ -498,7 +504,7 @@ func main() {
 
 	if enableWebhook {
 		setupWebhookReadinessCheck(mgr)
-		setupWebhooks(mgr)
+		setupWebhooks(context.Background(), mgr)
 	}
```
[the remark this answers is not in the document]
alegacy: Fixed

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -458,6 +458,46 @@ type FirmwareConfig struct {
...
+type NetworkInterface struct {
+	// Name of the network interface (e.g., "eth0", "ens1f0")
+	// This must match the name of a NIC in status.hardware.nics.
+	// If MACAddress is set then this field is ignored.
+	// +optional
+	// +kubebuilder:validation:Pattern=`^[a-zA-Z0-9._-]+$`
+	Name string `json:"name,omitempty"`
+
+	// MAC address of the network interface.  This must match the MAC address
+	// of the network interface in status.hardware.nics.
```
dtantsur: nit: `status.hardware` is soft-deprecated, we should refer to HardwareData resources

apis/metal3.io/v1alpha1/hostnetworkattachment_types.go
```diff
@@ -0,0 +1,102 @@
...
+// HostNetworkAttachmentRef references a HostNetworkAttachment for interface configuration.
+type HostNetworkAttachmentRef struct {
+	// Name of the HostNetworkAttachment resource
+	// +kubebuilder:validation:MaxLength=253
+	Name string `json:"name,omitempty"`
+
+	// Namespace of the HostNetworkAttachment (defaults to BMH namespace)
+	// +optional
+	// +kubebuilder:validation:MaxLength=63
+	Namespace string `json:"namespace,omitempty"`
```
dtantsur, line 31: I'm still somewhat sceptical about cross-namespace access here.. especially after the avalanche of CVEs we've fixed recently

apis/metal3.io/v1alpha1/hostnetworkattachment_types.go
```diff
@@ -0,0 +1,102 @@
...
+	MTU *int `json:"mtu,omitempty"`
+}
+
+// HostNetworkAttachment defines switchport configuration for BMH network interfaces.
+// Spec fields are mutable when no BMH references the attachment, immutable when in use.
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="Mode",type="string",JSONPath=".spec.mode",description="Switchport mode"
+// +kubebuilder:printcolumn:name="Native VLAN",type="integer",JSONPath=".spec.nativeVLAN",description="Native VLAN ID"
+// +kubebuilder:printcolumn:name="MTU",type="integer",JSONPath=".spec.mtu",description="Maximum Transmission Unit"
```
dtantsur: nit: not sure if MTU is useful by default.. or is it going to be set in most cases?

internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -427,3 +438,54 @@ func validatePowerStatus(host *metal3api.BareMetalHost) error {
 	}
 	return nil
 }
+
+// networkInterfacesMutableStates lists the states in which networkInterfaces
+// can be added, modified, or removed. All other states are considered
+// immutable to prevent disruption to provisioned or in-progress hosts.
```
dtantsur: We generally don't guard changes on state like this. We could allow changes but not actually apply them to Ironic unless in a suitable state.

internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -427,3 +438,54 @@ func validatePowerStatus(host *metal3api.BareMetalHost) error {
...
+var networkInterfacesMutableStates = map[metal3api.ProvisioningState]bool{
+	metal3api.StateNone:         true,
+	metal3api.StateUnmanaged:    true,
+	metal3api.StateRegistering:  true,
+	metal3api.StateInspecting:   true,
+	metal3api.StatePreparing:    true,
```
dtantsur: You write "prevent disruption to ... in-progress hosts", but these two states are in-progress ones.

internal/webhooks/metal3.io/v1alpha1/hostnetworkattachment_validation.go
```diff
@@ -0,0 +1,155 @@
...
+	var warnings admission.Warnings
+
+	// Check if any BMH still references this attachment
+	references, err := webhook.findBMHReferences(ctx, attachment)
+	if err != nil {
+		return warnings, fmt.Errorf("failed to check BMH references: %w", err)
+	}
+
+	if len(references) > 0 {
+		warnings = append(warnings, "This attachment is referenced by: "+strings.Join(references, ", "))
```
dtantsur, line 100: I'm not really fond of accessing Kubernetes API inside a webhook, cannot we use a finalizer for the same purpose?

internal/webhooks/metal3.io/v1alpha1/hostnetworkattachment_webhook.go
```diff
@@ -0,0 +1,98 @@
...
+	// Register field indexer for efficient BMH reference lookups
+	// This allows us to quickly find all BMHs that reference a specific HostNetworkAttachment
+	if err := mgr.GetFieldIndexer().IndexField(
```
dtantsur, line 41: Neat! Maybe it's better than handling finalizers indeed.

apis/metal3.io/v1alpha1/baremetalhost_types.go
alegacy: fixed

apis/metal3.io/v1alpha1/hostnetworkattachment_types.go
alegacy: Our use case is that BMHs are created ahead of time in some namespace(s), then later some subset of those BMHs are used to form a cluster.   The BMHs may all be in the same namespace or they may be in different namespaces.   There is a separate namespace created for the cluster and that's likely where the HNA would live along side other cluster-specific CRDs.

I understand that having a cross-namespace reference between, say, a BMH and Secret may be problematic, but in this case it would be a BMH and an HNA.  Both should be admin owned and the HNA doesn't contain any sensitive data so there shouldn't be much risk of allowing that access from BMO.

apis/metal3.io/v1alpha1/hostnetworkattachment_types.go
alegacy: removed

internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
alegacy: Ok, that's going to require some rework of my next commits, but I'm willing to remove this now and try to find an alternative approach to those other commits.   May have to revisit this later.

alegacy: Removed

internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
alegacy: Removed (see other comments)

internal/webhooks/metal3.io/v1alpha1/hostnetworkattachment_validation.go
alegacy: Based on your other comment in _webhook.go it looks like maybe you are ok with this approach now?

alegacy: if not ... how would you see finalizers working for this purpose?   We need to ensure that a HNA isn't removed while a BMH refers to it.   Many BMHs may reference a single HNA so we certainly don't want to have to add many finalizers (1:1 w/ BMH).  Adding just 1 finalizer whenever _any_ BMH refers to a HNA might be challenging to manage safely without race conditions.

dtantsur: Let's keep it this way unless somebody else objects

apis/metal3.io/v1alpha1/baremetalhost_types.go
dtantsur: Doesn't look fixed to me (can be a follow-up).

internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -446,3 +452,25 @@ func validatePowerStatus(host *metal3api.BareMetalHost) error {
...
+	var errs []error
+	seen := make(map[string]bool)
+
+	for i, iface := range networkInterfaces {
+		// At least one identifier (Name or MACAddress) must be specified
+		if !iface.IsValid() {
+			errs = append(errs, fmt.Errorf("networkInterfaces[%d]: must specify either name or macAddress", i))
+			continue
+		}
+
```
dtantsur, line 467: Since Name is ignored when MACAddress is provided, maybe make them exclusive?

internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
```diff
@@ -446,3 +452,25 @@ func validatePowerStatus(host *metal3api.BareMetalHost) error {
 	}
 	return nil
 }
+
+// validateNetworkInterfaces validates NetworkInterface specifications.
+func validateNetworkInterfaces(networkInterfaces []metal3api.NetworkInterface) []error {
+	var errs []error
+	seen := make(map[string]bool)
```
dtantsur: nit: use `len(networkInterfaces)` as capacity

internal/webhooks/metal3.io/v1alpha1/hostnetworkattachment_validation.go
```diff
@@ -0,0 +1,155 @@
...
+	// Fail-closed: if we cannot verify references, reject the update
+	references, err := webhook.findBMHReferences(ctx, oldAttachment)
+	if err != nil {
+		// Unable to verify if attachment is in use - reject update for safety
+		return warnings, fmt.Errorf("failed to check BMH references, cannot safely allow update: %w", err)
+	}
+
+	if len(references) > 0 {
+		// Only enforce immutability if there are active references
+		warnings = append(warnings, fmt.Sprintf("Cannot modify attachment while in use by: %s. Remove references first or create a new attachment.",
```
dtantsur: Hmm, why have both a warning and an error?

apis/metal3.io/v1alpha1/hostnetworkattachment_types.go
dtantsur: We don't have a controller yet (right?) but maybe we'll need to provide a knob for an admin to tune this behavior? For instance, always allow HNA from the same namespace, but create an option that tunes which other namespaces can be used (including "any")?

One case that is definitely going to be annoying: if someone uses your HNA, the namespace is now not deletable. This is going to wedge tools that love deleting namespaces like GitOps.

internal/webhooks/metal3.io/v1alpha1/hostnetworkattachment_validation.go
```diff
@@ -0,0 +1,155 @@
...
+			strings.Join(references, ", ")))
+		errs = append(errs, fmt.Errorf("HostNetworkAttachment spec is immutable while referenced by BMH interfaces: %s",
+			strings.Join(references, ", ")))
+	} else {
+		hostnetworkattachmentlog.V(1).Info("no BMH references found, allowing update",
+			"namespace", newAttachment.Namespace, "name", newAttachment.Name)
+	}
+
+	if len(errs) > 0 {
+		return warnings, kerrors.NewAggregate(errs)
```
dtantsur: nit: we now have `errors.Join` in stdlib (dunno if NewAggregate is objectively better)

alegacy: Given the rest of the code base uses NewAggregate I'd prefer to keep it as is until the rest are changed.

internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
alegacy: ok, that should be fine given having both isn't useful.  will change

apis/metal3.io/v1alpha1/baremetalhost_types.go
alegacy: sorry, I thought I did... will change now.

apis/metal3.io/v1alpha1/hostnetworkattachment_types.go
alegacy: If you are ok leaving it is as is for now, I can look at that as a follow-up.  I don't recall seeing that pattern elsewhere (e.g., a knob controlling which namespaces a CR can refer out to, but that doesn't mean it doesn't exist).  Not sure it would be incredibly valuable though... if 1 other namespace is allowed you might as well allow them all.  Like I said earlier its not like these are secrets.  Lets revisit.

internal/webhooks/metal3.io/v1alpha1/hostnetworkattachment_validation.go
alegacy: fixed

internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
alegacy: fixed

internal/webhooks/metal3.io/v1alpha1/baremetalhost_validation.go
alegacy: fixed

apis/metal3.io/v1alpha1/baremetalhost_types.go
alegacy: comment fixed

apis/metal3.io/v1alpha1/hostnetworkattachment_types.go
dtantsur: Yep, it can be done in the same PR as the controller logic.

I'm trying to be conservative by default. Defaulting to only same-namespace HNA makes it easy to reason about the scope of the system.

dtantsur:
/copilot-review

Let's get another pass with all the updates.

config/base/crds/bases/metal3.io_hostnetworkattachments.yaml
```diff
@@ -0,0 +1,102 @@
...
+                allOf:
+                - enum:
+                  - access
+                  - trunk
+                  - hybrid
+                - enum:
+                  - access
+                  - hybrid
+                  - trunk
+                default: access
```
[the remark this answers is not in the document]
alegacy: This looks like it is because of a limitation of the Makefile targets that was introduced by 4e1a22ed9.    I'll fix it in a separate commit and include here.

config/base/crds/bases/metal3.io_baremetalhosts.yaml
```diff
@@ -345,6 +345,72 @@ spec:
...
                 x-kubernetes-map-type: atomic
+              networkInterfaces:
+                description: |-
+                  NetworkInterfaces defines the network configuration for each interface.
+                  This will be used to configure switch ports for the host.  Interface
+                  names must correspond to actual NICs discovered in hardware details by
+                  the hardware inspection process.  They are referenced by either the name
+                  or MAC address of the NIC.
+                  This field is immutable unless the host is in an early lifecycle states
+                  (registering, inspecting, preparing, available).
```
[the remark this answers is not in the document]
alegacy: This looks like it is because of a limitation of the Makefile targets that was introduced by 4e1a22ed9.    I'll fix it in a separate commit and include here.

internal/webhooks/metal3.io/v1alpha1/hostnetworkattachment_validation.go
```diff
@@ -0,0 +1,152 @@
...
+		return warnings, nil
+	}
+
+	// Spec has changed - check if any BMH references this attachment
+	// Fail-closed: if we cannot verify references, reject the update
+	references, err := webhook.findBMHReferences(ctx, oldAttachment)
+	if err != nil {
+		// Unable to verify if attachment is in use - reject update for safety
+		return warnings, fmt.Errorf("failed to check BMH references, cannot safely allow update: %w", err)
+	}
```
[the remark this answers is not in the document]
alegacy: fixed

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -1156,6 +1206,19 @@ func (raid *RAIDConfig) GetRootVolumeCount() (rootCount int) {
 	return rootCount
 }
 
+// IsValid returns true if the network interface is valid.
```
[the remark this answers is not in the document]
alegacy: fixed

config/base/crds/bases/metal3.io_hostnetworkattachments.yaml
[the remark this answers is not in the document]
alegacy: fixed

config/base/crds/bases/metal3.io_baremetalhosts.yaml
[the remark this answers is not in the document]
alegacy: fixed

apis/metal3.io/v1alpha1/hostnetworkattachment_types.go
```diff
@@ -0,0 +1,101 @@
...
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=4094
+	NativeVLAN int `json:"nativeVLAN"`
+
+	// AllowedVLANs specifies a list of VLAN IDs that are allowed on this network attachment.
+	// This is typically used in trunk or hybrid mode to specify which tagged VLANs can be carried on the interface.
+	// +optional
+	// +kubebuilder:validation:items:Minimum=1
+	// +kubebuilder:validation:items:Maximum=4094
+	AllowedVLANs []int `json:"allowedVLANs,omitempty"`
```
alegacy: @dtantsur  Oh, I just realized that this is wrong.  I intended for this to be an array of strings so that a user could supply ranges if needed (e.g., 2,3,10-40,55) without having to enumerate them all.

alegacy: I've added an additional commit to handle those cases.  Would you mind taking another look?

PR #3341, opened 2026-06-12
:bug: Move PreprovisioningImage controller back to public package

honza:
The PPI controller is designed to be imported by other projects, so it should not be under internal/. This reverts the accidental move from 592a226.


dtantsur:
If it's designed to be reused, let's put it under `pkg/`.

There is also some paperwork you need to do - see the failing jobs.

adilGhaffarDev (commented):
/lgtm

PR #3358, opened 2026-06-17
:bug: Clean-up temporary CA files when Ironic CR are used

dtantsur:
Currently, temporary stored files keep accummulating.

### DIFF
--- a/pkg/provisioner/ironic/factory.go
+++ b/pkg/provisioner/ironic/factory.go
@@ -247,6 +247,14 @@ func (f ironicProvisionerFactory) ironicProvisioner(ctx context.Context, hostDat
 	if f.ironicName != "" && f.ironicNamespace != "" && f.k8sClient != nil {
 		createClient = func() (*gophercloud.ServiceClient, error) {
 			ironicEndpoint, ironicAuth, tlsConf, err := f.loadConfigFromIronicCR(ctx)
+			// NOTE(dtantsur): the file may be created even if an error is returned, so plan deletion before checking err.
+			if tlsConf.TrustedCAFileIsTemporary {
+				defer func() {
+					if rmErr := os.Remove(tlsConf.TrustedCAFile); rmErr != nil && !os.IsNotExist(rmErr) {
+						provisionerLogger.Error(rmErr, "failed to remove temporary CA file", "file", tlsConf.TrustedCAFile)
+					}
+				}()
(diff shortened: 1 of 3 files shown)


pkg/provisioner/ironic/factory.go
```diff
@@ -247,6 +247,13 @@ func (f ironicProvisionerFactory) ironicProvisioner(ctx context.Context, hostDat
 	if f.ironicName != "" && f.ironicNamespace != "" && f.k8sClient != nil {
 		createClient = func() (*gophercloud.ServiceClient, error) {
 			ironicEndpoint, ironicAuth, tlsConf, err := f.loadConfigFromIronicCR(ctx)
+			if tlsConf.TrustedCAFileIsTemporary {
+				defer func() {
```
zaneb, line 252: It's safe to delete it at the end of this function because by the time we have constructed the ironic client, the file has already been read and is not needed again when we actually use it?

pkg/provisioner/ironic/factory.go
```diff
@@ -247,6 +247,13 @@ func (f ironicProvisionerFactory) ironicProvisioner(ctx context.Context, hostDat
 	if f.ironicName != "" && f.ironicNamespace != "" && f.k8sClient != nil {
 		createClient = func() (*gophercloud.ServiceClient, error) {
 			ironicEndpoint, ironicAuth, tlsConf, err := f.loadConfigFromIronicCR(ctx)
```
zaneb, line 249: There's at least one place in `loadConfigFromIronicCR()` where an error would still cause the file returned by `loadTLSConfigFromIronicCR()` to be leaked.

pkg/provisioner/ironic/factory.go
dtantsur: Yeah, apparently we're using a helper that loads the files to `[]bytes` (but cannot accept `[]bytes` directly)

pkg/provisioner/ironic/factory.go
```diff
@@ -511,10 +522,15 @@ func writeTempFile(prefix string, data []byte) (string, error) {
 	}
 	defer tmpFile.Close()
```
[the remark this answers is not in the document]
dtantsur: No Windows support expected

iurygregory (commented):
/lgtm

PR #3359, opened 2026-06-18
🌱 Use proper IPA cache address

Rozzii:
This commit:
 - Makes sure that the correct IPA Nordix proxy is used.

Previous address circumvented the proper cache refresh trigger so in case the cache was empty after a cleanup, cache refresh was not triggered.


smoshiur1237 (commented):
/lgtm

tuminoid:
/retitle 🌱 Use proper IPA cache address

Not product bug.

lentzi90 (commented):
/approve

PR #3378, opened 2026-06-25
🌱 vbmctl: Make VM template portable across distros

adilGhaffarDev:
The VM.xml.tpl template had values hardcoded for Ubuntu's QEMU packages
that prevented vbmctl from working on CentOS/RHEL:

1. `machine='pc-q35-6.2'` — a versioned machine type only available in
   Ubuntu's QEMU. CentOS/RHEL uses different versions (`pc-q35-rhel9.*`),
   causing "unsupported machine type" errors.

2. `<emulator>/usr/bin/qemu-system-x86_64</emulator>` — Ubuntu's QEMU
   binary path. On CentOS/RHEL it's at `/usr/libexec/qemu-kvm`.

3. `<model type='cirrus' .../>` — the Cirrus VGA adapter is deprecated
   and not available on all QEMU builds.

Fix:
- Use the `q35` machine type alias, which resolves to the latest
  available q35 version on any distro.
- Remove the `<emulator>` element so libvirt auto-detects the correct
  QEMU binary path for the host.
- Switch video model from `cirrus` to `virtio`, which is supported
  across all modern QEMU/libvirt versions.

Fixes #

**Checklist:**

### DIFF
--- a/test/vbmctl/pkg/libvirt/templates/VM.xml.tpl
+++ b/test/vbmctl/pkg/libvirt/templates/VM.xml.tpl
@@ -4,7 +4,7 @@
   <currentMemory unit='KiB'>{{ .Memory }}</currentMemory>
   <vcpu placement='static'>{{ .VCPUs }}</vcpu>
   <os firmware="efi">
-    <type arch='x86_64' machine='pc-q35-6.2'>hvm</type>
+    <type arch='x86_64' machine='q35'>hvm</type>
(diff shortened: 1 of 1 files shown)


lentzi90 (commented):
Thanks @adilGhaffarDev !
/approve

Sunnatillo (commented):
/lgtm

PR #3381, opened 2026-06-26
🌱 Rename getSecretData() to getSecretDataWithFinalizer()

MahnoorAsghar:
Rename getSecretData() to getSecretDataWithFinalizer(), as the function can add a finalizer to the Secret, and is not a pure read-only function

Related comment: https://github.com/metal3-io/baremetal-operator/pull/3265#discussion_r3360575275
**Checklist:**

### DIFF
--- a/internal/controller/metal3.io/host_config_data.go
+++ b/internal/controller/metal3.io/host_config_data.go
@@ -36,9 +36,10 @@ type hostConfigData struct {
 }
 
 // Generic method for data extraction from a Secret. Function uses dataKey
-// parameter to detirmine which data to return in case secret contins multiple
-// keys.
-func (hcd *hostConfigData) getSecretData(ctx context.Context, name, namespace, dataKey string, addFinalizer bool) (string, error) {
+// parameter to determine which data to return in case secret contains multiple
+// keys. The addFinalizer parameter indicates whether or not a finalizer will be
+// added to the secret.
+func (hcd *hostConfigData) getSecretDataWithFinalizer(ctx context.Context, name, namespace, dataKey string, addFinalizer bool) (string, error) {
 	if namespace != hcd.host.Namespace {
 		return "", fmt.Errorf("%s secret must be in same namespace as host %s/%s", dataKey, hcd.host.Namespace, hcd.host.Name)
 	}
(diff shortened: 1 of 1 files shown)


adilGhaffarDev (commented):
/lgtm

internal/controller/metal3.io/host_config_data.go
```diff
@@ -36,9 +36,10 @@ type hostConfigData struct {
 }
 
 // Generic method for data extraction from a Secret. Function uses dataKey
-// parameter to detirmine which data to return in case secret contins multiple
-// keys.
-func (hcd *hostConfigData) getSecretData(ctx context.Context, name, namespace, dataKey string, addFinalizer bool) (string, error) {
+// parameter to determine which data to return in case secret contains multiple
+// keys. The addFinalizer parameter indicates whether or not a finalizer will be
+// added to the secret.
+func (hcd *hostConfigData) getSecretDataWithFinalizer(ctx context.Context, name, namespace, dataKey string, addFinalizer bool) (string, error) {
```
adilGhaffarDev, line 42:
```suggestion
func (hcd *hostConfigData) getSecretDataAndAddFinalizer(ctx context.Context, name, namespace, dataKey string, addFinalizer bool) (string, error) {
```

adilGhaffarDev (commented):
Maybe this is better.

internal/controller/metal3.io/host_config_data.go
MahnoorAsghar: I'll let others weigh in as well :)

dtantsur: I agree that `WithFinalizer` does not precisely convey the the finalizer is added (it sounds like it's retrieved). Note that we already have `ObtainSecretWithFinalizer`, so there is prior art. We'd need to rename it too. It's also going to be odd to have an `addFinalizer` argument to `getSecretDataAndAddFinalizer` :)

I think I'd go with Mahnoor's suggestion and move on. I also don't mind if everything is renamed, but let's converge on one option.

dtantsur:
/approve

The naming is not perfect but it's an improvement. We can iterate more on this, but we may as well have better things to do with our lives :)

PR #3382, opened 2026-06-27
:bug: Make NeedsMAC semantics consistent across BMC drivers

### LINKED ISSUE #3305: Inconsistent NeedsMAC implementations in various drivers
NeedsMAC is supposed to define whether the driver requires BootMACAddress to be populated.

The current implementations are inconsistent in what they imply:
1. For Redfish virtual media, it is set to false because the Redfish driver in Ironic can pre-populated MAC addresses during inspection and because booting over virtual media does not strictly require having a MAC address.
2. For Redfish with network booting, it is set to true because network booting requires a MAC address, even though Redfish may be able to pre-populate MACs.
3. In case of IPMI, it's only set to true for the largely unused `libvirt://` scheme. The apparent reasoning is that inspection has a fallback to `inspector.ipxe` (so called "unmanaged" inspection) if the MAC address is not recognized.

Every reasoning here makes certain sense, but they all are incompatible with each other. We need to settle on one, and my proposal would be to drop NeedsMAC entirely and assume that BootMACAddress is required when inspection is disabled.

The only problem is that it is still required for virtual BMC, and I don't understand why and how to handle it.

/kind bug

mvanhorn:
## Summary

`NeedsMAC()` on the `AccessDetails` interface is meant to declare whether a
driver requires `BootMACAddress` to be populated, but the per-driver
implementations were mutually inconsistent (#3305):

- Redfish virtual media returned `false`.
- Redfish network boot and `redfish-uefihttp` returned `true`.
- IPMI returned `true` only for the `libvirt://` (VirtualBMC) scheme, while
  plain `ipmi://` returned `false` despite also booting via iPXE.

This change settles on a single rule: `NeedsMAC()` reports only the
requirement that holds **regardless of inspection state**. That is true
solely for VirtualBMC, whose IPMI addresses are non-unique and cannot be
fetched from inside the VM, so the MAC can never be discovered. All
Redfish-family drivers now return `false`, matching the existing behaviour of
the plain IPMI driver: Ironic can pre-populate MAC addresses during
inspection, so a `BootMACAddress` is not required up front.

The separate "MAC required when inspection is disabled" guarantee is left to
the consumers. The validating webhook already applied
`NeedsMAC() || host.InspectionDisabled()`; this aligns the ironic `Register`
backstop to the same combined condition by threading `DisableInspection`
through `ManagementAccessData`, so both call sites enforce the rule
identically.

## Why this matters

As the issue author describes, each individual `NeedsMAC()` reasoning made
local sense but the set as a whole was incoherent, and the proposed direction
is to drop the ad-hoc divergence and require `BootMACAddress` only when
inspection is disabled, while still honoring the genuine VirtualBMC
requirement. This makes the driver-level booleans consistent and keeps the
inspection-state rule in one place (the consumers), rather than duplicating
conflicting heuristics across drivers.

## Testing

- `pkg/hardwareutils/bmc`: updated the per-scheme `NeedsMAC()` table cases
  (`redfish`, `redfish uefi http boot`, `idrac redfish` now expect `false`)
  and ran the package tests.
- `pkg/provisioner/ironic`: added `TestRegisterNoMACInspectionDisabled`
  covering the combined condition (a non-VirtualBMC driver with inspection
  disabled and no `BootMACAddress` is rejected), plus the existing
  `TestRegisterNoMAC` / `TestRegisterMACOptional` cases.
- `go vet` and `gofmt` clean on the touched packages; both affected modules
  (main and `pkg/hardwareutils`) pass their unit tests.

Fixes #3305

### DIFF
--- a/internal/controller/metal3.io/baremetalhost_controller.go
+++ b/internal/controller/metal3.io/baremetalhost_controller.go
@@ -956,6 +956,7 @@ func (r *BareMetalHostReconciler) registerHost(ctx context.Context, prov provisi
 			DisablePowerOff:            info.host.Spec.DisablePowerOff,
 			CPUArchitecture:            getHostArchitecture(info.host),
 			HardwareData:               info.hardwareData,
+			DisableInspection:          info.host.InspectionDisabled(),
 		},
 		credsChanged,
 		info.host.Status.ErrorType == metal3api.RegistrationError)
--- a/pkg/hardwareutils/bmc/access.go
+++ b/pkg/hardwareutils/bmc/access.go
@@ -55,8 +55,12 @@ type AccessDetails interface {
 	// will be used to communicate with it.
 	Type() string
 
-	// NeedsMAC returns true when the host is going to need a separate
-	// port created rather than having it discovered.
+	// NeedsMAC returns true when the driver requires a BootMACAddress to be
+	// populated regardless of whether inspection is enabled, for example
+	// VirtualBMC where the MAC cannot be discovered. Drivers whose MAC can be
+	// discovered during inspection return false; for those, a MAC is only
+	// required when inspection is disabled, which the callers enforce
+	// separately via host.InspectionDisabled().
 	NeedsMAC() bool
 
 	// The name of the driver to instantiate the BMC with. This may differ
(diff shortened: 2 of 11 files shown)


Rozzii (commented):
/approve

Rozzii:
Thanks for the contribution @mvanhorn !

Rozzii:
/retest

I think this is a bug of the test itself, I think the "timing" for checking the provisioning state is not correct.

Rozzii:
> /retest
> 
> I think this is a bug of the test itself, I think the "timing" for checking the provisioning state is not correct.

Yep, second time it has passed.

PR #3387, opened 2026-06-30
:sparkles: surface missing or inaccessible referenced config Secrets in BMH status

### LINKED ISSUE #3307: Improve BMH status visibility when referenced CRs (userData, secrets) are missing or inaccessible
When using BMO as a BMaaS platform (provisioning arbitrary bare-metal instances, not only k8s nodes), non-admin users encounter scenarios where provisioning silently stalls because a CR referenced by the BareMetalHost — such as `userData` or a Secret — is not present or not accessible due to RBAC.

In these cases:
- Provisioning does not progress
- The BMH status does not indicate *why* it's stuck
- The end-user has no way to diagnose the issue without access to BMO controller logs (which typically require admin privileges)

This makes BMO difficult to use in multi-tenant or self-service scenarios where operators (non-admins) submit BMH provisioning requests but cannot inspect controller logs.
originally posted by @jacob-anders [in the PR comment](https://github.com/metal3-io/baremetal-operator/pull/3111#issuecomment-4515016475)

mvanhorn:
## What this PR does

When a `BareMetalHost` references a `userData`, `networkData`, or `metaData` Secret that does not exist or is not readable by the controller, the reason is now surfaced on the BMH status and as an Event.

The config-data Secrets are read through `getSecretDataWithFinalizer` in `internal/controller/metal3.io/host_config_data.go`, which previously returned the raw error from `secretManager.ObtainSecretWithFinalizer`. That error propagated through `prov.Provision` and was handled by `actionProvisioning` as a generic `actionError`, which requeues but leaves the status empty.

- Adds a typed `SecretAccessError` in `errors.go` that carries the Secret name and the reference kind (`userData` / `networkData` / `metaData`) and implements `Unwrap()`, so callers can still inspect the underlying error (for example `k8serrors.IsNotFound`).
- In `getSecretDataWithFinalizer`, wraps the retrieval failure in `SecretAccessError` only when the underlying error is `IsNotFound` or `IsForbidden` (the two cases this issue describes: the Secret is missing or inaccessible). Other failures, such as a transient conflict while adding the finalizer, are returned unwrapped so they are still retried as ordinary action errors rather than recorded as a fatal provisioning error.
- In `actionProvisioning`, detects `SecretAccessError` with `errors.As` and reports it through the existing `recordActionFailure` path, which sets the status error message naming the Secret and publishes an Event. The host is requeued with the standard backoff, so provisioning resumes once the Secret is created, mirroring the existing BMC-secret handling (`ResolveBMCSecretRefError`).

The deletion-flow handling in `PreprovisioningNetworkData` is unchanged: `k8serrors.IsNotFound` still matches through `Unwrap`, and the wrong-key path still returns `NoDataInSecretError` directly.

## Why we need it

When BMO is used as a bare-metal-as-a-service platform, non-admin users submit `BareMetalHost` resources but cannot read the controller logs. If a referenced Secret is missing or blocked by RBAC, provisioning stalls and the BMH status gives no reason, so the user has no way to diagnose the stall. Events are visible to non-admin API users, so publishing one alongside the status message lets the operator see and fix the problem without admin access. This is what makes BMO usable in multi-tenant and self-service scenarios.

Fixes #3307

## Checklist

### DIFF
--- a/internal/controller/metal3.io/baremetalhost_controller.go
+++ b/internal/controller/metal3.io/baremetalhost_controller.go
@@ -1413,6 +1413,10 @@ func (r *BareMetalHostReconciler) actionProvisioning(ctx context.Context, prov p
 		ImagePullSecret: authSecret,
 	}, forceReboot)
 	if err != nil {
+		var secretErr SecretAccessError
+		if errors.As(err, &secretErr) {
+			return recordActionFailure(info, metal3api.ProvisioningError, secretErr.Error())
+		}
 		return actionError{fmt.Errorf("failed to provision: %w", err)}
 	}
 
--- a/internal/controller/metal3.io/errors.go
+++ b/internal/controller/metal3.io/errors.go
@@ -44,3 +44,22 @@ type NoDataInSecretError struct {
 func (e NoDataInSecretError) Error() string {
 	return fmt.Sprintf("Secret %s does not contain key %s", e.secret, e.key)
 }
+
+// SecretAccessError is returned when a Secret referenced by the host for
+// configuration data (userData, networkData, or metaData) cannot be retrieved
+// because it does not exist or is not accessible. It wraps the underlying error
+// so callers can still inspect it with errors.Is/As (for example
+// k8serrors.IsNotFound).
+type SecretAccessError struct {
+	secret string
+	key    string
+	err    error
+}
+
+func (e SecretAccessError) Error() string {
+	return fmt.Sprintf("could not retrieve %s secret %q: %v", e.key, e.secret, e.err)
+}
+
+func (e SecretAccessError) Unwrap() error {
+	return e.err
+}
--- a/internal/controller/metal3.io/host_config_data.go
+++ b/internal/controller/metal3.io/host_config_data.go
@@ -51,6 +51,14 @@ func (hcd *hostConfigData) getSecretDataWithFinalizer(ctx context.Context, name,
 
 	secret, err := hcd.secretManager.ObtainSecretWithFinalizer(ctx, key, addFinalizer)
 	if err != nil {
+		// Only treat a missing or inaccessible Secret as a SecretAccessError so the
+		// reason is surfaced in the BMH status. Other failures (for example a
+		// transient conflict while adding the finalizer) are returned unwrapped so
+		// they are retried as ordinary action errors rather than recorded as a fatal
+		// provisioning error.
+		if k8serrors.IsNotFound(err) || k8serrors.IsForbidden(err) {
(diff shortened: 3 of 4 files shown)


mvanhorn:
Thanks for running the review, @tuminoid. Both points are addressed in d545370.

The first one was right and the consequence was worse than extra events: a Secret that simply had not been created yet took the host through a full deprovision / available / provision cycle. `actionProvisioning` now records that condition through a retryable path that surfaces it on the host and requeues at the fixed wait-for-Secret delay without touching ErrorCount, and clears it once the data becomes readable. `handleProvisioning` aborts into deprovisioning only on a fatal provisioning error.

On coverage: added the Forbidden wrapper case, unrelated retrieval failures staying unwrapped, only `SecretAccessError` producing status and Event output, and a state-machine test that pins staying in provisioning and then recovering.

One thing I want to flag rather than bury, because it is a design choice you may not like. To tell the retryable case from a fatal one I match the persisted `Status.ErrorMessage` against the shape `SecretAccessError.Error()` produces. That works, but it is brittle in two ways: reword the message and the retry behavior silently reverts to deprovisioning, and any other provisioning error whose message happens to match would stop being fatal.

The clean version is a distinct `ErrorType` constant rather than sniffing text, and as far as I can tell nothing else in the tree pattern-matches `Status.ErrorMessage`. I did not do that because it adds a ninth value to a public API enum and that is your call, not mine. Happy to switch it if you would prefer, and if so, whether you want a new type or a different discriminator entirely.

mvanhorn:
Rebased onto `main` in `b0ac54a`; the branch was conflicting and every prow job was red as a result.

Two conflicts, both same-anchor drift in `host_state_machine.go` and its test:

- `main` added a `ProvisioningFailCount` increment inside the `ErrorType != ""` branch. I kept it, gated inside this PR's narrowed condition (`fatalProvisioningError() || provisioningCancelled()`), so a retried missing-Secret never inflates the count while every error that still aborts is counted exactly as `main` does.
- The mock provisioner struct gained `registerErr` / `preprovImageFormats` on `main` and `provisionErr` / `provisionCalls` here. Unioned both.

One test assertion also had to go: `TestActionProvisioningSecretAccessErrorOnly` asserted `reconcileResult.Requeue`, which controller-runtime v0.24 deprecates and `actionFailed.Result()` no longer sets. `RequeueAfter == hostErrorRetryDelay` is asserted on the next line and is the real signal, which is the convention the sibling assertions in this file already follow.

`go test ./internal/controller/metal3.io/...` is green here and `gofmt` is clean.

PR #3392, opened 2026-07-01
:bug: Fix common logging code and move to reusable location

dtantsur:
Follow-up to 903bb5cee08130958398fa79c23998ca15f0d07c:

Move the logging code to a common location so that we can use the same
constants in the provisioners and potentially API. Allow dotimports for
these to avoid bloating the code.

Fix incorrect guidance: things like "key decision points" and "important
branches" must be logged as INFO, this is implied by the words "key" and
"importants".

As an example, I'm adjusting (back) some of the changes log messages.
From my experience, these are critical to reconstruct the sequence of events.
I won't be able to fix anyone with their bugs if they don't have these on.

### DIFF
--- a/internal/controller/metal3.io/hostfirmwarecomponents_controller.go
+++ b/internal/controller/metal3.io/hostfirmwarecomponents_controller.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/go-logr/logr"
 	metal3api "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
+	. "github.com/metal3-io/baremetal-operator/pkg/logging"
 	"github.com/metal3-io/baremetal-operator/pkg/provisioner"
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
--- a/internal/controller/metal3.io/hostfirmwarecomponents_controller.go
+++ b/internal/controller/metal3.io/hostfirmwarecomponents_controller.go
@@ -255,11 +255,11 @@ func (r *HostFirmwareComponentsReconciler) updateEventHandler(e event.UpdateEven
 	// NOTE(dtantsur): the only realistic case of a changed owner reference is when pre-created HFC is adopted by a BMH that was created later.
 	// In this case, it's reasonable to reconcile the resource using the information from the new BMH.
 	if !reflect.DeepEqual(e.ObjectNew.GetOwnerReferences(), e.ObjectOld.GetOwnerReferences()) {
-		r.Log.Info("processing event for changed owner reference", "namespace", e.ObjectNew.GetNamespace(), "name", e.ObjectNew.GetName())
+		r.Log.V(VerbosityLevelDebug).Info("processing event for changed owner reference", "namespace", e.ObjectNew.GetNamespace(), "name", e.ObjectNew.GetName())
 		return true
 	}
 
-	r.Log.V(1).Info("ignoring event that did not change generation or owners", "namespace", e.ObjectNew.GetNamespace(), "name", e.ObjectNew.GetName())
+	r.Log.V(VerbosityLevelDebug).Info("ignoring event that did not change generation or owners", "namespace", e.ObjectNew.GetNamespace(), "name", e.ObjectNew.GetName())
 	return false
 }
 
@@ -281,7 +281,7 @@ func (r *HostFirmwareComponentsReconciler) publishEvent(ctx context.Context, req
 		"message", event.Message)
 	err := r.Create(ctx, &event)
 	if err != nil {
-		reqLogger.V(VerbosityLevelDebug).Info("failed to record event, ignoring",
+		reqLogger.Info("failed to record event, ignoring",
 			LogFieldReason, event.Reason,
 			"message", event.Message,
 			LogFieldError, err)
--- a/internal/controller/metal3.io/hostfirmwaresettings_controller.go
+++ b/internal/controller/metal3.io/hostfirmwaresettings_controller.go
@@ -229,7 +229,7 @@ func (r *HostFirmwareSettingsReconciler) updateStatus(ctx context.Context, info 
 	for k, v := range info.hfs.Spec.Settings {
 		if statusVal, ok := newStatus.Settings[k]; ok {
 			if v.String() != statusVal {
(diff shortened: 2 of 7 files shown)


kashifest (commented):
overall lgtm, 2 questions inline

kashifest:
> overall lgtm, 2 questions inline

wait where are the comments? 😄

internal/controller/metal3.io/hostfirmwarecomponents_controller.go
```diff
@@ -254,11 +255,11 @@ func (r *HostFirmwareComponentsReconciler) updateEventHandler(e event.UpdateEven
 	// NOTE(dtantsur): the only realistic case of a changed owner reference is when pre-created HFC is adopted by a BMH that was created later.
 	// In this case, it's reasonable to reconcile the resource using the information from the new BMH.
 	if !reflect.DeepEqual(e.ObjectNew.GetOwnerReferences(), e.ObjectOld.GetOwnerReferences()) {
-		r.Log.Info("processing event for changed owner reference", "namespace", e.ObjectNew.GetNamespace(), "name", e.ObjectNew.GetName())
+		r.Log.V(VerbosityLevelDebug).Info("processing event for changed owner reference", "namespace", e.ObjectNew.GetNamespace(), "name", e.ObjectNew.GetName())
```
kashifest, line 258: IMO change of owner reference should still go to .Info? wdyt?

internal/controller/metal3.io/hostfirmwarecomponents_controller.go
```diff
@@ -280,10 +281,9 @@ func (r *HostFirmwareComponentsReconciler) publishEvent(ctx context.Context, req
 		"message", event.Message)
 	err := r.Create(ctx, &event)
 	if err != nil {
-		reqLogger.V(VerbosityLevelDebug).Info("failed to record event, ignoring",
+		reqLogger.Error(err, "failed to record event, ignoring",
```
kashifest: so this we are changing from .Info to Error? The message still says, `ignoring`, what's the reasoning here?

internal/controller/metal3.io/hostfirmwarecomponents_controller.go
dtantsur: This is usually just BMO doing its job. The real use case here is creating HFS before BMH is created, in which case the HFS controller has to be awakened to find the freshly created host. Do you think it's something operators should be aware of?

internal/controller/metal3.io/hostfirmwarecomponents_controller.go
dtantsur: I felt like this being an unexpected error, it at least deserves highlighting in the logs at a much higher level. If you disagree, I can demote it back to INFO, but I'd rather not hide it entirely.

kashifest: No its fine and reasonable change I would say, shall we remove `ignoring` from the message?

internal/controller/metal3.io/hostfirmwarecomponents_controller.go
kashifest: ok, cant argue heavily on that. so the rule of thumb is log at Info() when the code does a host-visible state transition or action an operator may need to reason about, and keep it at debug when it’s just internal wake-up / ownership reconciliation like this?

internal/controller/metal3.io/hostfirmwarecomponents_controller.go
dtantsur: We're still ignoring it, as in: we don't fail the reconciliation. I'll move back to INFO since there are 3 more places at INFO level.

internal/controller/metal3.io/hostfirmwarecomponents_controller.go
dtantsur: >  the rule of thumb is log at Info() when the code does a host-visible state transition or action an operator may need to reason about

That's a good way to put it.

internal/controller/metal3.io/hostfirmwaresettings_controller.go
```diff
@@ -228,7 +229,7 @@ func (r *HostFirmwareSettingsReconciler) updateStatus(ctx context.Context, info
 	for k, v := range info.hfs.Spec.Settings {
 		if statusVal, ok := newStatus.Settings[k]; ok {
 			if v.String() != statusVal {
-				info.log.V(VerbosityLevelDebug).Info("spec value different than status",
+				info.log.Info("spec value different than status",
```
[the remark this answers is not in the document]
dtantsur: This is a fair point. The problem is that it's also the first place where we learn about the change in spec... I'll see if I can find a better place.

internal/controller/metal3.io/hostfirmwaresettings_controller.go
```diff
@@ -246,8 +257,16 @@ func (r *HostFirmwareSettingsReconciler) updateStatus(ctx context.Context, info
...
-	if specMismatch {
+	if len(specMismatch) > 0 {
 		if setCondition(generation, &newStatus, info, metal3api.FirmwareSettingsChangeDetected, metav1.ConditionTrue, reason, "") {
+			// This is the first time we detect a change, log the diff
+			diffParts := make([]string, 0, len(specMismatch))
+			for _, part := range specMismatch {
+				diffParts = append(diffParts,
+					fmt.Sprintf("%s: %q != %q", part.Name, part.Spec, ptr.Deref(part.Status, "<nil>")))
+			}
+			info.log.Info("current HostFirmwareSettings do not match the requested settings",
```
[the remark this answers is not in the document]
dtantsur: `newStatus` is already filtered for passwords, so this is only a problem if somebody adds a password to the spec for some reason. Which was equally a problem before my change.

Rozzii (commented):
I am fine with the log level adjustment.

But could you please explain what is the utility of moving the logging.go into a reusable package, I mean it has constants that are specific to BMO controllers, or are you planning to use these constants in the logging structure of the provisioner?

dtantsur:
> But could you please explain what is the utility of moving the logging.go into a reusable package, I mean it has constants that are specific to BMO controllers, or are you planning to use these constants in the logging structure of the provisioner?

Yep. From the user's perspective, the logs are the logs. We need to be consistent across the code base, not just in an arbitrary part of it.

Rozzii (commented):
/lgtm

PR #3396, opened 2026-07-04
🐛 Fix flaky Go installation in e2e setup

### LINKED ISSUE #2962: Failure to install go in e2e workflows
### Which jobs are flaking?

I think this affects all e2e jobs since they all install go the same way.
Here is one example of a failing job: https://github.com/metal3-io/baremetal-operator/actions/runs/21616532469

### Which tests are flaking?

Installation of go that happens before the tests actually start.

### Since when has it been flaking?

First occurrence I have seen was on 1st February: https://github.com/metal3-io/baremetal-operator/actions/runs/21557005916/job/62115095091

### Jenkins link

_No response_

There is some kind of issue downloading and extracting the archive. Probably it was corrupted during download.

```
+ /home/ubuntu/_work/baremetal-operator/baremetal-operator/hack/e2e/ensure_go.sh
+ MINIMUM_GO_VERSION=go1.24.12
+ verify_go_version
++ command -v go
+ '[' -x '' ']'
+ [[ linux-gnu == \l\i\n\u\x\-\g\n\u ]]
+ echo 'go not found, installing'
+ curl -sLo /tmp/go1.24.12.linux-amd64.tar.gz https://go.dev/dl/go1.24.12.linux-amd64.tar.gz
go not found, installing
+ sudo tar -C /usr/local -xzf /tmp/go1.24.12.linux-amd64.tar.gz

gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
Error: Process completed with exit code 2.
```

As a fix/workaround we could add a retry mechanism and file integrity check.

### Label(s) to be applied

/kind flake
One or more /area label. See https://github.com/metal3-io/baremetal-operator/labels for the list of labels.

Ahmedaltu:
Adds retry loop and checksum verification to the Go tarball download in `ensure_go.sh`. Previously a corrupted or truncated download was piped directly into `tar`, causing `gzip: stdin: not in gzip format` failures.

The fix downloads both the tarball and its published `.sha256`, verifies the checksum before extraction, and retries up to three
times on any failure.

Fixes #2962

### DIFF
--- a/hack/e2e/ensure_go.sh
+++ b/hack/e2e/ensure_go.sh
@@ -18,13 +18,20 @@ verify_go_version() {
         fi
         if [[ "${OSTYPE}" == "linux-gnu" ]]; then
             echo "go not found, installing"
+            local GO_TARBALL="${MINIMUM_GO_VERSION}.linux-amd64.tar.gz"
+            for attempt in 1 2 3; do
+                echo "Downloading Go (attempt ${attempt}/3)"
+                if curl -fsSL -o "/tmp/${GO_TARBALL}" "https://go.dev/dl/${GO_TARBALL}" && \
+                    curl -fsSL -o "/tmp/${GO_TARBALL}.sha256" "https://dl.google.com/go/${GO_TARBALL}.sha256" && \
+                    echo "$(cat "/tmp/${GO_TARBALL}.sha256")  /tmp/${GO_TARBALL}" | sha256sum --check --quiet; then
+                    break
+                fi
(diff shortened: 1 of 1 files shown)


Sunnatillo:
@Ahmedaltu Squash commits into one

Ahmedaltu:
Done!

peppi-lotta:
/retitle :seedling:  Fix flaky Go installation in e2e setup

Re titling to seeding because we don't want e2e test fixes under bug fixes in release notes.

PR #3406, opened 2026-07-06
:sparkles: Implementation of the HostClaim controller

pierrecregut:

The controller can be disabled if needed.
The structure of the controller is very similar to the code of the Metal3Machine controller.

As there is only one factory function, it is directly added to the reconciler state.
The missing core actions (update/delete) are stubs. The handling of pause (for pivot) is not integrated yet.

**Checklist:**

### DIFF
--- a/internal/controller/metal3.io/hostclaim_controller.go
+++ b/internal/controller/metal3.io/hostclaim_controller.go
@@ -37,10 +37,10 @@ import (
 // HostClaimReconciler reconciles a HostClaim object.
 type HostClaimReconciler struct {
 	client.Client
-	Scheme         *runtime.Scheme
-	Log            logr.Logger
-	APIReader      client.Reader
-	NewHostManager func(client client.Client, log logr.Logger, claim *metal3api.HostClaim, apireader client.Reader) hostclaimManager.HostManagerInterface
+	Scheme              *runtime.Scheme
+	Log                 logr.Logger
+	APIReader           client.Reader
+	NewHostClaimManager func(client client.Client, log logr.Logger, claim *metal3api.HostClaim, apireader client.Reader) hostclaimManager.ManagerInterface
 }
 
 //+kubebuilder:rbac:groups=metal3.io,resources=hostclaims,verbs=get;list;watch;create;update;patch;delete
@@ -73,39 +73,39 @@ func (r *HostClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	}()
 
 	// Create a helper for managing the baremetal container hosting the machine.
-	hostMgr := r.NewHostManager(r.Client, log, hostClaim, r.APIReader)
+	hostClaimMgr := r.NewHostClaimManager(r.Client, log, hostClaim, r.APIReader)
 
 	// Handle deleted machines
 	if !hostClaim.ObjectMeta.DeletionTimestamp.IsZero() {
-		return r.reconcileDelete(ctx, hostMgr)
+		return r.reconcileDelete(ctx, hostClaimMgr)
 	}
 
 	// Handle non-deleted machines
-	return r.reconcileNormal(ctx, hostMgr)
+	return r.reconcileNormal(ctx, hostClaimMgr)
 }
 
 func (r *HostClaimReconciler) reconcileNormal(ctx context.Context,
-	hostMgr hostclaimManager.HostManagerInterface,
+	hostClaimMgr hostclaimManager.ManagerInterface,
 ) (ctrl.Result, error) {
 	// If the HostClaim doesn't have finalizer, add it.
-	hostMgr.SetFinalizer()
+	hostClaimMgr.SetFinalizer()
 
 	// if the machine is already provisioned, update and return
-	if hostMgr.IsProvisioned() {
-		return checkHostClaimError(hostMgr.Update(ctx), "Failed to update the HostClaim")
+	if hostClaimMgr.IsProvisioned() {
+		return checkHostClaimError(hostClaimMgr.Update(ctx), "Failed to update the HostClaim")
 	}
 
 	// Check if the hostclaim was associated with a baremetalhost
-	if !hostMgr.IsAssociated() {
-		err := hostMgr.Associate(ctx)
+	if !hostClaimMgr.IsAssociated() {
+		err := hostClaimMgr.Associate(ctx)
 		if err != nil {
 			return checkHostClaimError(err, "failed to associate the HostClaim to a BaremetalHost")
 		}
 	} else {
 		// Update Condition to reflect that we have an associated BMH
-		hostMgr.SetConditionHostToTrue(metal3api.AssociatedCondition, metal3api.BareMetalHostAssociatedReason)
+		hostClaimMgr.SetConditionHostToTrue(metal3api.AssociatedCondition, metal3api.BareMetalHostAssociatedReason)
 	}
-	err := hostMgr.Update(ctx)
+	err := hostClaimMgr.Update(ctx)
 	if err != nil {
 		return checkHostClaimError(err, "failed to update BaremetalHost")
 	}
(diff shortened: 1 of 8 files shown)


pkg/hostclaim/hostclaim_manager.go
```diff
@@ -67,13 +68,48 @@ const (
...
 		Log:       log,
 		APIReader: apireader,
-	}, nil
+	}
+}
+
+// SetFinalizer sets finalizer on the host.
+func (m *HostManager) SetFinalizer() {
+	if !controllerutil.ContainsFinalizer(m.HostClaim, metal3api.HostClaimFinalizer) {
+		controllerutil.AddFinalizer(m.HostClaim, metal3api.HostClaimFinalizer)
```
zaneb, line 96: It's super confusing that HostManager doesn't manage a Host but a Claim.

dtantsur: I agree, let's plan on renaming it (can be a follow-up).

internal/controller/metal3.io/hostclaim_controller.go
```diff
@@ -18,32 +18,187 @@ package controllers
...
+	// if the machine is already provisioned, update and return
+	if hostMgr.IsProvisioned() {
+		return checkHostClaimError(hostMgr.Update(ctx), "Failed to update the Host")
+	}
+
+	// Check if the hostclaim was associated with a baremetalhost
+	if !hostMgr.IsAssociated() {
+		err := hostMgr.Associate(ctx)
+		if err != nil {
+			return checkHostClaimError(err, "failed to associate the Host to a BaremetalHost")
```
dtantsur: typo: s/Host/HostClaim/

internal/controller/metal3.io/hostclaim_controller.go
```diff
@@ -18,32 +18,187 @@ package controllers
...
+	if err != nil {
+		return checkHostClaimError(err, "failed to update BaremetalHost")
+	}
 	return ctrl.Result{}, nil
 }
 
+func (r *HostClaimReconciler) reconcileDelete(ctx context.Context,
+	hostMgr hostclaimManager.HostManagerInterface,
+) (ctrl.Result, error) {
+	// set machine condition to Deleting
```
dtantsur: s/machine/HostClaim/ (although this comment is quite obvious and can be dropped)

internal/controller/metal3.io/hostclaim_controller.go
```diff
@@ -18,32 +18,187 @@ package controllers
...
+	hostMgr.SetConditionHostToFalse(metal3api.AssociatedCondition,
+		metal3api.HostClaimDeletingReason,
+		"")
+
+	// delete the machine
```
dtantsur: same

internal/controller/metal3.io/hostclaim_controller.go
```diff
@@ -18,32 +18,187 @@ package controllers
...
+	return ctrl.Result{}, nil
+}
+
+// patchHostClaim patch the HostClaim and ensures that the conditions are initialized.
+// Can be used several times without creating a conflict.
+func patchHostClaim(ctx context.Context, patchHelper *patch.Helper, claim *metal3api.HostClaim, options ...patch.Option) error {
+	// Always update the readyCondition by summarizing the state of other conditions.
+	sumOption := conditions.ForConditionTypes{
+		metal3api.AssociatedCondition, metal3api.SynchronizedCondition, metal3api.ProvisionedCondition}
+	if err := conditions.SetSummaryCondition(claim, claim, clusterv1.ReadyCondition, sumOption); err != nil {
```
dtantsur: Let's not use clusterv1 constants even if the value is the same

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -67,13 +68,48 @@ const (
 // An error used when there is no BMH satisfying the constraints.
 var ErrNoAvailableBMH = errors.New("no available BareMetalHost")
 
-func NewHostManager(client client.Client, log logr.Logger, claim *metal3api.HostClaim, apireader client.Reader) (*HostManager, error) {
+type HostManagerInterface interface {
```
dtantsur: nit: add short descriptions to the methods

internal/controller/metal3.io/hostclaim_controller.go
```diff
@@ -18,32 +18,187 @@ package controllers
...
+	hostMgr.UnsetFinalizer()
+
+	return ctrl.Result{}, nil
+}
+
+// patchHostClaim patch the HostClaim and ensures that the conditions are initialized.
+// Can be used several times without creating a conflict.
+func patchHostClaim(ctx context.Context, patchHelper *patch.Helper, claim *metal3api.HostClaim, options ...patch.Option) error {
+	// Always update the readyCondition by summarizing the state of other conditions.
+	sumOption := conditions.ForConditionTypes{
```
dtantsur, line 141: nit: I find it somewhat odd that such a trivial thing as `controllerutil.RemoveFinalizer(m.HostClaim, metal3api.HostClaimFinalizer)` is in HostClaimManager, but this condition logic stays in the controller.

internal/controller/metal3.io/hostclaim_controller.go
```diff
@@ -18,32 +18,187 @@ package controllers
...
+	if err := hostMgr.Delete(ctx); err != nil {
+		hostMgr.SetConditionHostToFalse(metal3api.AssociatedCondition,
+			metal3api.HostClaimDeletionFailedReason,
+			err.Error())
+		return checkHostClaimError(err, "failed to delete Host")
+	}
+
+	// hostclaim is marked for deletion and ready to be deleted,
+	// so remove the finalizer.
+	hostMgr.UnsetFinalizer()
```
dtantsur: Why not make this a part of `hostMgr.Delete` as well?

pkg/hostclaim/hostclaim_manager.go
```diff
@@ -157,6 +193,14 @@ func (m *HostManager) Associate(ctx context.Context) error {
 	return nil
 }
 
+func (m *HostManager) Delete(_ context.Context) error {
+	return nil
+}
```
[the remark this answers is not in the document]
dtantsur: This seems to be a valid concern, at least from a quick read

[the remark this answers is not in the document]
pierrecregut: Either we build the functions first and implement the controller last or the reverse.  Delete is supposed to be a commit on its own.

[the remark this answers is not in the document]
dtantsur: Won't the AI suggestion just work?

I don't mind a separate commit too much, but it's indeed a bit of an inconsistency that Associate is already implemented and Delete is not.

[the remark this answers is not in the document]
pierrecregut: It will  be rewritten (shared getBmh,  fields to clean after update, etc.)

internal/controller/metal3.io/hostclaim_controller.go
pierrecregut: This was metal3machine choice and I think they were right. You look at the structure and SetFinalizer and UnsetFinalizer are clearly visible at the right place. This would be harder to check with unset in Delete.

dtantsur:
/lgtm

Please do consider following up with the nits.

The fixture failure is something I've just seen on another patch. I haven't dug into it.

internal/controller/metal3.io/hostclaim_controller.go
pierrecregut: If we push it to the manager, we will get the hostClaim from the manager but the patchHelper will probably stay in the controller because it is outside the defer. I am not sure it would help readability and it is probably why it was done this way in metal3machine.

pkg/hostclaim/hostclaim_manager.go
pierrecregut: I have added a commit to be squashed that addresses the renaming. Linter rightfully mentioned that as we are in a hostclaim package (not the case with metal3 machines for example), the manager can be named Manager rather than HostClaimManager.  So HostManager and HostManagerInterface become Manager and ManagerInterface. Variable hostMgr becomes hostClaimMgr in the controller and factory field is NewHostClaimManager.

dtantsur:
/lgtm

Thanks for the update!

/cc @Rozzii @elfosardo

elfosardo:
/approve
also backing up Dmitry's request for a follow up

PR #3407, opened 2026-07-06
:seedling: End to end tests for HostClaims

pierrecregut:

End to end tests for HostClaims
    
First end to end test for hostclaim.
Creates a BMH and a security policy authorizing the right namespace.
Associates the hostclaim in that namespace.
The test will be amended when other core life-cycle functions are  implemented.

**Checklist:**

### DIFF
--- a/config/overlays/e2e/cluster-role.yaml
+++ b/config/overlays/e2e/cluster-role.yaml
@@ -1,0 +1,27 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: baremetal-operator-manager-clusterrole
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - namespaces
+  verbs:
+  - get
+  - list
+  - watch
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: baremetal-operator-manager-clusterrolebinding
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: baremetal-operator-manager-clusterrole
+subjects:
+- kind: ServiceAccount
+  name: baremetal-operator-controller-manager
+  namespace: baremetal-operator-system
+---
--- a/hack/check-e2e.sh
+++ b/hack/check-e2e.sh
@@ -6,7 +6,8 @@ REPO_ROOT=$(realpath "$(dirname "${BASH_SOURCE[0]}")/..")
 cd "${REPO_ROOT}" || exit 1
 
 WATCH_NS="$(grep -A1 WATCH_NAMESPACE config/overlays/e2e/namespaced-manager-patch.yaml | grep 'value:' | awk '{ print $2; }')"
-EXCEPTIONS=upgrade
+# hostclaims requires two namespaces hostclaims-tenant for hostclaim and hostclaims-infra for bmh.
+EXCEPTIONS=upgrade,hostclaims
 
 EXITCODE=0
 for spec in $(grep -E --no-filename ' *specName +:?=' test/e2e/*_test.go | grep -o '".*"' | tr -d '"'); do
--- a/test/e2e/common.go
+++ b/test/e2e/common.go
@@ -1196,3 +1196,19 @@ func ContainCondition(conditionType string, conditionStatus metav1.ConditionStat
 		),
 	)
 }
+
+type WaitForHostClaimConditionInput struct {
+	Client        client.Client
+	HostClaim     *metal3api.HostClaim
+	ConditionType string
+	Status        metav1.ConditionStatus
+}
+
+func WaitForHostClaimCondition(ctx context.Context, input WaitForHostClaimConditionInput, intervals ...interface{}) {
+	Eventually(func(g Gomega) {
+		hostclaim := metal3api.HostClaim{}
+		key := types.NamespacedName{Namespace: input.HostClaim.Namespace, Name: input.HostClaim.Name}
+		g.Expect(input.Client.Get(ctx, key, &hostclaim)).To(Succeed())
+		g.Expect(hostclaim.Status.Conditions).To(ContainCondition(input.ConditionType, input.Status))
+	}, intervals...).Should(Succeed())
+}
--- a/test/e2e/common.go
+++ b/test/e2e/common.go
@@ -1189,3 +1189,19 @@ func ContainCondition(conditionType string, conditionStatus metav1.ConditionStat
 		),
 	)
 }
+
+type WaitForHostClaimConditionInput struct {
+	Client        client.Client
+	HostClaim     *metal3api.HostClaim
+	ConditionType string
+	Status        metav1.ConditionStatus
+}
+
+func WaitForHostClaimCondition(ctx context.Context, input WaitForHostClaimConditionInput, intervals ...interface{}) {
+	Eventually(func(g Gomega) {
+		hostclaim := metal3api.HostClaim{}
+		key := types.NamespacedName{Namespace: input.HostClaim.Namespace, Name: input.HostClaim.Name}
+		g.Expect(input.Client.Get(ctx, key, &hostclaim)).To(Succeed())
+		g.Expect(hostclaim.Status.Conditions).To(ContainCondition(input.ConditionType, input.Status))
+	}, intervals...).Should(Succeed())
+}
--- a/test/e2e/config/ironic.yaml
+++ b/test/e2e/config/ironic.yaml
@@ -92,6 +92,8 @@ intervals:
   default/wait-reconcile: [ "1s", "10ms" ]
   baremetalswitch/wait-secret-updated: [ "30s", "500ms" ]
   default/wait-user-data: [ "5s", "1s" ]
+  hostclaims/wait-associated: [ "10s", "1s" ]
+  hostclaims/wait-deleted: [ "20s", "10ms" ]
 
 kindExtraPortMappings:
 # Expose Ironic ports so they are reachable outside of kind
--- a/test/e2e/config/ironic.yaml
+++ b/test/e2e/config/ironic.yaml
@@ -96,6 +96,8 @@ intervals:
   default/wait-reconcile: [ "1s", "10ms" ]
   baremetalswitch/wait-secret-updated: [ "30s", "500ms" ]
   default/wait-user-data: [ "5s", "1s" ]
+  hostclaims/wait-associated: [ "10s", "1s" ]
+  hostclaims/wait-deleted: [ "20s", "10ms" ]
 
 kindExtraPortMappings:
 # Expose Ironic ports so they are reachable outside of kind
--- a/test/e2e/hostclaim_provisioning_test.go
+++ b/test/e2e/hostclaim_provisioning_test.go
@@ -1,0 +1,198 @@
+//go:build e2e
+// +build e2e
+
+package e2e
+
+import (
+	"context"
+	"path"
+
+	metal3api "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/cluster-api/test/framework"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	selectorLabel = "test.metal3.io/class"
+	selectorValue = "selected"
+)
+
+func createBmh(name, namespace string, labels map[string]string, hardwareDetails string) *metal3api.BareMetalHost {
+	bmh := &metal3api.BareMetalHost{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+			Labels:    labels,
+			Annotations: map[string]string{
+				metal3api.HardwareDetailsAnnotation: hardwareDetails,
+			},
+		},
+		Spec: metal3api.BareMetalHostSpec{
+			Online: false,
+			BMC: metal3api.BMCDetails{
+				Address:                        bmc.Address,
+				CredentialsName:                "bmc-credentials",
+				DisableCertificateVerification: bmc.DisableCertificateVerification,
+			},
+			BootMode:              metal3api.BootMode(e2eConfig.GetVariable("BOOT_MODE")),
+			BootMACAddress:        bmc.BootMacAddress,
+			AutomatedCleaningMode: metal3api.CleaningModeDisabled,
+			InspectionMode:        metal3api.InspectionModeDisabled,
+			RootDeviceHints:       &bmc.RootDeviceHints,
+		},
+	}
+	return bmh
+}
+
+func createHostclaim(namespace string) *metal3api.HostClaim {
+	hostClaim := &metal3api.HostClaim{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "claim",
+			Namespace: namespace,
+		},
+		Spec: metal3api.HostClaimSpec{
+			Image: &metal3api.Image{
+				URL:      e2eConfig.GetVariable("IMAGE_URL"),
+				Checksum: e2eConfig.GetVariable("IMAGE_CHECKSUM"),
+			},
+			HostSelector: metal3api.HostSelector{
+				MatchLabels: map[string]string{selectorLabel: selectorValue},
+			},
+			PoweredOn: true,
+		},
+	}
+	err := clusterProxy.GetClient().Create(ctx, hostClaim)
+	Expect(err).NotTo(HaveOccurred())
+	return hostClaim
+}
+
+func createHostDeployPolicy(namespaceBMH, namespaceClaim string) *metal3api.HostDeployPolicy {
+	hostDeployPolicy := &metal3api.HostDeployPolicy{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "hdp",
+			Namespace: namespaceBMH,
+		},
+		Spec: metal3api.HostDeployPolicySpec{
+			HostClaimNamespaces: &metal3api.HostClaimNamespaces{
+				Names: []string{namespaceClaim},
+			},
+		},
+	}
+	err := clusterProxy.GetClient().Create(ctx, hostDeployPolicy)
+	Expect(err).NotTo(HaveOccurred())
+	return hostDeployPolicy
+}
+
+func checkAssociation(specName string, hostClaim *metal3api.HostClaim, bmh *metal3api.BareMetalHost) {
+	WaitForHostClaimCondition(ctx, WaitForHostClaimConditionInput{
+		Client:        clusterProxy.GetClient(),
+		HostClaim:     hostClaim,
+		ConditionType: metal3api.AssociatedCondition,
+		Status:        metav1.ConditionTrue,
+	}, e2eConfig.GetIntervals(specName, "wait-associated")...)
+	currentHC := metal3api.HostClaim{}
+	key := client.ObjectKeyFromObject(hostClaim)
+	Expect(clusterProxy.GetClient().Get(ctx, key, &currentHC)).To(Succeed())
+	Expect(currentHC.Status.BareMetalHost).NotTo(BeNil())
+	Expect(currentHC.Status.BareMetalHost.Name).To(Equal(bmh.Name))
+	Expect(currentHC.Status.BareMetalHost.Namespace).To(Equal(bmh.Namespace))
+
+	key = client.ObjectKeyFromObject(bmh)
+	Expect(clusterProxy.GetClient().Get(ctx, key, bmh)).To(Succeed())
+	Expect(bmh.Spec.ConsumerRef).NotTo(BeNil())
+	Expect(bmh.Spec.ConsumerRef.Name).To(Equal(hostClaim.Name))
+	Expect(bmh.Spec.ConsumerRef.Namespace).To(Equal(hostClaim.Namespace))
+}
+
+func checkNoConsumer(specName string, bmh *metal3api.BareMetalHost) {
+	Eventually(func(g Gomega) {
+		currentBmh := metal3api.BareMetalHost{}
+		key := types.NamespacedName{Namespace: bmh.Namespace, Name: bmh.Name}
+		g.Expect(clusterProxy.GetClient().Get(ctx, key, &currentBmh)).To(Succeed())
+		g.Expect(currentBmh.Spec.ConsumerRef).To(BeNil())
+	}, e2eConfig.GetIntervals(specName, "wait-deleted")...).Should(Succeed())
+}
+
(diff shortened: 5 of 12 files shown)


pierrecregut:
@dtantsur This will follow #3406 to implement a basic e2e test scenario.

config/overlays/e2e/multi-bmh/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- ../roles-rolebindings
+- namespace.yaml
+
+namespace: multi-bmh
```
dtantsur: Took me a second to understand why two namespaces. I wonder if calling them something like `hostclaims-infra` and `hostclaims-tenant` would be clearer.

hack/check-e2e.sh
```diff
@@ -6,7 +6,7 @@ REPO_ROOT=$(realpath "$(dirname "${BASH_SOURCE[0]}")/..")
 cd "${REPO_ROOT}" || exit 1
 
 WATCH_NS="$(grep -A1 WATCH_NAMESPACE config/overlays/e2e/namespaced-manager-patch.yaml | grep 'value:' | awk '{ print $2; }')"
-EXCEPTIONS=upgrade
+EXCEPTIONS=upgrade,multi
```
dtantsur: Let's leave a comment why `multi` needs to be an exception.

test/e2e/config/ironic.yaml
```diff
@@ -91,6 +91,7 @@ intervals:
   default/wait-reconcile: [ "1s", "10ms" ]
   baremetalswitch/wait-secret-updated: [ "30s", "500ms" ]
   default/wait-user-data: [ "5s", "1s" ]
+  multi/wait-associated: [ "2m", "1s" ]
```
dtantsur: 2m timeout seems to be an overkill, shouldn't it be more like seconds?

test/e2e/common.go
```diff
@@ -1189,3 +1190,21 @@ func ContainCondition(conditionType string, conditionStatus metav1.ConditionStat
...
+	ConditionType string
+	Status        metav1.ConditionStatus
+}
+
+func WaitForHostClaimCondition(ctx context.Context, input WaitForHostClaimConditionInput, intervals ...interface{}) {
+	Eventually(func(g Gomega) {
+		hostclaim := metal3api.HostClaim{}
+		key := types.NamespacedName{Namespace: input.HostClaim.Namespace, Name: input.HostClaim.Name}
+		g.Expect(input.Client.Get(ctx, key, &hostclaim)).To(Succeed())
+		condition := conditions.Get(&hostclaim, input.ConditionType)
```
dtantsur: nit: we also have a `ContainCondition` matcher

test/e2e/hostclaim_provisioning_test.go
```diff
@@ -0,0 +1,201 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func createBmh(name, namespace string, labels map[string]string, hardwareDetails string) *metal3api.BareMetalHost {
+	bmh := &metal3api.BareMetalHost{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+			Annotations: map[string]string{
+				metal3api.InspectAnnotationPrefix:   "disabled",
```
dtantsur: nit: use `InspectionMode` instead

test/e2e/hostclaim_provisioning_test.go
```diff
@@ -0,0 +1,201 @@
...
+				metal3api.HardwareDetailsAnnotation: hardwareDetails,
```
dtantsur, line 32: If you use this, you shouldn't create HardwareData, otherwise the controller will race with your code

test/e2e/hostclaim_provisioning_test.go
```diff
@@ -0,0 +1,201 @@
...
+		Spec: metal3api.BareMetalHostSpec{
+			Online: false,
+			BMC: metal3api.BMCDetails{
+				Address:                        bmc.Address,
+				CredentialsName:                "bmc-credentials",
+				DisableCertificateVerification: bmc.DisableCertificateVerification,
+			},
+			BootMode:              metal3api.Legacy,
+			BootMACAddress:        bmc.BootMacAddress,
+			AutomatedCleaningMode: "disabled",
```
dtantsur: nit: use constant from metal3api

test/e2e/hostclaim_provisioning_test.go
```diff
@@ -0,0 +1,201 @@
...
+			Labels: labels,
+		},
+		Spec: metal3api.BareMetalHostSpec{
+			Online: false,
+			BMC: metal3api.BMCDetails{
+				Address:                        bmc.Address,
+				CredentialsName:                "bmc-credentials",
+				DisableCertificateVerification: bmc.DisableCertificateVerification,
+			},
+			BootMode:              metal3api.Legacy,
```
dtantsur: Should be `metal3api.BootMode(e2eConfig.GetVariable("BOOT_MODE"))`

test/e2e/hostclaim_provisioning_test.go
```diff
@@ -0,0 +1,201 @@
...
+			Expect(err).NotTo(HaveOccurred())
+			toCleanupBMH = append(toCleanupBMH, hwData)
+
+			By("Waiting for the BMH to become available")
+			WaitForBmhInProvisioningState(ctx, WaitForBmhInProvisioningStateInput{
+				Client: clusterProxy.GetClient(),
+				Bmh:    *bmh,
+				State:  metal3api.StateAvailable,
+			}, e2eConfig.GetIntervals(specName, "wait-available")...)
+			By("Creating a suitable hostDeployPolicy")
```
dtantsur, line 173: nit: space before By for readability

test/e2e/hostclaim_provisioning_test.go
```diff
@@ -0,0 +1,201 @@
...
+			By("Creating a secret with BMH credentials")
+			bmcCredentialsData := map[string]string{
+				"username": bmc.User,
+				"password": bmc.Password,
+			}
+			CreateSecret(ctx, clusterProxy.GetClient(), namespaceBMH.Name, secretName, bmcCredentialsData)
+
+			By("Creating a BMH with inspection disabled and hardware details added")
+			hardwareDetails := hardwareDetailsFor(&bmc)
+			bmh := createBmh("bmh", namespaceBMH.Name, map[string]string{"test.meta3.io/class": "selected"}, hardwareDetails)
```
dtantsur: Typo on the annotation (`meta3`). Actually, let's have a constant for it.

dtantsur:
Great start, some early comments posted.

config/overlays/e2e/multi-bmh/kustomization.yaml
pierrecregut: yes much better

test/e2e/config/ironic.yaml
pierrecregut: timeout diminished. The BMH is ready so there is no risk of waiting for it.

test/e2e/hostclaim_provisioning_test.go
pierrecregut: done

test/e2e/hostclaim_provisioning_test.go
pierrecregut: done.

pierrecregut:
Added a delete of the hostclaim and a check that bmh consumerRef is back to nil.

test/e2e/hostclaim_provisioning_test.go
dtantsur: Sorry, by "space" I meant "new line"

dtantsur:
Are you sure you've pushed a new revision? I don't see any changes.

pierrecregut:
My bad, pushed it to the wrong remote.

test/e2e/config/ironic.yaml
```diff
@@ -91,6 +91,7 @@ intervals:
   default/wait-reconcile: [ "1s", "10ms" ]
   baremetalswitch/wait-secret-updated: [ "30s", "500ms" ]
   default/wait-user-data: [ "5s", "1s" ]
+  hostclaims/wait-associated: [ "10s", "1s" ]
```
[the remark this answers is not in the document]
pierrecregut: Good point but it is probably default/wait-deleting in fixture.yaml that should be removed as I see no use for it.

pierrecregut:
Just added "ironic" label to the test.

pierrecregut:
tag removed: see discussion #3439

test/e2e/hostclaim_provisioning_test.go
```diff
@@ -0,0 +1,165 @@
...
+			specName       = "hostclaims"
+			secretName     = "bmc-credentials"
+			namespaceBMH   *corev1.Namespace
+			namespaceClaim *corev1.Namespace
+			cancelWatches  context.CancelFunc
+			toCleanupBMH   []client.Object
+			toCleanupClaim []client.Object
+		)
+
+		BeforeEach(func() {
```
elfosardo, line 133: we don't have a cleanup reset here as we do in other tests, it's harmless as we have  asingle It block, but maybe worth adding a comment at leats in case we add more ?

test/e2e/hostclaim_provisioning_test.go
```diff
@@ -0,0 +1,165 @@
...
+
+			By("Deleting the hostclaim")
+			err = clusterProxy.GetClient().Delete(ctx, hostClaim)
+			Expect(err).NotTo(HaveOccurred())
+			Eventually(func(g Gomega) {
+				currentBmh := metal3api.BareMetalHost{}
+				key := types.NamespacedName{Namespace: bmh.Namespace, Name: bmh.Name}
+				g.Expect(clusterProxy.GetClient().Get(ctx, key, &currentBmh)).To(Succeed())
+				g.Expect(currentBmh.Spec.ConsumerRef).To(BeNil())
+			}, e2eConfig.GetIntervals(specName, "wait-deleted")...).Should(Succeed())
```
elfosardo: wait-delete falls back to the default value here as we don't have a specific entry for hotsclaims, is this intended?

test/e2e/hostclaim_provisioning_test.go
pierrecregut: Good point as we will probably need more It block in the future.

test/e2e/hostclaim_provisioning_test.go
pierrecregut: I have put a hostclaims/wait-deleted so that the default delay can be changed without impact on hostclaims.

pierrecregut:
rebased (namespaced-manager-patch.yaml: added dataimage e2e test created conflict)

test/e2e/hostclaim_provisioning_test.go
```diff
@@ -0,0 +1,167 @@
...
+			toCleanupBMH = append(toCleanupBMH, bmh)
+
+			By("Waiting for the BMH to become available")
+			WaitForBmhInProvisioningState(ctx, WaitForBmhInProvisioningStateInput{
+				Client: clusterProxy.GetClient(),
+				Bmh:    *bmh,
+				State:  metal3api.StateAvailable,
+			}, e2eConfig.GetIntervals(specName, "wait-available")...)
+
+			By("Creating a suitable hostDeployPolicy")
```
tuminoid, line 173: We need negative test case too, deny-path / isolation, for better coverage.

test/e2e/hostclaim_provisioning_test.go
```diff
@@ -0,0 +1,167 @@
...
+			toCleanupBMH = append(toCleanupBMH, hostDeployPolicy)
+
+			By("Creating a hostClaim")
+			hostClaim := &metal3api.HostClaim{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      specName,
+					Namespace: namespaceClaim.Name,
+				},
+				Spec: metal3api.HostClaimSpec{
+					Image: &metal3api.Image{
```
tuminoid: HostManager.Update is no-op, so Image and PoweredOn never reach BMH and arent' asserted?

test/e2e/hostclaim_provisioning_test.go
```diff
@@ -0,0 +1,167 @@
...
+			BootMACAddress:        bmc.BootMacAddress,
+			AutomatedCleaningMode: metal3api.CleaningModeDisabled,
+			InspectionMode:        metal3api.InspectionModeDisabled,
+			RootDeviceHints:       &bmc.RootDeviceHints,
+		},
+	}
+	return bmh
+}
+
+var _ = Describe("Associate a hostclaim to a BMH and delete the claim.", Label("required", "provision", "hostclaim"),
```
tuminoid: This doesn't actually provision anything, so label is wrong? Description says associate+delete only.

tuminoid (commented):
Good work, I see no major issues. Just some nits in test cases.

test/e2e/hostclaim_provisioning_test.go
pierrecregut: Yes I will remove it for now and I will put it back when update is implemented.

test/e2e/hostclaim_provisioning_test.go
pierrecregut: With just two namespaces, we can only prevent the hostclaim from associating and there is no visible result. May be we should have a test with several namespaces for both hostclaims and bmh to check that they associate correctly (but this is not a proof as it could happen by chance). It could be a separate test with no provisioning.

test/e2e/hostclaim_provisioning_test.go
pierrecregut: Yes: we decided at some point that it was better to have e2e test early on an incomplete controller so that we can check the integration of what we have so far.

test/e2e/hostclaim_provisioning_test.go
dtantsur: > With just two namespaces, we can only prevent the hostclaim from associating and there is no visible result

I think it's still useful to do: this is a fundamental security feature.

test/e2e/hostclaim_provisioning_test.go
```diff
@@ -0,0 +1,167 @@
...
+			})
+		})
+
+		It("Create a claim, associate it to a BMH", func() {
+			By("Creating a secret with BMH credentials")
+			bmcCredentialsData := map[string]string{
+				"username": bmc.User,
+				"password": bmc.Password,
+			}
+			CreateSecret(ctx, clusterProxy.GetClient(), namespaceBMH.Name, secretName, bmcCredentialsData)
```
dtantsur: nit: add the resulting secret to toCleanup to guarantee its deletion

test/e2e/hostclaim_provisioning_test.go
```diff
@@ -0,0 +1,167 @@
...
+			toCleanupClaim = append(toCleanupClaim, hostClaim)
+
+			By("Waiting for the HostClaim to become associated")
+			WaitForHostClaimCondition(ctx, WaitForHostClaimConditionInput{
+				Client:        clusterProxy.GetClient(),
+				HostClaim:     hostClaim,
+				ConditionType: metal3api.AssociatedCondition,
+				Status:        metav1.ConditionTrue,
+			}, e2eConfig.GetIntervals(specName, "wait-associated")...)
+
```
dtantsur, line 183: Let's fetch the BMH and make sure it gets the right ConsumerRef. Let's also check that the link from the HostClaim is correct.

test/e2e/hostclaim_provisioning_test.go
```diff
@@ -0,0 +1,167 @@
...
+			By("Deleting the hostclaim")
+			err = clusterProxy.GetClient().Delete(ctx, hostClaim)
+			Expect(err).NotTo(HaveOccurred())
+			Eventually(func(g Gomega) {
+				currentBmh := metal3api.BareMetalHost{}
+				key := types.NamespacedName{Namespace: bmh.Namespace, Name: bmh.Name}
+				g.Expect(clusterProxy.GetClient().Get(ctx, key, &currentBmh)).To(Succeed())
+				g.Expect(currentBmh.Spec.ConsumerRef).To(BeNil())
+			}, e2eConfig.GetIntervals(specName, "wait-deleted")...).Should(Succeed())
+		})
```
dtantsur, line 188: Let's make sure that ConsumerRef is deleted from the BMH.

test/e2e/hostclaim_provisioning_test.go
dtantsur: Another important thing to test: create a HostClaim with a forged BMH link in the status, make sure that BMH does not get a ConsumerRef (in the future also Image and UserData), and that the HostClaim reports an error.

config/overlays/e2e/cluster-role.yaml
dtantsur, line 1: I'm curious why this role hasn't been needed so far. What does it do even?

test/e2e/hostclaim_provisioning_test.go
```diff
@@ -0,0 +1,243 @@
...
+	Expect(bmh.Spec.ConsumerRef.Namespace).To(Equal(hostClaim.Namespace))
+}
+
+func checkNoConsumer(specName string, bmh *metal3api.BareMetalHost) {
+	Eventually(func(g Gomega) {
+		currentBmh := metal3api.BareMetalHost{}
+		key := types.NamespacedName{Namespace: bmh.Namespace, Name: bmh.Name}
+		g.Expect(clusterProxy.GetClient().Get(ctx, key, &currentBmh)).To(Succeed())
+		g.Expect(currentBmh.Spec.ConsumerRef).To(BeNil())
+	}, e2eConfig.GetIntervals(specName, "wait-deleSted")...).Should(Succeed())
```
dtantsur: typo: wait-deleted?

test/e2e/hostclaim_provisioning_test.go
```diff
@@ -0,0 +1,243 @@
...
+
+			By("Waiting for the HostClaim to become associated")
+			checkAssociation(specName, hostClaim, bmh)
+
+			By("Deleting the hostclaim")
+			err = clusterProxy.GetClient().Delete(ctx, hostClaim)
+			Expect(err).NotTo(HaveOccurred())
+			checkNoConsumer(specName, bmh)
+
+			By("Creating a bad claim")
```
dtantsur: nit: "Creating a claim with forged BMH link"

test/e2e/hostclaim_provisioning_test.go
```diff
@@ -0,0 +1,243 @@
...
+					},
+				},
+			}
+			Expect(clusterProxy.GetClient().Create(ctx, hostClaim)).To(Succeed())
+
+			By("Wait for association to fail and BareMetalHost reference reset to nil")
+			WaitForHostClaimCondition(ctx, WaitForHostClaimConditionInput{
+				Client:        clusterProxy.GetClient(),
+				HostClaim:     hostClaim,
+				ConditionType: metal3api.AssociatedCondition,
```
dtantsur: Is there an error condition that we can wait for? Associated should be false initially.

config/overlays/e2e/cluster-role.yaml
pierrecregut: This is linked with HostDeployPolicy and the selection of namespaces based on their labels.

test/e2e/hostclaim_provisioning_test.go
pierrecregut: I have a forthcoming test (rather for a separate PR/commit) that creates two tenant namespaces and one infra (bmh). I create hostclaims with some delay and the first one is not authorized by the policy. I check that the bound is always with the second. This is one of the reasons I extracted some of the code of the test in separate functions so that it can be reused.

test/e2e/hostclaim_provisioning_test.go
pierrecregut: It is rather undefined. But no there is no real condition we can wait for. The controller removes incorrect links silently whether they are a result of malice from the tenant or bad reconfiguration of the bmh.

test/e2e/hostclaim_provisioning_test.go
pierrecregut: Is the check at line 155 above wrong ?

test/e2e/hostclaim_provisioning_test.go
pierrecregut: The test to push in a later commit is here : https://github.com/metal3-io/baremetal-operator/commit/fcc6f7bd2e48db87a8470a78fc6d12b6f9253652

test/e2e/hostclaim_provisioning_test.go
pierrecregut: done in checkAssociation

pierrecregut:
Fixed a bug: the bad hostclaim manifest has a status set. It will not be pushed with just a create. The status must be updated explicitly.

test/e2e/hostclaim_provisioning_test.go
pierrecregut: > Another important thing to test: create a HostClaim with a forged BMH link in the status, make sure that BMH does not get a ConsumerRef (in the future also Image and UserData), and that the HostClaim reports an error.

@dtantsur I have removed this part. When the logic is right (patched Status), ... nothing happens and this is the expected result so far because the role of Associate is not to check the association only build it. It is the next step Update that performs a getBmh, find something fishy and remove the baremetalhost from the status. 

But Update is not there yet. So the test can only fail: nothing bad happens like the consumerRef of the bmh being updated, but nothing else either. I will put it back as part of the updated "Update" commit.

As a side note, with associated baremetalhost stored in status and not as a label, a well configured rbac should prevent regular user from updating the status of their claims.

dtantsur: Let's not forget to add it please. With RBAC or not, it's a critical security boundary.

pierrecregut: As HostDeployPolicy tests will be in a separate commit and Forged status  will be with the Update code, I don't think there is anything pending left.

elfosardo:
/approve

it looks ok to me, I see only one oustanding question from @dtantsur still open

test/e2e/hostclaim_provisioning_test.go
pierrecregut: As the code has evolved and the lines were moved, to make the above comment clear, this is performed by checkNoConsumer.

test/e2e/hostclaim_provisioning_test.go
pierrecregut: implicitly resolved for this commit because of https://github.com/metal3-io/baremetal-operator/pull/3407#discussion_r3925416721 above.

config/overlays/e2e/cluster-role.yaml
dtantsur: Don't we need the same role for main BMO, outside of e2e tests?

dtantsur:
/lgtm

The question remains but can be addressed separately (if needed)

config/overlays/e2e/cluster-role.yaml
pierrecregut: Yes : https://github.com/metal3-io/baremetal-operator/blob/830814242f3a8969f5cd761860f5cbeb8b53a0bc/internal/controller/metal3.io/hostclaim_controller.go#L50 . But it is the only rule that is not converted to a role.

Issue #2626, opened 2025-08-22
Expand E2E suite with metadata test


lentzi90:
We lack tests for the BareMetalHost userData, networkData and metaData functionality.
This issue is for implementing such a test.

One basic thing we should check is that the content actually appears on the disk as it should.
A test could look something like this:

1. Create a secret with some metadata (userdata and networkdata can be trickier since they would change things on the host, metadata is not expected to do something on its own).
2. Provision the BMH with some image + the metadata (and perhaps some userdata and networkdata to be able to access it).
3. Access the host and look for the config drive where the metadata should be.
It can be seen with the `blkid` command. Look for `LABEL="config-2"`.
4. Mount the config drive and verify that the metadata is the same as that in the secret.

Example commands and output:

```console
ubuntu@ubuntu-rqqzc-z9kb7:~$ sudo blkid
/dev/nvme0n1p16: LABEL="BOOT" UUID="d247b12f-2037-462d-8751-fddf35d29b81" BLOCK_SIZE="4096" TYPE="ext4" PARTUUID="261a1bdf-6568-4711-bbbb-13dff71487ee"
/dev/nvme0n1p1: LABEL="cloudimg-rootfs" UUID="ee4654d0-951c-4342-ab5e-e8f80c71b13f" BLOCK_SIZE="4096" TYPE="ext4" PARTUUID="b7541bc4-4492-439d-a2d5-886a7f6f202e"
/dev/nvme0n1p2: BLOCK_SIZE="2048" UUID="2025-08-20-12-53-13-00" LABEL="config-2" TYPE="iso9660" PARTUUID="be4b443c-6567-45c7-a492-7d10182e5f08"
/dev/nvme0n1p15: LABEL_FATBOOT="UEFI" LABEL="UEFI" UUID="0002-E691" BLOCK_SIZE="512" TYPE="vfat" PARTUUID="c7c38019-259f-4cea-b01d-38ddfd5f8a68"
/dev/nvme0n1p14: PARTUUID="02f0a446-fcf3-4661-bc68-259aa3392940"
ubuntu@ubuntu-rqqzc-z9kb7:~$ mkdir config
ubuntu@ubuntu-rqqzc-z9kb7:~$ sudo mount --read-only --source LABEL=config-2 --target config
ubuntu@ubuntu-rqqzc-z9kb7:~$ ls -l config/openstack/latest/
total 7
-r--r--r-- 1 root root  307 Aug 20 12:53 meta_data.json
[...]

PR #3408, opened 2026-07-06
:seedling: e2e: test network data

dtantsur:
Change the existing preprovisioningNetworkData test to also cover:

1. Defaulting NetworkData and PreprovisioningNetworkData
2. Passing NetworkData to the instance

Unfortunately, we cannot test the effect of NetworkData because
Cirros does not support it.

Partial fix for #2626

### DIFF
--- a/test/e2e/network_data_test.go
+++ b/test/e2e/network_data_test.go
@@ -12,7 +12,6 @@ import (
 	"strings"
 
 	metal3api "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
-	metal3bmc "github.com/metal3-io/baremetal-operator/pkg/hardwareutils/bmc"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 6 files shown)


test/e2e/network_data_test.go
```diff
@@ -0,0 +1,308 @@
...
+	sshClient := EstablishSSHConnection(e2eConfig, connectIP)
+	defer sshClient.Close()
+
+	var output string
+	Eventually(func(g Gomega) {
+		command := "sudo mkdir -p /mnt/cdtest && sudo mount $(blkid -L config-2) /mnt/cdtest && cat /mnt/cdtest/openstack/latest/network_data.json"
+		var err error
+		output, err = executeSSHCommand(sshClient, command)
+		g.Expect(err).NotTo(HaveOccurred(), "Failed to read network_data.json from the config drive")
+	}, e2eConfig.GetIntervals(specName, "wait-user-data")...).Should(Succeed())
```
pierrecregut, line 56: May be at some point "wait-user-data" should be renamed in "wait-ssh-connect" because that is what we are really waiting for and 'user-data' is a little overloaded.

test/e2e/network_data_test.go
```diff
@@ -0,0 +1,308 @@
...
+			By("Verifying the network data file on the config drive")
+			parsed := getNetworkData(specName, bmc.IPAddress)
+			Expect(parsed.Links).To(ContainElement(HaveKeyWithValue("ethernet_mac_address", expectedMAC)))
+			Expect(parsed.Networks).To(ContainElement(HaveKeyWithValue("ip_address", staticIP)))
+		} else {
+			Logf("WARNING: Skipping SSH check since SSH_CHECK_PROVISIONED != true")
+		}
+	})
+
+	It("should apply network data during provisioning", func() {
```
pierrecregut, line 205: To be thorough, one could add a new test (or replace this one) with both preprovisioning-network-data and network-data provided with some minor differences between the two templates (eg. network-id) to guarantee the priority of network-data over preprovisioning.

pierrecregut (commented):
Some minor comments inlined.
/lgtm

test/e2e/network_data_test.go
dtantsur: This is a good point, hmm. I'll try to remember about it, I'm running out of time for this work.

PR #3410, opened 2026-07-08
🐛Delete BMHs before other resources in cleanup for e2e tests, when BMO is ns-scoped

MahnoorAsghar:
When cleaning up resources in e2e tests, when the BMO is namespace-scoped, delete the BMHs before other resources. This will ensure that resources like Secrets that the BMHs might need, are not deleted before the BMHs are. 

**Checklist:**

### DIFF
--- a/test/e2e/common.go
+++ b/test/e2e/common.go
@@ -410,13 +410,20 @@ func Cleanup(ctx context.Context, clusterProxy framework.ClusterProxy, namespace
 	} else {
 		// When namespace-scoped, the namespace is reused between tests; clean up
 		// individual resources to allow the next run to start clean.
+		// Delete BMHs before other tracked resources so that dependent secrets
+		// remain available for deprovisioning and finalizer handling.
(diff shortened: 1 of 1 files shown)


test/e2e/common.go
```diff
@@ -410,13 +410,20 @@ func Cleanup(ctx context.Context, clusterProxy framework.ClusterProxy, namespace
 	} else {
 		// When namespace-scoped, the namespace is reused between tests; clean up
 		// individual resources to allow the next run to start clean.
+		// Delete BMHs before other tracked resources so that dependent secrets
+		// remain available for deprovisioning and finalizer handling.
+		bmhIntervals := cfg.GetIntervals("default", "wait-bmh-deleted")
 		for _, obj := range toCleanup {
```
dtantsur, line 416: Won't it be enough to walk toCleanup in a reverse order?

MahnoorAsghar: It relies on the tests using the correct ordering still

PR #3423, opened 2026-07-13
:sparkles: Add TLS curve preferences support

hroyrh:
Extend TLS settings by including TLS curves preferences cli flag.

### DIFF
--- a/main.go
+++ b/main.go
@@ -59,9 +59,20 @@ const (
 )
 
 type TLSOptions struct {
-	TLSMaxVersion   string
-	TLSMinVersion   string
-	TLSCipherSuites string
+	TLSMaxVersion       string
+	TLSMinVersion       string
+	TLSCipherSuites     string
+	TLSCurvePreferences string
+}
+
+// Supported TLS Curves mapping for go1.25, make sure to update it
+// when go version is updated.
+var supportedTLSCurvesPreferences = map[string]tls.CurveID{
+	"X25519":         tls.X25519,
+	"CurveP256":      tls.CurveP256,
+	"CurveP384":      tls.CurveP384,
+	"CurveP521":      tls.CurveP521,
+	"X25519MLKEM768": tls.X25519MLKEM768,
 }
 
 var (
@@ -156,6 +167,10 @@ func main() {
 	var leaseDurationSeconds string
 	var renewDeadlineSeconds string
 	var retryPeriodSeconds string
+	var supportedTLSCurvesNames = make([]string, 0, len(supportedTLSCurvesPreferences))
+	for name := range supportedTLSCurvesPreferences {
+		supportedTLSCurvesNames = append(supportedTLSCurvesNames, name)
+	}
 
 	// From CAPI point of view, BMO should be able to watch all namespaces
 	// in case of a deployment that is not multi-tenant. If the deployment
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -199,6 +200,11 @@ func main() {
 			"If omitted, the default Go cipher suites will be used. \n"+
 			"Preferred values: "+strings.Join(tlsCipherPreferredValues, ", ")+". \n"+
 			"Insecure values: "+strings.Join(tlsCipherInsecureValues, ", ")+".")
+	flag.StringVar(&tlsOptions.TLSCurvePreferences, "tls-curve-preferences", "",
+		"Comma-separated list of IANA TLS Supported Group IDs for curve/group preferences. "+
+			"If omitted, the default Go curve preferences will be used. "+
+			"Common values: 29 (X25519), 23 (P-256), 24 (P-384), 25 (P-521), 4588 (X25519MLKEM768). "+
```
dtantsur: I've never seen these numbers used in any user-facing configuration (in fact, I hadn't been aware of these numbers until right now).

main.go
```diff
@@ -567,6 +573,30 @@ func GetTLSOptionOverrideFuncs(options TLSOptions) ([]func(*tls.Config), error)
...
+			if idStr == "" {
+				return nil, errors.New("invalid TLS curve preference: empty value in comma-separated list")
+			}
+			id, err := strconv.ParseUint(idStr, 10, 16)
+			if err != nil {
+				return nil, fmt.Errorf("invalid TLS curve preference %q: must be a numeric IANA group ID", idStr)
+			}
+			curve := tls.CurveID(id)
+			if strings.HasPrefix(curve.String(), "CurveID(") {
+				setupLog.Info("TLS curve preference not recognized by current Go version, will be ignored during handshake", "id", id)
```
dtantsur: Is the guidance that we ignore unknown curves this way?

main.go
hroyrh: Switched back to the Curves names to IANA id mapping for better transparency.

main.go
hroyrh: Should we ignore the unsupported Curves or raise an error [here](https://github.com/metal3-io/baremetal-operator/pull/3423/changes#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R592) ?

main.go
```diff
@@ -567,6 +572,31 @@ func GetTLSOptionOverrideFuncs(options TLSOptions) ([]func(*tls.Config), error)
 		})
 	}
 
+	if options.TLSCurvePreferences != "" {
+		curveNames := strings.Split(options.TLSCurvePreferences, ",")
+		var curves []tls.CurveID
+		for _, name := range curveNames {
+			name = strings.TrimSpace(name)
+			switch name {
+			case "X25519":
```
dtantsur: Let's create a `map[string]tls.CurveID` with valid values and also used to generate the help string on line 206

main.go
```diff
@@ -567,6 +572,31 @@ func GetTLSOptionOverrideFuncs(options TLSOptions) ([]func(*tls.Config), error)
...
+				curves = append(curves, tls.X25519)
+			case "CurveP256":
+				curves = append(curves, tls.CurveP256)
+			case "CurveP384":
+				curves = append(curves, tls.CurveP384)
+			case "CurveP521":
+				curves = append(curves, tls.CurveP521)
+			case "X25519MLKEM768":
+				curves = append(curves, tls.X25519MLKEM768)
+			default:
```
dtantsur: Any reason you don't include `SecP256r1MLKEM768` and `SecP384r1MLKEM1024`?

main.go
```diff
@@ -567,6 +572,31 @@ func GetTLSOptionOverrideFuncs(options TLSOptions) ([]func(*tls.Config), error)
 		})
 	}
 
+	if options.TLSCurvePreferences != "" {
+		curveNames := strings.Split(options.TLSCurvePreferences, ",")
+		var curves []tls.CurveID
```
dtantsur: nit: `curves := make([]tls.CurveID, 0, len(curveNames))`

main.go
hroyrh: They are not supported in go1.25 and will be available from 1.26 onward.

dtantsur: Okay, let's leave a TODO to update the list.

main.go
hroyrh: Done

main.go
hroyrh: Done

Rozzii (commented):
/approve

PR #3425, opened 2026-07-15
🌱 Require explicit GitHub token for release notes tool

migiFi:
Harden release tool by allowing the github token to be passed explicitly via `--githubToken` flag instead of reading from the environment.

**Changes:**
- Add a `--githubToken` flag to `notes.go`, it takes precedence over the environment.
- Keep `GITHUB_TOKEN` as a fallback, with a log reminder to prefer the flag.
- Update the release-notes make target to build the tool with a -tag tools and pass `--releaseTag` and `--githubToken` explicitly.

**Improvements:**
- Makes token provisioning to the tool explicit and intentional rather than implicitly inherited from the environment.
- Documents least-privilege expectations.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -459,8 +459,11 @@ $(RELEASE_NOTES_DIR):
 	mkdir -p $(RELEASE_NOTES_DIR)/
 
 .PHONY: release-notes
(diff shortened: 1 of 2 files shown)


smoshiur1237 (commented):
/lgtm

Rozzii (commented):
/approve

PR #3433, opened 2026-07-20
:seedling: e2e: use --no-verbose with wget

dtantsur:
With --quiet, it's not even possible to understand why the download
fails. Using --no-verbose should give at least some clues.


Rozzii (commented):
/approve

MahnoorAsghar:
looks good to me

PR #3437, opened 2026-07-21
🌱 e2e: Add user data secrets to cleanup list

### LINKED ISSUE #3436: e2e: userData secret cleanup
createSSHSetupUserdata does not return the created secret. There is no chance it is ever added to toCleanup list in the test scenarios where it is used. The same is true for createDiskTestUserData.

Note: other uses of CreateSecret are either upgrade scenarios, or the secret is recreated.

/kind bug

pierrecregut:
Ensure that functions generating user-data secrets return the generated secret and append the value to toCleanup in the scenarios using userData.

Fixes: #3436


Rozzii (commented):
/approve
/cc @lentzi90 
I could have sworn that I reviewed this before...

lentzi90 (commented):
/retitle 🌱 e2e: Add user data secrets to cleanup list
(changing 🐛 -> 🌱 since it is just about test code)
/lgtm

PR #3439, opened 2026-07-22
🌱 e2e: introduce a new tag "ironic"

dtantsur:
This tag marks tests that exercise the Ironic provisioner, not just BMO
internals. The idea is to use GINKGO_FOCUS=ironic on ironic-image and
Ironic itself to avoid running unrelated tests.

While here, fix a couple of Describe() statements to produce readable
test descriptions (they're combined with It() in the report).

### DIFF
--- a/test/e2e/README.md
+++ b/test/e2e/README.md
@@ -36,7 +36,7 @@ the e2e Makefile target directly with `make test-e2e`, you may need to adjust yo
 
 Currently there are two sets of tests, which cannot be ran together in the same
 cluster. One is the "optional" set, currently consists of only the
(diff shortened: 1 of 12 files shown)


pierrecregut:
So this tag is currently added to all tests but upgrade tests, metrics service, external-inspection and baremetalswitch (which only looks at the custom resource).

/lgtm

Rozzii (commented):
/approve

dtantsur:
> So this tag is currently added to all tests but upgrade tests, metrics service, external-inspection and baremetalswitch

Correct. We'll probably skip it on HostClaims tests too. While they may exercise deployments and cleanings, it's not their primary focus.

PR #3442, opened 2026-07-25
🌱 3 minor improvement for hardware switch management

alegacy:
This is addresses minor review comments from a downstream merge.  See: https://github.com/openshift/baremetal-operator/pull/510

In detail:

 - Replace discarded AddToScheme returns with require.NoError so setup
failures are caught immediately rather than silently ignored.

 - Replace reflect.DeepEqual with equality.Semantic.DeepEqual when
checking if the HNA spec changed during updates. reflect.DeepEqual
treats nil and empty slices as different, which would incorrectly
block metadata-only updates when a client sends an empty AllowedVLANs
slice instead of omitting the field.

- Add ^ and $ anchors to kubebuilder validation patterns for
NetworkInterface.MACAddress and SwitchPort.SwitchID to prevent
accepting strings that merely contain a valid MAC as a substring.

### DIFF
--- a/apis/metal3.io/v1alpha1/baremetalhost_types.go
+++ b/apis/metal3.io/v1alpha1/baremetalhost_types.go
@@ -466,7 +466,7 @@ type FirmwareConfig struct {
 // SwitchPort defines the attributes required to identify a switch port.
 type SwitchPort struct {
 	// SwitchID is expected to be the management MAC address of the switch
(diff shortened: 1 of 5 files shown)


Rozzii (commented):
/approve

As this is 3 distinct minor commit, let's not squash.

s3rj1k (approved):

Rozzii:
@s3rj1k approved so I am shifting to
/lgtm

PR #3455, opened 2026-07-28
:seedling: Bump go modules across two directories

nuhakala:
Bump dependencies to fix vulnerabilities in preparations for v1.14.0 beta release.


lentzi90 (commented):
Is it expected that `x/crypto` is downgraded? Also `x/text` is "synced" so that we have v0.39.0 consistently, but we already had v0.40.0 in the test module. Would it make more sense to set it to v0.40.0 in the root module?

nuhakala:
> Is it expected that `x/crypto` is downgraded? Also `x/text` is "synced" so that we have v0.39.0 consistently, but we already had v0.40.0 in the test module. Would it make more sense to set it to v0.40.0 in the root module?

Fixed.

lentzi90 (commented):
/approve

Rozzii (commented):
/lgtm

Rozzii:
I think this will be a timing issue, I might have seen this before.

Rozzii:
I can't re trigger the test neither with command nor manually, please change the commit message or something insignificant like that.

Sunnatillo (commented):
/lgtm

PR #3461, opened 2026-07-30
:seedling: vbmctl: Refactor BMC emulator config handling

terror96:
Create sushy-tools and vmbc specific configuration structures. Update config loading, defaults, CLI creation flow, and related tests to use the new structures consistently.

The config file still uses the same flat YAML shape, so `UnmarshalYAML` and `MarshalYAML` bridge the old on-disk format with the new structured representation. This keeps backward compatibility for existing configs and avoids forcing users to rewrite their configuration files. Currently, however, `MarshalYAML` is not used, but is included for completeness.

Currently, the `BMCEmulatorConfig` structure contains all possible configuration parameters, including those specific to only one BMC emulator type. Moving type-specific parameters into their own dedicated structures makes the configuration easier to maintain and allows the list of supported emulator types to be expanded more cleanly.

### DIFF
--- a/test/vbmctl/cmd/vbmctl/create.go
+++ b/test/vbmctl/cmd/vbmctl/create.go
@@ -453,19 +453,20 @@ func newCreateBMCEmulatorCmd() *cobra.Command {
 			if image == "" {
 				if effective.Image != "" {
 					image = effective.Image
-				} else if emulatorType == config.BMCEmulatorTypeSushyTools {
+				} else if emulatorType == vbmctlapi.BMCEmulatorTypeSushyTools {
 					image = config.DefaultBMCEmulatorSushyToolsImage
 				} else {
 					image = config.DefaultBMCEmulatorVBMCImage
 				}
 			}
 
-			if emulatorType == config.BMCEmulatorTypeSushyTools {
+			if emulatorType == vbmctlapi.BMCEmulatorTypeSushyTools {
+				sushyConfig := &effective.SushyToolsConfig
 				if configFile == "" {
 					// There is no default config file for sushy-tools, so no
 					// need to check if the value from config file is empty
 					// before using it.
-					configFile = effective.ConfigFile
+					configFile = sushyConfig.ConfigFile
 				}
 
 				// If using command line arguments to configure sushy-tools, apply the
@@ -473,15 +474,15 @@ func newCreateBMCEmulatorCmd() *cobra.Command {
 				// explicitly set. In case a sushy-tools config file is specified,
 				// defaults are not applied.
 				if listenAddress == "" {
-					if effective.ListenAddress != "" {
-						listenAddress = effective.ListenAddress
+					if sushyConfig.ListenAddress != "" {
+						listenAddress = sushyConfig.ListenAddress
 					} else if configFile == "" {
 						listenAddress = config.DefaultNetworkAddress
 					}
 				}
 				if listenPort == 0 {
-					if effective.ListenPort != 0 {
-						listenPort = effective.ListenPort
+					if sushyConfig.ListenPort != 0 {
+						listenPort = sushyConfig.ListenPort
 					} else if configFile == "" {
 						listenPort = config.DefaultBMCEmulatorSushyToolsListenPort
 					}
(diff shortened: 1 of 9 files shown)


terror96:
Ping @lentzi90 @nuhakala

terror96:
@copilot

test/vbmctl/cmd/vbmctl/create.go
```diff
@@ -491,27 +497,31 @@ func newCreateBMCEmulatorCmd() *cobra.Command {
...
-	cmd.Flags().StringVar(&listenAddress, "listen-address", "", "address for the BMC emulator to listen on for incoming connections (only applicable for sushy-tools, default is "+config.DefaultNetworkAddress+" if not set in config file and no config file is used)")
-	cmd.Flags().Uint16Var(&listenPort, "listen-port", 0, "port for the BMC emulator to listen on for incoming connections (only applicable for sushy-tools, default is "+strconv.Itoa(int(config.DefaultBMCEmulatorSushyToolsListenPort))+" if not set in config file and no config file is used)")
+	cmd.Flags().StringVar(&listenAddress, "listen-address", "", "address for the BMC emulator to listen on for incoming connections (only applicable for sushy-tools, default is "+config.DefaultNetworkAddress+" if not set in config file)")
+	cmd.Flags().Uint16Var(&listenPort, "listen-port", 0, "port for the BMC emulator to listen on for incoming connections (only applicable for sushy-tools, default is "+strconv.Itoa(int(config.DefaultBMCEmulatorSushyToolsListenPort))+" if not set in config file)")
```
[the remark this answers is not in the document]
terror96: This made me to realize that this needs to be further clarified in the help test. The first config file here is the `vbmctl config file` whereas the second refers to the `Sushy config file`.

test/vbmctl/pkg/api/types.go
```diff
@@ -64,31 +68,15 @@ type VolumeMount struct {
...
-	StoragePool string `json:"storagePool" yaml:"storagePool"`
-
-	// LibvirtURI is the libvirt URI that sushy-tools should connect to.
-	LibvirtURI string `json:"libvirtURI" yaml:"libvirtUri"`
-
 	// Image is the container image to use for the BMC emulator.
 	Image string `json:"image" yaml:"image"`
 
+	// SpecificConfig contains emulator-specific settings for the selected type.
+	SpecificConfig interface{} `json:"-" yaml:"-"`
```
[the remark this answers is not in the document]
terror96: Should be fixed.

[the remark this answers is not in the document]
terror96: <s>Should be fixed.</s>

EDIT: code uses `yaml.Unmarshal()` and there is no need for `json` helpers because it can parse `json` files and use the YAML helpers to construct `SpecificConfig`.

test/vbmctl/pkg/api/types.go
```diff
@@ -102,6 +90,131 @@ type BMCEmulatorConfig struct {
...
+			ConfigFile:    raw.ConfigFile,
+			ListenAddress: raw.ListenAddress,
+			ListenPort:    raw.ListenPort,
+			StoragePool:   raw.StoragePool,
+			LibvirtURI:    raw.LibvirtURI,
+		}
+	case "vbmc":
+		c.SpecificConfig = &VBMCConfig{}
+	default:
+		return fmt.Errorf("unsupported BMC emulator type: %s", c.Type)
```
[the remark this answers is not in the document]
terror96: Makes sense.

terror96:
> ## Pull request overview
> 
> Copilot reviewed 9 out of 9 changed files in this pull request and generated no new comments.
> Suppressed comments (1)
> 
> **test/vbmctl/pkg/api/types.go:76**
> 
>     * `SpecificConfig` accepts any value, but all consumers recognize only `*SushyToolsConfig` or `*VBMCConfig`. For example, `SetSpecificConfig(SushyToolsConfig{...})` compiles, yet `GetSushyToolsConfig` returns nil and validation/defaulting replaces the supplied settings with an empty config; a type/config mismatch is similarly discarded. Constrain this field/setter to the supported config types or reject invalid values instead of silently losing configuration.
> 
> 
> ```
> 	// SpecificConfig contains emulator-specific settings for the selected type.
> 	SpecificConfig interface{} `json:"-" yaml:"-"`
> ```

Dropped the single polymorphic setter (and getter) and implemented separate setters instead.

test/vbmctl/pkg/api/types.go
```diff
@@ -64,31 +66,15 @@ type VolumeMount struct {
...
-	StoragePool string `json:"storagePool" yaml:"storagePool"`
-
-	// LibvirtURI is the libvirt URI that sushy-tools should connect to.
-	LibvirtURI string `json:"libvirtURI" yaml:"libvirtUri"`
-
 	// Image is the container image to use for the BMC emulator.
 	Image string `json:"image" yaml:"image"`
 
+	// SpecificConfig contains emulator-specific settings for the selected type.
+	SpecificConfig interface{} `json:"-" yaml:"-"`
```
lentzi90: This may not be the best way to do it. If I understand correctly, we now say that `SpecificConfig` can be anything and then we parse it into either `SushyToolsConfig` or `VBMCConfig`. As an API this is hard to document, validate and understand.
A more common way of doing it (I think) is to have one field for each type. They can then be strictly validated and easily documented. Something like:

```go
type BMCEmulatorConfig struct {
	// BMC Emulator type.
	Type string `json:"type" yaml:"type"`
	// SushyToolsConfig. Required when Type is sushy-tools.
	SushyToolsConfig SushyToolsConfig `json:"sushyToolsConfig" yaml: "sushyToolsConfig"`
	// VBMCConfig. Required when Type is vbmc.
	VBMCConfig VBMCConfig `json:"vbmcConfig" yaml:"vbmcConfig"`
```

What do you think about this?

terror96: You have a valid point, and I don't have a strong opinion either way. My thinking started with a C-style `void *` approach and got a bit carried away with AI magic.

However, if re-refactor this according to your suggestion, I think the implementation and the API will be cleaner for also reasons you pointed out. Let's see if this needs to wait until I get back...

terror96: Pushing a new suggestion: I now eliminated also the set/get helpers, because it seemed a little bit overkill to have extra code for accessing a field in the structure which is always present; just need to check the type first.

terror96:
> ## Pull request overview
> 
> Copilot reviewed 8 out of 8 changed files in this pull request and generated no new comments.
> Suppressed comments (2)

Basing on this, reworked the output a bit. Now sushy-tools specific values are printed only if the type is sushy-tools. Zero port value is now omitted from output. The port can be zero if a sushy-tools configuration file is provided: this file may or may not contain the values, but if they are missing the `sushy-tools` binary uses the default values: all local interfaces and port 8000. Parsing (and possibly filling in default values) in case of a configuration file just for the purpose of `config view` does not sound reasonable.

lentzi90 (commented):
Thanks this looks good now!
/approve

test/vbmctl/pkg/containers/bmc_emulator.go
```diff
@@ -111,40 +112,44 @@ func createSushyToolsEmulatorInstance(ctx context.Context, cfg *vbmctlapi.BMCEmu
...
+
 	switch cfg.Type {
-	case config.BMCEmulatorTypeVBMC:
+	case vbmctlapi.BMCEmulatorTypeVBMC:
 		return createVBMCEmulatorInstance(ctx, cfg)
-	case config.BMCEmulatorTypeSushyTools:
+	case vbmctlapi.BMCEmulatorTypeSushyTools:
 		return createSushyToolsEmulatorInstance(ctx, cfg)
 	default:
 		return fmt.Errorf("unsupported BMC emulator type: %s", cfg.Type)
```
nuhakala, line 155: TBH, this is still little tangled. You are creating separate structure for sushy-tools but this structure is accessed always through the embedding `BMCEmulatorConfig`. So the benefit of grouping emulator specific configuration together is reduced because it does not change the flow of data here.

IMO it would be more clear if here we give the emulator specific structure as parameter to the `createSushyToolsEmulatorInstance` and then put the internal variables from `BMCEmulatorConfig` into a new structure (`emulatorInstanceoptions` or something similar) which can be internal to this package. And this new structure is passed from `createSushyToolsEmulatorInstance` to the function that actually creates the container.

Why I think this is better:

- Separation of the config file and internal values. The structure `BMCEmulatorConfig` would be only related to the values/fields in the config file. It is not used to pass around any internal values. Makes the purpose and usage of this structure more clear.
- The actual container creation flow becomes more clear when we pass specific arguments instead of the whole large structure here.
- vbmc config structure would actually be used instead of being defined but not used anywhere

Downsides:

- We need extra structure in this package, which basically increases boilerplate
- This pattern is basically a midlayer between the creation command and container creation. One could argue whether midlayer pattern is good or not in the first place, but I think it does not really matter here because we are most likely not gonna support dozens of emulators, but only a handful.

nuhakala (commented):
looks good but I would like to get your thoughts on the comment below before giving lgtm

test/vbmctl/pkg/containers/bmc_emulator.go
terror96: I did not have this on my mind with the PR, but the PR has been around for awhile and I don't see a problem in working a bit more on this. The same thought has crossed my mind, and maybe the above justification gives it a good reason to bake this refactoring as part of this PR. I agree, keeping config variables and variables for internal usage only separate make things easier to follow. What you think @lentzi90? I guess do not have the same approach with other configuration structure so why not align this with the rest of the structures too.

nuhakala: If you were planning other refactoring PR then we can merge this and do more refactoring in another PR

terror96: As discussed, I can make another commit in the same PR, because it still lands nicely under the same "refactor" topic.

lentzi90: > As discussed, I can make another commit in the same PR, because it still lands nicely under the same "refactor" topic.

Sounds good :+1:

terror96: Done: moved internal variables to `bmc_emulator.go` (see the second commit for details).

PR #3463, opened 2026-07-30
🌱 Harden hack scripts against WORKDIR volume mount injection

migiFi:
Affects the following scripts under the hack/ directory:

- generate.sh
- gomod.sh
- manifestlint.sh
- markdownlint.sh
- shellcheck.sh

These scripts read WORKDIR from the env and insert it directly into the container's --volume mount specification. Because the volume spec is colon-delimited (`host:target:options`), a WORKDIR value containing a colon could alter how the mount is parsed.

This change removes the WORKDIR var and hardcodes the container path to /workdir, which matches the pattern already used in several other metal3 repos. Since /workdir was already the default, there is no change in behavior for normal use, but the injection surface is eliminated.

### DIFF
--- a/hack/generate.sh
+++ b/hack/generate.sh
@@ -9,7 +9,6 @@ set -eux
 IS_CONTAINER="${IS_CONTAINER:-false}"
 ARTIFACTS="${ARTIFACTS:-/tmp}"
 CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-podman}"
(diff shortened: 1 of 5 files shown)


lentzi90 (commented):
/lgtm

adilGhaffarDev (approved):
/approve

lentzi90:
Oh, we can switch labels @adilGhaffarDev 
/lgtm cancel
/approve

PR #3464, opened 2026-07-30
🌱 Pin ironic version to 35.0 in default e2e overlay

UncivilizedGhost:
**What this PR does / why we need it**:Use the plain e2e overlay instead of a versioned one, patched to
pin spec.version to "35.0" rather than "latest".

Fixes #

**Checklist:**

### DIFF
--- a/test/e2e/data/ironic-standalone-operator/ironic/overlays/e2e/ironic.yaml
+++ b/test/e2e/data/ironic-standalone-operator/ironic/overlays/e2e/ironic.yaml
@@ -4,7 +4,4 @@ metadata:
   name: ironic
   namespace: baremetal-operator-system
 spec:
-  images:
-    deployRamdiskDownloader: "quay.io/metal3-io/ironic-ipa-downloader:main"
-    ironic: "quay.io/metal3-io/ironic:main"
-    keepalived: "quay.io/metal3-io/keepalived"
+  version: "35.0"


lentzi90 (commented):
31 is too old for BMO 0.13. I can see we have an overlay already for 35, let's use that.

lentzi90:
Oh how irritating. The versioned overlays are used for the upgrade tests, and those use a different IP to avoid conflicts. That is why the tests fail. You can see that they include the `upgrade-ip` here: https://github.com/metal3-io/baremetal-operator/blob/e2041b8819d8503d0bfc07b779d555034914acd5/test/e2e/data/ironic-standalone-operator/ironic/overlays/e2e-release-35.0/kustomization.yaml#L11
We will need a separate overlay or changing the `e2e` overlay. I would suggest that we change the `e2e` overlay here: https://github.com/metal3-io/baremetal-operator/blob/release-0.13/test/e2e/data/ironic-standalone-operator/ironic/overlays/e2e/ironic.yaml. Remove the `images` and put `version: "33.0"` instead.

lentzi90:
Sorry, I don't know why I said 33.0. We should use 35.0 here. Too early in the morning :yawning_face: 
At least it looks like this is working

lentzi90 (commented):
/approve

adilGhaffarDev (commented):
/lgtm

PR #3465, opened 2026-07-30
⚠️ Deprecate BMH Taints field

### LINKED ISSUE #3453: The .spec.taints of BareMetalHosts is ignored

https://github.com/metal3-io/baremetal-operator/blob/5fbeafac6a41124c80c9fab73a212288d342c6d3/apis/metal3.io/v1alpha1/baremetalhost_types.go#L513-L517

However, setting it has no effect.

Either it has effect, or the field is removed (or at least there is a warning that this is not _yet_ implemented).

**Environment:**

- Baremetal Operator version: v0.13.0
- Environment (metal3-dev-env or other):

/kind bug

dtantsur:
This field was added in the very first revision of the BMH API but has
never been implemented in BMO and CAPM3. Deprecate it to avoid confusion.

Also remove an obvious comment from the same time.

Fixes: #3453 

### DIFF
--- a/apis/metal3.io/v1alpha1/baremetalhost_types.go
+++ b/apis/metal3.io/v1alpha1/baremetalhost_types.go
@@ -507,12 +507,10 @@ type NetworkInterface struct {
 
 // BareMetalHostSpec defines the desired state of BareMetalHost.
 type BareMetalHostSpec struct {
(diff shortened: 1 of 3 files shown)


Sunnatillo (commented):
/lgtm

Rozzii (commented):
/approve

PR #3469, opened 2026-07-30
✨ API types and Ironic provisioner support for switch port configuration (ToR networking part1)

alegacy:
## Summary                                                

  - Add API types for switch port configuration on BareMetalHost: `NetworkInterface`,
    `SwitchPortConfig`, `SwitchPortIdentifier`, `AppliedPortConfig`, and the
    `NetworkInterfacesValid` status condition
  - Implement port configuration management in the Ironic provisioner: resolve
    switchport configs from `ManagementAccessData`, apply VLAN/MTU/mode settings
    to Ironic ports at creation and update time, with LLDP fallback for
    `local_link_connection`

  ## Details

  The API commit defines the data model. Each `NetworkInterface` on a BMH maps a
  host NIC (by name or MAC) to a `HostNetworkAttachment` that carries the desired
  switchport configuration.  `AppliedPortConfig` in status records the concrete
  values last written to Ironic so the reconciler can detect drift.

  The provisioner commit extends `Register` to accept resolved port configs and
  apply them to Ironic ports.  Existing ports are updated when config drifts,
  but only outside active provisioning operations to avoid conflicting with
  Ironic workflows.  A fixture provisioner update uses the boot MAC from
  `HostData` for consistency.

Partially Fixes # https://github.com/metal3-io/metal3-docs/pull/586

**Checklist:**


apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -463,15 +466,33 @@ type FirmwareConfig struct {
 	SriovEnabled *bool `json:"sriovEnabled,omitempty"`
 }
 
-// SwitchPort defines the attributes required to identify a switch port.
-type SwitchPort struct {
+// SwitchPortIdentifier defines the attributes required to identify a switch port.
+type SwitchPortIdentifier struct {
```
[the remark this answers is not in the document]
dtantsur: I think we can let this slide given that networking API is not usable yet.

pkg/provisioner/ironic/ironic.go
```diff
@@ -285,21 +309,100 @@ func (p *ironicProvisioner) createNodePort(ctx context.Context, uuid string, mac
...
+		Address:    macAddress,
+		PXEEnabled: &nic.PXE,
+	}
+
+	// Set port configuration if networking is enabled
+	if p.config.enableNetworking {
+		if portConfig, found := p.portConfigs[macAddress]; found {
+			createOpts.Extra = map[string]interface{}{
+				"switchport": buildSwitchPortFromConfig(&portConfig.SwitchPortConfig),
+			}
```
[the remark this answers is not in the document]
alegacy: fixed

pkg/provisioner/ironic/ironic_networking.go
```diff
@@ -0,0 +1,143 @@
...
+		return nil
+	}
+
+	connection := make(map[string]interface{})
+
+	if nic.LLDP.SwitchID != "" {
+		connection["switch_id"] = nic.LLDP.SwitchID
+	}
+	if nic.LLDP.PortID != "" {
+		connection["port_id"] = nic.LLDP.PortID
```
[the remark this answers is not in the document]
alegacy: fixed

pkg/provisioner/ironic/ironic.go
```diff
@@ -285,21 +309,100 @@ func (p *ironicProvisioner) createNodePort(ctx context.Context, uuid string, mac
...
+	_, err := ports.Create(ctx, p.client, createOpts).Extract()
 	if err != nil {
 		return fmt.Errorf("failed to create ironic port for node %s, MAC: %s: %w", uuid, macAddress, err)
 	}
 
 	return nil
 }
 
+// updateNodePort updates port with local_link_connection and switchport data.
+func (p *ironicProvisioner) updateNodePort(ctx context.Context, existingPort ports.Port, portConfig *provisioner.PortConfig) error {
```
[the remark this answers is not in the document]
alegacy: fixed

pkg/provisioner/ironic/register.go
```diff
@@ -303,36 +318,60 @@ func (p *ironicProvisioner) createPortsForNode(ctx context.Context, ironicNode *
 		return nil
 	}
 
-	ironicNodePorts, err := p.getPorts(ctx, ironicNode.UUID, "")
+	detail := portFieldLevel(updateExisting)
+	ironicNodePorts, err := p.getPorts(ctx, ironicNode.UUID, "", detail)
```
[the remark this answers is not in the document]
alegacy: fixed

pkg/provisioner/ironic/factory.go
```diff
@@ -379,7 +379,17 @@ func loadConfigFromEnv(havePreprovImgBuilder bool) (ironicConfig, error) {
 		}
 	}
 
-	c.provNetDisabled = strings.ToLower(os.Getenv("PROVISIONING_NETWORK_DISABLED")) == "true"
+	const envTrue = "true"
+
+	c.provNetDisabled = strings.ToLower(os.Getenv("PROVISIONING_NETWORK_DISABLED")) == envTrue
+
+	// Enable management of switch ports if networking is enabled
+	c.enableNetworking = strings.ToLower(os.Getenv("IRONIC_NETWORKING_ENABLED")) == envTrue
```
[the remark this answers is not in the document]
alegacy: fixed

apis/metal3.io/v1alpha1/baremetalhost_types.go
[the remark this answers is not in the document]
alegacy: Ignoring since the API is not complete

pkg/provisioner/ironic/register.go
```diff
@@ -303,36 +318,60 @@ func (p *ironicProvisioner) createPortsForNode(ctx context.Context, ironicNode *
...
-	if _, ok := ironicNodePortsList[p.bootMACAddress]; p.bootMACAddress != "" && !ok {
-		portMacsToCreate[p.bootMACAddress] = true
+	bootMACAddress := strings.ToLower(p.bootMACAddress)
+	if _, ok := ironicNodePortsList[bootMACAddress]; bootMACAddress != "" && !ok {
+		if _, ok := portsToCreate[bootMACAddress]; !ok {
+			portsToCreate[bootMACAddress] = metal3api.NIC{
+				MAC: bootMACAddress,
+				PXE: true,
+			}
+		}
```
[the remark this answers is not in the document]
alegacy: fixed

pkg/provisioner/ironic/ironic_networking_test.go
```diff
@@ -0,0 +1,278 @@
+package ironic
```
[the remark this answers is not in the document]
alegacy: There are no copyright headers in any other provisioner files.

pkg/provisioner/ironic/ironic_networking.go
```diff
@@ -0,0 +1,143 @@
+package ironic
```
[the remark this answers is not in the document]
alegacy: There are no copyright headers in any other provisioner files.

pkg/provisioner/provisioner.go
```diff
@@ -90,6 +99,22 @@ type ManagementAccessData struct {
 	CPUArchitecture            string
 	HardwareData               *metal3api.HardwareData
 	DisableInspection          bool
+	PortConfigs                map[string]*PortConfig
```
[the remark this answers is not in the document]
alegacy: that's expected.  the changes are in a follow-up PR #3470

pkg/provisioner/ironic/register.go
```diff
@@ -306,36 +321,64 @@ func (p *ironicProvisioner) createPortsForNode(ctx context.Context, ironicNode *
...
+	// The detail flag on the initial listing ensures we have the Extra
+	// and LocalLinkConnection fields needed for comparison.
+	if updateExisting && p.config.enableNetworking && p.portConfigs != nil {
+		p.log.Info("updating ports for node", "nodeUUID", ironicNode.UUID, "MACs", slices.Collect(maps.Keys(ironicNodePortsList)))
+		for _, port := range ironicNodePorts {
+			mac := strings.ToLower(port.Address)
+			config := p.portConfigs[mac]
+			if err := p.updateNodePort(ctx, port, config); err != nil {
+				return err
+			}
```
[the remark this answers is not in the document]
alegacy: This is design intent.  If there is no resolve port config for a port then we definitely want to remove the corresponding attributes from the ironic port config attributes.

dtantsur:
CI fails with `Could not find the following interface in the 'ironic.hardware.interfaces.network' entrypoint: ironic-networking. Valid interfaces are ['noop']`, which makes me think that networking is not properly enabled in IrSO.

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -813,6 +834,16 @@ type OperationHistory struct {
 	Deprovision OperationMetric `json:"deprovision,omitempty"`
 }
 
+// AppliedPortConfig records the port configuration that was last successfully
+// applied to an Ironic port for a network interface.
+type AppliedPortConfig struct {
+	// Name is the network interface name (e.g., "eno1np0")
+	Name             string           `json:"name"`
+	SwitchPortConfig SwitchPortConfig `json:"switchPortConfig"`
```
dtantsur, line 846: nit: let's have docstrings for all fields, even obvious ones

pkg/provisioner/ironic/factory.go
```diff
@@ -379,7 +379,24 @@ func loadConfigFromEnv(havePreprovImgBuilder bool) (ironicConfig, error) {
...
+	networkingValue := os.Getenv("IRONIC_NETWORKING_ENABLED")
+	if networkingValue != "" {
+		var parseErr error
+		c.enableNetworking, parseErr = strconv.ParseBool(networkingValue)
+		if parseErr != nil {
+			return c, fmt.Errorf("invalid IRONIC_NETWORKING_ENABLED value %q: %w", networkingValue, parseErr)
+		}
+	}
+
+	c.networkInterface = os.Getenv("IRONIC_NETWORK_INTERFACE")
```
dtantsur, line 403: I don't remember, what was the motivation for this variable? It does not look like we're getting more standalone interfaces soon.

pkg/provisioner/ironic/ironic.go
```diff
@@ -285,21 +309,100 @@ func (p *ironicProvisioner) createNodePort(ctx context.Context, uuid string, mac
...
+					op = ports.ReplaceOp
+				}
+				updateOpts = append(updateOpts, ports.UpdateOperation{
+					Op:    op,
+					Path:  "/local_link_connection",
+					Value: llc,
+				})
+			}
+		}
+	} else if existingPort.Extra != nil && existingPort.Extra["switchport"] != nil {
```
dtantsur: nit: I believe you can read from a `nil` map

pkg/provisioner/provisioner.go
```diff
@@ -90,6 +99,22 @@ type ManagementAccessData struct {
...
+}
+
+// IsActiveOperation returns true when the given provisioning state
+// indicates an active operation where port modifications should not
+// be performed.
+func IsActiveOperation(state metal3api.ProvisioningState) bool {
```
dtantsur: nit: I'd go with something less generic like `CanUpdateSwitchConfig`

pkg/provisioner/provisioner.go
```diff
@@ -90,6 +99,22 @@ type ManagementAccessData struct {
...
+	switch state {
+	case metal3api.StateProvisioning, metal3api.StateProvisioned,
```
dtantsur: I wonder if it's easier to list states where it's allowed. I mean, it's also dangerous during things like inspection, right? So maybe only allow updates in Registering, Available, and.. well, that's all?

alegacy: Ok, I'll reverse the logic.

pkg/provisioner/ironic/factory.go
alegacy: Ok, I can hardcode it to "ironic-networking" for now without reading the var at all.

alegacy: Actually... we need to be able to override this for e2e tests to "noop" so we need this var.

dtantsur: Sigh, this is unfortunate. Maybe call it appropriately, like `TESTING_USE_NOOP_NETWORKING=true`? To avoid impression that this is something that users should ever change?

pkg/provisioner/provisioner.go
alegacy: changed to `registering` and `available` only.

pkg/provisioner/provisioner.go
alegacy: Changed to `CanUpdateSwitchPortConfig`

pkg/provisioner/ironic/ironic.go
alegacy: Simplified.

pkg/provisioner/ironic/factory.go
alegacy: I kept the name the same but added a comment explaining its purpose.   I hope this satisfies the concern.

apis/metal3.io/v1alpha1/baremetalhost_types.go
alegacy: Done

pkg/provisioner/ironic/register.go
```diff
@@ -172,7 +180,10 @@ func (p *ironicProvisioner) Register(ctx context.Context, data provisioner.Manag
 		// Try to create ports from two sources
 		// bootMACAddress if available
 		// HardwareData whenever inspection data is available.
-		err = p.createPortsForNode(ctx, ironicNode, data.HardwareData)
+		// Avoid updating existing ports during active operations
+		// (provisioning, deprovisioning, etc.) to prevent disruption.
+		updateExisting := provisioner.CanUpdateSwitchPortConfig(data.State)
```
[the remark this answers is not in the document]
alegacy: I don't think this is valid or at least not a real concern.  If we're registering a new or existing host then it means that it isn't already in Ironic so the ports will get created NOT updated.   There shouldn't be a case where we are registering and we find ports already there, IIUC.

s3rj1k:
One concern I have about this, is that it is very coupled to Ironic, we would benefit in the long run if we make this similarly to Provisioner interface and should be done as a followup so not to complicate current PR.

alegacy:
> One concern I have about this, is that it is very coupled to Ironic, we would benefit in the long run if we make this similarly to Provisioner interface and should be done as a followup so not to complicate current PR.

@s3rj1k  Can you elaborate a bit on what you have in mind?

s3rj1k:
> > One concern I have about this, is that it is very coupled to Ironic, we would benefit in the long run if we make this similarly to Provisioner interface and should be done as a followup so not to complicate current PR.
> 
> @s3rj1k Can you elaborate a bit on what you have in mind?

I'd like to be able, in future, to replace ToR configuration implementation, so instead of relaying on Ironic, I would for example have my own controller that does all this and I want to reuse Metal3 objects against custom implementation.

Just like now there is a possibility to replace Ironic Provisioning part with custom provisioner.

dtantsur:
> One concern I have about this, is that it is very coupled to Ironic

The logic is already contained in the Provisioner, so I'm not sure why it's a concern.

apis/metal3.io/v1alpha1/baremetalhost_types.go
```diff
@@ -811,6 +832,18 @@ type OperationHistory struct {
...
+// AppliedPortConfig records the port configuration that was last successfully
+// applied to an Ironic port for a network interface.
+type AppliedPortConfig struct {
+	// Name is the network interface name (e.g., "eno1np0")
+	Name string `json:"name"`
+	// SwitchPortConfig is the switch port configuration that was applied
+	SwitchPortConfig SwitchPortConfig `json:"switchPortConfig"`
+	// LocalLinkConnection is the local link connection data from Ironic
+	// +optional
+	LocalLinkConnection *SwitchPortIdentifier `json:"localLinkConnection,omitempty"`
```
dtantsur: A bit on naming: I believe we don't use LocalLinkConnection as a term anywhere else? You call it SwitchPort above, which is going to be confusing with SwitchPortConfig.

Even if it means some loss of technical precision, maybe SwitchPortConfig should be called something like NetworkAttachmentConfig to reflect that it's coming from an HNA, while this field should be SwitchConfig for consistency with the Spec field?

alegacy: I agree.  LLC is about out of place here.  

How about something like this:
```
type AppliedNetworkInterface struct {
	// Name is the network interface name (e.g., "eno1np0")
	Name string `json:"name"`
	// SwitchPortConfig is the switch port configuration that was applied
	SwitchPortConfig SwitchPortConfig `json:"switchPortConfig"`
	// SwitchPortIdentifier is the identifier of the switch port that was
	// configured.  If not provided then LLDP information will be used.
	// +optional
	SwitchPortIdentifier *SwitchPortIdentifier `json:"switchPort,omitempty"`
}
```

`switchPort` as a json tag aligns with the equivalent in the NetworkInterface
`switchPortConfig` represents the config applied to the switch port


or ... `AppliedNetworkAttachmentConfig` if you prefer.

dtantsur: For me `AppliedNetworkAttachmentConfig` is much better because it refers to the CRD name. I'd go down this path if you're fine with it.

alegacy: Ok, so I'll refactor to this:

```
type AppliedNetworkAttachmentConfig struct {
	// Name is the network interface name (e.g., "eno1np0")
	Name string `json:"name"`

	// SwitchPortConfig is the switch port configuration that was applied
	SwitchPortConfig SwitchPortConfig `json:"switchPortConfig"`

	// SwitchPortIdentifier is the identifier of the switch port that was
	// configured.  If not provided then LLDP information will be used.
	// +optional
	SwitchPortIdentifier *SwitchPortIdentifier `json:"switchPort,omitempty"`
}
```

dtantsur: > SwitchPortIdentifier *SwitchPortIdentifier `json:"switchPort,omitempty"`

The JSON field should match

alegacy: Refactored.

pkg/provisioner/ironic/factory.go
```diff
@@ -379,7 +379,31 @@ func loadConfigFromEnv(havePreprovImgBuilder bool) (ironicConfig, error) {
 		}
 	}
 
-	c.provNetDisabled = strings.ToLower(os.Getenv("PROVISIONING_NETWORK_DISABLED")) == "true"
+	const envTrue = "true"
```
dtantsur, line 382: nit: leftover from using "true" in more than one place?

pkg/provisioner/ironic/factory.go
```diff
@@ -379,7 +379,31 @@ func loadConfigFromEnv(havePreprovImgBuilder bool) (ironicConfig, error) {
...
+
+	c.provNetDisabled = strings.ToLower(os.Getenv("PROVISIONING_NETWORK_DISABLED")) == envTrue
```
dtantsur, line 384: nit: we should change this one to ParseBool for consistency (can be done separately)

pkg/provisioner/ironic/ironic.go
```diff
@@ -266,11 +289,12 @@ func (p *ironicProvisioner) findExistingHost(ctx context.Context, bootMACAddress
 	return nil, nil //nolint:nilnil
 }
 
-func (p *ironicProvisioner) createNodePort(ctx context.Context, uuid string, macAddress string, pxe bool) error {
-	p.log.Info("creating ironic port for node", "NodeUUID", uuid, "MAC", macAddress, "PXE status", pxe)
+func (p *ironicProvisioner) createNodePort(ctx context.Context, uuid string, nic metal3api.NIC) error {
+	macAddress := strings.ToLower(nic.MAC)
+	p.log.Info("creating ironic port for node", "NodeUUID", uuid, "MAC", macAddress, "PXE status", nic.PXE)
```
dtantsur, line 294: nit: this should be logged after we check for a duplicate (or rephase to avoid implying that the creating process will happen in any case)

pkg/provisioner/ironic/ironic_networking.go
```diff
@@ -0,0 +1,135 @@
...
+	if config.SwitchSystemName != "" {
+		llc["switch_info"] = config.SwitchSystemName
+	}
+	return llc
+}
+
+// buildLocalLinkFromNIC creates a local_link_connection map from stored LLDP
+// data. Ironic requires both switch_id and port_id when local_link_connection
+// is set, so this returns nil if either is missing.
+func buildLocalLinkFromNIC(nic metal3api.NIC) map[string]interface{} {
```
dtantsur, line 30: nit: `interface{}` -> `any`

pkg/provisioner/ironic/ironic.go
```diff
@@ -285,21 +309,105 @@ func (p *ironicProvisioner) createNodePort(ctx context.Context, uuid string, mac
...
+
+	// Add switch port config if available; otherwise remove
+	if portConfig != nil {
+		existing, exists := existingPort.Extra["switchport"]
+		equal := false
+		if exists && existing != nil {
+			var err error
+			equal, err = switchPortConfigsEqual(existing, &portConfig.SwitchPortConfig)
+			if err != nil {
+				return fmt.Errorf("failed to parse switchport config for port %s: %w", existingPort.UUID, err)
```
dtantsur, line 359: I wonder what the recovery path here is. Currently, a user must fix their Ironic directly. Can we say "the port has garbage, we'll overwrite it"?

dtantsur:
/lgtm

A bunch of minor comments inline, all can be addressed in a follow-up.

We also need to figure out the feature gates topic (see metal3-dev), but this can happen in parallel, I hope.

dtantsur:
/cc @lentzi90 @Rozzii 

Please take a look. Once this merges, I intend to follow-up with a feature gate to cover it.

pkg/provisioner/ironic/ironic.go
alegacy: I think this is fairly unlikely to happen, but yes, if you prefer we could theoretically just consider this a !equal case and overwrite it.

pkg/provisioner/ironic/ironic_networking.go
alegacy: fixed

pkg/provisioner/ironic/ironic.go
alegacy: fixed

pkg/provisioner/ironic/factory.go
alegacy: This line isn't part of these changes (beyond the `envTrue` usage) so I'll skip this.

pkg/provisioner/ironic/factory.go
alegacy: reverted the `envTrue`

alegacy:
@dtantsur had to rebase so lost your `lgtm`... took the opportunity to address your nits since I had to push a change anyway.  Please take another look.

pkg/provisioner/ironic/ironic.go
dtantsur: Yeah, let's probably consider "it's junk" as the same as "not equal". (Can be done in a follow-up.)

PR #3483, opened 2026-08-04
:seedling: Add items section for ironic-cacert secret in tls.yaml

peppi-lotta:
The `cert-ironic-ca` volume mounts the `ironic-cacert` Secret without an `items` section adding `tls.key` to both the `ironic` and `ironic-httpd` containers. These containers only need the CA certificate `tls.crt` for trust verification. Add items section to prevent unnecessary `tls.key` from being added to the containers. This is a small security hardening.


Sunnatillo (commented):
/lgtm

PR #3490, opened 2026-08-06
🌱 e2e: fix a typo in the expected state

dtantsur:
When setting externallyProvisioned to false, the result should be
Provisioned, not ExternallyProvisioned. The test only passes because
BMO does not update the host quickly enough.


lentzi90 (commented):
/lgtm

dtantsur:
Okay, assuming the test did not exist in 0.13

PR #3505, opened 2026-08-11
🌱 Remove zizmor inline ignore

lentzi90:

Removed the `# zizmor: ignore[artipacked]` inline suppression comment from `.github/workflows/release.yaml`. The `artipacked` warning is now handled via a targeted ignore rule in `.zizmor.yml` at the repo root instead, since `persist-credentials: false` cannot be used here as credentials are needed later in the workflow.

This inline comment was also confusing dependabot when trying to bump the checkout action version.

Fixes #

**Checklist:**

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -20,7 +20,7 @@ jobs:
     if: github.repository == 'metal3-io/baremetal-operator'
     steps:
     - name: Checkout code
-      uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1 # zizmor: ignore[artipacked]
+      uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       with:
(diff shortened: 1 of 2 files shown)


tuminoid (commented):
Can we do this instead https://github.com/metal3-io/ironic-standalone-operator/pull/654 ?

lentzi90:
Oh, definitely!

lentzi90:
Hmm wait a sec, that is something else. We still need the checkout, right?

smoshiur1237 (commented):
/lgtm

lentzi90:
Yeah, just confirmed that we do it the same way in IrSO:
- https://github.com/metal3-io/ironic-standalone-operator/blob/c560d1e9ddcf3b5aaa0403c8312b74d86c940439/.github/workflows/release.yaml#L23-L26
- https://github.com/metal3-io/ironic-standalone-operator/blob/main/.zizmor.yml

kashifest (commented):
/approve

PR #3511, opened 2026-08-12
⚠️ Drop support for non-HFS firmware settings

dtantsur:
The Firmware field and the corresponding AccessDetails API are no longer
implemented for any drivers. Remove everything except for the API field
itself.

While I recognize that some downstream consumers might have drivers that
support this feature, maintaining it complicates the already complex
firmware management code without an ability to even test it e2e.
Such consumers should provide their own API on top of HFS instead.


lentzi90 (commented):
/lgtm

PR #3509, opened 2026-08-11
:seedling: Bumpi CAPI to 1.14.0, k8s.io to v0.36.3, and logr to v1.4.4

peppi-lotta:
Bumpi CAPI to 1.14.0, k8s.io to v0.36.3, and logr to v1.4.4

**Checklist:**

- [ ] Documentation has been updated, if necessary.
- [ ] Unit tests have been added, if necessary.
- [ ] E2E tests have been added, if necessary.
- [ ] Integration tests have been added, if necessary.

PR #3512, opened 2026-08-12
:seedling: Add delete permission for secrets in e2e role

peppi-lotta:
Add delete permission for secrets in e2e role

**Checklist:**

### DIFF
--- a/config/overlays/e2e/roles-rolebindings/roles-rolebindings.yaml
+++ b/config/overlays/e2e/roles-rolebindings/roles-rolebindings.yaml
@@ -16,6 +16,7 @@ rules:
   resources:
   - secrets
   verbs:
+  - delete
   - get
   - list
   - update


lentzi90 (commented):
/approve
For the record, this was a miss when restoring the permissions in https://github.com/metal3-io/baremetal-operator/pull/3310. The original PR (#3113) correctly handled the e2e manifests, but those where missed in the restore.
The new CAPI framework version enables the admission controller that detects these permission issues, as seen in #3509. This is why we need this PR. And it should also be backported all the way to 0.11. (See https://github.com/metal3-io/baremetal-operator/pull/3509#issuecomment-5264319223)

Sunnatillo (commented):
/lgtm

smoshiur1237 (commented):
/lgtm

PR #3478, opened 2026-08-03
🌱 Remove MariaDB support from BMO

migiFi:
Removes the MariaDB kustomize component and its wiring, following the mariadb-image archival and the removal of the `IRONIC_USE_MARIADB` option from  metal3-dev-env.

- delete ironic-deployment/components/mariadb and the basic-auth_tls_keepalived_mariadb overlay
- drop the `-m` flag and MariaDB handling from tools/deploy.sh and the hack/tools/deploy-cli tool + templates
- remove the set-manifest-image-mariadb Make target and the leftover `IRONIC_MARIADB_PASSWORD` in run_local_ironic.sh
- update ironic-deployment docs

The hardcoded default MariaDB ("changme") password shipped in the component's `secretGenerator` is removed with the component.

Depends on [#1710](https://github.com/metal3-io/metal3-dev-env/pull/1710) merging first, dev-env must stop passing -m before BMO drops it.

- [x] Documentation has been updated, if necessary.

PR #3532, opened 2026-08-17
🌱 Restore missing ironic entrypoint in run_local_ironic.sh

migiFi:
PR #3478 removed the MariaDB password line, which also carried `--entrypoint /bin/runironic`. Without it the ironic container fell back to `/bin/bash` and exited. This restores the entrypoint.


lentzi90 (commented):
/approve

PR #3541, opened 2026-08-19
🌱 Use IRSO integration for credentials in E2E tests 

### LINKED ISSUE #3331: Use IRSO integration for credentials in E2E tests
We currently configure credentials explicitly for BMO in the E2E tests. This can be seen for example here: https://github.com/metal3-io/baremetal-operator/blob/4c8b3540c2d8910ecb80a252ffca6d90b1513b6a/config/overlays/e2e/kustomization.yaml#L45-L56

BMO now has IRSO integration that makes it possible to just point to the Ironic instance instead. See [this kustomization](https://github.com/metal3-io/baremetal-operator/tree/main/config/use-irso) for how to do it.

Clean up all the e2e overlays under config/overlays to use this method instead of explicitly setting the credentials. Also remember to adjust the script that generates the credentials.

NOTE: Leave the Ironic configuration for now (under test/e2e/data/ironic-standalone-operator). IRSO could be used to create the credentials, but that can be done as a separate task.

UncivilizedGhost:
Remove ironic-credentials secretGenerator from BMO overlays since
credentials are now managed solely by IRSO.

Switch e2e overlays (e2e, release-0.12, release-0.13, upgrade) from
the to config/base and replace the hardcoded
IRONIC_ENDPOINT with IRONIC_NAME for Ironic CR.

Add a overlay granting BMO  RBAC to read the Ironic CR and its secrets
Have BMO wastch the samespace
Add DNS SAN to the Ironic TLS certificate, since BMO now connects by service DNS name
Trim whitespace from credentials

Simplify ci-e2e to only genrate credentials for release overlays
(as they have no IRONIC_NAME support)
Add a dedicated e2e-upgrade overlay for the upgrade tests since otherwise previous release's `ironic` ConfigMap are left in  place, which pin the upgraded BMO to the IRONIC_ENDPOINT path

Fixes https://github.com/metal3-io/baremetal-operator/issues/3331

**Checklist:**


config/overlays/e2e-release-0.12/ironic.env
```diff
@@ -1,3 +1,3 @@
+IRONIC_NAME=ironic
```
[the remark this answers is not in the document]
UncivilizedGhost: The upstream manifests don't consume IRONIC_ENDPOINT. The env var is read by the BMO binary at runtime.

hack/ci-e2e.sh
```diff
@@ -233,25 +233,17 @@ EOF
...
-for overlay in "${BMO_OVERLAYS[@]}"; do
-  echo "${IRONIC_USERNAME}" > "${overlay}/ironic-username"
-  echo "${IRONIC_PASSWORD}" > "${overlay}/ironic-password"
-done
+# release-0.11 still uses basic-auth and needs credential files in its overlay.
+echo "${IRONIC_USERNAME}" > "${REPO_ROOT}/config/overlays/e2e-release-0.11/ironic-username"
+echo "${IRONIC_PASSWORD}" > "${REPO_ROOT}/config/overlays/e2e-release-0.11/ironic-password"
 
 IRSO_IRONIC_AUTH_DIR="${REPO_ROOT}/test/e2e/data/ironic-standalone-operator/components/basic-auth"
 echo "${IRONIC_USERNAME}" > "${IRSO_IRONIC_AUTH_DIR}/ironic-username"
```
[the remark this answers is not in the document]
UncivilizedGhost: This was already being done for all overlays. I simply removed the newer versions.

config/overlays/e2e-upgrade/kustomization.yaml
```diff
@@ -1,16 +1,9 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
-# This overlay extends the e2e overlay with non-default Ironic ports
-# for use in upgrade tests. This allows upgrade tests to run on
-# a separate kind cluster without conflicting with the main e2e cluster.
+# This overlay extends the e2e overlay for use in upgrade tests.
+# This allows upgrade tests to run on a separate kind cluster
+# without conflicting with the main e2e cluster.
```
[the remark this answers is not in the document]
UncivilizedGhost: Added.

dtantsur:
`failed to list *v1alpha1.Ironic: ironics.ironic.metal3.io is forbidden` something off with RBAC?

UncivilizedGhost:
I believe the issue was BMO now needs to watch the Ironic CR which is in baremetal-operator-system but they aren't  any roles there. I added a role and a binding which should fix it.
**EDIT:** It didn't

UncivilizedGhost:
The issue seems to have been due to the credentials not being trimmed. It seems fixed now

pkg/provisioner/ironic/factory.go
lentzi90, line 1: We should not mix e2e config changes with actual code changes. The goal is to only change the e2e config and keep the behavior otherwise.

config/overlays/e2e-upgrade/kustomization.yaml
```diff
@@ -1,16 +1,14 @@
...
+# It is used to run upgrade tests on a separate kind cluster without
+# conflicting with the main e2e cluster. Conflict avoidance no longer
+# depends on overriding ironic endpoint/port settings in this overlay;
+# it relies on the inherited ironic service name/service discovery
+# configuration from the e2e overlay instead.
```
lentzi90: This means the overlay is also pretty useless, right? As long as we apply it in the correct cluster, it will discover the correct Ironic and the correct endpoint.
Perhaps it still makes sense to have it because of other settings though. The `e2e` overlays enables Ironic networking, and could have other new features included that we may not need/want for upgrades.
In any case, this comment should be updated to reflect this. We can also try to completely remove the upgrade overlay.

config/overlays/e2e/baremetal-operator-system/kustomization.yaml
lentzi90, line 1: This deserves a comment. If I understand correctly, we need this now because BMO needs to check secrets in `baremetal-operator-system` to get the credentials (and other things, e.g. the `Ironic` object). Previously the credentials and config were directly mounted on the pod.

lentzi90 (commented):
We need to run the optional tests in order to actually try the upgrade overlays.
Holding to make sure these tests pass before we merge.
/hold

config/overlays/e2e/baremetal-operator-system/kustomization.yaml
UncivilizedGhost: I added a small comment explaining it

config/overlays/e2e-upgrade/kustomization.yaml
UncivilizedGhost: I have removed it since it's probably cleaner. If we ever need it in the future it can be re-added.

pkg/provisioner/ironic/factory.go
UncivilizedGhost: Those changes are needed, otherwise the e2e tests fail. Is there another way I should add them?

lentzi90: Why are they needed? Why are the periodic tests on main (where we don't have this change) not failing?
This indicates a change in the tests. Either this PR discovered a bug in BMO that this is fixing, or the tests are now doing something "strange" that we are adapting to by changing BMO itself.

If it is a real bug, it is a good find! But then we need to understand it. Why did we miss it before? How are the tests here catching it?

If there is no bug and we are just blindly adapting the code to make the tests green, then we are doing things wrong.

UncivilizedGhost: I believe both changes were caused because I started using the ironic CR


Previously, ironic credentials were mounted at .../auth/ironic/ and read using the readAuthFile function  which used trimSpace to remove newlines.
With the ironic CR the credentials are fetched directly from kubernetes secret.  This means they aren't trimmed so I added trimSpace to avoid authentication issues. I think trimming is just good practice in general and this shouldn't cause any issues but I can trim them in ci-e2e script instead.



Previously, BMO used to read ironic endpoint from an env variable which had v1 appended to to the url. But now it makes the url from the name and namespace in the loadConfigFromIronicCR functions so I added the v1 there. 
Without the v1 this change seems to fail some tests. I force pushed the same commit a few times and got different results
- In one run the BMH provisioning test fails since it doesn't de-provision properly
- In another test [cleanup fails](https://github.com/metal3-io/baremetal-operator/actions/runs/33064142224/job/98490156006?pr=3541) due to provisioning failing
- Other runs pass with no issues (including the current one)
This seems be due to the whether a client is cached or not although I haven't investigated it deeply.

I think this v1 addition makes sense because of ironic endpoint was removed but if not I can try and pinpoint what's causing the issue

lentzi90: Ok, so the credentials we generate in ci-e2e probably contains newlines if I understand this correctly. I'd say the correct thing to do here then is to trim in ci-e2e.
Changing the endpoint does not make sense to me. I have used BMO with IrSO many times without this code change so it should be possible.

hack/ci-e2e.sh
```diff
@@ -233,29 +233,26 @@ EOF
...
 export IRONIC_PASSWORD
 
-for overlay in "${BMO_OVERLAYS[@]}"; do
-  echo "${IRONIC_USERNAME}" > "${overlay}/ironic-username"
-  echo "${IRONIC_PASSWORD}" > "${overlay}/ironic-password"
+# Release overlays still use basic-auth and need credential files.
+for overlay in \
+  "${REPO_ROOT}/config/overlays/e2e-release-0.11" \
+  "${REPO_ROOT}/config/overlays/e2e-release-0.12" \
+  "${REPO_ROOT}/config/overlays/e2e-release-0.13"; do
```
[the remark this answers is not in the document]
UncivilizedGhost: pr was outated. Changed it

config/overlays/e2e/ironic.env
```diff
@@ -1,4 +1,4 @@
-IRONIC_ENDPOINT=https://192.168.222.2:6385/v1/
+IRONIC_NAME=ironic
```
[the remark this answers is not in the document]
UncivilizedGhost: the v1 was a false alert

PR #3560, opened 2026-08-24
🐛 Set image_type in Ironic to avoid an unnecessary HEAD call

dtantsur:
When image_type is not set, the validation API ends up doing a HEAD call
to the image. This is not necessary in the Metal3 case since we don't
support partition images. Set image_type to "whole-disk" to disable this
path.

See https://bugs.launchpad.net/ironic/+bug/2164916


dtantsur:
/retest

IrSO timeout?

tuminoid (commented):
/lgtm

PR #3585, opened 2026-09-05
:bug: Build the manager and plugins with -trimpath

s3rj1k:
Source paths are part of the package fingerprint plugin.Open compares, so without -trimpath a plugin loads only when built from the same directory as the manager (/workspace for the release image), even when the toolchain and every shared module version already match.

Dockerfile.plugin-test now builds against BMO at /bmo-src rather than bmo-builder's /bmo, so it fails if -trimpath is dropped again.

### DIFF
--- a/.github/workflows/build-images-action.yml
+++ b/.github/workflows/build-images-action.yml
@@ -22,6 +22,7 @@ jobs:
     with:
       image-name: 'baremetal-operator'
       pushImage: true
+      image-build-args: MODULE_VERSION=${{ startsWith(github.ref, 'refs/tags/') && github.ref_name || '' }}
       generate-sbom: true
       sign-image: true
     secrets:
--- a/.github/workflows/plugin-test.yml
+++ b/.github/workflows/plugin-test.yml
@@ -7,9 +7,11 @@ on:
     - 'main'
     - 'release-*'
     paths:
+    - 'Dockerfile'
     - 'Dockerfile.plugin-test'
     - 'Makefile'
     - 'cmd/plugin-load-test/**'
+    - 'hack/plugin-build-flags.sh'
     - 'hack/plugin-test/**'
     - 'pkg/provisioner/**'
     - 'go.mod'
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -155,6 +155,7 @@ jobs:
       image-name: 'baremetal-operator'
       pushImage: true
       ref: ${{ needs.push_release_tags.outputs.release_tag }}
+      image-build-args: MODULE_VERSION=${{ needs.push_release_tags.outputs.release_tag }}
       generate-sbom: true
       sign-image: true
     secrets:
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,6 +8,12 @@ ARG BASE_IMAGE=gcr.io/distroless/base-debian13:nonroot@sha256:a557d784ac275c287d
 # this image to guarantee toolchain and module-version parity with BMO.
 FROM $BUILD_IMAGE AS sdk
 
+# Record our own packages as module@version, the form a dependency gets, so an
+# out-of-tree plugin can be built from any directory. Must match the Makefile.
+# See docs/plugin-provisioners.md.
+ARG MODULE_VERSION=v0.0.0-dev
+ENV MODULE_VERSION=${MODULE_VERSION}
+
 WORKDIR /workspace
 
 COPY go.mod go.sum ./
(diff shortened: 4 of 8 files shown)


pierrecregut:
This is quite a tricky MR. The only real issue is that it hides two commits that must be reviewed independently because they address different problems but only the second is referenced in the description of the merge request. But each commit message really explains its goal. 

For the commit related to the consistency of plugin version, check-submodule-version.sh does 3 things: first it checks that go.mod and test/go.mod are synced. Then it checks that the version referenced exist and if not just stop. This is important otherwise it is not possible to update to next version, finally it checks with the referenced version (and without replace) it is possible to recompile the provisioner. This exercise both apis and pkg/hardwareutils so it seems ok. Now I have two questions: 

* If as a developer  developing on main (last version is vN) I change let say apis, and the provisioner. The code is still referencing vN in go.mod and the provisioner recompilation should fail. So lint will fail. But I am just a dev, not a member of the release team, I do not know what vN+1 is, so to get through I must change go.mod and test/go.mod to an arbitrary non existing v-dev and for the plugin developers who want to consume directly this commit, this will not help them. Did I miss something ? So I am not really sure that developers should care about the version exposed in go.mod and so I am not sure that check-submodule-version.sh is required. 
* the second question is that it is quite involved and if somebody in a few months adds a new pkg/feature with its go.mod, are we confident that he will know what to update to cover his new module ? May be it could be useful to check in that the number of go.mod is always the same and otherwise print a warning pointing to an explanation stating how to change check-submodule-version, the relevant functions in verify-release.sh and if it is fine change the referenced number of go.mod (could be a pointer to a comment in check-submodule-version.sh

The added function in verify-release.sh ensures that if we are releasing vN, both go.mod and test/go.mod are referencing apis@vN and pkg/hardwareutils@vN and this is the important part. nit: the way the version referenced is extracted is different from the one used in check-submodule-version.sh.

s3rj1k:
> This is quite a tricky MR. The only real issue is that it hides two commits that must be reviewed independently because they address different problems but only the second is referenced in the description of the merge request. But each commit message really explains its goal.

Splitting into 2 PRs will ease review?

pierrecregut:
For the second commit, the core is plugin-build-flags.sh. In some sense it invert the replace (but also the absolute path of the bmo itself) and replace the file path with module versions. The script seems to be work as expected. On the CI, it is clear what MODULE_VERSION will be, this is less clear if I recompile locally and what this mean if I am developing a new plugin.

s3rj1k:
> if I am developing a new plugin

well, the suggest approach of developing **new** plugin is to first get it a working state in-tree (don't need to publish code, just get it working locally), when extract

pierrecregut:
So the answer to my last question is something like v0.14.0-rc.0-97-g34ebcf1c28165d1498e19396ea543f84b0e2efd9-dirty. But as all parts (core and plugins) use the same reference, it is probably fine for the plugin developer.

pierrecregut:
> Splitting into 2 PRs will ease review?

It would have. Comments if there are will be on two unrelated issues and the requirements synchronization lacks a proper description in the PR summary.

tuminoid:
/cc @dtantsur @lentzi90 @Rozzii 
More eyes on this.

pierrecregut:
> * Did I miss something ? So I am not really sure that developers should care about the version exposed in go.mod and so I am not sure that check-submodule-version.sh is required.

My comment was too long so I summarize.I did not find issues in the code itself but I still would like to understand the benefit of forcing the devs changing one of the submodule to change the versions. This is not their responsibility. If the team follows the additional guidance in verify-release, then the lint requirements are enforced at the time of the release and for versions with significant submodule modification between release, I do not think there is a simple answer (pin on commits sha ?). For such situations I would put the burden of adding replaces on the plugin developer.

s3rj1k:
> but I still would like to understand the benefit of forcing the devs changing one of the submodule to change the versions.

It is a common practice to have same tag version on submodules in monorepo Golang project, so it would be easier to consume those submodules.

(Same tag for all)

Of course we can still have different tags for different components and offload all this mess to plugin writer, but do consider that having same tag also simplifies CI.

pierrecregut:
I do agree with you on released version of bmo. But the problem arises when as a bmo dev, you change a submodule and use that change in the provisioner. Suddenly the lint rule will complain because the code does not compile with the previous released version of submodule  and will ask for a change of the submodule tags. I do not think that this is the developer problem to supply those new tags and there is not even a satisfactory solution. The plugin developer can wait for the new bmo release or use replace if he is impatient.

s3rj1k:
> because the code does not compile with the previous released version of submodule

I mean with go-plugin system we should keep in mind that this is not possible, meaning you can't use older submodule with newer BMO.

pierrecregut:
Here is a practical example. Starting from your commit. I add a type Foo to apis (probably for a new field in BMH spec). In provisioner I add a Foo field type metal3api.Foo to ManagementAccessDetail type. Now lint fails and tells me to bump apis and hardwareutils in go.mod and test/go.mod. What should I put ? v0.14.1 ? v0.15.0 ? Whatever I put does not really make sense and it is beyond my responsibility. Those kind of changes will be rare, but that is what the new linter catch and  they should not require to wait until the api is officially tagged to implement the corresponding behavior in the provisioner.

s3rj1k:
> Here is a practical example. Starting from your commit. I add a type Foo to apis (probably for a new field in BMH spec). In provisioner I add a Foo field type metal3api.Foo to ManagementAccessDetail type. Now lint fails and tells me to bump apis and hardwareutils in go.mod and test/go.mod. What should I put ? v0.14.1 ? v0.15.0 ? Whatever I put does not really make sense and it is beyond my responsibility. Those kind of changes will be rare, but that is what the new linter catch and they should not require to wait until the api is officially tagged to implement the corresponding behavior in the provisioner.

Do consider that provisioner as an Out-of-tree should still target only released BMO versions, so this inconvenience will not be a problem for Out-of-tree plugin, this is only a problem for development.

Open to suggestions on how to resolve all of this.

pierrecregut:
Then as out of tree plugin target released bmo versions, it means that the lint check is useless and worse counter-productive. For released versions, verify-release guarantee that the provisioner will compile because on that commit filepath and tag will exactly coincide.

s3rj1k:
@pierrecregut Dropped intermediate commit, will be a standalone PR

https://github.com/metal3-io/baremetal-operator/pull/3618

pierrecregut:
The remaining part looks sound. nit: may be document the exact compiler options you are adding with the shell script: a global -trimpath and a trimpath rewriting specifications for go tool compile  replacing the absolute paths of each module with their equivalent module descriptor. The replacement is limited to github.com/metal3-io/baremetal-operator/ subpackage by a prefix for gcflags. Is it required ?

/lgtm

PR #3586, opened 2026-09-05
🌱 Add `Anaconda` provisioner plugin

s3rj1k:
Adds the anaconda provisioner as an out-of-tree plugin exemplar under `test/anaconda`, plus an e2e test.

Anaconda installs a host by booting an installer ISO over Redfish virtual media and serving it a per host kickstart. No Ironic, no agent in the host, no Inspection.

Backport of https://github.com/s3rj1k/metal3-baremetal-operator-oot/blob/anaconda_minimal/out-of-tree/README.md


s3rj1k:
https://github.com/user-attachments/assets/87e6c72f-8034-40a5-99c1-c8367c04c6e8

dtantsur (changes requested):
The whole point of the plugin interface was that we don't add all possible provisioners into BMO code base. We can discuss have a separate repo along the lines of metal3-io/contrib-provisioners, but I don't want us to become responsible for even more code that most of us don't use or test.

s3rj1k:
> We can discuss have a separate repo along the lines of metal3-io/contrib-provisioners

Sure, I see no problem with that, until we have a decision how we want to handle this, and discussion how we should implement CI for this out-of-tree repo, let's keep this PR as a working reference

Rozzii:
> The whole point of the plugin interface was that we don't add all possible provisioners into BMO code base. We can discuss have a separate repo along the lines of metal3-io/contrib-provisioners, but I don't want us to become responsible for even more code that most of us don't use or test.

I think the idea is quite nice that we would test the plugin interface with one or more actual plugin(s), but I would move the plugin internal code & docker image etc... to utility-images same as we did with fake IPA and other test/specialized tools. And we could provide a "non voting" periodic test for testing with this plugin.

Or we could have ofc a separate repo just for provisioner plugins I am okay with that too.

s3rj1k:
> but I would move the plugin internal code

one benefit of this is simpler CI, so we keep CI things here and just move the actual code outside.

s3rj1k:
> PR needs rebase.

Holding off rebase until https://github.com/metal3-io/baremetal-operator/pull/3585 lands

PR #3594, opened 2026-09-07
:seedling: Add dataimage attach e2e test

hroyrh:
Extend the dataimage e2e test suite by adding dataimage attach test.
 
Assisted-By: Claude enterprise <noreply@anthropic.com>

### DIFF
--- a/test/e2e/dataimage_test.go
+++ b/test/e2e/dataimage_test.go
@@ -15,6 +15,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/cluster-api/test/framework"
+	"sigs.k8s.io/cluster-api/util/patch"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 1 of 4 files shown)


test/e2e/dataimage_test.go
```diff
@@ -307,6 +308,127 @@ var _ = Describe("DataImage Finalizer", Label("required", "dataimage"), func() {
 		}, "20m", "10s").Should(BeTrue())
 	})
 
+	It("should attach a data image to a provisioned host", func() {
+		bmhName := specName + "-attach"
+		secretName := bmhName + "-bmc-creds"
+
```
dtantsur, line 314: Please skip this test on non-vmedia drivers

dtantsur: You can check `UsesIPXE` in `common.go`

test/e2e/dataimage_test.go
```diff
@@ -307,6 +308,127 @@ var _ = Describe("DataImage Finalizer", Label("required", "dataimage"), func() {
 		}, "20m", "10s").Should(BeTrue())
 	})
 
+	It("should attach a data image to a provisioned host", func() {
```
dtantsur, line 311: Why cannot we expand the existing test instead?

test/e2e/dataimage_test.go
```diff
@@ -307,6 +308,127 @@ var _ = Describe("DataImage Finalizer", Label("required", "dataimage"), func() {
...
+		}, e2eConfig.GetIntervals(specName, "wait-power-state")...)
+
+		By("Powering on the BMH to trigger DataImage attachment")
+		err = clusterProxy.GetClient().Get(ctx, types.NamespacedName{Name: bmhName, Namespace: namespace.Name}, &bmh)
+		Expect(err).NotTo(HaveOccurred())
+		helper, err = patch.NewHelper(&bmh, clusterProxy.GetClient())
+		Expect(err).NotTo(HaveOccurred())
+		bmh.Spec.Online = true
+		Expect(helper.Patch(ctx, &bmh)).To(Succeed())
+
```
dtantsur, line 416: Maybe SSH into the machine and check that it's actually attached?

test/e2e/dataimage_test.go
```diff
@@ -307,6 +308,127 @@ var _ = Describe("DataImage Finalizer", Label("required", "dataimage"), func() {
...
+		By("Creating a DataImage with the same name as the BMH")
+		dataImageURL := e2eConfig.GetVariable("ISO_IMAGE_URL")
+		di := &metal3api.DataImage{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      bmhName,
+				Namespace: namespace.Name,
+			},
+			Spec: metal3api.DataImageSpec{
+				URL: dataImageURL,
```
dtantsur, line 378: I don't think you should rely on the fake URL in this test. Maybe check what live ISO tests are using?

PR #3614, opened 2026-09-09
🌱Add id-token permission to build-images-pr workflow

smoshiur1237:
https://github.com/metal3-io/cluster-api-provider-metal3/issues/3701
Add id-token to fix the workflow

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -13,6 +13,7 @@ on:
 
 permissions:
   contents: read
+  id-token: write
 
 jobs:
   build:


lentzi90 (commented):
This is not needed. The workflow is already working.
https://github.com/metal3-io/baremetal-operator/actions/workflows/build-pr.yml
But perhaps it should be migrated to use the project-infra workflow? Then it will need the id-token permission.

smoshiur1237:
BMO is not using the workflow from project-infra. So closing this one

PR #3617, opened 2026-09-09
🐛 Avoid returning RequeueAfter=0

dtantsur:
Apparently, the deprecated Requeue flag does not do anything any more,
so returning RequeueAfter 0 means no requeue. This can block deletion,
for example.

This commit:
- changes Result in the controller to never return requeue delay of 0 if
  requeue is expected and drops the deprecated Requeue flag
- changes deletion in the Ironic backend to not requeue (deletions are
  synchronous)
- changes automated clean sync to not requeue (patching nodes is
  synchronous too)

I've left TODOs for revisiting power and maintenance actions since
changing them requires modifying a lot of caller places and tests.

### DIFF
--- a/internal/controller/metal3.io/action_result.go
+++ b/internal/controller/metal3.io/action_result.go
@@ -18,6 +18,8 @@ const (
 	defaultBackoff  = 0.5
 )
 
+var tinyDelay = 10 * time.Millisecond
+
 // actionResult is an interface that encapsulates the result of a Reconcile
 // call, as returned by the action corresponding to the current state.
 type actionResult interface {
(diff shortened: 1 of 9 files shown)


dtantsur:
/hold

I'll test if this fixes my issue, this will take some time.

dtantsur:
/retest

SSH timeout in servicing, presumably unrelated

lentzi90 (commented):
> Apparently, the deprecated Requeue flag does not do anything any more,
so returning RequeueAfter 0 means no requeue. This can block deletion,
for example.

Where did you get this from? As far as I could tell it is just deprecated with functionality remaining.
That said, I think it is correct to go this way of having RequeueAfter > 0 in all places. Seems to be the preferred way.

/approve

dtantsur:
> Where did you get this from? As far as I could tell it is just deprecated with functionality remaining.

I don't see any clear message on this. But I got a fresh OpenShift cluster, where a host would get stuck in "deleting" state with no conductor actions whatsoever after Ironic reports that the node is deleted. When I touch the host (modify anything), it's immediately gone. The only conclusion I can make is that rescheduling did not work. Could be something with OpenShift. But we'll get into this situation eventually anyway.

dtantsur:
/hold cancel

Works for me. Meanwhile, I was able to reproduce the problem with upstream bits too.

iurygregory (commented):
/lgtm

PR #3618, opened 2026-09-09
:sparkles: Require the released apis and hardwareutils versions

s3rj1k:
The root `go.mod` and `test/go.mod` require `apis` and `pkg/hardwareutils` at
v0.5.1, hidden by the local `replace` directives. Consumers get no replace, so
they resolve v0.5.1 and `pkg/provisioner` does not compile against it.

Both files move to v0.14.0. Lint builds `pkg/provisioner` the way a consumer
resolves it, and `verify-release.sh` fails unless both files name the version
being tagged.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -181,6 +181,7 @@ lint: $(GOLANGCI_LINT)
 	cd pkg/hardwareutils; $(GOLANGCI_LINT) run -v $(GOLANGCI_LINT_EXTRA_ARGS) ./... --timeout=10m
 	cd hack/tools; $(GOLANGCI_LINT) run -v $(GOLANGCI_LINT_EXTRA_ARGS) ./... --timeout=10m
 	./hack/check-e2e.sh
+	./hack/check-submodule-versions.sh
 
 .PHONY: lint-fix
 lint-fix: $(GOLANGCI_LINT) ## Lint the codebase and run auto-fixers if supported by the linter
--- a/docs/releasing.md
+++ b/docs/releasing.md
@@ -69,6 +69,30 @@ by using following command `git remote -v`.
 - Fetch the remote (`metal3-io`): `git fetch upstream`
 This makes sure that all the tags are accessible.
 
+### Bumping the submodule requirements
+
+The root `go.mod` and `test/go.mod` require `apis` and `pkg/hardwareutils` at
+a version that their `replace` directives hide, so only consumers ever resolve
+it. Release tagging creates `v0.x.y`, `apis/v0.x.y`, `test/v0.x.y` and
+`pkg/hardwareutils/v0.x.y` from one commit, so both files have to require
+`v0.x.y` before the release notes are merged. Otherwise the released root
+module points consumers at submodule code that no released submodule version
+contains, and an out of tree plugin built against it hits a package
+fingerprint mismatch. See [plugin provisioners](plugin-provisioners.md).
+
+- Bump the two requirements in `go.mod` and `test/go.mod` to the version you
+   are about to release, in a pull request of its own, and merge it first. The
+   release notes commit has to contain the notes file and nothing else, so the
+   bump cannot ride along with it.
+
+- Until those tags exist, `hack/check-submodule-versions.sh` reports the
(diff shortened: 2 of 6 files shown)


pierrecregut:
As this commit was initially part of another PR #3585, Here is a practical example of my main concern: https://github.com/metal3-io/baremetal-operator/pull/3585#issuecomment-5601789230 . I will complete my analysis. Let say I chose v0.14.1 and my change is committed to main, but my modification is not included in the patch release (kept for v0.15.0). When the release team creates v0.14.1, the behavior of the linter will change. It will try to check that provisioner can really be compiled with v0.14.1 and as Foo is not defined in it, the CI will suddenly be stuck for everybody. 

verify-release.sh looks good to me, but not the lint part.

s3rj1k:
@pierrecregut How do you propose running "./hack/check-submodule-versions.sh", manually?

>  It will try to check that provisioner can really be compiled with v0.14.1 and as Foo is not defined in it, the CI will suddenly be stuck for everybody.

Change is now independent of provisioner interface

pierrecregut:
The only place where running check-submodule-version.sh makes sense is on the commit of a newly tagged release. The bad news is that it is already too late to correct anything. The good news is that if verify-release recommendation are followed there is no risk it fails. So I see no reason for running this script. 

One of the consequence that you pointed out is that out of tree plugins can only compile against tagged BMO release. So I would not replace v0.5.2 with v0.14.0 but rather with v0.0.0 on main to make this more obvious.

s3rj1k:
> The bad news is that it is already too late to correct anything

Why? v0.14 is last BMO release?

> One of the consequence that you pointed out is that out of tree plugins can only compile against tagged BMO release. 

Should, not `can`, they can be compiled to any hash, it just does not make sense to do so from an SBOM point-of-view. I want to use exactly verifiable by-hash BMO binary with my custom `.so` file, but I can use technically any commit of BMO I would need to build that binary myself in that case.

dtantsur:
> So I would not replace v0.5.2 with v0.14.0 but rather with v0.0.0 on main to make this more obvious.

I agree. In fact, the dependency here is not a normal dependency. If you keep v0.14.0 in place, then as main moves on, it gets broken again (because realistically we depend on whatever is in the tree right now).

I think we need to document that downstream plugins must use the same replacements as BMO.

s3rj1k:
> I think we need to document that downstream plugins must use the same replacements as BMO.

PR is independent of plugin system, and that is already documented in other PR.

What we need to do, if syncing submodule versions to module versions is unacceptable, is to document how exactly we should bump versions of submodules.

If we continue in-spirit the talk we had in recent community meeting, we need to establish a document that says that:

- we forbid to change submodule API version in-between BMO `patch` versions
- when we need to update submodule API contract - it directly means `minor` version BMO bump
- for each `minor` version BMO bump we bump `minor` version of all submodules
- if and only if submodule API contract change is an addition (new field), that is proven to be stable for all `minor` BMO versions, we can allow patch level submodule API bump, but this should be threated as exception to a rule, we generaly should prefer bumping versions with new BMO `minor` releases.

---

v0.5.2 -> v0.0.0 - IMO: this communicates to community that there is zero API stability guarantee, it would have been ok if we did not already migrate to tagging submodules

pierrecregut:
Could the check of coherence in verify-release.sh be added to .github/workflows/release.yaml somewhere between "Create or checkout release branch" and "Create Release Tag" ? This way it would be not only suggested but really enforced.

s3rj1k:
> Could the check of coherence in verify-release.sh be added to .github/workflows/release.yaml somewhere between "Create or checkout release branch" and "Create Release Tag" ? This way it would be not only suggested but really enforced.

I think there is no consensus, we need to produce governance docs on how we want to enforce submodule versioning.

dtantsur:
Okay, I see what you mean. If we don't care about the skew between tags, then I have one point only:

> we forbid to change submodule API version in-between BMO patch versions

This is the only part I'm not 100% comfortable with. But I guess we don't change the API often.

I also hope that the replacement will allow us to bump versions to ones that don't exist yet (bumping to 0.15 will happen right before 0.15 is tagged).

Issue #3578, opened 2026-09-03
E2E IPv6 only testing


nuhakala:
Currently all e2e testing is done over IPv4. All the components support IPv6, and there has been users who want at least dual-stack. Adding support for IPv6 only tests will help us to catch IPv6 related issues easier. Testing IPv6 only will also ensure that everything work with IPv6, we could do dual-stack testing, but the risk there is that IPv6 is not really used even though it is enabled, hence IPv6 only is simpler.

PR #3620, opened 2026-09-10
:seedling: Add support for IPv6 only e2e setup

nuhakala:

This PR adds support for IPv6 only environment to run the e2e tests in. It is not yet used anywhere, but the goal is to eventually test IPv6 in CI as well.

- Turns BMC and e2econfigs into templates, because IP addresses need to be injected into them.
- Also creates templates from IrSO and Ironic manifests.
- Adds support for handling IPv6 addresses in the tests.
- Collects all addresses into ack/e2e/ip_addressing.sh along with IPv6 only related options.
- Removes `kindExtraPortMappings` as they are not needed anymore

**Please note that the changes require support from IPA image as well.** Upstream IPA image _disables DHCPv6 by default_, which prevents IPA image from getting an IPv6 address. This requires IPA image built with `DIB_SIMPLE_INIT_NO_DHCP_FALLBACK=1`. This option disables Glean from falling back to DHCP if no explicit network configuration is provided (and hence disabling DHCPv6). Disabling DHCP fallback causes Glean to not generate ifcfg-rh network script which disables DHCPv6, instead the image will fall back to default NetworkManager configuration which has DHCPv6 enabled in that case.

This problem with IPA image need to be solved before we can start running this in CI.

Related issue: #3578

**Checklist:**


hack/e2e/ip_addressing.sh
```diff
@@ -0,0 +1,86 @@
...
+	test/e2e/data/ironic-standalone-operator/operator/irso-v0.11/kustomization.yaml
+	test/e2e/config/bmcs-fixture.yaml
+	test/e2e/config/bmcs-ipmi.yaml
+	test/e2e/config/bmcs-redfish-virtualmedia.yaml
+	test/e2e/config/bmcs-redfish.yaml
+	config/overlays/e2e/ironic.env
+)
+
+for tmpl in "${templates[@]}"; do
+	envsubst < "${REPO_ROOT}/${tmpl}.tmpl" > "${REPO_ROOT}/${tmpl}"
```
terror96: nit: in order to avoid unexpected `envsubst` replaces, consider listing keywords that are expected to be replaced by the `envsubst` command instead of replacing all environment variables i.e.

`envsubst '${HOST_ADDRESS},${VM_HOST1_IP},...' < "${REPO_ROOT}/${tmpl}.tmpl" > "${REPO_ROOT}/${tmpl}"`

it will be somewhat a lengthy list because it is a union of all variables present in the templates, but it prevents accidental replacements of variables that should not be replaced at this point (if not applicable now, could be the case in the future).

hack/ci-e2e.sh
```diff
@@ -258,7 +257,7 @@ echo "${IRONIC_USERNAME}" > "${IRSO_IRONIC_AUTH_DIR}/ironic-username"
 echo "${IRONIC_PASSWORD}" > "${IRSO_IRONIC_AUTH_DIR}/ironic-password"
 
 # shellcheck disable=SC2016
-SSH_PUB_KEY_CONTENT="${pub_ssh_key}" envsubst '${SSH_PUB_KEY_CONTENT}' < \
+SSH_PUB_KEY_CONTENT="${pub_ssh_key}" envsubst < \
```
terror96: Is there a reason why `'${SSH_PUB_KEY_CONTENT}'` is removed?

hack/ci-e2e.sh
```diff
@@ -137,7 +136,7 @@ if [[ ! -f "${IMAGE_DIR}/${IPA_FILE}" ]]; then
 fi
 
 # shellcheck disable=SC2016
-envsubst '${BMO_E2E_EMULATOR},${IP_ADDRESS},${BMO_E2E_IMAGE},${BMO_E2E_LISTEN_PORT},${IMAGE_DIR}' < \
+envsubst '${BMO_E2E_EMULATOR},${IP_ADDRESS},${BMO_E2E_IMAGE},${BMO_E2E_LISTEN_PORT},${IMAGE_DIR},${SUBNET_MASK}' < \
```
terror96, line 139: Just a note for others: `vbmctl.yaml.tmpl` still uses `${IP_ADDRESS}` instead of `${HOST_ADDRESS}`, and this is correct.

test/e2e/config/fixture.yaml.tmpl
```diff
@@ -29,15 +29,16 @@ variables:
   # Since there is no ironic involved, we do not risk having any port collisions.
   UPGRADE_USE_EXISTING_CLUSTER: "false"
 
-  IMAGE_URL: "http://192.168.222.1/cirros-0.6.2-x86_64-disk.img"
-  ISO_IMAGE_URL: "http://192.168.222.1/cirros.iso"
+  IMAGE_URL: "http://${IP_ADDRESS}/cirros-0.6.2-x86_64-disk.img"
+  ISO_IMAGE_URL: "http://${IP_ADDRESS}/minimal_linux_live-v2.iso"
```
terror96: Should we be using brackets here i.e. `${HOST_ADDRESS}` ?

test/e2e/config/ironic.yaml.tmpl
```diff
@@ -45,25 +45,26 @@ variables:
   # This allows upgrade tests to run alongside other tests, but requires a separate
   # provisioning IP (see UPGRADE_IRONIC_PROVISIONING_IP).
   UPGRADE_USE_EXISTING_CLUSTER: "false"
-  UPGRADE_IRONIC_PROVISIONING_IP: "192.168.222.3"
+  UPGRADE_IRONIC_PROVISIONING_IP: "${UPGRADE_IRONIC_PROVISIONING_IP}"
 
-  IMAGE_URL: "http://192.168.222.1/cirros-0.6.2-x86_64-disk.img"
-  ISO_IMAGE_URL: "http://192.168.222.1/minimal_linux_live-v2.iso"
+  IMAGE_URL: "http://${IP_ADDRESS}/cirros-0.6.2-x86_64-disk.img"
+  ISO_IMAGE_URL: "http://${IP_ADDRESS}/minimal_linux_live-v2.iso"
```
terror96: Should we be using brackets here i.e. ${HOST_ADDRESS} ?

test/e2e/data/ironic-standalone-operator/ironic/base/ironic.yaml.tmpl
```diff
@@ -7,16 +7,16 @@ spec:
...
+      - "00:60:2f:31:81:01,${VM_HOST1_HOST}"
+      - "00:60:2f:31:81:02,${VM_HOST2_HOST}"
     interface: "eth0"
-    ipAddress: "192.168.222.2"
+    ipAddress: "${IRONIC_PROVISIONING_IP}"
     ipAddressManager: "keepalived"
   deployRamdisk:
     sshKey: "${SSH_PUB_KEY_CONTENT}"
-    extraKernelParams: "console=ttyS0 bmo-e2e-kernel-params-ok"
+    extraKernelParams: "console=ttyS0 bmo-e2e-kernel-params-ok sshkey=\"${SSH_PUB_KEY_CONTENT}\""
```
terror96, line 26: It would be good to add a comment here that this improves debugging.

test/e2e/data/ironic-standalone-operator/operator/irso-v0.10/kustomization.yaml.tmpl
```diff
@@ -11,7 +11,7 @@ generatorOptions:
 configMapGenerator:
 - name: ironic-operator-config
   literals:
-  - IPA_BASEURI=http://192.168.222.1
+  - IPA_BASEURI=${HOST_ADDRESS}
```
terror96: This omits `http://`.

test/e2e/data/ironic-standalone-operator/operator/irso-v0.11/kustomization.yaml.tmpl
```diff
@@ -11,7 +11,7 @@ generatorOptions:
 configMapGenerator:
 - name: ironic-operator-config
   literals:
-  - IPA_BASEURI=http://192.168.222.1
+  - IPA_BASEURI=${HOST_ADDRESS}
```
terror96: This omits `http://`.

terror96 (commented):
I have only those a few small comments, but otherwise this lgtm. I'll have another look after a new push.

hack/ci-e2e.sh
nuhakala: The reason was that also the IP addresses needed to be substituted there, and that would have resulted in long list of vars. I refactored this and added this back.

In relation to the above comment of adding the variable list into `ip_addressing.sh`: I edited this so that `ip_addressing.sh` injects ip addresses to every file where they are needed and it also specifies variable list to inject only the IP addresses. Hence we can only inject the SSH key here, but because the corresponding file had already been injected, needed to move it and call envsubst to the moved file. Added also a comment explaining this.

hack/e2e/ip_addressing.sh
nuhakala: Added this

test/e2e/config/fixture.yaml.tmpl
nuhakala: fixed

test/e2e/config/ironic.yaml.tmpl
nuhakala: fixed

test/e2e/data/ironic-standalone-operator/ironic/base/ironic.yaml.tmpl
nuhakala: added comment

test/e2e/data/ironic-standalone-operator/operator/irso-v0.10/kustomization.yaml.tmpl
nuhakala: fixed

test/e2e/data/ironic-standalone-operator/operator/irso-v0.11/kustomization.yaml.tmpl
nuhakala: fixed

hack/ci-e2e.sh
```diff
@@ -63,6 +63,9 @@ export PATH="/usr/local/go/bin:${PATH}"
 # CAPI test framework uses kubectl in the background
 "${REPO_ROOT}/hack/e2e/ensure_kubectl.sh"
 "${REPO_ROOT}/hack/e2e/ensure_yq.sh"
+# Source the correct address set
+export USE_IPV6="${USE_IPV6:-false}"
+. "${REPO_ROOT}/hack/e2e/ip_addressing.sh"
 
 sudo apt-get update
 sudo apt-get install -y libvirt-dev pkg-config gettext-base curl
```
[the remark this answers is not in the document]
terror96: This is a valid finding.

hack/e2e/ip_addressing.sh
```diff
@@ -0,0 +1,94 @@
+#!/usr/bin/env bash
+
+# -----------------------------------------------------------------------------
+# Description: IPv4 and IPv6 specific setups including IP addresses. This file
+#              collects all IP-addresses and templates them into configuration
+#              files that are only missing these IP addresses.
+# Usage:       The file can be executed or sourced.
+# -----------------------------------------------------------------------------
+
+set -eu
```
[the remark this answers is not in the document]
terror96: I don't think it is needed: the script does not currently use pipes and also not all existing scripts use it either.

test/e2e/network_data_test.go
```diff
@@ -28,11 +28,17 @@ const networkDataTemplate = `{
   "networks": [
     {
       "id": "network0",
+      "link": "iface1",
+      "type": "ipv4_dhcp",
+      "network_id": "test-pp-network_dummy"
+    },
+    {
```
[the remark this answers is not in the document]
terror96: I think the purpose is to create a non-working interface, because Glean needs an IPv4 interface/block but it does not have to work. Thus, this is fine. Maybe comment this?

hack/ci-e2e.sh
```diff
@@ -257,9 +256,14 @@ IRSO_IRONIC_AUTH_DIR="${REPO_ROOT}/test/e2e/data/ironic-standalone-operator/comp
...
 
+# NOTE(nuhakala): ironic.yaml has already been once envsubstituted when
+# injecting IP addresses, but the SSH key was not injected. Hence, first move
+# the file under temp so that we can create it again, now fully substituted.
+# Doing in-place substitution can be buggy using shell redirection/pipe.
+mv "${REPO_ROOT}/test/e2e/data/ironic-standalone-operator/ironic/base/ironic.yaml" /tmp/ironic-base.yaml.tmpl
 # shellcheck disable=SC2016
 SSH_PUB_KEY_CONTENT="${pub_ssh_key}" envsubst '${SSH_PUB_KEY_CONTENT}' < \
-  "${REPO_ROOT}/test/e2e/data/ironic-standalone-operator/ironic/base/ironic.yaml.tmpl" > \
+  "/tmp/ironic-base.yaml.tmpl" > \
```
terror96, line 266: I also think this is a better approach because this way we can keep all address relates substitutions in `ip_addressing.sh` which is also the file where the environmental variables used in the process are defined. `SSH_PUB_KEY_CONTENT` is defined in this file which make this substitution logical here.

Makefile
```diff
@@ -150,6 +151,11 @@ verify-e2e-prerequisites: ## Check that required tools exist for e2e tests
 	@echo "Ensure the local environment is ready for e2e tests..."
 	VERIFY_ONLY=1 ./hack/e2e/ensure_e2e_prerequisites.sh
 
+.PHONY: template-addresses
+template-addresses: ## Put IP addresses into templates
+	@echo "Insert IP addresses into configuration..."
+	./hack/e2e/ip_addressing.sh
```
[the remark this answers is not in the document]
terror96: This is a valid finding.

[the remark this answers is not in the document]
nuhakala: I actually added the makefile target only because the fixture test was not invoking the ci-e2e script and hence the IP addresses were not injected there. I think the makefile target `test-e2e` is supposed to be called from the ci-e2e script and not directly, so then this would not be a problem? @lentzi90 
However, maybe it would be a good idea to refactor it regardless so that the makefile target test-e2e calls the ip_addressing anyway.

test/e2e/e2e_suite_test.go
```diff
@@ -91,6 +91,7 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 			Name:              "bmo-e2e",
 			Images:            e2eConfig.GetClusterctlImages(),
 			ExtraPortMappings: e2eConfig.KindExtraPortMappings,
+			IPFamily:          e2eConfig.GetVariable("IP_FAMILY"),
```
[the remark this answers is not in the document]
nuhakala: This is good point, I intentionally omitted everything related to the upgrade tests because I thought that it would be enough to get this working with basic tests only. Any opinions @lentzi90

Makefile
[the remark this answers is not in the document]
terror96: > I think the makefile target test-e2e is supposed to be called from the ci-e2e script and not directly, so then this would not be a problem?

We also do mention `make test-e2e` in https://github.com/Nordix/baremetal-operator/blob/11b3d01c2ed4453353627ef4d5b5062a98dbdf9f/test/e2e/README.md.

btw. are clean target and scripts in sync with these changes?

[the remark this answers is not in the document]
nuhakala: This could be done so that before substituting the variables we check if the .yaml file already exists and use that as a template if it does. But the downside of this is that the values won't be updated on subsequent runs. Would not matter in the CI but if running tests manually and the IP version needs to be changed, it would be annoying to start deleting/overwriting the files.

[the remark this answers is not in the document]
nuhakala: > > I think the makefile target test-e2e is supposed to be called from the ci-e2e script and not directly, so then this would not be a problem?
> 
> We also do mention `make test-e2e` in https://github.com/Nordix/baremetal-operator/blob/11b3d01c2ed4453353627ef4d5b5062a98dbdf9f/test/e2e/README.md.
> 
> btw. are clean target and scripts in sync with these changes?

Hmm, do we need to edit the cleaning as well? Maybe if we want to remove the rendered yaml files but I guess that is not necessary?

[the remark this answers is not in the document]
terror96: Cleaning is also mentioned in the documentation as an action to take between test runs...
`After the tests are run, please ensure proper cleanup before running them again. The due process for ensuring all is clean for the next run is (in the root directory)`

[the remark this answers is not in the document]
terror96: >  But the downside of this is that the values won't be updated on subsequent runs. Would not matter in the CI but if running tests manually and the IP version needs to be changed, it would be annoying to start deleting/overwriting the files.

Generally I would prefer that the procedure (whatever it requires) is documented even though it would contain inconvenient steps.

hack/ci-e2e.sh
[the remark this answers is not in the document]
nuhakala: fixed

Makefile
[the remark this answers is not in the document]
nuhakala: Added partial envsubst with which we can inject new variables to the templates without overriding the old ones. Also, added cleanup for the configs which the `clean-e2e.sh` will call. Now injecting the IP addresses to templates is also done automatically when running `make teste-e2e` so that the configs are generated for fixture tests without any extra steps.

test/e2e/network_data_test.go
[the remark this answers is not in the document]
nuhakala: Added comment there.

PR #3621, opened 2026-09-10
:seedling: Feature gate for HostClaims

pierrecregut:

Hide Hostclaims implementation behind a feature gate:

* disabled by default (alpha)
* HostClaim controller disabled unless the feature is enabled. In that case, it can still be disabled with --hostclaims=false (which can be useful if only the part in bmh controller or the webhooks are wanted).

### DIFF
--- a/config/base/manager.yaml
+++ b/config/base/manager.yaml
@@ -23,7 +23,6 @@ spec:
         - /baremetal-operator
         args:
         - --enable-leader-election
-        - --hostclaims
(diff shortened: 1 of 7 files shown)


pierrecregut:
/cc @dtantsur 
Is it what you had in mind for the actual implementation of the gates ?

* webhook disabling CR creation and controller.
*  Even if the gate is enabled it is possible to disable the controller.

PR #3638, opened 2026-09-15
🌱 Fix zizmor self-repository findings and bump zizmor-action to v0.6.3

### LINKED ISSUE #3635: Update workflows to fix new zizmor findings
See: https://github.com/metal3-io/baremetal-operator/pull/3603

Zizmor fails because it now has some new rules that we do not follow.
Update the workflows so that it passes. You need to update zizmor in the same PR to see that it really works.

aryanpatel99:

Fixes the 8 new `self-repository` findings introduced by zizmor 1.30.0 (shipped with `zizmor-action` v0.6.3).
The new [`self-repository` audit rule](https://docs.zizmor.sh/audits/#self-repository) recommends using GitHub's dedicated self-repository syntax (`$/...`) instead of relative paths (`./...`) when referencing reusable workflows in the same repository.

Fixes #3635

**Checklist:**


aryanpatel99:
This is ready for review whenever someone has a moment.

lentzi90 (commented):
/ok-to-test
/approve

smoshiur1237 (commented):
Thanks for the fix
/lgtm

PR #3639, opened 2026-09-16
🌱 e2e: test re-running fast inspection

dtantsur:


pierrecregut:
The code is fine. I have only a small concern: inspection.go and re_inspection.go have similar topic. inspection now completely covers fast inspection tests (first inspection and reinspection) but only first inspection for standard one. This is done by re_inspection with a test that also checks that it corrects mistakes rather than recreate the hardwareData. There is nothing wrong, just a little confusing.
/lgtm

dtantsur:
> inspection.go and re_inspection.go have similar topic

Yeah, it's a weird split. Maybe we should merge them eventually.

MahnoorAsghar:
lgtm

PR #3648, opened 2026-09-17
🌱 Avoid Nordix Artifactory

lentzi90:

Nordix Artifactory is currently down. While we wait for it to be fixed, let's avoid using it.

Fixes #

**Checklist:**


lentzi90:
This (artifactory down) also highlighted an issue with the fallback. It lacks timeout setting so it currently never proceeds to the fallback since the connection to artifactory is never established.
Once the situation is sorted out I think we should add a `--connect-timeout=10` or similar.

For now we avoid it completely though, since we don't know when or in what form it will be back.

kashifest (commented):
/approve

nuhakala (commented):
/lgtm

